PR #2, opened 2022-04-09
Update pr-validation job abd tycho scm parameters

sravanlakkimsetti:
Fixes #1

PR #4, opened 2022-04-09
[RelEng] Perform code freeze period check in GitHub workflow

HannesWell:
Perform the code freeze period check in a dedicated reusable workflow instead of in the Jenkins-Build-Pipeline.

The workflow being called was created in:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/206

- [x] GH-Actions have to be enabled for this repository

The currently pending PR #2 adds a Jenkinsfile that contains the `Check freeze period`. The PR should either be updated to not introduce this stage at all or this PR should be updated after #2 was merged to remove the stage.

### DIFF
--- a/.github/workflows/verifyFreezePeriod.yml
+++ b/.github/workflows/verifyFreezePeriod.yml
@@ -1,0 +1,12 @@
+# This workflow calls the Code-Freeze-Period check
+
+name: Check Code Freeze Period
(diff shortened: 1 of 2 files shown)


HannesWell:
Since #2 has been merged this PR is now ready.

Issue #22, opened 2022-04-18
ASTStructuralPropertyTest needs an update after bug 578608


stephan-herrmann:
AST node types introduced via https://bugs.eclipse.org/578608 are outside the range that is tested in `org.eclipse.jdt.core.tests.dom.ASTStructuralPropertyTest.testNodeClassForType()`.

I suggest to simply increase the loop limit (now `nodeType < 110`) by a larger step, say 10 or even more. Then you will see that the assert also needs adjusting - to correspond to the last constant in ASTNode.

SarikaSinha:
Thanks @stephan-herrmann

PR #25, opened 2022-04-19
ASTStructuralPropertyTest needs an update after bug 578608

SarikaSinha:
Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/22


jarthana:
It would have been nicer if we could find the last node type ID from ASTNode. That not being available, we are left with hard-coding this value in the tests every time the value goes out of range. But for a test, this is fine for me.

Issue #23, opened 2022-04-18
[null][external] Option to refer to a dependency in a container for external annotations


stephan-herrmann:
Assume a set of plug-in projects that want to maintain a set of **common external annotations** for all their third party dependencies, because repeating the same eea in each project is cumbersome. 

It is trivial to define a project / plug-in that ships eea files via (one of) its source folder.

We can also (since https://bugs.eclipse.org/574603) tell JDT to search for annotations in all classpath locations. While this is a viable solution, it may come at the price of performance, as significant more lookups need to be tried (for each type search all classpath entries for a matching .eea).

Here a new style of defining `annotationPath` in `.classpath` would help. 
* Assume your external annotations are in a plug-in `my.annots`
* Assume you define a dependency on my.annots via `MANIFEST.MF` (i.e., let PDE figure out where to find my.annots)
* The classpath entry for the container `org.eclipse.pde.core.requiredPlugins` should be able to select one of its resolved entries to contain eea for (potentially) all other entries.

I developed draft new lookup that would be able to resolve an annotation path of this shape:
```
	<attribute name="annotationpath" value="org.eclipse.pde.core.requiredPlugins/my.annots"/>
```

This would indicate:
* resolve the annotation path relative to the container `org.eclipse.pde.core.requiredPlugins`
* within the container, an entry `my.annots` should be searched
  * likely, any suffix indicating a version number, plus a `.jar` extension should be ignored during matching
  * as a heuristic, the specified name should be a prefix of the resolved container member, and the next char should not be a letter or digit (e.g., `-` or `_` ...), so a specified `my.annot` matches to a resolved entry `my.annot_1.0.0.202204181800.jar`, while we don't need to know the actual syntax of the tail.

One might wonder if explicitly saying `org.eclipse.pde.core.requiredPlugins/` is needed, because typically this declaration would happen right inside the element of that container, **but** also eea for the JRE could be declared using this mechanism, where the connection to the classpath container would otherwise not be given.

The nice part about using a container name to select this kind of resolving is: it can equally be used for non-plugin projects, by saying, e.g.,
```
	<attribute name="annotationpath" value="org.eclipse.m2e.MAVEN2_CLASSPATH_CONTAINER/my.annots"/>
```
JDT wouldn't need to know about the difference, which is key here :smile:

stephan-herrmann:
resolved via #48 (after test freezes in #28)

PR #28, opened 2022-04-19
[null][external] Option to refer to a dependency in a container for external annotations (#23)

stephan-herrmann:
- includes new API for JDT/UI to facilitate supporting this new option
- streamlined tests to focus on configuration, not content


iloveeclipse:
I'm currently trying to help Jay and Stephan to understand why their PR's are hanging...
See https://github.com/eclipse-jdt/eclipse.jdt.core/pull/38

While doing that, I've found #45 and that all opened PR's seem to be automatically rebuilt by Jenkins on every change in master branch. 

I **believe**, the PR's should be rebased in github on latest master (that has #45 merged) to avoid that crazy behavior.

It would be nice if @stephan-herrmann would do that now to verify my assumption, save our (jenkins) resources and may be prevent hangs for other PR's.

stephan-herrmann:
abandoning in favor of #48

PR #34, opened 2022-04-21
Bug 578874 - faster java autobuild cancel

jukzi:
Poll interrupt flag from JobManager like xtext does.

Trade off between UI responsiveness and additional work to restart
autobuild: We wait 3 seconds before interrupt build.

Can be configured with system property
"org.eclipse.jdt.MillisecondsBeforeInterruptAutoBuild=3000"

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/BuildNotifier.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/BuildNotifier.java
@@ -13,6 +13,8 @@
  *******************************************************************************/
 package org.eclipse.jdt.internal.core.builder;
 
+import java.util.function.BooleanSupplier;
+
 import org.eclipse.core.resources.*;
 import org.eclipse.core.runtime.*;
 
@@ -40,6 +42,14 @@ public static int FixedErrorCount = 0;
 public static int NewWarningCount = 0;
 public static int FixedWarningCount = 0;
 
+private static final int millisecondsBeforeCancelAutoBuild = Integer
+		.getInteger("org.eclipse.jdt.MillisecondsBeforeCancelAutoBuild", 1000); //$NON-NLS-1$
+private static final int millisecondsBeforeInterruptAutoBuild = Integer
+		.getInteger("org.eclipse.jdt.MillisecondsBeforeInterruptAutoBuild", 3000); //$NON-NLS-1$
+private BooleanSupplier interruptSupplier;
+private long startTimeNanos;
+private int buildKind;
+
 public static void resetProblemCounters() {
 	NewErrorCount = 0;
 	FixedErrorCount = 0;
@@ -47,8 +57,10 @@ public static void resetProblemCounters() {
 	FixedWarningCount = 0;
 }
 
-public BuildNotifier(IProgressMonitor monitor, IProject project) {
+public BuildNotifier(IProgressMonitor monitor, int buildKind, BooleanSupplier interruptSupplier) {
 	this.monitor = monitor;
+	this.buildKind = buildKind;
+	this.interruptSupplier = interruptSupplier;
 	this.cancelling = false;
 	this.newErrorCount = NewErrorCount;
 	this.fixedErrorCount = FixedErrorCount;
@@ -56,6 +68,7 @@ public BuildNotifier(IProgressMonitor monitor, IProject project) {
 	this.fixedWarningCount = FixedWarningCount;
 	this.workDone = 0;
 	this.totalWork = 1000000;
+	this.startTimeNanos = System.nanoTime();
 }
 
 /**
@@ -76,8 +89,32 @@ public void begin() {
  * Check whether the build has been canceled.
  */
 public void checkCancel() {
-	if (this.monitor != null && this.monitor.isCanceled())
-		throw new OperationCanceledException();
+	if (this.monitor != null && this.monitor.isCanceled()) {
+		// User requested cancel.
+		if (this.buildKind == IncrementalProjectBuilder.AUTO_BUILD) {
+			// Since jdt can not resume the incremental build we follow that request only after a period of grace
+			// to increase the chance the cancel happens between project builds and not during a project build
+			if (getBuildDurationInMs() > millisecondsBeforeCancelAutoBuild) {
+				throw new OperationCanceledException();
+			}
+		} else {
+			// non autobuild will be canceled immediately.
+			throw new OperationCanceledException();
+		}
+	}
+	if (this.interruptSupplier.getAsBoolean()) {
+		// Automatic request to interrupt build due to a conflicting user action.
+		// Since jdt can not resume the incremental build we follow that request only after a period of grace:
+		if (getBuildDurationInMs() > millisecondsBeforeInterruptAutoBuild) {
+			// This is a trade off between UI responsiveness and additional work to restart autobuild.
+			// We cancel autobuild to avoid possible UI freeze where the user can not manually cancel.
+			throw new OperationCanceledException();
+		}
+	}
+}
+
+private long getBuildDurationInMs() {
+	return ((System.nanoTime() - this.startTimeNanos) / 1_000_000);
 }
 
 /**
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/JavaBuilder.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/JavaBuilder.java
@@ -173,7 +173,8 @@ protected IProject[] build(int kind, Map ignored, IProgressMonitor monitor) thro
 	if (DEBUG)
 		System.out.println("\nJavaBuilder: Starting build of " + this.currentProject.getName() //$NON-NLS-1$
 			+ " @ " + new Date(System.currentTimeMillis())); //$NON-NLS-1$
-	this.notifier = new BuildNotifier(monitor, this.currentProject);
+	this.notifier = new BuildNotifier(monitor,kind,
+			kind == IncrementalProjectBuilder.AUTO_BUILD ? this::isInterrupted : ()->false);
 	this.notifier.begin();
 	boolean ok = false;
 	try {
@@ -303,7 +304,7 @@ protected void clean(IProgressMonitor monitor) throws CoreException {
 	if (DEBUG)
 		System.out.println("\nJavaBuilder: Cleaning " + this.currentProject.getName() //$NON-NLS-1$
 			+ " @ " + new Date(System.currentTimeMillis())); //$NON-NLS-1$
-	this.notifier = new BuildNotifier(monitor, this.currentProject);
+	this.notifier = new BuildNotifier(monitor,CLEAN_BUILD, ()->false);
 	this.notifier.begin();
 	try {
 		this.notifier.checkCancel();


org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/BuildNotifier.java
```diff
@@ -76,8 +89,35 @@ public void begin() {
  * Check whether the build has been canceled.
  */
 public void checkCancel() {
-	if (this.monitor != null && this.monitor.isCanceled())
-		throw new OperationCanceledException();
+	if (this.monitor != null && this.monitor.isCanceled()) {
+		// User requested cancel.
+		if (this.buildKind == IncrementalProjectBuilder.AUTO_BUILD) {
```
tsmaeder, line 94: I'm not sure we should be delaying user-induced cancellation: the user is boss! Anyway...the chance that a user can click cancel on an autobuild job before one second has elapsed seems small to me.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/BuildNotifier.java
```diff
@@ -40,22 +42,33 @@ public class BuildNotifier {
...
+	this.buildKind = buildKind;
+	this.interruptSupplier = interruptSupplier;
 	this.cancelling = false;
 	this.newErrorCount = NewErrorCount;
 	this.fixedErrorCount = FixedErrorCount;
 	this.newWarningCount = NewWarningCount;
 	this.fixedWarningCount = FixedWarningCount;
 	this.workDone = 0;
 	this.totalWork = 1000000;
+	this.startTimeNanos = System.nanoTime();
```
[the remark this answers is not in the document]
jukzi: yes! nanoTime is guaranteed to be monotonic, while currentTimeMillis are not. so for taking a difference millis are not. nanotime is the way to go for measure elapsed time.
https://stackoverflow.com/questions/2978598/will-system-currenttimemillis-always-return-a-value-previous-calls

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/BuildNotifier.java
jukzi: thats an excellent question.  Think of my workspace with ~500 projects where the autobuild can take several minutes. the user wants to cancel the autobuild because he does not want to wait minutes, but he does not care about a second. However if he presses cancel he has little luck to cancel autobuild between 2 projects. Higher chance he cancels just while a single project is building. That would result in a complete unbuild project which could take minutes to full-build on next autobuild. while when he just waited 1 second to get a good moment between projects then all autobuilds can resume incremental in the next round.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/BuildNotifier.java
```diff
@@ -76,8 +89,35 @@ public void begin() {
...
+	}
+	if (this.interruptSupplier.getAsBoolean()) {
+		// Automatic request to interrupt build due to a conflicting user action.
+		// Since jdt can not resume the incremental build we follow that request only after a period of grace:
+		if (getBuildDurationInMs() > millisecondsBeforeInterruptAutoBuild) {
+			// This is a trade off between UI responsiveness and additional work to restart autobuild.
+			// We cancel autobuild to avoid possible UI freeze where the user can not manually cancel.
+			throw new OperationCanceledException();
+		}
+		// XXX we could also use this.workDone to get an estimate how long the build will need and cancel immediately if
```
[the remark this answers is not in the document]
jukzi: 1. bugzilla is going to be shut down and to be replaced with github issues. 2. nobody is following this bug so adding any comment would reach nobody. 3. i have found several helpfull comments in the code that i would have never found well hidden in ancient bugzilla discussions.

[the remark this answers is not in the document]
gayanper: Should we document this in the wiki ? Not sure if we will move the Eclipse JDT wiki into github wiki. @mpalat will we keep the JDT wiki or will it be moved into github ?

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/BuildNotifier.java
gayanper: @jukzi Why don't we cancel at the first project state save operation to avoid a full build ? Or if you are trying to avoid a full build because we cancel before a delta build finishes, may be we can check for project state write with the build kind or cancel if the time elapse, which ever occurs first ?

jukzi: @gayanper that sound like you got a better idea - however my jdt knowledge is to limited to fully understand what that meand or how to implement your suggestion. i simply tried to extend the obvious existing cancel checks.
Could you please provide a alternative solution if my change is too bad?

jukzi:
unrelated org.eclipse.jdt.core.tests.compiler.regression.TryWithResourcesStatementTest.(?)
please rebuild

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/BuildNotifier.java
gayanper: @jukzi your change is good, i just want to make sure that we will not cancel in middle of delta changes which could result in full builds. I’m not sure the timeout we have right now in this patch is sufficient for all cases. There is a method that does update the build state at the end of delta builds. May be you can check that to see if that can be use to cancel without waiting and let the current logic handle all other scenarios.

i also came with these suggestions when i checked the code after reviewing this patch. So take my suggestion with a pinch of salt 😉

jukzi: Do you mean org.eclipse.jdt.internal.core.builder.JavaBuilder.buildDeltas(SimpleLookupTable)? it does not sound as an improvement to me to add there another cancel check as it is not within any loop. between projects the BuildManager will cancel immediatly anyway. org.eclipse.core.internal.events.BuildManager.checkCanceled(int, IProgressMonitor)

jukzi:
I think I covered all requested changes but the review is still -1, what do you think is missing?

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/BuildNotifier.java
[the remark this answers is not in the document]
tsmaeder: I doubt that makes a difference when we're talking about 1000 ms, but fair enough.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/BuildNotifier.java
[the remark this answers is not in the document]
tsmaeder: I'd like to invoke the Kantian categorical imperative here: we don't want _all_ code to look like this, so it shouldn't in this instance, either.  Anyway...a PR review is the wrong place to settle that question. This needs to go to the mailing list for all to see.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/BuildNotifier.java
```diff
@@ -76,8 +89,35 @@ public void begin() {
...
+		// that duration is greater then the period of grace.
+		// But since that estimate may not be very accurate it may be better to just wait the period of grace.
+	}
+}
+
+private int getBuildDurationInMs() {
+	return (int) ((System.nanoTime() - this.startTimeNanos) / 1_000_000);
```
tsmaeder, line 120: The cast to in seems unneccessary.

tsmaeder:
> I think I covered all requested changes but the review is still -1, what do you think is missing?

I'd like to get some clarity on the commenting issue.  This needs to be solved in general.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/BuildNotifier.java
jukzi: Type mismatch: cannot convert from long to int

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/BuildNotifier.java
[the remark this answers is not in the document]
tsmaeder: As I understand it, bugzilla is not going away (see mail to eclipse-dev).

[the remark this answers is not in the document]
jukzi: I removed the comment. I don't want the discussion about bugzilla be part of this change.

jukzi:
> I'd like to get some clarity on the commenting issue. This needs to be solved in general.

ok, clarify it, but please do not let the change depend on it.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/BuildNotifier.java
tsmaeder: Why not just let the method return long instead of int?

jukzi: done. i wonder why it works to compare int with long without typecast

jukzi:
> @jukzi would like to test a bit before merging. Do you have a good way to test this change?

That would be nice! It could be like (if you don't have a sufficient slow building workspace the only chance i see is to put breakpoints or a sleep into jdt build):

A) test cancel between project:
1. Have many (so many that total incremental build time is >>3sec) small java projects in workspace that build <3s each.
2. trigger autobuild that will build 2 files in each project. 
3. manual cancel autobuild in progress monitor.
4. wait
5. => autobuild should be stopped within 3sec between projects (next run should be INCREMENTAL), because delay is <3sec
 [without this change you would have got a cancel within porject -> most likely NOT incremental build]

B) test interrupt:
1. have 1 large project in workspace (incremental buildtime >>3sec)
2. trigger autobuild. 
3. trigger interrupt (touch file again)
4. wait
5. => autobuild should be stopped within 3sec, next run should be NOT incremental.  [without this change you could not interrupt it]

C) test cancel inside project:
1. have 1 large project in workspace (incremental buildtime >>3sec)
2. trigger autobuild (touch a file). 
3. manual cancel autobuild in progress monitor
4. wait 
5.=> autobuild should be stopped within 3sec, next run should be NOT incremental. [without this change the cancel might have been a bit faster]

D) play a bit around with many projects: every build should stop within 3ec regardless of interrupt or cancel.

To have high incremental build time its best to change something global, like class path ;-)

iloveeclipse:
I'm currently trying to help Jay and Stephan to understand why their PR's are hanging...
See https://github.com/eclipse-jdt/eclipse.jdt.core/pull/38

While doing that, I've found #45 and that all opened PR's seem to be automatically rebuilt by Jenkins on every change in master branch. 

I **believe**, the PR's should be rebased in github on latest master (that has #45 merged) to avoid that crazy behavior.

It would be nice if @jukzi would do that now to verify my assumption, save our (jenkins) resources and may be prevent hangs for other PR's.

tsmaeder (approved):

tsmaeder:
I'm fine to merge once the tests pass. Maybe just needs a rebase on master?

jukzi:
> I'm fine to merge once the tests pass. Maybe just needs a rebase on master?

ok, i rebased. i even squashed. i don't get this workflow - since its needed to  force push a rebase you won't be able to see the history of changes. I could have changed anything without notice. You might want to test again. After that i could be asked to rebase again and the story begins again ;-)
I guess it has a reason that github does not offer a rebase.

tsmaeder:
No need to squash if you want to keep the history. Actually it's better not to until right before merge.

jukzi:
thanks @tsmaeder for review & testing!

Issue #42, opened 2022-04-26
1100 formatter tests failed with exact same problem on Windows/Java 11 I20220425-1800


iloveeclipse:
~1100 formatter tests failed with exact same problem on Windows/Java 11 build **only**:
- https://download.eclipse.org/eclipse/downloads/drops4/I20220425-1800/testResults.php
- first failed test was `org.eclipse.jdt.core.tests.formatter.FormatterCommentsBugsTest.testBug121728`

Failure seem to be a problem with restoring workspace preferences **after** the test. I don't think it is Windows related, it looks like some random error.

```
--------- Difference is ----------
expected:<... compliance level: 1[.8\n
- JDK source level: 1.8\n
- JDK target level: 1.8]\n
- verbose : OFF\...> but was:<... compliance level: 1[1\n
- JDK source level: 11\n
- JDK target level: 11]\n
- verbose : OFF\...>
at org.eclipse.jdt.core.tests.junit.extension.TestCase.assertStringEquals(TestCase.java:266)
at org.eclipse.jdt.core.tests.junit.extension.TestCase.assertEquals(TestCase.java:242)
at org.eclipse.jdt.core.tests.model.AbstractJavaModelTests.tearDown(AbstractJavaModelTests.java:3909)
```

Note: the code checks that `JavaCore.getOptions()` equals to expected **defaults** from `org.eclipse.jdt.core.JavaCore.getDefaultOptions()`.

I see no immediate reason why this particular test would mess up options, nor recent related changes.

However I also do not see any code that would "fix" broken options to defaults.
I will add that to avoid following tests might be affected by some previous failure.

stephan-herrmann:
@iloveeclipse have you seen that this issue has a long history?
* https://bugs.eclipse.org/bugs/show_bug.cgi?id=302850
* https://bugs.eclipse.org/bugs/show_bug.cgi?id=482991 plus dupe 456147

After your change, is the original check still performed (didn't check the code)?

iloveeclipse:
@stephan-herrmann : sure, the change only prevents the problem to show 1000 times. 
Thanks for pointers, I was not aware the issue is old one, haven't seen such high number of fails yet.

PR #43, opened 2022-04-26
1100 formatter tests failed with exact same problem (#42)

iloveeclipse:
Restore expected workspace options on teardown and don't let all
following tests use broken defaults and fail too

Should fix https://github.com/eclipse-jdt/eclipse.jdt.core/issues/42


iloveeclipse:
Surprising result: 499 test fails !?!?
https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-43/1/#showFailuresLink

I can't understand how is that supposed to work if restoring defaults on teardown is **not** OK, but checking for defaults **is** OK. 

Someone changes defaults may be, so they match non-default options that are somewhere changed too???

Issue #41, opened 2022-04-26
Search is waiting for a background job, but no thread has any matching activity


iloveeclipse:
After jumping through some more maven hoops, I managed to get stacktraces from the freeze: [freeze-jstack.txt](https://github.com/eclipse-jdt/eclipse.jdt.core/files/8562807/freeze-jstack.txt)

two interesting portions extracted here:
```
"main" #1 prio=5 os_prio=0 cpu=32921,05ms elapsed=181,43s tid=0x00007fdd680261e0 nid=0x35536 in Object.wait()  [0x00007fdd6fb78000]
   java.lang.Thread.State: TIMED_WAITING (on object monitor)
	at java.lang.Object.wait(java.base@17/Native Method)
	- waiting on <no object reference available>
	at org.eclipse.jdt.internal.core.search.processing.JobManager.performConcurrentJob(JobManager.java:284)
	- locked <0x00000000c05bbbe0> (a org.eclipse.jdt.internal.core.search.indexing.IndexManager)
	at org.eclipse.jdt.internal.core.search.BasicSearchEngine.searchAllTypeNames(BasicSearchEngine.java:1923)
	at org.eclipse.jdt.internal.core.search.BasicSearchEngine.searchAllTypeNames(BasicSearchEngine.java:1751)
	at org.eclipse.jdt.core.search.SearchEngine.searchAllTypeNames(SearchEngine.java:1100)
	at org.eclipse.jdt.core.tests.model.AbstractJavaModelTests.waitUntilIndexesReady(AbstractJavaModelTests.java:3967)
	at org.eclipse.jdt.core.tests.dom.ASTConverterBugsTest.setUpSuite(ASTConverterBugsTest.java:41)
...
"Java indexing" #31 daemon prio=5 os_prio=0 cpu=871,25ms elapsed=178,36s tid=0x00007fdd68640200 nid=0x355bb in Object.wait()  [0x00007fdd402f9000]
   java.lang.Thread.State: WAITING (on object monitor)
	at java.lang.Object.wait(java.base@17/Native Method)
	- waiting on <no object reference available>
	at java.lang.Object.wait(java.base@17/Object.java:338)
	at org.eclipse.jdt.internal.core.search.processing.JobManager.run(JobManager.java:419)
	- locked <0x00000000c05bbbe0> (a org.eclipse.jdt.internal.core.search.indexing.IndexManager)
	at java.lang.Thread.run(java.base@17/Thread.java:833)
```

[...]

PR #46, opened 2022-04-26
Enable JobManager verbose mode for model tests

iloveeclipse:
For debugging https://github.com/eclipse-jdt/eclipse.jdt.core/issues/41

Doesn't fix anything, just trying to see if that extra info would be helpful or not.


iloveeclipse:
I don't plan to merge that, console explodes from debug output.

PR #61, opened 2022-05-03
scan all versions in multi-version-jar

rPraml:
This PR refers to
https://bugs.eclipse.org/bugs/show_bug.cgi?id=577790

### DIFF
--- a/org.eclipse.jdt.core/batch/org/eclipse/jdt/internal/compiler/batch/ClasspathMultiReleaseJar.java
+++ b/org.eclipse.jdt.core/batch/org/eclipse/jdt/internal/compiler/batch/ClasspathMultiReleaseJar.java
@@ -49,9 +49,13 @@ public class ClasspathMultiReleaseJar extends ClasspathJar {
 			} catch (FileSystemAlreadyExistsException e) {
 				this.fs = FileSystems.getFileSystem(uri);
 			}
-			this.releasePath = this.fs.getPath("/", "META-INF", "versions", this.compliance); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
-			if (!Files.exists(this.releasePath)) {
-				this.releasePath = null;
+			for (int version = Integer.parseInt(this.compliance); version >= 9; version--) {
+				this.releasePath = this.fs.getPath("/", "META-INF", "versions", String.valueOf(version)); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
+				if (Files.exists(this.releasePath)) {
+					break;
+				} else {
+					this.releasePath = null;
+				}
 			}
 		}
 	}
--- a/org.eclipse.jdt.core/batch/org/eclipse/jdt/internal/compiler/batch/ModuleFinder.java
+++ b/org.eclipse.jdt.core/batch/org/eclipse/jdt/internal/compiler/batch/ModuleFinder.java
@@ -237,10 +237,13 @@ public class ModuleFinder {
 		try {
 			zipFile = new ZipFile(file);
 			if (release != null) {
-				String releasePath = "META-INF/versions/" + release + "/" + path; //$NON-NLS-1$ //$NON-NLS-2$
-				ZipEntry entry = zipFile.getEntry(releasePath);
-				if (entry != null) {
-					path = releasePath;
+				for (int version = Integer.parseInt(release); version >= 9; version--) {
+					String releasePath = "META-INF/versions/" + version + "/" + path; //$NON-NLS-1$ //$NON-NLS-2$
+					ZipEntry entry = zipFile.getEntry(releasePath);
+					if (entry != null) {
+						path = releasePath;
+						break;
+					}
 				}
 			}
 			ClassFileReader reader = ClassFileReader.read(zipFile, path);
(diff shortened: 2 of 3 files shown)


iloveeclipse:
@rPraml : thanks for contributing. 
1) Please sign SCA agreement, otherwise we can't accept the patch
2) Please check contribution guidelines / git history for appropriate commit title
3) Please describe in the commit message the fix
4) ideally we would like to see a regression test added. 

See https://github.com/eclipse-jdt/.github/blob/main/CONTRIBUTING.md

org.eclipse.jdt.core/batch/org/eclipse/jdt/internal/compiler/batch/ClasspathMultiReleaseJar.java
```diff
@@ -49,9 +49,13 @@ public void initialize() throws IOException {
 			} catch (FileSystemAlreadyExistsException e) {
 				this.fs = FileSystems.getFileSystem(uri);
 			}
-			this.releasePath = this.fs.getPath("/", "META-INF", "versions", this.compliance); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
-			if (!Files.exists(this.releasePath)) {
-				this.releasePath = null;
+			for (int version = Integer.parseInt(this.compliance); version >= 9; version--) {
+				this.releasePath = this.fs.getPath("/", "META-INF", "versions", String.valueOf(version)); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
```
rPraml, line 53: according to
https://docs.oracle.com/javase/10/docs/specs/jar/jar.html#multi-release-jar-files
all versions down to 9 have to be checked.
`this.compliance` should be always a major version accroding to the spec, so `parseInt` should work

org.eclipse.jdt.core/batch/org/eclipse/jdt/internal/compiler/batch/ModuleFinder.java
```diff
@@ -237,10 +237,13 @@ private static IModule extractModuleFromArchive(File file, Classpath pathEntry,
 		try {
 			zipFile = new ZipFile(file);
 			if (release != null) {
-				String releasePath = "META-INF/versions/" + release + "/" + path; //$NON-NLS-1$ //$NON-NLS-2$
-				ZipEntry entry = zipFile.getEntry(releasePath);
-				if (entry != null) {
-					path = releasePath;
+				for (int version = Integer.parseInt(release); version >= 9; version--) {
```
rPraml, line 240: This is practically the same code as above.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/ClasspathJar.java
```diff
@@ -125,13 +125,18 @@ IModule initializeModule() {
...
 	try {
 		file = new ZipFile(this.zipFilename);
-		String releasePath = "META-INF/versions/" + this.compliance + '/' + IModule.MODULE_INFO_CLASS; //$NON-NLS-1$
 		ClassFileReader classfile = null;
-		try {
-			classfile = ClassFileReader.read(file, releasePath);
-		} catch (Exception e) {
-			e.printStackTrace();
-			// move on to the default
+		for (int version = Integer.parseInt(this.compliance); version >= 9; version--) {
```
rPraml, line 129: and here, we search for the `module-info.class`, which should be the fix for the bug

rPraml (commented):
I've identified 3 places, where multi-release-jars are not properly handled according to the spec.
These changes should make ecj more javac comatible.

It would be great, if someone can pick up this work and write a test, as I have not much knowledge how to do this in the eclipse ecosystem

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/ClasspathJar.java
jarthana: Any reason why It's being done in ClasspathJar instead of org.eclipse.jdt.internal.core.builder.ClasspathMultiReleaseJar ?

rPraml: @jarthana Do you mean, why I have made this change or why this code is here at all?

IMHO it sounds logical, if this is handled in ClasspathMultiReleaseJar as it does not require to have duplicate code.
(But I'm really not the expert to do such refactoring in this project)

jarthana: I was just wondering about having this in the super class ClasspathJar but in case of the compiler.batch package, you have made it in the subclass ClasspathMultiReleaseJar. Haven't looked at the code in detail, but IIRC, we had similar code and hierarchy in both the compiler.batch and core.builder packages.

org.eclipse.jdt.core/batch/org/eclipse/jdt/internal/compiler/batch/ClasspathMultiReleaseJar.java
laeubi: Probably it would be good to parse the compliance only once (same below) and cache it in a field?

PR #63, opened 2022-05-05
Moved detecting out-of-bounds continues to FlowContext. Fixes #53

### LINKED ISSUE #53: `continue` without label is incorrectly handled in a switch expression
Tested on Eclipse 2022-03; jdt.core version 3.29.0.v20220214-1307.

The JDT compiler allows a `continue` statement inside a (case of a) switch expression to continue to a loop declared outside of the switch expression, iff that `continue` doesn't target an explicit label.
The JLS clearly forbids this: "It is a compile-time error if the continue target contains any [...] switch expression that encloses the continue statement. That is, there are no non-local jumps." in https://docs.oracle.com/javase/specs/jls/se17/html/jls-14.html#jls-14.16 / https://docs.oracle.com/javase/specs/jls/se18/html/jls-14.html#jls-14.16

MWE:
```java
public class ContinueFromSwitchExpression
{
    public static String string = "a";

    public static void main(String[] args)
    {
        loop: for(;;)
        {
            System.out.println("In loop before switch");
            int result = 123 + switch(string)
            {
                case "a" ->
                {
                    continue; // incorrectly compiles in JDT
                    //continue loop; // correctly flagged as error ("Continue out of switch expressions not permitted")
                    // javac (correctly) outputs "error: attempt to continue out of a switch expression" for both continue statements
                }
                default -> 456;
            };
            System.out.println("After switch. result: " + result);
        }
    }
}
```

Resulting classfile:
```
public class ContinueFromSwitchExpression {
  public static java.lang.String string;

[... default constructor and <clinit> ...]

  public static void main(java.lang.String[]);
    Code:
       0: getstatic     #23                 // Field java/lang/System.out:Ljava/io/PrintStream;
       3: ldc           #29                 // String In loop before switch
       5: invokevirtual #31                 // Method java/io/PrintStream.println:(Ljava/lang/String;)V
       8: bipush        123
      10: getstatic     #12                 // Field string:Ljava/lang/String;
      13: dup
      14: astore_2
      15: invokevirtual #37                 // Method java/lang/String.hashCode:()I
      18: lookupswitch  { // 1
                    97: 36
               default: 45
          }
      36: aload_2
      37: ldc           #10                 // String a
      39: invokevirtual #43                 // Method java/lang/String.equals:(Ljava/lang/Object;)Z
      42: ifne          0
      45: sipush        456
      48: iadd
      49: istore_1
      50: getstatic     #23                 // Field java/lang/System.out:Ljava/io/PrintStream;
      53: new           #47                 // class java/lang/StringBuilder
      56: dup
      57: ldc           #49                 // String After switch. result:
      59: invokespecial #51                 // Method java/lang/StringBuilder."<init>":(Ljava/lang/String;)V
      62: iload_1
      63: invokevirtual #53                 // Method java/lang/StringBuilder.append:(I)Ljava/lang/StringBuilder;
      66: invokevirtual #57                 // Method java/lang/StringBuilder.toString:()Ljava/lang/String;
      69: invokevirtual #31                 // Method java/io/PrintStream.println:(Ljava/lang/String;)V
      72: goto          0
}
```

Trying to run the resulting classfile throws a verification error because of the value 123 remaining on the stack when the `continue` jumps to the loop beginning. Removing `123 + ` causes the JVM to not throw any errors; it then prints "In loop before switch" repeatedly and indefinitely as expected (but still incorrectly).

Haspamelodica:
Fixing #53

### DIFF
--- a/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/ast/ContinueStatement.java
+++ b/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/ast/ContinueStatement.java
@@ -51,6 +51,9 @@ public FlowInfo analyseCode(BlockScope currentScope, FlowContext flowContext, Fl
 	if (targetContext == FlowContext.NotContinuableContext) {
 		currentScope.problemReporter().invalidContinue(this);
 		return flowInfo; // pretend it did not continue since no actual target
+	} else if (targetContext == FlowContext.ContinuedContextCrossingSwitchExpression) {
+		currentScope.problemReporter().switchExpressionsContinueOutOfSwitchExpression(this);
+		return flowInfo; // pretend it did not continue since no actual target
 	}
 	this.initStateIndex =
 		currentScope.methodScope().recordInitializationStates(flowInfo);
--- a/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/ast/SwitchExpression.java
+++ b/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/ast/SwitchExpression.java
@@ -389,11 +389,9 @@ public class SwitchExpression extends SwitchStatement implements IPolyExpression
 	static class OOBLFlagger extends ASTVisitor {
 		Set<String> labelDecls;
 		Set<BreakStatement> referencedBreakLabels;
-		Set<ContinueStatement> referencedContinueLabels;
 		public OOBLFlagger(SwitchExpression se) {
 			this.labelDecls = new HashSet<>();
 			this.referencedBreakLabels = new HashSet<>();
-			this.referencedContinueLabels = new HashSet<>();
 		}
 		@Override
 		public boolean visit(SwitchExpression switchExpression, BlockScope blockScope) {
@@ -406,12 +404,6 @@ public class SwitchExpression extends SwitchStatement implements IPolyExpression
 						continue;
 					if (!this.labelDecls.contains(new String(bs.label)))
 						blockScope.problemReporter().switchExpressionsBreakOutOfSwitchExpression(bs);
-				}
-				for (ContinueStatement cs : this.referencedContinueLabels) {
-					if (cs.label == null || cs.label.length == 0)
-						continue;
-					if (!this.labelDecls.contains(new String(cs.label)))
-						blockScope.problemReporter().switchExpressionsContinueOutOfSwitchExpression(cs);
 				}
 			} catch (EmptyStackException e) {
 				// ignore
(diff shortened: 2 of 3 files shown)


iloveeclipse:
Few formal issues to address (nkte, I have no idea about code itself) :
1) Please sign ECA
2) Please provide appropriate commit message containing link to the issue and describing the fix
3) There are few test fails - please check and ideally not only fix bit add new test that demonstrates the fix for the original problem.

Thank you. Please don't give up.

Haspamelodica:
This PR isn't meant to be ready to be merged yet 😉 However, since you mentioned it:
1. I already signed the ECA a few times today, and clicked on "Revalidate". It doesn't seem to have an effect, though.

Haspamelodica:
Ah, I just did so again (accept the ECA and click Revalidate); now it works.

stephan-herrmann:
> Regarding the failing tests: The failing cases seem to rely on the incorrect behaviour I found. This makes me wonder if I'm correct in my assumption that I found a compiler bug. I would like someone more "experienced" to have another look.

From only a brief look, I see different situations:
* testBug544073_076: checks for "'continue' [...] cannot be the last statement [...]" and **additionally** finds "your" error => it's plausible the test author simply didn't notice the double wrong :)
* testBug558067_005: checks for "Continue out of switch expressions not permitted" and **additionally** finds "The label LABEL is never explicitly referenced" => this looks like an undesirable side-effect of the change :-1: 
* testBug558067_003: like above **plus** "Unreachable code" => this is acceptable (perhaps it wasn't reported previously because the primary error stopped further flow analysis).

srikanth-sankaran:
> Fixing #53

I think as of result of "Fixing" instead of "Fixes", this PR didn't get linked with the issue prominently in the title itself with a "May be fixed by <PR>" and I seem to have missed it and attempted a parallel fix - sorry! 

That said https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1056 is more comprehensive and completely rips out the abstraction SwitchExpression.OOBLFlagger and uniformly handles continue/break/return by integrating with the flow analysis - where these checks belong rather than in resolve as you rightly point out.

https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1056 is ready to be merged in and I'll go ahead and close this one - sorry once again for having inadvertently overlooked your effort

srikanth-sankaran:
> > Regarding the failing tests: The failing cases seem to rely on the incorrect behaviour I found. This makes me wonder if I'm correct in my assumption that I found a compiler bug. I would like someone more "experienced" to have another look.
> 
> From only a brief look, I see different situations:
> 
> * testBug544073_076: checks for "'continue' [...] cannot be the last statement [...]" and **additionally** finds "your" error => it's plausible the test author simply didn't notice the double wrong :)
> * testBug558067_005: checks for "Continue out of switch expressions not permitted" and **additionally** finds "The label LABEL is never explicitly referenced" => this looks like an undesirable side-effect of the change 👎
> * testBug558067_003: like above **plus** "Unreachable code" => this is acceptable (perhaps it wasn't reported previously because the primary error stopped further flow analysis).

The additional error for the label not explicitly referenced is arguably correct or at least not patently incorrect. There is no valid reference that binds to the label. 

compare with:
```
public class X {
		label:
			System.out.println();
		while (true) {
			break label;
		}
	}
}
```

ECJ reports:
The label label is missing AND
The label label is never explicitly referenced

Javac says:
X.java:6: error: undefined label: label
			break label;
			^
1 error

Both set of messages can be argued with :)

PR #69, opened 2022-05-09
MT hardening JobManager and IndexManager (hopefully fixes issue #41)

jukzi:
to fix CI hung in JobManager.run -> wait()

* distinct private mutexes for both classes (they did share "this")
* added missing "synchronized" accesses
* documented synchronizing
* SynchronizedSet for IndexManager.metaIndexUpdates
* AtomicInteger for JobManager.awaitingClients
* added missing notify in activateProcessing()
* wait instead of sleep

### DIFF
--- a/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/IndexManager.java
+++ b/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/IndexManager.java
@@ -26,7 +26,6 @@ import java.util.Collections;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
-import java.util.ListIterator;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Objects;
@@ -95,24 +94,34 @@ public class IndexManager extends JobManager implements IIndexConstants {
 	// disabled by default until Bug574171 is fixed.
 	private static final boolean DISABLE_META_INDEX = Boolean.parseBoolean(System.getProperty("org.eclipse.jdt.disableMetaIndex", "false")); //$NON-NLS-1$ //$NON-NLS-2$
 
-	// key = containerPath, value = indexLocation path
-	// indexLocation path is created by appending an index file name to the getJavaPluginWorkingLocation() path
+	/**
+	 * synchronized by IndexManager.this
+	 * <br>
+	 * key = containerPath, value = indexLocation path
+	 * <br>
+	 * indexLocation path is created by appending an index file name to the getJavaPluginWorkingLocation() path
+	 */
 	public SimpleLookupTable indexLocations = new SimpleLookupTable();
-	// key = indexLocation path, value = an index
-	private SimpleLookupTable indexes = new SimpleLookupTable();
 
 	/**
-	 * The new indexer is disabled, see bug 544898
+	 * key = indexLocation path, value = an index
+	 * <br>
+	 * synchronized by IndexManager.this
 	 */
-//	private Indexer indexer = Indexer.getInstance();
+	private SimpleLookupTable indexes = new SimpleLookupTable();
 
 	/* need to save ? */
-	private boolean needToSave = false;
+	private volatile boolean needToSave;
 	private IPath javaPluginLocation = null;
 
-	/* can only replace a current state if its less than the new one */
-	// key = indexLocation path, value = index state integer
-	private SimpleLookupTable indexStates = null;
+	/**
+	 * can only replace a current state if its less than the new one
+	 * <br>
+	 * key = indexLocation path, value = index state integer
+	 * <br>
+	 * synchronized by IndexManager.this
+	*/
+	private SimpleLookupTable indexStates;
 	private File indexNamesMapFile = new File(getSavedIndexesDirectory(), "indexNamesMap.txt"); //$NON-NLS-1$
 	private File participantIndexNamesFile = new File(getSavedIndexesDirectory(), "participantsIndexNames.txt"); //$NON-NLS-1$
 	private boolean javaLikeNamesChanged = true;
@@ -124,9 +133,14 @@ public class IndexManager extends JobManager implements IIndexConstants {
 
 	private final IndexNamesRegistry nameRegistry = new IndexNamesRegistry(new File(getSavedIndexesDirectory(),
 			"savedIndexNames.txt"), getJavaPluginWorkingLocation()); //$NON-NLS-1$
-	// search participants who register indexes with the index manager
-	private SimpleLookupTable participantsContainers = null;
-	private boolean participantUpdated = false;
+	/**
+	 * search participants who register indexes with the index manager
+	 * <br>
+	 * synchronized by IndexManager.this
+	 */
+	private SimpleLookupTable participantsContainers;
+	/** synchronized by IndexManager.this */
+	private boolean participantUpdated;
 	private volatile MetaIndex metaIndex;
 
 	// should JDT manage (update, delete as needed) pre-built indexes?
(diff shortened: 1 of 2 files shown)


jukzi:
unrelated errors:
[org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_003](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-69/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug501162_003/)
[org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_022](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-69/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug501162_022/)
[org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug519980_001](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-69/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug519980_001/)
[org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug530016_001](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-69/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug530016_001/)
[org.eclipse.jdt.core.tests.model.ResolveTests2.testBug232880b](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-69/9/testReport/junit/org.eclipse.jdt.core.tests.model/ResolveTests2/testBug232880b/)

no stacktrace logged

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/processing/JobManager.java
```diff
@@ -558,4 +561,25 @@ public String toString() {
 		}
 		return buffer.toString();
 	}
+
+	protected synchronized boolean findLastJob(Predicate<IJob> testJob) {
```
iloveeclipse, line 565: This change / code can be further simplified with the merging of same code block above.

iloveeclipse (changes requested):
I haven't yet reviewed changed semantics of using more / different locks, but I will push an updated cleaned up PR to make review easier

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/processing/JobManager.java
```diff
@@ -29,32 +31,54 @@
...
+	private synchronized Thread getProcessingThread() {
+		return this.processingThread;
+	}
 	/**
 	 * Invoked exactly once, in background, before starting processing any job
 	 */
-	public void activateProcessing() {
+	public synchronized void activateProcessing() {
 		this.activated = true;
+		this.notifyAll(); // someone may wait for awaitingJobsCount()
```
iloveeclipse, line 81: Not sure this call is needed here (and synchronized method call), as it doesn't necessarily means there job count is changed - it **may** change later on.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/processing/JobManager.java
```diff
@@ -142,6 +168,7 @@ public void discardJobs(String jobFamily) {
 						}
 						currentJob.cancel();
 						it.remove();
+						notifyAll(); // notify waiters for awaitingJobsCount()
```
iloveeclipse, line 171: Since we hold lock during the loop anyway (so waiters will stop sleeping but will be blocked), shouldn't this be called only once **after** the while loop?

iloveeclipse (commented):
Jörg, two questions regarding JobManager changes.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/processing/JobManager.java
jukzi: awaitingJobsCount reads activated which is changed here

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/processing/JobManager.java
jukzi: That should be no difference. Don't know if the compiler does that anyway.

iloveeclipse: No, it doesn't (hint: use Bytecode view to see what compiler does). 
I would move that also below outside the loop. Just feels not natural to call that N times if in reality it only needs to be done once.

jukzi: It's the hotspot compiler which would do such optimizations at runtime. I don't care.

iloveeclipse:
OK, Jörg, please check `JobManager` I've just pushed - that should be OK now from my POV.
Next (for me): `IndexManager` review.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/IndexManager.java
```diff
@@ -251,10 +266,10 @@ public void addSource(IFile resource, IPath containerPath, SourceElementParser p
 	IndexLocation indexLocation = computeIndexLocation(containerPath);
 	scheduleDocumentIndexing(document, containerPath, indexLocation, participant);
 }
-/*
+/**
  * Removes unused indexes from disk.
  */
-public void cleanUpIndexes() {
+public synchronized void cleanUpIndexes() {
```
iloveeclipse, line 272: Note: this is called from `org.eclipse.core.internal.resources.SaveManager.save()` that will lock the workspace before. Means, we might block workspace save and in worst case we have other code that locks the index manager before and needs workspace lock to do something, just like the code few lines below calling to `BasicSearchEngine.createWorkspaceScope()`.

jukzi:
JobManager has some minor odds left:
The logging of this.enableCount may not be meaningful, as it is not synchronized.
The thread returned by getProcessingThread() is used after the synchronization which may then not be the right thread anymore.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/IndexManager.java
```diff
@@ -1155,26 +1161,11 @@ public void saveIndex(Index index) throws IOException {
 			return;
 		}
 	}
+	IPath containerPath = new Path(index.containerPath);
+	if (hasPendingJobMatching(job -> (job instanceof IndexRequest) && (((IndexRequest) job).containerPath.equals(containerPath)))) {
+		return;
+	}
 	synchronized (this) {
```
iloveeclipse, line 1168: The scope of the `synchronized(this)` is reduced and allows non-sync changes between the check / return and following code. I will move the check above into the sync scope back.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/IndexManager.java
iloveeclipse: OK, I've checked now the callees of the code inside, the change should be safe.
```
cleanUpIndexes() ->  synchronized
	conditionally called removeIndexesState() (synchronized)
	called getIndexes(IndexLocation[], IProgressMonitor) (non synchronized)
		called getIndex(IndexLocation) (synchronized)
		called conditionally getIndex(IPath, IndexLocation, boolean, boolean) (synchronized)
		called conditionally updateIndexState(IndexLocation, Integer) (synchronized)
```

iloveeclipse (approved):
The rest looks good. 
Let's hope that fixes sporadic test hangs.

Issue #72, opened 2022-05-10
endless running Daemon Thread [Compiler Source File Reader] 


jukzi:
Every time i build a large project and then cancel the build there is one additional Thread created and never terminated.
Those threads getting more and more...

![image](https://user-images.githubusercontent.com/51790620/167650056-6db61a1b-5c2a-4342-9e03-11d18df04a8d.png)

jukzi:
in org.eclipse.jdt.internal.compiler.Compiler.internalBeginToCompile(ICompilationUnit[], int) the shutdown() is not initiated finally.
i will push a solution

PR #73, opened 2022-05-10
Always shutdown readManager (closes issue #72)

jukzi:
Each cancel had left one "Compiler Source File Reader" thread alive.


iloveeclipse:
Thanks.

PR #88, opened 2022-05-17
Bug 543817 - [null] better null analysis exceptions for java.util.Object's methods and java.util.Collections factories

stephan-herrmann:
- avoid spurious warning against Objects.requireNonNull()


stephan-herrmann:
Instead of auto-linking: [Bug 543817](https://bugs.eclipse.org/bugs/show_bug.cgi?id=543817)

PR #143, opened 2022-06-16
Complete inline tags with blank (fixes #139)

### LINKED ISSUE #139: Javadoc code completion for inline tags misses a space
* Have the cursor inside Javadoc
* type "@"
* trigger code completion
* select "{@code}", return

Result "{@code|}" (where | denotes the cursor). The user always has to type a blank before writing the actual tag content. To me it looks like this was caused by https://bugs.eclipse.org/bugs/show_bug.cgi?id=121026. That issue was about having one blank too much when _wrapping_ existing characters into inline tags via code completion. Unfortunately the fix modified both the code for just inserting the plain tag and the code for wrapping. Half of the patch needs to be reverted therefore.

I'll provide a fix.

Bananeweizen:
## What it does
Complete those inline tags with an additional blank character as
completion suffix which take additional text from the user, e.g.
`{@code }` instead of `{@code}`.

This reverts half of
https://bugs.eclipse.org/bugs/show_bug.cgi?id=121026, which meant to
only suppress superfluous blanks when adding the completion around
existing text, but wrongly removed blanks also from completion that does
not wrap around input. The test case of that linked bug still works as expected after this change.

Be aware that this change has a visual side effect during completion. The suggested completions in the drop down also contain the newly added blank. I would create another change in jdt.ui to remove the blank from the label provider in the drop down to make this look like before.

image

As you can see from the screenshot, not _all_ tags get an additional blank, but only a known list of tags. Therefore docRoot remains as before.

The result after completion looks like this:
image

## How to test
* Have the cursor inside Javadoc
* type "@"
* trigger code completion
* select "{@code}", return

## Author checklist
- I have thoroughly tested my changes (_only the modified classes_)


akurtakov:
Is this one still relevant? If yes please rebase it on master.

jarthana:
I have rebased on master and also removed all the white space only changes and force-pushed. At a glance, looks like the patch may still be relevant. But without any tests, can't confirm.

PR #152, opened 2022-06-19
Don't open InputStreamReader without closing it

carstenartur:
Just a pull request as base for tests how long explicitly unclosed stream/reader need to be closed automatically. (depending on java vm?)

## What it does
Instead of open and never explicitly closing stream/reader to check for UnsupportedEncodingException to find out if an encoding is supported it makes use of explicit api.

Charset.availableCharsets().keySet().contains(name)

instead of 

new InputStreamReader(new ByteArrayInputStream(new byte[0]), stringValue);

Should there ever be interest to merge it Charset.availableCharsets().keySet() might better be cached.
But for a test through the build this is not necessary.

image

So no single word about what specification the *encoding name* has to follow.
In fact it is the java.io name that is expected as long as you do not make use of nio methods. This is however changed using this patch - so I'm not sure this would be possible at all for commandline parameters. Many encoding names are the same but some important ones are not. I added a small mapping but I'm not sure if it is worth it.
Here you can see the currently used encoding names in the column "Canonical Name for java.io API and java.lang API".
https://docs.oracle.com/javase/7/docs/technotes/guides/intl/encoding.doc.html
So I want to check if I can see how long the GC needs to close the readers. After that we can close - don't care..

## How to test
Check non standard and not existing encodings on the platform.

## Author checklist


jarthana:
notify @jarthana

Issue #147, opened 2022-06-17
Array type is getting generated in the .class where java.lang.Object should have been generated


sravanlakkimsetti:
Based on https://github.com/eclipse-equinox/p2/issues/92

for the code [return methodHandle.invoke(self);](https://github.com/eclipse-equinox/p2/blob/245b912c1d082336e68f0df882b601dd458fd706/bundles/org.eclipse.equinox.p2.metadata/src/org/eclipse/equinox/internal/p2/metadata/expression/Member.java#L73), We are generating 
```
INVOKEVIRTUAL java/lang/invoke/MethodHandle.invoke ([Ljava/lang/Object;)Ljava/lang/Object;
```

sravanlakkimsetti:
Notify @jarthana

jarthana:
This is turns out to be a result of consuming "java/lang/invoke/MethodHandle" with --release option. This particular class file, when --release is used is having wrong content. I.e., the RuntimeVisibleAnnotations attribute is missing. As a result of this, the @PolymorphicSignature from the MethodHandle#invoke() is ignored, which further affects how we generate code for a variable arity method. As far JDT is concerned, the relevant code is in Statement#generateArguments() where binding.isVarargs() returns false since the method is not created as polymorphic. If we have a version of JDK with correct .class files, we should take them and close this as not eclipse.

sravanlakkimsetti:
I went ahead with looking into class file changes between java 17.0.1 and 17.0.3. There are no changes between them. I am really not confident on calling this as fixed in 17.0.3 yet. 

Will continue investigation on this one.

jarthana:
> I went ahead with looking into ct.sym file changes between java 17.0.1 and 17.0.3. There are no changes between them. I am really not confident on calling this as fixed in 17.0.3 yet.
> 
> Will continue investigation on this one.

Actually, I find the produced .class file having the same issue with 17.0.3.1 as well.

sravanlakkimsetti:
We are depending on PolymorphicSignature annotation to determine whether the method is polymorphic or not. But according to https://docs.oracle.com/javase/specs/jls/se18/html/jls-15.html#jls-15.12.3

_A method is signature polymorphic if all of the following are true:_

- _It is declared in the java.lang.invoke.MethodHandle class or the java.lang.invoke.VarHandle class._
[...]

PR #153, opened 2022-06-20
Adding Polymorphic method deduction

sravanlakkimsetti:
## What it does
We are depending on PolymorphicSignature annotation to determine whether
the method is polymorphic or not. But according to
https://docs.oracle.com/javase/specs/jls/se11/html/jls-15.html#jls-15.12.3

A method is signature polymorphic if all of the following are true:

- It is declared in the java.lang.invoke.MethodHandle class or the java.lang.invoke.VarHandle class.
- It has a single variable arity parameter (§8.4.1) whose declared type is Object[].
- It is native.

I guess we should deduce PolymorphicSignature annotation based on above definition

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/147

## How to test
Test case has been added for this particular issue

Can test with code snippet in https://github.com/eclipse-jdt/eclipse.jdt.core/issues/147#issuecomment-1160323735 with --release and and without --release

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/BatchCompilerTest2.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/BatchCompilerTest2.java
@@ -26,7 +26,7 @@ import junit.framework.Test;
 public class BatchCompilerTest2 extends AbstractBatchCompilerTest {
 
 	static {
-		TESTS_NAMES = new String[] { "testIssue147" };
+//		TESTS_NAMES = new String[] { "testIssue147" };
 //		TESTS_NUMBERS = new int[] { 306 };
 //		TESTS_RANGE = new int[] { 298, -1 };
 	}
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/BatchCompilerTest2.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/BatchCompilerTest2.java
@@ -461,7 +461,7 @@ public void testIssue147() throws Exception {
 			"}\n"
 		},
 		"\"" + OUTPUT_DIR +  File.separator + "X.java\"" +
-				" --release " + CompilerOptions.VERSION_11 + " ",
+				" --release " + CompilerOptions.VERSION_9 + " ",
 				"",
 				"",
 				true);
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/BatchCompilerTest2.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/BatchCompilerTest2.java
@@ -26,7 +26,7 @@ import junit.framework.Test;
 public class BatchCompilerTest2 extends AbstractBatchCompilerTest {
 
 	static {
-//		TESTS_NAMES = new String[] { "test440477" };
+		TESTS_NAMES = new String[] { "testIssue147" };
 //		TESTS_NUMBERS = new int[] { 306 };
 //		TESTS_RANGE = new int[] { 298, -1 };
 	}
@@ -438,4 +438,34 @@ public void testIssue114() {
 		);
 }
 
+public void testIssue147() throws Exception {
+	runConformTest(
+		new String[] {
+			"X.java",
+			"import java.lang.invoke.MethodHandle;\n" +
+			"import java.lang.invoke.MethodHandles;\n" +
+			"import java.lang.reflect.Method;\n" +
+			"\n" +
+			"\n" +
+			"public class X {\n" +
+			"	public final Object invoke(Object self) throws Throwable {\n" +
+			"\n" +
+			"			Method method = null;\n" +
+			"			try {\n" +
+			"				MethodHandle methodHandle = MethodHandles.lookup().unreflect(method );\n" +
+			"				return methodHandle.invoke(self);\n" +
+			"			} catch (IllegalArgumentException e) {\n" +
+			"				throw e;\n" +
+			"			} \n" +
+			"		}\n" +
+			"}\n"
+		},
+		"\"" + OUTPUT_DIR +  File.separator + "X.java\"" +
+				" --release " + CompilerOptions.VERSION_11 + " ",
+				"",
+				"",
+				true);
+	String expectedOutput = "java.lang.invoke.MethodHandle.invoke(java.lang.Object)";
+	checkDisassembledClassFile(OUTPUT_DIR + File.separator + "X.class", "X", expectedOutput);
+}
 }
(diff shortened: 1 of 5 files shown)


jarthana:
Here are some points to consider:

- Scope#computeCompatibleMethod() is a high traffic method. Not sure we want to do all that string comparisons there. Considering that this is only required when the annotation tagbit is not set, I think we should move this code to the else of the following check:

             if ((method.tagBits & TagBits.AnnotationPolymorphicSignature) != 0) {

- We should move all the easier checks like isNative(), isVarargs() etc. to the front to avoid the overhead.
- Also, take a look at TypeConstants and TypeIds that we use to code type names etc.

jarthana:
I would still like Stephan to do a review since it relates to the signature polymorphic method, which I am not an expert in.

@stephan-herrmann I hope you can spare some time for this.

sravanlakkimsetti:
@stephan-herrmann Can you please take a look at this one? This issue is causing problem in generation class files in one of the equinox projects while using --release 11 with java 17 jvm.

Please note @PolymophicSignature does not exist in ct.sym.

stephan-herrmann:
Sorry for the delay.

### Testing
To exercise the bug the test must be run on a JRE > 11 (for related reason I used `--release 9` in PR #156 - tests always run on a JRE > 9 :) ).

### Strategy discussion
It seems like Java invented an interesting notion of "internal": 

The annotation type `PolymorphicSignature` is package private in `MethodHandle`, which renders its use outside `java.lang.invoke` illegal. They could've stopped here.

As the annotation has retention `RUNTIME` it must be present in `MethodHandle.class`.

I was surprised to see even `RUNTIME` as the annotation is intended (as per javadoc) to aid "the Java compiler", but fair enough, likely the VM needs it, too.

But then, in `--release` mode, `MethodHandle.class` isn't even used, but some `MethodHandle.sig`. 

Seeing lack of file `PolymorphicSignature.sig` in `ct.sym` could be understood as making use of this annotation in user code doubly illegal (_why?_).

But seeing methods like `invokeExact()` defined in `MethodHandle.sig` *without any annotation* (source code has two), is - weird.

As a result, the tool (java compiler) that is the main client of the annotation cannot see the annotation in `--release` mode.

If only `PolymorphicSignature.sig` was missing from `ct.sym` I _would_ recommend to interpret the name even if it cannot be resolved, but as even the annotation reference in `invokeExact()` and friends is missing, we don't seem to be allowed to do what the spec requires us to do.

In this light, the find in [jls-15.12.3](https://docs.oracle.com/javase/specs/jls/se11/html/jls-15.html#jls-15.12.3) is precious! :)  That's the best possible kludge we can clutch at.

I only see one downside in the deduction strategy: ecj gets more fragile vis-a-vis changes in JLS (that may or may not be announced), whereas reading a manifest annotation would be future proof.

### Implementation

**Nitpick**: couldn't the invocation of `deduceSignaturePolymorphicAnnotation` be pushed under the guard of  `if (parameterCompatibilityLevel(...)...)`?

**Completeness**: Some more code locations rely on `TagBits.AnnotationPolymorphicSignature`, some of which lead to invoking the same `LE.createPolymorphicMethod()`. I would assume that all these need the on-demand deduction, notably `Scope.findExactMethod()` and `BindingKeyResolver.consumeMethod()`. The distance between `Scope` and `BindingKeyResolver` makes me think the new method should likely be defined in `MethodBinding`, as a transparent query `hasPolymorphicSignature()` or such, WDYT? (just make sure we won't confuse it vs. `MethodBinding.isPolymorphic()`.)

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/BatchCompilerTest2.java
```diff
@@ -438,4 +438,34 @@ public void testIssue114() {
...
+			"				MethodHandle methodHandle = MethodHandles.lookup().unreflect(method );\n" +
+			"				return methodHandle.invoke(self);\n" +
+			"			} catch (IllegalArgumentException e) {\n" +
+			"				throw e;\n" +
+			"			} \n" +
+			"		}\n" +
+			"}\n"
+		},
+		"\"" + OUTPUT_DIR +  File.separator + "X.java\"" +
+				" --release " + CompilerOptions.VERSION_11 + " ",
```
stephan-herrmann, line 464: do you want to say VERSION_9 to be sure?

org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/Scope.java
```diff
@@ -868,6 +868,9 @@ protected final MethodBinding computeCompatibleMethod(MethodBinding method, Type
 			if (CompilerOptions.tolerateIllegalAmbiguousVarargsInvocation && compilerOptions.complianceLevel < ClassFileConstants.JDK1_7)
 				tiebreakingVarargsMethods = false;
 		}
+
+		deduceSignaturePolymorphicAnnotation(method);
+
 		if ((parameterCompatibilityLevel(method, arguments, tiebreakingVarargsMethods)) > NOT_COMPATIBLE) {
 			if ((method.tagBits & TagBits.AnnotationPolymorphicSignature) != 0) {
```
stephan-herrmann, line 875: if `deduceSignaturePolymorphicAnnotation` stays in Scope, then invocation between the two `if`s would be my recommendation.

stephan-herrmann (commented):
2 nitpicks as mentioned in the detailed comment :)

sravanlakkimsetti:
Thanks @stephan-herrmann for the detailed review. I will work on the changes required.

sravanlakkimsetti:
Created `hasPolymorphicSignature()` in `MethodBinding` to identify whether we have polymorphicSignature. Based on the result the tagbits are applied again.

sravanlakkimsetti:
@stephan-herrmann Can you please take a look again?

Thanks

org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/MethodBinding.java
```diff
@@ -1464,4 +1464,38 @@ public void updateTypeVariableBinding(TypeVariableBinding previousBinding, TypeV
...
+	}
+	if (this.isNative()	&& this.isVarargs() && this.parameters.length == 1) {
+		/*
+		 *  here type will be arrayType we will come here only if the method is of type
+		 *  varargs(represented by arraytype) and with only one parameter.
+		 */
+		if (this.parameters[0].leafComponentType().id == TypeIds.T_JavaLangObject) {
+			ReferenceBinding declaringClassLocal = this.declaringClass;
+			if ((declaringClassLocal != null)
+					&&(declaringClassLocal.id == scope.getJavaLangInvokeMethodHandle().id||declaringClassLocal.id == scope.getJavaLangInvokeVarHandle().id)) {
```
jarthana, line 1493: Please format this line. We could use some spacing around the operators.

jarthana (approved):

jarthana:
Changes look good to me and in the interest of getting this in M1, let's release this changes once the test passes.

Stephan, I have verified your comments have been taken care of. If anything, we can take this up post M1.

stephan-herrmann:
> Stephan, I have verified your comments have been taken care of. If anything, we can take this up post M1.

Thanks.

For a second I was surprised that the new method doesn't set the tagBit before returning `true`, but then I saw that all callers do so, so all is fine!

sravanlakkimsetti:
> > Stephan, I have verified your comments have been taken care of. If anything, we can take this up post M1.
> 
> Thanks.
> 
> For a second I was surprised that the new method doesn't set the tagBit before returning `true`, but then I saw that all callers do so, so all is fine!

I had a thought about it. My reasoning is we are checking whether the signature is polymorphic or not. In my opinion we should not modify any parameters during the check. That's the reason I did not set the tagBit there.

PR #164, opened 2022-06-23
NumberFormatException: For input string: "17.8" #167

jarthana:
ClasspathJrtWithReleaseOption accepts any version that is of the format x.y. But it should only be accepted when it's of the format 1.x. When this condition is not met, we should throw a better error rather than failing with an exception.

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ModuleBuilderTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ModuleBuilderTests.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2020 IBM Corporation and others.
+ * Copyright (c) 2016, 2022 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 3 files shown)


jarthana:
Another note on the change: I also updated the JarPackageFragmentRoot to use versionToJdkLevel() instead of releaseToJDKLevel(), since we receive this value from compliance and it is safer to use the former.

iloveeclipse (approved):
LGTM

iloveeclipse:
@jarthana : the new test fails on Java 17 build: https://download.eclipse.org/eclipse/downloads/drops4/I20220815-1800/testResults.php

Could you please take a look?

jarthana:
Thanks Andrey for reporting.

The problem is, with JDK 17, the code that checks for the validity of the --release is not executed. This is because, before we create ClasspathJrtWithReleaseOption, we ask JavaCore.compareJavaVersions() whether the --release supplied is same as the current version and this in turn uses CompilerOptions.versionToJdkLevel(), which is more tolerant when it comes to handling versions. So, basically we figure --release 17.8 is same as --release 17 and ignore the value.

Now, we have two options:
1. Accept this and adjust this test. But this will be inconsistent with what we wanted from this bug fix.
2. Move the code that checks for --release validity further up so.

Any thoughts from Andrey or others?

iloveeclipse:
Jay, since the code works as expected (fails on "other" releases and doesn't use "release" on the same JDK), I would fix the test. Changing "17.8" to "16.8" in the new test fixes it.

WDYT.

Issue #94, opened 2022-05-23
[19] DOM AST support for JEP 405: Record Patterns


SarikaSinha:
Adding the DOM AST support for Record pattern.

SarikaSinha:
#85 JEP 405: Record Patterns

Work will start after we have compiler AST available

SarikaSinha:
[] ASTNode addition
[] ASTMatcher addition
[] ASTVisitor addition
[] DefaultASTVisitor addition
[] ASTConverter Change
[] RecordPattern node addition
[] Test

PR #174, opened 2022-06-24
RecordPattern DOM AST changes

SarikaSinha:
fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/94

## Author checklist


SarikaSinha:
@jarthana Even with latest JDT core source change, it gives the same Bad AST structure error.

jarthana:
Thanks Sarika, this brought out a bug in the source range for the pattern variables in record patterns. I will release a fix for this. However, this is not sufficient to make this test pass. The problem is we are having overlapping source ranges for patternVariable and patterns. I think we should split the pattern variable into type and name that way we can have the patterns list between them.

Issue #181, opened 2022-06-30
Extract ECJ compiler code from jdt.core to ecj bundle and make ecj a real plug-in


iloveeclipse:
Follow up on https://github.com/eclipse-platform/eclipse.platform.ua/issues/18#issuecomment-1164373052.

Currently JDT has LOT of hacks around compiler part (dedicated source folders, dedicated build files, dedicated deployment steps, dedicated validation bundle) - all because the standalone Eclipse Compiler for Java (ECJ) code is located inside the jdt.core project that contains additionally all the IDE specific extensions.

Ideally ECJ could be just extracted from the jdt.core bundle so it can be used by jdt.core and other bundles as a regular bundle.

This would remove most of the custom hacks around building, testing and shipping ecj we have today.

I will provide a draft idea, but I have no enough time/knowledge to perform all the build/maven magic required to get that working.
Technically **I believe** this should work without breaking existing jdt.core clients.

stephan-herrmann:
> Do you have any idea why `org.eclipse.jdt.compiler.apt` and `org.eclipse.jdt.compiler.tool` are _fragments_ of jdt.core? Just a hack to get the standalone part compiled separately?

I can only guess / infer from what I see in our history. Both fragments implement additional capabilities connected to the compiler that were introduced around Java 6 (JSR 269 and JSR 199), right? As such the had to set BREE 1.6, while the rest of JDT/Core was still at 1.4 (!). That's why the implementation *had* to be kept separate.

*If* that was the only reason, then the reason is gone in times where JDT/Core requires 11.

For authoritative answers you'd have to ask Walter Harley or Olivier Thomann, I believe. I haven't heard from Walter for ages, but Olivier might still respond?

guw:
+1 for merging the fragments into a new ECJ bundle. They make ECJ a better offering in the Java ecosystem.

My hypothesis is that they are separate because they have been created long time ago while JDT was much more strict with who and what goes into jdt.core.

stephan-herrmann:
> My hypothesis ...

The JDK 6 dependency was a sufficient reason to keep those fragments separate (see my previous comment). I just wasn't sure if that was the only reason.

As of today I cannot think of any reason against merging.

jarthana:
> @stephan-herrmann , @jarthana : WDYT about #182 ?

[...]

PR #182, opened 2022-06-30
[DRAFT] An idea how to restructure JDT core to get proper ecj bundle

iloveeclipse:
The org.eclipse.jdt.core.ecj.validation was a "dummy" bundle so far,
used only to validate compilation issues in IDE.

That one should be renamed (org.eclipse.jdt.core.ecj.validation ->
org.eclipse.jdt.core.ecj) and be a proper maven library.

It is actually the ecj compiler library without any dependencies, that
could be consumed by JDT and the rest of the world.

It must be required and re-exported by JDT core.
Unfortunately, there are two split packages:

org.eclipse.jdt.internal.compiler
org.eclipse.jdt.internal.compiler.parser

Not sure what's the best way to deal with it, may be rename those
smaller packages in in jdt.core to avoid split packages issue, they are
internal, so it shouldn't break many clients (but most likely will, like
Xtext), but I haven't tried it and don't know if they use package
protected API.

org.eclipse.jdt.compiler.apt and org.eclipse.jdt.compiler.tool were
fragments of jdt.core, now they would be fragments of
org.eclipse.jdt.core.ecj, but this is problematic.

This breaks tests/clients that required jdt.core before, it seems that
by re-exporting the ecj host API, jdt core doesn't re-export the
fragments of ecj. Not sure why both fragments were fragments before,
might be we could just put them into the ecj bundle directly.

So now rg.eclipse.jdt.compiler.apt and org.eclipse.jdt.compiler.tool are
just bundles. We could merge them together into ecj, I honestly don't
know why they were fragments.

What I did NOT tried is to re-write all the magic scripts that build and
package separated ecj library out of jdt.core.

But this is just an idea how the "real ecj" could look like.

See
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/181
- https://github.com/eclipse-platform/eclipse.platform.ua/issues/18

## Author checklist


guw:
@iloveeclipse +1 for moving the fragments **into** the ECJ bundle. They offer additional functionality to make the ECJ better integrate with the Java ecosystem. Technically they are not required for the IDE. However, they do make ECJ a better offering with those APIs and should be shipped together. It doesn't make sense to have them be separate fragments.

My hypothesis is that they are separate because they have been created long time ago while JDT was much more strict with who and what goes into `jdt.core`.

iloveeclipse:
OK, got rebase working with `git rebase -s subtree master` (so git considered moved subtrees). At least in the IDE still no errors...

laeubi:
@iloveeclipse I have no idea why but the unit `org.eclipse.jdt.core.compiler.batch` is explicitly _excluded_ from the target resolution here:
https://github.com/eclipse-jdt/eclipse.jdt.core/blob/53f5b7bdd17ceee2710becccc8db75eaed678799/pom.xml#L97-L101

if I remove this the build can resolve ecj.core...

iloveeclipse:
This is probably because it was "artificially" generated **out of** jdt core and other bundles. OK, I will remove that later to see where the build will fail next.

mpalat:
Hey Andrey,
I know its a little late to comment/ask info, but neverthless - better late than never..comments/queries inline.
> It is actually the ecj compiler library without any dependencies, that could be consumed by JDT and the rest of the world.
First of all, I believe this is indeed a great initiative to bring this structure to JDT.. Thank you!

> It must be required and re-exported by JDT core. Unfortunately, there are two split packages:
Here, the query is what did you mean by required by JDT Core? I understand this as you meant that the ecj (the standalone part) to be required by *Other Parts of JDT Core - like Search/Content-Assist/DOM etc Part of the JDT Core as we know now* and then re-exported by this part. Is this understanding correct?

iloveeclipse:
@mpalat : sorry I haven't updated the patch here on last commit, below is the last state I've commented on #181. 

The split package is solved by reexporting packages, and yes, your understanding is correct.

**Update**

With the help from @laeubi I was able to fix first round of maven errors.

**Something** was built, so that even (all gerrit?) tests were executed.

So far only 24 failures in [org.eclipse.jdt.apt.pluggable.tests](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-182/11/testReport/org.eclipse.jdt.apt.pluggable.tests/).

The same tests run fine in the IDE, and I don't see clear pointer why don't they run in maven.
If anyone has an idea, what is special about org.eclipse.jdt.apt.pluggable.tests, don't hesitate to leave a comment.

laeubi:
@stephan-herrmann can you help with the tests maybe?

I see in one test:

> junit.framework.AssertionFailedError: Processor did not run

this seems that on the maven setup something is either missing (e.g. "the" Processor) or such an item has unresolved requirements and thus can't be loaded.

@iloveeclipse I don't know if there are maybe any resources relevant here, there is a stupid behaviour in PDE that it includes **all resources** in the IDE, but maintains a build.properties what resources are included when the bundle is build... so it might be that a resource that is required here is missing in the build-jar.

Beside that, maybe it could reveal some issues if one is not using jdt.core but the compiler bundle directly where appropriate? But thats just a wild guess...

org.eclipse.jdt.core.compiler.batch/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,32 @@
...
+Bundle-ManifestVersion: 2
+Bundle-Name: Eclipse Compiler for Java(TM)
+Bundle-SymbolicName: org.eclipse.jdt.core.compiler.batch
+Bundle-Version: 3.32.0.qualifier
+Bundle-ClassPath: .
+Bundle-Vendor: Eclipse.org
+Automatic-Module-Name: org.eclipse.jdt.core.compiler.batch
+Bundle-ActivationPolicy: lazy
+Bundle-RequiredExecutionEnvironment: JavaSE-11
+Export-Package: org.eclipse.jdt.core,
```
laeubi, line 12: I think at least the public packages should get a version, (e.g. 1.0.0) and include a use-clause (you can calculate this on the export package tab).

org.eclipse.jdt.core.compiler.batch/build.properties
```diff
@@ -16,7 +16,6 @@ source.. = src/
 output.. = bin/
 bin.includes = META-INF/,\
                .,\
-               about.html,\
-               compiler_tool_fragment.properties
+               about.html
 src.includes = about.html
-jars.extra.classpath = lib/javax18api.jar
+jars.extra.classpath = lib/javax18api.jar,platform:/plugin/org.apache.ant/lib/ant.jar
```
laeubi, line 21: Why do the compile requires ant?

org.eclipse.jdt.core.compiler.batch/pom.xml
```diff
@@ -0,0 +1,281 @@
...
+
+  <properties>
+    <defaultSigning-excludeInnerJars>true</defaultSigning-excludeInnerJars>
+    <code.ignoredWarnings>-warn:+fieldHiding,-unavoidableGenericProblems</code.ignoredWarnings>
+  </properties>
+
+  <build>
+    <plugins>
+    <plugin>
+		<artifactId>maven-antrun-plugin</artifactId>
```
laeubi, line 31: @akurtakov can you maybe help here so we replace this by maven resource filtering? seems more suitable than an ant-run ... maybe similar to

https://github.com/eclipse-tycho/tycho/blob/master/tycho-core/pom.xml#L30

akurtakov: It's in my todo list to look at this one at some point.

org.eclipse.jdt.core.compiler.batch/build.properties
iloveeclipse: There was/is an ant adapter in jdt core, but now it can be removed from batch plugin - I will try to do so in a moment

iloveeclipse: > There was/is an ant adapter in jdt core, but now it can be removed from batch plugin - I will try to do so in a moment

Scratch that, the thing **is** in batch compiler in the root package, see `JDTCompilerAdapter`

org.eclipse.jdt.core.compiler.batch/META-INF/MANIFEST.MF
iloveeclipse: Let discuss that later, to not lost the main focus. AFAIK this is unusual for Eclipse bundles to version packages **additionally** to the bundle version. Note: the bundle **is not new** (it was just constructed"on the fly by ant"), it is already published under same id, so version is not 1.0.0 anyway.

laeubi: This is not additionally. Currently the package is completely _unversioned_ that means every consumer must use require bundle, see for example: 

https://docs.osgi.org/whitepaper/semantic-versioning/010-executive-summary.html

so it seems a good opportunity to start with that.

org.eclipse.jdt.core.compiler.batch/build.properties
laeubi: Do sou think the Ant part can be removed from the batch compiler and instead be part of jtd.core, or maybe even better an jdt.compiler.ant module? I think having compiler.batch as standalone and technique agnostic as possible would be of great value.

org.eclipse.jdt.core.compiler.batch/META-INF/MANIFEST.MF
iloveeclipse: > it seems a good opportunity to start with that

Let us first get that thing working, OK? After that feel free to propose any improvements. I just don't want to loose focus here and don't have extra time anyway. My goal is to do whatever is **needed** first, ideally in 4.27 M1.

laeubi: Sure, I just found that such things often help to find the culprit, but we can leave that out for sure as I have no idea why it should make a difference.

iloveeclipse:
I *believe* the problem with [org.eclipse.jdt.apt.pluggable.tests](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-182/11/testReport/org.eclipse.jdt.apt.pluggable.tests/) could be that the org.eclipse.jdt.apt.pluggable.core imported wrong packages from old apt fragments of jdt core because the packages merged into batch compiler were not re-exported to jdt.core.
I will push a patch in a moment to see if tuning dependencies a bit helps.

org.eclipse.jdt.core.compiler.batch/META-INF/MANIFEST.MF
iloveeclipse: The problem here is that we have split packages / fragments / whatsoever dependencies with existing jdt core code that might cause unexpected side effects. Create a task if you want to follow up.

iloveeclipse:
Hmm, now it says
```
junit.framework.AssertionFailedError: apt.pluggable.tests plug-in was not found in project factory path
	at junit.framework.Assert.fail(Assert.java:57)
	at junit.framework.Assert.assertTrue(Assert.java:22)
	at junit.framework.TestCase.assertTrue(TestCase.java:192)
	at org.eclipse.jdt.apt.pluggable.tests.InfrastructureTests.testFactoryPathContents(InfrastructureTests.java:91)
```
and I can reproduce in the IDE if I **exclude** compiler.apt fragment from the target platform.

Same with test below
```
expected:<NO[ ERRORS]> but was:<NO[T RUN]>
	at org.eclipse.jdt.core.tests.junit.extension.TestCase.assertStringEquals(TestCase.java:266)
	at org.eclipse.jdt.core.tests.junit.extension.TestCase.assertEquals(TestCase.java:242)
	at org.eclipse.jdt.apt.pluggable.tests.FilerTests.testNoCollisionInSameModule(FilerTests.java:453)
```

So that thing somehow can't find re-exported packages at runtime. batch plugin re-exports them to jdt core and it is fine for compilation, but that seem to be not enough at runtime.

stephan-herrmann:
> junit.framework.AssertionFailedError: apt.pluggable.tests plug-in was not found in project factory path

Have a look at `org.eclipse.jdt.apt.core.internal.AptPlugin.canRunJava6Processors()`: this line no longer works:
```
		return Platform.getBundle("org.eclipse.jdt.compiler.apt") != null; //$NON-NLS-1$
```
Perhaps it can just be removed for good, given that we no longer support ecj on Java < 6.

iloveeclipse:
> Have a look at `org.eclipse.jdt.apt.core.internal.AptPlugin.canRunJava6Processors()`

Yep, that was it.

iloveeclipse:
Cool, no test fails, let merge for RC1 !!! :-)

Now the boring part starts: 

# understand which non-Java resources from JDT.core should be also copied or moved into batch compiler
# polish probably poms and check that we produce "same" bundles
# understand which ant, maven and script build steps are not needed / must be adopted in JDT/SDK area for the new **real** bundle.
# check if the SDK resulted from that can be used as IDE
# check if the ECJ resulted from that can be used on command line / ant / maven / tycho compilation
# TBC, whatever else.

iloveeclipse:
image

iloveeclipse:
**@ALL** : if **this** state builds & tests are OK would it make sense to merge and "break" SDK build? 

This would at least make sure everyone else interested to get things done in SDK would need to help here.

akurtakov:
Yes, please. Now is the perfect time for that work.

iloveeclipse:
OK, the build is green, I will merge that now, with ~100% probability to break SDK build.
I would like to trigger SDK build, but I didn't find it anymore reading https://github.com/eclipse-platform/eclipse.platform.releng.aggregator and looking at https://ci.eclipse.org/releng/view/Builds/

laeubi:
@iloveeclipse why do you think this will break SDK build (and what needs to be done/changed to mitigate this?).

iloveeclipse:
Because batch compiler / ecj jar had extra treatment and I'm pretty sure some of "extra" steps missing we don't run in Jenkins to "manually merge & copy & publish not-really-maven library to some SDK special places". 

E.g. there is a dedicated section for JDT Core Batch Compiler & source jars on download page: https://download.eclipse.org/eclipse/downloads/drops4/I20221201-0500/

I bet this will not work, as I have no idea where the final jars were produced and if the existing code finds them now in maven target directory.

iloveeclipse:
@akurtakov : I must be blind, I can't find where 4.27 build jobs are, I don't see them here: https://ci.eclipse.org/releng/view/Builds/
image

iloveeclipse:
Arrgh. That's somehow hidden. I will trigger one now.

akurtakov:
The jobs are autocreated from git content in https://ci.eclipse.org/releng/job/Builds/ and https://ci.eclipse.org/releng/job/Builds/ . Real structure instead of jenkins view.

stephan-herrmann:
First breakage can be fixed well inside JDT: https://github.com/eclipse-jdt/eclipse.jdt/blob/master/org.eclipse.jdt-feature/feature.xml

PR #199, opened 2022-07-07
Bug 576983 - fix ClassCastException for Types#asMemberOf of a Record

raouldh:
Bug 576983 - fix ClassCastException for Types#asMemberOf of a RecordComponent. Cast to VariableBinding instead.

A ClassCastException occurs when binding a member of a Java Record component to a FieldBinding, cast to supertype of FieldBinding instead(VariableBinding). See bugreport.

With this patch applied to my Eclipse installation I can now open all my Java Record classes without the ClassCastException.

### DIFF
--- a/org.eclipse.jdt.compiler.apt/src/org/eclipse/jdt/internal/compiler/apt/model/TypesImpl.java
+++ b/org.eclipse.jdt.compiler.apt/src/org/eclipse/jdt/internal/compiler/apt/model/TypesImpl.java
@@ -45,6 +45,7 @@ import org.eclipse.jdt.internal.compiler.lookup.ParameterizedTypeBinding;
 import org.eclipse.jdt.internal.compiler.lookup.ReferenceBinding;
 import org.eclipse.jdt.internal.compiler.lookup.TypeBinding;
 import org.eclipse.jdt.internal.compiler.lookup.TypeVariableBinding;
+import org.eclipse.jdt.internal.compiler.lookup.VariableBinding;
 
 /**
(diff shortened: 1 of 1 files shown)


filiphr:
I am the project lead of MapStruct. We have received multiple different bug reports about certain problems around this problem in our project.

Is it possible for someone from the JDT team to look into this PR? It would be very appreciated.

cc: @jarthana, @mpalat, @iloveeclipse

iloveeclipse:
@filiphr : I would recommend to provide a test case that demonstrates the problem and the fix, see for example a5e6a8967df6ce6eb6b2226d37eaefd8a33f0bd4 or 12f5e79231520db7114560416c2bd1714317f4c9.

I personally have no idea about this code, so I can't help you here.

filiphr:
good point @iloveeclipse. @raouldh can you look into adding a test case that would fail without your fix?

gzsombor:
I've added a test case, because I need this fix as well ... :wink:

gzsombor:
Can anyone review or merge my patch?

jarthana:
The patch looks good to me. But looks the like the testcase commit is not in the same branch and hence can't squash/merge. Can you take care of that please?

gzsombor:
> The patch looks good to me. But looks the like the testcase commit is not in the same branch and hence can't squash/merge. Can you take care of that please?

The fix with the test is on a different pull request, as I - of course - couldn't modify someone else branch: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/321

Issue #183, opened 2022-07-01
[17] Compilation problem when running in parallel with Java 17


laeubi:
We currently discovered [a problem in Tycho](https://github.com/eclipse/tycho/issues/1099) that makes compilation failing on Java 17 with error:

`The type java.lang.Object cannot be resolved`

Current reproducer:

- git clone https://github.com/eclipse/tycho.git
- cd tycho
- Make sure java -version shows 17
- mvn clean package **-T1** -DskipTests -> **succeds**
- mvn clean package **-T16** -DskipTests -> **fails** (but not always at the same step)

Using java 11 the build succeeds!

Failure message include:

- String cannot be resolved to a type
- The type java.lang.Object cannot be resolved. It is indirectly referenced from required .class files
- The import java.util cannot be resolved
- Implicit super constructor Object() is undefined for default constructor. Must define an explicit constructor

laeubi:
It might be necessary to configure a Java 11 JDK in the `<userhome>/.m2/toolchains.xml`:

```
<?xml version="1.0" encoding="UTF-8"?>
<toolchains>
	<toolchain>
		<type>jdk</type>
		<provides>
			<id>JavaSE-11</id>
			<version>11</version>
		</provides>
		<configuration>
			<jdkHome>/path/to/jdk-11</jdkHome>
		</configuration>
	</toolchain>
	<toolchain>
		<type>jdk</type>
		<provides>
			<id>JavaSE-17</id>
			<version>17</version>
		</provides>
		<configuration>
			<jdkHome>/path/to/jdk-17</jdkHome>
		</configuration>
	</toolchain>
</toolchains>
```

stephan-herrmann:
What options does tycho pass into ecj? Is `--release` included? For which target version?

laeubi:
We are calling the compiler with the follwoing arguments:
```
-properties <projectdir>/.settings/org.eclipse.jdt.core.prefs
 -s <classes output dir>generated-sources/annotations
 -d <classes output dir>
 -classpath <classpath>
<classes output dir>
<a java file>
 -g
 -nowarn
 -target 11
 -source 11
 --release 11
 -encoding UTF-8
```
It might be an issue in how tycho is calling the compiler, but strange enough it reliable fails when using JDK17 but works with JDK11 and I'm not familiar enough with EJC to tell if it is a Tycho or JDT issue, so any hint would be appreciated!

laeubi:
[...]

PR #213, opened 2022-07-14
Fix the init of FileSystem in various places

laeubi:
## What it does
Currently there is a possible race if two or more threads compete for
creating the filesystem for the same URI. This change uses a
helper method to simply fetch the filesystem again if it was created in the meanwhile.

## How to test
See for example test case here: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/183#issuecomment-1183979985

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/batch/org/eclipse/jdt/internal/compiler/batch/ClasspathJep247Jdk12.java
+++ b/org.eclipse.jdt.core/batch/org/eclipse/jdt/internal/compiler/batch/ClasspathJep247Jdk12.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2019, 2021 IBM Corporation.
+ * Copyright (c) 2019, 2022 IBM Corporation.
  * All rights reserved. This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License v2.0
  * which accompanies this distribution, and is available at
@@ -7,15 +7,13 @@
  *
  * Contributors:
  *     IBM Corporation - initial API and implementation
+ *     Christoph Läubrich - use Filesystem helper method
  *******************************************************************************/
 package org.eclipse.jdt.internal.compiler.batch;
 
 import java.io.File;
 import java.io.IOException;
-import java.net.URI;
 import java.nio.file.DirectoryStream;
-import java.nio.file.FileSystemNotFoundException;
-import java.nio.file.FileSystems;
 import java.nio.file.FileVisitResult;
 import java.nio.file.FileVisitor;
 import java.nio.file.Files;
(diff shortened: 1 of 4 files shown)


laeubi:
@iloveeclipse WDYT?

org.eclipse.jdt.core/batch/org/eclipse/jdt/internal/compiler/batch/ClasspathJep247Jdk12.java
```diff
@@ -122,14 +125,22 @@ public void initialize() throws IOException {
...
+			try {
+				return FileSystems.getFileSystem(uri);
+			} catch (FileSystemNotFoundException fne) {
+				// create it
+				HashMap<String, ?> env = new HashMap<>();
+				try {
+					return FileSystems.newFileSystem(uri, env);
+				} catch (IOException e) {
+					// can't do anything useful then, will produce error below
+					return null;
```
iloveeclipse, line 138: One should catch java.nio.file.FileSystemAlreadyExistsException and return getFileSystem(). 

Also would be good to have similar error handling like in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/192 and not just silently hide original errors. Exception below doesn't provide the root cause.

iloveeclipse: note: there are few places where similar problem with newFileSystem exists, one should go over all of them

iloveeclipse: See constructor of CtSym, this can be also done without a map, JDK knows already created file systems.

laeubi: This FileSystems API is really a bit messy... According to the docs, even if the FS exits, it can throw exception because it was closed but that's provider specific ... So actually anything bad can happen any time, I actually wonder is there any advantage in using `FileSystem` here at all? Why not simply ZipFile?

laeubi:
I have now merged all the different impl into one helper method

iloveeclipse:
@laeubi : I'm not aware about any changes in the related JDT code, so I assume we just hit some new tycho issue:
```
08:29:41  [ERROR] Cannot resolve project dependencies:
08:29:41  [ERROR]   Software being installed: org.eclipse.jdt.core.tests.compiler 3.12.2000.qualifier
08:29:41  [ERROR]   Missing requirement: org.eclipse.jdt.core.tests.compiler 3.12.2000.qualifier requires 'osgi.bundle; org.eclipse.jdt.annotation [2.0.0,3.0.0)' but it could not be found
08:29:41  [ERROR] 
```

Note that the `org.eclipse.jdt.core.tests.compiler` really want **two** ` org.eclipse.jdt.annotation` bundles because these two provide different API's: https://github.com/eclipse-jdt/eclipse.jdt.core/blob/8b1f036ff06b8529f24054b1d7895755ab1bef5a/org.eclipse.jdt.core.tests.compiler/META-INF/MANIFEST.MF#L24

```
 org.eclipse.jdt.annotation;bundle-version="[1.1.0,2.0.0)";resolution:=optional,
 org.eclipse.jdt.annotation;bundle-version="[2.0.0,3.0.0)";resolution:=optional,
```

laeubi:
@iloveeclipse I didn't even know its valid nor would even consider such construct I think that is reported here https://github.com/eclipse/tycho/issues/1129 it seems we really need an [integration-test](https://github.com/eclipse/tycho/wiki#providing-an-integration-test) for this....

laeubi:
This should be fixed now, seems that GH uses my default mail address (what is not my eclipse mail address)

iloveeclipse (approved):

PR #217, opened 2022-07-14
Cleanup JrtFileSystem init code (fixes #216)

### LINKED ISSUE #216: Cleanup JrtFileSystem init  code
Biggest part of the https://github.com/eclipse-jdt/eclipse.jdt.core/pull/194 is not the actual change of classpath behavior in presence of `--release` argument, but generic cleanup of the JrtFileSystem init  code.

- There is code only required if JDT would run on Java 8 and not needed anymore because we run only on 11
- Some code duplication
- Really unnecessary complicated init sequence

So the plan is to extract most of the refactoring from https://github.com/eclipse-jdt/eclipse.jdt.core/pull/194 and apply to JDT without changing any compiler behavior.

iloveeclipse:
- Removed code only required if JDT would run on Java 8 (we run only on
11+)
- Removed some code duplication
- Fixed really unnecessary complicated init sequence
- added some basic tests

### DIFF
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/TestAll.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/TestAll.java
@@ -26,6 +26,7 @@ package org.eclipse.jdt.core.tests.compiler.regression;
 import java.util.ArrayList;
 
 import org.eclipse.jdt.core.tests.compiler.util.HashtableOfObjectTest;
+import org.eclipse.jdt.core.tests.compiler.util.JrtUtilTest;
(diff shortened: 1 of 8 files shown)


iloveeclipse:
Maven / Jenkins issue, looks like another build toolchain problem
https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-217/1/consoleFull
```
12:51:10  [ERROR] Cannot resolve project dependencies:
12:51:10  [ERROR]   Software being installed: org.eclipse.jdt.core.tests.compiler 3.12.2000.qualifier
12:51:10  [ERROR]   Missing requirement: org.eclipse.jdt.core.tests.compiler 3.12.2000.qualifier requires 'osgi.bundle; org.eclipse.jdt.annotation [2.0.0,3.0.0)' but it could not be found
```

PR #227, opened 2022-07-15
Consider external paths in search scope

### LINKED ISSUE #190: Search engine fails to find results in external classpath container with source attachment
(Issue imported from https://bugs.eclipse.org/bugs/show_bug.cgi?id=578102)

Using the JDT search engine to find references of an annotation type does not produce any findings in the required plug-ins of a plug-in project.

Attached is a JUnit [test project](https://github.com/eclipse-jdt/eclipse.jdt.core/files/9046858/search-engine-test.zip) that does the following things:

1. Create a plug-in project that requires another plug-in (which includes a reference to an annotation)
2. Open the required plug-in as a workspace project. When searching for the annotation reference, it is found successfully
3. Close the required project, so that the required plug-in resolves to a platform bundle. When searching for the annotation reference again, it can't be found.

The strange thing is that the references **are** found when extending the scope to the entire workspace (instead of the respective plug-in project).

gerking:
## What it does
Fixes #190

## How to test
Test case added in commit 2c8a3b3

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/JavaSearchScope.java
+++ b/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/JavaSearchScope.java
@@ -302,9 +302,10 @@ public void add(IJavaElement element) throws JavaModelException {
 private void add(String projectPath, String relativePath, String containerPath, boolean isPackage, AccessRuleSet access) {
 	// normalize containerPath and relativePath
 	containerPath = normalize(containerPath);
+	containerPath = convertInternalToExternalPath(containerPath);
 	relativePath = normalize(relativePath);
-	int length = this.containerPaths.length,
-		index = (containerPath.hashCode()& 0x7FFFFFFF) % length;
+	int length = this.containerPaths.length;
+	int index = (containerPath.hashCode() & 0x7FFFFFFF) % length;
 	String currentRelativePath, currentContainerPath;
 	while ((currentRelativePath = this.relativePaths[index]) != null && (currentContainerPath = this.containerPaths[index]) != null) {
 		if (currentRelativePath.equals(relativePath) && currentContainerPath.equals(containerPath))
@@ -370,6 +371,7 @@ public boolean encloses(String resourcePathString) {
 private int indexOf(String fullPath) {
 	// cannot guess the index of the container path
 	// fallback to sequentially looking at all known paths
+	fullPath = convertInternalToExternalPath(fullPath);
 	for (int i = 0, length = this.relativePaths.length; i < length; i++) {
 		String currentRelativePath = this.relativePaths[i];
 		if (currentRelativePath == null) continue;
@@ -396,9 +398,10 @@ private int indexOf(String fullPath) {
  *   4. (empty)
  */
 private int indexOf(String containerPath, String relativePath) {
+	containerPath = convertInternalToExternalPath(containerPath);
+	int length = this.containerPaths.length;
 	// use the hash to get faster comparison
-	int length = this.containerPaths.length,
-		index = (containerPath.hashCode()& 0x7FFFFFFF) % length;
+    int index = (containerPath.hashCode()& 0x7FFFFFFF) % length;
 	String currentContainerPath;
 	while ((currentContainerPath = this.containerPaths[index]) != null) {
 		if (currentContainerPath.equals(containerPath)) {
@@ -411,6 +414,29 @@ private int indexOf(String containerPath, String relativePath) {
 		}
 	}
 	return -1;
+}
+
+/**
+ * If the given path is internal but represents an external folder,
+ * converts it to the corresponding external path.
+ * No conversion takes place if the given path does not represent an external folder.
+ * @param given the given path to convert if necessary
+
+ * @return the external path that corresponds to the given path,
+ * or the given path itself if no conversion is necessary
+ */
+private String convertInternalToExternalPath(String given) {
+	IPath givenPath = new Path(given);
+	if (ExternalFoldersManager.isInternalPathForExternalFolder(givenPath)) {
+		IResource targetResource = JavaModel.getWorkspaceTarget(givenPath);
+		if (targetResource != null) {
+			IPath targetLocation = targetResource.getLocation();
+			if (targetLocation != null) {
+				return targetLocation.toOSString();
+			}
+		}
+	}
+	return given;
 }
 
 /*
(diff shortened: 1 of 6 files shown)


iloveeclipse:
I see that you had an example with unit test on the issue. Could you please convert this unit test to "regular" JDT unit test that demonstrates the problem and helps to catch a possible regression later? There are few tests already existing, see subclasses of `org.eclipse.jdt.core.tests.model.AbstractJavaSearchTests`.

gerking:
> Could you please convert this unit test to "regular" JDT unit test that demonstrates the problem and helps to catch a possible regression later?

Easier said than done. Creating a JDT unit test was pretty easy, and it successfully demonstrates the fix during a local JUnit run. But when running the test with Surefire, the test project (that resides in the runtime workspace) cannot resolve its classpath. In particular, the `org.eclipse.pde.core.requiredPlugins` entry fails to resolve, so that required bundles such as `org.eclipse.jdt.annotation` cannot be found. But to demonstrate the problem, this bug needs a plug-in project with required bundles.

Can you provide any hints on how to make bundles such as JDT Annotations resolvable by a test project? Is PDE available during the Surefire runs? Is the amount of launched target bundles somehow restricted?

I hope that this is not the same problem discovered here: https://bugs.eclipse.org/bugs/show_bug.cgi?id=578744

gerking:
> Is PDE available during the Surefire runs?

Obviously it was not, so I had to add PDE as a Surefire dependency to the `org.eclipse.jdt.core.tests.model` plugin. Commit 6ee074f. 

Another challenge was that the `simulateRestart()` routine of class `AbstractJavaModelTests` removes the listeners required by PDE (this is the same problem described by @stephan-herrmann in [bugzilla 314850](https://bugs.eclipse.org/bugs/show_bug.cgi?id=314850#c6). Thus if some test is running after another test triggered a workspace restart, PDE can't resolve the classpath of test projects anymore. Since PDE itself seems to provide no API to do a proper reset that would reestablish the required listeners (correct me if I'm wrong), I also had to modify the `AbstractJavaModelTests` to save/load those listeners before/after the workspace restart: Commit 4bc3718.

With these two additions, the test finally succeeds within Surefire. Interestingly the problem is not even reproducible with Surefire, so the test will always succeed regardless of whether the fix is present or not (false negative). But when exceuted with JUnit, the problem (and the fix) can be demonstrated reliably.

gerking:
> Could you please convert this unit test to "regular" JDT unit test that demonstrates the problem and helps to catch a possible regression later?

Done in commit 2c8a3b3. I finally managed to reproduce the bug without a PDE dependency. Accordingly, the problem is more general and occurs on external classpath containers in general. But interestingly, only those containers with a source attachment seem to be affected. Therefore my unit test reuses the TestContainer from `TestExternalLibContainerInitializer` and attaches a source path to it.

Fixed in commit 52ed5be by converting internal to external paths before checking the index.

iloveeclipse:
Looking into that, sorry I had more urgent issues to work on.
I see the last commit was a merge => I will rebase other two commits, squash & push now to see what the tests say on current master.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/JavaSearchScope.java
```diff
@@ -374,6 +374,7 @@ private int indexOf(String fullPath) {
 		String currentRelativePath = this.relativePaths[i];
 		if (currentRelativePath == null) continue;
 		String currentContainerPath = this.containerPaths[i];
+		currentContainerPath = convertInternalToExternalPath(currentContainerPath, fullPath);
```
iloveeclipse, line 377: I would avoid doing conversion on every container path in the loop during the `indexOf`, instead I would convert to the external paths immediately at creating the search containers and only convert once given path.
Same below.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/JavaSearchScope.java
```diff
@@ -413,6 +415,34 @@ private int indexOf(String containerPath, String relativePath) {
...
+private String convertInternalToExternalPath(String given, String reference) {
+	IPath givenPath = new Path(given);
+	if (ExternalFoldersManager.isInternalPathForExternalFolder(givenPath)) {
+		IPath referencePath = new Path(reference);
+		if (!ExternalFoldersManager.isInternalPathForExternalFolder(referencePath)) {
+			IResource targetResource = JavaModel.getWorkspaceTarget(givenPath);
+			if (targetResource != null) {
+				IPath targetLocation = targetResource.getLocation();
+				if (targetLocation != null)
+					return targetLocation.toString();
```
iloveeclipse, line 438: I believe we should use `toOSString()` here as that is what used in this class elsewhere for container paths

iloveeclipse (changes requested):
I will push changes in a moment

iloveeclipse:
With latest patch we see test fail below

```
junit.framework.ComparisonFailure: 
Unexpected search results.
ExternalType (not open) [in ExternalType.class [in p [in /home/jenkins/agent/workspace/eclipse.jdt.core-Github_PR-227/org.eclipse.jdt.core.tests.model/target/work/externalLib]]]

---------------------- ----------
 expected:<[ExternalType (not open) [in ExternalType.class [in p [in /home/jenkins/agent/workspace/eclipse.jdt.core-Github_PR-227/org.eclipse.jdt.core.tests.model/target/work/externalLib]]]]> but was:<[]>
	at org.eclipse.jdt.core.tests.junit.extension.TestCase.assertStringEquals(TestCase.java:267)
	at org.eclipse.jdt.core.tests.junit.extension.TestCase.assertEquals(TestCase.java:242)
	at org.eclipse.jdt.core.tests.model.AbstractJavaModelTests.assertSearchResults(AbstractJavaModelTests.java:677)
	at org.eclipse.jdt.core.tests.model.AbstractJavaModelTests.assertSearchResults(AbstractJavaModelTests.java:663)
	at org.eclipse.jdt.core.tests.model.AbstractJavaModelTests.assertSearchResults(AbstractJavaModelTests.java:660)
	at org.eclipse.jdt.core.tests.model.JavaSearchTests.testExternalFolder2(JavaSearchTests.java:646)
```

The reason is that  `JavaSearchScope.packageFragmentRoot()` now uses **resolved** external paths to find matching `IPackageFragmentRoot` and that doesn't work. I will update PR.

PR #203, opened 2022-07-10
Insufficient triggers for classpath validation

stephan-herrmann:
fixes #202

one-line fix, I still need to provide tests for it.

PR #230, opened 2022-07-15
ModuleBuilderTests.testAddExportsIllegal test is unstable

iloveeclipse:
Ensure marker order before assert

See https://github.com/eclipse-jdt/eclipse.jdt.core/pull/203


stephan-herrmann:
Thanks, @iloveeclipse and sorry I didn't think of sorting in the first place.

PR #231, opened 2022-07-16
[19] - Support search for record pattern

### LINKED ISSUE #215: [19] - Support search for record pattern
public class X {
	static void print(Object r) {
		if(r instanceof String s11) {
			s11.charAt(0);
		}
		if (r instanceof (Rectangle(ColoredPoint(Point(int x, int y), Color c),
		                         ColoredPoint lr) r11)) { //here
		  System.out.println("Upper-left corner: " + r11);
		}
	}
	public static void main(String[] obj) {
		print(new Rectangle(new ColoredPoint(new Point(0, 0), Color.BLUE), 
		                         new ColoredPoint(new Point(10, 15), Color.RED)));
	}
}
record Point(int x, int y) {}
enum Color { RED, GREEN, BLUE }
record ColoredPoint(Point p, Color c) {}
record Rectangle(ColoredPoint upperLeft, ColoredPoint lowerRight) {}``

Select r11 where comment says here. Do Command-Shift-G. It should select 1 reference in the next line.

Actual result : Please select a valid java name.

vik-chand:
## What it does

Fixes #215 

1) For selectionParser, for selectionNode to be found, the indentifierPtr needs to be decremented post local variable creation else this.indexOfAssistIdentifier() is negative and SelectionOnLocalName cant be created

2) For selection of component in record pattern scenario, the traverse function of InstanceofExpression is updated to traverse the record patterns too.

## How to test

Check JavaSearchBugs19Tests for testIssue215_001 and testIssue215_002 to get examples that has been fixed. Press Command-Shift-G  ( or Control-Shift-G) after selection r11 and xyz respectively. Rest of the test case tests other scenarios.  Most ( if not all)  test case would fail without this fix. 

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/ast/InstanceOfExpression.java
+++ b/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/ast/InstanceOfExpression.java
@@ -385,6 +385,14 @@ public void traverse(ASTVisitor visitor, BlockScope scope) {
 			this.type.traverse(visitor, scope);
 		}
 	}
+	if (this.pattern instanceof RecordPattern) {
+		if (((RecordPattern) this.pattern).patterns != null) {
+			Pattern[] rPatterns = ((RecordPattern) this.pattern).patterns;
+			for (Pattern pat : rPatterns) {
+				pat.traverse(visitor, scope);
+			}
+		}
+	}
 	visitor.endVisit(this, scope);
 }
 }
--- a/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/ast/InstanceOfExpression.java
+++ b/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/ast/InstanceOfExpression.java
@@ -379,14 +379,11 @@ public void tagAsUnnecessaryCast(Scope scope, TypeBinding castType) {
 public void traverse(ASTVisitor visitor, BlockScope scope) {
 	if (visitor.visit(this, scope)) {
 		this.expression.traverse(visitor, scope);
-		if (this.elementVariable != null) {
-			this.elementVariable.traverse(visitor, scope);
+		if (this.pattern != null) {
+			this.pattern.traverse(visitor, scope);
 		} else {
 			this.type.traverse(visitor, scope);
 		}
-	}
-	if (this.pattern !=null) {
-		this.pattern.traverse(visitor, scope);
 	}
 	visitor.endVisit(this, scope);
 }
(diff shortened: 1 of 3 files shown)


org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/parser/Parser.java
```diff
@@ -10928,9 +10928,10 @@ protected void consumeRecordPatternWithId() {
 	this.astPtr -= length;
 	// Identifier
 	char[] identifierName = this.identifierStack[this.identifierPtr];
-	long namePositions = this.identifierPositionStack[this.identifierPtr--];
-	TypeReference type = getTypeReference(0);
+	long namePositions = this.identifierPositionStack[this.identifierPtr];
 	LocalDeclaration local = createLocalDeclaration(identifierName, (int) (namePositions >>> 32), (int) namePositions);
+    this.identifierPtr--;
```
jarthana, line 10933: Just wondering if this can be wrapped into the previous statement retrieving the namePositions, like we do in other places, like this:
`long namePositions = this.identifierPositionStack[this.identifierPtr--];`

org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/ast/InstanceOfExpression.java
```diff
@@ -385,6 +385,14 @@ public void traverse(ASTVisitor visitor, BlockScope scope) {
 			this.type.traverse(visitor, scope);
 		}
 	}
+	if (this.pattern instanceof RecordPattern) {
```
jarthana, line 388: After the recent changes, I think it is possible to simply delegate to the this.pattern regardless of the pattern type. That way we can get rid of the elementVariable.traverse() call too.

jarthana (changes requested):

org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/ast/InstanceOfExpression.java
vik-chand: Thanks ! I updated accordingly. Looks much cleaner.

org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/parser/Parser.java
vik-chand: That will not work as indexOfAssistIdentifier will be negative and no valid Java element will be found.

jarthana (approved):

PR #107, opened 2022-05-29
Bug 529197 - Fix lambdas in constructor call in inner class referencing outer classes

coehlrich:
Based on https://git.eclipse.org/r/c/jdt/eclipse.jdt.core/+/128897 with changes addressing the comments and fixing the merge conflict

## Changes made on top of gerrit change

- Fix the merge conflict in `LambdaExpressionsTest`
- Keep the `mapSyntheticEnclosingTypes` only in `LambdaExpression`
- Move the `addSyntheticArgument` method call to `tagAsAccessingEnclosingInstanceStateOf` so that 
`mapSyntheticEnclosingTypes` loop can be merged with the `outerLocalVariables` loop
- Move the `synEAL` local variable to where it's used
- Replace the `enclosingType.equals(tmp)` call with a call to `TypeBinding.equalsEquals`
## Author checklist

- I have thoroughly tested my changes
- The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

PR #270, opened 2022-07-27
Bug 529197 - Fix lambdas in constructor call in inner class referencing outer classes #107

stephan-herrmann:
Fix adopted from gerrit with 2 rounds of review-improvements.

https://bugs.eclipse.org/529197


stephan-herrmann:
For previous discussion (mostly related to git workflows) see #107

Issue #269, opened 2022-07-27
SourceTypeBinding.superInterfaces() returns missing type for "cyclic" static imports


trancexpress:
To reproduce, import project from attached archive: [TypeHierarchyBugReproducer.zip](https://github.com/eclipse-jdt/eclipse.jdt.core/files/9199404/TypeHierarchyBugReproducer.zip)

Then try to compute the type hierarchy of ```test.Test```, observe that ```test.internal.TestImpl``` is not on the hierarchy according to the view.

This is the case, due to this stack trace exiting early:

```
        at org.eclipse.jdt.internal.core.hierarchy.HierarchyResolver.reportHierarchy(HierarchyResolver.java:607)
        at org.eclipse.jdt.internal.core.hierarchy.HierarchyResolver.resolve(HierarchyResolver.java:894)
        at org.eclipse.jdt.internal.core.hierarchy.IndexBasedHierarchyBuilder.buildForProject(IndexBasedHierarchyBuilder.java:265)
        at org.eclipse.jdt.internal.core.hierarchy.IndexBasedHierarchyBuilder.buildFromPotentialSubtypes(IndexBasedHierarchyBuilder.java:376)
        at org.eclipse.jdt.internal.core.hierarchy.IndexBasedHierarchyBuilder.build(IndexBasedHierarchyBuilder.java:166)
        at org.eclipse.jdt.internal.core.hierarchy.TypeHierarchy.compute(TypeHierarchy.java:323)
        at org.eclipse.jdt.internal.core.hierarchy.TypeHierarchy.refresh(TypeHierarchy.java:1319)
        - locked <0x00000001b44008c0> (a org.eclipse.jdt.internal.core.hierarchy.TypeHierarchy)
        at org.eclipse.jdt.internal.ui.typehierarchy.TypeHierarchyLifeCycle.doHierarchyRefresh(TypeHierarchyLifeCycle.java:331)
        at org.eclipse.jdt.internal.ui.typehierarchy.TypeHierarchyLifeCycle.doHierarchyRefreshBackground(TypeHierarchyLifeCycle.java:272)
        at org.eclipse.jdt.internal.ui.typehierarchy.TypeHierarchyLifeCycle$1.run(TypeHierarchyLifeCycle.java:225)
        at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```
(stack trace is from 4.21, where we initially observed the problem)

[...]

PR #275, opened 2022-07-28
Fix type hierarchy for cyclic static imports #269

trancexpress:
If two classes import each other with static imports, computing a type
hierarchy can run into problems. In particular:

1. CompilationUnitScope.checkAndSetImports() will try to resolve the
static import in the 1st type.
2. checkAndSetImports() for the 2nd type will try to resolve the static
import.
3. This will result in using the (currently) incomplete scope of the 1st
type.
4. Super types (interfaces in the case of #269) are not resolved
correctly, since the bindings of the 1st type are not yet fully
computed.

This change adjusts CompilationUnitScope.checkAndSetImports() to first
resolve non-static imports and then resolve static imports. This ensures
the statically imported types are resolved as correctly.

Fixes: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/269

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/TypeHierarchyTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/TypeHierarchyTests.java
@@ -15,9 +15,12 @@
  *******************************************************************************/
 package org.eclipse.jdt.core.tests.model;
 
+import static org.junit.Assert.assertNotEquals;
+
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Paths;
+import java.util.Arrays;
 import java.util.HashMap;
 
 import junit.framework.Test;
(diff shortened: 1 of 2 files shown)


org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/CompilationUnitScope.java
```diff
@@ -226,11 +228,35 @@ void checkAndSetImports() {
...
+		this.imports = resolvedNonStaticImports;
+	} else {
+		/*
+		 * We have static imports, those need to know the "normal" imports
+		 * in order to be resolved correctly. So temporarily set the imports
+		 * to the non-static ones, and resolve the static ones.
+		 * See: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/269
+		 */
+		this.imports = resolvedNonStaticImports;
+		ImportBinding[] resolvedStaticImports = getResolvedImports(importReference -> !importReference.isStatic());
```
iloveeclipse, line 244: is this a typo: `!importReference.isStatic()` ?

trancexpress: > is this a typo: `!importReference.isStatic()` ?

Yes.

iloveeclipse: Could you push the fix, I see lot of tests failing. Or will they fail anyway, because of the patch?

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/TypeHierarchyTests.java
```diff
@@ -3674,4 +3677,68 @@ public void testIndexQualificationBinaryNestedSubTypes_SearchForNestedSuperType(
 	}
 }
 
+/**
+ * @bug GitHub 269: Wrong type hierarchy computed for types with cyclic static imports.
+ * @see "https://github.com/eclipse-jdt/eclipse.jdt.core/issues/269"
+ */
+public void testBugGh269() throws Exception {
```
trancexpress, line 3684: Since we are now moved to GitHub, and everything is more cool and 50 times better, we have 2 bug trackers and will have 2 types if bug numbers. "Normal", old ones from the gone Eclipse bugzilla, and new ones from GitHub.

I don't know how to meaningfully name tests because of this. Should I ask if we have guidelines on cross platform dev, or do I assume nobody has even thought about this and just use whatever names I want?

iloveeclipse: Please ask, I assume no one cared so far. "Gh269" makes sense for me.

trancexpress:
I'll check the fails tomorrow, likely something broke. The code that is now moved to another method didn't look very robust (it does a lot of "low level" array modifications).

org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/CompilationUnitScope.java
```diff
@@ -240,10 +241,34 @@ void checkAndSetImports() {
 	resolvedImports[0] = getDefaultImports()[0];
 	int index = 1;
 
+	Predicate<ImportReference> isStaticImport = i -> i.isStatic();
+	Predicate<ImportReference> isNotStaticImport = Predicate.not(isStaticImport);
+
+	// GitHub 269: resolve non-static imports first, so that cyclic static imports can be resolved correctly
+	index = resolveImports(numberOfStatements, resolvedImports, index, isNotStaticImport);
```
trancexpress, line 248: This is ugly, but preserves the original code as much as possible. While still going over the non-static imports first, so that static imports can be properly resolved afterward.

The previous commit tried to do the same but less ugly, likely some check was not being done properly due to looping over incomplete imports somewhere in that code.

trancexpress: Seems like tests pass now. @iloveeclipse can you review?

iloveeclipse (commented):
Just small nits, the change is OK in general.

org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/CompilationUnitScope.java
```diff
@@ -240,16 +241,42 @@ void checkAndSetImports() {
...
+		}
+
 		char[][] compoundName = importReference.tokens;
 
 		// skip duplicates or imports of the current package
 		for (int j = 0; j < index; j++) {
 			ImportBinding resolved = resolvedImports[j];
 			if (resolved.onDemand == ((importReference.bits & ASTNode.OnDemand) != 0) && resolved.isStatic() == importReference.isStatic()) {
 				if (CharOperation.equals(compoundName, resolvedImports[j].compoundName)) {
-					continue nextImport;
```
trancexpress, line 252: Any reason this was not changed in your commit @iloveeclipse ?

iloveeclipse: > Any reason this was not changed in your commit @iloveeclipse ?

Yes, this is the **only** reason why the label was needed - to jump out the inner loop and continue with next import.

All other places were simple "continue" cases.

iloveeclipse: Please restore this

trancexpress: > Yes, this is the **only** reason why the label was needed - to jump out the inner loop 

Ah sorry, I missed the outer loop.

iloveeclipse: Me either...

Issue #286, opened 2022-08-02
AIOOBE in ProblemReporter.invalidType()


iloveeclipse:
Most recent SDK build as of today (but can be reproduced on 4.21 too).
Right click `org.eclipse.jdt.internal.compiler.batch.ClasspathJar.file` field and say "Open Call Hierarchy".
"Call Hierarchy" view will be empty but an error will be reported to the log.

Opening call hierarchy on **other** fields in same class seem to work fine.

```
eclipse.buildId=4.25.0.I20220801-1800
java.version=17.0.4
java.vendor=Eclipse Adoptium
BootLoader constants: OS=linux, ARCH=x86_64, WS=gtk, NL=en_US
Command-line arguments:  -data /data/4x_platform_workspace -os linux -ws gtk -arch x86_64

org.eclipse.jdt.ui
Error
Tue Aug 02 14:30:26 CEST 2022
Internal Error

java.lang.ArrayIndexOutOfBoundsException: Index 3 out of bounds for length 2
	at org.eclipse.jdt.internal.compiler.problem.ProblemReporter.invalidType(ProblemReporter.java:4915)
	at org.eclipse.jdt.internal.compiler.ast.TypeReference.reportInvalidType(TypeReference.java:593)
	at org.eclipse.jdt.internal.compiler.ast.ParameterizedQualifiedTypeReference.internalResolveLeafType(ParameterizedQualifiedTypeReference.java:223)
	at org.eclipse.jdt.internal.compiler.ast.ParameterizedQualifiedTypeReference.internalResolveType(ParameterizedQualifiedTypeReference.java:206)
	at org.eclipse.jdt.internal.compiler.ast.ParameterizedQualifiedTypeReference.resolveType(ParameterizedQualifiedTypeReference.java:472)
	at org.eclipse.jdt.internal.compiler.ast.TypeReference.resolveType(TypeReference.java:621)
	at org.eclipse.jdt.internal.compiler.lookup.SourceTypeBinding.resolveTypesWithSuspendedTempErrorHandlingPolicy(SourceTypeBinding.java:2803)
	at org.eclipse.jdt.internal.compiler.lookup.SourceTypeBinding.resolveTypesFor(SourceTypeBinding.java:2705)
	at org.eclipse.jdt.internal.compiler.lookup.SourceTypeBinding.methods(SourceTypeBinding.java:2276)
[...]

PR #289, opened 2022-08-03
Prevent AIOOBE in ProblemReporter.invalidType() #286

trancexpress:
This change adds index checks for ParameterizedQualifiedTypeReference
and ArrayQualifiedTypeReference in ProblemReporter.invalidType(),
similar to the existing index check for QualifiedTypeReference in the
same method.

Fixes: #286

## Author checklist


trancexpress:
@iloveeclipse I don't know if its worth the time to investigate the cause of the problem. There already is an index check in code right next to the code that throws AIOOBE, I guess we just do the same to prevent the exceptions.

It might take a few days to understand what the real problem is. If there are no test fails, at most we are preventing an OOB exception.

PR #299, opened 2022-08-09
Improve indexing performance for .jar in classpath.

jukzi:
Reuse open zip files while indexing until indexing is finished.

When multiple .java files are indexed which depend on the same .jar
libraries in the classpath then it is faster to only close those files
after indexing is done.

### DIFF
--- a/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/SourceIndexer.java
+++ b/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/SourceIndexer.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2016 IBM Corporation and others.
+ * Copyright (c) 2000, 2022 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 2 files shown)


jukzi:
I sampled the timing in a big workspace.
Before the change:
image
After the change:
image
=> I/O reduced. Indexing almost twice as fast.

jukzi:
unrelated build error (see https://github.com/eclipse-jdt/eclipse.jdt.core/pull/69)
18:07:06  Possible frozen test case
18:07:06  "main": TIMED_WAITING
18:07:06      java.base@18.0.2/java.lang.Object.wait(Native Method)
18:07:06      org.eclipse.jdt.internal.core.search.processing.JobManager.performConcurrentJob(JobManager.java:312)
18:07:06      org.eclipse.jdt.internal.core.search.indexing.IndexManager.waitForIndex(IndexManager.java:176)
...

please rebuild

iloveeclipse:
Jörg, here and in other PR's you've updated: could you please avoid merge commits and instead rebase your branch on master?

jukzi:
done. could you review please, Andrey?

rgrunber:
I can definitely confirm that the `getZipFile` call is completely gone but I'm not seeing any difference in the total time the Java Indexer thread spends running on project import.

 I was trying through JDT-LS so maybe that had some effect, but it should be importing projects in a similar manner. I tried on a pure Maven project (via m2e), and an Eclipse project that is set up via a target platform but there didn't seem to be a noticeable difference (after ~3 attempts both ways).

Issue #167, opened 2022-06-10
NumberFormatException: For input string: "17.8"


laeubi:
I'm recently getting this when perform a java search after a workspace crash:
```
java.lang.NumberFormatException: For input string: "17.8"
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.base/java.lang.Integer.parseInt(Integer.java:652)
	at java.base/java.lang.Integer.parseInt(Integer.java:770)
	at org.eclipse.jdt.internal.compiler.util.CtSym.getReleaseCode(CtSym.java:429)
	at org.eclipse.jdt.internal.core.builder.ClasspathJrtWithReleaseOption.initialize(ClasspathJrtWithReleaseOption.java:109)
	at org.eclipse.jdt.internal.core.builder.ClasspathJrtWithReleaseOption.<init>(ClasspathJrtWithReleaseOption.java:82)
	at org.eclipse.jdt.internal.core.builder.ClasspathLocation.forJrtSystem(ClasspathLocation.java:157)
	at org.eclipse.jdt.internal.core.search.matching.JavaSearchNameEnvironment.mapToClassPathLocation(JavaSearchNameEnvironment.java:267)
	at org.eclipse.jdt.internal.core.search.matching.JavaSearchNameEnvironment.computeClasspathLocations(JavaSearchNameEnvironment.java:200)
	at org.eclipse.jdt.internal.core.search.matching.JavaSearchNameEnvironment.<init>(JavaSearchNameEnvironment.java:94)
	at org.eclipse.jdt.internal.core.search.matching.IndexBasedJavaSearchEnvironment.create(IndexBasedJavaSearchEnvironment.java:27)
	at org.eclipse.jdt.internal.core.search.matching.MatchLocator.initialize(MatchLocator.java:1239)
	at org.eclipse.jdt.internal.core.search.matching.MatchLocator.locateMatches(MatchLocator.java:1277)
	at org.eclipse.jdt.internal.core.search.matching.MatchLocator.locateMatches(MatchLocator.java:1390)
	at org.eclipse.jdt.internal.core.search.matching.MatchLocator.locateMatches(MatchLocator.java:1509)
	at org.eclipse.jdt.internal.core.search.JavaSearchParticipant.locateMatches(JavaSearchParticipant.java:135)
	at org.eclipse.jdt.internal.core.search.BasicSearchEngine.findMatches(BasicSearchEngine.java:251)
[...]

PR #312, opened 2022-08-16
Fix failing test after fix for #167

jarthana:
## Author checklist


iloveeclipse:
@jarthana : I would prefer to merge this: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/311 , which will still run test on Java 17 even if we don't support Java 11 anymore.

jarthana:
> 



> @jarthana : I would prefer to merge this: #311 , which will still run test on Java 17 even if we don't support Java 11 anymore.

I was only thinking that the test would fail again if we run it with JDk 16 (of course, unlikely)
But what you are saying is fair point too. Let's go with your change since it's test only change anyway.

iloveeclipse:
OK, I've merged #311 .

PR #315, opened 2022-08-16
ecj: The import org.eclipse.jdt.core.Signature cannot be resolved

### LINKED ISSUE #314: ecj: The import org.eclipse.jdt.core.Signature cannot be resolved
image
Bug 513880 - [null][external] JDT null analysis "Illegal format in annotation" exception is ambiguous and unhelpful (#310)
@stephan-herrmann

stephan-herrmann:
fixes #314


jukzi:
long lives copy&paste ;-)
thanks.

stephan-herrmann:
> long lives copy&paste ;-)

copy & paste & **simplify** in this case :)

Alternatively, I could have pulled `Signature` into the realm of ecj, but that seemed like overkill. Plus I didn't want to even think about implications of splitting package `org.eclipse.jdt.core` between source folders (later via #181: between bundles).

PR #320, opened 2022-08-17
Implement CompilationUnit.hashCode to fix #319

### LINKED ISSUE #319: Slow "organize imports" - Memory Leak
image

I could just multiple times (but not deterministically) reproduce a very slow (>10s) "organize imports"  (https://bugs.eclipse.org/bugs/show_bug.cgi?id=573172) 
It was during editing eclipse with eclipse. 
Sampling with VisualVm:
 
image
Most time was spend in JavaElement.equals() during a HashMap.put()
That feels like a hash collision.

So i took a Heap Dump after the operation. ~ 524_288 items in JavaModelCache.childrenCache:
image
many keys where null:
image
how many? more then 495_000:
image
image
When i reproduced it again still only 30_000 elements in the hash table but JavaModelCache.childrenCache had doubled in size.
i.e. all new elements had the exact same hash key as previously. The old model elements had not been removed.
That's a memory leak.

The most nodes with the most collisions contained mainly LocalVariable. For example two variables "fullPath" collided
image

Both example instances refered to the same file (Path.java) that i was editing and the same source range 5768-5754:
image
image
But still they both have made their way into the same HashMap. So either there is a broken "equals()" contract somewhere? 

The "owner" referenced in CompilationUnit is of type CompilationUnit$3. i guess that is the anonymous WorkingCopyOwner class defined in org.eclipse.jdt.internal.core.CompilationUnit.getWorkingCopy(IProgressMonitor) (CompilationUnit.java:1007) which does not even define an equals() method.

So the working copy will never be removed from the map because it can not equal any other working copy and the map will grow with elements that are of "same" but not technically not equal() content.
The ongrowing number of hash collisions then produces the performance degradation.

Well the performance problem could probably resolved:
Since CompilationUnit.equals(Object) compares CompilationUnit.owner the owner's hashcode should just be used in hashCode() too.

But it would even more help if someone fixes the Leak.

jukzi:
To avoid hash collisions between JavaElements of different
WorkingCopyOwner in JavaModelCache.childrenCache.
Should improve slow Organize imports.

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/CompilationUnit.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/CompilationUnit.java
@@ -523,6 +523,12 @@ public boolean equals(Object obj) {
 	CompilationUnit other = (CompilationUnit)obj;
 	return this.owner.equals(other.owner) && super.equals(obj);
 }
+
+@Override
+public int hashCode() {
+	return Util.combineHashCodes(super.hashCode(), this.owner.hashCode());
+}
+
 /**
  * @see ICompilationUnit#findElements(IJavaElement)
  */


jukzi:
@jjohnstn  got time to review a one-liner?

stephan-herrmann:
> Implement CompilationUnit.hasCode

Even this tiny typo confuses readers of the git history. Yes, `hasCode()` would be a relevant query for a compilation unit!

jarthana:
> > Implement CompilationUnit.hasCode
> 
> Even this tiny typo confuses readers of the git history. Yes, `hasCode()` would be a relevant query for a compilation unit!

Good catch, Stephan! Sorry I missed that.

PR #326, opened 2022-08-18
Bug 576735 - "Type cannot be resolved. It is indirectly referenced from required .class files" error message not helpful

stephan-herrmann:
From https://bugs.eclipse.org/576735

This was actually an old WIP branch ... I'll just await tests to decide if I'll discard or improve.


stephan-herrmann:
#327 is the newer approach for this.

stephan-herrmann:
The test case from this PR has been adopted in #327

I think the newer approach (#327) of showing the referencing type is more helpful than forcing an error location (which only shows a trigger, not a reason for the problem).

=> closing.

PR #327, opened 2022-08-18
Bug 576735 - "Type cannot be resolved. It is indirectly referenced from required .class files" error message not helpful

stephan-herrmann:
From https://bugs.eclipse.org/576735

+ propagate the ReferenceBinding requesting a type lookup
+ include that type's qualified name in the error message
+ test case modeled with findings from https://bugs.eclipse.org/577019

Mostly, LookupEnvironment acts as the information hub, but also UnresolvedReferenceBinding is used, to remember, which .class file caused the URB to be created. In case the URB is later referenced also from source we can still report the binary file referencing this type.

While I doubt that this will cover all cases of "indirectly referenced" problems, but I hope it improves the message for a significant number of cases.


stephan-herrmann:
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=577019#c4 for a slightly expanded explanation of the relevance of URB creation & resolving.

stephan-herrmann:
I might call out that I'm upping jdt.core's version because the change adds new API in IProblem.

Actually I was reluctant to introduce a new constant, but these constants control the error message shown to users, and as I am not sure additional information is always available, I didn't see how I could abandon the old (poorer) error message.

Issue #332, opened 2022-08-22
[Java 11] Call hierarchy doesn't work for java.util.Timer.sched(TimerTask, long, long) 


trancexpress:
See: https://bugs.eclipse.org/bugs/show_bug.cgi?id=572989

```
1) Have a workspace with both Java 8 & Java 11 JDK's configured.
2) Open java.util.Timer class in each and select java.util.Timer.sched(TimerTask, long, long) method in the Outline view.
3) Right click -> Open Call Hierarchy.
4) In Java 8 JDK you will get a list with 6 callers inside same Timer class.
5) In Java 11 JDK you will get ... zero callers. At same time References -> Workspace menu shows all 6 callers in Timer class (for both JDK's).

I see same behavior on 4.15 & 4.20 master.
```

Its still seen with 4.25.

trancexpress:
Will have to double check, after rebasing the bug is still seen with the change... Maybe I was checking with the class from the wrong JRE...?

trancexpress:
Seems like the pipe is only added for jars.

iloveeclipse:
Verified with I20220919-0240

jukzi:
testModuleJarSearchBugGh332 fails on win
https://download.eclipse.org/eclipse/downloads/drops4/I20221010-1840/testresults/html/org.eclipse.jdt.core.tests.model_ep426I-unit-win32-java11_win32.win32.x86_64_11.html
```
junit.framework.AssertionFailedError: Unexpected search results:
C:\openjdk\jdk-11\lib\jrt-fs.jar void java.util.Timer.schedule(java.util.TimerTask, long)
C:\openjdk\jdk-11\lib\jrt-fs.jar void java.util.Timer.schedule(java.util.TimerTask, java.util.Date)
C:\openjdk\jdk-11\lib\jrt-fs.jar void java.util.Timer.schedule(java.util.TimerTask, long, long)
C:\openjdk\jdk-11\lib\jrt-fs.jar void java.util.Timer.schedule(java.util.TimerTask, java.util.Date, long)
C:\openjdk\jdk-11\lib\jrt-fs.jar void java.util.Timer.scheduleAtFixedRate(java.util.TimerTask, long, long)
C:\openjdk\jdk-11\lib\jrt-fs.jar void java.util.Timer.scheduleAtFixedRate(java.util.TimerTask, java.util.Date, long) expected:<6> but was:<1>
```

trancexpress:
[...]

PR #333, opened 2022-08-22
Call hierarchy doesn't work for Java 11 java.util.Timer.sched #332

trancexpress:
During the call hierarchy computation of Timer.sched() from JRE 11,
JavaSearchScope.encloses() is called with the following values:

path=java.base|java/util/Timer.class
enclosingPath=java/util

The check then fails, since the provided enclosing path does not enclose
the path of the type - its missing the module name.

This change adjusts JavaSearchScope to remove the module part of the
path, before checking whether the path is enclosed. In addition, a
warning is logged if JavaSearchScope.encloses(String, String, int) is
reached with a path that contains the symbol '|'.

Fixes: #332

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchScopeTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchScopeTests.java
@@ -15,23 +15,40 @@ package org.eclipse.jdt.core.tests.model;
 
 import java.net.MalformedURLException;
 import java.net.URL;
+import java.util.Arrays;
 import java.util.HashMap;
-
-import junit.framework.Test;
+import java.util.LinkedHashSet;
+import java.util.List;
+import java.util.Set;
 
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IWorkspaceRunnable;
 import org.eclipse.core.resources.IncrementalProjectBuilder;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.Path;
-import org.eclipse.jdt.core.*;
-import org.eclipse.jdt.core.search.*;
+import org.eclipse.jdt.core.IClasspathContainer;
+import org.eclipse.jdt.core.ICompilationUnit;
+import org.eclipse.jdt.core.IJavaElement;
+import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.jdt.core.IMethod;
+import org.eclipse.jdt.core.IPackageFragment;
+import org.eclipse.jdt.core.IPackageFragmentRoot;
+import org.eclipse.jdt.core.IType;
+import org.eclipse.jdt.core.JavaCore;
+import org.eclipse.jdt.core.search.IJavaSearchConstants;
+import org.eclipse.jdt.core.search.IJavaSearchScope;
+import org.eclipse.jdt.core.search.SearchEngine;
+import org.eclipse.jdt.core.search.SearchParticipant;
+import org.eclipse.jdt.core.search.SearchPattern;
 import org.eclipse.jdt.core.tests.model.AbstractJavaSearchTests.JavaSearchResultCollector;
 import org.eclipse.jdt.core.tests.model.AbstractJavaSearchTests.TypeNameMatchCollector;
 import org.eclipse.jdt.internal.core.JavaModelManager;
 import org.eclipse.jdt.internal.core.index.IndexLocation;
 import org.eclipse.jdt.internal.core.search.indexing.IndexManager;
+
+import junit.framework.Test;
 
 /**
  * Tests the Java search engine accross multiple projects.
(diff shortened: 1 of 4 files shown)


org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/JavaSearchScope.java
```diff
@@ -59,6 +60,7 @@ public class JavaSearchScope extends AbstractJavaSearchScope {
 	private int[] projectIndexes; // Indexes of projects in list
 	private String[] containerPaths; // path to the container (e.g. /P/src, /P/lib.jar, c:\temp\mylib.jar)
 	private String[] relativePaths; // path relative to the container (e.g. x/y/Z.class, x/y, (empty))
+	private String[] modulePaths; // path relative to the container (e.g. x/y/Z.class, x/y, (empty))
```
trancexpress, line 63: Will update the comment.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/JavaSearchScope.java
```diff
@@ -436,15 +444,26 @@ private boolean encloses(String enclosingPath, String path, int index) {
...
 		// or if the given path path is the folder path (see bug 13919 Declaration for package not found if scope is not project)
-		if (path.startsWith(enclosingPath)
-			&& ((enclosingPath.length() == path.lastIndexOf('/'))
-				|| (enclosingPath.length() == path.length()))) {
+		if (isEnclosing(enclosingPath, path)) {
 			return true;
 		}
+		String module = this.modulePaths[index];
+		if (module != null) {
+			String enclosingPathWithModule = module + org.eclipse.jdt.core.search.IJavaSearchScope.JAR_FILE_ENTRY_SEPARATOR + enclosingPath;
```
trancexpress, line 452: I'm guessing there are multiple cases here, since the pipe that we see is for jars. I'll check if the separator remains a pipe for workspace projects with modules.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/JavaSearchScope.java
```diff
@@ -239,19 +241,24 @@ public void add(IJavaElement element) throws JavaModelException {
...
 				String relativePath = Util.concatWith(((PackageFragment) element).names, '/');
+				IModuleDescription moduleDescription = root.getModuleDescription();
+				String moduleName = null;
+				if (moduleDescription != null) {
+					moduleName = moduleDescription.getElementName();
+				}
 				containerPath = root.getPath();
 				containerPathToString = containerPath.getDevice() == null ? containerPath.toString() : containerPath.toOSString();
-				add(projectPath, relativePath, containerPathToString, true/*package*/, null);
+				add(projectPath, relativePath, containerPathToString, moduleName, true/*package*/, null);
```
trancexpress, line 261: I'm guessing, more calls to ```add``` will need the module name passed. If we find more bugs due to including the module name in the path arguments here. For now I think its best to fix 1 thing at a time.

trancexpress:
Will try to add a test.

trancexpress:
> Will try to add a test.

Unfortunately the test passes even without the change. I tried to mirror how the search is done during the call hierarchy computation, the only difference is workspace vs project scope (I tried the workspace scope, same problem).

I'm not sure if I'm doing the search "wrong" or if the bug is in the search done by the call hierarchy computation...

trancexpress:
> I'm not sure if I'm doing the search "wrong" or if the bug is in the search done by the call hierarchy computation...

I missed this call in `org.eclipse.jdt.internal.corext.callhierarchy.CallerMethodWrapper.findChildren(IProgressMonitor)`:

```
	private IJavaSearchScope getAccurateSearchScope(IJavaSearchScope defaultSearchScope, IMember member) throws JavaModelException {
		if (! JdtFlags.isPrivate(member))
			return defaultSearchScope;

		if (member.getCompilationUnit() != null) {
			return SearchEngine.createJavaSearchScope(new IJavaElement[] { member.getCompilationUnit() });
		} else if (member.getClassFile() != null) {
			// member could be called from an inner class-> search
			// package fragment (see also bug 109053):
			return SearchEngine.createJavaSearchScope(new IJavaElement[] { member.getAncestor(IJavaElement.PACKAGE_FRAGMENT) });
		} else {
			return defaultSearchScope;
		}
	}
```

This actually replaces the workspace scope with a more accurate scope (in this case, the module which contains the type). After adding similar code to the test, the test fails too without the fix.

So IMO this can be reviewed. Having the test, maybe we can find a better fix.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchScopeTests.java
```diff
@@ -1124,4 +1141,68 @@ public void testBug397818() throws CoreException {
...
+				}
+			}
+			if (isFound) {
+				notFound.remove(expectedResult);
+			}
+		}
+		assertTrue("Unexpected search result\nExpected:\n" + String.join("\n", expectedResults) + "\nActual:\n" + foundReferences + "\nExpected but not found:\n" + String.join("\n", notFound),
+				notFound.isEmpty());
+	} finally {
+		JavaCore.setOptions(getDefaultJavaCoreOptions());
```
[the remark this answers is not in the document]
trancexpress: > Why is JavaCore.setOptions call needed? I don't see it above

The test is changing some options and fails on tear down because of that. I don't know which part of the test changes the options though. I can debug to find out if you want, to me it was simpler to reset the options.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/JavaSearchScope.java
iloveeclipse: Right, I've found #401 while experimenting with this patch.

iloveeclipse:
Bingo: I see now warnings, for example for this path: `/usr/lib/jvm/java-17/lib/jrt-fs.jar|java.base|java/util/SplittableRandom.class`
```
!ENTRY org.eclipse.jdt.core 2 4 2022-09-16 15:04:17.773
!MESSAGE Unexpected symbol '|' in path "java.base|java/util/SplittableRandom.class".
!STACK 0
java.lang.IllegalArgumentException
	at org.eclipse.jdt.internal.core.search.JavaSearchScope.encloses(JavaSearchScope.java:422)
	at org.eclipse.jdt.internal.core.search.JavaSearchScope.indexOf(JavaSearchScope.java:406)
	at org.eclipse.jdt.internal.core.search.JavaSearchScope.packageFragmentRoot(JavaSearchScope.java:605)
	at org.eclipse.jdt.internal.core.util.HandleFactory.getJarPkgFragmentRoot(HandleFactory.java:323)
	at org.eclipse.jdt.internal.core.util.HandleFactory.createOpenable(HandleFactory.java:96)
	at org.eclipse.jdt.internal.core.search.matching.MatchLocator.locateMatches(MatchLocator.java:1497)
	at org.eclipse.jdt.internal.core.search.JavaSearchParticipant.locateMatches(JavaSearchParticipant.java:135)
	at org.eclipse.jdt.internal.core.search.BasicSearchEngine.findMatches(BasicSearchEngine.java:251)
	at org.eclipse.jdt.internal.core.search.BasicSearchEngine.search(BasicSearchEngine.java:602)
	at org.eclipse.jdt.core.search.SearchEngine.search(SearchEngine.java:670)
	at org.eclipse.jdt.internal.corext.callhierarchy.CallerMethodWrapper.findChildren(CallerMethodWrapper.java:155)
	at org.eclipse.jdt.internal.corext.callhierarchy.MethodWrapper.performSearch(MethodWrapper.java:266)
	at org.eclipse.jdt.internal.corext.callhierarchy.MethodWrapper.doFindChildren(MethodWrapper.java:197)
	at org.eclipse.jdt.internal.corext.callhierarchy.MethodWrapper.getCalls(MethodWrapper.java:88)
	at org.eclipse.jdt.internal.ui.callhierarchy.DeferredMethodWrapper.getCalls(DeferredMethodWrapper.java:70)
	at org.eclipse.jdt.internal.ui.callhierarchy.DeferredMethodWrapper.fetchDeferredChildren(DeferredMethodWrapper.java:78)
	at org.eclipse.ui.progress.DeferredTreeContentManager$1.run(DeferredTreeContentManager.java:219)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```
This however doesn't cause "visible" regression, because we "only" find the right package fragment not immediately but later (as fallback code iterated over all projects in the workspace).

Fix for that warning is:
```
diff --git a/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/JavaSearchScope.java b/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/JavaSearchScope.java
index 82f370b..0665326 100644
--- a/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/JavaSearchScope.java
+++ b/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/JavaSearchScope.java
@@ -604,2 +604,7 @@
 		String relativePath = resourcePathString.substring(jarSeparatorIndex+1);
+		// Module in the jar
+		int moduleSeparatorIndex = relativePath.indexOf(JAR_FILE_ENTRY_SEPARATOR);
+		if (moduleSeparatorIndex != -1) {
+			relativePath = relativePath.substring(moduleSeparatorIndex+1);
+		}
 		index = indexOf(jarPath, relativePath);
```

Could you please apply it?

Issue #350, opened 2022-08-26
[null analysis]  `animal instanceof Dog dog` fails to assume dog as @NonNull


mickaelistria:
named instance of (eg `animal instanceof Dog dog`) do not assume the new variable (`dog` here) is `@NonNull`. It leads to false warning reports with null analysis.

iloveeclipse:
@stephan-herrmann : do you have some time/interest?

stephan-herrmann:
Interest? Sure. Time? Let's see. 
Might also be a good learning issue for someone new to this aspect of the code :)

BTW, we already have two bugs for this: 
- https://bugs.eclipse.org/bugs/show_bug.cgi?id=571933
- https://bugs.eclipse.org/bugs/show_bug.cgi?id=577181

stephan-herrmann:
GitHub forgot to close this issue, when #351 got merged

PR #351, opened 2022-08-29
[null] Assume instanceof local declaration non-null

mickaelistria:
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/350

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/NullAnnotationTests18.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/NullAnnotationTests18.java
@@ -541,4 +541,56 @@ public class NullAnnotationTests18 extends AbstractNullAnnotationTest {
 		runner.expectedOutputString = "0";
 		runner.runConformTest();
 	}
+
+	public void testInstanceOfPatternIsNonNull() {
+		Runner runner = getDefaultRunner();
+		runner.testFiles = new String[] {
+				"X.java",
+				"import org.eclipse.jdt.annotation.*;\n" +
+				"public class X {\n" +
+				"	public static void consumeNonNull(@NonNull String s) {\n" +
+				"		System.out.println(\"nonnull\");\n" +
+				"	}\n" +
+				"	public static void main(String... args) {\n" +
+				"		Object o = Math.random() < 0 ? new Object() : \"blah\";\n" +
+				"		if (o instanceof String message) {\n" +
+				"			consumeNonNull(message);\n" +
+				"		}\n" +
+				"	}\n" +
+				"}\n"
+			};
+		runner.expectedCompilerLog = "";
+		runner.expectedOutputString = "nonnull";
+		runner.runConformTest();
+	}
+
+	public void testInstanceOfPatternIsLaterAssignedNull() {
+		Runner runner = getDefaultRunner();
+		runner.testFiles = new String[] {
+				"X.java",
+				"import org.eclipse.jdt.annotation.*;\n" +
+				"public class X {\n" +
+				"	public static void consumeNonNull(@NonNull String s) {\n" +
+				"		System.out.println(\"nonnull\");\n" +
+				"	}\n" +
+				"	public static void main(String... args) {\n" +
+				"		Object o = Math.random() >= 0 ? new Object() : \"blah\";\n" +
+				"		if (o instanceof String message) {\n" +
+				"			consumeNonNull(message);\n" +
+				"			message = null;\n" +
+				"			consumeNonNull(message);\n" +
+				"		}\n" +
+				"	}\n" +
+				"}\n"
+			};
+		runner.expectedCompilerLog =
+				"----------\n" +
+				"1. ERROR in X.java (at line 11)\n" +
+				"	consumeNonNull(message);\n" +
+				"	               ^^^^^^^\n" +
+				"Null type mismatch: required \'@NonNull String\' but the provided value is null\n" +
+				"----------\n";
+		runner.runNegativeTest();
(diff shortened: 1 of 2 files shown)


mickaelistria:
I can see an error locally, that I'm failing to understand, when this change is applied:
```
org.eclipse.jdt.core.tests.compiler.regression.NullAnnotationTests18#testNamedInstanceOfIsNonNull
[ERR]:Error: LinkageError occurred while loading main class X
	java.lang.ClassFormatError: Illegal field name " instanceOfPatternExpressionValue" in class X
```
I tried to debug it and I don't get how the change I wrote here actually affects the resulting AST in that way.
@stephan-herrmann any idea which trap caught me here?

mickaelistria:
Please ignore former comment, the issue was a configuration one on my end. Things should be good enough for review now.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/NullAnnotationTests18.java
```diff
@@ -541,4 +541,28 @@ public void testBug576329() {
...
+		Runner runner = getDefaultRunner();
+		runner.testFiles = new String[] {
+				"X.java",
+				"import org.eclipse.jdt.annotation.*;\n" +
+				"public class X {\n" +
+				"	public static void consumeNonNull(@NonNull String s) {\n" +
+				"		System.out.println(\"nonnull\");\n" +
+				"	}" +
+				"	public static void main(String... args) {\n" +
+				"		Object o = Math.random() > 0.5 ? new Object() : \"blah\";\n" +
```
stephan-herrmann, line 556: Please don't create a test that may randomly fail. (yes, I saw it fail)

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/NullAnnotationTests18.java
```diff
@@ -541,4 +541,28 @@ public void testBug576329() {
...
+	@org.junit.Test
+	public void testNamedInstanceOfIsNonNull() {
+		Runner runner = getDefaultRunner();
+		runner.testFiles = new String[] {
+				"X.java",
+				"import org.eclipse.jdt.annotation.*;\n" +
+				"public class X {\n" +
+				"	public static void consumeNonNull(@NonNull String s) {\n" +
+				"		System.out.println(\"nonnull\");\n" +
+				"	}" +
```
stephan-herrmann, line 554: in case we observe a compile error (which I did with an experimental variant of this test), missing \n makes it hard to match line numbers.

stephan-herrmann (changes requested):
I assume we are targeting 4.26, right? (I remember a time when bug reports had a defined target milestone ...)

The direction looks good. 

For a second I was pondering, whether we should instead tag the local with `TagBits.AnnotationNonNull`, but that would actually be wrong as witnessed by this variant of the test:
```
if (o instanceof String message) {
	consumeNonNull(message);
	message = null; // OK
	consumeNonNull(message); // NOK
}
```

If we'd fake a `@NonNull` on the variable, we'd get an error on "OK", but by using the flowInfo, we indeed get an error only at "NOK", which is correct.
Might be a good idea to add the above as a negative test.

mickaelistria:
> I assume we are targeting 4.26, right?

Yes.

> (I remember a time when bug reports had a defined target milestone ...)

Only project committers can set the milestone on GitHub issues/PRs. So if you wish to add it, it would be welcome.

I'm working on implemented the improvements you suggest.

stephan-herrmann:
> Only project committers can set the milestone on GitHub issues/PRs. So if you wish to add it, it would be welcome.

last time I looked, no useful milestones were available, but I see this has been fixed for meanwhile.

mickaelistria:
I updated the patch as you suggested.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/NullAnnotationTests18.java
```diff
@@ -541,4 +541,54 @@ public void testBug576329() {
...
+				"		if (o instanceof String message) {\n" +
+				"			consumeNonNull(message);\n" +
+				"		}\n" +
+				"	}\n" +
+				"}\n"
+			};
+		runner.expectedCompilerLog = "";
+		runner.expectedOutputString = "nonnull";
+		runner.runConformTest();
+		//
```
stephan-herrmann, line 566: nitpick: we normally prefer one test per test method, unless there's a very good excuse like reusing lots of test data.

mickaelistria:
@stephan-herrmann anything else you think must be improved?

stephan-herrmann:
> @stephan-herrmann anything else you think must be improved?

It's in my queue of things to look into now that master is re-opened.

A successful build would certainly help.

mickaelistria:
> A successful build would certainly help.

I'll rebase so we can see how things work or not, but I have the impression that the master branch isn't really in a stable state, so the build failure may be coming from a previous patch.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/NullAnnotationTests18.java
stephan-herrmann: I still don't see the benefit of using Math.random() in tests, but at least id no longer causes harm.

stephan-herrmann (approved):

stephan-herrmann:
Thanks, @mickaelistria

Issue #344, opened 2022-08-25
[19] Rename record pattern variable not working 


noopur2507:
```
package pack11;

@SuppressWarnings("preview")
public class X {
	public static void print(Object p) {
		if (p instanceof Pair(Student(Object n1,int i),Teacher(Object n)) p1) {
			System.out.println(i); // rename 'i'
		}
	}

	public static void main(String[] args) {
		print(new Pair(new Student("abc", 15), new Teacher("123")));
		print(new Pair(new Teacher("123"), new Student("abc", 1)));
	}
}

sealed interface Person permits Student, Teacher {
	String name();
}

record Student(String name, int id) implements Person {
}

record Teacher(String name) implements Person {
}

record Pair(Person s, Person s1) {
}
```
Rename 'i' in the sysout statement above. It doesn't rename the variable declaration 'int i'.

vik-chand:
Changing the scope of this bug to code select

PR #368, opened 2022-09-05
[19] Rename record pattern variable not working

vik-chand:
Signed-off-by: Vikas Chandra <Vikas.Chandra@in.ibm.com>

## What it does

When there is RecordPattern, also push the local variable on the ASTStack so that the codeSelect can find it.

## How to test

Fixes [344](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/344)
See the comment#0 of the bug. Also see the test case

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchBugs19Tests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchBugs19Tests.java
@@ -700,6 +700,86 @@ public class JavaSearchBugs19Tests extends AbstractJavaSearchTests {
 			javaProject.setOption(JavaCore.COMPILER_PB_ENABLE_PREVIEW_FEATURES, old);
 		}
 	}
+	public void testIssue344_001() throws CoreException {
+		this.workingCopies = new ICompilationUnit[1];
+		this.workingCopies[0] = getWorkingCopy("/JavaSearchBugs/src/X.java",
+				"public class X {\n"
+						+ "  static void print(Rectangle r) {\n"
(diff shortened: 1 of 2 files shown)


jarthana:
We should also keep in mind the following case when providing a fix:

		switch (r) {
		case Rectangle r1 when (r instanceof (Rectangle(ColoredPoint upperLeft2, ColoredPoint lowerRight) r2)):
			System.out.println("Upper-left corner: " + upperLeft2);
			break;
			default :
				break;
		} 

The proposed fix doesn't appear to fix this case.

vik-chand:
The last fix still doesnt fix the switch-case issue. I am working on that.

srikanth-sankaran:
I believe this PR is wrong on a few counts. It only works by accident/co-incidence. If you debug the junit test 
`org.eclipse.jdt.core.tests.model.JavaSearchBugs19Tests.testIssue344_002()` you will find that when `org.eclipse.jdt.internal.codeassist.select.SelectionParser.consumeCaseLabelElement(CaseLabelKind)' is reached
this chunk of code is executed:

```
                        ASTNode[] ps = this.patternStack;
			if (ps[0] instanceof RecordPattern) {
				pushLocalVariableFromRecordPatternOnAstStack((RecordPattern) ps[0]);
			}
```

this.patternPtr == -1

which means the patternStack is empty and we have no business to access it.

Also I believe the approach taken by `pushLocalVariableFromRecordPatternOnAstStack` itself is questionable although it may be building on prior bad art - `org.eclipse.jdt.internal.codeassist.select.SelectionParser.consumeInstanceOfExpressionWithName()` also pushes things onto astStack - this is incorrect approach I believe. 

I will attempt to roll back this fix and reimplement an alternate one and when the PR is available document it here

PR #397, opened 2022-09-15
Add note about typo in method declaration parethesis position formatter option

b-fein:
## What it does

Added a note to a configuration option that contains a typo to highlight it for people manually editing a formatter configuration file and to explain that it has to remain.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/formatter/org/eclipse/jdt/core/formatter/DefaultCodeFormatterConstants.java
+++ b/org.eclipse.jdt.core/formatter/org/eclipse/jdt/core/formatter/DefaultCodeFormatterConstants.java
@@ -1178,6 +1178,10 @@ public class DefaultCodeFormatterConstants {
 	 *     - possible values:   { COMMON_LINES, SEPARATE_LINES_IF_NOT_EMPTY, SEPARATE_LINES_IF_WRAPPED, SEPARATE_LINES, PRESERVE_POSITIONS }
 	 *     - default:           COMMON_LINES
 	 * </pre>
+	 *
+	 * Note that there is a typo ({@code delcaration} vs. {@code declaration})
+	 * in the option name. It has to remain, as fixing it would be a breaking
+	 * change.
 	 * @see #COMMON_LINES
 	 * @see #SEPARATE_LINES_IF_NOT_EMPTY
 	 * @see #SEPARATE_LINES_IF_WRAPPED


org.eclipse.jdt.core/formatter/org/eclipse/jdt/core/formatter/DefaultCodeFormatterConstants.java
```diff
@@ -1185,7 +1185,7 @@ public class DefaultCodeFormatterConstants {
 	 * @see #PRESERVE_POSITIONS
 	 * @since 3.12
```
iloveeclipse, line 1186: `@since 3.12`

I fear it is too late, it is released few years ago, so anyone used this would be broken after this patch.

@mateusz-matela : WDYT?

iloveeclipse (changes requested):

org.eclipse.jdt.core/formatter/org/eclipse/jdt/core/formatter/DefaultCodeFormatterConstants.java
mateusz-matela: I agree, just changing it like that would cause more trouble than it's worth.
If we really care to fix the typo, we can handle the old key as deprecated - that would need changes in `DefaultCodeFormatterOptions.setDeprecatedOptions()` and probably also `ProfileVersionerCore` in jdt-ui.

mateusz-matela: Actually, build errors suggest that either the API constant cannot be changed at all, or with major version bump to 4.0.0 (I just bumped into similar problem in https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/280)
So the constant with a typo would need to stay as `@Deprecated` in API, and a different constant created, and the formatter changed to use the new constant. I guess this would only add to the confusion, so I'd recommend leaving it and adding a comment that we are aware of the typo but it can't be fixed.

b-fein: Fair enough, when creating the PR I should have specified that I wasn’t sure how to handle it, as I’ve seen some projects allow breaking changes for typos, others not.

I reverted the change to the constant value and added a note to the Javadoc instead. Is that okay or did you intend it to be regular code comment?

mateusz-matela (approved):

mateusz-matela:
Looks like @iloveeclipse must approve as well. I see an option to dismiss his change request, but I don't want to be rude ;)

iloveeclipse (approved):

PR #414, opened 2022-09-16
Fix class file generic varargs method reference signature

MaisiKoleni:
## What it does
This fixes the problem that
```java
import java.util.Optional;

public interface ClassFileBug {

	static void signatureBug1() {
		Optional.<Stream<String>>empty().orElseGet(Stream::of);
	}

	static <T> Stream<T> signatureBug2() {
		return Optional.<Stream<T>>empty().orElseGet(Stream::of);
	}

		for (var m : ClassFileBug.class.getDeclaredMethods())
			m.toGenericString();
	}
}
```
fails with 
```
Exception in thread "main" java.lang.reflect.GenericSignatureFormatError: Signature Parse error: Expected Field Type Signature
	Remaining input: +Ljava/util/stream/Stream<Ljava/lang/String;>;
```
due to the `+` in the signature and also other problems related to the generic types that get "disconnected" in the lambda signature. Currently, this is only a test that expects the same behavior as the `javac` compiler does, which is no method signature in such cases.

## How to test
See the provided automatic JUnit test.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericTypeSignatureTest.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericTypeSignatureTest.java
@@ -18,8 +18,6 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
-import java.lang.reflect.GenericSignatureFormatError;
-import java.lang.reflect.Method;
 
 import junit.framework.Test;
 
@@ -1329,34 +1327,6 @@ public class GenericTypeSignatureTest extends AbstractRegressionTest {
 		}
 	}
 
-	/**
-	 * javac does not generate a method signature in that case, and the JDT compiler
-	 * produced
-	 *
-	 * <pre>{@code Signature: ()+Ljava/util/stream/Stream<Ljava/lang/String;>;}</pre>
-	 *
-	 * That is not a valid signature due to the {@code '+'} at the beginning. This
-	 * makes various tools that analyze the class files fail, including the JDK
-	 * reflection API (like, e.g., {@link Method#toGenericString()}) with a
-	 * {@link GenericSignatureFormatError}. Removing it would make the signature
-	 * valid but has the potential to make it nonsensical, like in the following
-	 * example:
-	 *
-	 * <pre>{@code
-	 * // Signature: ()+Ljava/util/stream/Stream<TT;>;
-	 * private static synthetic java.util.stream.Stream lambda$2();
-	 * }</pre>
-	 *
-	 * The type parameters that originate from the location of the lambda / method
-	 * reference are not defined here (note that the lambda is static). The
-	 * signature would need to be
-	 *
-	 * <pre>{@code Signature: <T:Ljava/lang/Object;>()Ljava/util/stream/Stream<TT;>;}</pre>
-	 *
-	 * Generating valid and useful generic method signatures for synthetic methods
-	 * like lambdas does likely cost more than it is worth, so JDT should probably
-	 * follow the javac behavior.
-	 */
 	public void testGenericVarargsMethodReferenceLambdasHaveNoSignature() {
 		// uses lambdas
 		if (this.complianceLevel < ClassFileConstants.JDK1_8)
@@ -1397,11 +1367,11 @@ public class GenericTypeSignatureTest extends AbstractRegressionTest {
 
 				IBinaryMethod m1Lambda = methods[2];
 				assertEquals("Wrong name", "lambda$2", new String(m1Lambda.getSelector()));
-				assertEquals("Wrong signature", "()Ljava/util/stream/Stream<TT;>;", new String(m1Lambda.getGenericSignature()));
+				assertNull("Wrong signature - non denotable should have been suppressed", m1Lambda.getGenericSignature());
 
 				IBinaryMethod m2Lambda = methods[3];
 				assertEquals("Wrong name", "lambda$3", new String(m2Lambda.getSelector()));
-				assertEquals("Wrong signature", "()Ljava/util/stream/Stream<TT;>;", new String(m2Lambda.getGenericSignature()));
+				assertNull("Wrong signature - non denotable should have been suppressed", m2Lambda.getGenericSignature());
 			} catch (ClassFormatException e) {
 				assertTrue(false);
 			} catch (IOException e) {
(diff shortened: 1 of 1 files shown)


MaisiKoleni:
I would be pleased about input from more experienced contributors in that area before I attempt to fix this issue. Is the behavior I expect in the test the correct one?

jarthana:
@Stephan, can you please provide your thoughts/inputs if you can find some time?

stephan-herrmann:
I'm confused between:
> This fixes the problem that ...

and

> Currently, this is only a test ...

Did you intend to submit a fix?

MaisiKoleni:
> I'm confused between:
> > This fixes the problem that ...
> 
> and
> 
> > Currently, this is only a test ...
> 
> Did you intend to submit a fix?
> 

@stephan-herrmann Yes, see my comment why the fix is not there yet. This PR is still in draft state, but it would be good to know if the direction I am going is correct.

What I would change: remove the method signature from lambdas in class files.

stephan-herrmann:
> What I would change: remove the method signature from lambdas in class files.

I see. Questions to be answered:
* in which situations exactly should the signature be omitted?
  * what is the code location to detect this condition?
* is it legal in that situation to omit the signature attribute?

I'll try to find time for a more detailed answer by this weekend.

Until then please have a look at this previous, related issue: https://bugs.eclipse.org/bugs/show_bug.cgi?id=494198

stephan-herrmann:
@MaisiKoleni I'm sorry I haven't yet found the time for a deeper look into the issue. Do the questions from my previous comment already help you to make progress? Perhaps you could just comment back what you _think_ should be done, before implementing?

MaisiKoleni:
@stephan-herrmann Yes, I would take a look when `javac` emits signatures for lambdas and try to replicate that behavior. I will also try to look into the specification on this matter, if there are any rules given.

stephan-herrmann:
With #756 merged, this test case would produce the following signature for both lambdas:
`()Ljava/util/stream/Stream<TT;>;`

@MaisiKoleni  do you still see a need for suppressing signature generation?

MaisiKoleni:
@stephan-herrmann Not necessarily, if the generated signature is valid, it should be ok.

stephan-herrmann:
> @stephan-herrmann Not necessarily, if the generated signature is valid, it should be ok.

Thanks.

I adjusted the test to confirm the current behavior, but no hurry to merge this, given Eclipse is already in the endgame for 4.28.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericTypeSignatureTest.java
```diff
@@ -1326,4 +1329,83 @@ public void testBug460491() {
...
+	 *
+	 * The type parameters that originate from the location of the lambda / method
+	 * reference are not defined here (note that the lambda is static). The
+	 * signature would need to be
+	 *
+	 * <pre>{@code Signature: <T:Ljava/lang/Object;>()Ljava/util/stream/Stream<TT;>;}</pre>
+	 *
+	 * Generating valid and useful generic method signatures for synthetic methods
+	 * like lambdas does likely cost more than it is worth, so JDT should probably
+	 * follow the javac behavior.
```
stephan-herrmann, line 1358: We might still update this comment :)

stephan-herrmann (approved):
Meanwhile ecj generates a legal signature attribute, so no code changes are necessary any more.

Still the test case is valuable to prevent future regressions.

iloveeclipse:
> adjusted the test to confirm the current behavior, but no hurry to merge this, given Eclipse is already in the endgame for 4.28

Test changes still can be merged.

srikanth-sankaran:
A slightly modified form of the test case from top morphs this defect into the same state as https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1048

As wondered at https://github.com/eclipse-jdt/eclipse.jdt.core/issues/756 that fix is not fully enough. When it encodes a signature of the form ()Ljava/util/stream/Stream<TT;>; whether there is a T in scope become crucial to whether reflection API methods can process the signature. I don't think we have heard the end of it. So I would not close this in a hurry merging in the test case.

I will continue to investigate this and try to provide a fuller fix than  https://github.com/eclipse-jdt/eclipse.jdt.core/issues/756  that would also address this test case and https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1048

srikanth-sankaran:
> A slightly modified form of the test case from top morphs this defect into the same state as #1048
> 
> As wondered at #756 that fix is not fully enough. When it encodes a signature of the form ()Ljava/util/stream/Stream<TT;>; whether there is a T in scope become crucial to whether reflection API methods can process the signature. I don't think we have heard the end of it. So I would not close this in a hurry merging in the test case.
> 
> I will continue to investigate this and try to provide a fuller fix than #756 that would also address this test case and #1048

And if a type variable T were to be in scope, that may well be a dangling incorrect reference. I think problem originates from opting to encode the signature from erased form of the method. In the case of wildcards we should reach for the bounds and use that to parameterize the generic type rather than by using the source type variable

srikanth-sankaran:
> > A slightly modified form of the test case from top morphs this defect into the same state as #1048
> > As wondered at [#756](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/756#issuecomment-1531317345) that fix is not fully enough. When it encodes a signature of the form ()Ljava/util/stream/Stream<TT;>; whether there is a T in scope become crucial to whether reflection API methods can process the signature. I don't think we have heard the end of it. So I would not close this in a hurry merging in the test case.
> > I will continue to investigate this and try to provide a fuller fix than #756 that would also address this test case and #1048
> 
> And if a type variable T were to be in scope, that may well be a dangling incorrect reference. I think problem originates from opting to encode the signature from erased form of the method. In the case of wildcards we should reach for the bounds and use that to parameterize the generic type rather than by using the source type variable

The more I think about it, this looks troublesome. Even if we were to avoid encoding alien type variables but encode the bounds somehow, that is no guarantee of absence of trouble downstream - It is hard to know the set of all manipulations the generic signature may be put to and the approximations may back fire.

Another approach may be to enhance the lambda implementation method with type variables that (are declared elsewhere but) get encoded in the signature - this is completely inventive territory and we have veered off too far and wide from JVMS or reference implementation.

I am inclined to think we should just punt and stop emitting signatures if the relevant types are non-denotable rather than do these ad-hoc extra constitutional excursions.

@stephan-herrmann - do you disagree ??

(in the test case I added for https://github.com/eclipse-jdt/eclipse.jdt.core/issues/756 -  org.eclipse.jdt.core.tests.compiler.regression.LambdaExpressionsTest.testIssue756() , the test class X happens to have a type variable T while in the test case in https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1048#issuecomment-1550938091 there is no T in scope - see the difference in behavior of reflection APIs)

stephan-herrmann:
> I am inclined to think we should just punt and stop emitting signatures if the relevant types are non-denotable rather than do these ad-hoc extra constitutional excursions.
> 
> @stephan-herrmann - do you disagree ??

No disagreement.

After re-reading https://bugs.eclipse.org/bugs/show_bug.cgi?id=494198 I believe the key message from that bug for the current ticket is
> Types that have no encoding via JVMS should never occur in *Signature* attributes. Where this seems to be necessary nevertheless, this is a bug in JLS.

This should allow us to just skip signature generation if there's no good encoding.

The strategy of erasure was targeted more for LVTT, where not generating is not an option. Sorry for sending you down that alley wrt Signature :(

See [this comment](https://bugs.eclipse.org/bugs/show_bug.cgi?id=494198#c10)

srikanth-sankaran:
I think this PR can be closed if the test is cleaned up to capture the current behavior and validate that behavior for the snippet relevant here and merged in

Issue #415, opened 2022-09-17
The package xxx is accessible from more than one module


jdneo:
After searching around, I found a detailed explanation of this problem: https://bugs.eclipse.org/bugs/show_bug.cgi?id=536928#c53

But, unfortunately, the [JDK side bug](https://bugs.openjdk.org/browse/JDK-8215739) is just there being inactive for ~4 years. It's embarrassing that although Eclipse did the right thing, the behavior is different from javac/build tools/other IDEs, which confuse users quite a lot. (I'm not blaming Eclipse, Eclipse did things right).

So I'm thinking that, is it possible that adding a preference to specify whether the module system should be enabled or not when the project is not a modular project, but the JDK is higher than `1.8`? In that case, user will not bother excluding dependencies or creating module description file. And more importantly, the behavior will be aligned with the build tools.

After the JDK bug is fixed, we can deprecate this preference.

iloveeclipse:
> And more importantly, the behavior will be aligned with the build tools

You mean with the javac that has the bug in question?

> After the JDK bug is fixed, we can deprecate this preference.

Well, this means we build something in the compiler that will be broken as soon as javac fix will be there, and Eclipse again would need a fix.

Please ask Java maintainers on mailing list again to look at this bug, I **personally** don't think it is wise to make workarounds in Eclipse compiler for javac bugs that violate the Java specs.

@stephan-herrmann : WDYT?

testforstephen:
Since the default behavior of ECJ still follows the strict JPMS rules, it doesn't break existing tools. I think it makes sense from a user perspective to have such a new preference to maintain compatibility with other popular Java build tools (e.g. Maven, Intellij IDEA).

stephan-herrmann:
You might as well link to https://stackoverflow.com/questions/51094274/eclipse-cant-find-xml-related-classes-after-switching-build-path-to-jdk-10/53824670#53824670

* That bug in javac concerns one specific situation, the proposed change in #424 is way too broad.
* Since Java 9, there is no longer an option to avoid JPMS, even if all your code is in an unnamed module, JPMS rules do apply and the system libraries **are** modular.
* What the community needs to do most urgently: put pressure on Oracle to fix that bug. The votes on SO might be a good start :)
  * Makeing ecj bug compatible releases any such pressure and Java will degrade even more
[...]

PR #424, opened 2022-09-20
Add a preference to ignore the unnamed module for split packages

jdneo:
Signed-off-by: Sheng Chen <sheche@microsoft.com>

Hi team! 👋

I was trying to make a draft to illustrate the idea mentioned in #415. It will be great and helpful if anyone from the team can take a look and let me know if this idea is acceptable, thanks!

## What it does

Add a preference to control whether the unnamed module should be ignored when it comes to a split package.

In https://openjdk.org/projects/jigsaw/spec/sotms/#the-unnamed-module says

```
If a package is defined in both a named module and the unnamed module then
the package in the unnamed module is ignored. This preserves reliable configuration
even in the face of the chaos of the class path, ensuring that every module still reads
at most one module defining a given package.
```

## How to test

Sample: [test1.zip](https://github.com/eclipse-jdt/eclipse.jdt.core/files/9605378/test1.zip)

In `.settings/org.eclipse.jdt.core.prefs`, if `org.eclipse.jdt.core.compiler.ignoreUnnamedModuleForSplitPackage` is set to `enabled`. After import the sample project, everything looks fine. If the preference is set to `disabled`, `The package xxx is accessible from more than one module` can be observed.

## Author checklist


org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/LookupEnvironment.java
```diff
@@ -110,6 +110,7 @@ public class LookupEnvironment implements ProblemReasons, TypeConstants {
 	private SimpleLookupTable uniqueGetClassMethodBinding; // https://bugs.eclipse.org/bugs/show_bug.cgi?id=300734
 
 	boolean useModuleSystem;					// true when compliance >= 9 and nameEnvironment is module aware
+	boolean ignoreUnnamedModule;
```
stephan-herrmann, line 113: I'd prefer not to further pollute the state of `LookupEnvironment`, given that the field is just an alias for `this.globalOptions.ignoreUnnamedModuleForSplitPackage`. It's perfectly normal use to ask `environment.globalOptions.xyz` in any client code.

org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/SplitPackageBinding.java
```diff
@@ -258,23 +258,33 @@ public boolean isDeclaredIn(ModuleBinding moduleBinding) {
...
+		PlainPackageBinding bindingInUnnamedModule = null;
 		for (PlainPackageBinding incarnation : this.incarnations) {
 			if (incarnation.hasCompilationUnit(false)) {
 				if (preferLocal && incarnation.enclosingModule == clientModule) {
 					return incarnation;
 				} else {
 					if (clientModule.canAccess(incarnation)) {
-						visibleCount++;
-						unique = incarnation;
+						if (incarnation.enclosingModule.isUnnamed()) { // TODO: check preference
```
stephan-herrmann, line 270: stale TODO (?)

org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/SplitPackageBinding.java
```diff
@@ -258,23 +258,33 @@ public boolean isDeclaredIn(ModuleBinding moduleBinding) {
...
+						} else {
+							visibleCountInNamedModules++;
+							uniqueInNamedModules = incarnation;
+						}
 					}
 				}
 			}
 		}
-		if (visibleCount > 1)
+		if (visibleCountInNamedModules > 1)
```
stephan-herrmann, line 280: As the control structure is growing, please add braces for readability, at least to the outer if-else.

stephan-herrmann (commented):
please find some nitpicks inline.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaCorePreferenceInitializer.java
```diff
@@ -63,6 +63,7 @@ public void initializeDefaultPreferences() {
 		defaultOptionsMap.put(JavaCore.COMPILER_TASK_CASE_SENSITIVE, JavaCore.ENABLED);
 		defaultOptionsMap.put(JavaCore.COMPILER_DOC_COMMENT_SUPPORT, JavaCore.ENABLED);
 		defaultOptionsMap.put(JavaCore.COMPILER_PB_FORBIDDEN_REFERENCE, JavaCore.ERROR);
+		defaultOptionsMap.put(JavaCore.COMPILER_IGNORE_UNNAMED_MODULE_FOR_SPLIT_PACKAGE, JavaCore.DISABLED);
```
stephan-herrmann, line 66: On second look, I don't understand why this is added (under the heading "Override some compiler defaults"). Default is already disabled, or am I missing smth?

org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
```diff
@@ -2645,6 +2645,23 @@ public final class JavaCore extends Plugin {
 	 * @since 3.18
 	 */
 	public static final String COMPILER_PB_REPORT_PREVIEW_FEATURES = PLUGIN_ID + ".compiler.problem.reportPreviewFeatures"; //$NON-NLS-1$
+
+	/**
+	 * Compiler option ID: Ignore unnamed module for split package.
+	 * <p>
+	 * When enabled, the compiler will ignore the unnamed module when a package is defined in both a named module and the unnamed module.
+	 * This preference is introduced due to the explanation described in https://openjdk.org/projects/jigsaw/spec/sotms/#the-unnamed-module.
```
stephan-herrmann, line 2653: I think without further context, this will leave readers confused. I'm afraid we'll have to explain a bit more, like:
```
	 * <p>
	 * With this option the compiler will deliberately accept programs violating JLS in a specific way.
	 * Instead the compiler will behave in accordance to the original, but unmaintained document
	 *  <a href="https://openjdk.org/projects/jigsaw/spec/sotms/#the-unnamed-module">"The State of the Module System"</a>,
	 * which indicates that different semantics had been intended.
	 * </p>
```

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/UnnamedModuleTest.java
```diff
@@ -66,4 +71,35 @@ public void testBug522326() {
...
+	String[] libs = new String[len+1];
+	System.arraycopy(defaultLibs, 0, libs, 0, len);
+	libs[len] = path;
+	Map<String, String> compilerOptions = getCompilerOptions();
+	compilerOptions.put(CompilerOptions.OPTION_IgnoreUnnamedModuleForSplitPackage, CompilerOptions.ENABLED);
+	this.runConformTest(
+			true,
+			new String[] {
+				"X.java",
+				"import org.xml.sax.SAXParseException;\n" +
```
stephan-herrmann, line 91: This successfully tests the code path originating from `CompilationUnitScope.faultInImports()`.

Additionally, I'd like to see one test covering the case where this is triggered from `QualifiedTypeReference.getTypeBinding(Scope)`.

Currently, we know that `ProblemReporter.conflictingPackagesFromModules()` can only be triggered by a specific outcome from `SplitPackageBinding.getVisibleFor()`, good. But from a testing p.o.v. it might be prudent to take a more black-boxish stand, i.e., test all scenarios that can pull in a package into complation - including qualified type references.

Bonus questions: what happens when code uses a `QualifiedNameReference` pointing directly to a static member of an affected type? Do we actually detect the error? If so, is the new option effective in that case, too?

stephan-herrmann (changes requested):
We're getting closer, great!

With more diligence applied, here's the next round of comments

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaCorePreferenceInitializer.java
jdneo: Ah, it's not necessary.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/UnnamedModuleTest.java
jdneo: Thanks for the suggestion, more cases are added in the latest commit.

In terms of the bonus question, before this change:

If the code is written as:

```java
import static a.split.package.type.staticMember;
```
JDT will report the split package error.

But if the fully qualified name is directly used in the code:

```java
String s = a.split.package.type.staticMember;
```

JDT will not report any errors.

But anyway, I put both of the 2 cases into the tests to avoid regression in the future.

jdneo:
Hi @stephan-herrmann, I've updated the PR, would you have time to take a look? :)

stephan-herrmann (approved):
All looks well now. Thanks!

laeubi:
@stephan-herrmann @jdneo I just came across the preference

`org.eclipse.jdt.core.compiler.ignoreUnnamedModuleForSplitPackage=enabled`

but it is quite hard to discover and I was not aware about this only after some google-search with a lot of unhelpful results.

Do you think it would be useful to enhance the error message with a hint to the property? e.g

> The package xxx is accessible from more than one module, if you want to allow this use org.eclipse.jdt.core.compiler.ignoreUnnamedModuleForSplitPackage=enabled in your org.eclipse.jdt.core preferences.

xnhp:
Can this option also be set when using `eclipsec` but not the Eclipse IDE? For instance, IntelliJ comes with `eclipsec`. I've tried putting some variations in the compiler CLI arguments in IntelliJ but none were recognised.

stephan-herrmann:
> Can this option also be set when using `eclipsec` but not the Eclipse IDE? For instance, IntelliJ comes with `eclipsec`. I've tried putting some variations in the compiler CLI arguments in IntelliJ but none were recognised.

I've never heard the name "eclipsec" :smile: but ecj should understand the option via `-properties <file>` (haven't tried it myself, though).

xnhp:
Thank you :pray: It does not complain about some unknown option, but it does not seem to help with the split package error either. I think I'm having the same issue as [here](https://github.com/eclipse-jdtls/eclipse.jdt.ls/issues/3074).

laeubi:
@xnhp You should ask at the supportforum of your tool (**IntelliJ**) how to set compiler settings for the Eclipse Compiler there.

Issue #432, opened 2022-09-22
Search for reference not working for method reference in anonymous class


trancexpress:
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=566435.

To reproduce, compute the call hierarchy of method `missingReference1` in the following snippet:

```
public class MissingReference2 {

	public static void main(String[] args) {
		new Runnable() {
			@Override
			public void run() {
				// here is the usage of the method
				Stream.empty().filter(MissingReference2::missingReference1).count();
			}
		}.run();
	}

	// Show references of this method says "0 references"
	public static boolean missingReference1(Object arg) {
		return true;
	}
}
```

Observe that there is nothing found.

This is a regression from the fix for https://bugs.eclipse.org/bugs/show_bug.cgi?id=429279.

In particular, the entire body of `main` is removed by this call:

```
diff --git a/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/SourceIndexer.java b/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/SourceIndexer.java
index 39d388adb1..ac4820fc87 100644
--- a/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/SourceIndexer.java
+++ b/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/SourceIndexer.java
@@ -168,7 +168,7 @@ public class SourceIndexer extends AbstractIndexer implements ITypeRequestor, Su
                        // Use a non model name environment to avoid locks, monitors and such.
                        INameEnvironment nameEnvironment = IndexBasedJavaSearchEnvironment.create(Collections.singletonList((IJavaProject)javaProject), JavaModelManager.getJavaModelManager().getWorkingCopies(DefaultWorkingCopyOwner.PRIMARY, true/*add primary WCs*/));
                        this.lookupEnvironment = new LookupEnvironment(this, this.options, problemReporter, nameEnvironment);
-                       reduceParseTree(this.cud);
+                       //reduceParseTree(this.cud);
                        this.lookupEnvironment.buildTypeBindings(this.cud, null);
                        this.lookupEnvironment.completeTypeBindings();
                        this.cud.scope.faultInTypes();
```

When changing the snippet slightly, the problem is not seen:

```
public class MissingReference2 {

	public static void main(String[] args) {
		new Runnable() {
			@Override
			public void run() {
[...]

PR #433, opened 2022-09-22
No match found for method reference in anonymous class #432

trancexpress:
The performance improvement done for Eclipse bug 429279 introduces
pruning of method bodies, if a method does not have functional interface
types. Checking for the presence of functional interface types however
does not consider anonymous types defined in the method body.

This change prevents the pruning done for bug 429279, in case a method
defines an anonymous type.

A test is also added for a few example cases.

See:
* https://bugs.eclipse.org/bugs/show_bug.cgi?id=429279
* https://bugs.eclipse.org/bugs/show_bug.cgi?id=566435

Fixes: #432

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchTests.java
@@ -4563,34 +4563,157 @@ public void testAnonymousTypeMethodReferenceJarSearchGh375() throws Exception {
 		root.attachSource(srcZip.getFullPath(), null, null);
 		waitUntilIndexesReady();
 
-		JavaSearchResultCollector testResultCollector = new JavaSearchResultCollector();
-		String typeFqn = "TestGh375";
-		IType testType = project.findType(typeFqn);
-		assertNotNull("Failed to find test type: " + typeFqn, testType);
-		String methodName = "helloWorld";
-		IMethod testMethod = null;
-		for (IMethod method : testType.getMethods()) {
-			if (methodName.equals(method.getElementName())) {
-				testMethod = method;
-			}
-		}
-		IJavaSearchScope scope = SearchEngine.createJavaSearchScope(new IJavaElement[] { testMethod.getAncestor(IJavaElement.PACKAGE_FRAGMENT) });
-		assertNotNull("Failed to find method: " + methodName + ", in type: " + typeFqn, testMethod);
-		int agnosticMatchRule = SearchPattern.R_EXACT_MATCH | SearchPattern.R_CASE_SENSITIVE | SearchPattern.R_ERASURE_MATCH;
-		SearchPattern pattern = SearchPattern.createPattern(testMethod, IJavaSearchConstants.REFERENCES, agnosticMatchRule);
-		SearchEngine searchEngine = new SearchEngine();
-		searchEngine.search(pattern, new SearchParticipant[] {SearchEngine.getDefaultSearchParticipant()}, scope, testResultCollector, new NullProgressMonitor());
-
-		String foundReferences = testResultCollector.toString();
-		assertFalse("Expected search to find references of method: " + testMethod, foundReferences.isEmpty());
+		IMethod testMethod = findMethod(project, "TestGh375", "helloWorld");
+		String foundReferences = searchForMethodReferences(testMethod);
+		assertFalse("Expected search to find references of method: " + testMethod + ", in snippet:\n" + libSource, foundReferences.isEmpty());
 		List<String> results = Arrays.asList(foundReferences.split(System.lineSeparator()));
 		String[] expectedResults = {
-				"libGh375.jar java.lang.Void <anonymous>.call()",
+				"call() (not open) [in <anonymous> [in TestGh375$2.class [in <default> [in libGh375.jar [in P]]]]]",
 		};
 		assertEquals("Unexpected search result", String.join(System.lineSeparator(), expectedResults), String.join(System.lineSeparator(), results));
 	} finally {
 		JavaCore.setOptions(getDefaultJavaCoreOptions());
 		deleteProject("P");
 	}
+}
+/*
+ * Test that we can find method references called in an anonymous type defined in a method.
+ * https://github.com/eclipse-jdt/eclipse.jdt.core/issues/432
+ */
+public void testAnonymousTypeMethodReferenceSearchGh432() throws Exception {
+	String testProjectName = "gh432MethodReferencesSearchBug";
+	String snippet1 = "package p;\n" +
+			"public class TestGh432 {\n" +
+			"  public static void main(String[] args) {\n" +
+			"    class R implements Runnable {\n" +
+			"      public void run() {\n" +
+			"        java.util.stream.Stream.empty().filter(TestGh432::missingReference).count();\n" +
+			"      }\n" +
+			"    }\n" +
+			"    new R().run();\n" +
+			"  }\n" +
+			"  public static boolean missingReference(Object arg) {\n" +
+			"    return true;\n" +
+			"  }\n" +
+			"}";
+	String snippet2 = "package p;\n" +
+			"public class TestGh432 {\n" +
+			"  private static void foo(Runnable r) {\n" +
+			"    r.run();\n" +
+			"  }\n" +
+			"  public static void main(String[] args) {\n" +
+			"    foo(new Runnable() {\n" +
+			"      public void run() {\n" +
+			"        java.util.stream.Stream.empty().filter(TestGh432::missingReference).count();\n" +
+			"      }\n" +
+			"    });\n" +
+			"  }\n" +
+			"  public static boolean missingReference(Object arg) {\n" +
+			"    return true;\n" +
+			"  }\n" +
+			"}";
+	String snippet3 = "package p;\n" +
+			"public class TestGh432 {\n" +
+			"  public static void main(String[] args) {\n" +
+			"    new Runnable() {\n" +
+			"      public void run() {\n" +
+			"        new Runnable() {\n" +
+			"          public void run() {\n" +
+			"            java.util.stream.Stream.empty().filter(TestGh432::missingReference).count();\n" +
+			"          }\n" +
+			"        }.run();\n" +
+			"      }\n" +
+			"    }.run();\n" +
+			"  }\n" +
+			"  public static boolean missingReference(Object arg) {\n" +
+			"    return true;\n" +
+			"  }\n" +
+			"}";
+	String snippet4 = "package p;\n" +
+			"public class TestGh432 {\n" +
+			"  private static void foo(Runnable r) {\n" +
+			"    r.run();\n" +
+			"  }\n" +
+			"  public static void main(String[] args) {\n" +
+			"    foo(() -> {\n" +
+			"      new Runnable() {\n" +
+			"        public void run() {\n" +
+			"          java.util.stream.Stream.empty().filter(TestGh432::missingReference).count();\n" +
+			"        }\n" +
+			"      };\n" +
+			"    });\n" +
+			"  }\n" +
+			"  public static boolean missingReference(Object arg) {\n" +
+			"    return true;\n" +
+			"  }\n" +
+			"}";
+	String[] testSnippets = {
+			snippet1,
+			snippet2,
+			snippet3,
+			snippet4,
+	};
+	String[] expectedMatches = {
+			"void run() {key=Lp/TestGh432$89$R;.run()V} [in R [in main(String[]) [in TestGh432 [in TestGh432.java [in p [in src [in gh432MethodReferencesSearchBug]]]]]]]",
+			"void run() {key=Lp/TestGh432$148;.run()V} [in <anonymous #1> [in main(String[]) [in TestGh432 [in TestGh432.java [in p [in src [in gh432MethodReferencesSearchBug]]]]]]]",
+			"void run() {key=Lp/TestGh432$138;.run()V} [in <anonymous #1> [in run() [in <anonymous #1> [in main(String[]) [in TestGh432 [in TestGh432.java [in p [in src [in gh432MethodReferencesSearchBug]]]]]]]]]",
+			"run() (not open) {key=Lp/TestGh432$162;.run()V} [in <anonymous #1> [in run() [in <lambda #1> [in main(String[]) [in TestGh432 [in TestGh432.java [in p [in src [in gh432MethodReferencesSearchBug]]]]]]]]]",
+
+	};
+	try {
+		IJavaProject p = setupModuleProject(testProjectName, new String[] {"src"}, new String[0], null);
+		String packageFolder = "/" + testProjectName + "/src/p";
+		createFolder(packageFolder);
+		String sourceFile = packageFolder + "/TestGh432.java";
+		waitForAutoBuild();
+		for (int i = 0; i < testSnippets.length; ++i) {
+			String fileContent = testSnippets[i];
+			String expectedResults = expectedMatches[i];
+			deleteFile(sourceFile);
+			waitForAutoBuild();
+			waitUntilIndexesReady();
+			createFile(sourceFile, fileContent);
+			waitForAutoBuild();
+			waitUntilIndexesReady();
+
+			IMethod testMethod = findMethod(p, "p.TestGh432", "missingReference");
+			String foundReferences = searchForMethodReferences(testMethod);
+			assertFalse("Expected search to find references of method: " + testMethod + ", in snippet:\n" + fileContent, foundReferences.isEmpty());
+			List<String> results = Arrays.asList(foundReferences.split(System.lineSeparator()));
+			assertEquals("Unexpected search result for snippet:\n" + fileContent, expectedResults, String.join(System.lineSeparator(), results));
+		}
+	} finally {
+		JavaCore.setOptions(getDefaultJavaCoreOptions());
+		deleteProject(testProjectName);
+	}
+}
+private static String searchForMethodReferences(IMethod testMethod) throws CoreException {
+	class JavaSearchResultCollector extends SearchRequestor {
+		private final StringBuilder result = new StringBuilder();
+		public void acceptSearchMatch(SearchMatch match) throws CoreException {
+			JavaSearchResultCollector.this.result.append(match.getElement());
+			JavaSearchResultCollector.this.result.append(System.lineSeparator());
+		}
+	}
+	JavaSearchResultCollector testResultCollector = new JavaSearchResultCollector();
+	IJavaSearchScope scope = SearchEngine.createJavaSearchScope(new IJavaElement[] { testMethod.getAncestor(IJavaElement.PACKAGE_FRAGMENT) });
+	int agnosticMatchRule = SearchPattern.R_EXACT_MATCH | SearchPattern.R_CASE_SENSITIVE | SearchPattern.R_ERASURE_MATCH;
+	SearchPattern pattern = SearchPattern.createPattern(testMethod, IJavaSearchConstants.REFERENCES, agnosticMatchRule);
+	SearchEngine searchEngine = new SearchEngine();
+	searchEngine.search(pattern, new SearchParticipant[] {SearchEngine.getDefaultSearchParticipant()}, scope, testResultCollector, new NullProgressMonitor());
+	String foundReferences = testResultCollector.result.toString();
+	return foundReferences;
+}
+private static IMethod findMethod(IJavaProject project, String typeFqn, String methodName) throws JavaModelException {
+	IType testType = project.findType(typeFqn);
+	assertNotNull("Failed to find test type: " + typeFqn, testType);
+	IMethod testMethod = null;
+	for (IMethod method : testType.getMethods()) {
+		if (methodName.equals(method.getElementName())) {
+			testMethod = method;
+		}
+	}
+	assertNotNull("Failed to find method: " + methodName + ", in type: " + typeFqn, testMethod);
+	return testMethod;
 }
 }
--- a/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/SourceIndexer.java
+++ b/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/SourceIndexer.java
@@ -196,7 +196,14 @@ public class SourceIndexer extends AbstractIndexer implements ITypeRequestor, Su
 		AbstractMethodDeclaration[] methods = type.methods;
 		for (int j = 0, length = methods == null ? 0 : methods.length; j < length; j++) {
 			AbstractMethodDeclaration method = methods[j];
-			if (method != null && (method.bits & ASTNode.HasFunctionalInterfaceTypes) == 0) {
+			/*
+			 * In case the method defines a local type, skip purging the method body.
+			 * We don't know if the local type defines a method that uses a method reference or a lambda.
+			 * See:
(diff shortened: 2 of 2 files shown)


trancexpress:
@iloveeclipse can you take a look here?

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/SourceIndexer.java
```diff
@@ -208,6 +215,35 @@ private void purgeMethodStatements(TypeDeclaration type) {
 				purgeMethodStatements(memberTypes[i]);
 	}
 
+	private boolean definesAnonymousType(AbstractMethodDeclaration method) {
```
trancexpress, line 218: The code here is IMO bad, but I don't know how else to write the code. The types in the CU are not yet resolved at this point, so (I think) we can't depend on more reliable APIs to find anonymous/inner types.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/SourceIndexer.java
```diff
@@ -196,7 +203,7 @@ private void purgeMethodStatements(TypeDeclaration type) {
 		AbstractMethodDeclaration[] methods = type.methods;
 		for (int j = 0, length = methods == null ? 0 : methods.length; j < length; j++) {
 			AbstractMethodDeclaration method = methods[j];
-			if (method != null && (method.bits & ASTNode.HasFunctionalInterfaceTypes) == 0) {
+			if (method != null && (method.bits & ASTNode.HasFunctionalInterfaceTypes) == 0 && !definesAnonymousType(method)) {
```
szarnekow, line 206: I don't have the sources checked out locally and github code navigation doesn't help me with it: Can we use the same tag-bit HasFunctionalInterfaceTypes (maybe rename to HasFunctionalInterfaceTypesOrLocalDeclarations)  and make sure it is correctly assigned when a local class declaration is encountered?

Does the bug manifest itself with a local named class declaration, too?
```
		class R implements Runnable {
			@Override
			public void run() {
				// here is the usage of the method
				Stream.empty().filter(MissingReference2::missingReference1).count();
			}
		};
	}
```
Does it make sense to have test cases that are deeper nested, too?

trancexpress: > I don't have the sources checked out locally and github code navigation doesn't help me with it: Can we use the same tag-bit HasFunctionalInterfaceTypes (maybe rename to HasFunctionalInterfaceTypesOrLocalDeclarations) and make sure it is correctly assigned when a local class declaration is encountered?

I can check. If you can point me to the code I should be checking, that would speed things up.

> Does the bug manifest itself with a local named class declaration, too?
> 
> ```
> 	public static void main(String[] args) {
> 		class R implements Runnable {
> 			@Override
> 			public void run() {
> 				// here is the usage of the method
> 				Stream.empty().filter(MissingReference2::missingReference1).count();
> 			}
> 		};
> 	}
> ```

Yes, there is a snippet for this in the new test. The name of the method needs work, yes.

> Does it make sense to have test cases that are deeper nested, too?

There is one snippet for this in the new test.

szarnekow: > There is one snippet for this in the new test.

Sorry, I missed that when reading through this.

> If you can point me to the code I should be checking, that would speed things up.

I'd love to but I cannot seem to find the piece where this tagbit is set on the ASTNode via the browser. I don't have a local JDT clone at hand right now.

trancexpress: > I'd love to but I cannot seem to find the piece where this tagbit is set on the ASTNode via the browser. I don't have a local JDT clone at hand right now.

OK, I'll check where the flag is set tomorrow.

trancexpress: Given this snippet:

```
		class R implements Runnable {
			public void run() {
				// here is the usage of the method
				java.util.stream.Stream.empty().filter(TestGh432::missingReference).count();
			}
		}
		new R().run();
	}
```

The stack trace at which the flag `ASTNode.HasFunctionalInterfaceTypes` is set is:

```
"Java indexing" #60 daemon prio=5 os_prio=0 cpu=315.32ms elapsed=806.03s tid=0x00007ffff0c57000 nid=0xf269 at breakpoint [0x00007fff5bcfb000]
   java.lang.Thread.State: RUNNABLE
        at org.eclipse.jdt.internal.compiler.parser.Parser.markEnclosingMemberWithLocalOrFunctionalType(Parser.java:12767)
        at org.eclipse.jdt.internal.compiler.parser.Parser.consumeReferenceExpression(Parser.java:9362)
        at org.eclipse.jdt.internal.compiler.parser.Parser.consumeReferenceExpressionTypeForm(Parser.java:9300)
        at org.eclipse.jdt.internal.compiler.parser.Parser.consumeRule(Parser.java:7915)
        at org.eclipse.jdt.internal.compiler.parser.Parser.parse(Parser.java:13149)
        at org.eclipse.jdt.internal.compiler.parser.Parser.parse(Parser.java:13505)
        at org.eclipse.jdt.internal.compiler.ast.MethodDeclaration.parseStatements(MethodDeclaration.java:251)
        at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.parseMethods(TypeDeclaration.java:1143)
        at org.eclipse.jdt.internal.compiler.parser.Parser.getMethodBodies(Parser.java:12167)
        at org.eclipse.jdt.internal.compiler.SourceElementParser.parseCompilationUnit(SourceElementParser.java:1138)
        at org.eclipse.jdt.internal.core.search.indexing.SourceIndexer.indexDocument(SourceIndexer.java:122)
        at org.eclipse.jdt.internal.core.search.JavaSearchParticipant.indexDocument(JavaSearchParticipant.java:83)
        at org.eclipse.jdt.internal.core.search.indexing.IndexManager.indexDocument(IndexManager.java:664)
        at org.eclipse.jdt.internal.core.search.indexing.IndexManager$2.execute(IndexManager.java:1298)
        at org.eclipse.jdt.internal.core.search.processing.JobManager.run(JobManager.java:478)
        at java.lang.Thread.run(java.base@11.0.10/Thread.java:834)
```

So far I don't see how to traverse the AST upward. The stack `Parser.astStack` contains only the inner type and its method, not the outer type method that defines it:

```
class R implements Runnable {
}
```
```
public void run() {
}
```

The method that is being parsed (see the stack trace, `MethodDeclaration.java:251`) is `main`, i.e. we are at the correct spot. Maybe we can do something after that line...

trancexpress:
@szarnekow how does the change look like? To me it seems good now.

@iloveeclipse I think this change is ready for a review, but lets wait for what Sebastian has to say.

org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/parser/Parser.java
```diff
@@ -12772,6 +12772,20 @@ protected void markEnclosingMemberWithLocalOrFunctionalType(LocalTypeKind contex
 				case LOCAL:
 					node.bits |= ASTNode.HasLocalType;
 			}
+			// mark enclosing methods as having functional interface types, if the current node has such
+			if ((node.bits & ASTNode.HasFunctionalInterfaceTypes) != 0) {
+				List<Object> ancestors = new ArrayList<>(i);
+				ancestors.add(this.referenceContext);
+				for (int j = i - 1; j >= 0; j--) {
```
trancexpress, line 12779: I guess for very high nesting this might cause trouble, as would be quadratic complexity. If e.g. there is 100% nesting level and each time a lambda or a method reference is in the code.

Maybe we can stop the iteration over ancestor nodes if we see a method that is already marked with the flag... It should help reduce operations regardless of how the AST is traversed (BFS or DFS). Though as long as the flag is the same (`HasFunctionalInterfaceTypes`) I'm not sure we are allowed to do that. Since e.g. a parent method might itself already have a functional type, but its parent method still needs to be marked due to containing the node that is being visited "here".

szarnekow: Seeing the AstNode.HasLocalType here (line 12773), I got an idea:

Reverting the adjustment to Parser.java and adjusting `org.eclipse.jdt.internal.core.search.indexing.SourceIndexer.purgeMethodStatements(TypeDeclaration)` like this makes the newly added tests green:

```java
	private void purgeMethodStatements(TypeDeclaration type) {
		AbstractMethodDeclaration[] methods = type.methods;
		for (int j = 0, length = methods == null ? 0 : methods.length; j < length; j++) {
			AbstractMethodDeclaration method = methods[j];
			if (method != null && (method.bits & (ASTNode.HasFunctionalInterfaceTypes | ASTNode.HasLocalType)) == 0) {
				method.statements = null;
				method.javadoc = null;
			}
		}

		TypeDeclaration[] memberTypes = type.memberTypes;
		if (memberTypes != null)
			for (int i = 0, l = memberTypes.length; i < l; i++)
				purgeMethodStatements(memberTypes[i]);
	}
```

Not sure if / how I can push this to your branch / merge request, though.

trancexpress: > Not sure if / how I can push this to your branch / merge request, though.

Me neither, on idea how that works with forked repos. I believe the PR can be edited by committers/reviewers, but I don't know how that works either.

I've adjusted the change, seems like a good suggestion.

trancexpress:
@szarnekow if you have some extra time, could you also take a look at #438 ? I understand the problem meanwhile, but have no idea how a fix should look.

Essentially, `SourceIndexer.indexResolvedDocument()` will go over functional expressions and try to resolve them. If they cannot be resolved, the indexer doesn't "remember" those expressions.

E.g. a method reference is ignored, if the method cannot be resolved. Adding a type later on (e.g. some type could not compile so far, now it does, or the type was really added "just now") doesn't result in an extra run of `SourceIndexer` for the source that contains the method reference, and so the indexer is in an incomplete state - despite the build state being fine.

I'm not sure which direction the fix should go into. Try to re-index the file with the method reference? Or add "placeholders" (however we want to achieve that) in the index, and try to resolve referenced methods *again* when some search is asking for them? Assuming either option can even be achieved.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchTests.java
```diff
@@ -4593,4 +4576,108 @@ public void testAnonymousTypeMethodReferenceJarSearchGh375() throws Exception {
...
+			"public class TestGh432 {\n" +
+			"  public static void main(String[] args) {\n" +
+			"    class R implements Runnable {\n" +
+			"      public void run() {\n" +
+			"        java.util.stream.Stream.empty().filter(TestGh432::missingReference).count();\n" +
+			"      }\n" +
+			"    }\n" +
+			"    new R().run();\n" +
+			"  }\n" +
+			"  public static boolean missingReference() {\n" +
```
szarnekow, line 4594: Is it intentional that the method `missingReference` does not match the signature of the Predicate that needs to be passed to Stream.filter?

trancexpress: > Is it intentional that the method `missingReference` does not match the signature of the Predicate that needs to be passed to Stream.filter?

No, this is an oversight. I'll adjust it.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchTests.java
```diff
@@ -4593,4 +4576,108 @@ public void testAnonymousTypeMethodReferenceJarSearchGh375() throws Exception {
...
+			"      public void run() {\n" +
+			"        java.util.stream.Stream.empty().filter(TestGh432::missingReference).count();\n" +
+			"      }\n" +
+			"    });\n" +
+			"  }\n" +
+			"  public static boolean missingReference(Object arg) {\n" +
+			"    return true;\n" +
+			"  }\n" +
+			"}";
+	String snippet3 = "package p;\n" +
```
szarnekow, line 4614: Any idea why the test would fail for this case?
I see that the method body of main is retained but still the class defined in the lambda is not processed such the method reference is not found.

```java
class TestGh432 {
	private static void foo(Runnable r) {
		r.run();
	}

		foo(() -> {
			new Runnable() {
				public void run() {
					java.util.stream.Stream.generate(TestGh432::missingReference).count();
				}
			};
		});
	}

	public static boolean missingReference() {
		return true;
	}
}
```

szarnekow (commented):
Nice work.
I found another case (see comment) but it's probably a different bug.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchTests.java
trancexpress: > Any idea why the test would fail for this case? I see that the method body of main is retained but still the class defined in the lambda is not processed such the method reference is not found.

Something with the lambda probably, method references and lambdas have a lot of bugs around them from my experience.

I'll be on vacation next week, I'll check the case above once I'm back.

trancexpress: Looks like #438 to an extent. During the following stack trace, `FunctionalExpression.binding` is `null` and so the indexer "skips" the lambda:

```
"Java indexing" #61 daemon prio=5 os_prio=0 cpu=923.10ms elapsed=207.80s tid=0x00007ffff0c2e000 nid=0x4b4e runnable  [0x00007fff59ca0000]
   java.lang.Thread.State: RUNNABLE
        at org.eclipse.jdt.internal.core.search.indexing.SourceIndexer.indexResolvedDocument(SourceIndexer.java:226)
        at org.eclipse.jdt.internal.core.search.JavaSearchParticipant.indexResolvedDocument(JavaSearchParticipant.java:103)
        at org.eclipse.jdt.internal.core.search.indexing.IndexManager.indexResolvedDocument(IndexManager.java:677)
        at org.eclipse.jdt.internal.core.search.indexing.IndexManager$2.execute(IndexManager.java:1303)
        at org.eclipse.jdt.internal.core.search.processing.JobManager.run(JobManager.java:478)
        at java.lang.Thread.run(java.base@11.0.10/Thread.java:834)
```

trancexpress: OK looks like the project doesn't have the JRE on its classpath, so `Runnable` is not resolved and then the rest of the problems I mentioned follow. I'll try to get the JRE on the classpath... unfortunately that is not so trivial. E.g. here I had to commit a `.classpath` file to do JRE type searches:

https://github.com/eclipse-jdt/eclipse.jdt.core/pull/333/files#diff-c6d731dbfd19fc5fe3c6c0c41bc57229e6d579e3f37f2af0507b3cb2d7b1a634

szarnekow: > Looks like https://github.com/eclipse-jdt/eclipse.jdt.core/issues/438 to an extent

Curious since it's only a single file while the bug seems to rely on multiple files to be present.

Chewing on the question whether the source indexer is the correct tool to resolve the bindings ... I wonder how it can work together with `var` typed locals that are used in method references, e.g. 
```
  var someList = new ArrayList<String>();
  withAcceptor(someList.subList(0,1)::add);
```
Is the source indexer in the position to compute the correct type information for "all" shapes of local vars such that method references can work?

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchTests.java
```diff
@@ -4593,4 +4576,137 @@ public void testAnonymousTypeMethodReferenceJarSearchGh375() throws Exception {
...
+			snippet1,
+			snippet2,
+			snippet3,
+			snippet4,
+	};
+	String[] expectedMatches = {
+			"src/p/TestGh432.java void void p.TestGh432.main(String[]):R#1.run() [missingReference]",
+			"src/p/TestGh432.java void void p.TestGh432.main(String[]):<anonymous>#1.run() [missingReference]",
+			"src/p/TestGh432.java void void void p.TestGh432.main(String[]):<anonymous>#1.run():<anonymous>#1.run() [missingReference]",
+			"src/p/TestGh432.java \nJava Model Exception: Error in Java Model (code 969): run() {key=Lp/TestGh432$162;.run()V} [in <anonymous #1> [in run() [in <lambda #1> [in main(String[]) [in TestGh432 [in TestGh432.java [in p [in src [in gh432MethodReferencesSearchBug]]]]]]]]] does not exist",
```
trancexpress, line 4660: I don't understand the `JavaModelException` occurring here. I'm debugging to find out what is causing it. Stack trace of its creation is:

```
"main" #1 prio=5 os_prio=0 cpu=2537.33ms elapsed=41.73s tid=0x00007ffff0018000 nid=0x1183f at breakpoint [0x00007ffff7fc8000]
   java.lang.Thread.State: RUNNABLE
        at org.eclipse.jdt.core.JavaModelException.<init>(JavaModelException.java:75)
        at org.eclipse.jdt.internal.core.JavaElement.newNotPresentException(JavaElement.java:574)
        at org.eclipse.jdt.internal.core.JavaElement.openWhenClosed(JavaElement.java:617)
        at org.eclipse.jdt.internal.core.JavaElement.getElementInfo(JavaElement.java:328)
        at org.eclipse.jdt.internal.core.JavaElement.getElementInfo(JavaElement.java:314)
        at org.eclipse.jdt.internal.core.hierarchy.HierarchyResolver.remember(HierarchyResolver.java:441)
        at org.eclipse.jdt.internal.core.hierarchy.HierarchyResolver.rememberWithMemberTypes(HierarchyResolver.java:542)
        at org.eclipse.jdt.internal.core.hierarchy.HierarchyResolver.rememberAllTypes(HierarchyResolver.java:527)
        at org.eclipse.jdt.internal.core.hierarchy.HierarchyResolver.resolve(HierarchyResolver.java:874)
        at org.eclipse.jdt.internal.core.hierarchy.HierarchyResolver.resolve(HierarchyResolver.java:669)
        at org.eclipse.jdt.internal.core.hierarchy.HierarchyBuilder.buildSupertypes(HierarchyBuilder.java:129)
        at org.eclipse.jdt.internal.core.hierarchy.IndexBasedHierarchyBuilder.build(IndexBasedHierarchyBuilder.java:170)
        at org.eclipse.jdt.internal.core.hierarchy.TypeHierarchy.compute(TypeHierarchy.java:323)
        at org.eclipse.jdt.internal.core.hierarchy.TypeHierarchy.refresh(TypeHierarchy.java:1319)
        - locked <0x0000000111beb4a8> (a org.eclipse.jdt.internal.core.hierarchy.TypeHierarchy)
        at org.eclipse.jdt.internal.core.CreateTypeHierarchyOperation.executeOperation(CreateTypeHierarchyOperation.java:94)
        at org.eclipse.jdt.internal.core.JavaModelOperation.run(JavaModelOperation.java:740)
        at org.eclipse.jdt.internal.core.JavaModelOperation.runOperation(JavaModelOperation.java:806)
        at org.eclipse.jdt.internal.core.SourceType.newSupertypeHierarchy(SourceType.java:858)
        at org.eclipse.jdt.internal.core.SourceType.newSupertypeHierarchy(SourceType.java:809)
        at org.eclipse.jdt.internal.core.search.IndexSelector.getFocusedElementsAndTypes(IndexSelector.java:172)
        at org.eclipse.jdt.internal.core.search.IndexSelector.initializeIndexLocations(IndexSelector.java:233)
        at org.eclipse.jdt.internal.core.search.IndexSelector.getIndexLocations(IndexSelector.java:304)
        at org.eclipse.jdt.internal.core.search.JavaSearchParticipant.selectIndexURLs(JavaSearchParticipant.java:160)
        at org.eclipse.jdt.internal.core.search.PatternSearchJob.getIndexes(PatternSearchJob.java:186)
        at org.eclipse.jdt.internal.core.search.PatternSearchJob.ensureReadyToRun(PatternSearchJob.java:99)
        at org.eclipse.jdt.internal.core.search.processing.JobManager.performConcurrentJob(JobManager.java:220)
        at org.eclipse.jdt.internal.core.search.BasicSearchEngine.findMatches(BasicSearchEngine.java:235)
        at org.eclipse.jdt.internal.core.search.BasicSearchEngine.search(BasicSearchEngine.java:602)
        at org.eclipse.jdt.core.search.SearchEngine.search(SearchEngine.java:670)
        at org.eclipse.jdt.core.tests.model.JavaSearchTests.searchForMethodReferences(JavaSearchTests.java:4696)
        at org.eclipse.jdt.core.tests.model.JavaSearchTests.testAnonymousTypeMethodReferenceSearchGh432(JavaSearchTests.java:4680)
```

The exception occurs also when using Eclipse to compute the call hierarchy in this snippet...

trancexpress: Though by the looks if an extra `JavaModelException` occurs in the result collector used for the test...

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchTests.java
trancexpress: > Chewing on the question whether the source indexer is the correct tool to resolve the bindings ... I wonder how it can work together with `var` typed locals that are used in method references, e.g.
> 
> ```
>   var someList = new ArrayList<String>();
>   withAcceptor(someList.subList(0,1)::add);
> ```
> 
> Is the source indexer in the position to compute the correct type information for "all" shapes of local vars such that method references can work?

I'm not sure I understand your question. You expect `SourceIndexer` to struggle with this example? There is no problem with this snippet:

```
package test;

import java.util.ArrayList;
import java.util.List;
import java.util.function.Consumer;

public class TestVarMethodReference {

	static class TestList<T> {

		private final List<T> list;

		TestList() {
			this(new ArrayList<>());
		}

		private TestList(List<T> list) {
			this.list = list;
		}

		public void add(T o) {
			list.add(o);
		}

		public TestList<T> sub(int s, int e) {
			return new TestList<>(list.subList(s, e));
		}

		@Override
		public String toString() {
			return list.toString();
		}
	}

		var list = new TestList<String>();
		list.add("1");
		list.add("2");
		list.add("3");
		list.add("4");
		withAcceptor(list.sub(0, 1)::add);
		System.out.println(list);
	}

	private static void withAcceptor(Consumer<String> l) {
		l.accept(".1");
		l.accept(".2");
	}
}
```

The method reference to the `add` method is found.

I'm guessing everything is resolved here and so the method reference is also "seen".

trancexpress:
The current fail is:

```
junit.framework.AssertionFailedError: Compilation unit should not exist
	at org.eclipse.jdt.core.tests.model.ClasspathTests.testNoResourceChange03(ClasspathTests.java:4983)
```

No such fail when ran locally.

iloveeclipse:
Simeon, I've rebased, slightly fixed & force pushed your patch. The only change I've made was to remove empty project files.

Issue #438, opened 2022-09-23
Call hierarchy does not find some uses via method reference


trancexpress:
See: https://bugs.eclipse.org/bugs/show_bug.cgi?id=571913

To reproduce:

1. Create a new Java project, JRE is 11. No `module-info.java` file.
2. First paste this snippet into the project:
```
___ MethodRef.java ___
import java.util.ArrayList;
import java.util.List;

class MethodRef {
	List<User> users = new ArrayList<>();
	
	void printStuff() {
		doNames();
	}

	private void doNames() {
		users.forEach(User::printName);
	}
}
```
3. Then paste this snippet into the project:
```
___ User.java ___
class User {
	String name;

	void printName() {
		System.out.println(name);
	}
}
```
4. Compute the call hierarchy of `User.printName()`, observe that its empty.

When reversing the paste order, the call hierarchy is not empty.

trancexpress:
When debugging, I see the following stack trace computes a `null` `MethodBinding` for the method reference. And so it doesn't add the method reference to the index:

```
"Java indexing" #61 daemon prio=5 os_prio=0 cpu=372.19ms elapsed=330.62s tid=0x00007ffff0bdd000 nid=0xdbef runnable  [0x00007fff5bdfc000]
   java.lang.Thread.State: RUNNABLE
        at org.eclipse.jdt.internal.core.search.indexing.SourceIndexer.indexResolvedDocument(SourceIndexer.java:280)
        at org.eclipse.jdt.internal.core.search.JavaSearchParticipant.indexResolvedDocument(JavaSearchParticipant.java:103)
        at org.eclipse.jdt.internal.core.search.indexing.IndexManager.indexResolvedDocument(IndexManager.java:677)
        at org.eclipse.jdt.internal.core.search.indexing.IndexManager$2.execute(IndexManager.java:1303)
        at org.eclipse.jdt.internal.core.search.processing.JobManager.run(JobManager.java:478)
        at java.lang.Thread.run(java.base@11.0.10/Thread.java:834)
```

Adding the missing source however does result in a re-compile of the source. I don't know why the indexer doesn't go over this code again to add the method reference once it can be resolved...

PR #439, opened 2022-09-23
No call hierarchy for method reference after fixing compile error #438

trancexpress:
SourceIndexer.indexResolvedDocument() adds an index entry for a method
reference only if the expression is resolved. If the expression becomes
resolved later on, e.g. due to fixing compile errors in another file,
SourceIndexer is not triggered again. So the index for the file remains
incomplete and the method reference is not found e.g. when computing the
call hierarchy of the referenced method.

This change adds a new index category, an unresolved method reference.
References of methods with the same name and different arguments count
will not be distinguished in this new category, as the argument count is
not known at the time of adding the unresolved method reference to the
index.

Fixes: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/438

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchBugsTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchBugsTests.java
@@ -21,6 +21,7 @@ import static org.eclipse.jdt.core.search.IJavaSearchScope.SYSTEM_LIBRARIES;
 import java.io.File;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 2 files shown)


trancexpress:
This is band-aid but I don't know what else to do here.

I don't know `jdt.core` well enough to trigger an extra indexing of a source, if its functional expressions got fixed during a compile round. And I don't know where else code can check whether the functional expression has been resolved "meanwhile", when asking the indexer during the search.

trancexpress:
@iloveeclipse this is a workaround for the problem and it won't work for referenced methods that share the same name but have different argument count.

Still, IMO better than nothing and its the only simple fix I could come up with.

iloveeclipse:
@gayanper : could you please review this, if you have time?

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/IIndexConstants.java
```diff
@@ -21,6 +21,7 @@ public interface IIndexConstants {
 	char[] REF= "ref".toCharArray(); //$NON-NLS-1$
 	char[] ANNOTATION_REF= "annotationRef".toCharArray(); //$NON-NLS-1$
 	char[] METHOD_REF= "methodRef".toCharArray(); //$NON-NLS-1$
+	char[] METHOD_REF_UNRESOLVED= "methodRefUnresolved".toCharArray(); //$NON-NLS-1$
```
iloveeclipse, line 24: I know not much about indexer, but it sounds strange that we add **unresolved** methods to the index, I don't see any comparable methods in the existing code.

Assuming the reference was really wrong and was just a "typo" in the code - what kind of side effects can that "wrong" reference in the index cause?

gayanper:
@trancexpress first of all thanks for trying out to fix the issue. it really good that you have gone through the indexer code and try to figure out a way for fix it. Really appreciate.

Instead of indexing unresolved information, which could lead to other issues if we cannot extract information like parameter and return type information which is required for more complex search operations, have you thought about finding away by listening into the compiler to see what documents we should index ? For example when we skip the referenced document in indexing because this method ref in unresolved, we could push it to some place and start listening into to compiler, when a compiler cycle is finished successfully we could retry to index these pending documents. I think if you look at the IndexManager and JobManager you could find a sweet spot to intercept since JobManager is responsible for controlling build jobs and indexing jobs.

Let me know if you have more questions around this area.

trancexpress:
> Instead of indexing unresolved information
See: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/439#issuecomment-1291676906

> I don't know `jdt.core` well enough to trigger an extra indexing of a source, if its functional expressions got fixed during a compile round.

I wouldn't trigger extra indexing in general after a compile of a file, no idea how much extra processing this will result in overall. Aside from the bug here it seems like it would be unnecessary; I'm not aware of other bugs like this one. If the file had compile errors and they are gone after a compile, triggering indexing would be fine from my POV (since compile errors should not be the norm and extra processing after fixing them would be fine).

If @iloveeclipse wants to invest more time in the bug, I could check. From my POV the bug is too rare for that.

gayanper:
> I wouldn't trigger extra indexing in general after a compile of a file, no idea how much extra processing this will result in >overall. Aside from the bug here it seems like it would be unnecessary;
It will not be an unnecessary indexing i think, You will only index files which had compiler errors and failed to index because they had unresolved symbols.

>If the file had compile errors and they are gone after a compile, triggering indexing would be fine from my POV (since >compile errors should not be the norm and extra processing after fixing them would be fine).
We will not do a full reindex, only files that were skipped due to unresolved information + was edited, when you edit a file it cause a change event (if i remember correct some kind of a delta event) which triggers building and indexing. So the indexer will only index that file which was changed by that change event. jdt.core IndexManager class is a good place to start looking at, it has all these fine grain methods which act on those events.

jarthana:
Tagging @mpalat, who might have some idea about the indexer.

trancexpress:
I have some time now to look into this, I've pushed updated suggestion: trigger re-indexing if JDT error markers are being removed from a compilation unit.

We maybe want to trigger re-indexing if no JDT error markers remain.

@gayanper do you have time / interest to check the new patch?

gayanper (approved):
LGTM

iloveeclipse:
Last merge commit was unfortunate. Ideally Simeon could just pick the first commit alone, rebase on master and push again.

PR #441, opened 2022-09-25
Improve scanner "when" token disambiguation

MaisiKoleni:
## What it does
The contextual keyword "when" for the JLS 19 preview feature pattern matching in switch is currently not detected correctly.

Example:
```java
class X {
    void someMethod(Predicate<Object> condition) {
        SomeClass.when(condition).doSomething();
    }
}
```
The problem:
```
Syntax error on token "RestrictedIdentifierWhen", Identifier expected
```
I noticed that problem with JavaFX EasyBind, but this is probably a common method name in libraries.

This PR also contains a refactoring and small semantic change regarding `lookBack` in the `Scanner`, see the provided Javadoc + test case for that and the commit messages.

The new rules for `when` are based on https://docs.oracle.com/javase/specs/jls/se19/preview/specs/patterns-switch-record-patterns-jls.html#jls-14.11.1

## How to test
See the provided example or the provided test.

## Author checklist

- [x] I have thoroughly tested my changes (on my machine, all parser and compiler regression tests still pass)

### DIFF
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/ScannerTest.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/ScannerTest.java
@@ -1836,4 +1836,25 @@ public class ScannerTest extends AbstractRegressionTest {
 			assertTrue(false);
 		}
 	}
+
+	public void testWhenAsIdentifier() {
+		String source =
+				"public void when(Object when) {\n" +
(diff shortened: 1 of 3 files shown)


MaisiKoleni:
Please tell me whether the (new) approach of having only allowed tokens in the switch for the look back is OK. I can also adjust the way `lookBack` works or revert the change, however solving the problem might get more difficult.

jarthana (approved):
Honestly, I can't decide about the TokenNameNotAToken in the switch/case (TODO) change. I guess we can let it be.

Issue #485, opened 2022-10-20
Java editor cannot resolve import due to classpath filter


trancexpress:
To reproduce:

1. Import projects from attached archive: [ClasspathFilterBugExample.zip](https://github.com/eclipse-jdt/eclipse.jdt.core/files/9830786/ClasspathFilterBugExample.zip)
2. Open source `F.java`.
3. Observe that the only import cannot be resolved in the Java editor, although the projects have no compile errors.

I don't see the problem with JRE 8 on the classpath, but I do see it with JRE 11. I went as far back as Eclipse 4.10 and still see the same behavior (bug with JRE 11, no bug with JRE 8).

trancexpress:
Looks like `ModuleBinding.declaredPackages` doesn't contain the `com` package from one of the example projects. This results in the following code running in `ModuleBinding.getVisiblePackage(PackageBinding, char[])` (which, without a classpath filter, doesn't run):

```
						for (char[] declaringModuleName : declaringModuleNames) {
							ModuleBinding declaringModule = this.environment.root.getModule(declaringModuleName);
							if (declaringModule != null) {
								PlainPackageBinding declaredPackage = declaringModule.getDeclaredPackage(fullFlatName);
								binding = SplitPackageBinding.combine(declaredPackage, binding, this);
							}
						}
```

This then uses the `com` package from the JRE (due to `com.sun....`) and creates a split package that doesn't contain the unnamed module. So the `com.g` package from the example is not found in that split package for `com`. Then the problem is reported in the editor.

During a compile, `ModuleBinding.declaredPackages` does have the `com` package from the example project, so the compile works fine.

trancexpress:
Having debugged more, the exact problem seems to be this code in `SearchableEnvironment.getModulesDeclaringPackage(char[][], char[])`

```
case AnyNamed:
	char[][] names = CharOperation.NO_CHAR_CHAR;
	// narrow down candidates of roots (https://bugs.eclipse.org/566498)
	IPackageFragmentRoot[] matchingRoots = this.nameLookup.findPackageFragementRoots(pkgName);
	if(matchingRoots != null) {
```

When the classapth inclusion filter is set as in the attached reproducer, the respective project has only 1 package fragment root, `com.g.f.t.f`. So the code above searches with `com`, finds `com` from the JRE, and then associates `com` with the  JRE modules. It doesn't know the project (from the reproducer) also contributes `com` from the unnamed module.

[...]

PR #490, opened 2022-10-24
Java editor cannot resolve import due to classpath filter #485

trancexpress:
When using the classpath "including" and "excluding" entry properties, its possible that a top level package is not listed as a IPackageFragmentRoot. As a result,
SearchableEnvironment.getModulesDeclaringPackage() fails to recognize such a package as belonging to the unnamed module. This can result in Java editor unresolvable type compile errors.

This change adjusts SearchableEnvironment.getModulesDeclaringPackage() to consider parents of packages in the unnamed module to also belong in the unnamed module.

Fixes: #485

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/InclusionPatternsTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/InclusionPatternsTests.java
@@ -458,6 +458,14 @@ public void testIncludeCUOnly01_new() throws CoreException {
 			workingCopy.discardWorkingCopy();
 	}
 }
+
+/**
+ * Same test as above but checking in modular environment, where search failed to see package in unnamed module
+ */
+public void testIncludeCUOnly01_new_modular() throws CoreException {
+	this.project.setOption(JavaCore.COMPILER_COMPLIANCE, "11");
+	testIncludeCUOnly01_new();
+}
 /**
  * Ensure that a type can be resolved if it is included but not its super packages.
  * (regression test for bug 119161 classes in "deep" packages not fully recognized when using tight inclusion filters)
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/SearchableEnvironment.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/SearchableEnvironment.java
@@ -87,6 +87,8 @@ public class SearchableEnvironment
 	private long timeSpentInGetModulesDeclaringPackage;
 	private long timeSpentInFindTypes;
 
+	private List<IPackageFragmentRoot> unnamedModulePackageFragmentRoots;
+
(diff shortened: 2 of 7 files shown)


trancexpress:
@iloveeclipse who can review this? We need someone with good module knowledge.

I don't know if a package can be in the unnamed module, while a parent of that package belongs to an actual module.

Also this doesn't fix the cases where modules are actually defined and used. I'm sure cases like that can be constructed and the change here doesn't fix such cases.

iloveeclipse:
@stephan-herrmann : if you would have some free time to look at this PR?
The example project that shows the error is at #485.

@trancexpress : would it be possible to add a test case that shows the problem (even if the patch is wrong, test could be used to show the problem easier)?

trancexpress:
> @trancexpress : would it be possible to add a test case that shows the problem (even if the patch is wrong, test could be used to show the problem easier)?

I'll try to do that next, yes.

trancexpress:
@iloveeclipse , this is ready for review.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/InclusionPatternsTests.java
```diff
@@ -458,6 +458,14 @@ public void testIncludeCUOnly01_new() throws CoreException {
 			workingCopy.discardWorkingCopy();
 	}
 }
+
+/**
+ * Same test as above but checking in modular environment, where search failed to see package in unnamed module
+ */
+public void testIncludeCUOnly01_new_modular() throws CoreException {
+	this.project.setOption(JavaCore.COMPILER_COMPLIANCE, "11");
```
iloveeclipse, line 466: This is from https://github.com/eclipse-jdt/eclipse.jdt.core/issues/646

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/SearchableEnvironment.java
```diff
@@ -102,10 +104,10 @@ public SearchableEnvironment(JavaProject project, org.eclipse.jdt.core.ICompilat
 			|| !JavaCore.IGNORE.equals(project.getOption(JavaCore.COMPILER_PB_DISCOURAGED_REFERENCE, true));
 		this.workingCopies = workingCopies;
 		this.nameLookup = project.newNameLookup(workingCopies, excludeTestCode);
-		if (CompilerOptions.versionToJdkLevel(project.getOption(JavaCore.COMPILER_COMPLIANCE, true)) >= ClassFileConstants.JDK9) {
```
iloveeclipse, line 105: I've removed that continuous checking of same condition and merged three if blocks in one.
Boy Scout rule. It was dirty before, we should not add more dirt. It is cleaner now.

iloveeclipse (approved):
LGTM. Waiting for tests results to merge.

Issue #504, opened 2022-10-29
Add support for array type completion


gayanper:
```
Object[] array = new O|
```
Taking the above code snippet, the developer would expect a completion which complete the above statement as

```
Object[] array = new Object[]
```

So that the UI or JavaLS can provide linked editing support to complete the statement as follows

```
Object[] array = new Object[|]
```

or 

```
Object[] array = new Object[]|
```

*Note*: Note that the `|` denotes the cursor location on a editor.

PR #507, opened 2022-10-30
Add support for array type completions #504

gayanper:
Enhance the existing support for discovering array component type by
adding extra information into CompletionProposal so that the consumers
can decide how the edits should be handled in the editors like jdt.ui
and jdt.ls

### DIFF
--- a/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
+++ b/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
@@ -12187,8 +12187,17 @@ public final class CompletionEngine
 
 				checkCancel();
 
-				if(this.expectedTypes[i] instanceof ReferenceBinding) {
-					ReferenceBinding refBinding = (ReferenceBinding)this.expectedTypes[i];
+				TypeBinding expectedType = this.expectedTypes[i];
+				TypeBinding originalExpectedType = expectedType;
+				int dimensions = 0;
+				if (expectedType instanceof ArrayBinding) {
+					ArrayBinding arrayBinding = (ArrayBinding) expectedType;
+					expectedType = arrayBinding.leafComponentType();
+					dimensions = arrayBinding.dimensions();
+				}
+
+				if (expectedType instanceof ReferenceBinding) {
+					ReferenceBinding refBinding = (ReferenceBinding) expectedType;
 
 					if (typeLength > 0) {
 						if (typeLength > refBinding.sourceName.length) continue next;
@@ -12270,7 +12279,9 @@ public final class CompletionEngine
 						relevance += computeRelevanceForResolution();
 						relevance += computeRelevanceForInterestingProposal(refBinding);
 						relevance += computeRelevanceForCaseMatching(token, typeName);
-						relevance += computeRelevanceForExpectingType(refBinding);
+						// if an array, then we need to use the original expected type.
+						relevance += computeRelevanceForExpectingType(
+								dimensions > 0 ? originalExpectedType : refBinding);
 						relevance += computeRelevanceForQualification(isQualified);
 						relevance += computeRelevanceForRestrictions(accessibility);
 
@@ -12284,7 +12295,7 @@ public final class CompletionEngine
 						}
 
 						if (proposeType &&
-								(!this.assistNodeIsConstructor ||
+								(!this.assistNodeIsConstructor || dimensions > 0 ||
 										!allowingLongComputationProposals ||
 										hasStaticMemberTypes(refBinding, scope.enclosingSourceType() ,this.unitScope)) ||
 										hasArrayTypeAsExpectedSuperTypes()) {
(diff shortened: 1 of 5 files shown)


gayanper:
@rgrunber @stephan-herrmann I add both of you from jdt and jdt.ls perspective. My plan is to initially add editor support in jdt.ls for array completions.

jdneo:
Vote for this one. I had a [try](https://git.eclipse.org/r/c/jdt/eclipse.jdt.core/+/191208) to support array completion

This one look neater and does not require to create a new proposal type 👍

stephan-herrmann:
@gayanper just a quick ACK: saw your request and will try to find time to look into it soonish.

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
```diff
@@ -12083,135 +12083,157 @@ private void findTypesFromExpectedTypes(char[] token, Scope scope, ObjectVector
...
+				boolean receiverIsArrayType = false;
 				if(this.expectedTypes[i] instanceof ReferenceBinding) {
-					ReferenceBinding refBinding = (ReferenceBinding)this.expectedTypes[i];
+					refBinding = (ReferenceBinding) this.expectedTypes[i];
+				} else if (this.expectedTypes[i] instanceof ArrayBinding) {
+					receiverIsArrayType = true;
+					TypeBinding leaf = ((ArrayBinding) this.expectedTypes[i]).leafComponentType();
+					refBinding = this.lookupEnvironment
+							.getType(CharOperation.splitOn('.', CharOperation.concat(leaf.qualifiedPackageName(),
+									leaf.qualifiedSourceName(), '.')));
```
stephan-herrmann, line 12095: I'm not totally sure what's going on here:
* why the detour from an existing binding via its name to a new binding?
* what if `refBinding` is not a valid type binding afterwards?
* I'm missing an explicit check for array-of-primitive, perhaps that's what the detour achieves in an obscure way? :)

stephan-herrmann (commented):
Sorry for the delay.

The change generally looks good. See one inline question.

Confidence could perhaps still be increased by some new tests, and/or a companion PR to JDT/UI where the new flag is evaluated to bring some added value.

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
gayanper: I will check this again @stephan-herrmann and update. I think I have to refresh my approach here :).

gayanper:
@stephan-herrmann I reworked the solution, eliminating some white space changes and improving on the logic as you have pointed out in earlier fix.

stephan-herrmann:
> [gayanper](https://github.com/gayanper) [force-pushed](https://github.com/eclipse-jdt/eclipse.jdt.core/compare/f40e24d63d1a06b26d6240a41ac0278ed8a4443e..a89797dd95c78159de2bdc98372e1fe19e2ed6cc) the add_array_completions branch from[ f40e24d ](https://github.com/eclipse-jdt/eclipse.jdt.core/commit/f40e24d63d1a06b26d6240a41ac0278ed8a4443e)to a89797d yesterday

Didn't we agree on avoiding force push to facilitate re-reviews? :smile:

Anyway, I'm looking at this now ...

stephan-herrmann (approved):
From jdt.core side this looks good, now.

I recall in my previous review I was asking about seeing this in action in a client (just the review appears to be gone due to force push).

Anyway, I still suggest to demonstrate in either jdt.ls or jdt.ui that the new API serves the purpose. Merging this PR early in 4.28 should allow as to make any required changes before the next release...

gayanper:
> > [gayanper](https://github.com/gayanper) [force-pushed](https://github.com/eclipse-jdt/eclipse.jdt.core/compare/f40e24d63d1a06b26d6240a41ac0278ed8a4443e..a89797dd95c78159de2bdc98372e1fe19e2ed6cc) the add_array_completions branch from[ f40e24d ](https://github.com/eclipse-jdt/eclipse.jdt.core/commit/f40e24d63d1a06b26d6240a41ac0278ed8a4443e)to [a89797d](https://github.com/eclipse-jdt/eclipse.jdt.core/commit/a89797dd95c78159de2bdc98372e1fe19e2ed6cc) yesterday
> 
> Didn't we agree on avoiding force push to facilitate re-reviews? 😄
> 
> Anyway, I'm looking at this now ...

Yes we agreed, But for this change since I reworked to avoid unwanted whitespace changes it was easy for me to start from a new branch and push to the same PR remote branch to keep the PR open. Didn't want to spend lot of time fiddling with git. Sorry for the inconvenience with force push.

Your previous commands are there, they are just collapsed since I resolved the discussion thread and they are also outdated since the code is changed.

gayanper:
> From jdt.core side this looks good, now.
> 
> I recall in my previous review I was asking about seeing this in action in a client (just the review appears to be gone due to force push).
> 
> Anyway, I still suggest to demonstrate in either jdt.ls or jdt.ui that the new API serves the purpose. Merging this PR early in 4.28 should allow as to make any required changes before the next release...

Since I now use vscode as my main driver I can work on adding the support into jdt.ls. But with this change for unsupported consumer of JDT there are not significant changes. Here is a screencast

https://user-images.githubusercontent.com/7471994/222981296-0d0762a2-ff5d-473d-bdde-3a426e3d5ba3.mp4

as you can see it suggest the type reference only and for primitives, I think we need to handle that from jdt.ui side.

PR #508, opened 2022-10-30
Failed to infer types during nested generic method invocation

### LINKED ISSUE #506: Failed to infer types during nested generic method invocation
```java

class Convert<A> {
    public static <B> void test(final B a) {
        Convert<B> res1 = convert(arg -> create(arg), a); // ok

        Convert<B> res2 = convert(arg -> wrap(create(arg)), a); // error: Type mismatch
    }

    public static <C, D> Convert<D> convert(final Function<C, Convert<D>> func, final C value) {
        return null;
    }

    public static <E> E wrap(final E a) {
        return null;
    }

    public static <F> Convert<F> create(final F initial) {
        return null;
    }
}
```

javac compiles fine on the snippet above, but ecj reports an error about `Type mismatch: cannot convert from Convert<Object> to Convert<B>` on `res2`.

stephan-herrmann:
Fixes #506

## What it does
This is an opportunistic attempt at reconciling the existing fix in https://github.com/eclipse-jdt/eclipse.jdt.core/commit/51ed0947dc59c8e4cadd7d626d5ed089ccf3bd98  with the example in #506

I glimpsed that https://bugs.eclipse.org/511252#c17 motivated changes in 51ed0947 by the need to defer decision between strict vs. varargs modes of inference. Obviously this is only relevant if the target method is indeed varargs.

## How to test
See the new test case, but beware that we have no backing by JLS in this area. Other corner cases not yet covered by our test suite may regress due to this change.

It would be interesting to see if the same problem in #506 can actually be reproduced with a varargs method, in which case the attempted fix is expected to fail.

### DIFF
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericsRegressionTest_1_8.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericsRegressionTest_1_8.java
@@ -10379,4 +10379,31 @@ public void testBug508834_comment0() {
 				"}\n"
 			});
 	}
+	public void testIssue506() {
+		runConformTest(
+			new String[] {
+				"Convert.java",
+				"import java.util.function.Function;\n" +
+				"\n" +
+				"class Convert<A> {\n" +
+				"    public static <B> void test(final B a) {\n" +
+				"        Convert<B> res1 = convert(arg -> create(arg), a); // ok\n" +
+				"\n" +
+				"        Convert<B> res2 = convert(arg -> wrap(create(arg)), a); // error: Type mismatch\n" +
+				"    }\n" +
+				"\n" +
+				"    public static <C, D> Convert<D> convert(final Function<C, Convert<D>> func, final C value) {\n" +
+				"        return null;\n" +
+				"    }\n" +
+				"\n" +
+				"    public static <E> E wrap(final E a) {\n" +
+				"        return null;\n" +
+				"    }\n" +
+				"\n" +
+				"    public static <F> Convert<F> create(final F initial) {\n" +
+				"        return null;\n" +
+				"    }\n" +
+				"}\n"
+			});
+	}
 }
--- a/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/ParameterizedGenericMethodBinding.java
+++ b/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/ParameterizedGenericMethodBinding.java
@@ -299,7 +299,7 @@ public class ParameterizedGenericMethodBinding extends ParameterizedMethodBindin
 							methodSubstitute = new PolyParameterizedGenericMethodBinding(methodSubstitute);
 						}
 					} finally {
-						if (allArgumentsAreProper) {
+						if (allArgumentsAreProper || !methodSubstitute.isVarargs()) {
 							if (invocationSite instanceof Invocation)
 								((Invocation) invocationSite).registerInferenceContext(methodSubstitute, infCtx18); // keep context so we can finish later
 							else if (invocationSite instanceof ReferenceExpression)


testforstephen:
@stephan-herrmann thanks for the quick fix. Tried it on the example in #506, it works fine.

And then I modified the example to a varargs version, it fails as expected.
```java {.lineNo}

class Convert<A> {
    public static <B> void test(final B a) {
        // Convert<B> res1 = convert(arg -> create(arg), a); // ok

        Convert<B[]> res2 = convert(arg -> wrap(create(arg)), a); // Type mismatch error
    }

    public static <C, D> Convert<D[]> convert(final Function<C, Convert<D[]>> func, final C value) {
        return null;
    }

    public static <E> E wrap(final E a) {
        return null;
    }

    public static <F> Convert<F[]> create(final F... initial) {
        return null;
    }
}
```

However, javac still compiles fine on the varargs case. 

Take a look at the javac inference details from the command `javac --debug=verboseResolution=all Convert.java`. It seems it always infers inner recursively, and eventually resolves the actual return type of `convert()` as Convert<B[]>. 

Regarding how to deal with the varargs method `create`, it first applies the applicability inference in strict mode and returns false, then falls back to VARARITY mode.
```
Convert.java:3: Note: resolving method <init> in type Object to candidate 0
class Convert<A> {
^
  phase: BASIC
  with actuals: no arguments
  with type-args: no arguments
  candidates:
      #0 applicable method found: Object()
Convert.java:7: Note: resolving method convert in type Convert to candidate 0
        Convert<B[]> res2 = convert(arg -> wrap(create(arg)), a); // Type mismatch error
                            ^
  phase: BASIC
  with actuals: <none>,B
  with type-args: no arguments
  candidates:
      #0 applicable method found: <C,D>convert(Function<C,Convert<D[]>>,C)
        (partially instantiated to: (Function<B,Convert<Object[]>>,B)Convert<Object[]>)
  where B,C,D are type-variables:
    B extends Object declared in method <B>test(B)
    C extends Object declared in method <C,D>convert(Function<C,Convert<D[]>>,C)
    D extends Object declared in method <C,D>convert(Function<C,Convert<D[]>>,C)
Convert.java:7: Note: resolving method wrap in type Convert to candidate 0
        Convert<B[]> res2 = convert(arg -> wrap(create(arg)), a); // Type mismatch error
                                           ^
  phase: BASIC
  with actuals: Convert<F[]>
  with type-args: no arguments
  candidates:
      #0 applicable method found: <E>wrap(E)
        (partially instantiated to: (Convert<B[]>)Convert<B[]>)
  where F,E,B are type-variables:
    F extends Object declared in method <F>create(F...)
    E extends Object declared in method <E>wrap(E)
    B extends Object declared in method <B>test(B)
Convert.java:7: Note: erroneous resolution for method create in type Convert
        Convert<B[]> res2 = convert(arg -> wrap(create(arg)), a); // Type mismatch error
                                                ^
  phase: BASIC
  with actuals: B
  with type-args: no arguments
  candidates:
      #0 not applicable method found: <F>create(F...)
        (cannot infer type-variable(s) F
          (argument mismatch; B cannot be converted to F[]))
  where B,F are type-variables:
    B extends Object declared in method <B>test(B)
    F extends Object declared in method <F>create(F...)
Convert.java:7: Note: erroneous resolution for method create in type Convert
        Convert<B[]> res2 = convert(arg -> wrap(create(arg)), a); // Type mismatch error
                                                ^
  phase: BOX
  with actuals: B
  with type-args: no arguments
  candidates:
      #0 not applicable method found: <F>create(F...)
        (cannot infer type-variable(s) F
          (argument mismatch; B cannot be converted to F[]))
  where B,F are type-variables:
    B extends Object declared in method <B>test(B)
    F extends Object declared in method <F>create(F...)
Convert.java:7: Note: resolving method create in type Convert to candidate 0
        Convert<B[]> res2 = convert(arg -> wrap(create(arg)), a); // Type mismatch error
                                                ^
  phase: VARARITY
  with actuals: B
  with type-args: no arguments
  candidates:
      #0 applicable method found: <F>create(F...)
        (partially instantiated to: (B[])Convert<B[]>)
  where B,F are type-variables:
    B extends Object declared in method <B>test(B)
    F extends Object declared in method <F>create(F...)
Convert.java:7: Note: Deferred instantiation of method <C,D>convert(Function<C,Convert<D[]>>,C)
        Convert<B[]> res2 = convert(arg -> wrap(create(arg)), a); // Type mismatch error
                                   ^
  instantiated signature: (Function<B,Convert<B[]>>,B)Convert<B[]>
  target-type: Convert<B[]>
  where C,D,B are type-variables:
    C extends Object declared in method <C,D>convert(Function<C,Convert<D[]>>,C)
    D extends Object declared in method <C,D>convert(Function<C,Convert<D[]>>,C)
    B extends Object declared in method <B>test(B)
Convert.java:7: Note: resolving method metafactory in type LambdaMetafactory to candidate 0
        Convert<B[]> res2 = convert(arg -> wrap(create(arg)), a); // Type mismatch error
                                    ^
  phase: BASIC
  with actuals: Lookup,String,MethodType,MethodType,MethodHandle,MethodType
  with type-args: no arguments
  candidates:
      #0 applicable method found: metafactory(Lookup,String,MethodType,MethodType,MethodHandle,MethodType)
Note: Convert.java uses unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
```

stephan-herrmann:
> And then I modified the example to a varargs version, it fails as expected.

Great. So we know that my opportunistic fix was wrong in switching behaviour based on `isVarargs()`.

> Take a look at the javac inference details from the command javac --debug=verboseResolution=all Convert.java

That's valuable information indeed. From discussion with javac developer Maurizio Cimadamore I know that javac organizes inference differently from JLS in terms of what is processed top-down vs bottom-up and how information is transferred between nested inferences.

In particular I assume that javac finds inner applicability inference in strict/lose modes to fail due to information that is not available in ecj at that point in time (notably we don't yet have a proper target type).

So javac and ecj differ in such "organizational" aspects, which ideally shouldn't influence observable behavior, but we know that unfortunately these difference are relevant, indeed. 

With this in mind we'll have to shift gears from faithfully implementing JLS towards our own decisions how to connect the individual steps of inference (which are well understood, specified and implemented) into an algorithm that handles complex input in ways that are similar in effect to what javac is doing.

Implementationwise my partially successful experiment suggests, that we might want to focus on `MessageSend.inferenceContexts`, which connects the location of applicability inference (in `PMGB.computeCompatibleMethod18`) to `ConstraintExpressionFormula.reduce(InferenceContext18)`, where compilation made the "wrong" decision.

TODO 1: In the end we will need to inspect other implementations of `Invocation` to check whether the kind of invocation has an impact on our solution. This is just to be sure that focusing on `MessageSend.inferenceContexts` didn't lead us down a wrong alley.

Using the call hierarchy on field  `MessageSend.inferenceContexts` (all references) we get a pretty clear picture of which code locations are potentially involved.
* via `registerInferenceContext` it's only the well known location in `PMGB.computeCompatibleMethod18`, :+1: 
* via `getInferenceContext` see 6 callers, one of which is `ConstraintExpressionFormula.reduce` indeed
* call paths through `cleanUpInferenceContexts` can be ignored in this investigation as these happen only much later during compilation when resolution with type inference has long completed :+1: 

TODO 2: after dealing with interaction between these two methods, we need to inspect the other 5 callers of `getInferenceContext` to see if those might be affected by any change we create, or need individual treatment.

I don't have a strategy _how_ exactly our implementation should be changed, since "restarting in varargs mode" may not be technically feasible right at the moment when we see a failure. But perhaps the following idea could spark some experiments towards such solution:

What if `PMGB.computeCompatibleMethod18` would never *skip* `registerInferenceContext()`, but set a new flag inside the `InferenceContext18` which could then be used in `ConstraintExpressionFormula.reduce` in order to take the steps necessary at that point and in knowledge of more intermediate inference results? Foremost we need to avoid the block commented with "assume it wasn't really poly after all". At some point you will see a number of regressions which might give an idea (during debugging) what needs to be done _instead_ of that block which is now known to be wrong for the cases at hand. 

Do you want to experiment in this direction and report back?

testforstephen:
> Do you want to experiment in this direction and report back?

Sure, let me have a try on this.

testforstephen:
Hey @stephan-herrmann, I have experimented with your suggestion, it solved the issues above with no regression.

Here are the full changes I made.

```diff
diff --git a/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/ast/ASTNode.java b/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/ast/ASTNode.java
index c8d6ba877f..fb30edf64e 100644
--- a/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/ast/ASTNode.java
+++ b/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/ast/ASTNode.java
@@ -775,7 +775,7 @@ public abstract class ASTNode implements TypeConstants, TypeIds {
 		}
 		if (method instanceof ParameterizedGenericMethodBinding) {
 			InferenceContext18 ic18 = invocation.getInferenceContext((ParameterizedMethodBinding) method);
-			if (ic18 != null)
+			if (ic18 != null && !ic18.isInexactVarargsInference())
 				ic18.flushBoundOutbox(); // overload resolution is done, now perform the push of bounds from inner to outer
 		}
 		if (problemMethod != null)
diff --git a/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/ConstraintExpressionFormula.java b/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/ConstraintExpressionFormula.java
index 966053565d..4918b232ac 100644
--- a/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/ConstraintExpressionFormula.java
+++ b/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/ConstraintExpressionFormula.java
@@ -122,6 +122,9 @@ class ConstraintExpressionFormula extends ConstraintFormula {
 								return FALSE;
 							return ConstraintTypeFormula.create(exprType, this.right, COMPATIBLE, this.isSoft);
 						}
+						if (innerCtx.isInexactVarargsInference()) {
+							return TRUE;
+						}
 						if (innerCtx.stepCompleted >= InferenceContext18.APPLICABILITY_INFERRED) {
 							inferenceContext.integrateInnerInferenceB2(innerCtx);
 						} else {
diff --git a/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java b/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
index ce774b634b..fe024279dc 100644
--- a/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
+++ b/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
@@ -172,6 +172,7 @@ public class InferenceContext18 {
 	private boolean directlyAcceptingInnerBounds = false;
 	/** Not per JLS: pushing bounds from inner to outer may have to be deferred till after overload resolution, store here a runnable to perform the push. */
 	private Runnable pushToOuterJob = null;
+	private boolean isInexactVarargsInference = false;
 
 	public static boolean isSameSite(InvocationSite site1, InvocationSite site2) {
 		if (site1 == site2)
@@ -707,7 +708,7 @@ public class InferenceContext18 {
 			if (innerMethod instanceof ParameterizedGenericMethodBinding)
 				 innerContext = invocation.getInferenceContext((ParameterizedGenericMethodBinding) innerMethod);
 
-			if (innerContext != null) {
+			if (innerContext != null && !innerContext.isInexactVarargsInference()) {
 				MethodBinding shallowMethod = innerMethod.shallowOriginal();
 				innerContext.outerContext = this;
 				if (innerContext.stepCompleted < InferenceContext18.APPLICABILITY_INFERRED) // shouldn't happen, but let's play safe
@@ -1910,4 +1911,12 @@ public class InferenceContext18 {
 		this.b2 = null;
 		this.currentBounds = null;
 	}
+
+	public boolean isInexactVarargsInference() {
+		return this.isInexactVarargsInference;
+	}
+
+	public void setInexactVarargsInference(boolean isInexactVarargsInference) {
+		this.isInexactVarargsInference = isInexactVarargsInference;
+	}
 }
diff --git a/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/ParameterizedGenericMethodBinding.java b/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/ParameterizedGenericMethodBinding.java
index ab1a84f1f4..5cc6c611de 100644
--- a/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/ParameterizedGenericMethodBinding.java
+++ b/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/ParameterizedGenericMethodBinding.java
@@ -243,10 +243,12 @@ public class ParameterizedGenericMethodBinding extends ParameterizedMethodBindin
 			final boolean isPolyExpression = invocationSite instanceof Expression &&   ((Expression) invocationSite).isTrulyExpression() &&
 					((Expression)invocationSite).isPolyExpression(originalMethod);
 			boolean isDiamond = isPolyExpression && originalMethod.isConstructor();
+			boolean isInexactVarargsInference = false;
 			if (arguments.length == parameters.length) {
 				infCtx18.inferenceKind = requireBoxing ? InferenceContext18.CHECK_LOOSE : InferenceContext18.CHECK_STRICT; // engine may still slip into loose mode and adjust level.
 				infCtx18.inferInvocationApplicability(originalMethod, arguments, isDiamond);
 				result = infCtx18.solve(true);
+				isInexactVarargsInference = (result != null && originalMethod.isVarargs() && !allArgumentsAreProper);
 			}
 			if (result == null && originalMethod.isVarargs()) {
 				// check for variable-arity applicability
@@ -299,12 +301,11 @@ public class ParameterizedGenericMethodBinding extends ParameterizedMethodBindin
 							methodSubstitute = new PolyParameterizedGenericMethodBinding(methodSubstitute);
 						}
 					} finally {
-						if (allArgumentsAreProper) {
-							if (invocationSite instanceof Invocation)
-								((Invocation) invocationSite).registerInferenceContext(methodSubstitute, infCtx18); // keep context so we can finish later
-							else if (invocationSite instanceof ReferenceExpression)
-								((ReferenceExpression) invocationSite).registerInferenceContext(methodSubstitute, infCtx18); // keep context so we can finish later
-						}
+						infCtx18.setInexactVarargsInference(isInexactVarargsInference);
+						if (invocationSite instanceof Invocation)
+							((Invocation) invocationSite).registerInferenceContext(methodSubstitute, infCtx18); // keep context so we can finish later
+						else if (invocationSite instanceof ReferenceExpression)
+							((ReferenceExpression) invocationSite).registerInferenceContext(methodSubstitute, infCtx18); // keep context so we can finish later
 					}
 					return methodSubstitute;
 				}
diff --git a/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/Scope.java b/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/Scope.java
index e310a642e6..61491db335 100644
--- a/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/Scope.java
+++ b/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/Scope.java
@@ -844,7 +844,7 @@ public abstract class Scope {
 			if (compilerOptions.sourceLevel >= ClassFileConstants.JDK1_8 && method instanceof ParameterizedGenericMethodBinding && invocationSite instanceof Invocation) {
 				Invocation invocation = (Invocation) invocationSite;
 				InferenceContext18 infCtx = invocation.getInferenceContext((ParameterizedGenericMethodBinding) method);
-				if (infCtx != null)
+				if (infCtx != null && !infCtx.isInexactVarargsInference())
 					return method; // inference is responsible, no need to recheck.
 			}
 		} else if (genericTypeArguments != null && compilerOptions.complianceLevel < ClassFileConstants.JDK1_7) {
```

- Make `PGMB.computeCompatibleMethod18` never to skip `registerInferenceContext()`, and also add a flag `isInexactVarargsInference` to reflect if the applicability inference of varargs method is inferred with proper arguments.
- The change on the caller `InferenceContext18.addConstraintsToC_OneExpr` is used to pass the test case https://github.com/eclipse-jdt/eclipse.jdt.core/blob/351571327d22c10c6f53943892e0bf4a3c8226ba/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericsRegressionTest_1_8.java#L7133
- The change on the caller `ASTNode.resolvePolyExpressionArguments` and `Scope.computeCompatibleMethod` is used to pass the usual case `Convert<B> res1 = convert(arg -> create(arg), a);` on the varargs example above.
- Finally on the caller `ConstraintExpressionFormula.reduce`, I just made it to return `TRUE` if the innercontext is from inexact varargs inference. This can pass the varargs example above on this issue.

  The inexact varargs inference happens during the first round type inference while `MessageSend.findMethodBinding` (The expr `convert((<no type> arg) -> wrap(create(arg)), a)`) looks up the method binding. At this time, the invocation arguments are not proper type (It's inference variable `C`), which causes the strict applicability inference of `create(F[])` to be passed. If we transfer its inference result to outer, that will result in a bound set where `C <: F[]`, `B <: C`, `B <: F[]` (which is FALSE), and finally the invocation type inference fails on the first round type inference. My proposal is to let `ConstraintExpressionFormula.reduce` to return TRUE if the innercontext is from inexact varargs inference.
https://github.com/eclipse-jdt/eclipse.jdt.core/blob/351571327d22c10c6f53943892e0bf4a3c8226ba/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/ast/MessageSend.java#L1018-L1020

  On the second round type inference while `MessageSend.findMethodBinding` tries to `resolvePolyExpressionArguments`, the proper type `B` is passed as arguments while inferring varargs method `create(F[])`. The strict applicability inference would fail, but VARARG check would pass. This is the expected result, transferring its result to outer can help produce better type inference.
https://github.com/eclipse-jdt/eclipse.jdt.core/blob/351571327d22c10c6f53943892e0bf4a3c8226ba/org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/ast/MessageSend.java#L1027

If you're OK with it, I can go ahead and add some unit tests for it. @stephan-herrmann WDYT?

stephan-herrmann:
> Hey @stephan-herrmann, I have experimented with your suggestion, it solved the issues above with no regression.

Wow, you have found your way around some complex code areas in little time! :+1: 

Looking from a distance you changes are well motivated.

May I suggest to create a new PR with your changes including unit tests? That would help me reviewing / playing with those changes. 

This current PR I will abandon as your additional test proved my opportunistic approach wrong. :smile:

testforstephen:
> May I suggest to create a new PR with your changes including unit tests?

@stephan-herrmann Here is the new PR https://github.com/eclipse-jdt/eclipse.jdt.core/pull/524.

stephan-herrmann:
looks like I forgot to close this, when moving over to https://github.com/eclipse-jdt/eclipse.jdt.core/pull/524.

PR #513, opened 2022-11-03
Export o.e.jdt.core.compiler.batch package in same named bundle

akurtakov:
This package is exported by org.eclipse.jdt.core so this looks like simply being forgotten rather than on purpose


rgrunber (approved):
I can confirm it's missing from the ECJ jar, but like you mention, no idea why. Been like this since the manifest was created. Only class is BatchCompiler, but I guess if one is using the ECJ jar, the Export-Package statements are an afterthought ?

Package was already bumped to 3.32.0 so the "exposing" of api should be fine.

Issue #529, opened 2022-11-11
 Make AbstractCompilerTest recognize Java 20 JVM


akurtakov:
That would allow running eclipse I-build tests (https://ci.eclipse.org/releng/job/AutomatedTests/job/ep426I-unit-cen64-gtk3-java20 ) with Java 20 even if no specific Java 20 implementation is done.
Other projects already recognize new bytecode version for months in releases even (https://gitlab.ow2.org/asm/asm/-/commit/8b7090dc7fc280aeff270473658766179cb2024b) so this would allow us to at least proper our existing codebase runs fine on Java 20. 
Majority of failures seen are "java.lang.IllegalStateException: Unrecognized Java version: 20-ea" so these have to be gone to see what's left for runtime support on Java 20(ea).

akurtakov:
Downgrading the issue to just making AbstractCompilerTest recognize Java 20 is newer JVM

PR #530, opened 2022-11-11
 Make AbstractCompilerTest recognize Java 20 JVM 

akurtakov:
Constants for jvm version, bytecode version, tests runtime detection. 
Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/529

### DIFF
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/util/AbstractCompilerTest.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/util/AbstractCompilerTest.java
@@ -321,7 +321,8 @@ public class AbstractCompilerTest extends TestCase {
 	public static int getPossibleComplianceLevels() {
 		if (possibleComplianceLevels == UNINITIALIZED) {
 			String specVersion = System.getProperty("java.specification.version");
-			isJRE19Plus = CompilerOptions.VERSION_19.equals(specVersion);
+			isJRE19Plus = CompilerOptions.VERSION_19.equals(specVersion)
+					|| Integer.parseInt(specVersion) >= Integer.parseInt(CompilerOptions.VERSION_19);
 			isJRE18Plus = isJRE19Plus || CompilerOptions.VERSION_18.equals(specVersion);
 			isJRE17Plus = isJRE18Plus || CompilerOptions.VERSION_17.equals(specVersion);
 			isJRE16Plus = isJRE17Plus || CompilerOptions.VERSION_16.equals(specVersion);


jarthana:
Alex, are you sure you want this in master? I plan to include this in the yet-to-be-created BETA_JAVA20 branch.

akurtakov:
The only public part is public static final String VERSION_20 = "20" . I don't see this implying more than recognizing there is Java 20 .

jarthana:
> The only public part is public static final String VERSION_20 = "20" . I don't see this implying more than recognizing there is Java 20 .

Sure, but at this point, there's Java 20 in JDT :) 
And my point is, it's probably not necessary if the idea is just to be able run with JDK 20 and not run at Java 20. The latter is meaningless as of now.

akurtakov:
So is the last patch better? If not, let me know what exactly is the problem.

jarthana:
> Now that I thought more about it. Even the constant was in internal package thus should be totally safe to have them even up to 29 if we decide to do so.

Yes, perhaps it is harmless. But my concern is what we are trying to achieve here? 

1. Are we trying to run the tests with JRE/JDK20? 
     If so, then the changes to the constants are not required. Perhaps, we need to add a hack (to ignore the "-ea") to be able to run the tests with JDK 20. So far, we never had to do that, but if there's a good reason, we can do that. I will provide that fix if you want.
2. Are we trying to run at Java 20?
    If this is the case, I think it doesn't add any value at this point. We will be running at a level that we don't yet support. Plus, we will be adding tens of thousands of tests for no reason (because tests at 19 and 20 will be same)

Also, note that running at Java 20 means, we will be producing .class files with Java 20 level, which we are prohibited from doing on a branch like master.

akurtakov:
>     1. Are we trying to run the tests with JRE/JDK20?
>        If so, then the changes to the constants are not required. Perhaps, we need to add a hack (to ignore the "-ea") to be able to run the tests with JDK 20. So far, we never had to do that, but if there's a good reason, we can do that. I will provide that fix if you want.

That's what I try to achieve.
> 
>     2. Are we trying to run at Java 20?
>        If this is the case, I think it doesn't add any value at this point. We will be running at a level that we don't yet support. Plus, we will be adding tens of thousands of tests for no reason (because tests at 19 and 20 will be same)

Would you please point to where/how/why this patch adds tens of thousands of tests?

> 
> 
> Also, note that running at Java 20 means, we will be producing .class files with Java 20 level, which we are prohibited from doing on a branch like master.

If the class file version is only added as a constant how would ecj start producing Java 20 level?

akurtakov:
I figured the part about extra tests , it's the additions of new int[] {F_20, ClassFileConstants.MAJOR_VERSION_20}, - I'll remove it now.

jarthana:
> > ```
> > 1. Are we trying to run the tests with JRE/JDK20?
> That's what I try to achieve.

Then, we are doing a lot of things that could have been avoided.

> I figured the part about extra tests , it's the additions of new int[] {F_20, ClassFileConstants.MAJOR_VERSION_20}, - I'll remove it now.

Thanks, but I still think it's better to start recognizing the -ea in java.version if we plan to run tests with EA JDK in future also.

akurtakov:
So this is the minimal fix needed for tests to recognize Java 20.

akurtakov:
So last version of the patch fixes isJRE19Plus to really check for the "Plus" as right now it was equal. That should be fine to run tests on newer JVMs even if its 21, 22, etc.

akurtakov:
@jarthana Can we get this one reevaluated so testing on Java 20 can be looked further https://ci.eclipse.org/releng/job/AutomatedTests/job/ep426I-unit-cen64-gtk3-java20/9/#showFailuresLink ?

jarthana:
> @jarthana Can we get this one reevaluated so testing on Java 20 can be looked further https://ci.eclipse.org/releng/job/AutomatedTests/job/ep426I-unit-cen64-gtk3-java20/9/#showFailuresLink ?

I have an alternate proposal  [here](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/531)

I have run the tests locally with JDK 20 and can confirm tests pass at all levels 19 and below.

akurtakov:
I don't see any logical difference between the two so would you please push one?

akurtakov:
Should be fixed via https://github.com/eclipse-jdt/eclipse.jdt.core/pull/531 instead.

jarthana:
> I don't see any logical difference between the two so would you please push one?

The difference is, with your patch, you will have to adjust your new code once again for Java 21. I thought it was obvious, but then may be I am missing something.

akurtakov:
> > I don't see any logical difference between the two so would you please push one?
> 
> The difference is, with your patch, you will have to adjust your new code once again for Java 21. I thought it was obvious, but then may be I am missing something.

Sure in a place that has been adjusted for every single Java version even without this change - aka different line would have to be copy-pasted :).

jarthana:
> > > I don't see any logical difference between the two so would you please push one?
> > 
> > 
> > The difference is, with your patch, you will have to adjust your new code once again for Java 21. I thought it was obvious, but then may be I am missing something.
> 
> Sure in a place that has been adjusted for every single Java version even without this change - aka different line would have to be copy-pasted :).

Now you are sounding more like a troll. If your request was not running tests with JDK 20 even before us adding support for Java 20, we wouldn't even be talking about this change. Do you want to do this next time also for Java 21?

akurtakov:
Most likely I will have to.

jarthana:
> Most likely I will have to do smth (if not exactly that) for Java 21 at some point too.

My point is, just because we do this exercise once every Java release, doesn't mean we have to do it twice. So, let's not worry about doing it again next time. Whatever required to run tests at Java 20/21/... will be done on our customary beta branches and we should be able to run tests with most recent version of JDK on master without having to do any changes.

akurtakov:
So should I recreate the initial patch with all the constants and etc. for the beta branch to have the ball rolling there? Right now the beta branch even misses the ability to run tests.

iloveeclipse:
> Right now the beta branch even misses the ability to run tests.

I'm not that experienced in the "next Java version support" topic, but wasn't the intention of #531 to fix that?

What we (non compiler developers) all miss here is some "How to update JDT for to next Java release for dummies" wiki.
At least few basic but important steps like bumping constants here and there, updating test configs etc.

@jarthana : do you see a chance to provide some basic information?

I've enabled wiki on JDT core: https://github.com/eclipse-jdt/eclipse.jdt.core/wiki, feel free to add a new page there.
It could be used as starting point to document what we do while updating to Java 20, so all future updates would be easier.

jarthana:
> So should I recreate the initial patch with all the constants and etc. for the beta branch to have the ball rolling there? Right now the beta branch even misses the ability to run tests.

Only if you want to take up contributing to Java 20 work. On the other hand, I am still planning to add the basic infrastructure support soon. No timeline yet, but Java 20 is still far away.

> What we (non compiler developers) all miss here is some "How to update JDT for to next Java release for dummies" wiki. At least few basic but important steps like bumping constants here and there, updating test configs etc.
> 
> @jarthana : do you see a chance to provide some basic information?

Fair enough. As I add stuff for Java 20, I will keep documenting these on the wiki, so it helps others not only to understand what is what and also to be able to contribute without fumbling around.

akurtakov:
> > So should I recreate the initial patch with all the constants and etc. for the beta branch to have the ball rolling there? Right now the beta branch even misses the ability to run tests.
> 
> Only if you want to take up contributing to Java 20 work. On the other hand, I am still planning to add the basic infrastructure support soon. No timeline yet, but Java 20 is still far away.
> 

I thought that intention was implicit by sending a PR :) . Even if I'm not a compiler guy (yet ), the amount of work to get new Java version supports spreads far wider than the compiler but majority of it starts at that level so it looks inevitable to not get my hands dirty. 
Java 20 might be far away for you but it's not that far for someone like me that has to find his path around.

Issue #533, opened 2022-11-12
MetaIndex values for type references are wrong


gayanper:
Based on the https://github.com/eclipse-jdt/eclipse.jdt.core/blob/7eeb9b4342d4d83f3a5a1d078a4e4dff0ab83bb2/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/AbstractIndexer.java#L256

It cause all type references which doesn't have a enclosing type to prefix with the '$' which cause metaindex hits to fail.

Once this is fixed the metaindex needs to rebuilt which means the a full index rebuilt must be done.

iloveeclipse:
I assume this can be closed now.

PR #532, opened 2022-11-11
Use meta index for reference search

gayanper:
## What it does
The meta index which was introduced to speedup subtype search could be used to speed up the reference search as well. When proper qualifier parameters are given with the pattern, the index selection will use that information to narrow down the correct indexes to search for.

This change will only add support for using meta index information for type reference and declaration search for now.

## How to test
- Integration tests which exist for Java Search must pass without any changes needed.
- Person type reference and declaration search in the IDE.

## Author checklist

- [x] I have thoroughly tested my changes
- [x] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [x] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

PR #534, opened 2022-11-12
Fix the meta index values for type declarations #533

gayanper:
## What it does
The type declaration qualifiers are added with a '$' which cause index misses. Since the indexes needs to be rebuilt after applying the fix, the disk index signature is updated.

## How to test
- Integration tests
- Test with #532

## Author checklist


gayanper:
@iloveeclipse please have a look, most importantly about the disk index signature bump to rebuild all indexes.

iloveeclipse:
Sorry Gayan, I'm as usually totally under water with other tasks, so can't guarantee anything here.

gayanper:
I would like to merge this soon so that index regeneration problems can be found early in the release cycle. @iloveeclipse @stephan-herrmann let me know if you see otherwise.

iloveeclipse:
Sure. Earlier is better.

gayanper:
Fixes #533

testforstephen:
@gayanper A small suggestion for PR merge, It might be better to use "**Squash and merge**" to keep the git history clean.

gayanper:
> @gayanper A small suggestion for PR merge, It might be better to use "**Squash and merge**" to keep the git history clean.

My bad 😕, thanks for pointing out @testforstephen. I missed to rebase before the merge. I should default to squash merge to be safe in future. 

May be we should look into configuring squashing as default merge strategy. Cannot remember if we had that discussion when moving to github. But will start a discussion for that.

testforstephen:
> May be we should look into configuring squashing as default merge strategy. Cannot remember if we had that discussion when moving to github. But will start a discussion for that.

Agree, to ensure a linear history, we should only turn on squashing or rebasing merge for PR in GitHub repo settings.

image

PR #541, opened 2022-11-16
Add Java 20 class version and other constants

akurtakov:
Constants for jvm version, bytecode version, tests runtime detection. 
Started with https://github.com/eclipse-jdt/eclipse.jdt.core/issues/529 but continues here as these have to go to a branch.

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/Main.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/Main.java
@@ -3173,6 +3173,9 @@ private String optionStringToVersion(String currentArg) {
 		case "19": //$NON-NLS-1$
 		case "19.0": //$NON-NLS-1$
 			return CompilerOptions.VERSION_19;
+		case "20": //$NON-NLS-1$
+		case "20.0": //$NON-NLS-1$
+			return CompilerOptions.VERSION_20;
 		default:
 			return null;
 	}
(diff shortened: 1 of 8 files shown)


iloveeclipse:
@akurtakov : it would be a good idea to document all changes you do in order to get Java 20 support working on a dedicated issue and link all PRs to that issue.
 This would allow us to perform same for Java 21 with less effort just by following steps.

sravanlakkimsetti:
> Once this patch is in we can start doing branch builds for Java 20. @sravanlakkimsetti would you please outline what should be done ?

Branch builds are done using Y-builds.
1. Update scripts in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/tree/master/cje-production/Y-build. should be very much similar to I-build scripts
2. Update the branch information in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/master/cje-production/streams/repositories_java19.txt also rename this file to repositories_java20.txt. Update this file only when you have a branch. Otherwise keep the repository at master. I normally ask jdt team to update this file when they are ready with the branch
3. Create Y-build test jobs and Y-build job

You can trigger the build once these steps are done. You can actually start Y-builds using master branch till the java20 branches are ready. This will be same as I-build

jarthana:
> Once this patch is in we can start doing branch builds for Java 20. @sravanlakkimsetti would you please outline what should be done ?

@akurtakov I uploaded a wiki page on things that need to be done for new Java versions. I updated the [other issue](https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/164), but for convenience, here it is: 

https://github.com/eclipse-jdt/eclipse.jdt.core/wiki/Adding-support-for-new-Java-version-and-Features-in-JDT

You can ignore the "Going forward" section, but the rest of the changes are better if committed in one batch.

akurtakov:
@jarthana Everything listed should be done now.

jarthana:
I see some tests failing. I suspect that is because the features continuing as preview need to be moved to 20. And I also think the new constant value `F_20 = 0x12000` is wrong. I will take these up in a follow-up commit.

PR #545, opened 2022-11-17
Fix NPE in ClasspathMultiDirectory#directoryList()

jdneo:
Signed-off-by: Sheng Chen <sheche@microsoft.com>

## What it does
Fix a potential NPE issue in `ClasspathMultiDirectory`. the member `this.directoryCache` could be null if `cleanup()` is called

## How to test

1. Use the sample code https://github.com/micronaut-projects/micronaut-examples/tree/master/hello-world-java
2. Enable Maven JDT APT (Window > Preferences > Maven > Annotation Processing > Automatically configure JDT APT)
3. Import as a Maven project
4. Build the project, and run the test case `HelloControllerTest`

#### Behavior without the change
1. Some class files are not generated into `target/test-classes/example`
2. The test case `HelloControllerTest` fails

#### Behavior with the change
1. class files are generated successfully into `target/test-classes/example`
2. The test case `HelloControllerTest` passes

## Author checklist


iloveeclipse:
What is missing is same change in the parent class, and a test.

jdneo:
@fbricon Not yet, I can investigate more tomorrow. But might need some time since I'm not quite familiar with the code base now.

If anyone has some background knowledge/hints that will be helpful :)

jdneo:
@iloveeclipse yeah, I could do that after I get some certain understanding of this problem

stephan-herrmann:
> Fix a potential NPE issue in ClasspathMultiDirectory. the member this.directoryCache could be null if cleanup() is called

Scanning the call hierarchy of `ClasspathDirectory.cleanup()` I see many clients who systematically discard the `NameEnvironment` after cleanup, where the name environment should hold the only reference to the classpath directory. Hence, nobody should be able to see the null in `directoryCache`.

Seams like one of those clients didn't observe that discipline?

jdneo:
I found that the [same NPE issues](https://bugs.eclipse.org/bugs/buglist.cgi?bug_status=__all__&content=directoryCache&no_redirect=1&order=Importance&product=JDT&query_format=specific) can be found in bugzilla as well, but for different use cases.

For this PR, looks like it is caused by multiple instances reference the same `NameEnvironment`. See following screenshots:

When triggering `update maven project`, the code goes to:
https://github.com/eclipse-jdt/eclipse.jdt.core/blob/efccc219eae2f64f507fa3031e1d68c32d644aa2/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/JavaBuilder.java#L271-L272

image

And then, during the build, some typebindings are generated with the same `nameEnvironment`

image

After [imageBuilder.build();](https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/JavaBuilder.java#L273), it starts to clean up
image

At this time, the `directoryCache` is set to null.

Then, the `testImageBuilder` starts to build and the code goes to:

image

And NPE happens.

I'm thinking that, maybe reset `directoryCache` is better than set it to null when cleaning up?

stephan-herrmann:
> To simply fix this issue, maybe we can just add null checks and return null directly if the directoryCache is null.

That sounds fair.

Still I'm a bit worried that perhaps we are leaking internal structures, inviting all kinds of unexpected usage.

@jarthana can you say a word on the life cycle of `IdeBuildProcessingEnvImpl`? It seems to hold on to stuff like a `LookupEnvironment` which may not be suitable for infinite usage by clients. Reminds me of similar problems where clean-up of scopes breaks clients of DOM bindings triggering on-demand resolving. How much out-off-band resolving can methods on `TypeElementImpl` and friends trigger, that might find the `LE` beyond its best-before date?

jarthana:
> @jarthana can you say a word on the life cycle of `IdeBuildProcessingEnvImpl`? It seems to hold on to stuff like a `LookupEnvironment` which may not be suitable for infinite usage by clients. Reminds me of similar problems where clean-up of scopes breaks clients of DOM bindings triggering on-demand resolving. How much out-off-band resolving can methods on `TypeElementImpl` and friends trigger, that might find the `LE` beyond its best-before date?

I will have to check, but a quick search for BaseProcessingEnvImpl.getLookupEnvironment() brings up a lot of results :)

Issue #549, opened 2022-11-18
Increased compilation time with Eclipse 4.12+


Domare90:
Switching to newer Eclipse versions we could identify a noticable increase in compilation time for the same code.
I've attached an example project containing some repeated generic code reproducing the actual constellation that we have. With Eclipse 2019-03 this compiles in around **1 minute**, while with all Eclipse versions beginning with 2020-03 it requires nearly **7 minutes** to compile, effectively leading to an increased compile time of hours in some larger projects. Also tested the latest 2022-09 which shows the same behavior.

In the profiling I can see that basically all of the time is spent in **org.eclipse.jdt.internal.compiler.lookup.CompilationUnitScope.faultInTypes()** drilling down to **org.eclipse.jdt.internal.compiler.lookup.TypeSystem.HashedParameterizedTypes**. Here you can also see a snippet of the profiling when compiling the attached test project with 2019-03 and 2022-06.
**2019-03**
![profiling-2019-03](https://user-images.githubusercontent.com/107324920/202706437-b76cb8da-7d3a-45ca-8ada-b7c7d050cdb9.JPG)
**2022-06**
![profiling-2022-06](https://user-images.githubusercontent.com/107324920/202706429-15a34c77-ea58-4fc6-8931-a9f5bf22f1df.JPG)

Analyzing this further I noticed that the hashCode() implementation of **org.eclipse.jdt.internal.compiler.lookup.TypeSystem.HashedParameterizedTypes.PTBKey** has been changed beginning with 2020-03. In 2019-03 it used all three fields *type*, *enclosingType* and *arguments*, while in the newer versions *enclosingType* has been dropped (while still being contained in the equals() implementation though).

As I don't have an JDT development setup, I've modified the bytecode of the hashCode() implementation in 2022-06 and afterwards the compile time then instantly dropped down to the same level as with 2019-03.

Example project: [compile-perf.zip](https://github.com/eclipse-jdt/eclipse.jdt.core/files/10041321/compile-perf.zip)

iloveeclipse:
> As I don't have an JDT development setup

Can you please try: https://github.com/eclipse-jdt/.github/blob/main/CONTRIBUTING.md#create-an-eclipse-development-environment

Would be great if you could provide a PR with your patch & ideally with a regression test (also OK if that would be a manual one, since performance tests tend to be unstable).

[...]

PR #551, opened 2022-11-21
Use hash of enclosingType if it is a ParameterizedTypeBinding

brychcy:
Fixes "Increased compilation time with Eclipse 4.12+" #549

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeSystem.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeSystem.java
@@ -98,6 +98,10 @@ public class TypeSystem {
 					ReferenceBinding enclosing = resolvedType.enclosingType();
 					if (enclosing != null) {
 						this.enclosingType = resolvedType.isStatic() ? enclosing : (ReferenceBinding) env.convertUnresolvedBinaryToRawType(enclosing); // needed when binding unresolved member type
+						if (this.enclosingType.getClass() == ParameterizedTypeBinding.class) {
+							throw new IllegalStateException("unexpected: resolved enclosing type of " //$NON-NLS-1$
+									+ new String(this.type.readableName(false)) + " is a ParameterizedTypeBinding"); //$NON-NLS-1$
+						}
 					}
 				}
 				if (this.arguments != null) {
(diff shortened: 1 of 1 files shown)


org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/TypeSystem.java
```diff
@@ -98,6 +98,9 @@ public void swapUnresolved(UnresolvedReferenceBinding unresolvedType, ReferenceB
 					ReferenceBinding enclosing = resolvedType.enclosingType();
 					if (enclosing != null) {
 						this.enclosingType = resolvedType.isStatic() ? enclosing : (ReferenceBinding) env.convertUnresolvedBinaryToRawType(enclosing); // needed when binding unresolved member type
+						if(this.enclosingType.getClass() == ParameterizedTypeBinding.class) {
```
iloveeclipse, line 101: May I ask for a "Ctrl + F" on selection? The code change doesn't look aligned with other formatting.

iloveeclipse (commented):
Till, many thanks for the patch. Do you think a regression test is possible?

org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/TypeSystem.java
```diff
@@ -123,6 +126,11 @@ final int hash(TypeBinding b) {
 			public int hashCode() {
 				final int prime=31;
 				int hashCode = 1 + hash(this.type);
+				if (this.enclosingType != null && this.enclosingType.getClass() == ParameterizedTypeBinding.class) {
```
eric-milles, line 129: Would "this.enclosingType != null && this.enclosingType.isParameterizedType()" cover it?  Or do you also need this for `RawTypeBinding`?

org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/TypeSystem.java
brychcy: done.

org.eclipse.jdt.core/compiler/org/eclipse/jdt/internal/compiler/lookup/TypeSystem.java
brychcy: So the base issue is that RawTypeBinding extends PTB and  raw instances can appear during URB resolution, so we cannot use "instanceof PTB".

And you're not wrong: isParameterizedType() would actually work here, but you have to hunt down all overloads to see this: TypeBinding defines it to returns false; it is overloaded to return true in PTB and then to return false again in RTB - and there are no other subclasses of PTB. 

Now here "exactly for PTB instances" is really meant so I think it is better to express this as direct as possible.

eric-milles: Right, but isn't ".getClass() == ParameterizedTypeBinding.class" much more expensive than ".isParameterizedType()"?

brychcy: No, at least hotspot optimizes this, see https://github.com/openjdk/jdk/blob/ccc6e169188019dab008df5ffb11b4279e14f8fc/src/hotspot/share/opto/subnode.cpp#L1068

PR #564, opened 2022-11-28
Distinguish shared indexes with different index versions

testforstephen:
Signed-off-by: Jinbo Wang <jinbwan@microsoft.com>

This is an enhancement to previous commit "[shared JDT Index folder among different workspaces](https://git.eclipse.org/r/c/jdt/eclipse.jdt.core/+/152349/)". The problem with shared indexes is that the index schema can evolve over time, and reusing outdated indexes from shared folders does not work. Currently, unless you read the original contents of the *.index file, it is not clear what index version is being used in the shared folder. To make it more intuitive, one suggestion is to force the shared folder structure to be organized as `<base_path>/<index_version>/*.index`, then you have to enable the system property in the format like `-Djdt.core.sharedIndexLocation=<base_path>`. 

This is also similar to a shared m2 dir, and allows multiple versions of indexes to coexist.

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaIndexTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaIndexTests.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2012, 2019 IBM Corporation and others.
+ * Copyright (c) 2012, 2022 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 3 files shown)


rgrunber:
I guess my only concern is about maintaining the older index files. Is it that expensive to simply regenerate the index files every time the version is updated ? The last index update was done for Photon (4.8) so it happens pretty rarely. Or is the goal to prevent the shared index folder from growing too large ?  If so, maybe clients should institute some kind of mechanism to remove old entries ?

Looking at updates to `SIGNATURE` :

1.130 -> 1.13.1 at bf6a4008daf8d98e0cec7c91279c54eace8d943b (Feb 2018)
1.129 -> 1.130 at a9993d1acf18d20bcdbeca323fb66ee7d9123e81 (Jan 2017)
1.128 -> 1.129 at 1b4ae3c306349bc12b5bd14c0b0208197346f4fe (Dec 2015)

rgrunber (approved):
The change works for me. As JDT-LS was the main motivation for this feature, I think we should be fine to break the previous behaviour (all lookups at the root of the folder). Future users of this feature will just have to be aware of the index version in order to get the lookup to work correctly.

PR #572, opened 2022-12-01
Removed dependencies to removed JDT fragments

iloveeclipse:
- org.eclipse.jdt.compiler.apt
 - org.eclipse.jdt.compiler.tool

See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/181


iloveeclipse:
Any help fixing apt tests would be appreciated...

akurtakov:
My guess is that the the new bundle doesn't register javax.tools.JavaCompiler properly.

akurtakov:
https://github.com/eclipse-jdt/eclipse.jdt.core/blob/457e2221c44b49b6f38fcdfe6d1d59467a04bf75/org.eclipse.jdt.compiler.apt.tests/META-INF/MANIFEST.MF#L21 is optional but this is where the service class lives so it doesn't look optional to me. What do you think about trying to change it?

iloveeclipse:
> What do you think about trying to change it?

Let see if the next attempt fixes tests. The service definition was for whatever reason **manually** copied to the build jar in one of the ant scripts. I guess because it was in a fragment. OMFG.

iloveeclipse:
> My guess is that the the new bundle doesn't register javax.tools.JavaCompiler properly.
> Edit: I am wrong.

That **was** the case. Why do you think you were wrong?

akurtakov:
> That **was** the case. Why do you think you were wrong?
```
Iterator<JavaCompiler> iter =ServiceLoader.load(JavaCompiler.class).iterator();
		while(iter.hasNext()) {
			JavaCompiler compiler = iter.next();
			System.out.println(compiler);
		}
```
with the jar produced by maven for the new bundle gave me both javac and ecj.

PR #573, opened 2022-12-01
Test fix for apt.tests

akurtakov:


iloveeclipse:
I think that isn't the fix, this one looks better: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/572/commits/54e4616b81469b3d02b8bf04b2a5560f132d8511

Issue #220, opened 2022-07-14
jdt.internal.compiler.apt.util copies jdt.internal.compiler.tool classes and they already derived


iloveeclipse:
I've stumbled over lot of duplicated and **slightly diverged** code in `org.eclipse.jdt.internal.compiler.apt.util` and `org.eclipse.jdt.internal.compiler.tool` packages.

I believe the divergence is mostly because of the bug fixes done in one package because no one expects the same code to be somewhere else.

Note: the packages are inside of two different **fragments** of `org.eclipse.jdt.core` bundle.

I believe the right solution would be to put the common classes into `org.eclipse.jdt.core` bundle and there into `org.eclipse.jdt.internal.compiler.batch.util` and use them from fragments.

iloveeclipse:
#222 was just a small fix for divergent implementation details, the implementations itself should ideally be fully unified and copies merged to one code.

iloveeclipse:
Consider this to be fixed with https://github.com/eclipse-jdt/eclipse.jdt.core/pull/574

PR #574, opened 2022-12-02
Removed 99% of duplicated apt/util code (#220)

iloveeclipse:
Remaining code is in
org.eclipse.jdt.internal.compiler.apt.util.EclipseFileManager and has slightly different behavior compared with the
org.eclipse.jdt.internal.compiler.tool.EclipseFileManager.

Both implementations were committed in same commit [1], so it seems that differences could be significant, but unfortunately it wasn't documented why.

[1] https://github.com/eclipse-jdt/eclipse.jdt.core/commit/c798229d351d6daa2232cff6c15750336659223e

See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/220


iloveeclipse:
@jarthana : since you've added both (slightly different) EclipseFileManager classes in same commit https://github.com/eclipse-jdt/eclipse.jdt.core/commit/c798229d351d6daa2232cff6c15750336659223e, could you please take a look at the code and give a hint, why the implementation differs in the way it does?

Note: both "original" org.eclipse.jdt.internal.compiler.apt.util.EclipseFileManager version and the current version that actually 1:1 copy of org.eclipse.jdt.internal.compiler.tool.EclipseFileManager pass through the tests.

But the org.eclipse.jdt.internal.compiler.apt.util.EclipseFileManager version with the four **un**commented out methods (that contain original code that differs from org.eclipse.jdt.internal.compiler.tool.EclipseFileManager) and with the rest left to org.eclipse.jdt.internal.compiler.tool.EclipseFileManager implementation does not work???

jarthana:
Sorry, missed this. Let me take a look at this first things tomorrow.

jarthana:
Sorry, I can't recollect what was the idea behind the difference even after going through the changes again. 
I don't recall ever thinking that these two files need to be different. So, it could be an oversight from my side.

And if the test pass when they being identical, then we should take it :)
I also vaguely remember cooking up ModuleLocationWrapper to make things work with the then evolving module system. And if the compiler.tool version is using it in more places, then I would imagine that being a better code.

PR #580, opened 2022-12-02
Remove generated jdtCompilerAdapter.jar from sources

iloveeclipse:
jdtCompilerAdapter.jar is generated by maven on build, why it is inside source tree ???

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchScopeTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchScopeTests.java
@@ -18,15 +18,19 @@ import java.net.URL;
 import java.util.Arrays;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Queue;
 import java.util.Set;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import junit.framework.Test;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IWorkspaceRunnable;
 import org.eclipse.core.resources.IncrementalProjectBuilder;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.ILogListener;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.Path;
+import org.eclipse.core.runtime.Platform;
 import org.eclipse.jdt.core.IClasspathContainer;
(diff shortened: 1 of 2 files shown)


iloveeclipse:
As we see (tests are green), the jar is not needed for maven build & tests, because it is generated & packaged into the jdt.core jar (!) anyway.
I believe the only reason to have this jar is the code in org.eclipse.ant.internal.ui.datatransfer.BuildFileCreator that, if running Eclipse from Eclipse, will fail to find & link the jar.

@SarikaSinha , @vik-chand : is my assumption correct and the jar **in the source tree** is only needed for debugging in the IDE?
If so, I would propose to fix BuildFileCreator & remove this jar from sources.

iloveeclipse:
OK, another really weird reason is [this one](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/592#issuecomment-1339681717): bundles that use this jar at **compile time** because it can be ["added" to PDE classpath](https://github.com/groovy/groovy-eclipse/blob/bacc5e539a6461e9a3c44ad17886adab97cca847/ide/org.codehaus.groovy.eclipse.ant/build.properties#L14) via 

`jars.extra.classpath=platform:/plugin/org.eclipse.jdt.core/jdtCompilerAdapter.jar`

Not sure if that would work if jdt bundle is in the workspace **without** the jar.

Ideally in both cases that should be taken not from the jar that is checked into the source tree but from jar that is in the platform.

iloveeclipse:
I think most practical solution would be to remove the jar from git and instead create this jar with external builder on every change from batch compiler code or by copying from the platform.

eric-milles:
`org.codehaus.groovy.eclipse.ant` depends on `platform:/plugin/org.eclipse.jdt.core/jdtCompilerAdapter.jar` (for `JDTCompilerAdapter`) much like `org.eclipse.jdt.core` depends on `platform:/plugin/org.apache.ant/lib/ant.jar` for basic Ant types.

In fact, noticing this, I was able to reduce my bundle dependencies by replacing `Require-Bundle: org.apache.ant,org.eclipse.jdt.core` (in MANIFEST.MF) with `jars.extra.classpath=platform:/plugin/org.apache.ant/lib/ant.jar,platform:/plugin/org.eclipse.jdt.core/jdtCompilerAdapter.jar` (in build.properties).

At this moment, both `org.eclipse.jdt.core` and `org.codehaus.groovy.eclipse.ant` projects do not create `jdtCompilerAdapter.jar` and `groovyAntEclipse.jar` respectively when built within PDE.  When launching Eclipse IDE from PDE, the Ant extensions are not present if these jars do not exist.  I think this is a reason JDT Core carries this jar in git.  But, as you describe it could be generated and this gap would be closed.

Issue #583, opened 2022-12-04
Add support for array constructor reference


gayanper:
When completing a code like this

```
Arrays.asList("1").stream().toArray(String[]::)
```

I would expect `new` keywords as a completion entry to choose from.

PR #584, opened 2022-12-04
Add support for array constructor reference :#583

gayanper:
## What it does
This change will provide new completion entry for method references on a array type reference.

## How to test
use the following code snippet `Arrays.asList("1").stream().toArray(String[]::)`

## Author checklist


gayanper:
Since this is a small change, I would like to merge it if @jarthana and @stephan-herrmann doesn't see any thing that should be done differently.

gayanper:
@testforstephen would you be able to have a look at this until others provide their feedbacks

testforstephen:
@gayanper It makes sense to auto suggest array constructor for method reference expression. What do you think if we also support autocompleting all Object methods?

image

gayanper:
> @gayanper It makes sense to auto suggest array constructor for method reference expression. What do you think if we also support autocompleting all Object methods?
> 
> image

Do you see any value with such suggestion as a method reference ?

testforstephen:
> Do you see any value with such suggestion as a method reference ?

I see other IDE like IJ provides a full list of methods there. Yes, the array constructor is the most frequently used for me as well. If the ROI is not worth it, it's good to me to only suggest `new` method.

gayanper:
> > Do you see any value with such suggestion as a method reference ?
> 
> I see other IDE like IJ provides a full list of methods there. Yes, the array constructor is the most frequently used for me as well. If the ROI is not worth it, it's good to me to only suggest `new` method.

If you don't see any issues in the implementation I would like to merge this.

iloveeclipse:
@gayanper : could sou please squash commits to ome and avoid merge commits?

testforstephen (approved):
LGTM

Issue #613, opened 2022-12-25
Support dynamic enable/disable of classpath entries


laeubi:
Currently it is not possible to have JDT  dynamically enable a classpath entry based on some system state, this might include:

- The os / ws / arch of the running system (e.g because some code is targeted for a special system, see https://github.com/eclipse-tycho/tycho/issues/632)
- The project JVM (e.g. because some code is targeted only for a certain java version see https://github.com/eclipse-jdt/eclipse.jdt.core/issues/265)

My proposal therefore would be to have a way marking a classpath entry as disabled, and then related code can handle it as if it is not defined at all (e.g. ignore it during compilation / indexing / ...).

This concept also is generic enough to apply it to any classpath entry type, even though most likely only a few would make use of it (as a first start, only CPE_SOURCE)

laeubi:
After thinking more about it I'm probably on the wrong track here regarding my initial idea of enable/disable to support:
- https://github.com/eclipse-tycho/tycho/issues/632

and the parts for
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/265

are better solved as you described here:
- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/614#issuecomment-1382746804
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/265#issuecomment-1383274075

so I'll close this for now and open another proposal in favor of this for the multi-platform compile.

_Originally posted by @laeubi in https://github.com/eclipse-jdt/eclipse.jdt.core/issues/614#issuecomment-1383519460_

Issue #265, opened 2022-07-27
[JEP 238] Multirelease support


laeubi:
[JEP 238: Multi-Release JAR Files](https://openjdk.org/jeps/238) describes the structure of Multirelease jars and when I add such a jar to my classpath it shows the corresponding selected classes so this looks fine.

Nerveless I have not found yet a way to produce such jar with JDT and [maven indicates](https://maven.apache.org/plugins/maven-compiler-plugin/multirelease.html)

> Up until the moment of writing all IDEs can only have one JDK per Maven Project, whereas with multi release you want to specify it per source folder.

We recently have [a similar request at m2e](https://github.com/eclipse-m2e/m2e-core/issues/842#issuecomment-1178704554) where it would be useful to specify different target release.

Because of this I'd like to propose preliminary support for JEP 238 in JDT in the following way:

1. The project VM must be the highest of all release settings
2. on a `classpathentry` entry, allow beside a different output folder (what would be required in such a case), to specify the release like this `<classpathentry kind="src" output="bin17" path="src17" release="17">`
3. when compiling code for such a source folder, use the appropriate release setting for the given source files

The advantage of this would be, that there is no need to duplicate all settings (so I'm not aiming for different compiler settings for each folder) and map 1:1 for a setup you could use in maven with different executions defined for each release used and thus m2e could also leverage this later on.

stephan-herrmann:
Let's briefly check whether or not there is a connection to #614.

I see a bit of contention between your statements:

> 2. on a classpathentry entry, allow [...] to specify the release
> 3. when compiling code for such a source folder, use the appropriate release setting for the given source files

vs. 

> I'm not aiming for different compiler settings for each folder

`--release` *is* of course a compiler setting. The question here seems to be, can we use different `--release` settings within the same build? I strongly doubt that's feasible. All of ecj's `LookupEnvironment` would need to provide different (release dependent) versions of each class of the JRE. If you ask me, that sounds like a huge refactoring (like adding one parameter to "50%" of methods of ecj). Not sure this is manageable.

So, if we have different source folders with different requirements, wouldn't that bring us back to multiple builds even in the IDE, where each build uses a specific classpath filter a la #614 ? To me this sounds like a situation, where calls to the `isEnabled()` test _must_ provide a token to identify the build target. Am I missing anything?

laeubi:
I think I falsely though #614 can be used here but thinking a bit further this won't work out well... so this is probably not related.

[...]

PR #614, opened 2022-12-25
Support dynamic enable/disable of classpath entries

laeubi:
This is WIP for:
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/613

This outlines the idea described in https://github.com/eclipse-jdt/eclipse.jdt.core/issues/613 for adding a way to mark classpath entries as disabled and add an implementation based on platform filtering.

One can then have a classpath entry like this:
```
<classpathentry kind="src">
    <attributes>
	    <attribute name="filter" value="(platform.os=win32)"/>
    </attributes>
</classpathentry>
```

What already works:

- [x] When adding such an attribute manually in the classpath file, the entry is disabled when not running under windows.

What is missing:
- [ ] There is no UI yet to set a filter
- [ ] I could not yet find out how JDT maps `CPE_SOURCE` entries to the actual things to compile, actually I though about this could be some special king of `excluding` files (that simply excludes everything) so hints would be appreciated.

Beside that I'd like to encourage everyone to participate in discussing and sharing ideas about this proposal, but please focus on the enable/disable part (the filtering might or might not be implemented in jdt directly and is just here to have a working example who such thing could be used!

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/core/IClasspathEntry.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/core/IClasspathEntry.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2019 IBM Corporation and others.
+ * Copyright (c) 2000, 2022 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -10,6 +10,7 @@
  *
  * Contributors:
  *     IBM Corporation - initial API and implementation
+ *     Christoph Läubrich - Support dynamic enable/disable of classpath entries
  *******************************************************************************/
 package org.eclipse.jdt.core;
 
@@ -529,4 +530,12 @@ public interface IClasspathEntry {
 	 * @since 3.30
 	 */
 	IPath getExternalAnnotationPath(IProject project, boolean resolve);
+
+	/**
+	 * Check if this entry is disabled, for example ebcause it does not match the current environment.
+	 *
+	 * @return <code>true</code> if this entry is disabled <code>false</code> otherwhise.
+	 * @since 3.34
+	 */
+	boolean isDisabled();
 }
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
@@ -185,6 +185,7 @@ import org.eclipse.jdt.internal.core.JavaModel;
 import org.eclipse.jdt.internal.core.JavaModelManager;
 import org.eclipse.jdt.internal.core.JavaProject;
 import org.eclipse.jdt.internal.core.PackageFragmentRoot;
+import org.eclipse.jdt.internal.core.PlatformPropertiesUtils;
 import org.eclipse.jdt.internal.core.Region;
 import org.eclipse.jdt.internal.core.SetContainerOperation;
 import org.eclipse.jdt.internal.core.SetVariablesOperation;
@@ -197,6 +198,9 @@ import org.eclipse.jdt.internal.core.util.Messages;
 import org.eclipse.jdt.internal.core.util.ModuleUtil;
 import org.eclipse.jdt.internal.core.util.Util;
 import org.osgi.framework.BundleContext;
+import org.osgi.framework.Filter;
+import org.osgi.framework.FrameworkUtil;
+import org.osgi.framework.InvalidSyntaxException;
 
 /**
  * The plug-in runtime class for the Java model plug-in containing the core
@@ -219,6 +223,10 @@ import org.osgi.framework.BundleContext;
 public final class JavaCore extends Plugin {
 
 	private static final IResource[] NO_GENERATED_RESOURCES = new IResource[0];
+
+	private final static String PLATFORM_WS = "platform.ws"; //$NON-NLS-1$
+	private final static String PLATFORM_OS = "platform.os"; //$NON-NLS-1$
+	private final static String PLATFORM_ARCH = "platform.arch"; //$NON-NLS-1$
 
 	private static Plugin JAVA_CORE_PLUGIN = null;
 	/**
@@ -5585,7 +5593,8 @@ public final class JavaCore extends Plugin {
 		if (extraAttributes == null) {
 			extraAttributes = ClasspathEntry.NO_EXTRA_ATTRIBUTES;
 		}
-		return new ClasspathEntry(
+
+		ClasspathEntry entry = new ClasspathEntry(
 			IPackageFragmentRoot.K_SOURCE,
 			IClasspathEntry.CPE_SOURCE,
 			path,
@@ -5598,6 +5607,22 @@ public final class JavaCore extends Plugin {
 			null,
 			false, // no access rules to combine
 			extraAttributes);
+		for (IClasspathAttribute attribute : extraAttributes) {
+			String name = attribute.getName();
+			if (name.equals("filter")) { //$NON-NLS-1$
+				try {
+					Filter filter = FrameworkUtil.createFilter(attribute.getValue());
+					String os = PlatformPropertiesUtils.getOS(System.getProperties());
+					String ws = PlatformPropertiesUtils.getWS(System.getProperties());
+					String arch = PlatformPropertiesUtils.getArch(System.getProperties());
+					boolean disabled = !filter.matches(Map.of(PLATFORM_OS, os, PLATFORM_WS, ws, PLATFORM_ARCH, arch));
+					entry.setDisabled(disabled);
+				} catch (InvalidSyntaxException e) {
+					getJavaCore().getLog().error("Can't evaluate filter", e); //$NON-NLS-1$
+				}
+			}
+		}
+		return entry;
 	}
 
 	/**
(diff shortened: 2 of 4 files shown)


laeubi:
This shows how this could be used in `jdt.debug` to skip disabled items for system libs:
- https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/169

laeubi:
@stephan-herrmann @iloveeclipse can you review or suggest someone?

stephan-herrmann:
> @stephan-herrmann @iloveeclipse can you review or suggest someone?

From a quick look, the PR only introduces a new property, but I could not find any code using that property. So who would be calling `org.eclipse.jdt.core.IClasspathEntry.isDisabled()` with which effect?

Also I'm not quite sure _what kind of dynamism_ "dynamic enable/disable..." should refer to. Wouldn't it suffice to implement "conditional computation of classpaths", i.e., once computed there is no further dynamism?

As for the split of computing vs applying a filter, I suggest to think of an analogy to **classpath containers**: it's a concept defined by JDT but implemented by others (like PDE, m2e). JDT doesn't claim to understand how a container is populated, but it knows that extensions can do so, to the end that JDT can then use the result for its _resolved classpath_. So far **source folders** didn't have any such need, but when JDT would define a **classpath filter extension point**, then any extender can mark arbitrary cp entries as invisible (this could also apply to context specific dependencies). 

In eclipse-tycho/tycho/issues/632 (which I only skimmed) I also saw mentioning of the "test" flag for comparison, which is special in the regard that JDT already performs two separate builds for a project using this flag: one build for main sources, one build for test sources with visibility of main sources plus test dependencies. If we were to generalize the platform thing with the main/test distinction, then the new extension point used for filtering might need to be more complex, like:
* listing all relevant build categories  (one extender might define "main" vs. "test", another would contribute "linux" vs. "win32", etc.)
* compute visibility of any classpath entry for any given set of categories

Then JDT could do smth like

```
List<Set<Category>> configurations = getAndCombineCategoriesFromExtensions(); // full cartesian product??
for (Set<Category> configuration : configurations) {
     resolvedClasspath = applyExtensionFilters(resolvedClasspath, configuration);
     build (resolvedClasspath);
}
```

That would cover the case of full builds, whereas incremental builds and change notification would require additional care.

I should note, that I don't propose to change the story of "test" vs "main" just yet, only using the example to steer us towards a maximally generic solution (if this can be done without unwanted complexity).

stephan-herrmann:
I looked at https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/169 only after commenting, so now I wonder, if the request is only relevant for the runtime classpath? Or compile time classpath? Or both? Should compile time apply different filtering than launching?

laeubi:
> From a quick look, the PR only introduces a new property, but I could not find any code using that property. So who would be calling org.eclipse.jdt.core.IClasspathEntry.isDisabled() with which effect?

I added a call here but it is sadly a different repository:
- https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/169

also see this discussion:
- https://github.com/eclipse-jdt/eclipse.jdt.core/discussions/615

> Also I'm not quite sure what kind of dynamism "dynamic enable/disable..." should refer to. Wouldn't it suffice to implement "conditional computation of classpaths", i.e., once computed there is no further dynamism?

Dynamic means, the state is computed (once) by some code running in the IDE an not persisted e.g. in the `.classpath` file (but the content of that file might influence the computation).

> So far source folders didn't have any such need, but when JDT would define a classpath filter extension point, then any extender can mark arbitrary cp entries as invisible (this could also apply to context specific dependencies).

Yes something might work as well thats why it is a draft at them moment, so probably an extension point would be good that is then implemented by jdt or others and finally results in a boolean true/false in the end (here I have coded it with just one example).

> I should note, that I don't propose to change the story of "test" vs "main" just yet, only using the example to steer us towards a maximally generic solution (if this can be done without unwanted complexity).

This sounds very interesting, especially in the context of
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/265

where one wants each source-folder to compile with a different `target-level` so if you have a better, more general idea how to cover both cases that's great, I just wanted to came up with a minimal starting point so thing start to get some spin  but I'm open to other approaches as long as they are approaching sometime :-)

> I looked at https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/169 only after commenting, so now I wonder, if the request is only relevant for the runtime classpath? Or compile time classpath? Or both? Should compile time apply different filtering than launching?

For me, this would be pure "compile-time" even though of course runtime is affected if one compiles some source and others are ignored, but I think this would work without special consideration.

stephan-herrmann:
So if you think, an extension point might be the right thing, then let's discuss the desired protocol.

In the end, JDT needs to ask extensions, which of a project's classpath entries should be enabled, so likely there will be a method taking an `IClasspathEntry` and returning `boolean`.

Then, which information will the extension use, to compute its answer? I could imagine the following sources:

1. the actual environment of the running Eclipse instance
2. an extension announces several configurations, and JDT builds for each configuration in turn (so the configuration to be built should be a parameter of the method)
3. I could also think that preferences be used to select configurations
  (a) of course the extension can always use its own preference, with no pre-planning in the extension point
  (b) also JDT could use a preference to select from a given set of configurations which one(s) should be built

For (2), where configurations must be communicated between JDT and extensions, we need to decide how a "configuration" should be represented. Is a String sufficient (who ensures uniqueness)? Or should we introduce a type `IBuildConfiguration`? To JDT, this type could be mostly opaque, just some `getName()` would be useful, should we ever have to show a configuration to the user (e.g., in progress bar "Building configuration FOO", or in preferences for en/disabling available configurations etc.). Other than that the meaning of each `IBuildConfiguration` would be known only to the extension (and JDT should probably remember which extension is responsible for filtering which configuration...).

Ah, so this brings us to the point, where potentially several extensions are involved, so how to handle combinations of configurations? That's where in my earlier comment I asked about the full cartesian product (one extension filters by Java version, another extension is interested in different windowing systems ...).

If the extension point controls the actual content of the resolved classpath, then several aspects of the IDE will automatically benefit from it: compilation (possible several builds), runtime (needs to select one config). OTOH, tooling features like search and all that depends on it may be at conflict with this multiverse of configurations (perhaps that's where preferences will help?), so will we need to support both views: unfiltered and filtered. That would get a bit tricky.

If the multitude of configurations gets overwhelming for tooling (and performance?), then perhaps the preferences approach is more suitable: we'd select one configuration at a time, and use it for all IDE operations from build, over refactor, to launch. We'd leave it to build tools to invoke things for several configurations in turn.

Or: a source folder could statically declare the configuration it is intended for, and then the compiler will check that it only ever uses dependencies of the same configuration.

This discussion also reminds me of how b3aggr files handle configurations (as triplets, right?), defining a list of concrete configurations for which a given aggregation should be validated.

While an extension point can provide a very powerful addition, it needs to be designed carefully, because once published the options to change it are quite limited, so we better do it right the first time.

laeubi:
I think such an extension just get the `IClasspathEntry` as the rest it implementation specific, if there are multiple ones, they should be and'ed together so if one think it should be disabled it must be disabled for all.

So something like:

```
interface ClasspathFilter {

     boolean isEnabled(IJavaProject project, IClasspathEntry entry);
  
}
```

I also won't assume that such a extension does performance critical operations, most of the time it would be:

1. Check if the type of classpath entry is something that's of interest for the extension (e.g. is a a source type and has it an attribute `release`, or has it an attribute `filter`)
2. Compare this with the system state (e.g. what is the projects JRE or does some system properties match)

and in most of the cases the check will end at step 1 because most of the time filtering is not required.

stephan-herrmann:
Thanks for mentioning the `IJavaProject` as I tend to forget that `IClasspathEntry` contains no back pointer.

> I think such an extension just get the `IClasspathEntry` as the rest it implementation specific

This would certainly work for the case of filtering by the current platform.

This would not work for a use case like the main/test distinction, where multiple builds are performed, and thus the extension needs to be asked relative to a specific variant (enabled for main? enabled for test?). Perhaps this argument is irrelevant, as we already have a solution for main/test, but what about multi-release support? Doesn't this need multiple builds for the same project, too?

It would be sad if we create an extension point for today's requirement, and tomorrow we find, that it lacks just one parameter to also satisfy another relevant requirement.

BTW, multiple builds is also the reason I mentioned performance, not the time spent testing enablement :)

laeubi:
I think multiple builds should be handled differently because here I want that some item are just **ignored** (e.g. I want them to be handled like I already can choose *exclude from classpath*) and these are specific to **one** project and the other case is that I want some items to be handled **differently** (e.g. compiled with different settings), where I agree that those "different settings" are global and apply to **multiple** projects.

Combining both approaches seem to complicate things a lot and as you mentioned the test/main thing is already there, so one probably want to generalize that for other use-cases but at the moment it is there an likley needs other consideration if generalized.

laeubi:
I think multiple builds should be handled differently because here I want that some item are just ignored (e.g. I want them to be handled like I already can choose exclude from classpath)

After thinking more about it I'm probably on the wrong track here regarding my initial idea of enable/disable to support:
- https://github.com/eclipse-tycho/tycho/issues/632

and the parts for
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/265

are better solved as you described here:
- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/614#issuecomment-1382746804
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/265#issuecomment-1383274075

so I'll close this for now and open another proposal in favor of this for the multi-platform compile.

HannesWell:
The concept proposed https://github.com/eclipse-jdt/eclipse.jdt.core/issues/613 would for example be useful for the org.eclipse.swt bundle development, where we currently have three different `.classpath_${os}` files, one for each OS that reference different src-folders.
When provisioning a dev workspace for SWT the right one has to be renamed to `.classpath`.
If entries could be enabled/disabled automatically by a property this would be much more convenient.

PR #618, opened 2022-12-26
Unnecessary error marker for record constructor that uses varargs

### LINKED ISSUE #342: Unnecessary error marker for record constructor that uses varargs
From https://github.com/redhat-developer/vscode-java/issues/2640

Version: 2022-09 (4.25)
Build id: I20220817-1800

Given the following files, I see `The constructor Sneaker(String, int, int, int, int) is undefined` error marker on the `Sneaker` constructor invocation in `Main` but not in the `Sneaker` class itself. Furthermore, the error marker does not appear in the problems view, nor does it generate an error annotation on the file/folder(s)/projects in the package explorer, or even on the editor part. I've included a screenshot to show that. The code seems to compile and run just fine.

**Sneaker.java**
```
package org.running;
public record Sneaker(String brand, float price, int ... sizes) {
	public void test () {
		Sneaker sn = new Sneaker("Eclipse", 100, 9, 10, 11);
	}
}
```

**Main.java**
```
package org.running;
public class Main {
		Sneaker sn = new Sneaker("Eclipse", 100, 1, 2, 3);
	}
}
```

record-constructor

snjeza:
Fixes #342 

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ReconcilerTests16.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ReconcilerTests16.java
@@ -24,6 +24,7 @@ import org.eclipse.jdt.core.JavaCore;
 import org.eclipse.jdt.core.JavaModelException;
 import org.eclipse.jdt.core.WorkingCopyOwner;
 import org.eclipse.jdt.core.dom.AST;
+
 import junit.framework.Test;
 
 public class ReconcilerTests16 extends ModifyingResourceTests {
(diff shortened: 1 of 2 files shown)


rgrunber:
The change seems to be working, but I'm trying to understand what's going wrong that requires this. I'm guessing this stems from the fact that the vararg is represented as an array, causing a failure to properly detect it later on.

At https://github.com/eclipse-jdt/eclipse.jdt.core/blob/7f2b6b230ae5f09345d2f0ebbddae00d9a72b9fe/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/AllocationExpression.java#L491 , a ProblemMethodBinding is returned when a valid constuctor binding can't be found. How does performing a `dietParse` avoid this.

snjeza:
> he change seems to be working, but I'm trying to understand what's going wrong that requires this. I'm guessing this stems from the fact that the vararg is represented as an array, causing a failure to properly detect it later on.

You are right.

>  How does performing a dietParse avoid this.

dietParse (or parse) detects varargs properly. Because of that the error marker does not appear in the problems view,

rgrunber (approved):
I think this would be fine to merge after the freeze. The parsing seems to ensure varargs is set to true in the method binding which is later used to find the compatibility between one representation having 3 arguments (String, int, int[]) and the other having 5 (String, int, int, int, int).

**Update:** CC'd @jjohnstn , @gayanper for additional review for RC1. The patch seems safe (just do the parse in a special case, varargs).

jjohnstn (approved):
I agree this should be safe for RC1.  Merging.

PR #621, opened 2022-12-30
ECJ compiler errors

### LINKED ISSUE #612: [compiler] ECJ compiler errors
Steps to reproduce:

- create the following class

```
public class Inheritance {
        Sub sub = new Sub();
        sub.method1();
    }

    static class Outer<T> {
        class Inner {
            void sayHi() {
                System.out.println("hello world!");
            }
        }
    }

    static class Sub extends Outer<Sub.Inner> {
        void method1() {
            Inner inner = new Inner();
            inner.sayHi();
        }
    }
}
```
- Eclipse generates
```

Inner cannot be resolved to a type	line 17	Java Problem
Inner cannot be resolved to a type	line 17	Java Problem
Sub.Inner cannot be resolved to a type	line 15	Java Problem

```
It works in javac
```
$ javac -version
javac 17.0.5
$ javac Inheritance.java 
$ java Inheritance 
hello world!
$ 
```
A related issue: https://github.com/redhat-developer/vscode-java/issues/2843

snjeza:
Fixes #612 

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
@@ -3027,6 +3027,25 @@ public abstract class Scope {
 	*/
 	public final ReferenceBinding getMemberType(char[] typeName, ReferenceBinding enclosingType) {
 		ReferenceBinding memberType = findMemberType(typeName, enclosingType);
+		if (enclosingType.isHierarchyBeingConnected() && memberType == null && this.kind == CLASS_SCOPE) {
+			ClassScope scope = (ClassScope) this;
+			TypeReference superTypeReference = scope.referenceContext.superclass;
+			if (superTypeReference != null && superTypeReference.resolvedType instanceof ReferenceBinding) {
+				memberType = findMemberType(typeName, (ReferenceBinding) superTypeReference.resolvedType);
+			}
+			if (memberType == null && scope.referenceContext.superInterfaces != null
+					&& scope.referenceContext.superInterfaces.length > 0) {
+				TypeReference[] interfaces = scope.referenceContext.superInterfaces;
+				for (TypeReference reference : interfaces) {
+					if (reference != null && reference.resolvedType instanceof ReferenceBinding) {
+						memberType = findMemberType(typeName, (ReferenceBinding) reference.resolvedType);
+						if (memberType != null) {
+							break;
+						}
+					}
+				}
+			}
+		}
 		if (memberType != null) return memberType;
 		char[][] compoundName = new char[][] { typeName };
 		return new ProblemReferenceBinding(compoundName, null, ProblemReasons.NotFound);
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ReconcilerTests16.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ReconcilerTests16.java
@@ -299,4 +299,84 @@ public void testissue342_001() throws Exception {
 		deleteProject(p);
 	}
 }
+public void testGH612_001() throws Exception {
+	if (!isJRE16)
+		return;
+	IJavaProject p = createJava16Project("p");
+	createFolder("/p/src/a");
+	try {
+		String source =
+				"package a;\n"
+				+ "public class X {\n"
+				+ "    public static void main(String[] args) {\n"
+				+ "        Sub sub = new Sub();\n"
+				+ "        sub.method1();\n"
+				+ "    }\n"
+				+ "\n"
+				+ "    static class Outer<T> { \n"
+				+ "        class Inner {\n"
+				+ "            void sayHi() {\n"
+				+ "                System.out.println(\"hello world!\");\n"
+				+ "            }\n"
+				+ "        }\n"
+				+ "    }\n"
+				+ "\n"
+				+ "    static class Sub extends Outer<Sub.Inner> {\n"
+				+ "        void method1() {\n"
+				+ "            Inner inner = new Inner();\n"
+				+ "            inner.sayHi();\n"
+				+ "        }\n"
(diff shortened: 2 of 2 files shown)


snjeza:
test this please

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ReconcilerTests16.java
```diff
@@ -299,4 +299,44 @@ public void testissue342_001() throws Exception {
...
+				+ "\n"
+				+ "    static class Outer<T> { \n"
+				+ "        class Inner {\n"
+				+ "            void sayHi() {\n"
+				+ "                System.out.println(\"hello world!\");\n"
+				+ "            }\n"
+				+ "        }\n"
+				+ "    }\n"
+				+ "\n"
+				+ "    static class Sub extends Outer<Sub.Inner> {\n"
```
stephan-herrmann, line 324: I can break your fix by changing `extends` to `implements` with the necessary changes to `Outer`.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
```diff
@@ -3027,6 +3027,13 @@ public final ReferenceBinding getJavaUtilIterator() {
 	*/
 	public final ReferenceBinding getMemberType(char[] typeName, ReferenceBinding enclosingType) {
 		ReferenceBinding memberType = findMemberType(typeName, enclosingType);
```
stephan-herrmann, line 3029: I wonder if the change can be made inside `findMemberType`, near the location where currently `null` is returned?

Otherwise, I suggest to let the new code below check if `enclosingType.isHierarchyBeingConnected()` to ensure that the additional code is invoked only in this particular situation.

stephan-herrmann (changes requested):
Thanks, @snjeza, the change is going in a good direction.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ReconcilerTests16.java
snjeza: Fixed.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
snjeza: > I wonder if the change can be made inside findMemberType, near the location where currently null is returned?

I tried that, but some compiler regression tests fail.

> Otherwise, I suggest to let the new code below check if enclosingType.isHierarchyBeingConnected() to ensure that the additional code is invoked only in this particular situation.

It has been fixed.

stephan-herrmann:
For posterity: in search for the best location of the fix I suggested to **push it into** `findMemberType()` which @snjeza  answered as causing regressions in some tests. I then made the experiment of **pulling it out** into `QualifiedTypeReference.findNextTypeBinding()`, like so:
```

	protected TypeBinding findNextTypeBinding(int tokenIndex, Scope scope, PackageBinding packageBinding) {
		LookupEnvironment env = scope.environment();
		try {
			env.missingClassFileLocation = this;
			if (this.resolvedType == null) {
				this.resolvedType = scope.getType(this.tokens[tokenIndex], packageBinding);
			} else {
// begin change:
				ReferenceBinding enclosingType = (ReferenceBinding) this.resolvedType;
				this.resolvedType = scope.getMemberType(this.tokens[tokenIndex], enclosingType);
				if (!this.resolvedType.isValidBinding()) {
					TypeBinding problemBinding = this.resolvedType;
					if (enclosingType.isHierarchyBeingConnected() && scope.kind == Scope.CLASS_SCOPE) {
						TypeDeclaration referenceContext = ((ClassScope) scope).referenceContext;
						TypeReference superTypeReference = referenceContext.superclass;
						if (superTypeReference != null && superTypeReference.resolvedType instanceof ReferenceBinding) {
							this.resolvedType = scope.findMemberType(this.tokens[tokenIndex], (ReferenceBinding) superTypeReference.resolvedType);
						}
						if ((this.resolvedType == null || !this.resolvedType.isValidBinding())
								&& referenceContext.superInterfaces != null && referenceContext.superInterfaces.length > 0) {
							TypeReference[] interfaces = referenceContext.superInterfaces;
							for (TypeReference reference : interfaces) {
								if (reference != null && reference.resolvedType instanceof ReferenceBinding) {
									this.resolvedType = scope.findMemberType(this.tokens[tokenIndex], (ReferenceBinding) reference.resolvedType);
									if (this.resolvedType != null) {
										break;
									}
								}
							}
						}
						if (this.resolvedType != null && this.resolvedType.isValidBinding()) {
							return this.resolvedType;
						}
					}

					this.resolvedType = new ProblemReferenceBinding(
						CharOperation.subarray(this.tokens, 0, tokenIndex + 1),
						(ReferenceBinding) problemBinding.closestMatch(),
						this.resolvedType.problemId());
// end change
				}
			}
			return this.resolvedType;
		} catch (AbortCompilation e) {
			e.updateContext(this, scope.referenceCompilationUnit().compilationResult);
			throw e;
		} finally {
			env.missingClassFileLocation = null;
		}
	}
```

On the one hand this would reflect the fact, that indeed a qualified type reference caused the original error. OTOH I cannot prove this to be a better location. That's why I'm just dumping that version here, should the change in `getMemberType()` cause trouble in a future incident.

stephan-herrmann (approved):

iloveeclipse:
Since we've just merged JAVA20 branch code to master, I 've rebased branch to make sure we don't introduce a regression.

stephan-herrmann:
OK, so my extended description came a bit late :)  
I had attempted to add:

> Ensure qualified type references are correctly resolved relative to a type whose hierarchy is currently being conntected

iloveeclipse:
@stephan-herrmann : any objections to merge now, or you are still on it?

PR #630, opened 2023-01-08
Critical compilation bug when null analysis enabled.Fixes #629

### LINKED ISSUE #629: Critical compilation bug when null analysis enabled
```
import org.springframework.context.ApplicationContextInitializer;
import org.springframework.context.ConfigurableApplicationContext;
import org.springframework.test.context.ContextConfiguration;
@ContextConfiguration(initializers = {AbstractTest.Initializer.class})
public abstract class AbstractTest {

  static class Initializer
      implements ApplicationContextInitializer<ConfigurableApplicationContext> {

    @Override
    public void initialize(ConfigurableApplicationContext configurableApplicationContext) {
        
    }
  }
}
```

When null analysis enabled I get this error

```
Type mismatch: cannot convert from Class<AbstractTest.Initializer> to Class<? extends ApplicationContextInitializer<?>>Java(16777233)
```

snjeza:
Fixes #629 

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
@@ -60,6 +60,7 @@ import org.eclipse.jdt.core.compiler.InvalidInputException;
 import org.eclipse.jdt.internal.compiler.ast.LambdaExpression;
 import org.eclipse.jdt.internal.compiler.ast.MethodDeclaration;
 import org.eclipse.jdt.internal.compiler.ast.NullAnnotationMatching;
+import org.eclipse.jdt.internal.compiler.ast.TypeReference;
 import org.eclipse.jdt.internal.compiler.classfmt.ClassFileConstants;
 import org.eclipse.jdt.internal.compiler.impl.CompilerOptions;
 import org.eclipse.jdt.internal.compiler.impl.JavaFeature;
@@ -1334,6 +1335,27 @@ public boolean implementsInterface(ReferenceBinding anInterface, boolean searchH
 				for (int b = 0; b < nextPosition; b++)
 					if (TypeBinding.equalsEquals(next, interfacesToVisit[b])) continue nextInterface;
 				interfacesToVisit[nextPosition++] = next;
+			}
+		}
+	}
+	// see https://github.com/eclipse-jdt/eclipse.jdt.core/issues/629
+	if (nextPosition == 0 && this instanceof SourceTypeBinding) {
+		SourceTypeBinding sourceType = (SourceTypeBinding) this;
+		if (sourceType.scope != null && sourceType.scope.referenceContext != null && sourceType.scope.compilerOptions().isAnnotationBasedNullAnalysisEnabled) {
+			TypeReference[] references = sourceType.scope.referenceContext.superInterfaces;
+			if (references == null || references.length == 0) {
+				return false;
+			}
+			for (TypeReference reference: references) {
+				if (!(reference.resolvedType instanceof ReferenceBinding)) {
+					reference.resolveType(sourceType.scope);
+				}
+				if (reference.resolvedType instanceof ReferenceBinding) {
+					ReferenceBinding binding = (ReferenceBinding) reference.resolvedType;
+					if (binding.isEquivalentTo(anInterface)) {
+						return true;
+					}
+				}
 			}
 		}
 	}
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/NullAnnotationTests18.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/NullAnnotationTests18.java
@@ -19,6 +19,7 @@ import java.io.File;
 import java.io.IOException;
 import java.util.Map;
 
+import org.eclipse.jdt.core.JavaCore;
 import org.eclipse.jdt.core.tests.util.Util;
 import org.eclipse.jdt.internal.compiler.impl.CompilerOptions;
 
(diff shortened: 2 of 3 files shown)


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
```diff
@@ -1337,6 +1338,24 @@ public boolean implementsInterface(ReferenceBinding anInterface, boolean searchH
 			}
 		}
 	}
+	// see https://github.com/eclipse-jdt/eclipse.jdt.core/issues/629
```
gayanper, line 1341: I think this solves the issue, But I doubt why superInterfaces are not resolved when the null annotation processing is enabled. Its something I spent last 2-3 days few hours each to see if I could 😞. 

But unless we get some input on why it's not properly resolved, This will be a good quick fix for the issue. We can change later if we find a better fix suggestion.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/NullAnnotationTests18.java
```diff
@@ -628,4 +629,32 @@ public void testNullableVar() {
 				"----------\n";
 		runner.runNegativeTest();
 	}
+	public void testGH629() {
```
gayanper, line 632: I guess we have tests to make sure this kind of resolutions are working when null annotation are not enabled. May be something to think about to see if such test is needed.

gayanper (approved):

gayanper:
@snjeza Let give few days for @stephan-herrmann or @jarthana to provide feedback on this, otherwise I suggest we merge this which we can later revisit if improvement is needed

stephan-herrmann:
Given the analysis in https://github.com/eclipse-jdt/eclipse.jdt.core/issues/629#issuecomment-1374858823 I can break the fix, by extending the test like so:

```
"App.java",
		"interface I<T> {}\n" +
		"@Annot(inits = {App.MyInit.class})\n" +
		"public class App {\n" +
		"    static class MyInit implements I<String>, Init<Configuration> {}\n" +
		"}\n"
```

Now we branch into all that on-demands stuff already when resolving `I<String>` at which point the second type reference  (`Init<Configuration>`) is not yet resolved :(

gayanper:
> 



> Given the analysis in [#629 (comment)](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/629#issuecomment-1374858823) I can break the fix, by extending the test like so:
> 
> ```
> "App.java",
> 		"interface I<T> {}\n" +
> 		"@Annot(inits = {App.MyInit.class})\n" +
> 		"public class App {\n" +
> 		"    static class MyInit implements I<String>, Init<Configuration> {}\n" +
> 		"}\n"
> ```
> 
> Now we branch into all that on-demands stuff already when resolving `I<String>` at which point the second type reference (`Init<Configuration>`) is not yet resolved :(

Thanks for the insight and another test case @stephan-herrmann, @snjeza can you look into see if we can provide a fix to solve what is described in the issue comment ?

snjeza:
>  can you look into see if we can provide a fix to solve what is described in the issue comment ?

@gayanper @stephan-herrmann I have updated the PR. Please review.

snjeza:
@gayanper @stephan-herrmann  org.eclipse.jdt.core.tests.model.InclusionPatternsTests tests fail randomly
- https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-630/7/testReport/junit/org.eclipse.jdt.core.tests.model/InclusionPatternsTests/testRenameIncludedCompilationUnit/
- https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-630/6/testReport/junit/org.eclipse.jdt.core.tests.model/InclusionPatternsTests/testNestedSourceFolder6/

I can't reproduce it in Eclipse. All 40 tests in org.eclipse.jdt.core.tests.model.InclusionPatternsTests always pass in Eclipse.
These tests are in another project (model).

gayanper:
@snjeza if I understood what @stephan-herrmann shared with us, I think correct way we should try to solve this issue is by deferring the effecting computation. @stephan-herrmann can provide more insight if my assumption is not accurate.

stephan-herrmann:
> @snjeza if I understood what @stephan-herrmann shared with us, I think correct way we should try to solve this issue is by deferring the effecting computation. @stephan-herrmann can provide more insight if my assumption is not accurate.

You correctly guess that I was leaning towards deferring, which would be in line with prior art near by, as the wiki link shows.

But I'm still not sure which direction is best, actually.

Generally, AST nodes should not be resolved more than once, which might serve as a reason to prefer deferring over on-demand resolving. But this general policy may not be necessary for TypeReferences (for comparison, I know that double resolving of NameReference can cause havoc).

Interestingly, org.eclipse.jdt.internal.compiler.ast.TypeReference.internalResolveType(Scope, int) has this comment:
```
	if (this.resolvedType != null) { // is a shared type reference which was already resolved
```
this sounds promising, but it must be checked whether all subclasses are safe in this regard.

I'm happy to see that the mentioned method already handles the case, when a previous resolve produced a problem binding.

Additionally, we might ask if _more_ on-demand resolving (ahead of time) can trigger yet more undesired resolving, and similar problems of the same kind.

snjeza:
> , I think correct way we should try to solve this issue is by deferring the effecting computation

@gayanper @stephan-herrmann I have tried to defer annotation processing but it causes a lot of tests to fail.

I have found yet another issue.
The following test fails:
```
public void testGH629a() {
		Map<String, String> options = getCompilerOptions();
		options.put(CompilerOptions.OPTION_Source, CompilerOptions.VERSION_18);
		options.put(JavaCore.COMPILER_NONNULL_ANNOTATION_NAME, "test.NonNull");
		options.put(JavaCore.COMPILER_NULLABLE_ANNOTATION_NAME, "test.Nullable");
		options.put(JavaCore.COMPILER_ANNOTATION_NULL_ANALYSIS, JavaCore.ENABLED);

		runNegativeTestWithLibs(
				new String[] {
						"Annot.java",
						"public @interface Annot {\n" +
								"    Class<? extends Init<? extends Configuration>>[] inits(); \n" +
								"}\n",
						"App.java",
						"@Annot(inits = {App.MyInit.class})\n" +
						"public class App {\n" +
						"    static class MyInit implements Init<Configuration> {}\n" +
						"}\n",
						"Configuration.java",
						"public interface Configuration {\n" +
								"}\n",
						"Init.java",
						"public interface Init<C extends Configuration> {\n" +
								"}\n"
				},
				options,
				"");
	}
```
Trying to fix it.

gayanper:
LGTM and @snjeza thanks for working on this. Just for curiosity did you find any issues when you tried to implement a differing solution, what kind of issues you found ? Was there any issues related to re-resolving AST nodes like Stephan mentioned in his comment ?

snjeza:
>  Just for curiosity did you find any issues when you tried to implement a differing solution, what kind of issues you found ? Was there any issues related to re-resolving AST nodes like Stephan mentioned in his comment ?

I have faced a lot of failed tests.

gayanper:
> > Just for curiosity did you find any issues when you tried to implement a differing solution, what kind of issues you found ? Was there any issues related to re-resolving AST nodes like Stephan mentioned in his comment ?
> 
> I have faced a lot of failed tests.

Were they due resolution issues ? And the failures happened on when you ran the suites locally is it ? Because time to time we have CI test failures.

snjeza:
> Were they due resolution issues ? And the failures happened on when you ran the suites locally is it 

Right.

gayanper:
Thanks @snjeza

stephan-herrmann:
> Additionally, we might ask if _more_ on-demand resolving (ahead of time) can trigger yet more undesired resolving, and similar problems of the same kind.

I'm afraid that exactly this might have happened causing #1007

stephan-herrmann:
For posterity: we have been going back and forth in this area: when exactly is it safe to resolve annotations? We kept seeing new problems with either too early or too late resolved annotations.  

Meanwhile, however, @srikanth-sankaran found a way to solve a root cause of all this: #4740 which allows us to update TypeVariableBindings in place, so that the update is automatically seen also in all sites referring to that TVB.

I'm documenting this connection to confirm that from the code base of 2023 the fix provided here looked reasonable. Just we hadn't really put the finger on the root cause (which existed since the introduction of type annotations).

PR #668, opened 2023-01-30
Bug 578073 - [18] JSR 269 - Implement Elements.getFileObjectOf(Element)

jarthana:
## Author checklist


jarthana:
This implementation is long overdue from our side. I was hesitant in the beginning since it meant we needed to provide a link between our binding and the resource path somehow. It wasn't that difficult with source binding since we had a way of finding the path from the CompilationResult. But binary binding is a different story. I had to alter ClassFileReader and IBinaryType/IBinaryModule to ensure that we stored the path at the time of creation and later use it in the ElementsImpl.

jarthana:
@stephan-herrmann @iloveeclipse This is still WIP since I have a test failure to address. But would like to get your opinion and see if I am missing a better way to achieve this.

As you can see, I am only storing the BinaryTypeBinding#path only when annotation processing is enabled. But still it means we keep a reference to the path and therefore to the file system itself, including the jrtfs. Do you see any problem with it? Or should I move this condition much earlier, i.e. even when I am creating the ClassFileReader itself?

stephan-herrmann:
@jarthana all those debug prints in the code make it a bit difficult to see which are the essential changes. Force pushing those changes also blurs the picture. Should we wait until you are done debugging?

jarthana:
> @jarthana all those debug prints in the code make it a bit difficult to see which are the essential changes. Force pushing those changes also blurs the picture. Should we wait until you are done debugging?

Yes, please. Sorry about it. Meanwhile, I moved away from storing the java.nio.Path to URI. Seems better to me. 

The barrage of syso statements are for the lone failing test, which was due to the fact that I was opening the FileSystem on a zip file which no longer exists under certain circumstances. Unfortunately, the test passes on my setup. Hence, the noise in the console.

jarthana:
> @jarthana all those debug prints in the code make it a bit difficult to see which are the essential changes. Force pushing those changes also blurs the picture. Should we wait until you are done debugging?

@stephan-herrmann Can you take a look now? There's still one scenario where I haven't applied the changes since it's breaking one of the tests. But the essence of the fix is still there.

jarthana:
@stephan-herrmann, you can ignore the second commit for now. It is experimental.

jarthana:
Something I am not very happy about this patch is in Util#getZipEntryClassFile().

This method opens a new FileSystem every time we want to read an entry from a zip file. Not a bad thing, except when we sent sequence of individual requests for the same Jar (but different file). This happens when we are searching and we have multiple search documents all pointing to the same archive. Other than that, I am reasonably happy that this doesn't add much cost to the existing code in terms of performance.

@stephan-herrmann Have you had a chance to look at this? In the interest of time, may be I can release this patch and then we can address the concerns if any soon after?

stephan-herrmann:
> @stephan-herrmann Have you had a chance to look at this?

sorry, not yet. 
Maybe on Thursday.

> In the interest of time, may be I can release this patch and then we can address the concerns if any soon after?

Sounds good to me.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/tool/PathFileObject.java
```diff
@@ -0,0 +1,228 @@
...
+
+import org.eclipse.jdt.internal.compiler.classfmt.ClassFileConstants;
+import org.eclipse.jdt.internal.compiler.classfmt.ClassFileReader;
+import org.eclipse.jdt.internal.compiler.classfmt.ClassFormatException;
+import org.eclipse.jdt.internal.compiler.util.JRTUtil;
+
+/**
+ * Implementation of a Java file object that corresponds to a file on the file system
+ */
+public class PathFileObject implements JavaFileObject {
```
stephan-herrmann, line 39: This new class seems to be touched by tests only superficially. I can see that much of the implementation is trivial, but still ...

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/classfmt/ClassFileReader.java
```diff
@@ -162,8 +166,7 @@ public static ClassFileReader read(
 	java.util.zip.ZipEntry ze = zip.getEntry(filename);
 	if (ze == null)
 		return null;
-	byte classFileBytes[] = Util.getZipEntryByteContent(ze, zip);
-	ClassFileReader classFileReader = new ClassFileReader(classFileBytes, filename.toCharArray());
+	ClassFileReader classFileReader = Util.getZipEntryClassFile(zip.getName(), filename);
```
stephan-herrmann, line 169: Please help me understand, why we can't use the `ze`, after it has already been retrieved successfully?
Passing the URI into the `ClassFileReader` should be possible without `getZipEntryClassFile()`, no? Or is the encoding of an URI relative to a zip file the problem?

If opening a `ZipFileSystem` is a potential performance issue, I'd think an URI like `jar:file:///tmp/eclipse-temp/lib/apttestprocessors8.jar!/org/eclipse/jdt/compiler/apt/tests/annotations/Type.class` can be inexpensively constructed by simple string operations?

In tests, this seemed to work:

```
	URI uri = URI.create("jar:"+Paths.get(zip.getName()).toUri()+"!/"+filename);
```
Does it lack any fine points of URI encoding?

jarthana:
Thanks @stephan-herrmann! All are valid concerns and I see Andrey has already taken care of the important ones. I will fix the remaining shortly.

jarthana:
The review comments are addressed in #733 

Having a tough time with most of them. Javac returns null for most methods, so can't really compare the output. And the code is a copy paste anyway from out existing classes. So...

Anyway, I have added few more tests for the PathFileObject.

PR #672, opened 2023-01-30
Create compile error in case 'var' is used with type arguments. Fixes #600

### LINKED ISSUE #600: Missing error message for generics on var
An error should be reported when there are generics on `var`, but the JDT compiler does not complain and compiles anyway. The following class currently compiles and runs, but an error should be reported:
```java
package test;

import java.util.ArrayList;

public class Test {
        var<String> list = new ArrayList<String>();
        System.out.println(list.size());
    }
}
```
When I run this class through `javac`, the following error is produced:
```
Test.java:7: error: illegal reference to restricted type 'var'
		var<String> list = new ArrayList<String>();
		^
```

simfis:
## What it does
In issue #600 it was noted that the JDT compiler accepts statements like 

`var<Integer> x = List.of(42);`

which do violate the Java language specification (see [here](https://docs.oracle.com/javase/specs/jls/se19/html/jls-14.html#jls-14.4)).

I've addressed this issue by adding a check in the resolve method of the _LocalDeclaration_ class, which will report a compile error using _ProblemReporter_ class. Additionally, I've created a new test in _JEP286Test_.

## How to test
Can be tested by running  _JEP286Test_ / letting the parser loose on statements like the one quoted above.

## Author checklist

- [x] I have thoroughly tested my changes (I ran _RunCompilerTests_)

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/core/compiler/IProblem.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/core/compiler/IProblem.java
@@ -2180,7 +2180,7 @@ void setSourceStart(int sourceStart);
 	int VarIsNotAllowedHere = Syntax + 1511; // ''var'' is not allowed here
 	/** @since 3.16 */
 	int VarCannotBeMixedWithNonVarParams = Syntax + 1512; // ''var'' cannot be mixed with explicit or implicit parameters
-	/** @since 3.33 */
+	/** @since 3.35 */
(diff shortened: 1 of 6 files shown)


simfis:
I've added another test to cover variable declaration in basic for statements. Unfortunately, in this case further errors are created in the condition expression and the increments statements. I'll investigate how to deal with this properly and create tests for enhanced for-loops and try-with-resources statements as well.

srikanth-sankaran:
Thanks for the patch - I will review this next week

srikanth-sankaran:
Brought this PR up to date by pulling and rebasing on top of all recent master changes in preparation for review

srikanth-sankaran (approved):
Looks good, Thanks a lot for your contribution.

Issue #674, opened 2023-01-31
Enhance the BuildContext with the discovered annotations


laeubi:
I'm writing a `CompilationParticipant` that wants to process annotations.

While I can get notified (in `CompilationParticipant.processAnnotations(BuildContext[])`) that annotations *are* found, there is no way to discover what kind of annotations are found, so currently the only way would be to scan the source (or [try using the classfile](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/673)).

Both approaches seems wasteful in terms of resources, effort and time, as JDT already *has* discovered the annotation, it would therefore be great if there would be a method `BuildContext#getAnnotations` that return the already parsed and discovered annotations, see also:

- https://github.com/eclipse-jdt/eclipse.jdt.core/discussions/523

PR #676, opened 2023-01-31
Enhance the BuildContext with the discovered annotations

laeubi:
## What it does
Currently one can only get the information that annotations are present on a file in a `BuildContext` / `CompilationParticipant` but sometimes it is much more interesting to know if a specific annotation is present so one don't need to scan files unnecessary.

This captures the found low-level annotations and adds a new method based on this `BuildContext.hasAnnotations(String)` so that `CompilationParticipant` can check if the file is of any interest for them.

Fix https://github.com/eclipse-jdt/eclipse.jdt.core/issues/674

## How to test

- Have a `CompilationParticipant`
- Have a project with a source file that contains an specific annotation and one without but containing another
- in the method `CompilationParticipant.processAnnotations(BuildContext[])` check that `buildContext.hasAnnotations("...")` returns true for you annotation of interest.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2021 IBM Corporation and others.
+ * Copyright (c) 2000, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -10,6 +10,7 @@
  *
  * Contributors:
  *     IBM Corporation - initial API and implementation
+ *     Christoph Läubrich -  Enhance the BuildContext with the discovered annotations #674
  *******************************************************************************/
 package org.eclipse.jdt.internal.core.builder;
 
@@ -23,6 +24,7 @@ import org.eclipse.jdt.internal.compiler.Compiler;
 import org.eclipse.jdt.internal.compiler.classfmt.ClassFileConstants;
 import org.eclipse.jdt.internal.compiler.env.ICompilationUnit;
 import org.eclipse.jdt.internal.compiler.impl.CompilerOptions;
+import org.eclipse.jdt.internal.compiler.lookup.AnnotationBinding;
 import org.eclipse.jdt.internal.compiler.lookup.TypeConstants;
 import org.eclipse.jdt.internal.compiler.problem.*;
 import org.eclipse.jdt.internal.compiler.util.SimpleSet;
@@ -60,7 +62,7 @@ protected boolean compiledAllAtOnce;
 private boolean inCompiler;
 
 protected boolean keepStoringProblemMarkers;
-protected Set<SourceFile> filesWithAnnotations = null;
+protected Map<SourceFile, AnnotationBinding[]> filesWithAnnotations = null;
 
 //2000 is best compromise between space used and speed
 public static int MAX_AT_ONCE = Integer.getInteger(JavaModelManager.MAX_COMPILED_UNITS_AT_ONCE, 2000).intValue();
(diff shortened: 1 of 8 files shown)


laeubi:
@stephan-herrmann or @iloveeclipse can you review this or suggest a reviewer?

This would be really helpful for PDE to not scan source files unnecessarily for annotations.

laeubi:
The testfailure seems unrelated?

iloveeclipse:
> The testfailure seems unrelated?

Yes. Can you please add some tests using new API?

laeubi:
I have now enhanced `ParticipantBuildTests` with a new testcase.

laeubi:
Build has passed now! 🥳

laeubi:
@iloveeclipse could this be merged or is anything left here?

iloveeclipse:
Someone need to review this PR. I could do it later, this weekend or next week, I'm busy right now with internal HR work and mostly not have access to my workstation.

laeubi:
@iloveeclipse any chance for this PR to be reviewed, or maybe suggest a reviewer for this?

jarthana:
Let's consider this for M1.

laeubi:
> Let's consider this for M1.

Please let me know if anything has to be done from my side!

jarthana:
I am considering this for merging this week. If anything, we can take up in follow up bugs. Let's see if we can get rid of the jenkins failures.

laeubi:
I think I found the problem was some refactorings in the meantime and the change of the verions so there was an API error.

laeubi:
@jarthana the testfailure seems unrelated as it can be seen in other PRs as well e.g. here: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1155

iloveeclipse:
Please rebase to get rid of #1150 issue

laeubi:
@iloveeclipse @jarthana this now is rebased and test are passing.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java
```diff
@@ -219,8 +221,11 @@ public void acceptResult(CompilationResult result) {
 					createProblemFor(compilationUnit.resource, null, Messages.build_inconsistentClassFile, JavaCore.ERROR);
 			}
 		}
-		if (result.hasAnnotations && this.filesWithAnnotations != null) // only initialized if an annotation processor is attached
-			this.filesWithAnnotations.add(compilationUnit);
+		if (result.hasAnnotations && this.filesWithAnnotations != null) {
```
jarthana, line 224: hasAnnotations is set even when the result is an annotation, in which case, we add this to the map even when the AnnotationBinding[] is empty. I believe we can avoid this being added to the map?

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/CompilationParticipantResult.java
```diff
@@ -143,9 +164,9 @@ public void recordNewProblems(CategorizedProblem[] newProblems) {
 	this.problems = merged;
 }
 
-void reset(boolean detectedAnnotations) {
+void reset(AnnotationBinding[] annotations) {
```
jarthana, line 167: This introduces a new warning about the parameter hiding a field.

jarthana (changes requested):
This is just an observation. I see that we have the types already being added to the AbstractImageBuilder.newState.typeLocators. I don't see a direct way to use them, though.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java
laeubi: So your suggestion is tho also check for `annotationbinding.length!=0`?

jarthana: Yes

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/CompilationParticipantResult.java
laeubi: This should be fixed now!

jarthana (approved):

jarthana:
Now it fails with a different test. I see it passed locally, but the fact that it is not a test that has failed before and that it is related to the annotations, makes me wonder.

laeubi:
I tried to enhance the assertin failure with some more information.

laeubi:
@jarthana I think the problem is the check for empty, because before the code was

`if (result.hasAnnotations && this.filesWithAnnotations != null) `

I will revert this, rebase (again) and then think it could be merged as is and if someone with deeper knowledge want to optimize it later on?

laeubi:
> Please fix this typo: precence => presence and then we are good to go.

done!

Issue #675, opened 2023-01-31
Wrong call hierarchy for jars with module conflict on compile classpath


trancexpress:
To reproduce:

1. Import projects from attached archive: [TestCallHierarchyBug.zip](https://github.com/eclipse-jdt/eclipse.jdt.core/files/10544219/TestCallHierarchyBug.zip)
2. Open type test.Test.
3. Select the only field testField, right-click -> "Open Call Hierarchy".
4. Observe no matches.
5. Select testField again, right-click -> "References" -> "Workspace".
6. Observe the expected match, denoted as "potential match".

After debugging, it looks like `JavaSearchNameEnvironment.getModulesDeclaringPackage()` determines multiple declaring modules for `testpackage` (the unnamed module and `TestLibraryWithModule`). From there, a `SplitPackageBinding` is created for `testpackage` and `CompilationUnitScope.findImport()` will skip the import, creating a problem for the statement `import testpackage.TestClass`. This in turn results in a `MissingTypeBinding` for `TestClass` (during the call hierarchy computation). And so the search match in `test.Test.bar()` is considered an inaccurate match (also seen when computing workspace references of `testField`) and not reported for the call hierarchy.

The jars with the module conflict are on the compile **classpath**, not on the module path. So its odd to see this problem. If the 2 jars are on the module path, then there is a compile problem. So for the jars on a module path, the call hierarchy not finding anything would be expected. But not for the jars on the classpath.

iloveeclipse:
In my eyes if "Reference search" can find calls, so "Call Hierarchy" should find them too, and vice versa.

Also from user point of view, *especially* if classpath/module path has some duplicates that need to be solved, one should be able to see where which code refers to what.

iloveeclipse:
Just a note for myself after debugging & reviewing the code.

The key point in the bug and in the fix is that original code (added for 9+ support) **automatically** added "module" info to the search environment from the jar with module-info.class inside, **even** if that jar itself was used in a non-modular project (and placed on classpath).

So the fix adds differentiation between "modular" / "non-modular" classpath entries given to the search. So for a non modular classpath entry `JavaSearchNameEnvironment` doesn't try to add module info to the jar's `ClasspathLocation` and it doesn't land in `moduleToClassPathLocations` so later this info doesn't appear in `JavaSearchNameEnvironment.getModulesDeclaringPackage()` that confuses later code, and so on.

PR #679, opened 2023-01-31
Wrong call hierarchy computed for type in conflicting modules

trancexpress:
JavaSearchNameEnvironment adds module infos to each PackageFragmentRoot of a project, whether the root fragment is on the compile classpath or module path. However the module info is only valid for fragments that are on the module path.

As a result, a search (call hierarchy, references) can encounter non-existent compile errors such as module conflicts. The project itself compiles fine, but the search detects such conflicts due to the module infos attached to each PackageFragmentRoot.

This change adjusts JavaSearchNameEnvironment to not add module info for a PackageFragmentRoot which is not on the module path of the project.

Fixes: #675

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/matching/JavaSearchNameEnvironment.java
+++ b/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/matching/JavaSearchNameEnvironment.java
@@ -26,6 +26,7 @@ import java.util.stream.Collectors;
 import org.eclipse.core.resources.IContainer;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
+import org.eclipse.jdt.core.IClasspathEntry;
 import org.eclipse.jdt.core.IJavaElement;
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.IModuleDescription;
@@ -281,7 +282,24 @@ private ClasspathLocation mapToClassPathLocation(JavaModelManager manager, Packa
 	}
 	JavaProject javaProject = root.getJavaProject();
 	if (isComplianceJava9OrHigher(javaProject)) {
-		addModuleClassPathInfo(root, defaultModule, cp);
+		boolean isOnModulePath = true; // if an exception occurs, assume yes
+		try {
+			IClasspathEntry classpathEntry = root.getRawClasspathEntry();
+			if (classpathEntry.getEntryKind() == IClasspathEntry.CPE_SOURCE) {
+				/*
+				 * Source classpath entries of a project never have the module attribute set,
+				 * so we cannot rely on the attribute.
+				 */
+				isOnModulePath = isModularProject(javaProject);
+			} else {
+				isOnModulePath = ClasspathEntry.isModular(classpathEntry);
+			}
+		} catch (JavaModelException e) {
+			Util.log(e, "Error checking whether PackageFragmentRoot is on module path!"); //$NON-NLS-1$
+		}
+		if (isOnModulePath) {
+			addModuleClassPathInfo(root, defaultModule, cp);
+		}
 	}
 	return cp;
 }
@@ -562,5 +580,11 @@ private static boolean isComplianceJava9OrHigher(IJavaProject javaProject) {
 		return false;
 	}
 	return CompilerOptions.versionToJdkLevel(javaProject.getOption(JavaCore.COMPILER_COMPLIANCE, true)) >= ClassFileConstants.JDK9;
+}
+
+private static boolean isModularProject(IJavaProject project) throws JavaModelException {
+	IModuleDescription module = project.getModuleDescription();
+	String modName = module == null ? null : module.getElementName();
+	return modName != null && modName.length() > 0;
 }
 }
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/AbstractJavaModelTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/AbstractJavaModelTests.java
@@ -4041,6 +4041,25 @@ public abstract class AbstractJavaModelTests extends SuiteOfTestCases {
 		}
 	}
 
+	protected void makeJCLModular(IJavaProject javaProject) throws JavaModelException {
+		IClasspathEntry[] classpath = javaProject.getRawClasspath();
+		for (int i = 0, length = classpath.length; i < length; i++) {
+			IClasspathEntry entry = classpath[i];
+			final IPath path = entry.getPath();
+			if (isJCLPath(path)) {
+					classpath[i] = JavaCore.newVariableEntry(
+							entry.getPath(),
+							entry.getSourceAttachmentPath(),
+							entry.getSourceAttachmentRootPath(),
+							entry.getAccessRules(),
+							moduleAttribute(),
+							entry.isExported());
+					break;
+			}
+		}
+		javaProject.setRawClasspath(classpath, null);
+	}
+
 	private static void logError(String errorMessage, CoreException e) {
 		Plugin plugin = JavaCore.getPlugin();
 		if (plugin != null) {
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchBugs15Tests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchBugs15Tests.java
@@ -135,6 +135,7 @@ public class JavaSearchBugs15Tests extends AbstractJavaSearchTests {
 	public void setUpSuite() throws Exception {
 		super.setUpSuite();
 		JAVA_PROJECT = setUpJavaProject("JavaSearchBugs", "16");
+		makeJCLModular(JAVA_PROJECT);
 	}
 
 	public void tearDownSuite() throws Exception {
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchBugs9Tests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchBugs9Tests.java
@@ -503,7 +503,7 @@ public void testBug501162_008() throws Exception {
 				"package pack22;\n" +
 				"public interface I22 {}\n");
 
-		addClasspathEntry(project1, JavaCore.newProjectEntry(project2.getPath()));
+		addModularProjectEntry(project1, project2);
 		project1.close(); // sync
 		project2.close();
 		project2.open(null);
@@ -1081,7 +1081,7 @@ public void testBug501162_019() throws Exception {
 				"    public ITwo i2;\n" +
 				"    public XOne X1;\n" +
 				"}\n");
-		addLibraryEntry(project1, "/JavaSearchBugs/lib/bzero501162.jar", false);
+		addModularLibraryEntry(project1, new Path("/JavaSearchBugs/lib/bzero501162.jar"), null);
 		project1.close(); // sync
 		project1.open(null);
 		SearchPattern pattern = SearchPattern.createPattern("pack.one",
@@ -1120,7 +1120,7 @@ public void testBug501162_020() throws Exception {
 				"    public ITwo i2;\n" +
 				"    public XOne X1;\n" +
 				"}\n");
-		addLibraryEntry(project1, "/JavaSearchBugs/lib/bzero501162.jar", false);
+		addModularLibraryEntry(project1, new Path("/JavaSearchBugs/lib/bzero501162.jar"), null);
 		project1.close(); // sync
 		project1.open(null);
 		SearchPattern pattern = SearchPattern.createPattern("pack.two",
@@ -1160,7 +1160,7 @@ public void testBug501162_021() throws Exception {
 				"    public ITwo i2;\n" +
 				"    public XOne X1;\n" +
 				"}\n");
-		addLibraryEntry(project1, "/JavaSearchBugs/lib/bzero501162.jar", false);
+		addModularLibraryEntry(project1, new Path("/JavaSearchBugs/lib/bzero501162.jar"), null);
 		project1.close(); // sync
 		project1.open(null);
 		SearchPattern pattern = SearchPattern.createPattern("pack.three",
@@ -1200,7 +1200,7 @@ public void testBug501162_022() throws Exception {
 				"    public ITwo i2;\n" +
 				"    public XOne X1;\n" +
 				"}\n");
-		addLibraryEntry(project1, "/JavaSearchBugs/lib/bzero501162.jar", false);
+		addModularLibraryEntry(project1, new Path("/JavaSearchBugs/lib/bzero501162.jar"), null);
 		project1.close(); // sync
 		project1.open(null);
 		SearchPattern pattern = SearchPattern.createPattern("java.base",
@@ -1386,7 +1386,7 @@ public void testBug501162_027() throws Exception {
 				"    public ITwo i2;\n" +
 				"    public XOne X1;\n" +
 				"}\n");
-		addLibraryEntry(project1, "/JavaSearchBugs/lib/bzero501162.jar", false);
+		addModularLibraryEntry(project1, new Path("/JavaSearchBugs/lib/bzero501162.jar"), null);
 		project1.close(); // sync
 		project1.open(null);
 		SearchPattern pattern = SearchPattern.createPattern("XOne",
@@ -1425,7 +1425,7 @@ public void testBug501162_028() throws Exception {
 				"    public ITwo i2;\n" +
 				"    public XOne X1;\n" +
 				"}\n");
-		addLibraryEntry(project1, "/JavaSearchBugs/lib/bzero501162.jar", false);
+		addModularLibraryEntry(project1, new Path("/JavaSearchBugs/lib/bzero501162.jar"), null);
 		project1.close(); // sync
 		project1.open(null);
 		SearchPattern pattern = SearchPattern.createPattern("XFourOne",
@@ -1462,7 +1462,7 @@ public void testBug501162_029() throws Exception {
 				"    public ITwo i2;\n" +
 				"    public XOne X1;\n" +
 				"}\n");
-		addLibraryEntry(project1, "/JavaSearchBugs/lib/bzero501162.jar", false);
+		addModularLibraryEntry(project1, new Path("/JavaSearchBugs/lib/bzero501162.jar"), null);
 		project1.close(); // sync
 		project1.open(null);
 		SearchPattern pattern = SearchPattern.createPattern("ITwo",
@@ -1502,7 +1502,7 @@ public void testBug501162_030() throws Exception {
 				"    public ITwo i2;\n" +
 				"    public XOne X1;\n" +
 				"}\n");
-		addLibraryEntry(project1, "/JavaSearchBugs/lib/bzero501162.jar", false);
+		addModularLibraryEntry(project1, new Path("/JavaSearchBugs/lib/bzero501162.jar"), null);
 		project1.close(); // sync
 		project1.open(null);
 		SearchPattern pattern = SearchPattern.createPattern("IThreeOne",
@@ -1539,7 +1539,7 @@ public void testBug501162_031() throws Exception {
 				"    public ITwo i2;\n" +
 				"    public XOne X1;\n" +
 				"}\n");
-		addLibraryEntry(project1, "/JavaSearchBugs/lib/bzero501162.jar", false);
+		addModularLibraryEntry(project1, new Path("/JavaSearchBugs/lib/bzero501162.jar"), null);
 		project1.close(); // sync
 		project1.open(null);
 		SearchPattern pattern = SearchPattern.createPattern("XThreeOne",
@@ -1576,7 +1576,7 @@ public void testBug501162_032() throws Exception {
 				"    public ITwo i2;\n" +
 				"    public XOne X1;\n" +
 				"}\n");
-		addLibraryEntry(project1, "/JavaSearchBugs/lib/bzero501162.jar", false);
+		addModularLibraryEntry(project1, new Path("/JavaSearchBugs/lib/bzero501162.jar"), null);
 		project1.close(); // sync
 		project1.open(null);
 		SearchPattern pattern = SearchPattern.createPattern("zero",
@@ -1918,7 +1918,7 @@ public void testBug501162_041() throws Exception {
 				"    public ITwo i2;\n" +
 				"    public XOne X1;\n" +
 				"}\n");
-		addLibraryEntry(project1, "/JavaSearchBugs/lib/bzero.src.501162.jar", false);
+		addModularLibraryEntry(project1, new Path("/JavaSearchBugs/lib/bzero.src.501162.jar"), null);
 		project1.close(); // sync
 		project1.open(null);
 		SearchPattern pattern = SearchPattern.createPattern("XOne",
@@ -1994,7 +1994,7 @@ public void testBug501162_043() throws Exception {
 				"    public ITwo i2;\n" +
 				"    public XOne X1;\n" +
 				"}\n");
-		addLibraryEntry(project1, "/JavaSearchBugs/lib/bzero.src.501162.jar", false);
+		addModularLibraryEntry(project1, new Path("/JavaSearchBugs/lib/bzero.src.501162.jar"), null);
 		project1.close(); // sync
 		project1.open(null);
 		SearchPattern pattern = SearchPattern.createPattern("ITwo",
@@ -2074,7 +2074,7 @@ public void testBug501162_045() throws Exception {
 				"    public ITwo i2;\n" +
 				"    public XOne X1;\n" +
 				"}\n");
-		addLibraryEntry(project1, "/JavaSearchBugs/lib/bzero.src.501162.jar", false);
+		addModularLibraryEntry(project1, new Path("/JavaSearchBugs/lib/bzero.src.501162.jar"), null);
 		project1.close(); // sync
 		project1.open(null);
 		SearchPattern pattern = SearchPattern.createPattern("XThreeOne",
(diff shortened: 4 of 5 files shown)


trancexpress:
@iloveeclipse WDYT?

I'm looking into adding a test next.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/matching/JavaSearchNameEnvironment.java
```diff
@@ -563,4 +573,18 @@ private static boolean isComplianceJava9OrHigher(IJavaProject javaProject) {
 	}
 	return CompilerOptions.versionToJdkLevel(javaProject.getOption(JavaCore.COMPILER_COMPLIANCE, true)) >= ClassFileConstants.JDK9;
 }
+
+private static boolean isOnModulePath(PackageFragmentRoot root) throws JavaModelException {
+	IClasspathEntry classpathEntry = root.getRawClasspathEntry();
+	IClasspathAttribute[] attributes = classpathEntry.getExtraAttributes();
```
iloveeclipse, line 579: I see similar code in `org.eclipse.jdt.internal.core.ClasspathEntry.isModular()` and `org.eclipse.jdt.launching.JavaRuntime.isModule(IClasspathEntry, IJavaProject)`.

Looks like this should be just a default method on `IClasspathEntry`

iloveeclipse:
Make sense, but see comment above regarding attribute checks.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/matching/JavaSearchNameEnvironment.java
trancexpress: If we want to introduce new API, best to open an extra issue for that. Can you check the current change? Is that enough? If so, should I adjust also `JavaRuntime` (that seems to be in another plug-in) to use the new method? Though it might be accessing internal code then.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/matching/JavaSearchNameEnvironment.java
```diff
@@ -281,7 +282,16 @@ private ClasspathLocation mapToClassPathLocation(JavaModelManager manager, Packa
 	}
 	JavaProject javaProject = root.getJavaProject();
 	if (isComplianceJava9OrHigher(javaProject)) {
-		addModuleClassPathInfo(root, defaultModule, cp);
+		boolean isOnModulePath = true; // if an exception occurs, assume yes
+		try {
+		    IClasspathEntry classpathEntry = root.getRawClasspathEntry();
+			isOnModulePath = ClasspathEntry.isModular(classpathEntry);
```
iloveeclipse, line 288: Something is wrong with indentation here

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/matching/JavaSearchNameEnvironment.java
iloveeclipse: Right, let split that in two changes.

iloveeclipse:
@trancexpress : either I've messed something with the merge (test had merge issues) or the changes on master cause now fails. Could you please check with your local branch copy?

iloveeclipse:
Actually few Java 9+ search tests have now failures. I assume they worked **without** module attributes set on classpath entries.

trancexpress:
I'll check tomorrow.

  | Test Result (30 failures  / -1)org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_008org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_009org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_012org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_013org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_014org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_015org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_016org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_017org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_018org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_019org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_020org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_021org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_022org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_027org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_028org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_029org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_030org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_031org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_032org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_041org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_043org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_045org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug519980_001org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug519980_002org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug520477_001org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug519151_023org.eclipse.jdt.core.tests.model.JavaSearchBugs15Tests.test566062_001org.eclipse.jdt.core.tests.model.JavaSearchBugs15Tests.test566062_002org.eclipse.jdt.core.tests.model.JavaSearchBugs15Tests.test566062_003org.eclipse.jdt.core.tests.model.JavaSearchBugs15Tests.test566062_004
-- | --


[Test Result](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/) (30 failures / -1)

    [org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_008](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug501162_008/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_009](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug501162_009/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_012](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug501162_012/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_013](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug501162_013/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_014](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug501162_014/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_015](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug501162_015/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_016](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug501162_016/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_017](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug501162_017/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_018](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug501162_018/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_019](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug501162_019/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_020](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug501162_020/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_021](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug501162_021/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_022](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug501162_022/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_027](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug501162_027/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_028](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug501162_028/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_029](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug501162_029/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_030](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug501162_030/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_031](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug501162_031/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_032](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug501162_032/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_041](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug501162_041/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_043](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug501162_043/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug501162_045](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug501162_045/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug519980_001](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug519980_001/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug519980_002](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug519980_002/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug520477_001](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug520477_001/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs9Tests.testBug519151_023](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs9Tests/testBug519151_023/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs15Tests.test566062_001](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs15Tests/test566062_001/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs15Tests.test566062_002](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs15Tests/test566062_002/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs15Tests.test566062_003](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs15Tests/test566062_003/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugs15Tests.test566062_004](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-679/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugs15Tests/test566062_004/)

trancexpress:
There is some weird classpath entry that is added by some of the test... E.g.:

```
<?xml version="1.0" encoding="UTF-8"?>
<classpath>
	<classpathentry kind="src" path="src"/>
	<classpathentry kind="var" path="JCL19_LIB">
		<attributes>
			<attribute name="module" value="true"/>
		</attributes>
	</classpathentry>
	<classpathentry kind="con" path="org.eclipse.jdt.MODULE_PATH"/>
	<classpathentry kind="src" path="/second"/>
	<classpathentry kind="output" path="bin"/>
</classpath>
```

Its maybe related to this bug, when `JavaCore.MODULE_PATH_CONTAINER_ID` was added: https://bugs.eclipse.org/bugs/show_bug.cgi?id=479610

https://github.com/eclipse-jdt/eclipse.jdt.core/commit/9b4c6fe06310162fcf691a59705683aa5c5ce814

```
	/**
	 * @since 3.14
	 */
	public static final String MODULE_PATH_CONTAINER_ID = "org.eclipse.jdt.MODULE_PATH"; //$NON-NLS-1$
```

The tests seem to add a container with this path:

```
addClasspathEntry(project2, JavaCore.newContainerEntry(new Path("org.eclipse.jdt.MODULE_PATH")));
```

I don't know what the significance of entry is, or how its supposed to indicate what is on the module path. But I think as soon as we see one, we should not make assumptions as to what is on the module path and what isn't. I've adjusted the change accordingly.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/matching/JavaSearchNameEnvironment.java
```diff
@@ -563,4 +573,15 @@ private static boolean isComplianceJava9OrHigher(IJavaProject javaProject) {
 	}
 	return CompilerOptions.versionToJdkLevel(javaProject.getOption(JavaCore.COMPILER_COMPLIANCE, true)) >= ClassFileConstants.JDK9;
 }
+
+private static boolean hasModulePathContainer(JavaProject javaProject) throws JavaModelException {
+	IClasspathEntry[] classpath = javaProject.getRawClasspath();
+	for (IClasspathEntry entry : classpath) {
+		IPath entryPath = entry.getPath();
+		if (entryPath != null && JavaCore.MODULE_PATH_CONTAINER_ID.equals(entryPath.toString())) {
```
iloveeclipse, line 581: This sounds like a hack on top of another hack. Would be great to understand what that `MODULE_PATH_CONTAINER` is good for and if that should be better properly configured instead. I believe at the time it was introduced there were no "module" classpath attribute.

trancexpress: And? Projects like that will exist, same as during the tests. Just because its a hack doesn't mean its not being used.

iloveeclipse: Simeon, you didn't get it. I mean, this container looks like a hack for JDT tests, nothing more. It will be NOT exist in any code outside JDT.

trancexpress: Hi @mpalat ,

can you clarify the purpose and usage of `JavaCore.MODULE_PATH_CONTAINER_ID`/`ModulePathContainer`?

Was this some hack to test the compiler module path with Eclipse projects? Can we expect that projects having a classpath entry as follows, are only used in tests and don't exist as actual projects in "real" workspaces?

```
<classpathentry kind="con" path="org.eclipse.jdt.MODULE_PATH"/>
```

If this was added for tests only, we can adjust the tests that have problems with the change (to have a proper module path).

Best regards and thanks,
Simeon

trancexpress: I've prepared a 2nd branch with how tests would look without using `JavaCore.MODULE_PATH_CONTAINER_ID`:

https://github.com/trancexpress/eclipse.jdt.core/commit/404f642d03eb2f1e32013552867a24d595db176a

Its not a small change.

iloveeclipse: > Its not a small change.

In that case we should separate the change with the actual fix from the patch that changes the usage of "MODULE_PATH" container.

Regarding https://github.com/trancexpress/eclipse.jdt.core/commit/404f642d03eb2f1e32013552867a24d595db176a : I'm not sure I can follow. The "MODULE_PATH" seem to be a shortcut to include all required modules of a project module, the patch I see in the branch seem to simply add the "module" attribute to all classpath entries, not the required modules. `org.eclipse.jdt.internal.core.ModulePathContainer.getClasspathEntries()`.

So I don't understand what exactly breaks with the original patch in "MODULE_PATH" and how the second patch supposed to fix that.

Would be nice first to explain the fail and discuss the right solution after that?

trancexpress: Opened #717 to check if/why we need this entry in tests.

> Would be nice first to explain the fail and discuss the right solution after that?

The new fails are due to expecting the `module` attribute for classpath entries on the module path, in order to add module information during the search. The tests that fail require having classpath entries on the module path, but don't explicitly make those entries modular. This worked, since the search environment adds module infos for every classpath entry (assuming JLS > 9). Now it no longer works.

trancexpress:
Some fails in `JavaSearchBugs9Tests` remain, I'm not sure how to fix those. I think we have a general problem with source classpath entries not having the module attribute. Since they don't have the attribute, we would not add the module info... But its possible that the project has a module path and everything is on the module path. I don't know what JDT does to ensure the sources are also on the module path for the compile...

trancexpress:
> So we have to figure out what is the best way to fix them.

That is simple, we can adjust the change to not exclude source classpath entries.

**tl;dr** I don't see a reliable way to see if a source folder of a project is on the module path or on the classpath, during the search

However the new test fails. We can of course change the new test to use a library, same as the original problem we ran into. Of course the case where actual referenced projects are on the classpath (and not on the module path) will have the same problem.

Unfortunately the package fragment roots that pass through the changed code are the source folders of each individual project. I.e. if I have this structure:

```
project1 -> depends on project2
    src
project2
    src
```

What I see is `src` from project1` and `src` from `project2`. Neither of which has the `module` attribute. I.e. I see 2 classpath entries of 2 different projects:

```
<classpathentry kind="src" path="src"/>
```
```
<classpathentry kind="src" path="src"/>
```

If instead the environment was created with `src` from project` and `src` from the dependency to `src` from `project2`, there would be something to work with. Since the latter has the `module` attribute if its on the module path; we can rely on its presence or absence:

```
	<classpathentry kind="src" path="src"/>
	<classpathentry kind="src" path="/project2"/>
```

I'm not sure what we can do about this... In theory, we can examine the projects referencing the current project, checking their whether the reference is modular or not... But that assumes there is 1 "main" project with which to judge.

trancexpress:
Now the new test fails. But we can rewrite it to use jars instead of project references and open a new issue to try to fix the case of referenced projects. I really don't know what to do for that case. Without having a "main" project from which the search is initiated there is nothing to be done I think. I'll see if the search environment knows from where it was initiated...

We might want to define `JavaSearchNameEnvironment.isModularProject(IJavaProject)` somewhere where `JavaRuntime` in JDT debug can use it.

trancexpress:
@iloveeclipse see my last 2 comments, I think we have to discuss how to proceed here. I was naive to think the `module` attribute is always present. It isn't for source folders of projects containing those source folders. This complicates things a lot.

We can focus on fixing the initial case, problems in libraries (for those the `module` attribute is set).

iloveeclipse:
`JavaSearchNameEnvironment.addModuleClassPathInfo(PackageFragmentRoot, IModuleDescription, ClasspathLocation)` asks for module descriptor and if that is present on a **library** entry, we should check "module" attribute, not before. So new logik should go a bit later => no problems with source folders anymore.

trancexpress:
> `JavaSearchNameEnvironment.addModuleClassPathInfo(PackageFragmentRoot, IModuleDescription, ClasspathLocation)` asks for module descriptor and if that is present on a **library** entry, we should check "module" attribute, not before. So new logik should go a bit later => no problems with source folders anymore.

The new test will still fail. It shows the problem with source folders, not libraries. I.e. as said above, we'll fix only a part of the problem.

trancexpress:
I've adjusted the new test to use libraries and not referenced projects.

> `JavaSearchNameEnvironment.addModuleClassPathInfo(PackageFragmentRoot, IModuleDescription, ClasspathLocation)` asks for module descriptor and if that is present on a **library** entry, we should check "module" attribute, not before. So new logik should go a bit later

I didn't do this since there are many `addModuleClassPathInfo()` methods doing different things based on module information available here and there. From my POV better to add code in one place than all the places in the different `addModuleClassPathInfo()` methods.

trancexpress:
I've created an extra branch, to showcase what I meant with my previous comments:

https://github.com/trancexpress/eclipse.jdt.core/commit/4eaa334e6a5104f5f1595e22da8ab01240896aae

If we want to fix also the case for referenced projects, we can likely refine that commit. If we pass the project from which the search was started properly (currently its the first element in the passed `ArrayList` of projects, I'm using this to showcase the bug fix), maybe its worthwhile to make that change as well.

It would be a bigger change though. And as I've commented, I'm not 100% sure how valid this is. For the example in the new test its valid to do this, since the type used in the search is defined in the project. But if the search is initiated for a type found in a library, I'm not sure its OK to do this... Since the library can be on the classpath of any number of projects, none of which would be the "main" one then...

trancexpress:
> I've created an extra branch, to showcase what I meant with my previous comments:
> 
> [trancexpress@4eaa334](https://github.com/trancexpress/eclipse.jdt.core/commit/4eaa334e6a5104f5f1595e22da8ab01240896aae)
> 
> If we want to fix also the case for referenced projects, we can likely refine that commit. If we pass the project from which the search was started properly (currently its the first element in the passed `ArrayList` of projects, I'm using this to showcase the bug fix), maybe its worthwhile to make that change as well.

I guess another option would be to check which projects are on the classpath and which are on the module path, in `MatchLocator`. Then pass this information to `JavaSearchNameEnvironment`. Then when we see a source entry from a project, we would know if its on the module path.

We can maybe do this only for searches where the type has a source compilation unit. That way we are sure the search really is about the project where the search started.

rgrunber:
Is there any way this change might be related to type signature resolution of local variable type signatures ?

We recently started seeing test failures in a project that uses the JDT Core API, https://ci.eclipse.org/lsp4mp/job/lsp4mp/job/master/211/testReport/ . If I import jdt.core into my workspace (as opposed to from the target platform), the test failures begin happening at this commit.

If I take a `LocalVariable` element (eg. whose type is `String`) and calls `getTypeSignature()`. The result used to be `Ljava.lang.String` whereas now it is `LString`. I can't find any example/references to `LString` as a valid signature. It should either be `QString`, or `Ljava.lang.String`.

iloveeclipse:
There was a regression, see https://github.com/eclipse-jdt/eclipse.jdt.core/issues/740, and the fix was just merged, see https://github.com/eclipse-jdt/eclipse.jdt.core/pull/746.

So if that was the problematic commit, please check next I build.

iloveeclipse:
@rgrunber : I20230214-1800 build should contain the fix. Please test.

iloveeclipse:
Thanks Roland!

PR #694, opened 2023-02-03
Enhance the BuildContext with the produced output class

laeubi:
## What it does
Currently a CompilationParticipant can't get the produced class files to
further process them.

This adds new method to access the generated classfiles name, location
and bytes.

## How to test

A new test method has been added.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/core/compiler/IClassContent.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/core/compiler/IClassContent.java
@@ -1,0 +1,50 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Christoph Läubrich and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Christoph Läubrich - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.jdt.core.compiler;
+
+import org.eclipse.core.runtime.CoreException;
+
+/**
+ * Encapsulates the produced class of the compile operation
+ * @since 3.35
+ * @noimplement This interface is not intended to be implemented by clients.
+ */
+public interface IClassContent {
+
+	/**
+	 * Provides access to the raw class bytes of this class content.
+	 *
+	 * @see #setBytes(byte[]) for how to update the class bytes
+	 * @return the class bytes of this class content or <code>null</code> if no content is avaiable
+	 * @throws CoreException
+	 *             if gathering the bytes failed
+	 */
+	byte[] getBytes() throws CoreException;
+
+	/**
+	 * Set the class bytes to the provided array
+	 *
+	 * @see #getBytes() for gathering the current class bytes
+	 * @param classBytes
+	 * @throws CoreException
+	 */
+	void setBytes(byte[] classBytes) throws CoreException;
+
+	/**
+	 * Describe the relative filename of the class that could be used to store the contetn for example in a jar file or on disk.
+	 *
+	 * @return the filename of the class as represented for example in a jar file
+	 */
+	String getFileName();
+}
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/CompilationParticipantResult.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/CompilationParticipantResult.java
@@ -28,6 +28,7 @@ public class CompilationParticipantResult extends BuildContext {
 	protected CategorizedProblem[] problems; // new problems to report against this compilationUnit
 	protected String[] dependencies; // fully-qualified type names of any new dependencies, each name is of the form 'p1.p2.A.B'
 	private boolean isTestCode;
+	private ClassContent[] classfiles;
 
 protected CompilationParticipantResult(SourceFile sourceFile, boolean isTestCode) {
 	this.sourceFile = sourceFile;
@@ -47,6 +48,14 @@ protected CompilationParticipantResult(SourceFile sourceFile, boolean isTestCode
 @Override
 public char[] getContents() {
 	return this.sourceFile.getContents();
+}
+
+@Override
+public IClassContent[] getClassContent() {
+	if (this.classfiles == null) {
+		return super.getClassContent();
+	}
+	return this.classfiles.clone();
 }
 
 /**
@@ -171,6 +180,10 @@ void reset(AnnotationBinding[] newAnnotations) {
 	this.deletedFiles = null;
 	this.problems = null;
 	this.dependencies = null;
+}
+
+void setClassFiles(ClassContent[] classfiles) {
+	this.classfiles = classfiles;
 }
 
 @Override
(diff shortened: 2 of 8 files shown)


org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/CompilationParticipantResult.java
```diff
@@ -143,9 +173,19 @@ public void recordNewProblems(CategorizedProblem[] newProblems) {
...
-	this.hasAnnotations = detectedAnnotations;
+	this.annotations = annotations;
+	this.classfiles = null;
+	this.addedFiles = null;
+	this.deletedFiles = null;
+	this.problems = null;
+	this.dependencies = null;
+}
+
+void reset(ClassContent[] classfiles) {
```
szarnekow, line 186: I haven't read it in depth yet, but I wonder if `reset(AnnotationBinding[])` should differently from `reset(ClassContent[]).

If the participant is interested in producing more artifacts based on the annotations + class content, it would need to maintain some internal state about the build context. I suspect reset(ClassContent[]) should not discard all the other state as opposed to reset(AnnotationBinding[]) which is called at the beginning of the compilation and should discard prior state.

laeubi: `reset(AnnotationBinding[])` is called *after* the compilation, and at the moment the compile method is left then (so no further processing possible), `reset(ClassContent[])` is just called after that, I just though that if when annotation processing resets the previous state, class processing should do the same, but as the results are getting out of scope afterwards it might be enough to just reset the annotations (that are only valid when calling), or one could just allow access to these as well, I would be fine with anything here :-)

One might even make the class data available already in the `processAnnotations`, just let me know.

szarnekow: You're absolutely right, I misinterpreted the API - should have looked at the code first :)

> One might even make the class data available already in the processAnnotations, just let me know.

I think it does make sense. Is the idea to allow bytecode transformation from the processor?

laeubi: I need to process the classes afterwards and access the bytes, currently it is quite inconvenient as:

1. A source file can produce multiple class files, so simply take the name to find the output does not work
2. I need to know where the class files get written

Poor-Man-Soloution is to resolve the classpath in `buildFinished`, collect all output folders, read all directories and guess that if it starts with something like `SourceFileName` and ends with `.class`, then read the file contents... that seems really vastfull and error prone given that JDT already has gathered the data :-)

szarnekow: True, and it wouldn't be correct either since package private classes don't need to have the same name as the source file.

So modifications to the generated bytecode are a non-goal for this?
I was thinking about the following use-case: Add a compiler participant that understands non-null annotations and adds runtime assertions to the bytecode.

laeubi: I guess the idea can be enhanced for sure, just keep in mind that currently the class-file is already written to disk, but it would be quite easy to have another method `setBytes(byte[])` that writes new content to the class-file location, if you think it is useful I can add such method as well.

laeubi: @szarnekow I now added a `setBytes` method that also removes some code duplication in JDT itself.

laeubi: @szarnekow I have now changed this so the class content is already set in the `processAnnotations` phase and not reseting the state.

laeubi:
By the way this also includes the changes from this PR:
- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/676

maybe you want to review this as well?

laeubi:
@szarnekow @iloveeclipse @stephan-herrmann any chance to make progress here?

szarnekow:
> @szarnekow @iloveeclipse @stephan-herrmann any chance to make progress here?

Not much that I can do since I'm not a committer.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/CompilationParticipantResult.java
szarnekow: FTR: I think this is reasonable.

org.eclipse.jdt.core/model/org/eclipse/jdt/core/compiler/IClassContent.java
```diff
@@ -0,0 +1,48 @@
...
+ * @noimplement This interface is not intended to be implemented by clients.
+ */
+public interface IClassContent {
+
+	/**
+	 * @return the class bytes of this class content or <code>null</code> if no content was written
+	 * @throws CoreException if gathering the bytes failed
+	 */
+	byte[] getBytes() throws CoreException;
+
```
jarthana, line 31: To me, readBytes() sounds better if reading from a stream is involved here. It makes it more apparent of what happens.

org.eclipse.jdt.core/model/org/eclipse/jdt/core/compiler/IClassContent.java
```diff
@@ -0,0 +1,48 @@
...
+	/**
+	 * Set the given classbytes to the provided array
+	 * @param classBytes
+	 * @throws CoreException
+	 */
+	void setBytes(byte[] classBytes) throws CoreException;
```
jarthana, line 37: Same as getBytes(), name similar to writeBytes() makes it apparent.
Having said that, who, other than the compiler is going to be using this method? What's the need to expose this to the outside world? If someone gets hold of an IClassContent and invoke this method, what is supposed to happen?

jarthana (changes requested):

org.eclipse.jdt.core/model/org/eclipse/jdt/core/compiler/IClassContent.java
laeubi: The content is not necessarily read from a stream, it might be cached or gathers from somewhere else (see setBytes), so probably I should adjust the javadoc to not contain "written" but "is not available"

org.eclipse.jdt.core/model/org/eclipse/jdt/core/compiler/IClassContent.java
laeubi: See above, the bytes are not necessarily written as a consequence of this call (they might finally end up on disk even though).
The idea for this method is that for example a class-weaving function can get the bytes, do some weaving and later on update the bytes for further processing by other participants.

laeubi (commented):
Thanks for looking into it, see comments above.

laeubi:
@jarthana I have now rebased  / squashed everything, adjusted the javadoc and removed the "file-nature" from the interface as it is an implementation detail here.

jarthana:
OK, I am happy with the explanation provided. But I would like to add some documentation on IClassContent on how the clients of the API are expected it to use or what they can expect. For e.g, we should specify that the compilation participants may pass these around among one another. And that setBytes() does not have any impact on the compiler whatsoever or something along those lines. It's better if we set the expectation in the beginning itself to avoid confusion later.

laeubi:
Yes I need to review the changes already made here previously by https://github.com/eclipse-jdt/eclipse.jdt.core/pull/984

jarthana:
I am off for the next two days. I can take a quick look over the weekend or Monday if the patch is up-to-date and all green. If any other committers are interested in taking this forward, welcome to do so.

PR #705, opened 2023-02-05
Completions for switch expressions inside lambda.Fix #697

### LINKED ISSUE #697: No completions for switch case labels in switch expressions inside lambda
Take the following sample

```
	static Map<Thread.State, String> toMap(){
		return Stream.of(Thread.State.values()).collect(Collectors.toMap(Function.identity(), state -> {
			return switch (state) {
				case |
			};		
		}));
	}
```

Try to complete code near `|`, you don't get any completions

gayanper:
The fix tries to correct the parser snapshot position based on the current lambda block the scanner is at.

## What it does
Fix completions for switch labels in switch expression inside lambda block.

## How to test
UnitTest and code example in the issue #697 
## Author checklist


gayanper:
@stephan-herrmann here is a test and some partial changes I did to solve the consumeBlock assertion error.

gayanper:
@stephan-herrmann please check the second commit specially, I really doubt if its the best fix to get the case expression inside the already found switch expression.

stephan-herrmann:
@gayanper your changes led me to make a funny observation: my changes in https://git.eclipse.org/r/c/jdt/eclipse.jdt.core/+/125001 somehow entered through the backdoor, such that I believed `commit()` should match snapshots according to `this.scanner.startPosition`. Only [towards the end of that gerrit](https://git.eclipse.org/r/c/jdt/eclipse.jdt.core/+/125001/9..10/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/impl/AssistParser.java) I noticed a case, where I needed to use `this.blockStarts[this.realBlockPtr]` instead.

What I didn't check back then: do we ever *need* to use `this.scanner.startPosition`? The answer seems to be: *no*.

I quickly replaced the increasingly complex impl of `commit(boolean)` with this:
```
void commit(boolean isStart) {
	int newSnapShotPosition = this.blockStarts[this.realBlockPtr];
	if (this.snapShotPtr == -1 || newSnapShotPosition != this.snapShotPositions[this.snapShotPtr]) {
		// first commit, or existing snapshot doesn't match
		addNewSnapShot(newSnapShotPosition);
	}
	this.snapShotStack[this.snapShotPtr].copyState(this);
}
```
With this `RunCompletionParserTests` and `RunCompletionModelTests` still run green.

And we could even remove the `boolean isStart` argument.

I haven't yet replayed my line of reasoning in https://bugs.eclipse.org/535743, so this simplification still needs some double checking against original intentions, but generally speaking this looks promising, if you ask me :)

The asserts you mentioned are an indication that I was afraid of bugs in stack management, and indeed, the current bug is exactly of that kind.

I haven't yet checked your changes in `CompletionParser`. The co-existence of a `SwitchStatement` and a `SwitchExpression` for the same text looked weird at first, but I can't yet tell if this is sign of a deeper bug or acceptable.

gayanper:
@stephan-herrmann let me know the CompletionParser changes as well, then i can try to improve both Completion and Assist Parsers together and run tests.

gayanper:
@stephan-herrmann I improved the fix to support more scenarios and also applies the changes you suggested to `commit` method.

jarthana:
I reran the job but it still fail. I don't know why. The calendar correctly marks Stabilization only from tomorrow.
Gayan, are you trying this for RC1?

gayanper:
> I reran the job but it still fail. I don't know why. The calendar correctly marks Stabilization only from tomorrow.
> Gayan, are you trying this for RC1?

No @jarthana this needs one more improvement, Lets plan this for next release.

gayanper:
@stephan-herrmann need some help on the AIOB error getting for `testGH697_CompletionOnSwitchExpressionVariable` from the scanner.

gayanper:
@stephan-herrmann seems like removing the AIOBE from the scanner does seems to solve the issue. Its on my latest commit.

gayanper:
@jarthana @stephan-herrmann I also found another scenario, `testGH697_CompletionOnSwitchExpressionAsFirstParameter` seems like the type parameters are not getting resolved since we have not parsed the whole MessageSend, the last parameter is important.

To avoid issue like this, could we just force the parser to parse the whole statement where the completion happens even if we find syntax errors ?

gayanper:
@stephan-herrmann can you check on the changes I have done after last you provided some feedback. I was running this change with my daily usage and found few issues and improved, But i'm not sure if I have taken the correct path.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
```diff
@@ -1395,7 +1395,8 @@ public int scanIdentifier() throws InvalidInputException {
 }
 public void ungetToken(int unambiguousToken) {
 	if (this.nextToken != TokenNameNotAToken) {
-		throw new ArrayIndexOutOfBoundsException("Single cell array overflow"); //$NON-NLS-1$
+		return;
+		// throw new ArrayIndexOutOfBoundsException("Single cell array overflow"); //$NON-NLS-1$
```
stephan-herrmann, line 1399: I'm not quite comfortable with this change. I can only assume that the exception is thrown because that situation _should never happen_, and we want to be alerted, so we can fix this some other way.

The conflict seems to arise because `fallBackToSpringForward()` is invoked twice without making real progreess. The first invocation sets `this.currentToken = TokenNameSEMICOLON` and the second invocation tries to push that exact token back into the scanner (where it actually never was observed).

Why is it invoked twice to begin with?

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/complete/CompletionParser.java
```diff
@@ -6373,7 +6383,12 @@ private boolean foundToken(int token) {
 @Override
 protected int actFromTokenOrSynthetic(int previousAct) {
 	int newAct = tAction(previousAct, this.currentToken);
-	if (this.hasError && !this.diet && newAct == ERROR_ACTION && this.scanner.currentPosition > this.cursorLocation) {
+	if (this.hasError
+			&& (!this.diet || 
+			(topKnownElementKind(COMPLETION_OR_ASSIST_PARSER) == K_BETWEEN_CASE_AND_COLON && 
+			isIndirectlyInsideFieldInitialization())) // diet or a complex field init with switch expression
+			&& newAct == ERROR_ACTION
+			&& this.scanner.currentPosition > this.cursorLocation) {
```
stephan-herrmann, line 6391: This change I haven't yet looked at.

stephan-herrmann (commented):
I'm sorry I don't have definite answers (yet), only vague concerns and questions.

Perhaps you can first answer, whether or not the intermediate lambda is essential for the problem you are fixing?

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/complete/CompletionParser.java
```diff
@@ -6099,7 +6103,13 @@ protected int resumeAfterRecovery() {
 	if (this.assistNode != null) {
 
 		if (requireExtendedRecovery()) {
-			if (this.unstackedAct != ERROR_ACTION) {
+			if (this.unstackedAct != ERROR_ACTION
+					// resume also if we are inside a MessageSend to complete collecting all paramaters
+					// when using generics, it helps for proper resolution of the MessageSend when all
+					// paramaters are collected. This is essential when next few paramaters are the ones
+					// that decide a TypeParameter of a current completing paramater type.
+					|| (topKnownElementKind(COMPLETION_OR_ASSIST_PARSER) == K_BETWEEN_CASE_AND_COLON
```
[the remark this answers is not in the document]
gayanper: To be honest I'm not proud of this condition, it's just trail and error to get the parser to parts the whole invocation so that the lamda variable is resolved to have proper completions inside switch expressions. Thats one reason i want to get feedback from you, because I'm sure there should be a more elegant way to handle this.

I do agree, if we generlize this behavior or could generalize it, so that when we are completing on a method invocation which has type parameters, if we could continue until the invocation expression ends, it could benifit other cases as well isn't it ?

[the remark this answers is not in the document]
gayanper: > Seeing this code change under the condition of `requireExtendedRecovery()` I wonder if all this is indeed specific to the situation of a switch expression inside a lambda (which it what requireExtendedRecovery tests)? Or would the same tweaks also be necessary when the switch itself (i.e., its value) is passed into a method invocation?

No for normal method invocation, the current implementation works. Isn't only when inside lambda.

gayanper:
@stephan-herrmann I think your questions led me to simplify the solution with some additional recovery tokens.

gayanper:
@stephan-herrmann can you have a look at the new changes and let me know if the changes looks good ?

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/impl/AssistParser.java
```diff
@@ -142,7 +142,7 @@ public abstract class AssistParser extends Parser {
 
 	public int cursorLocation = Integer.MAX_VALUE;
 
-	protected static final int[] RECOVERY_TOKENS = { TokenNameSEMICOLON, TokenNameRPAREN, TokenNameRBRACE, TokenNameRBRACKET};
+	protected static final int[] RECOVERY_TOKENS = { TokenNameSEMICOLON, TokenNameRPAREN, TokenNameRBRACE, TokenNameRBRACKET, TokenNameARROW, TokenNameCOLON};
```
stephan-herrmann, line 145: can you provide a test case needing `TokenNameARROW`? I couldn't find any.

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/impl/AssistParser.java
```diff
@@ -1942,6 +1942,22 @@ protected boolean isInsideMethod(){
...
+		switch (this.elementKindStack[i]) {
+			case K_TYPE_DELIMITER : return false;
+			case K_METHOD_DELIMITER : return false;
+			case K_FIELD_INITIALIZER_DELIMITER : return false;
+			case K_SELECTOR: return true;
+		}
+		i--;
+	}
+	return false;
+}
```
stephan-herrmann, line 1960: no longer used, right? :)

stephan-herrmann (approved):
You found a very lean solution here. Congrats!

I dropped just some final nitpicks, questions, ponderings ...

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/complete/CompletionParser.java
```diff
@@ -2698,6 +2698,9 @@ protected void consumeSwitchExpression() {
 	if (this.assistNode != null) {
 		SwitchExpression expr = (SwitchExpression) this.expressionStack[this.expressionPtr];
 		expr.resolveAll = true;
+		if (this.assistNodeParent instanceof SwitchStatement) {
+			expr.statements = ((SwitchStatement) this.assistNodeParent).statements;
+		}
```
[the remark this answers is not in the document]
gayanper: Yes seems we don't need that anymore.

Issue #722, opened 2023-02-10
ambigious call for jdk >=13


jukzi:
in org.eclipse.jdt.internal.compiler.util.Util.getZipEntryClassFile(String, String)
![image](https://user-images.githubusercontent.com/51790620/218073019-36100b68-519d-4ee2-8058-5e8cf27a2eca.png)
@jarthana 
you can use `FileSystems.newFileSystem(zipfile, (ClassLoader) null)`

java.nio.file.FileSystems.newFileSystem(Path, Map<String, ?>) was added in 13

jarthana:
This happens when BREE 11 is mapped to a newer JDK. Now, if I change the code to make it work with JDK > 13, we might get a warning about an unused type cast. May be we can take this up when we move up to Java 11, eventually.

jarthana:
> This happens when BREE 11 is mapped to a newer JDK. Now, if I change the code to make it work with JDK > 13, we might get a warning about an unused type cast. May be we can take this up when we move up to Java 11, eventually.

OK, I was wrong about the warning part. I don't see a warning about the cast. I will fix this for M3.

iloveeclipse:
> I will fix this for M3.

Jay, that is already fixed.

PR #727, opened 2023-02-11
Don't use FileSystem just to read zip file entries

iloveeclipse:
Using a FileSystem is not correct way to read zip file entries in multi-threaded environment, because FileSystem is a global object and may be already created or closed while we are reading from it, and vice versa.

Beside this, this change fixes
org.eclipse.jdt.ui.tests.leaks.JavaLeakTest failing because of the unexpected use of ZipFileSystem on opening JDT editor and compilation error if compiling JDT with Java 17.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/722


jukzi:
thanks @iloveeclipse

PR #728, opened 2023-02-11
Tests should detect duplicate problem logs - Fixes #710

### LINKED ISSUE #710: Tests should detect duplicate problem logs
During #707 I found that our compiler tests don't detect when the exact same problem is reported twice. When it happens, it is an indication about unclear responsibilities among different parts of the implementation and should therefor be avoided.

With such detection enabled, I noticed that several more situations of redundant null annotations do contribute duplicate problems, which should be cleaned-up as well, thus hardening #683.

stephan-herrmann:
+ Includes cleanup of where @NonNullByDefault should be checked
+ draftily add RecordComponentBinding.fillInDefaultNonNullness()
+ remove duplicate reporting of some record related problems

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ASTNode.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ASTNode.java
@@ -56,7 +56,6 @@ import java.util.Map;
 
 import org.eclipse.jdt.core.compiler.CharOperation;
 import org.eclipse.jdt.internal.compiler.ASTVisitor;
-import org.eclipse.jdt.internal.compiler.ast.TypeReference.AnnotationPosition;
 import org.eclipse.jdt.internal.compiler.classfmt.ClassFileConstants;
 import org.eclipse.jdt.internal.compiler.env.AccessRestriction;
 import org.eclipse.jdt.internal.compiler.lookup.AnnotationBinding;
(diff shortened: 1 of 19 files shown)


stephan-herrmann:
@mpalat I noticed that some record related problems were reported redundantly from 2 different code locations. As this affects some code written by you, do you have a minute to check if my changes broke anything?

Relevant locations:
* `Parser.convertToFields()`: don't call `recordComponentCannotBeVoid()`, will be done later by `SourceTypeBinding.resolveTypeFor(RecordComponentBinding)`.
  * reporting semantic errors during resolving looks safer to me, in particular in light of syntax recovery, which might invoke the same parser methods more than once.
* `SourceTypeBinding.resolveTypesWithSuspendedTempErrorHandlingPolicy(MethodBinding)`: invoke `checkAndFlagHeapPollution()` only conditionally, like it was already done for `argumentTypeCannotBeVoid()`.

I also saw null related code being tricked by the duplication of record component & field, see `FieldBinding.fillInDefaultNonNullness()`. Similar for `MethodBinding.fillInDefaultNonNullness18` regarding implicit constructors.

What's your general idea about dividing responsibilities of the various parts generated from a record header?

stephan-herrmann:
I noticed several more tests triggering duplicate problem entries, and so for now I made the test infrastructure to complain about duplicate entries only if not explicitly expected.

Some tests already had duplication in their expectations, so nothing changed for them atm:
* `AnnotationTest.testBug376429a()`
* `NullTypeAnnotationTest.testBug484926redundantNNBD()`
* `NegativeLambdaExpressionsTest.testUnderScoreParameter()`
* `JEP323VarLambdaParamsTest.testBug534787_negative_001()`
* `SwitchExpressionsYieldTest.testBug544073_020()`
* `SwitchExpressionsYieldTest.testBug544073_079()`

@jarthana with this strategy I "excused" two of your tests by making duplicate problems explicit. See [5201c03](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/728/commits/5201c0316ee49a6aa76c537cf1133363c3b502ed). At a quick glance I found repeated invocations of `TypePattern.collectPatternVariablesToScope()` slightly confusing, in particular relating to the "side effect" of error reporting. Is this something you are aware of?

Issue #740, opened 2023-02-14
Call Hierarchy and references search don't see obvious references


iloveeclipse:
I'm baffled seeing following, using I20230213-1800 build with latest master state of related projects.

Preconditions: workspace with `org.eclipse.swt` and `org.eclipse.ui.ide.application` projects checked out.

1) Try to open Call Hierarchy on `org.eclipse.swt.graphics.Resource.nonDisposedReporter` field: **zero matches**.
2) Try to open Call Hierarchy on `org.eclipse.swt.graphics.Resource.setNonDisposeHandler(Consumer<Error>)` method: **zero matches**.
3) Try to search for references on `org.eclipse.swt.graphics.Resource.nonDisposedReporter` field: : **zero matches**
4) Try to search for **write** references on `org.eclipse.swt.graphics.Resource.nonDisposedReporter` field: : **zero matches**
5) Try to search for references on `org.eclipse.swt.graphics.Resource.setNonDisposeHandler(Consumer<Error>)` method: two **potential** matches.

![image](https://user-images.githubusercontent.com/964108/218682322-c0d31af4-e7fd-4622-b37e-4432df19163f.png)

This is insane! Sure there are **direct** matches for both the field and method!

Just to be sure, I've rebuilt JDT index and recompiled projects in question - with no change.

**Updated steps to reproduce**

1) Create java project referencing Java 9+ **without** module info file.
2) Change .classpath to remove "modular" attribute from container, like
```
<?xml version="1.0" encoding="UTF-8"?>
<classpath>
	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-17"/>
	<classpathentry kind="src" path="src"/>
	<classpathentry kind="output" path="bin"/>
</classpath>
```
3) Paste this to the source folder:
```
import java.util.function.Consumer;
public class Z {
	private static Consumer<Error> field;
	private static class Inner {
		static {
			setField(e -> {});
		}
	}
	public static void setField(Consumer<Error> c) {
		field = c;
	}
}
```
Try Call Hierarchy or references search on `field` or `setField()`.

trancexpress:
I wonder if we (or someone else) broke something recently?

I do see a call hierarchy for `Resource.nonDisposedReporter` with `I20230207-1800`. But not with `I20230211-0330`...

iloveeclipse:
[...]

PR #746, opened 2023-02-14
Call Hierarchy and references search don't see obvious references

trancexpress:
The fix for #675 neglects the fact that a classpath JRE 9+ is on the module path despite not having a classpath entry to indicate this. This change disables the change for classpath container entries.

Fixes: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/740

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchBugs9Tests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchBugs9Tests.java
@@ -25,7 +25,11 @@ import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.Path;
+import org.eclipse.jdt.core.IClasspathAttribute;
+import org.eclipse.jdt.core.IClasspathContainer;
+import org.eclipse.jdt.core.IClasspathEntry;
 import org.eclipse.jdt.core.ICompilationUnit;
+import org.eclipse.jdt.core.IField;
 import org.eclipse.jdt.core.IJavaElement;
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.ILocalVariable;
@@ -44,6 +48,9 @@ import org.eclipse.jdt.core.search.SearchMatch;
 import org.eclipse.jdt.core.search.SearchParticipant;
 import org.eclipse.jdt.core.search.SearchPattern;
 import org.eclipse.jdt.core.search.TypeReferenceMatch;
+import org.eclipse.jdt.core.tests.model.ClasspathTests.TestContainer;
+import org.eclipse.jdt.core.tests.util.Util;
+import org.eclipse.jdt.internal.core.ClasspathEntry;
 import org.eclipse.jdt.internal.core.JavaElement;
 import org.eclipse.jdt.internal.core.LocalVariable;
 import org.eclipse.jdt.internal.core.TypeParameter;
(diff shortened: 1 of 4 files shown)


trancexpress:
I'll see if I can add a test case.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/matching/JavaSearchNameEnvironment.java
```diff
@@ -582,6 +568,30 @@ private static boolean isComplianceJava9OrHigher(IJavaProject javaProject) {
...
+	try {
+		IClasspathEntry classpathEntry = root.getRawClasspathEntry();
+		if (classpathEntry.getEntryKind() == IClasspathEntry.CPE_SOURCE) {
+			/*
+			 * Source classpath entries of a project never have the module attribute set,
+			 * so we cannot rely on the attribute.
+			 */
+			isOnModulePath = isModularProject(javaProject);
+		/*
+		 * The JRE container is on the module path based on its version,
```
iloveeclipse, line 582: Note, this is not correct assumption and comment. There can be different containers on classpath, not necessarily `JRE_CONTAINER`. I first was tempted to say we should check `if (classpathEnty instanceof IModulePathEntry)` but it seems that `IModulePathEntry` can also represent non-modular code.

So far I see no reliable way to detect if some container entry is on module path...

May be we could use given `PackageFragmentRoot`. Obviously, if that is `instanceof JrtPackageFragmentRoot` it will be always modular. 

But I'm not sure what kind of other `PackageFragmentRoot` objects can be contributed by `CPE_CONTAINER` entries that have no "modular" attribute but still host `PackageFragmentRoot` from modular world.

iloveeclipse (changes requested):

trancexpress:
> I'll see if I can add a test case.

Seems to not be simple, since the tests add the JRE without a container:

```
<?xml version="1.0" encoding="UTF-8"?>
<classpath>
        <classpathentry kind="src" path="src"/>
        <classpathentry kind="var" path="JCL11_LIB">
                <attributes>
                        <attribute name="module" value="true"/>
                </attributes>
        </classpathentry>
        <classpathentry kind="output" path="bin"/>
</classpath>
```

trancexpress:
I added a test, lets clarify the review questions and we can merge.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/matching/JavaSearchNameEnvironment.java
iloveeclipse: @trancexpress : the comment hast to be changed, it is not correct.

Something like "we have no idea what the container provides, so we assume it contributes something to module path"

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchBugs9Tests.java
```diff
@@ -4747,5 +4754,52 @@ public void testBug547095_type_patter_search_modular() throws Exception {
 	}
 }
 
+/*
+ * Test for call hierarchy of a field with a type parameter coming from a JRE module.
```
iloveeclipse, line 4758: Note: the test **example** doesn't match the comment and method name. This is extremely confusing.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/matching/JavaSearchNameEnvironment.java
iloveeclipse: Also please move the comment to the **corresponding** if case,

Issue #748, opened 2023-02-14
Handle JRE container when adding module class path info in Java search


trancexpress:
See #740 and the fix #746. Due to the missing module attribute for the JRE container (despite the JRE container being on the module path), we are forced to assume every container is on the module path.

We should look into the special handling JDT does for the JRE container and how to make the fix in #746 apply only for such containers (that JDT adds to the module path without a module attribute).

trancexpress:
The build path page decides the JRE is on the module path with this stack trace:

```
"main" #1 prio=6 os_prio=0 cpu=7618.77ms elapsed=2454.16s tid=0x00007f6560015440 nid=0x3e19 runnable  [0x00007f65647fc000]
   java.lang.Thread.State: RUNNABLE
        at org.eclipse.jdt.internal.ui.wizards.buildpaths.CPListElement.addModuleNodes(CPListElement.java:214)
        at org.eclipse.jdt.internal.ui.wizards.buildpaths.CPListElement.<init>(CPListElement.java:178)
        at org.eclipse.jdt.internal.ui.wizards.buildpaths.CPListElement.create(CPListElement.java:772)
        at org.eclipse.jdt.internal.ui.wizards.buildpaths.CPListElement.create(CPListElement.java:710)
        at org.eclipse.jdt.internal.ui.wizards.buildpaths.BuildPathsBlock.getCPListElements(BuildPathsBlock.java:446)
        at org.eclipse.jdt.internal.ui.wizards.buildpaths.BuildPathsBlock.init(BuildPathsBlock.java:341)
        at org.eclipse.jdt.internal.ui.preferences.BuildPathsPropertyPage.createWithJava(BuildPathsPropertyPage.java:170)
        at org.eclipse.jdt.internal.ui.preferences.BuildPathsPropertyPage.createContents(BuildPathsPropertyPage.java:88)
        at org.eclipse.jface.preference.PreferencePage.createControl(PreferencePage.java:244)
        at org.eclipse.jdt.internal.ui.preferences.BuildPathsPropertyPage.createControl(BuildPathsPropertyPage.java:99)
        at org.eclipse.jface.preference.PreferenceDialog.createPageControl(PreferenceDialog.java:1434)
        at org.eclipse.jface.preference.PreferenceDialog$8.run(PreferenceDialog.java:1197)
```

With this code:

```
IModuleDescription module= fragmentRoots[0].getModuleDescription();
if (module != null && module.isSystemModule())
[...]

PR #751, opened 2023-02-15
Handle JRE container when adding module class path info in Java search

trancexpress:
The JRE container on a projects classpath doesn't always have the module attribute set, despite being on the module path. Due to not considering this, the fix for #675 introduces a bug - types in the JRE modules are not resolved correctly.

The fix for #740 disables the fix for #675 for classpath containers to prevent this bug from occurring.

This change ensures classpath containers with system modules (such as the JRE container) have their classpath module info attached, while the module info is not attached for other containers without the module classpath attribute. This ensures the bug in #675 is fixed also for containers on the classpath.

Fixes: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/748

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/matching/JavaSearchNameEnvironment.java
+++ b/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/matching/JavaSearchNameEnvironment.java
@@ -578,13 +578,10 @@ private static boolean isOnModulePath(JavaProject javaProject, PackageFragmentRo
 			 * so we cannot rely on the attribute.
 			 */
 			isOnModulePath = isModularProject(javaProject);
-		} else if (classpathEntry.getEntryKind() == IClasspathEntry.CPE_CONTAINER) {
+		} else if (hasSystemModule(root)) {
 			/*
 			 * The JRE 9+ container is on the module path without the 'module' classpath attribute being set.
-			 * So for any container with modules, we assume its on the module path. Since its difficult to know
-			 * whether we have the JRE container or not.
-			 * We want to improve this check with https://github.com/eclipse-jdt/eclipse.jdt.core/issues/748,
-			 * so that only module path containers have their module infos added.
+			 * We detected the JRE container by checking the container for system modules.
 			 */
 			isOnModulePath = true;
 		} else {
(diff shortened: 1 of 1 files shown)


org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/matching/JavaSearchNameEnvironment.java
```diff
@@ -602,4 +599,19 @@ private static boolean isModularProject(IJavaProject project) throws JavaModelEx
...
 }
+
+private static boolean hasSystemModule(JavaProject javaProject, IClasspathEntry classpathEntry) {
+	if (classpathEntry.getEntryKind() == IClasspathEntry.CPE_CONTAINER) {
+		IPackageFragmentRoot[] fragmentRoots = javaProject.findPackageFragmentRoots(classpathEntry);
+		if (fragmentRoots != null) {
+			for (IPackageFragmentRoot fragmentRoot : fragmentRoots) {
+				IModuleDescription module = fragmentRoot.getModuleDescription();
+				if (module != null && module.isSystemModule()) {
+					return true;
```
trancexpress, line 610: @iloveeclipse see this comment: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/748#issuecomment-1430211639

This is what the build path page for libraries does. I've not determined what the compiler does , I think its done via the `instanceof JrtPackageFragmentRoot` statements in `SearchableEnvironment`.

iloveeclipse: Simeon, I think this can be simplified. We **have** `PackageFragmentRoot` from `isOnModulePath` that is being to be added to search path, we do not want to loop over a list of **other** PFR's from the project.

iloveeclipse (changes requested):

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/matching/JavaSearchNameEnvironment.java
trancexpress: This is iterating over package fragments from the classpath **container** entry. The entry for which we are checking whether to add module infos or not.

Or am I misunderstanding you?

iloveeclipse: isOnModulePath asks about PFR. We get a classpath entry from PFR and call `hasSystemModule`  with that entry, from which we **again** asks for **all** PFR and not that one that we've got initially. This makes no sense. We **have already** PFR we want to check, it has to be provided as argument to `hasSystemModule`.

iloveeclipse (approved):

Issue #790, opened 2023-02-24
Call Hierarchy crashes with AIOOBE in QualifiedTypeReference.getTypeBinding()


iloveeclipse:
Initially I've found strange that Call Hierarchy called on `org.eclipse.osgi.internal.module.MappedList.internal` showed no results - even if the field is referenced in the same class.
Then I've realized the search silently crashed with this stack in the error log:
```
java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
	at org.eclipse.jdt.internal.compiler.ast.QualifiedTypeReference.getTypeBinding(QualifiedTypeReference.java:133)
	at org.eclipse.jdt.internal.compiler.ast.TypeReference.internalResolveType(TypeReference.java:522)
	at org.eclipse.jdt.internal.compiler.ast.TypeReference.resolveType(TypeReference.java:634)
	at org.eclipse.jdt.internal.compiler.ast.TypeReference.resolveType(TypeReference.java:630)
	at org.eclipse.jdt.internal.compiler.ast.TypeReference.resolveSuperType(TypeReference.java:602)
	at org.eclipse.jdt.internal.compiler.lookup.ClassScope.findSupertype(ClassScope.java:1673)
	at org.eclipse.jdt.internal.compiler.lookup.ClassScope.connectSuperclass(ClassScope.java:1176)
	at org.eclipse.jdt.internal.compiler.lookup.ClassScope.connectTypeHierarchy(ClassScope.java:1454)
	at org.eclipse.jdt.internal.compiler.lookup.ClassScope.connectMemberTypes(ClassScope.java:1143)
	at org.eclipse.jdt.internal.compiler.lookup.ClassScope.connectTypeHierarchy(ClassScope.java:1464)
	at org.eclipse.jdt.internal.compiler.lookup.CompilationUnitScope.connectTypeHierarchy(CompilationUnitScope.java:394)
	at org.eclipse.jdt.internal.compiler.lookup.LookupEnvironment.completeTypeBindings(LookupEnvironment.java:521)
	at org.eclipse.jdt.internal.core.search.matching.MatchLocator.locateMatches(MatchLocator.java:1338)
	at org.eclipse.jdt.internal.core.search.matching.MatchLocator.locateMatches(MatchLocator.java:1400)
	at org.eclipse.jdt.internal.core.search.matching.MatchLocator.locateMatches(MatchLocator.java:1519)
[...]

PR #793, opened 2023-02-25
AIOOBE in QualifiedTypeReference.getTypeBinding() for call hierarchy

trancexpress:
This change adjusts the QualifiedTypeReference created for a record to have matching 'sources' and 'poss' array lengths, preventing an AIOOBE in QualifiedTypeReference.getTypeBinding() due to mismatched lengths.

Fixes: #790

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchBugs16Tests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchBugs16Tests.java
@@ -16,9 +16,11 @@ import java.io.IOException;
 
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.jdt.core.ICompilationUnit;
+import org.eclipse.jdt.core.IField;
 import org.eclipse.jdt.core.IJavaElement;
 import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.ILocalVariable;
+import org.eclipse.jdt.core.IType;
 import org.eclipse.jdt.core.JavaModelException;
 import org.eclipse.jdt.core.WorkingCopyOwner;
 import org.eclipse.jdt.core.search.IJavaSearchScope;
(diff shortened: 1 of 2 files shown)


iloveeclipse:
Looks like jenkins detected new deadlock in the indexer (not because of this PR). 
Most likely introduced by https://github.com/eclipse-jdt/eclipse.jdt.core/pull/69. 
We should investigate that hang in a dedicated issue.

iloveeclipse:
> Looks like jenkins detected new deadlock in the indexer (not because of this PR).

Simeon, I see that you've pushed new commit and moved tests around and they are green now. So the deadlock was reproducible and root cause in the test code itself?

trancexpress:
> Simeon, I see that you've pushed new commit and moved tests around and they are green now. So the deadlock was reproducible and root cause in the test code itself?

I dont think so, I moved the test because its better placed in the Java 16 tests. I assumed the deadlock was sporadic.

iloveeclipse:
I've created https://github.com/eclipse-jdt/eclipse.jdt.core/issues/797 for the sporadic hang.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchBugs16Tests.java
```diff
@@ -621,6 +623,41 @@ public void testBug574870_4() throws CoreException {
 
 			}
 		}
+
+	/*
+	 * An inner record in a JRE class causes an AIOOBE during a call hierarchy computation.
```
gayanper, line 628: Could we improve the comment here ? Even though description mentions about call hierarchy, we are testing a reference search isn't it ? And also the name `RecordClass` is not clear here since there is no java record involvement.

gayanper (approved):

trancexpress:
Are we merging this @iloveeclipse ? Code freeze seems to be over (or is this again some problem with the code freeze check?).

iloveeclipse:
> Code freeze seems to be over 

Have you seen **any** mail saying that master is open for 4.28?

> (or is this again some problem with the code freeze check?).

Yes.

iloveeclipse:
**Now** we can merge :-)

Issue #765, opened 2023-02-19
ModuleBuilderTests.testPatch2 fails on Windows


iloveeclipse:
This issue is not a new one, just nobody cared so far.

See
https://download.eclipse.org/eclipse/downloads/drops4/I20230218-1800/testresults/html/org.eclipse.jdt.core.tests.model_ep427I-unit-win32-java17_win32.win32.x86_64_17.html

Stack:
```
Unexpected errors in project org.astro.patch

junit.framework.AssertionFailedError: Unexpected errors in project org.astro.patch
at junit.framework.Assert.fail(Assert.java:57)
at junit.framework.Assert.assertTrue(Assert.java:22)
at junit.framework.Assert.assertFalse(Assert.java:39)
at junit.framework.TestCase.assertFalse(TestCase.java:208)
at org.eclipse.jdt.core.tests.model.ModuleBuilderTests.assertNoErrors(ModuleBuilderTests.java:9061)
at org.eclipse.jdt.core.tests.model.ModuleBuilderTests.testPatch2(ModuleBuilderTests.java:5628)
```

The problem is that the patch module arguments use different separator for patch module files in test compared to the expected one.

And even more interesting is that the expected one is system dependent.
Means, one can't reuse same classpath between different OS's if the patch module has multiple arguments.

I wonder if that is a bug in the original implementation.
See 
- org.eclipse.jdt.core.IClasspathAttribute.PATCH_MODULE
- org.eclipse.jdt.internal.core.JavaProject.getPatchedModules(IClasspathEntry)

I see unfortunately no clear specs in Java, except this occurrence in https://docs.oracle.com/en/java/javase/19/docs/specs/man/java.html
> --patch-module module=file(;file)*
> Overrides or augments a module with classes and resources in JAR files or directories.

which seem to use `;` and not path separator for separating files.

iloveeclipse:
@stephan-herrmann : could you please check the point regarding "paths separator" assumed by JDT vs ";" in Java specs? 
Also I wonder why the attribute written to the .classpath file (and so shared between different systems) should be system specific?
The test fix itself is trivial, see draft PR https://github.com/eclipse-jdt/eclipse.jdt.core/pull/766.
I don't want to merge it yet because it would then hide an assumed bigger issue.

jarthana:
[...]

PR #806, opened 2023-03-03
Bug 566243 - testPatch2 failure

jarthana:
Fix the hard-coded File path separator

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ModuleBuilderTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ModuleBuilderTests.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2016, 2022 IBM Corporation and others.
+ * Copyright (c) 2016, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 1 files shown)


iloveeclipse:
Jay, please see my comment on https://github.com/eclipse-jdt/eclipse.jdt.core/issues/765. I'm not sure the way how this **feature** is implemented is correct, because it seem to be OS dependent and so .classpath is not portable.

jarthana:
Oops., sorry I didn't care to search for the duplicates. I still miss bugzilla in that aspect. Will close this.

iloveeclipse:
@jarthana : but what du you think about the code in question? [Is this a bug or just test issue](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/765#issuecomment-1435958342)?

jarthana:
I just looked at the Oracle's help page and I am even more confused.

The help page for Javac has this example:
`--patch-module module=file(:file)*`

whereas the one for java has this:
`--patch-module module=file(;file)*`

The spec for Javac clearly specifies this:
`--patch-module module=file(;file)*`
but the spec for Javac is mum about this

But I think the savior is this piece at the top of all options:
`In the following lists of options, an argument of path represents a search path, composed of a list of file system locations separated by the platform path separator character, (semicolon ; on Windows, or colon : on other systems.) Depending on the option, the file system locations may be directories, JAR files or JMOD files.`

So, I think a test fix is sufficient here.

stephan-herrmann:
> Jay, please see my comment on #765. I'm not sure the way how this **feature** is implemented is correct, because it seem to be OS dependent and so .classpath is not portable.

@iloveeclipse  Regarding the command line, that was Oracle's decision, but unfortunately I missed the chance to do better regarding `.classpath` ([Bug 546653](https://bugs.eclipse.org/bugs/show_bug.cgi?id=546653)): Since this is our own abstraction, in that file I could have chosen a platform-independent separator. Is it worth discussing, e.g., a lenient interpretation that accepts both separators?

iloveeclipse:
> Is it worth discussing, e.g., a lenient interpretation that accepts both separators?

Not on **this** PR for sure.
Honesty I don't think there are many people that use patch modules, but this could be very biased opinion because we don't use modules at all at my day job.

PR #812, opened 2023-03-07
Make Open Type dialog icons consistent with Outline View

jjohnstn:
- add logic to properly calculate the effective modifiers for classes and interfaces that are nested in interfaces
- fixes Bug 567663

## What it does

Fixes the Open Type dialog to properly show the correct icons for nested classes/interfaces in an interface.

## How to test

See: https://bugs.eclipse.org/bugs/show_bug.cgi?id=567663

## Author checklist


iloveeclipse:
The 15 fails seem to be consistent. Other branches don't see them.

jjohnstn:
I'll look at it.

jjohnstn:
@iloveeclipse The problem is that the change requires element info and after the call to getElementInfo(), displaying the toString() changes.  There is a better way to fix this in JDT UI so I will abandon this patch.

Issue #821, opened 2023-03-10
Call hierarchy incomplete for method of private inner type implementing public interface


trancexpress:
See: https://bugs.eclipse.org/bugs/show_bug.cgi?id=552774

For some reason we have an empty call hierarchy in the attached example, for `CDOCommitContextImpl.postCommit(CommitTransactionResult)`: 
[ws.zip](https://github.com/eclipse-jdt/eclipse.jdt.core/files/10941296/ws.zip)

The problem starts with this `ProblemReferenceBinding`:

```
"Worker-8: Fetching children of postCommit" #74 prio=5 os_prio=0 cpu=1355.56ms elapsed=1013.65s tid=0x00007f72c4018d70 nid=0x385b at breakpoint [0x00007f72b01fd000]
   java.lang.Thread.State: RUNNABLE
        at org.eclipse.jdt.internal.compiler.lookup.ProblemReferenceBinding.<init>(ProblemReferenceBinding.java:29)
        at org.eclipse.jdt.internal.compiler.lookup.Scope.findMemberType(Scope.java:1556)
        at org.eclipse.jdt.internal.compiler.lookup.Scope.getMemberType(Scope.java:3029)
        at org.eclipse.jdt.internal.compiler.lookup.Scope.getType(Scope.java:3292)
        at org.eclipse.jdt.internal.core.search.matching.MatchLocator.getType(MatchLocator.java:972)
        at org.eclipse.jdt.internal.core.search.matching.MatchLocator.getMethodBinding0(MatchLocator.java:1144)
        at org.eclipse.jdt.internal.core.search.matching.MatchLocator.getMethodBinding(MatchLocator.java:982)
        at org.eclipse.jdt.internal.core.search.matching.MethodLocator.matchMethod(MethodLocator.java:357)
        at org.eclipse.jdt.internal.core.search.matching.MethodLocator.resolveLevel(MethodLocator.java:787)
        at org.eclipse.jdt.internal.core.search.matching.MethodLocator.resolveLevel(MethodLocator.java:718)
        at org.eclipse.jdt.internal.core.search.matching.MatchLocator.reportMatching(MatchLocator.java:2807)
        at org.eclipse.jdt.internal.core.search.matching.MatchLocator.process(MatchLocator.java:2046)
        at org.eclipse.jdt.internal.core.search.matching.MatchLocator.locateMatches(MatchLocator.java:1362)
[...]

PR #823, opened 2023-03-12
Call hierarchy incomplete for method of private nested type

trancexpress:
An incomplete call hierarchy is computed given the following conditions:

1. A private nested class implements an interface.
2. The interface has a method with a nested type.
3. The call hierarchy starts at the private nested class level.

This is a regression from the fix for:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=431357

In particular the fix uses `MethodBinding.parameters`, which is empty for the callee method in the described scenario.

A `MethodBinding` with empty `parameters` is computed due to the following sequence:

1. The method caller is not able to see the private private nested class with the callee method, resulting in creating a `ProblemReferenceBinding` for the nested private class.
2. `MatchLocator.getMethodBinding(MethodPattern)` fails to retrieve a `MethodBinding` via `MatchLocator.getMethodBinding0(MethodPattern)`, due to the `ProblemReferenceBinding`.
3. `getMethodBinding()` resorts to parsing source code, but retrieves a `MethodBinding` with unresolved parameters.

This change adjusts the part of `getMethodBinding()` which resorts to reading a source file, so that method parameters are resolved if they are not resolved yet.

Fixes: #821

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/matching/MatchLocator.java
+++ b/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/matching/MatchLocator.java
@@ -984,6 +984,10 @@ public MethodBinding getMethodBinding(MethodPattern methodPattern) {
     	return methodBinding; // known to be valid.
     // special handling for methods of anonymous/local types. Since these cannot be looked up in the environment the usual way ...
     if (methodPattern.focus instanceof SourceMethod) {
+    	MethodBinding binding = getClosestMatchMethodBinding(methodPattern);
+    	if (binding != null) {
+    		return binding;
+    	}
     	char[] typeName = PatternLocator.qualifiedPattern(methodPattern.declaringSimpleName, methodPattern.declaringQualification);
     	if (typeName != null) {
     		IType type = methodPattern.declaringType;
@@ -1012,24 +1016,22 @@ public MethodBinding getMethodBinding(MethodPattern methodPattern) {
     } else if (methodPattern.focus instanceof BinaryMethod &&
     		methodPattern.declaringType instanceof BinaryType &&
     		this.unitScopeTypeBinding instanceof ProblemReferenceBinding) {//Get binding from unit scope for non-visible member of binary type
-    	char[] typeName = PatternLocator.qualifiedPattern(methodPattern.declaringSimpleName, methodPattern.declaringQualification);
-    	if (typeName != null) {
-    		IType type = methodPattern.declaringType;
-    		IType enclosingType = type.getDeclaringType();
-    		while (enclosingType != null) {
-    			type = enclosingType;
-    			enclosingType = type.getDeclaringType();
-    		}
-    		typeName = type.getFullyQualifiedName().toCharArray();
-    		TypeBinding typeBinding = this.unitScopeTypeBinding;
-    		if (typeBinding instanceof ProblemReferenceBinding) {
-    			ProblemReferenceBinding problemReferenceBinding = (ProblemReferenceBinding) this.unitScopeTypeBinding;
-    			ReferenceBinding closestMatch = (problemReferenceBinding.problemId() == ProblemReasons.NotVisible) ?
-    					problemReferenceBinding.closestReferenceMatch() : null;
-    					return closestMatch != null ?  getMethodBinding(methodPattern, closestMatch) : null;
-    		}
-    	}
+    	return getClosestMatchMethodBinding(methodPattern);
     }
+	return null;
+}
+
+private MethodBinding getClosestMatchMethodBinding(MethodPattern methodPattern) {
+	TypeBinding typeBinding = this.unitScopeTypeBinding;
+	if (typeBinding instanceof ProblemReferenceBinding) {
+		ProblemReferenceBinding problemReferenceBinding = (ProblemReferenceBinding) this.unitScopeTypeBinding;
+		if (problemReferenceBinding.problemId() == ProblemReasons.NotVisible) {
+			ReferenceBinding closestMatch = problemReferenceBinding.closestReferenceMatch();
+			if (closestMatch != null) {
+				return getMethodBinding(methodPattern, closestMatch);
+			}
+		}
+	}
 	return null;
 }
 
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchBugsTests2.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchBugsTests2.java
@@ -2981,6 +2981,57 @@ public class JavaSearchBugsTests2 extends AbstractJavaSearchTests {
 		}
 	}
 
+	/**
+	 * <br>For the following scenario, no call hierarchy is found:</br>
+	 * A nested private class implements a public nested interface.
+	 * The overridden method from the interface also has an argument of the interface type.
+	 * No call hierarchy is found for the overridden method, when searching from within the nested class.
+	 * See: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/821
+	 */
+	public void testCallHierarchyWithNestedInterfacesGh821() throws Exception {
+		String projectName = "TestCallHierarchyWithNestedInterfacesGh821";
+		try {
+			IJavaProject project = createJavaProject(projectName, new String[] { "src" }, new String[] { "JCL15_LIB" }, "bin", "1.5");
+			String packageFolder = "/" + projectName + "/src/test";
+			createFolder(packageFolder);
+			String snippet1 =
+					"""
+					package test;
+					public class TestNestedPrivateClass {
+						public NestedInterface nm = new NC();
+						public interface NestedInterface {
+							public void foo(NestedInterface arg);
+						}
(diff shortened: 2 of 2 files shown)


trancexpress:
@iloveeclipse can you review this? Its part of the main bug for search problems.

gayanper:
@trancexpress could you please fix the conflicts and if @iloveeclipse is busy let me know I can have a look at it.

trancexpress:
> @trancexpress could you please fix the conflicts and if @iloveeclipse is busy let me know I can have a look at it.

Rebased, please review. Andrey is unfortunately busy.

trancexpress:
@gayanper do you have time to review this change?

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/matching/MatchLocator.java
```diff
@@ -1002,6 +1002,11 @@ public MethodBinding getMethodBinding(MethodPattern methodPattern) {
     				if (unit != null) {
     					AbstractMethodDeclaration amd = new ASTNodeFinder(unit).findMethod((IMethod) methodPattern.focus);
     					if (amd != null && amd.binding != null && amd.binding.isValidBinding()) {
+    						// Check if we "forgot" to resolve parameter types, see: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/821
```
gayanper, line 1005: I just tried the same workaround for BinaryMethod which is trying to resolve the method by using the following expression as a test `getMethodBinding(methodPattern, ((ProblemReferenceBinding) this.unitScopeTypeBinding).closestReferenceMatch())` and following is the output.

image

trancexpress: > I just tried the same workaround for BinaryMethod which is trying to resolve the method by using the following expression as a test `getMethodBinding(methodPattern, ((ProblemReferenceBinding) this.unitScopeTypeBinding).closestReferenceMatch())` and following is the output.

Sorry I don't understand this comment. Do you mean you have an example where the bug from #821 is seen also in a library binary file?

If so, what change are you applying and is the result satisfactory or bad?

gayanper: @trancexpress sorry for not being clear. I checked you changes, if I understand correctly, you try to resolve the method binding with parameters using the fix. 

What i tried is the fix that is there in the if else block, which tries the resolve the method of hidden type because its nested private. 

So in my previous comment the expression i tried out. So if you add a debug point in the if condition which checks for SourceMethod and run that expression, you will get the output in screenshot. 

So does this output is different than what you expect from your fix ? Because if not we can both merge the binary and source method resolutions login in this method right ?

trancexpress: As long as the added test and the original problem in #821 are fixed, I don't have a preference. It has been a while so I don't remember the exact specifics (and so I cannot tell from the screenshot whether the bug is fixed). Your suggestion sounds like it will simplify code, so we can try it out. If nothing fails and the problems are fixed, sure, we should do this.

Since you've already adjusted the code on your end, can you amend the commit?

gayanper: @trancexpress not i just did a debug evaluation without all the if guards to see if that provides the method binding with parameters resolved

trancexpress: In what context did you try to evaluate `getMethodBinding(methodPattern, ((ProblemReferenceBinding) this.unitScopeTypeBinding).closestReferenceMatch())`?

There is no `ProblemReferenceBinding` in the bug case. So when I evaluate this expression I get:

```
ClassCastException: Cannot cast org.eclipse.jdt.internal.compiler.lookup.SourceTypeBinding (id=303) to org.eclipse.jdt.internal.compiler.lookup.ProblemReferenceBinding		
```

gayanper: I did while in debug mode with the tests in this PR. May be you can check adding a conditional breakpoint and see if you every come to this method which satisfies that. Otherwise I can check a bit more tomorrow.

iloveeclipse:
Please remove "test" commit.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/matching/MatchLocator.java
trancexpress: OK, I was setting a breakpoint too deep in the source case. When setting a breakpoint at the start of the source case, I see the `ProblemReferenceBinding`.

Can you check the current changes? I've pushed an extra commit so its easier to see the changes.

Unfortunately this is the best I can do. Without the remaining code in the source case there are at least fails in `JavaSearchBugsTests2`. Probably more in other test cases.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/matching/MatchLocator.java
```diff
@@ -1012,27 +1016,25 @@ public MethodBinding getMethodBinding(MethodPattern methodPattern) {
     } else if (methodPattern.focus instanceof BinaryMethod &&
     		methodPattern.declaringType instanceof BinaryType &&
     		this.unitScopeTypeBinding instanceof ProblemReferenceBinding) {//Get binding from unit scope for non-visible member of binary type
-    	char[] typeName = PatternLocator.qualifiedPattern(methodPattern.declaringSimpleName, methodPattern.declaringQualification);
-    	if (typeName != null) {
```
trancexpress, line 1016: Removing this part might cause different behavior, but to me its dead code. `typeName` is not read by this block, just set below (maybe code was deleted at some point, or this was copy-paste).

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/matching/MatchLocator.java
trancexpress: Though I wonder if we should still make sure `amd.binding` has completed type bindings... I'm not sure if there will be cases where `amd.binding` is returned and its incomplete (due to having no type bindings for its parameters).

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/matching/MatchLocator.java
gayanper: agree lets remove, we can visit if something goes wrong in ibuilds.

iloveeclipse (changes requested):
Please get rid of the "test" commit

trancexpress:
> Please get rid of the "test" commit

I read this the first time. Once the review is done I'll squash.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/matching/MatchLocator.java
```diff
@@ -1002,6 +1002,11 @@ public MethodBinding getMethodBinding(MethodPattern methodPattern) {
     				if (unit != null) {
     					AbstractMethodDeclaration amd = new ASTNodeFinder(unit).findMethod((IMethod) methodPattern.focus);
     					if (amd != null && amd.binding != null && amd.binding.isValidBinding()) {
+    						// Check if we "forgot" to resolve parameter types, see: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/821
+    						if (amd.binding.parameterNames != null && amd.binding.parameters != null &&
+    								amd.binding.parameterNames.length != amd.binding.parameters.length) {
+    							sourceTypeBinding.resolveTypesFor(amd.binding);
```
trancexpress: @gayanper , are you sure we shouldn't keep this? From my debugging, as soon as the code is here the parameter types might be unresolved. I've not tried to construct a case where this is a problem, but maybe there are such cases.

gayanper: If we can find a situation where the closest match doesn't resolve, I think we can have this, But i think for this specific issue the closest match solve the resolution issue right ?

But as you say we might have other scenarios where method bindings are not property resolved, when looking at the implementation of `resolveTypesFor` it actually use `resolveTypesWithSuspendedTempErrorHandlingPolicy` which sounds like tring to resolve ignore any errors. So using this method might be usefull in some scenarios.

If we keep it, I would like we make the comment more clear rather than highlighting it as a fix for this issue since it is not anymore at least from the scenarios we know right ?

What if we have a method which has some issues like incompatible generic usage or some issue with parameters definitions (not access issue like this one), will this code block helps to still resolve the call hierarchy ?

I'm more leaning towards removing this block unless it doesn't solve any scenario related to search. @iloveeclipse  WDYT ?

trancexpress:
@gayanper , can you check this again? The change should be ready for merging.

gayanper (approved):

PR #829, opened 2023-03-13
20 record pattern type inference

mpalat:
## Author checklist

Major part for the Record Type Inference

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/core/compiler/IProblem.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/core/compiler/IProblem.java
@@ -2545,4 +2545,9 @@ void setSourceStart(int sourceStart);
 	 * @noreference preview feature
 	 */
 	int RawTypeInRecordPattern =  PreviewRelated + 1915;
+	/**
+	 * @since 3.33
+	 * @noreference preview feature
+	 */
+	int CannotInferRecordPatternTypes = PreviewRelated + 1940;
 }
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
@@ -32,6 +32,7 @@ import org.eclipse.jdt.internal.compiler.ast.Expression;
 import org.eclipse.jdt.internal.compiler.ast.FunctionalExpression;
 import org.eclipse.jdt.internal.compiler.ast.Invocation;
 import org.eclipse.jdt.internal.compiler.ast.LambdaExpression;
+import org.eclipse.jdt.internal.compiler.ast.RecordPattern;
 import org.eclipse.jdt.internal.compiler.ast.ReferenceExpression;
 import org.eclipse.jdt.internal.compiler.ast.SwitchExpression;
 import org.eclipse.jdt.internal.compiler.ast.Wildcard;
@@ -791,6 +792,12 @@ public class InferenceContext18 {
 			return null;
 		InferenceVariable[] alpha = addInitialTypeVariableSubstitutions(a);
 
+		createAdditionalBounds(a, alpha);
+		TypeBinding falpha = substitute(functionalInterface);
+		return falpha.getSingleAbstractMethod(this.scope, true).parameters;
+	}
+
+	private void createAdditionalBounds(TypeBinding[] a, InferenceVariable[] alpha) {
 		for (int i = 0; i < a.length; i++) {
 			TypeBound bound;
 			if (a[i].kind() == Binding.WILDCARD_TYPE) {
@@ -813,8 +820,6 @@ public class InferenceContext18 {
 			}
 			this.currentBounds.addBound(bound, this.environment);
 		}
-		TypeBinding falpha = substitute(functionalInterface);
-		return falpha.getSingleAbstractMethod(this.scope, true).parameters;
 	}
 
 	public boolean reduceWithEqualityConstraints(TypeBinding[] p, TypeBinding[] q) {
@@ -1019,6 +1024,15 @@ public class InferenceContext18 {
 	 * @throws InferenceFailureException a compile error has been detected during inference
 	 */
 	public /*@Nullable*/ BoundSet solve(boolean inferringApplicability) throws InferenceFailureException {
+		return solve(inferringApplicability, false);
+	}
+	/**
+	 * Try to solve the inference problem defined by constraints and bounds previously registered.
+	 * @param skipLU see 18_5_5_item_5 for Record Type Inference
+	 * @return a bound set representing the solution, or null if inference failed
+	 * @throws InferenceFailureException a compile error has been detected during inference
+	 */
+	private /*@Nullable*/ BoundSet solve(boolean inferringApplicability, boolean skipLU) throws InferenceFailureException {
 
 		if (!reduce())
 			return null;
@@ -1027,7 +1041,7 @@ public class InferenceContext18 {
 		if (inferringApplicability)
 			this.b2 = this.currentBounds.copy(); // Preserve the result after reduction, without effects of resolve() for later use in invocation type inference.
 
-		BoundSet solution = resolve(this.inferenceVariables);
+		BoundSet solution = resolve(this.inferenceVariables, skipLU);
 
 		/* If inferring applicability make a final pass over the initial constraints preserved as final constraints to make sure they hold true at a macroscopic level.
 		   See https://bugs.eclipse.org/bugs/show_bug.cgi?id=426537#c55 onwards.
@@ -1050,12 +1064,15 @@ public class InferenceContext18 {
 	}
 
 	public /*@Nullable*/ BoundSet solve(InferenceVariable[] toResolve) throws InferenceFailureException {
+		return solve(toResolve, false);
+	}
+	public /*@Nullable*/ BoundSet solve(InferenceVariable[] toResolve, boolean skipLU) throws InferenceFailureException {
 		if (!reduce())
 			return null;
 		if (!this.currentBounds.incorporate(this))
 			return null;
 
-		return resolve(toResolve);
+		return resolve(toResolve, skipLU);
 	}
 
 	/**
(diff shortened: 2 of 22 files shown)


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
```diff
@@ -1794,7 +1794,7 @@ public ReferenceBinding inferRecordPatternParameterization(
 		if (!(typeBinding instanceof ReferenceBinding))
 			return null; // should not happen.
 		// 1.If T is not downcast convertible (5.5) to the raw type R, inference fails.
-		if (!typeBinding.isCompatibleWith(candidateT))
+		if (!isCompatibleWith(candidateT, typeBinding))
```
stephan-herrmann, line 1797: We have a specific understanding of `isCompatibleWith`, but the new method seems to do something different. 
Does the implementation conform to JLS 5.5? 
We didn't yet have an implementation that does this?
If the method is needed indeed, perhaps `isDowncastConvertibleTo` is a better name?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
```diff
@@ -1838,6 +1838,27 @@ public ReferenceBinding inferRecordPatternParameterization(
 		return getRecordPatternTypeFromUpwardsProjection(typeBinding, alphas, solution);
 	}
 
+	private boolean isCompatibleWith(TypeBinding candidateT, TypeBinding typeBinding) {
+		if (candidateT.isTypeVariable() || candidateT.isIntersectionType18()) {
+			// 18.5.5_item_3_bullet_3
+			/* If T is a type variable or an intersection type, then for each upper bound of the type
+			 * variable or element of the intersection type, this step and step 4 are repeated
+			 * recursively. All bounds produced in steps 3 and 4 are incorporated into a single bound set.*/
```
stephan-herrmann, line 1846: this comment looks misplaced.

stephan-herrmann (commented):
I dropped a few quick questions. Without a specific question I didn't really know what to look for :)

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/core/compiler/IProblem.java
```diff
@@ -2545,4 +2545,9 @@ public interface IProblem {
 	 * @noreference preview feature
 	 */
 	int RawTypeInRecordPattern =  PreviewRelated + 1915;
+	/**
+	 * @since 3.33
+	 * @noreference preview feature
+	 */
+	int CannotInferRecordPatternTypes = PreviewRelated + 1940;
```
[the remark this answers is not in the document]
mpalat: commit 383db211d66ab0bfd3506b70baa47e830b10a627 from Jay has changes pertaining to this - since he was working on this, didn't add this.. Am rebasing to that commit and we should be having the jenkins runs.

mpalat:
Thanks @stephan-herrmann for https://github.com/eclipse-jdt/eclipse.jdt.core/pull/829/commits/4d84d1478945631621d6e7bcb4f654223bba8e16 - this helped move faster. 

(1) Here: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/829/commits/7be9b9cdc03709aca676396c8da2c1e88ddcaba8, this implementation is not complete. I am passing the SwiSta.expression for resolving with expression - I don;t see any other possibility for T [ie candidateT]. WDYT?
(2) Also, ref the IC18 commented TODO. Should we check for the absence of type arguments in terms of length of the array?  
(3) looking into the PatternMatch16Test failures-- finally the jenkins is back

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
```diff
@@ -1910,4 +1784,412 @@ public void cleanUp() {
...
+		if (tPrime == null)
+			return false;
+		ReferenceBinding rAlpha = this.environment.createParameterizedType(
+				(ReferenceBinding) typeBinding.original(), alphas, typeBinding.enclosingType(), typeBinding.getTypeAnnotations());
+		TypeBinding[] rPrimes = this.currentBounds.condition18_5_5_item_4(rAlpha, alphas, tPrime, this);
+		if (rPrimes != null) {
+			// TODO:
+			/* The constraint formula ‹T' = R'› is reduced (18.2) and the resulting bounds are
+			 * incorporated into B1 to produce a new bound set, B2.
+			 * TODO: refactor the method below into two? - instead of creating array - here reusing. */
```
stephan-herrmann, line 1909: Re "TODO: refactor...":

I think `reduceWithEqualityConstraints` is just not needed here, what about
```
try {
	if (!reduceAndIncorporate(ConstraintTypeFormula.create(tPrime, rPrime, ReductionResult.SAME))) {
		 /* If B2 contains the bound false, inference fails. */
		return false;
	}
} catch (InferenceFailureException e) {
	return false;
}
```

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
```diff
@@ -1130,151 +1147,8 @@ public boolean reduceAndIncorporate(ConstraintFormula constraint) throws Inferen
...
-						if (tmpBoundSet.numUninstantiatedVariables(this.inferenceVariables) == oldNumUninstantiated)
-							return null; // abort because we made no progress
-						continue;
-					}
-					return null;
-				}
-			}
-		}
-		return tmpBoundSet;
+		return resolve(toResolve, false);
```
stephan-herrmann, line 1150: please avoid moving code around like this, it would kill this method's git history which is vital for future maintenance!

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ParameterizedRecordComponentBinding.java
```diff
@@ -0,0 +1,68 @@
...
+ * their signature did involve generics or not, so as to get the proper declaringClass for
+ * these record components.
+ */
+public class ParameterizedRecordComponentBinding extends RecordComponentBinding {
+
+    public RecordComponentBinding originalRecordComponent;
+
+public ParameterizedRecordComponentBinding(ParameterizedTypeBinding parameterizedDeclaringClass, RecordComponentBinding originalRecordComponent) {
+	super (originalRecordComponent.name,
+            (originalRecordComponent.modifiers & ClassFileConstants.AccEnum) != 0
```
stephan-herrmann, line 35: copied from `ParameterizedFieldBinding`? A field can be an enum constant, but can a record component?

stephan-herrmann (changes requested):
I dropped a few more suggestions, but the one regarding IC18.resolve() is really a request.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
```diff
@@ -1910,4 +1784,412 @@ public void cleanUp() {
...
+		}
+		return typeVariables;
+	}
+	 /* <b>JLS 18.4 for 18.5.5_item_3_bullet_5</b> Resolution
+	 * @return answer null if some constraint resolved to FALSE, otherwise the boundset representing the solution
+	 * @throws InferenceFailureException
+	 */
+	private /*@Nullable*/ BoundSet resolve(
+			InferenceVariable[] toResolve,
+			boolean skipLU) throws InferenceFailureException {
```
stephan-herrmann, line 2047: What does `skipLU` mean? Can we have a better name?

stephan-herrmann:
Investigating the situation of raw type R vs. its generic original() using `testRecordPatternTypeInference_008()` I noticed that resolving unduly reports 
```
R is a raw type. References to generic type R<T> should be parameterized
```

which can be avoided by setting `IgnoreRawTypeCheck` like so:
```
public TypeBinding resolveType(BlockScope scope, boolean isPatternVariable) {
	if (this.resolvedType != null)
		return this.resolvedType;
	super.resolveType(scope, isPatternVariable);

	if (this.local != null) {
		this.resolvedType = super.resolveType(scope);
	} else {
   		this.type.bits |= ASTNode.IgnoreRawTypeCheck;
		this.resolvedType = this.type.resolveType(scope);
	}
	...
```

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
```diff
@@ -1910,4 +1937,256 @@ public void cleanUp() {
...
+		Expression synthExpr = new Expression() {
+			@Override public StringBuffer printExpression(int indent, StringBuffer output) {
+				return output;
+			}
+		};
+		if (!synthExpr.checkCastTypesCompatibility(scope2, candidateT, typeBinding, synthExpr, false))
+			return null;
+//		if (!isDownwardCompatible(candidateT, typeBinding))
+//			return null;
+		//2. Otherwise, where P1, ..., Pn (n ≥ 1) are the type parameters of R,...
```
stephan-herrmann, line 1959: I believe that within JLS 18.5.5 the check for downcast convertibility is the only thing that should actually use the raw type R (which we correctly get from SingleTypeReference.resolved()). This implies that after that check we can safely update typeBinding:
```typeBinding = typeBinding.original()``` which will allow us to remove all `original()` calls in downstream code for simplification.

PS: if you're comfortable with `synthExpr.checkCastTypesCompatibility()` you may want to remove the commented code above. :)

stephan-herrmann (approved):

stephan-herrmann (commented):
While things are generally good-looking, this round of comments is already at the polish level. Perhaps avoiding `e.printStackTrace();` being the "most severe".

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
mpalat: Agree.. but I believe the //2.. needs to be kept for the steps.. keeping that, pl let me know if you think otherwise

Issue #837, opened 2023-03-14
Unexpected cast compile error for generic type argument


trancexpress:
See: https://bugs.eclipse.org/bugs/show_bug.cgi?id=562230

To reproduce, paste the following snippet in Eclipse:

```
package test;

import java.util.Collections;
import java.util.List;

public class Test {

	public static <Z extends Comparable<? super Z>> String test(List<Z> in) {
		List<String> x = (List<String>) in;
		return x.get(0);
	}

    public static void main(String[] args) {
        List<String> list = Collections.singletonList("hello");
        String string = Test.<String>test(list);
        System.out.println(string);
    }
}
```

Compile and run with `javac`/`java` (OpenJDK 17), observe no errors:

```
➜  src /usr/lib/jvm/java-17-openjdk/bin/javac test/Test.java
Note: test/Test.java uses unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
```
```
➜  src /usr/lib/jvm/java-17-openjdk/bin/java test.Test      
hello
```

With `ecj`/Eclipse though, observe compile error:

```
Description	Resource	Path	Location	Type
Cannot cast from List<Z> to List<String>	Test.java	/TestBug562230/src/test	line 9	Java Problem
```

The problem is created here:

```
"org.eclipse.jdt.internal.ui.text.JavaReconciler" #82 daemon prio=1 os_prio=0 cpu=1512.91ms elapsed=1783.09s tid=0x00007f381e1e6bb0 nid=0x1251 at breakpoint [0x00007f377cae6000]
   java.lang.Thread.State: RUNNABLE
        at org.eclipse.jdt.internal.compiler.problem.ProblemHandler.handle(ProblemHandler.java:130)
        at org.eclipse.jdt.internal.compiler.problem.ProblemHandler.handle(ProblemHandler.java:229)
        at org.eclipse.jdt.internal.compiler.problem.ProblemReporter.handle(ProblemReporter.java:2679)
        at org.eclipse.jdt.internal.compiler.problem.ProblemReporter.typeCastError(ProblemReporter.java:8542)
        at org.eclipse.jdt.internal.compiler.ast.CastExpression.resolveType(CastExpression.java:676)
        at org.eclipse.jdt.internal.compiler.ast.LocalDeclaration.resolve(LocalDeclaration.java:390)
[...]

PR #838, opened 2023-03-14
Unexpected cast compile error for generic type argument

trancexpress:
Given a generic type defined as `<Z extends Comparable<? super Z>>`, the compiler fails to cast `List<Z>` to e.g. `List<String>`. Despite `String` fulfilling the bound requirements for `Z` (due to `String implements Comparable<String>`).

For this particular case, the bug occurs due to:

For the `? super Z` bound, `TypeBinding.isTypeArgumentContainedBy()` calls `isCompatibleWith()` on `String`, with argument the lower bound `?`. Then in `ReferenceBinding.isCompatibleWith0()` the equivalence check fails, since `String` doesn't evaluate to being equivalent to `?` - `String` is not a parameterized type.

This change adds an extra equivalence check, ensuring `?` is tested for equivalence to `String` (or any other type that would fit the generic argument bounds).

Fixes: #837

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
@@ -1488,6 +1488,8 @@ private boolean isCompatibleWith0(TypeBinding otherType, /*@Nullable*/ Scope cap
 	// bound
 	if (isEquivalentTo(otherType))
 		return true;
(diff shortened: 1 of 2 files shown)


trancexpress:
Lets see how tests look. I'm not too sure about this change, maybe the extra equivalence check should be placed directly in `TypeBinding.isTypeArgumentContainedBy(TypeBinding)`. But then the case for `extend` would look weird without such a check...

trancexpress:
Not the right approach. I think we need to substitute the bound `? super Z` with `? super String` during the cast check, as opposed to eliminating the parameter from it. But I lack knowledge to do this.

Issue #831, opened 2023-03-13
Autocomplete blows up when trying to list the members for java.util.Optional.of(true ? 0 : "")


yeah-boi:
Autocomplete blows up when trying to list the members for:

`java.util.Optional.of(true ? 0 : "")`

See the following image:

![image](https://user-images.githubusercontent.com/30007588/224806588-293928d8-6da0-4c7d-95f9-973a59e25c54.png)

I'm using the following version:

Version: 2022-12 (4.26.0)
Build id: 20221201-1913

trancexpress:
```
java.lang.UnsupportedOperationException
	at org.eclipse.jdt.internal.compiler.lookup.IntersectionTypeBinding18.getPackage(IntersectionTypeBinding18.java:142)
	at org.eclipse.jdt.internal.compiler.lookup.TypeBinding.qualifiedPackageName(TypeBinding.java:1571)
	at org.eclipse.jdt.internal.codeassist.CompletionEngine.findLocalMethods(CompletionEngine.java:9606)
	at org.eclipse.jdt.internal.codeassist.CompletionEngine.findMethods(CompletionEngine.java:10964)
	at org.eclipse.jdt.internal.codeassist.CompletionEngine.findFieldsAndMethods(CompletionEngine.java:7701)
	at org.eclipse.jdt.internal.codeassist.CompletionEngine.completionOnMemberAccess(CompletionEngine.java:3195)
	at org.eclipse.jdt.internal.codeassist.CompletionEngine.complete(CompletionEngine.java:2027)
	at org.eclipse.jdt.internal.codeassist.CompletionEngine.complete(CompletionEngine.java:2346)
	at org.eclipse.jdt.internal.core.Openable.codeComplete(Openable.java:136)
	at org.eclipse.jdt.internal.core.CompilationUnit.codeComplete(CompilationUnit.java:367)
	at org.eclipse.jdt.internal.core.CompilationUnit.codeComplete(CompilationUnit.java:357)
	at org.eclipse.jdt.internal.ui.text.java.JavaCompletionProposalComputer.internalComputeCompletionProposals(JavaCompletionProposalComputer.java:256)
	at org.eclipse.jdt.internal.ui.text.java.JavaCompletionProposalComputer.computeCompletionProposals(JavaCompletionProposalComputer.java:218)
[...]

PR #841, opened 2023-03-15
Exception thrown during auto complete for Optional

trancexpress:
Add an override `IntersectionTypeBinding18.qualifiedPackageName()` to prevent exceptions in `IntersectionTypeBinding18.getPackage()`, when trying to retrieve the package name of a type intersection.

The new override is similar to the override for `qualifiedSourceName()`.

Fixes: #831

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/IntersectionTypeBinding18.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/IntersectionTypeBinding18.java
@@ -39,6 +39,8 @@ import org.eclipse.jdt.internal.compiler.classfmt.ClassFileConstants;
  */
 public class IntersectionTypeBinding18 extends ReferenceBinding {
 
+	private static final char[] INTERSECTION_PACKAGE_NAME = "<package intersection>".toCharArray(); //$NON-NLS-1$
+
 	public ReferenceBinding [] intersectingTypes;
 	private ReferenceBinding javaLangObject;
 	int length;
@@ -253,6 +255,11 @@ public class IntersectionTypeBinding18 extends ReferenceBinding {
 		if (classIdx > -1 && classIdx < Integer.MAX_VALUE)
 			return this.intersectingTypes[classIdx].erasure();
 		return this;
+	}
+
+	@Override
+	public char[] qualifiedPackageName() {
+		return INTERSECTION_PACKAGE_NAME;
 	}
 
 	@Override
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests18.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests18.java
@@ -6776,4 +6776,27 @@ public void testGH767() throws JavaModelException {
 	assertTrue(String.format("Result doesn't contain expected method (%s)", result),
 			result.contains("length[METHOD_REF]{length(), Ljava.lang.String;, ()I, length, null, 60}\n"));
 }
+// https://github.com/eclipse-jdt/eclipse.jdt.core/issues/831
+public void testIntersection18GH831() throws Exception {
+	this.workingCopies = new ICompilationUnit[1];
+	this.workingCopies[0] = getWorkingCopy(
+			"/Completion/src/Foo.java",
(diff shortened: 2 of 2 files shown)


trancexpress:
@stephan-herrmann since you've [commented on the original bug|https://bugs.eclipse.org/bugs/show_bug.cgi?id=559168#c1], could you check the changes here?

I'm not sure what the type names separated by `&` are meant to represent, but my guess is it doesn't hurt to do the same about the qualified package names.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/IntersectionTypeBinding18.java
```diff
@@ -266,6 +266,17 @@ public char[] qualifiedSourceName() {
 		return qualifiedSourceName.toString().toCharArray();
 	}
 
+	@Override
+	public char[] qualifiedPackageName() {
+		StringBuilder qualifiedPackageName = new StringBuilder(16);
```
gayanper, line 271: Is this implementation accurate according to https://docs.oracle.com/javase/specs/jls/se8/jls8.pdf 4.9 ? and also 4.4 ?

trancexpress: Hard for me to say. There is no mention of packages, only types. I can ask on the `compiler-dev` mailing list?

stephan-herrmann: `qualifiedPackageName()` is never really used in compiler land, so I see no need to consult JLS for this. Any mailing list outside JDT will not know what we use the package name for.

OTOH, I would be surprised if any code in JDT can make sense of an '&' joined sequence of package names.

I haven't debugged the test, but I assume, we only expect one property from this "package name": that it exists (and is equal to itself), am I right?

In other parts of JDT I see lots of callers of `qualifiedPackageName()` for which that &-joined name doesn't make any sense, notably when that is composed with a type name to yield a fully qualified type name.

Would it make sense to actually prevent the call that causes the exception during search?

Alternatively, to just prevent the exception, we could also consider returning a constant synthetic name, like `"<package intersection>"`, which is guaranteed to never occur in source?

stephan-herrmann:
> I'm not sure what the type names separated by `&` are meant to represent

That's an intersection type, meaning: each value of this type must be a member of the first type **and** the second type as well ("intersection" as in: all those instances that are members of both types). There is no similar notion for packages, as packages are disjoint by definition.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/IntersectionTypeBinding18.java
trancexpress: > In other parts of JDT I see lots of callers of `qualifiedPackageName()` for which that &-joined name doesn't make any sense, notably when that is composed with a type name to yield a fully qualified type name.
> 
> Would it make sense to actually prevent the call that causes the exception during search?

The many callers is why I wouldn't try to avoid the `qualifiedPackageName()` call. We can prevent the auto-completion case here, but in https://bugs.eclipse.org/bugs/show_bug.cgi?id=559168 the problem apparently occurred during a call hierarchy search. Who knows where calling `qualifiedPackageName()` will explode due to `getPackage()` throwing `UnsupportedOperationException`.

> Alternatively, to just prevent the exception, we could also consider returning a constant synthetic name, like `"<package intersection>"`, which is guaranteed to never occur in source?

Initially I wanted to return `CharOperation.NO_CHAR`, but I thought maybe something like `qualifiedSourceName()` makes sense. I'll add a constant as you suggest.

gayanper: > Hard for me to say. There is no mention of packages, only types. I can ask on the `compiler-dev` mailing list?

I think @stephan-herrmann answered this above. But my kind of cryptic question was whether returning concat packages with `&` is correct when reading how type binding is resolved as defined in JLS. But since packages are not used in compiler this will only cause a side effect if we try to use the package name for some type info lookup thinking its a real package name.

gayanper (approved):
LGTM

Issue #856, opened 2023-03-17
Call Hierarchy identifies incompatible methods with same signature


trancexpress:
See: https://bugs.eclipse.org/bugs/show_bug.cgi?id=117028

Snippet:

```
package eclipse;

public class CallHiearchyTest {
    // case 1
    public static interface Action1 {
        public void doit();
    }

    public static void action11( Action1 action ) {
        action.doit();
    }

    public static void action12( Action1 action ) {
        action.doit();
    }

    // case 2
    public static interface Action2 {
        public void doit();
    }

    public static void action2( Action2 action ) {
        action.doit();
    }

    public static void test21() {
        action2( new Action2() {
            public void doit() {
                // (2) view call hierachy of this method
                System.out.println( "Foo" );
            }
        } );
    }

    public static void test22() {
        action2( new Action2() {
            public void doit() {
                // (3) view call hierachy of this method
                System.out.println( "Bar" );
            }
        } );
    }

    public static void main( String[] args ) {
        action11( new Action1() {
            public void doit() {
                // (1) view call hierachy of this method
                System.out.println( "Yeah!" );
            }
        } );
    }

}
```

Call hierarchy of e.g. `eclipse.CallHiearchyTest.main(...).new Action1() {...}.doit()` has callers:

```
action11(Action1)
action12(Action1)
action2(Action2)
```

`action2(Action2)` should not be on the call hierarchy here, `Action1` and `Action2` don't implement a common interface containing `doit()`.

iloveeclipse:
This os one I hate most. Especially on methods named run(), getName() etc call hierarchy usually shows too many false matches.

trancexpress:
Checking with simplified snippet:

```
public class CallHiearchyTest {
    public static interface Action1 {
        public void doit();
    }

    public static interface Action2 {
    	public void doit();
    }

    public static void action1( Action1 action ) {
        action.doit();
    }

    public static void action2( Action2 action ) {
        action.doit();
    }

    public static void test22() {
        action2( new Action2() {
            public void doit() { // call hierarchy here
                System.out.println( "Bar" );
            }
        } );
    }
}
```

Call hierarchyt on `Action2.doit()` shows only `test22()`. So the call hierarchy doesn't seem to concern itself with the case of calling `action1(Action1)` with an instance of e.g. `Sub implements Action1, Action2`. Also replacing `interface` with `abstract class` results in the same unexpected contents under `callers`.

[...]

PR #866, opened 2023-03-19
Call Hierarchy identifies incompatible methods with same signature

trancexpress:
Given an anonymous type defined in a method, the call hierarchy of a
method in the anonymous type can contain wrong entries. This is due to
`SearchPattern.enclosingTypeNames()` adding `*` symbols to a part of the
search pattern, for anonymous types defined in a method. The `*` symbol
then matches more inner types than actual matches.

This change replaces the `*` symbol with type names, when possible. This
ensures no unexpected matches are seen in the call hierarchy.

Fixes: #856

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchBugsTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchBugsTests.java
@@ -13141,7 +13141,7 @@ public void testBug345807() throws CoreException {
 	}
 }
 // https://bugs.eclipse.org/bugs/show_bug.cgi?id=355605
-public void testBug355605() throws CoreException {
+public void testBug355605_1() throws CoreException {
 	try {
 	createJavaProject("P");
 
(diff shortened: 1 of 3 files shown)


trancexpress:
Looking at the failing test `JavaSearchBugsTests.testBug355605()`, `declaringQualification` is set to: `[X.R.*.*]`
(I don't find documentation for it)

The `*` symbol is coming from the method handling:

```
		case IJavaElement.METHOD:
			IType declaringClass = ((IMember) parent).getDeclaringType();
			return CharOperation.arrayConcat(
				enclosingTypeNames(declaringClass),
				new char[][] {declaringClass.getElementName().toCharArray(), IIndexConstants.ONE_STAR});
```

If we can check what type is being instantiated, maybe we can replace `*` with that type name...

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchBugsTests.java
```diff
@@ -13182,6 +13182,44 @@ public void testBug355605() throws CoreException {
...
+						+ "class R {\n"
+						+ "   class S {\n"
+						+ "     S(String s) {}\n"
+						+ "   	void setInfo(String x) {\n"
+						+ "   	}\n"
+						+ "   }\n"
+						+ "   class T {\n"
+						+ "   }\n"
+						+ "	T t = new T()  {\n"
+						+ "		S s = new S(\"test\") {\n"
```
trancexpress, line 13200: As commented on the PR, the string we are changing must match `X.R.1.new T(){}.1.new S(){}` during this test.

I've added a paranoia test (copy of the one above) with an argument in the constructor of `S`. Just to be sure the argument doesn't make it into the string `X.R.1.new T(){}.1.new S(){}`. It doesn't, but best to be sure with a test.

org.eclipse.jdt.core/search/org/eclipse/jdt/core/search/SearchPattern.java
```diff
@@ -2453,10 +2453,19 @@ private static char[][] enclosingTypeNames(IType type) {
 		case IJavaElement.FIELD:
 		case IJavaElement.INITIALIZER:
 		case IJavaElement.METHOD:
+			char[] typeName = IIndexConstants.ONE_STAR;
+			try {
+				String superclassName = type.getSuperclassName();
```
trancexpress, line 2458: Since we have an anonymous class, `getSuperclassName()` yields the interface or class that is being implemented resp. extended. No need to check interfaces (which I initially thought we also had to do).

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchBugsTests2.java
```diff
@@ -2894,6 +2894,54 @@ public void testOutOfBoundsIndexExceptionDuringSearchGh825() throws Exception {
...
 		}
 	}
+	public void testCallHierarchyAnonymousInnerTypeGh856() throws CoreException, IOException {
+		try {
+			IJavaProject p = createJavaProject("P", new String[] { "src" }, new String[] { "JCL11_LIB" }, "bin", "11");
+			createFolder("/P/src/test");
+			createFile("/P/src/test/X.java", String.join(System.lineSeparator(), new String[] {
+					"package test;\n",
+					"public class X {\n",
+					"    public static interface Action1 {\n",
```
gayanper, line 2904: what if we have a method or a field definition before the first member type definition ? does this test pass still. I'm working about the usage of `type.getOccurenceCount()` ?

gayanper:
@iloveeclipse other than pointed out comment, the changes LGTM.

org.eclipse.jdt.core/search/org/eclipse/jdt/core/search/SearchPattern.java
gayanper, line 2458: I would really like if you can add some code comments describing what we are doing here and why for future us or another developer :)

PR #768, opened 2023-02-19
content assist not working for method with unresolved arguments

snjeza:
Fixes #767

PR #867, opened 2023-03-20
Fix the require bundle version for compiler.batch

jarthana:
## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/.settings/.api_filters
+++ b/org.eclipse.jdt.core.compiler.batch/.settings/.api_filters
@@ -3,7 +3,7 @@
     <resource path="META-INF/MANIFEST.MF">
         <filter id="924844039">
             <message_arguments>
-                <message_argument value="3.33.100"/>
+                <message_argument value="3.33.50"/>
                 <message_argument value="3.33.0"/>
             </message_arguments>
         </filter>
--- a/org.eclipse.jdt.core.compiler.batch/META-INF/MANIFEST.MF
+++ b/org.eclipse.jdt.core.compiler.batch/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Main-Class: org.eclipse.jdt.internal.compiler.batch.Main
 Bundle-ManifestVersion: 2
 Bundle-Name: Eclipse Compiler for Java(TM)
 Bundle-SymbolicName: org.eclipse.jdt.core.compiler.batch
-Bundle-Version: 3.33.100.qualifier
+Bundle-Version: 3.33.50.qualifier
 Bundle-ClassPath: .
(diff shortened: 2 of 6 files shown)


jarthana:
@iloveeclipse Is this okay or do you think we should go for a range?

iloveeclipse:
> Is this okay or do you think we should go for a range?

I honestly would like to uunderstand the problem first before discussing a solution :-)
Do we have a ticket describing the problem?

jarthana:
> > Is this okay or do you think we should go for a range?
> 
> I honestly would like to uunderstand the problem first before discussing a solution :-) Do we have a ticket describing the problem?

The problem is, when I upped the version numbers for the BETA branch, I didn't update the bundle dependency version. Then someone pointed out that it is a tightly coupled to the current version and not a range. I am happy with just fixing the version, but just not sure if this tight dependency is intentional.

iloveeclipse:
If I understand it right, jdt core should have strict dependency relationship with batch compiler. 
It can be a range too, but range should set minimal ecj version to latest "tested" one. On beta branch it should be also higher as on master, isn't?

jarthana:
> If I understand it right, jdt core should have strict dependency relationship with batch compiler. It can be a range too, but range should set minimal ecj version to latest "tested" one. 

OK, actually on BETA, it makes no sense to have a different version. We added quite a few stuff in compiler.batch that will be used in jdt.core. So, looks like it's very unlikely we will ever have a range.

>On beta branch it should be also higher as on master, isn't?
We are going to be making patches for 4.27 off BETA branch. So, it only needs to be higher than 4.27 and preferably lower than master. Once we merge the BETA changes to master, we will bump the master anyway to 3.34 as there will be quite a few API changes.

iloveeclipse:
> we will bump the master anyway to 3.34 as there will be quite a few API changes.

Should we do that on master now?

> I also noticed that in master we have 3.33.101 in jdt.core and 33.100 in compiler.batch.

This is a mistake, see https://github.com/eclipse-jdt/eclipse.jdt.core/pull/768

jarthana:
> > we will bump the master anyway to 3.34 as there will be quite a few API changes.
> 
> Should we do that on master now?

We can if there are already API changes in master. Otherwise, since the Java GA (and hence the merge) is only a couple of days away, we can take it up at that time. I will shortly create a patch that needs to be pushed after the merge.

ktatavarthi:
> If I understand it right, jdt core should have strict dependency relationship with batch compiler. It can be a range too, but range should set minimal ecj version to latest "tested" one. On beta branch it should be also higher as on master, isn't?

If ther is going to be a tightly coupled dependency between jdt.core and batch compiler, this information needs to be documented as som eplace so that who ever is updating the batch compiler version does not miss updating the version in  the dependencies of jdt.core

Issue #193, opened 2022-07-06
Content assist does not work after switch and lambda


jubax:
If you have
```
import java.nio.file.StandardCopyOption;

@SuppressWarnings("nls")
public class EclipseTest {
    
    private void a(StandardCopyOption option) throws Exception {
	switch (option) {
	case ATOMIC_MOVE:
	    break;
	}
    }

    public void b() throws Exception {
	Runnable r = () -> {
	};

    }

    private fi  <==== Empty content assist box 
    
}
```

If you add the `private fi` and autocomplete then several suggestions are shown.

<img width="282" alt="image" src="https://user-images.githubusercontent.com/10562923/177520625-81ea4c40-d428-49d5-b848-8382701cec69.png">

vs

<img width="221" alt="image" src="https://user-images.githubusercontent.com/10562923/177520721-a2d66501-39bb-4a5d-8c07-c7f6ea5bfbc6.png">

Note that it depends on the other statements in the file, the content assist usually works fine for members at the end of a file.

ktatavarthi:
This was working in 4.16 Eclipse. This is a regression.

jubax:
This content assist regression still exists in 2022-09 (4.25.0).

jubax:
I debugged a bit and I found that
```
import java.nio.file.StandardCopyOption;

@SuppressWarnings("nls")
public class EclipseTest {
    
    private void a(StandardCopyOption option) throws Exception {
	switch (option) {
	case ATOMIC_MOVE:
	    break;
	}
        Runnable r = () -> {};
        fina<==== Empty content assist box 
    }
    
}
```
is sufficient to reproduce the error. The completion AST for that program is
```
import java.nio.file.StandardCopyOption;
public @SuppressWarnings("nls") class EclipseTest {
  public EclipseTest() {
  }
  private void a(StandardCopyOption option) throws Exception {
    Runnable r;
    <CompleteOnName:fina>;
  }
}
```
and at some point we land in `AssistParser.fallBackToSpringForward()`. In that method the current token is `TokenNameBeginCaseExpr = 72` and `this.referenceContext` is a MethodDeclaration `() {}`, so something went wrong with the lambda.

After that `CompletionScanner.ungetToken()` is called, which leads to a "Single cell array overflow" `ArrayIndexOutOfBoundsException`.

I'm not sure if I can really make progress from here, but maybe this comment helps anyone who actually understands what is going on.

jubax:
[...]

PR #878, opened 2023-03-21
Content assist does not work after switch and lambda

trancexpress:
The fix for bug 565844 introduces a regression, with a switch and a
lambda in different methods, followed by an incomplete method
definition.

This is due to Scanner.disambiguateArrowWithCaseExpr() including the
switch case statement when scanning the lambda expression. Specifically,
Scanner.caseStartPosition is set in Scanner.updateCase() but is never
set back to -1 once scanning the switch expression is done.

This change adjusts Scanner.mayBeAtCaseLabelExpr() to check
Scanner.lookBack for lambda definition symbols, "-> (". If so, assume no
case label expression. This prevents confusing the lambda for an arrow
case of a switch expression.

Fixes: #193

## Author checklist


trancexpress:
> `Scanner.updateCase(int)` is setting `Scanner.caseStartPosition`, maybe a proper fix is to set `caseStartPosition` to `-1` when the `switch` statement goes out of scope...

IMO its better to do this, but I have no idea how.

iloveeclipse:
Code changed in cb6774f3ba36fb8dafc3a4834eb22799072db6ef & I can't reproduce this bug anymore on I20240507-2220.

Issue #649, opened 2023-01-18
Prevent imports/package from being moved above first comments


fbricon:
this is a follow up on https://github.com/redhat-developer/vscode-java/issues/1993.

JBang scripts use directives in java files taking the form of single line comments. eg:

```java
///usr/bin/env jbang "$0" "$@" ; exit $?

public class hello {

  public static void main(String[] args) {
    if (args != null) {
      Arrays.stream(args).forEach(System.out::println);
    } else {
      System.out.println("no args");
    }
  }
}
```

here, the `///usr/bin/env jbang "$0" "$@" ; exit $?` needs to stay on the 1st line, as it's interpreted by bash/shell when invoked from CLI (see https://www.jbang.dev/documentation/guide/latest/faq.html).

When you organize imports or use autocompletion for `Arrays`, `import java.utils.Arrays` is added on the 1st line, instead of keeping the JBang directive there.

```java
import java.util.Arrays;
///usr/bin/env jbang "$0" "$@" ; exit $?


public class hello {

  public static void main(String[] args) {
    if (args != null) {
      Arrays.stream(args).forEach(System.out::println);
    } else {
      System.out.println("no args");
    }
  }
}
``` 

So, it'd be nice if the JDT formatter didn't touch those 1st comments. @rgrunber proposed the following patch, that might fix this issue:

> I think the entry point to fix this (or add an option to change the behaviour) should be at https://github.com/eclipse-jdt/eclipse.jdt.core/blob/c58f581d8cef2c38a339b8056b3f48f1d5365731/org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/rewrite/imports/ImportRewriteAnalyzer.java#L317 .
> 
> If there's no existing import region (no imports), and no existing package declaration (default package), then find the offset of the first non-comment node and begin from there.
>
> This pretty basic patch seems to do the job.

```diff
diff --git a/org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/rewrite/imports/ImportRewriteAnalyzer.java b/org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/rewrite/imports/ImportRewriteAnalyzer.java
index 200be49a4f..cd29dfcc5e 100644
--- a/org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/rewrite/imports/ImportRewriteAnalyzer.java
[...]

PR #889, opened 2023-03-23
Don't put package or import statements ahead of header comments

jjohnstn:
- modify CopyResourceElementsOperation.updatePackageStatement() to add any header comments to package statement if no package statement existed previously and delete the original comments so there aren't duplicates
- modify ImportRewriteAnalyzer.determineSurroundingRegion() to calculate start of imports to be after any header comments if no package statement currently exists
- for #649

## What it does

Fixes organize imports and move resource to preserve header comments when adding imports to a class that is in the default package or moving a class from the default package to another package that will end up adding package statement.  See issue.

## How to test

See examples in issue.  Tests will be added to jdt.ui once this patch is merged.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/rewrite/imports/ImportRewriteAnalyzer.java
+++ b/org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/rewrite/imports/ImportRewriteAnalyzer.java
@@ -316,13 +316,15 @@ public final class ImportRewriteAnalyzer {
 			}
 			else {
 				surroundingStart = 0;
-				ASTNode topNode = nodesTreeMap.firstEntry().getValue();
-				if (topNode instanceof Comment) {
-					for (ASTNode node : nodesTreeMap.values()) {
-						if (!(node instanceof Comment)) {
-							break;
-						} else {
-							surroundingStart = node.getStartPosition() + node.getLength();
+				if (!nodesTreeMap.isEmpty()) {
+					ASTNode topNode = nodesTreeMap.firstEntry().getValue();
+					if (topNode instanceof Comment) {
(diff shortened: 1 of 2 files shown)


org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/rewrite/imports/ImportRewriteAnalyzer.java
```diff
@@ -315,6 +316,16 @@ private static IRegion determineSurroundingRegion(CompilationUnit compilationUni
 			}
 			else {
 				surroundingStart = 0;
+				ASTNode topNode = nodesTreeMap.firstEntry().getValue();
```
rgrunber, line 319: I can't remember exactly, but I think with the old patch, I noted that `firstEntry()` here may be `null` when there is no entry in the comment mapping. I would just double check that. I don't think I updated my patch after I discovered it since it was just a quick POC.

jjohnstn: The node mapping puts non-comment nodes in the list as well such as types, imports, package statements, then sorts based on start position.  So, a class without a comment should have the first entry being a package or import or type.

jjohnstn: Actually, after testing, an empty file will cause an NPE.  I have added the check.

jjohnstn: it turned out the fix gets an NPE if you tried to organize imports in an empty file.  I have already added the check.

Issue #917, opened 2023-03-28
NPE in org.eclipse.jdt.apt.core.util.AptConfig#addProjectClasspath


ivy-lli:
The is a possible NPE in the org.eclipse.jdt.apt.core.util.AptConfig#addProjectClasspath method: https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/org.eclipse.jdt.apt.core/src/org/eclipse/jdt/apt/core/util/AptConfig.java#L368

The [IResource#getLocation](https://help.eclipse.org/latest/index.jsp?topic=%2Forg.eclipse.platform.doc.isv%2Freference%2Fapi%2Forg%2Feclipse%2Fcore%2Fresources%2FIResource.html) method can return null as here seen:

```
java.lang.NullPointerException: Cannot invoke "org.eclipse.core.runtime.IPath.toOSString()" because the return value of "org.eclipse.core.resources.IResource.getLocation()" is null
	at org.eclipse.jdt.apt.core.util.AptConfig.addProjectClasspath(AptConfig.java:368)
	at org.eclipse.jdt.apt.core.util.AptConfig.getProcessorOptions(AptConfig.java:255)
	at org.eclipse.jdt.internal.apt.pluggable.core.dispatch.IdeProcessingEnvImpl.getOptions(IdeProcessingEnvImpl.java:86)
	at org.hibernate.jpamodelgen.Context.<init>(Context.java:71)
	at org.hibernate.jpamodelgen.JPAMetaModelEntityProcessor.init(JPAMetaModelEntityProcessor.java:76)
	at org.eclipse.jdt.internal.apt.pluggable.core.dispatch.IdeAnnotationProcessorManager.discoverNextProcessor(IdeAnnotationProcessorManager.java:97)
	at org.eclipse.jdt.internal.compiler.apt.dispatch.RoundDispatcher.round(RoundDispatcher.java:119)
	at org.eclipse.jdt.internal.compiler.apt.dispatch.BaseAnnotationProcessorManager.processAnnotations(BaseAnnotationProcessorManager.java:172)
	at org.eclipse.jdt.internal.apt.pluggable.core.dispatch.IdeAnnotationProcessorManager.processAnnotations(IdeAnnotationProcessorManager.java:138)
	at org.eclipse.jdt.internal.compiler.Compiler.processAnnotations(Compiler.java:950)
	at org.eclipse.jdt.internal.compiler.Compiler.compile(Compiler.java:450)
[...]

PR #918, opened 2023-03-28
Issue 917: Fix NPE in AptConfig#addProjectClasspath

ivy-lli:
## What it does
Fixes issue: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/917

## How to test
Check: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/917

## Author checklist


org.eclipse.jdt.apt.core/src/org/eclipse/jdt/apt/core/util/AptConfig.java
```diff
@@ -364,8 +364,8 @@ private static void addProjectClasspath(
     		IPath binPath = otherJavaProject.getOutputLocation();
     		IResource binPathResource = root.findMember(binPath);
     		String binDirString;
-    		if (binPathResource != null) {
-    			binDirString = root.findMember(binPath).getLocation().toOSString();
+    		if (binPathResource != null && binPathResource.getLocation() != null) {
```
iloveeclipse, line 367: The fix is formally correct.
However, the `else` branch will still use same (not existing) path.
Why should the `else` branch work? Because someone creates the path **later on**?

ivy-lli: Why is the path in the else branch not existing?
The NPE was in the if branch which means that the binPath exists and also a member was found (binPathResource).

Or do you mean that if the binPathResource is found but the location can not be determined, the method should be aborted (by return)?

iloveeclipse: `binPath` is just a string value that we are trying to resolve via `root.findMember(binPath).getLocation()`.

So if there are no path variables used in the location, it is **same** path.

So if the `binPathResource.getLocation()` can't be resolved, it means it is either doesn't exist (anymore?) or path variable is wrong, or it is not local one.

So the "else" branch **seem** to be wrong in all of such cases.

=> therefore I wonder if you can provide the case where it would work.

iloveeclipse (changes requested):
OK, that makes more sense now, just some polishing needed.

1) Please add this line to the end of the commit message:
`Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/917`
2) See comments in code.

iloveeclipse (approved):

iloveeclipse:
Thanks.

Issue #401, opened 2022-09-15
No search results for inner class constructor with first argument of outer type


iloveeclipse:
1) Part one: on **Java 11** I can't get any match trying to see **callers** or **references** of all inner class constructors inside `java.util.SplittableRandom` class, like
`java.util.SplittableRandom.RandomDoublesSpliterator.RandomDoublesSpliterator(SplittableRandom, long, long, double, double)`.
Neither call hierarchy nor search for references on the constructors show any match.
2) Part two: on **Java 17** (where the code in question is changed) I can't get any match trying to see **callers** (but I get references!) of all inner class constructors inside `java.util.SplittableRandom` class, like
`java.util.SplittableRandom.AbstractSplittableGeneratorProxy.AbstractSplittableGeneratorProxy(SplittableRandom)`.
Call hierarchy doesn't find, but search for references find the constructor.

The second part will be fixed with #333, but the first part not.
So this (part 1) seem to be different to #332 issue and need another fix.

trancexpress:
Looks like the `SearchPattern` has `ConstructorPattern.parameterCount` set to 4, but the constructor has 5 arguments. So the calls to the constructor are ignored by the search. The argument count is decremented with this stack trace:

```
"Worker-11: Fetching children of RandomDoublesSpliterator" #97 prio=5 os_prio=0 cpu=38.88ms elapsed=158.86s tid=0x00007f8c9c001e90 nid=0x122b4 runnable  [0x00007f8cef5fe000]
   java.lang.Thread.State: RUNNABLE
        at org.eclipse.jdt.internal.core.search.matching.ConstructorPattern.<init>(ConstructorPattern.java:334)
        at org.eclipse.jdt.internal.core.search.matching.ConstructorPattern.<init>(ConstructorPattern.java:366)
        at org.eclipse.jdt.core.search.SearchPattern.createPattern(SearchPattern.java:2209)
        at org.eclipse.jdt.internal.corext.callhierarchy.CallerMethodWrapper.findChildren(CallerMethodWrapper.java:144)
        at org.eclipse.jdt.internal.corext.callhierarchy.MethodWrapper.performSearch(MethodWrapper.java:266)
        at org.eclipse.jdt.internal.corext.callhierarchy.MethodWrapper.doFindChildren(MethodWrapper.java:197)
        at org.eclipse.jdt.internal.corext.callhierarchy.MethodWrapper.getCalls(MethodWrapper.java:88)
[...]

PR #938, opened 2023-04-02
No search results for inner class constructor with outer type argument

trancexpress:
Constructors of inner types have a synthetic first argument, if the inner type is not static. Both the search and indexer code have various places where its not checked whether the inner type is static. Instead, the type name of the first argument of the constructor is checked - if it matches the name of the outer type, the first argument is assumed to be synthetic.

This causes issues when searching for references of a static inner type constructor which takes the outer type as the first argument. There is no synthetic argument in the constructor, but the search/index code assumes there is and so constructor invocations are not evaluated as matching (due to an argument count mismatch).

This change adjusts several such places that result in inaccurate searches, as well as adds a test case. The disk index version is also incremented, since existing stored indices are wrong w.r.t. static inner type constructors that take the outer type as their first argument.

Fixes: #401

## Author checklist


gayanper (approved):
LGTM

PR #939, opened 2023-04-03
Store declaring type arguments info when creating constructor completion

snjeza:
Rebased Gerrit change https://git.eclipse.org/r/c/jdt/eclipse.jdt.core/+/191217 
Author: jinbwan@microsoft.com

Related issues:
- https://github.com/eclipse/eclipse.jdt.ls/pull/2535#issuecomment-1493803864
- https://bugs.eclipse.org/bugs/show_bug.cgi?id=578969

It's cheap to get the type variable names and the compatibility info when creating constructor completion proposal in CompletionEngine. This can be used in the later completion processing to quick check if the diamond operator is needed.
In addition to storing type arguments, I also save isCompatibleProposal in CompletionProposal. See
LazyGenericTypeProposal#computeTypeArgumentProposals(), if proposed type does not inherit from expected type, it doesn't add any type arguments. The compatibility info can be reused later to quickly skip those incompatible proposals in computeTypeArgumentProposals().

### DIFF
--- a/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
+++ b/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
@@ -14158,9 +14158,20 @@ public final class CompletionEngine
 		relevance += computeRelevanceForInterestingProposal();
 		relevance += computeRelevanceForRestrictions(accessibility);
 		relevance += computeRelevanceForCaseMatching(this.completionToken, simpleTypeName);
-		relevance += computeRelevanceForExpectingType(typeBinding);
+		int relevanceForExpected = computeRelevanceForExpectingType(typeBinding);
+		relevance += relevanceForExpected;
 		relevance += computeRelevanceForQualification(isQualified);
 		relevance += computeRelevanceForConstructor();
+
+		boolean isCompatibleType = this.expectedTypesPtr < 0 || relevanceForExpected >= R_EXPECTED_TYPE;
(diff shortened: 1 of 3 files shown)


org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
```diff
@@ -14246,6 +14265,8 @@ private void proposeConstructor(
 						proposal.setReplaceRange(this.endPosition - this.offset, this.endPosition - this.offset);
 						proposal.setTokenRange(this.tokenStart - this.offset, this.tokenEnd - this.offset);
 						proposal.setRelevance(relevance);
+						proposal.setDeclarationTypeVariables(typeVariables);
+						proposal.setCompatibleProposal(isCompatibleType);
```
rgrunber, line 14269: Is there a reason why we don't call  `setDeclarationTypeVariables(..)` & `setCompatibleProposal(..)` for the default case (constructor with parameter) (in the else) ?

snjeza: Fixed.

rgrunber (approved):
I would schedule to merge this as soon as we open up for 4.29M1 as this isn't really a high priority item for JDT itself. It would mainly help JDT-LS for any clients that don't opt into the aggressive `resolve` logic that we now perform (ie. only produce the final completion text if selected, and not immediately on sending the full list).

PR #946, opened 2023-04-05
Issue eclipse-jdt#994 - Fixed external annotation when using JSR199 c…

### LINKED ISSUE #944: -annotationpath does not work with JSR199 EclipseCompiler
ECF allows to pass the path to external null annotation via the `-annotationpath` argument. This works perfectly when using the "batch" compiler (`org.eclipse.jdt.internal.compiler.batch.Main`). However, it does not work when using the JSR199 compiler (`org.eclipse.jdt.internal.compiler.tool.EclipseCompiler`). The latter is used by the `plexus-compiler-eclipse` Maven plug-in which allows using ECF in Maven builds. This means you cannot use annotation-based null analysis in Maven right now (javac does not support it and ECF ignores the external annotation).
A simple reproducer is [test.zip](https://github.com/eclipse-jdt/eclipse.jdt.core/files/11157640/test.zip)

The problem is in `org.eclipse.jdt.internal.compiler.tool.EclipseCompilerImpl.getCompilationUnits()` which overrides the same method from `Main` (the batch compiler). In [line 177](https://github.com/eclipse-jdt/eclipse.jdt.core/blob/457e2221c44b49b6f38fcdfe6d1d59467a04bf75/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/tool/EclipseCompilerImpl.java#L177) a `CompilationUnit` is created using a [constructor](https://github.com/eclipse-jdt/eclipse.jdt.core/blob/457e2221c44b49b6f38fcdfe6d1d59467a04bf75/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/CompilationUnit.java#L58) that sets the _annotationPathProvider_ to `null` and hence disabling the external annotations.

sithmein:
…ompiler

The external annotations path was not set on any compilation unit or classpath entry even though it was passed on the command line.

## What it does
It fixes #944.

## How to test
The issue contains a minimal reproducer.

## Author checklist


iloveeclipse:
@sithmein : please also [sign ECA](https://accounts.eclipse.org/legal/eca/validation/148742) and if already signed, please make sure mail used to sign commit is same used to sign ECA.

sithmein:
I revalidated the ECA.
I don't think `NullAnnotationBatchCompilerTest` is the right place because the batch compiler works without issues. It's only the JSR199 compiler that doesn't work.

iloveeclipse:
> I revalidated the ECA. 

Perfect.

> I don't think `NullAnnotationBatchCompilerTest` is the right place because the batch compiler works without issues. It's only the JSR199 compiler that doesn't work.

So what about `CompilerToolTests`?

sithmein:
Replaced by https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1120

Issue #954, opened 2023-04-10
Bug: module binding error message does not render correctly in diagnostics


ascopes:
```
8389908 - Cannot bind message for problem (id: 1300) "{0} cannot be resolved to a module" with arguments: {}
```

Appears that https://sourcegraph.com/github.com/eclipse-jdt/eclipse.jdt.core@17d644ffc8a0fc262a20baaac0e390b55076a088/-/blob/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/problem/ProblemReporter.java#L11448 is not passing an array of problem arguments into the handler, despite one parameter being required.

I'll put a PR up to fix this.

iloveeclipse:
The new test fails: https://download.eclipse.org/eclipse/downloads/drops4/I20230412-0120/testresults/html/org.eclipse.jdt.compiler.tool.tests_ep428I-unit-macM1-java17_macosx.cocoa.aarch64_17.html

PR #955, opened 2023-04-10
Fix GH-954: module binding error message not rendering correctly.

ascopes:
## What it does
Fixes GH-954: a message rendering issue for module resolution issues.

## How to test
Right now, I am integrating with the compiler from a project of mine, and I am coming across
the following diagnostic being output:

```
8389908 - Cannot bind message for problem (id: 1300) "{0} cannot be resolved to a module" with arguments: {}
```

I am still looking for the actual trigger of this error in my code, but it appears that the
log message is misconfigured.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.compiler.tool.tests/src/org/eclipse/jdt/compiler/tool/tests/CompilerToolTests.java
+++ b/org.eclipse.jdt.compiler.tool.tests/src/org/eclipse/jdt/compiler/tool/tests/CompilerToolTests.java
@@ -1416,6 +1416,30 @@ static final String[] FAKE_ZERO_ARG_OPTIONS = new String[] {
 				);
 	}
 
+	/**
+	 * See: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/954
+	 */
+	public void testModuleCannotBeResolvedErrorMessageGh955() throws Exception {
+		suppressTest(
+				"module-info.java",
+				"""
+				module testmodule {
+					requires aaa;
+				}
+				""",
+				"ERROR 2: aaa cannot be resolved to a module",
+				"""
+				----------
+				1. ERROR in /tmp/module-info.java (at line 2)
+					requires aaa;
+					         ^^^
+				aaa cannot be resolved to a module
+				----------
+				1 problem (1 error)
+				"""
+				);
+	}
+
 	public void testSupportedCompilerVersions() throws IOException {
 		Set<SourceVersion> sourceVersions = compiler.getSourceVersions();
 		SourceVersion[] values = SourceVersion.values();
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/problem/ProblemReporter.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/problem/ProblemReporter.java
@@ -75,6 +75,8 @@
  *								bug 527554 - [18.3] Compiler support for JEP 286 Local-Variable Type
  *     Ulrich Grave <ulrich.grave@gmx.de> - Contributions for
  *                              bug 386692 - Missing "unused" warning on "autowired" fields
+ *     Ashley Scopes - Contributions for
+ * 								GH-954 	   - Module binding error renders incorrectly for diagnostics
  ********************************************************************************/
 package org.eclipse.jdt.internal.compiler.problem;
 
@@ -11446,9 +11448,8 @@ public void invalidTypeArguments(TypeReference[] typeReference) {
 			typeReference[typeReference.length - 1].sourceEnd);
 }
 public void invalidModule(ModuleReference ref) {
-	this.handle(IProblem.UndefinedModule,
-		NoArgument, new String[] { CharOperation.charToString(ref.moduleName) },
-		ref.sourceStart, ref.sourceEnd);
+	String[] args = new String[] { CharOperation.charToString(ref.moduleName) };
+	this.handle(IProblem.UndefinedModule, args, args, ref.sourceStart, ref.sourceEnd);
 }
 public void missingModuleAddReads(char[] requiredModuleName) {
 	String[] args = new String[] { new String(requiredModuleName) };


ascopes:
ECA signed:
image

trancexpress:
Can you add a test here, or provide a reproducer? Something like this:

```
diff --git a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ModuleBuilderTests.java b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ModuleBuilderTests.java
index 422cfbeb15..bd6677a57c 100644
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ModuleBuilderTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ModuleBuilderTests.java
@@ -9061,6 +9061,28 @@ public class ModuleBuilderTests extends ModifyingResourceTests {
                        deleteProject(p1);
                }
        }
+       /**
+        * See: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/954
+        */
+       public void testModuleCannotBeResolvedErrorMessageGh955() throws Exception {
+               try {
+                       IJavaProject p = setupModuleProject("testGh955", new String[] {
+                                       "src/module-info.java",
+                                       "module testmodule {\n" +
+                                       "       requires aaa;\n" +
+                                       "}\n",
+                       });
+
+                       ResourcesPlugin.getWorkspace().build(IncrementalProjectBuilder.FULL_BUILD, null);
+                       waitForAutoBuild();
+
+                       IMarker[] markers = p.getProject().findMarkers(null, true, IResource.DEPTH_INFINITE);
+                       assertMarkers("Unexpected markers", "aaa cannot be resolved to a module",
+                                       markers);
+               } finally {
+                       deleteProject("testGh955");
+               }
+       }
        protected void assertNoErrors() throws CoreException {
                for (IProject p : getWorkspace().getRoot().getProjects()) {
                        int maxSeverity = p.findMaxProblemSeverity(null, true, IResource.DEPTH_INFINITE);
```

Note that the above test passes without the change here. I'm not sure how to run into #954.

ascopes:
@trancexpress outside of a project that is invoking this, I am not sure of the best way of reproducing it, since my use case is building from a custom diagnostic listener and file manager. #958 seems to suggest to me that this uses a totally different code flow to the standard configuration, as there are branches that detect if the default components are being used elsewhere that I can spot. However, inspecting the existing source code:

```
public void invalidModule(ModuleReference ref) {
	this.handle(IProblem.UndefinedModule,
		NoArgument, new String[] { CharOperation.charToString(ref.moduleName) },
		ref.sourceStart, ref.sourceEnd);
}
```
the message arguments are being provided correctly, but not the problem arguments.

It seems that the output ECJ is emitting usually makes use of the formatted string using the message arguments, but using the JSR-199 API with a DiagnosticListener, calling `Diagnostic.getMessage(Locale.ROOT)` produces the output described in the issue.

This appears tobe because `EclipseCompilerImpl$Jsr199ProblemWrapper` has a `diagnostic` field, which, upon being inspected in a debugger, is using the empty array problemArguments rather than the populated messageArguments.

image

Line 354 of EclipseCompilerImpl defines the code for JSR-199's `getMessage(Locale)` as this:

image

Which shows it is using the empty array that has been changed in this PR.

If you can advise of a good way to replicate this environment in a test pack, I'll be happy to try and push a test for this.

trancexpress:
> If you can advise of a good way to replicate this environment in a test pack, I'll be happy to try and push a test for this.

I see, please add this test:

```
diff --git a/org.eclipse.jdt.compiler.tool.tests/src/org/eclipse/jdt/compiler/tool/tests/CompilerToolTests.java b/org.eclipse.jdt.compiler.tool.tests/src/org/eclipse/jdt/compiler/tool/tests/CompilerToolTests.java
index becccd0394..671308880b 100644
--- a/org.eclipse.jdt.compiler.tool.tests/src/org/eclipse/jdt/compiler/tool/tests/CompilerToolTests.java
+++ b/org.eclipse.jdt.compiler.tool.tests/src/org/eclipse/jdt/compiler/tool/tests/CompilerToolTests.java
@@ -1416,6 +1416,30 @@ static final String[] FAKE_ZERO_ARG_OPTIONS = new String[] {
 				);
 	}
 
+	/**
+	 * See: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/954
+	 */
+	public void testModuleCannotBeResolvedErrorMessageGh955() throws Exception {
+		suppressTest(
+				"module-info.java",
+				"""
+				module testmodule {
+					requires aaa;
+				}
+				""",
+				"ERROR 2: aaa cannot be resolved to a module",
+				"""
+				----------
+				1. ERROR in /tmp/module-info.java (at line 2)
+					requires aaa;
+					         ^^^
+				aaa cannot be resolved to a module
+				----------
+				1 problem (1 error)
+				"""
+				);
+	}
+
 	public void testSupportedCompilerVersions() throws IOException {
 		Set<SourceVersion> sourceVersions = compiler.getSourceVersions();
 		SourceVersion[] values = SourceVersion.values();
```

trancexpress:
For the ECA, can you try adding a line like this at the end of the commit message:

```
Signed-off-by: ... <...@...>
```

E.g. if I were signing the commit, the line would look like this:

```
```

If you are using the Eclipse Git Staging view, you can amend the commit and use the "Add Signed-off-by" view button (assuming you have set `user.name` and `user.mail` for your forked git repository).

I'm not sure what hooks there are to try to identify your email, but it should be possible to use the line above too (unless I'm out of touch with the GitHub contribution process).

It could also be that GitHub needs some time to "see" the ECA is signed, I'm not sure.

ascopes:
@trancexpress updated the commit with your patch.

Also looks like the signoff worked as well, which is good. Thanks for all the help!

trancexpress:
@iloveeclipse this looks like a simple enough change. Can you review it?

trancexpress:
Thanks for the fix @ascopes !

Issue #795, opened 2023-02-27
No "Javadoc: Missing tag for return type" for "{@return ...}"


howlger:
For methods with a Javadoc comment containing an inline _return_ (**`{@return ...}`**), no **_Javadoc: Missing tag for return type_** problem should be displayed when _Project > Properties: Java Compiler > Javadoc: Missing Javadoc tags_ is set to _Error_, _Warning_ or _Info_.

```java
/** Sample */
public class Sample {

	/** {@return 42} */
	public int sample() {
		return 42;
	}

}
```

Tested in Eclipse 2023-03 M3:

<img alt="screenshot" src="https://user-images.githubusercontent.com/5197918/221596522-74780bd2-babf-4647-ad71-01d711b820ce.png" width="459">

See also:
- https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/340
- [Eclipse 4.27 - New and Noteworthy - Java development tools - Javadoc inline \@return](https://www.eclipse.org/eclipse/news/4.27/jdt.php#javadoc-inline-return)

iloveeclipse:
@mpalat, @jarthana : after quickly looking in debugger, I assume inline return tags are not parsed in `org.eclipse.jdt.internal.compiler.parser.AbstractCommentParser.commentParse()`.
Is this correct?

jarthana:
The reason appears to be this code in JavadocParser:

```
  if (!this.inlineTagStarted) {
    valid = parseReturn();
  }
```

I had no clue that I added this code 13 years ago!

If return tags are accepted as inline tags, then we should perhaps remove this check.

mpalat:
@jarthana moving this out of M1

iloveeclipse:
From #978 :
> I figured these would be a problem during my initial assessment. Let me see what we can do here.

@jarthana : do you plan to work on this? If not, please unset milestone.

jarthana:
> From #978 :
> 
> > I figured these would be a problem during my initial assessment. Let me see what we can do here.
> 
> @jarthana : do you plan to work on this? If not, please unset milestone.

Sorry, I don't yet have any reasonable response to the open questions. I am afraid we will have to move this out.

iloveeclipse:
See also following bugs related to missing Javadoc features:
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1026
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1027

PR #978, opened 2023-04-17
"Missing tag for return type" javadoc error reported for inline return

iloveeclipse:
- Updated JavadocParser code to allow return tags to be "inlined".
- Relaxed the error check for inline/block return tags mismatch added via  https://bugs.eclipse.org/bugs/show_bug.cgi?id=267833
- added regression tests

Note: because of the JLS change, JavadocTagConstants.JAVADOC_TAG_TYPE array doesn't work anymore as intended, because there are now tags that can be tags that support *both* block or inline style.

See https://docs.oracle.com/en/java/javase/17/docs/specs/javadoc/doc-comment-spec.html#return

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/795

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/JavadocParser.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/JavadocParser.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2022 IBM Corporation and others.
+ * Copyright (c) 2000, 2023 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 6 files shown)


iloveeclipse:
- `JavadocCompletionParserTest.test001` and other tests in same class fail because of unexpected return tag found. This can be "fixed" by removing added `TAG_RETURN` from `JavadocTagConstants.INLINE_TAGS`.
- `JavadocBugsTest.testBug267833_3` fail on different JVM versions because now we don't report mismatch error anymore on older JLS. 
 
Will fix that.

iloveeclipse:
Notes / open points:

1) because of the JLS change, `JavadocTagConstants.JAVADOC_TAG_TYPE`
array doesn't work anymore as intended, because there are now tags that
can be tags that support *both* block or inline style.
2) `JavadocTagConstants.INLINE_TAGS` should actually contain `TAG_RETURN`
but that causes failures in `JavadocCompletionParserTest` because of
return tag included twice.
3) ECJ reports javadoc error if @return tag is included both inline and
as block - but javac reports a warning here.

jarthana:
> Notes / open points:
> 
>     1. because of the JLS change, `JavadocTagConstants.JAVADOC_TAG_TYPE`
>        array doesn't work anymore as intended, because there are now tags that
>        can be tags that support _both_ block or inline style.
> 
>     2. `JavadocTagConstants.INLINE_TAGS` should actually contain `TAG_RETURN`
>        but that causes failures in `JavadocCompletionParserTest` because of
>        return tag included twice.

I figured these would be a problem during my initial assessment. Let me see what we can do here.

jarthana:
> Notes / open points:
> 
>     1. because of the JLS change, `JavadocTagConstants.JAVADOC_TAG_TYPE`
>        array doesn't work anymore as intended, because there are now tags that
>        can be tags that support _both_ block or inline style.
> 
>     2. `JavadocTagConstants.INLINE_TAGS` should actually contain `TAG_RETURN`
>        but that causes failures in `JavadocCompletionParserTest` because of
>        return tag included twice.
> 
>     3. ECJ reports javadoc error if @return tag is included both inline and
>        as block - but javac reports a warning here.

I am working on the first two. But I believe the third is because it's a configurable problem and we can simply set it to what we want. Let me confirm and if still an issue, will address that as well.

jarthana:
> But I believe the third is because it's a configurable problem and we can simply set it to what we want. Let me confirm and if still an issue, will address that as well.

I can confirm that this can be configured to error/warning or info. We only have a top level flag for all Javadoc validation, but that's still good enough to address this. My latest patch addresses all remaining issues.

PR #984, opened 2023-04-18
Extends CompilationParticipant to allow post-processing of compiled classes. Fixes #983

### LINKED ISSUE #983: Compiled java classes cannot be modifed before they are written to disk
### Issue
I'm currently working on a project requiring to manipulate the byte code of compiled java classes. Currently this is achieved by hooking into 
```java
org.eclipse.jdt.core.compiler.CompilationParticipant.buildFinished(IJavaProject)
``` 
, reading the compiled classes from disk, applying the byte code modification and writing the resulting byte code back to disk.

This works, but introduces quite some extra runtime because of the additional read + write operations. This may be acceptable if just a few classes are to be modified, but for larger projects this quickly becomes a serious performance bottleneck, especially if the workspace is located on a network file system (NFS).

### Proposal

For the default java compiler there are compiler extensions available which allow to hook into the compile framework and manipulate the created java classes **before** they are written to disk, eliminating additional read+write operations. For the eclipse java compiler such possibilities are not available.

My proposal would be to introduce a new hook in
```java
org.eclipse.jdt.core.compiler.CompilationParticipant
```
which gives access to the compiled byte stream (read+write access) and some meta information (e.g. the BuildContext) right before it is written to disk, allowing to perform arbitrary byte code manipulations as required by the developer.

For example this new hook could look like this:
```java
public byte[] postProcess(BuildContext file, byte[] bytes);
```

This would enhance the possibilities to hook into the compile process and provide a general approach to modify the compiled byte code in performance-critical applications.

steffschl:
## What it does
This PR introduces a new hook in org.eclipse.jdt.core.compiler.CompilationParticipant to allow post-processing of compiled classes.

The hook provides read and write access to the compiled byte stream and some meta information (org.eclipse.jdt.core.compiler.BuildContext), right before it is written to disk, allowing to perform arbitrary byte code manipulations in a performant way as required by the developer.

## How to test
Provide implementation of org.eclipse.jdt.core.compiler.CompilationParticipant.postProcess(BuildContext, byte[]). Manipulate the class content which is passed in as parameter and return the modified one. The returned class content is written out to disk by the compiler framework.

## Corresponding issue
#983 

## Author checklist

- [ x] I have thoroughly tested my changes
- [ x] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [ x] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)


iloveeclipse:
@steffschl : 
# please squash all changes into one commit and force push your change.
# please add tests (again, in same commit)

steffschl:
@iloveeclipse 
1.) done. PR should be updatedwithin the next minutes.
2.) Changes in CompilationParticipant and AbstractImageBuilder are covered by ParticipantBuildTests. BuildContext only contains javadoc changes, in ClassFile only a new setter was added, I couldn't find tests covering this class.

steffschl:
> @steffschl : please set "user.email" property in the git repository settings for eclipse.jdt.core, so all your commits are automatically get proper author.

Thanks, done.

iloveeclipse (changes requested):

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ClassFile.java
```diff
@@ -5764,6 +5764,17 @@ public byte[] getBytes() {
 		}
 		return this.bytes;
 	}
+
+	/**
+	 * EXTERNAL API
```
cdietrich, line 5769: what is the semantiv of EXTERNAL_API

org.eclipse.jdt.core/model/org/eclipse/jdt/core/compiler/CompilationParticipant.java
```diff
@@ -143,6 +143,22 @@ public void processAnnotations(BuildContext[] files) {
...
+ * class.
+ *
+ * @param file BuildContext of the generated class
+ * @param bytes byte representation of the generated class
+ * @return byte array with length > 0 representing the modified class or byte[0]
+ *         if no modification was applied.
+ * @since 3.34
+ */
+public byte[] postProcess(BuildContext file, byte[] bytes) {
+	return new byte[0];
```
cdietrich, line 159: do we ususally have contstants for such magic objects?

iloveeclipse: One can create one in this class, something like `NO_CHANGE`.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ClassFile.java
iloveeclipse: I believe this is just a hint for the maintainer, see `getBytes()`.

iloveeclipse:
@jarthana : may I ask your opinion about this proposal?

org.eclipse.jdt.core/model/org/eclipse/jdt/core/compiler/CompilationParticipant.java
```diff
@@ -143,6 +143,22 @@ public void processAnnotations(BuildContext[] files) {
...
+ * This allows to perform arbitrary byte code manipulation on the generated
+ * class.
+ *
+ * @param file BuildContext of the generated class
+ * @param bytes byte representation of the generated class
+ * @return byte array with length > 0 representing the modified class or byte[0]
+ *         if no modification was applied.
+ * @since 3.34
+ */
+public byte[] postProcess(BuildContext file, byte[] bytes) {
```
szarnekow, line 158: We strictly need to specify what an implementor of a CompilationParticipant can expect here if another CompilationParticipant performs post processing, too. Is there an order to be imposed? Is it deliberately random? Do we throw an exception, if two compilation participants modify the bytes? 

Please also note in the JavaDoc, how / if an implementor can / must record additional dependencies that are imposed due to the post processing. E.g. if a post processor works on the bytes for `class A` and introduces a super class reference in the bytecode to `class B`, `class A` must certainly be recompiled, if `B` changes. In that case, `file.recordDependencies` must be called, I assume. Also document, what is to be expected from UI services. E.g. would the introduced super type `B` be visible in the class hierarchy of `A`?

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java
```diff
@@ -190,6 +190,15 @@ public void acceptResult(CompilationResult result) {
 				if (result.checkSecondaryTypes && !qualifiedTypeName.equals(compilationUnit.initialTypeName))
 					acceptSecondaryType(classFile);
 			}
+			for (int j = 0, l = this.javaBuilder.participants == null ? 0 : this.javaBuilder.participants.length; j < l; j++) {
+				BuildContext buildContext = new CompilationParticipantResult(compilationUnit,
+						this.compilationGroup == CompilationGroup.TEST);
+				byte[] postProcessingResult = this.javaBuilder.participants[j].postProcess(buildContext,
+						classFile.getBytes());
```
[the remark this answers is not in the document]
szarnekow: Since it is safe to assume that the bytes should not be modified directly, but a new byte array is supposed to be made, I'd say the ByteArrayInputStream is a nice idea.

org.eclipse.jdt.core/model/org/eclipse/jdt/core/compiler/CompilationParticipant.java
szarnekow: I'm thinking about `Optional<byte[]>` which would communicate the intend arguably better.

jarthana:
I haven't yet looked at this, but looks like this is related to #694 ?

iloveeclipse:
> I haven't yet looked at this, but looks like this is related to #694 ?

Not directly.

In  #694 participants are receiving in the `ClassContent` the bytes that compiler generates, but they aren't asked to change the `classFile` itself.

This is what requested here - "class file weaving on the fly" **before** writing bytes to the disk, to avoid "generate class bytes -> write bytes to disk -> read bytes from disk -> modify class by participants > write class again" sequence.

The proposal in #694 would allow to skip `read bytes from disk` step and change 

`generate class bytes -> write bytes to disk -> read bytes from disk -> modify class by participants > write class again`

to 

`generate class bytes -> write bytes to disk -> read bytes from ClassContent -> modify class by participants > write class again`

**Here** the request is to change the order and reduce it to something like:

`generate class bytes -> read bytes from ClassFile -> modify ClassFile in memory by participants -> write bytes to disk`

jarthana:
> > I haven't yet looked at this, but looks like this is related to #694 ?
> 
> Not directly.

Thanks Andrey, for neatly summing it up. So, is it fair to say that this one makes (or can make) #694 redundant?

In other words, can we achieve what #694 intends to do with this PR itself? I see that the other PR is little more conservative by letting only annotation processors access the class content. That difference apart, may be we can merge these two issues together?

I looked at the patch and found it straight forward. But as you and others expressed, the real question is about the design and setting the right expectation from the new API for the clients. May be about time we start releasing our own preview features ;)

steffschl:
I've just pushed the latest state containing your review feedback.

iloveeclipse:
> In other words, can we achieve what #694 intends to do with this PR itself? 

We could, but it is not needed for this feature and can be implemented "on top" in (updated version of) #694.

> I see that the other PR is little more conservative by letting only annotation processors access the class content. That difference apart, may be we can merge these two issues together?

As said, the one is for manipulating the files, the other is for inspecting / using what was finally generated.
They are related but can be done iteratively one after each other.

> But as you and others expressed, the real question is about the design and setting the right expectation from the new API for the clients.

Yes. Here Sebastian raised good question, so I will try to answer what I could made of it.

> We strictly need to specify what an implementor of a CompilationParticipant can expect here if another CompilationParticipant performs post processing, too. Is there an order to be imposed? Is it deliberately random? 

Let's look at [`JavaModelManager.CompilationParticipants.getRegisteredParticipants()`](https://github.com/eclipse-jdt/eclipse.jdt.core/blob/fdacba6b017279ec9c9bb00f7fa60207847b8ec7/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java#L480)

There is already today a possibility to impose dependency order [via `requires` attribute](https://github.com/eclipse-jdt/eclipse.jdt.core/blob/fdacba6b017279ec9c9bb00f7fa60207847b8ec7/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java#L576).

So the order problem can be **partly** solved by already existing feature - because the participants need to know each other  to define dependency relationship.

However there can be other requirements like "I want be the first one after ecj" or "I want be last one before writing" or "I want be only one who modifies".

So one could add `postProcess` attribute with `first`, `last` and `exclusive`  and sort additionally on that. If there will be conflicting participants, the first one will win and all other will be not considered (sure with error reported).

But given the fact there are no "post processing" participants **yet**, and not many would be anyway, we can first provide a "simple" API (that assumes all participants are well behaving, not exclusive and can live with "requires" order) and refine it afterwards by adding extra extension point attribute. That would be not breaking I assume, if the first API version would not guarantee any order except one defined by "requires" and second one would allow extra options.

@szarnekow : If you however think a new attribute would be required, please raise your hand! I definitely have no insights into Xtext special use cases you might have in mind.

> Do we throw an exception, if two compilation participants modify the bytes?

I would say no. Given the fact that participants **would be able** specify order they want, they should be able to deal with modified class files.

> Please also note in the JavaDoc, how / if an implementor can / must record additional dependencies that are imposed due to the post processing. E.g. if a post processor works on the bytes for class A and introduces a super class reference in the bytecode to class B, class A must certainly be recompiled, if B changes. In that case, file.recordDependencies must be called, I assume. 

Good point, yes, this can be done via `recordParticipantResult(buildContext);`. So participants also could even report problems there happened during post processing.

> Also document, what is to be expected from UI services. E.g. would the introduced super type B be visible in the class hierarchy of A?

I would say that's beyond the scope of the API here. Compiler also doesn't explain where and how synthetic or bridge methods or inner classes or all the extra generated structures should be shown or not.

szarnekow:
> > We strictly need to specify what an implementor of a CompilationParticipant can expect here if another CompilationParticipant performs post processing, too. Is there an order to be imposed? Is it deliberately random?
> 
> There is already today a possibility to impose dependency order [via `requires` attribute](https://github.com/eclipse-jdt/eclipse.jdt.core/blob/fdacba6b017279ec9c9bb00f7fa60207847b8ec7/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java#L576).
> 
> [..]
> 
> @szarnekow : If you however think a new attribute would be required, please raise your hand! I definitely have no insights into Xtext special use cases you might have in mind.

I think `requires` is exactly what we need as a first step. I'm surprised that it's not mentioned on the CompilationParticipant itself in the JavaDoc. Sorry that I missed it in the extension point decl. Maybe add at least a hint to the class documentation.
Btw: Can a CompilationParticipant be registered for a compilation via the ecj command line compiler?

> I would say no. Given the fact that participants **would be able** specify order they want, they should be able to deal with modified class files.

Agreed.

> 
> > Please also note in the JavaDoc, how / if an implementor can / must record additional dependencies that are imposed due to the post processing. E.g. if a post processor works on the bytes for class A and introduces a super class reference in the bytecode to class B, class A must certainly be recompiled, if B changes. In that case, file.recordDependencies must be called, I assume.
> 
> Good point, yes, this can be done via `recordParticipantResult(buildContext);`. So participants also could even report problems there happened during post processing.


Maybe mention this in the Javadoc, too. Currently it states that the postProcess happens after finished class compilation. It's not obvious that problems might still be reported at that stage. 

> 
> > Also document, what is to be expected from UI services. E.g. would the introduced super type B be visible in the class hierarchy of A?
> 
> I would say that's beyond the scope of the API here. Compiler also doesn't explain where and how synthetic or bridge methods or inner classes or all the extra generated structures should be shown or not.

Agreed. It would be enough to mention that structural changes to the generated class are not reflected by any means in UI services. This likely implies that a newly introduced method is only available in byte code and cannot be used from another class without seeing error markers in the editor, true?

iloveeclipse:
> is it possible this changes the eventing
> and thus class file changes that were produced before are now no longer produced?

@cdietrich : could you please elaborate what "this" is, which events do you mean and what do you mean by  "class file changes that were produced before are now no longer produced"? It feel like you have some concrete bug in mind but I miss the context.

cdietrich:
if the byte code is changed with file.setcontents there might be different eventing in eclipse resouces
i am strugglong to adopt this change in xtext codebase
but this might be problem on our side

iloveeclipse:
> if the byte code is changed with file.setcontents there might be different eventing in eclipse resouces

Here is nothing changed in the way **how** files are written, only **what** is written. So there should be no different events (except generated bytecode causes later changes because it is used somewhere and the changes in bytecode change the logic later on).

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ClassFile.java
```diff
@@ -5764,6 +5764,16 @@ public byte[] getBytes() {
 		}
 		return this.bytes;
 	}
+
+	/**
+	 * Sets the actual bytes of the class file.
+	 *
+	 * This method can be used e.g. to store the modified byte representation of the class file after post processing.
+	 */
```
iloveeclipse, line 5772: It is not API, but nevertheless one should add a note that "this method is made public only to be accessible from AbstractImageBuilder during compilation post processing. Using this method for any other purpose is discouraged and may lead to unpredictable results".

org.eclipse.jdt.core.tests.builder/src/org/eclipse/jdt/core/tests/builder/ParticipantBuildTests.java
```diff
@@ -405,6 +405,110 @@ public void processAnnotations(BuildContext[] files) {
 		expectingNoProblems();
 	}
 
+	public void testPostProcessing() throws JavaModelException {
```
iloveeclipse, line 408: We need **another** test that tests that three participants configured with "requires" are:
1) Executed in expected order
2) See the output of each other
3) the final class bytes contain results of all three participants
4) Participants can add one more class to be recompiled via "dependencies" set on result
5) Participants can create extra compilation problems.

org.eclipse.jdt.core/model/org/eclipse/jdt/core/compiler/CompilationParticipant.java
```diff
@@ -143,6 +146,28 @@ public void processAnnotations(BuildContext[] files) {
 	// do nothing by default
 }
 
+/**
+ * Notifies this participant about a finished class compilation. This
+ * notification occurs right before the compiled class is written out to disk.
+ * This allows to perform arbitrary byte code manipulation on the generated
+ * class.
+ *
+ * There is no specific order in which potential multiple compilation participants are called. The received class data
```
iloveeclipse: Please start with "If the "requires" attribute is not used for participant definition, there will be no specific order ..."

steffschl:
Updated the PR with latest comments. I also introduces a new class comment describing the "requires" attribute.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java
```diff
@@ -190,6 +190,15 @@ public void acceptResult(CompilationResult result) {
 				if (result.checkSecondaryTypes && !qualifiedTypeName.equals(compilationUnit.initialTypeName))
 					acceptSecondaryType(classFile);
 			}
+			for (int j = 0, l = this.javaBuilder.participants == null ? 0 : this.javaBuilder.participants.length; j < l; j++) {
+				BuildContext buildContext = new CompilationParticipantResult(compilationUnit,
+						this.compilationGroup == CompilationGroup.TEST);
+				Optional<byte[]> postProcessingResult = this.javaBuilder.participants[j].postProcess(buildContext,
+						new ByteArrayInputStream(classFile.getBytes()));
+				if (postProcessingResult.isPresent()) {
```
iloveeclipse, line 198: Please add `recordParticipantResult(buildContext);` call, to record dependencies / problems (also if the class file was NOT modified.

org.eclipse.jdt.core/model/org/eclipse/jdt/core/compiler/CompilationParticipant.java
```diff
@@ -143,6 +151,32 @@ public void processAnnotations(BuildContext[] files) {
...
+ * class is written out to disk. This allows to perform arbitrary byte code manipulation on the generated class.
+ *
+ * If additional dependencies to other classes are introduced by the modified class, these dependencies have to be
+ * recorded by calling org.eclipse.jdt.core.compiler.BuildContext.recordDependencies(String[]).
+ *
+ * In case problems occur, they can be reported using
+ * org.eclipse.jdt.core.compiler.BuildContext.recordNewProblems(CategorizedProblem[]).
+ *
+ * The received class data could have already been transformed by another compilation participant. If the compilation
+ * participants have to be called in a specific order, the attribute "requires" can be set on the extension point
```
iloveeclipse, line 165: "can be set on the extension point entries" -> "can be used in the participant contribution entry"

iloveeclipse (changes requested):

org.eclipse.jdt.core.tests.builder/src/org/eclipse/jdt/core/tests/builder/TestBuilderParticipant.java
```diff
@@ -47,4 +50,8 @@ public boolean isAnnotationProcessor() {
 public void processAnnotations(BuildContext[] files) {
 	PARTICIPANT.processAnnotations(files);
 }
+
+public Optional<byte[]> postProcess(BuildContext file, ByteArrayInputStream bytes) {
```
cdietrich, line 54: is there a way to access the name of the to be generated class files
as org.eclipse.jdt.core.JavaCore.getGeneratedResources(IRegion, boolean)
would produce it?
in the build context i dont see it.
the caller has access to the classfile
but this one has it neither
org.eclipse.jdt.internal.core.builder.AbstractImageBuilder.writeClassFile(ClassFile, SourceFile, boolean)
will have it

@szarnekow our test is green anyway.
thus i wonder why we need to touch the class file

cdietrich: touch is needed in older versions like 2202203. so i assume the needrebuild fix also fixed that

steffschl: > is there a way to access the name of the to be generated class files as org.eclipse.jdt.core.JavaCore.getGeneratedResources(IRegion, boolean) would produce it? in the build context i dont see it. the caller has access to the classfile but this one has it neither org.eclipse.jdt.internal.core.builder.AbstractImageBuilder.writeClassFile(ClassFile, SourceFile, boolean) will have it

With the current implementation this is not possible (as far as I know) and also not required for this issue, but #694 intends to introduce this possibility.

cdietrich:
ok, looks like i have a successful adaptation of the new extension point for DebugSourceInstallingCompilationParticipant
https://github.com/eclipse/xtext/compare/cd_testCompilationParticipant2

org.eclipse.jdt.core.tests.builder/src/org/eclipse/jdt/core/tests/builder/ParticipantBuildTests.java
steffschl: > We need **another** test that tests that three participants configured with "requires" are:
> 
>     1. Executed in expected order
> 
>     2. See the output of each other
> 
>     3. the final class bytes contain results of all three participants
> 
>     4. Participants can add one more class to be recompiled via "dependencies" set on result
> 
>     5. Participants can create extra compilation problems.

@iloveeclipse 
Is there some way to register compilation participants on the fly? I would like to prevent to register them via plugin.xml since they would be visible for all tests if we need it just for one, but not sure if this is possible.

iloveeclipse: > Is there some way to register compilation participants on the fly

No, I don't see a way right now.
See 
- final org.eclipse.jdt.internal.core.JavaModelManager.compilationParticipants
- org.eclipse.jdt.internal.core.JavaModelManager.CompilationParticipants.getRegisteredParticipants()

But in theory the "dummy" test participants shouldn't harm if they are added in a way where they only work during test.

iloveeclipse: `CompilationParticipant.isActive(IJavaProject)` should just return false for any project except the one used in the test.

steffschl:
2 questions:

1. Should we mention in the class description that the "requires" attribute is non-transitive?
2. Regarding the following XML snippet:

```
<compilationParticipant
   class="org.eclipse.jdt.core.tests.builder.participants.TestCompilationParticipant1"
   id="org.eclipse.jdt.core.tests.builder.participants.testCompilationParticipant1">
</compilationParticipant>
<compilationParticipant
   class="org.eclipse.jdt.core.tests.builder.participants.TestCompilationParticipant2"
   id="org.eclipse.jdt.core.tests.builder.participants.testCompilationParticipant2">
   <requires id="org.eclipse.jdt.core.tests.builder.participants.testCompilationParticipant1"/>
</compilationParticipant>
```

I would expect testCompilationParticipant1 to be run before testCompilationParticipant2, but it is the other way. I also tested this behavior with more participants. Is this intended?

iloveeclipse:
> I would expect testCompilationParticipant1 to be run before testCompilationParticipant2, but it is the other way. I also tested this behavior with more participants. Is this intended?

I believe this is "expected" reverse order of contributions in xml. Don't ask me why, it is like this since ever.

iloveeclipse:
> 1. Should we mention in the class description that the "requires" attribute is non-transitive?

What do you mean by that? 
Participant A requires B, B requires C doesn't mean CBA execution oder? Would be a bug.

steffschl:
> > 1. Should we mention in the class description that the "requires" attribute is non-transitive?
> 
> What do you mean by that? Participant A requires B, B requires C doesn't mean CBA execution oder? Would be a bug.

Sorry, was a mistake in my setup. It's working as intended now.

org.eclipse.jdt.core.tests.builder/src/org/eclipse/jdt/core/tests/builder/participants/ParticipantBase.java
```diff
@@ -0,0 +1,46 @@
+package org.eclipse.jdt.core.tests.builder.participants;
```
iloveeclipse, line 1: File header is missimg

iloveeclipse (commented):
Have briefly reviewed in the browser, looks good.
Will check next week in IDE, especially tests.

org.eclipse.jdt.core.tests.builder/src/org/eclipse/jdt/core/tests/builder/participants/ParticipantBase.java
steffschl: added

org.eclipse.jdt.core.tests.builder/src/org/eclipse/jdt/core/tests/builder/participants/ParticipantBase.java
```diff
@@ -0,0 +1,59 @@
...
+
+import java.io.ByteArrayInputStream;
+import java.util.Optional;
+
+import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.jdt.core.compiler.BuildContext;
+import org.eclipse.jdt.core.compiler.CompilationParticipant;
+
+public class ParticipantBase extends CompilationParticipant {
+	protected CompilationParticipant getParticipant() {
```
iloveeclipse, line 24: Should be abstract and documented

iloveeclipse (changes requested):
The biggest problem I see now is **wrong** sort order of participants we have in `JavaModelManager.CompilationParticipants.sortParticipants()`.

I think that has to be changed, I'm still thinking how to deal with "sortParticipants" which was obviously always wrong.

I've created https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1000 and pushed https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1001 to see what would break.

For all the comments above I have patch prepared except for the "sortParticipants" issue, I will push an update in a moment.

iloveeclipse:
Note, I've also updated commit message.

org.eclipse.jdt.core.tests.builder/src/org/eclipse/jdt/core/tests/builder/ParticipantBuildTests.java
```diff
@@ -405,6 +405,253 @@ public void processAnnotations(BuildContext[] files) {
...
+		env.removePackageFragmentRoot(projectPath, "");
+		IPath root = env.addPackageFragmentRoot(projectPath, "src");
+		env.setOutputFolder(projectPath, "bin");
+
+		final String TEST_CLASS_NAME = "TestClass";
+		env.addClass(root, "", TEST_CLASS_NAME, "public class " + TEST_CLASS_NAME + " {}\n");
+
+		// The set "requires" attribute enforces calling order TestCompilationParticipant1 -->
+		// TestCompilationParticipant3 --> TestCompilationParticipant2
+		//
```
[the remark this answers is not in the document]
iloveeclipse: OK, I've fixed master with #1001 so the code/tests now do what one expects it to do.

iloveeclipse (approved):
@cdietrich, @szarnekow, @steffschl : please check the last commit.

From my POV that is ready to merge.

iloveeclipse:
> > I would expect testCompilationParticipant1 to be run before testCompilationParticipant2, but it is the other way. I also tested this behavior with more participants. Is this intended?
> 
> I believe this is "expected" reverse order of contributions in xml. Don't ask me why, it is like this since ever.

My apologies, I haven't noticed there was "requires" line in the snippet and assumed you were talking about default contribution reading order. In fact, this is #1000 and I've fixed that in master via #1001. I've realized that while reviewing your new test (original version) that proved nonsensical execution order.

cdietrich:
after adapting the Xtext patch use override isPostProcessor method tests are still green

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ClassFile.java
```diff
@@ -5764,6 +5764,18 @@ public byte[] getBytes() {
 		}
 		return this.bytes;
 	}
+
+	/**
+	 * Sets the actual bytes of the class file.
```
szarnekow, line 5769: Should this be called `internalSetBytes(..)`?

Also, mention that the byte array is stored as-is, and subsequent modification will be reflected and vice versa.

org.eclipse.jdt.core/model/org/eclipse/jdt/core/compiler/CompilationParticipant.java
```diff
@@ -143,6 +152,48 @@ public void processAnnotations(BuildContext[] files) {
...
+ * @return whether this participant is interested in post processing
+ * @since 3.34
+ */
+public boolean isPostProcessor() {
+	return false;
+}
+
+/**
+ * Notifies this participant about a finished class compilation. This notification occurs right before the compiled
+ * class is written out to disk. This allows to perform arbitrary byte code manipulation on the generated class.
```
szarnekow, line 174: Please add `Only sent to participants that answer true to {@link #isPostProcessor()}.`

szarnekow (approved):
This is a nice addition to the CompilationParticipant extension point.

steffschl:
Changes look good to me, thanks for your code contributions @iloveeclipse

PR #995, opened 2023-04-21
Expect type for first arg on constructor. Fix #979

### LINKED ISSUE #979: Expected type of the first parameter is not populated in context for constructors
Expected type of the first parameter is not populated in context for constructors.

```
public class App {
      public App(List<String> names) {}

     public static void foo() {
     App a = new App(|);  
     }

}
```

completions on `App(|` must populate `List<String>` as a expected type like it does for method invocations.

gayanper:
## What it does
The fix use the same strategy used for finding first argument type on CompletionOnMessageSend for CompletionOnQualifiedAllocationExpression.

## How to test
Unit Test and example provided in issue #979 

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests.java
@@ -5664,13 +5664,15 @@ public void testCompletionFindConstructor() throws JavaModelException {
     this.wc.codeComplete(cursorLocation, requestor, this.wcOwner);
 
     assertResults(
-            "expectedTypesSignatures=null\n"+
-            "expectedTypesKeys=null",
+            "expectedTypesSignatures={I}\n"+
+            "expectedTypesKeys={I}",
             requestor.getContext());
 
    assertResults(
 			"CompletionFindConstructor[ANONYMOUS_CLASS_DECLARATION]{, LCompletionFindConstructor;, (I)V, null, (i), "+(R_DEFAULT + R_RESOLVED + R_INTERESTING + R_NON_RESTRICTED)+"}\n" +
-			"CompletionFindConstructor[METHOD_REF<CONSTRUCTOR>]{, LCompletionFindConstructor;, (I)V, CompletionFindConstructor, (i), "+(R_DEFAULT + R_RESOLVED + R_INTERESTING + R_NON_RESTRICTED)+"}",
+			"CompletionFindConstructor[METHOD_REF<CONSTRUCTOR>]{, LCompletionFindConstructor;, (I)V, CompletionFindConstructor, (i), " + (R_DEFAULT + R_RESOLVED + R_INTERESTING + R_NON_RESTRICTED) + "}\n" +
+			"hashCode[METHOD_REF]{hashCode(), Ljava.lang.Object;, ()I, hashCode, null, " + (R_DEFAULT + R_RESOLVED + R_INTERESTING + R_CASE + R_UNQUALIFIED + R_NON_RESTRICTED) + "}\n" +
+			"x[LOCAL_VARIABLE_REF]{x, null, I, x, null, " + (R_DEFAULT + R_RESOLVED + R_INTERESTING + R_CASE + R_UNQUALIFIED + R_NON_RESTRICTED) + "}",
 			requestor.getResults());
 }
 
@@ -5755,13 +5757,14 @@ public void testCompletionFindConstructor4() throws JavaModelException {
     this.wc.codeComplete(cursorLocation, requestor, this.wcOwner);
 
     assertResults(
-            "expectedTypesSignatures=null\n"+
-            "expectedTypesKeys=null",
+            "expectedTypesSignatures={I}\n"+
+            "expectedTypesKeys={I}",
             requestor.getContext());
 
 	assertEquals(
 			"Constructor4[ANONYMOUS_CLASS_DECLARATION]{, Lzconstructors.Constructor4;, (I)V, null, (i), " + (R_DEFAULT + R_RESOLVED + R_INTERESTING + R_NON_RESTRICTED) + "}\n" +
-			"Constructor4[METHOD_REF<CONSTRUCTOR>]{, Lzconstructors.Constructor4;, (I)V, Constructor4, (i), " + (R_DEFAULT + R_RESOLVED + R_INTERESTING + R_NON_RESTRICTED) + "}",
+			"Constructor4[METHOD_REF<CONSTRUCTOR>]{, Lzconstructors.Constructor4;, (I)V, Constructor4, (i), " + (R_DEFAULT + R_RESOLVED + R_INTERESTING + R_NON_RESTRICTED) + "}\n" +
+			"hashCode[METHOD_REF]{hashCode(), Ljava.lang.Object;, ()I, hashCode, null, " + (R_DEFAULT + R_RESOLVED + R_INTERESTING + R_CASE + R_UNQUALIFIED + R_NON_RESTRICTED) + "}",
 			requestor.getResults());
 }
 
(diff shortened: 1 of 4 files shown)


org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
```diff
@@ -6091,6 +6117,7 @@ private void findConstructors(
 								continue next;
 						}
 
+					constructorsFound.add(new Object[]{constructor, currentType});
```
trancexpress, line 6120: Passing this looks weird, but considering the is 400+ lines long, I'm not sure the code gets much worse.

Though is it really necessary to gather the list and then add suggestions for it? You cannot add suggestions right here?

trancexpress:
Changing your example slightly:

```
public class App {
	public App(List<String> names) {
	}
	public App(Map<String, Integer> m, int i) {
	}

	public void bar(List<String> names) {
	}

	public void bar(Map<String, Integer> m, int i) {
	}

	public static void foo() {
		App a = new App(, 1);
		a.bar(, 1);
	}

	private static List<String> emptyList() {
		return null;
	}

}
```

When I invoke content assist at the cursor, I get a suggestion for `emptyList()`. This isn't expected? I see it also for the method call though, maybe this is just a bug that has existed so far and is not related to your changes (other than, the bug now manifests also for constructors due to having completion suggestions)?

gayanper:
> Changing your example slightly:
> 
> ```
> public class App {
> 	public App(List<String> names) {
> 	}
> 	public App(Map<String, Integer> m, int i) {
> 	}
> 
> 	public void bar(List<String> names) {
> 	}
> 
> 	public void bar(Map<String, Integer> m, int i) {
> 	}
> 
> 	public static void foo() {
> 		App a = new App(|, 1);
> 		a.bar(, 1);
> 	}
> 
> 	private static List<String> emptyList() {
> 		return null;
> 	}
> 
> }
> ```
> 
> When I invoke content assist at the cursor, I get a suggestion for `emptyList()`. This isn't expected? I see it also for the method call though, maybe this is just a bug that has existed so far and is not related to your changes (other than, the bug now manifests also for constructors due to having completion suggestions)?

I think it is as expected since the constructor is overloaded. At completion time we don't use the exact resolved binding in the AST tree as the ASTNode, so here we don't yet know which constructor is to be used precisely. So instead we use both constructors given that they have 1 or more parameters and use their first parameter types as expected types.

gayanper:
@trancexpress based on my above answer do you think we can move forward with this patch ?

trancexpress:
> @trancexpress based on my above answer do you think we can move forward with this patch ?

Overall looks fine, but can you check my comment above? https://github.com/eclipse-jdt/eclipse.jdt.core/pull/995#discussion_r1181807918

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
gayanper: This collected constructors are not used for creating completions. They are used for other purposes such as eliminating duplicate constructors and also trying to figure our parameter types. It is same as methodsFound and typesFound collections in other methods.

gayanper:
@trancexpress I replied your comment, sorry i missed it. If things looks good now please provide your review approval so that I can merge this.

trancexpress (approved):

iloveeclipse:
@gayanper : could you please check if this PR causes test fails in org.eclipse.jdt.text.tests.contentassist.CodeCompletionTest	testAnonymousTypeCompletion6? 
See
https://download.eclipse.org/eclipse/downloads/drops4/I20230505-1800/testresults/html/org.eclipse.jdt.text.tests_ep428I-unit-cen64-gtk3-java17_linux.gtk.x86_64_17.html

gayanper:
I will fix the tests under the same issue, seems like we need to update the expected proposals since now we get more proposals matching the first argument type. Probably I can push changes by end of the weekend.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests.java
```diff
@@ -23395,7 +23402,11 @@ public void testBug346454c_2() throws JavaModelException {
...
 			"Test[METHOD_REF<CONSTRUCTOR>]{, Ltest.Test<>;, (TT;)V, Test, (t), " + (R_DEFAULT + R_RESOLVED + R_INTERESTING + R_NON_RESTRICTED) + "}\n" +
 			"Test<>[ANONYMOUS_CLASS_DECLARATION]{, Ltest.Test<>;, (I)V, null, (i), " + (R_DEFAULT + R_RESOLVED + R_INTERESTING + R_NON_RESTRICTED) + "}\n" +
-			"Test<>[ANONYMOUS_CLASS_DECLARATION]{, Ltest.Test<>;, (TT;)V, null, (t), " + (R_DEFAULT + R_RESOLVED + R_INTERESTING + R_NON_RESTRICTED) + "}",
+			"Test<>[ANONYMOUS_CLASS_DECLARATION]{, Ltest.Test<>;, (TT;)V, null, (t), " + (R_DEFAULT + R_RESOLVED + R_INTERESTING + R_NON_RESTRICTED) + "}\n" +
+			"clone[METHOD_REF]{clone(), Ljava.lang.Object;, ()Ljava.lang.Object;, clone, null, " + (R_DEFAULT + R_RESOLVED + R_INTERESTING + R_CASE + R_UNQUALIFIED + R_NON_RESTRICTED) + "}\n" +
+			"getClass[METHOD_REF]{getClass(), Ljava.lang.Object;, ()Ljava.lang.Class;, getClass, null, " + (R_DEFAULT + R_RESOLVED + R_INTERESTING + R_CASE + R_UNQUALIFIED + R_NON_RESTRICTED) + "}\n" +
+			"hashCode[METHOD_REF]{hashCode(), Ljava.lang.Object;, ()I, hashCode, null, " + (R_DEFAULT + R_RESOLVED + R_INTERESTING + R_CASE + R_UNQUALIFIED + R_NON_RESTRICTED) + "}\n" +
```
stephan-herrmann, line 23408: @gayanper please double check your Eclipse settings: at this line the change introduced a trailing whitespace, which should never happen because the project has a save action configured which will remove all trailing white space.

I stumbled upon this when I edited the file and even after undo it contained a change wrt git index: the trailing white space was removed. I saw the same effect in another file (don't remember which and where), where it created several changes in different places, which makes creating a minimal patch for such file quite tedious. 

Thanks.

stephan-herrmann: > I saw the same effect in another file (don't remember which and where)

tripped over it again: `CompletionEngine` from the same commit.

gayanper: Sure i will check and fix the settings @stephan-herrmann in the workspace.

gayanper: Since this PR is merged, how do we handle it ?

stephan-herrmann: > Since this PR is merged, how do we handle it ?

I guess a dedicated commit only removing the trailing whitespace soon after opening for 4.29 would be best.

Did you find out how this could happen in the first place? The setting is actually persisted in git, so if `.settings/` is clean wrt git all should be good.

Invoking undo directly after saving will revert only the effect of save actions. Maybe that's what you did?

gayanper: I think it might be due to I import the projects as maven modules, But i will make sure the settings stays clean always to make sure i don't miss this.

PR #604, opened 2022-12-14
JavaModelManager: Use custom ForkJoinWorkerThreadFactory

sratz:
## What it does
In case a SecurityManager is installed, using the defaultForkJoinWorkerThreadFactory would result in worker threads with no permissions, which could prevent the model from being saved.

Use a custom ForkJoinWorkerThreadFactory implementation to work around this.

The same fix was also applied to org.eclipse.core.resources SaveManager.

See also:
https://github.com/eclipse-platform/eclipse.platform/issues/294
https://github.com/eclipse-platform/eclipse.platform/pull/295

## How to test
1. Have any Eclipse-plugin that installs a `SecurityManager`. Use attached demo plugin [foo.bar.securitymanager.zip](https://github.com/eclipse-platform/eclipse.platform/files/10208024/foo.bar.securitymanager.zip) (Main Menu `SecurityManager -> Install SecurityManager`) or the following snippet:
```java
Policy.getPolicy(); // init
Policy.setPolicy(new Policy() {
    @Override
    public boolean implies(ProtectionDomain domain, Permission permission) {
        return true;
    }
});
System.setSecurityManager(new SecurityManager());
```
2. Open ja Java project, edit some file, do not save
3. Close Eclipse -> Answer 'yes' when asked to save files

-> 
image
```
Problems occurred while trying to save the state of the workbench.
  Problems occurred during save.
    Error saving build states
    java.security.AccessControlException: access denied ("java.io.FilePermission" "C:\eclipse_workspaces\runtime-EclipseContribute\.metadata\.plugins\org.eclipse.core.resources\.projects\A\org.eclipse.jdt.core" "read")
```

## Author checklist
- [X] I have thoroughly tested my changes
- [X]  The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [X] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

PR #1017, opened 2023-04-27
PatterSearchJob: Use custom ForkJoinWorkerThreadFactory

Christopher-Hermann:
### What it does
In case a SecurityManager is installed, using the defaultForkJoinWorkerThreadFactory would result in worker threads with no permissions, which leads to not working java type search.

Use a custom ForkJoinWorkerThreadFactory implementation to work around this.

The same fix was also applied to org.eclipse.jdt.internal.core.JavaModelManager.java

See also:
https://github.com/eclipse-jdt/eclipse.jdt.core/pull/604
https://github.com/eclipse-platform/eclipse.platform/issues/294
https://github.com/eclipse-platform/eclipse.platform/pull/295

## How to test

1. Have any Eclipse-plugin that installs a SecurityManager. Use attached demo plugin [foo.bar.securitymanager.zip](https://github.com/eclipse-platform/eclipse.platform/files/10208024/foo.bar.securitymanager.zip) (Main Menu SecurityManager -> Install SecurityManager) or the following snippet:
```
Policy.getPolicy(); // init
Policy.setPolicy(new Policy() {
    @Override
    public boolean implies(ProtectionDomain domain, Permission permission) {
        return true;
    }
});
System.setSecurityManager(new SecurityManager());
```
3. Open a Java project and open the type search (CTRL+SHIFT+T)
4. As soon as you start typing a error is shown

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/PatternSearchJob.java
+++ b/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/PatternSearchJob.java
@@ -19,6 +19,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.ForkJoinWorkerThread;
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicLong;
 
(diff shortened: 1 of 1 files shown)


Christopher-Hermann:
Discussion in https://github.com/eclipse-platform/eclipse.platform.text/pull/185 will be relevant for this PR as well.

gayanper (approved):

gayanper:
Thanks @Christopher-Hermann, sorry for taking so long to approve. I was not well last week, slowly starting back.

gayanper:
@Christopher-Hermann seems like the build hangs. Could you investigate it ?

iloveeclipse (changes requested):
Please undo the merge commit and rebase on master head instead

Christopher-Hermann:
> `commonPool` is never shutdown now & created on each job execution.

Done, thanks for the review.

szarnekow:
This is still creating a new pool for each index search. Putting aside that all stream().parallel() pipelines that are not initiated from within an existing parallel task or forkjoin task do still run on the common pool, I fail to see a systematic approach. It might make more sense to offer an equivalent to the common pool, e.g in one of the platform bundles as a primed pool that will be shutdown when the workspace / RCP is closed.

PR #1022, opened 2023-05-01
Improve array completions Fixes #1021

### LINKED ISSUE #1021: Improve array completions for current type
```
			public class ArrayTest {
			  public void test() {
			 		ArrayTest[] arr = new\s
			  }
			}
```

Should provide  `ArrayTest` as an array completion with array dimension > 1 

Improvement on https://github.com/eclipse-jdt/eclipse.jdt.core/issues/504

gayanper:
## What it does
The current fix avoid skipping current type from completions processing if the original expected type is an Array. And for primitives types remove unwanted conditions `!this.assistNodeIsConstructor || !allowingLongComputationProposals` which proves it is not needed for correct completions for primitive arrays.

For enclosing type array expected types, already found enclosing type proposal is carried forward into expected type evaluation and if such proposal is found, that proposal is updated with array dimension and relevance instead of creating similar TYPE-REF completion.

## How to test
Unit Tests

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
+++ b/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
@@ -5562,6 +5562,21 @@ public final class CompletionEngine
 			int[] missingElementsStarts,
 			int[] missingElementsEnds,
 			boolean missingElementsHaveProblems) {
+		createTypeProposal(refBinding, typeName, accessibility, completionName, relevance, missingElements,
+				missingElementsStarts, missingElementsEnds, missingElementsHaveProblems, new ObjectVector());
+	}
+
+	private void createTypeProposal(
+			ReferenceBinding refBinding,
+			char[] typeName,
+			int accessibility,
+			char[] completionName,
+			int relevance,
+			Binding[] missingElements,
+			int[] missingElementsStarts,
+			int[] missingElementsEnds,
+			boolean missingElementsHaveProblems,
+			ObjectVector proposalsCreated) {
 
 		// Create standard type proposal
 		if(!this.isIgnored(CompletionProposal.TYPE_REF, missingElements != null) && (this.assistNodeInJavadoc & CompletionOnJavadoc.ONLY_INLINE_TAG) == 0) {
@@ -5593,6 +5608,7 @@ public final class CompletionEngine
 			if(DEBUG) {
 				this.printDebug(proposal);
 			}
+			proposalsCreated.add(proposal);
 		}
 
 		// Create javadoc text proposal if necessary
@@ -5615,6 +5631,7 @@ public final class CompletionEngine
 			if(DEBUG) {
 				this.printDebug(proposal);
 			}
+			proposalsCreated.add(proposal);
 		}
 	}
 	private void createTypeVariable(TypeVariableBinding typeVariable, Scope scope, StringBuffer completion) {
@@ -11683,6 +11700,7 @@ public final class CompletionEngine
 
 		boolean isEmptyPrefix = token.length == 0;
 
+		ObjectVector proposalsFound = new ObjectVector();
 		if ((proposeType || proposeConstructor) && this.unitScope != null) {
 
 			ReferenceBinding outerInvocationType = scope.enclosingSourceType();
@@ -11791,7 +11809,7 @@ public final class CompletionEngine
 								null,
 								null,
 								null,
-								false);
+								false, proposalsFound);
 				}
 
 				if (proposeConstructor) {
@@ -11821,12 +11839,12 @@ public final class CompletionEngine
 
 			checkCancel();
 
-			findTypesFromExpectedTypes(token, scope, typesFound, proposeType, proposeConstructor);
+			findTypesFromExpectedTypes(token, scope, typesFound, proposeType, proposeConstructor, proposalsFound);
 		}
 
 		if (isEmptyPrefix && !this.assistNodeIsAnnotation) {
 			if (!proposeConstructor) {
-				findTypesFromExpectedTypes(token, scope, typesFound, proposeType, proposeConstructor);
+				findTypesFromExpectedTypes(token, scope, typesFound, proposeType, proposeConstructor, proposalsFound);
 			} else {
 				findConstructorsFromSubTypes(scope, typesFound);
 			}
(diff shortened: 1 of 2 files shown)


trancexpress:
> And for primitives remove
> unwanted conditions which checks for long computing and assist node.

Where is this part of the commit message in the change? Or am I misunderstanding the sentence?

Aide from that, the change in the code looks fine.

But I don't understand what the change does when using Eclipse.

Without the change, using the snippet from the new test, I see this:

https://user-images.githubusercontent.com/24752155/235526286-e99e6857-331d-4495-9474-c283f4a6f862.mp4

With the change, I see an extra content assist suggestion, that does the same as the other/old suggestion for the array element type:

https://user-images.githubusercontent.com/24752155/235526377-54cfa39a-a3e3-44fd-bb07-68d4acd1f35f.mp4

Is something missing in my environment, or in the snippet?

gayanper:
This will not show up in eclipse ui since the support for array completions in only added in jdt.ls. But the duplicate behavior is something I need to fix since its not acceptable.

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
```diff
@@ -12332,32 +12352,52 @@ private void findTypesFromExpectedTypes(char[] token, Scope scope, ObjectVector
...
-								proposal.setAccessibility(accessibility);
-								proposal.setArrayDimensions(dimensions);
-								this.requestor.accept(proposal);
-								if(DEBUG) {
-									this.printDebug(proposal);
+								char[] signature = getSignature(refBinding);
+								boolean found = false;
+								// check if we already found this completion before. if we have found then change it's
+								// dimension and relevance to match against expected type.
+								for (int j = 0; j < proposalsFound.size(); j++) {
```
trancexpress, line 12365: Any idea if this can cause trouble due to quadratic complexity?

Other than this, the change looks fine now.

trancexpress:
Also please squash before merging, as usual.

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
gayanper: Agree, did some improvements by eliminating the need for altering the already found completions, while also eliminating the duplicates. Lets see how the tests react.

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
```diff
@@ -11778,29 +11785,16 @@ private void findTypesAndPackages(char[] token, Scope scope, boolean proposeBase
 			}
 		}
 
+		checkCancel();
```
trancexpress, line 11788: I would prefer to not change this, it makes reviewing harder. Maybe as a separate change that improves the code.

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
```diff
@@ -11778,29 +11785,16 @@ private void findTypesAndPackages(char[] token, Scope scope, boolean proposeBase
...
-
-			checkCancel();
-
-			findTypesFromExpectedTypes(token, scope, typesFound, proposeType, proposeConstructor);
-		}
-
+		checkCancel();
 		if (isEmptyPrefix && !this.assistNodeIsAnnotation) {
-			if (!proposeConstructor) {
-				findTypesFromExpectedTypes(token, scope, typesFound, proposeType, proposeConstructor);
```
trancexpress, line 11802: If I understand the changes up until this point, this line and the previous line 11797 were moved above, to line 11643?

Could you explain what that move achieves?

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
```diff
@@ -12369,9 +12364,7 @@ private void findTypesFromExpectedTypes(char[] token, Scope scope, ObjectVector
 				} else if (expectedType instanceof BaseTypeBinding) {
 					BaseTypeBinding baseType = (BaseTypeBinding) expectedType;
 					// only go further if we are completing on an array.
-					if (dimensions > 0 && proposeType &&
-							(!this.assistNodeIsConstructor || !allowingLongComputationProposals)
```
trancexpress, line 12373: `(!this.assistNodeIsConstructor || !allowingLongComputationProposals)` is gone now, could you explain why?

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
```diff
@@ -12357,7 +12352,7 @@ private void findTypesFromExpectedTypes(char[] token, Scope scope, ObjectVector
 							}
 						}
 
-						if (proposeConstructor && dimensions == 0) {
```
trancexpress, line 12360: I assume `dimensions` can't be negative, so this must be OK to do.

trancexpress:
I see the completion suggestion tat adds `ArrayTest()` is gone now, I think this is a good thing. I'm not aware of syntax where it makes sense to add `ArrayTest()` after `ArrayTest[] arr = new`.

trancexpress:
Overall looks fine, I just don't understand some parts of the change. Functionality is as expected for the changed case.

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
gayanper: mainly we don't need the allowingLongComputationProposals here, this was copied from `findNestedTypes` in the original commit which added support for array completions. Even though it make sense checking for this in nested types, the same doesn't make sense here.

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
gayanper: A: The `findTypesFromExpectedTypes` finds current source type as a expected type. and since its a array type we create a completion even though is the same source type `(!inSameUnit || (inSameUnit && refBinding.isMemberType()) || (inSameUnit && isArrayCompletion)` at 12286

B: At the same time at 11666 where we go through the current types in the enclosing source type we find the same. 

Now in previous code, We find the current source type when running the code block at B:, And then we also found the same time at A:, but it was too late to change it since we have already added the type completion with dimension 0 insteand of correct array dimension.

By reordering this A: and B: to run A: first and B: after that, we could stop processing the current type if it was already handled in the `findTypesFromExpectedTypes` as part of array expected type. The exclusion happens at line 11715.

trancexpress (approved):

PR #1024, opened 2023-05-02
Fix NPE in SelectionParser.attachOrphanCompletionNode

### LINKED ISSUE #1023: NPE in SelectionParser.attachOrphanCompletionNode
For the following code:

```
	private void initOtherButtons() {
		String[] buttonNames3= null;
		switch (fTypeKind) {
			case CLASS_TYPE:
				if (fIsSealedSupported) {
					buttonNames3= new String[] {
							NewWizardMessages.NewTypeWizardPage_none_label,
							NewWizardMessages.NewTypeWizardPage_modifiers_sealed,
							NewWizardMessages.NewTypeWizardPage_modifiers_non_sealed,
							NewWizardMessages.NewTypeWizardPage_modifiers_final
					};
				}
				break;
			case INTERFACE_TYPE:
				if (fIsSealedSupported) {
					buttonNames3= new String[] {
							NewWizardMessages.NewTypeWizardPage_none_label,
							NewWizardMessages.NewTypeWizardPage_modifiers_sealed,
							NewWizardMessages.NewTypeWizardPage_modifiers_non_sealed
					};
				}
				//$FALL-THROUGH$
			default:
				// do nothing

		}

```

Running a search for unused NLS messages causes an NPE in SelectionParser.attachOrphanCompletionNode on the following lines:

```
	// if casestatement and not orphan, then add switch statement as assist node parent
	if (this.expressionPtr > 0 && this.assistNodeParent == null) {
		if (this.astPtr >= 0) {
			if (this.astStack[this.astPtr] instanceof CaseStatement) {
				// add switch statement as assistNodeParent
				Expression expression = this.expressionStack[this.expressionPtr];
				SwitchStatement switchStatement = new SwitchStatement();
				switchStatement.expression = this.expressionStack[this.expressionPtr - 1];
				if (this.astLengthPtr > -1 && this.astPtr > -1) {
					int length = this.astLengthStack[this.astLengthPtr];
					int newAstPtr = this.astPtr - length;
					ASTNode firstNode = this.astStack[newAstPtr + 1];
					if (length != 0 && switchStatement.expression != null && firstNode.sourceStart > switchStatement.expression.sourceEnd) {
```

What is happening is that in the first case statement above, the new String[] { causes a null entry in the expression stack.  When the code tries to find the switchStatement.expression, it gets the NPE.

jjohnstn:
- add null check
- fixes #1023

## What it does

Fixes NPE caused when searching for broken NLS message in eclipse.jdt.ui.

## How to test

Run Source -> Find Broken NLS Strings on org.eclipse.jdt.ui plug-in.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/select/SelectionParser.java
+++ b/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/select/SelectionParser.java
@@ -145,35 +145,6 @@ protected void attachOrphanCompletionNode(){
 			this.currentToken = 0; // given we are not on an eof, we do not want side effects caused by looked-ahead token
 		}
 	}
-	// if casestatement and not orphan, then add switch statement as assist node parent
-	if (this.expressionPtr > 0 && this.assistNodeParent == null) {
-		if (this.astPtr >= 0) {
-			if (this.astStack[this.astPtr] instanceof CaseStatement) {
-				// add switch statement as assistNodeParent
-				Expression expression = this.expressionStack[this.expressionPtr];
-				SwitchStatement switchStatement = new SwitchStatement();
-				switchStatement.expression = this.expressionStack[this.expressionPtr - 1];
-				if (this.astLengthPtr > -1 && this.astPtr > -1) {
-					int length = this.astLengthStack[this.astLengthPtr];
-					int newAstPtr = this.astPtr - length;
-					ASTNode firstNode = this.astStack[newAstPtr + 1];
-					if (length != 0 && switchStatement.expression != null && firstNode.sourceStart > switchStatement.expression.sourceEnd) {
-						switchStatement.statements = new Statement[length + 1];
-						System.arraycopy(this.astStack, newAstPtr + 1, switchStatement.statements, 0, length);
-					}
-
-				}
-				CaseStatement caseStatement = new CaseStatement(expression, expression.sourceStart,
-						expression.sourceEnd);
-				if (switchStatement.statements == null) {
-					switchStatement.statements = new Statement[] { caseStatement };
-				} else {
-					switchStatement.statements[switchStatement.statements.length - 1] = caseStatement;
-				}
-				this.assistNodeParent = switchStatement;
-			}
-		}
-	}
 }
 private void buildMoreCompletionContext(Expression expression) {
 	ASTNode parentNode = null;
--- a/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/select/SelectionParser.java
+++ b/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/select/SelectionParser.java
@@ -157,7 +157,7 @@ protected void attachOrphanCompletionNode(){
 					int length = this.astLengthStack[this.astLengthPtr];
 					int newAstPtr = this.astPtr - length;
 					ASTNode firstNode = this.astStack[newAstPtr + 1];
-					if (length != 0 && firstNode.sourceStart > switchStatement.expression.sourceEnd) {
+					if (length != 0 && switchStatement.expression != null && firstNode.sourceStart > switchStatement.expression.sourceEnd) {
 						switchStatement.statements = new Statement[length + 1];
 						System.arraycopy(this.astStack, newAstPtr + 1, switchStatement.statements, 0, length);
 					}


jjohnstn:
@iloveeclipse Could you take a quick look?  The code added in 2021 causes an NPE when you search for broken NLS messages in org.eclipse.jdt.ui.  In particular, the file NewTypeWizardPage has a switch statement where a case sets a String array with NLS strings.  The code in question is expecting the expressions stack to never a null entry.  I just added a null check and the Find broken NLS strings code works.  If you think the fact that the expression stack has a null entry is the real problem, then the fix is just a band-aid.

iloveeclipse:
Jeff, I have no insights into this code, but I see another missing NP check at line https://github.com/eclipse-jdt/eclipse.jdt.core/blob/ddcc2736212e6faf0d65e47060162f05cb5603ee/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/select/SelectionParser.java#LL188C6-L188C32.

I see @jarthana changed `SelectionParser` code most, so may be Jay can review that?

jarthana:
> @iloveeclipse Could you take a quick look? The code added in 2021 causes an NPE when you search for broken NLS messages in org.eclipse.jdt.ui. In particular, the file NewTypeWizardPage has a switch statement where a case sets a String array with NLS strings. The code in question is expecting the expressions stack to never a null entry. I just added a null check and the Find broken NLS strings code works. If you think the fact that the expression stack has a null entry is the real problem, then the fix is just a band-aid.

It's very unlikely that we add a null entry in the expressions stack, which is easier to find. But I suspect (and usually) the pointer is not at the correct spot. Is this something you can try and find?

jjohnstn:
Let me describe what is happening and perhaps you can point me in the right direction.  I have initiated a Find Broken NLS strings on NewWizardMessages.java.  In one instance of finding messages, it comes across the following in AccessRulesDialog:

```
	private String getDescriptionString() {
		String desc;
		String name= BasicElementLabels.getResourceName(fCurrElement.getPath().lastSegment());
		switch (fCurrElement.getEntryKind()) {
			case IClasspathEntry.CPE_CONTAINER:
				try {
					name= JavaElementLabels.getContainerEntryLabel(fCurrElement.getPath(), fCurrElement.getJavaProject());
				} catch (JavaModelException e) {
				}
				desc= NewWizardMessages.AccessRulesDialog_container_description;
				break;
			case IClasspathEntry.CPE_PROJECT:
				desc=  NewWizardMessages.AccessRulesDialog_project_description;
				break;
			default:
				desc=  NewWizardMessages.AccessRulesDialog_description;
		}

		return Messages.format(desc, name);
	}
```

Now the code in question gets triggered for the second case (IClassPathEntry.CPE_PROJECT).  On the expression stack at the expressionPtr is: <SelectOnName:NewWizardMessages.AccessRulesDialog_project_description> which is a SelectionOnQualifiedNameReference which I believe is fine.  Preceding it on the expression stack is a SingleNameReference to desc.  So it creates a switch expression as follows:  switch (desc) {} and creates a case for the SelectionOnQualifiedNameReference.

When it processes the following code in NewTypeWizardPage:

```
	private void initOtherButtons() {
		String[] buttonNames3= null;
		switch (fTypeKind) {
			case CLASS_TYPE:
				if (fIsSealedSupported) {
					buttonNames3= new String[] {
							NewWizardMessages.NewTypeWizardPage_none_label,
							NewWizardMessages.NewTypeWizardPage_modifiers_sealed,
							NewWizardMessages.NewTypeWizardPage_modifiers_non_sealed,
							NewWizardMessages.NewTypeWizardPage_modifiers_final
					};
```
It is triggered for the first case and is pointing on the expression stack to: <SelectOnName:NewWizardMessages.NewTypeWizardPage_none_label> which is a SelectionOnQualifiedNameReference.  The preceding item on the expression stack is null and preceding that is the SingleNameReference to buttonsName3.  It appears that a null is put in the expression stack for the new String[] array creation.

From there, it is creating a switch statement with a null expression and then adds a case for the SelectionOnQualifiedNameReference.  If it tries to get the switch expression sourceEnd, it gets the NPE.

I added the null check but the switch statement having a null expression doesn't seem right. 

If you think that the new String[] should have put a non-null value on the expression stack or should have been omitted, then that is one path to look at.  Alternatively, if the switch expression could have looked farther up the expression stack, what would you expect the switch statement to look like (e.g. if it found buttonNames3, it would have to calculate it is an array and add an index such as: switch (buttonNames3[0]) {}

jjohnstn:
@jarthana @iloveeclipse I tried looking at the original commit that added the code in question which was from the Java 17 branch.  When I ran the tests that were added for that commit, they do not exercise that added code.  I tried deleting the code and all tests are running correctly as well as the original scenarios reported in the bug that fueled the change so my take is to simply remove the code as extraneous.  As seen above, everything is testing fine.  The logic of the code in question does not appear to be sound as it is generating switch statements with invalid expressions and case statements for NLS string references when used in a Find Broken Externalized Strings action.  The action works fine without the code so I believe it has become extraneous with subsequent changes to Java 17.

jarthana:
Sorry for responding late. I just managed to reproduce this yesterday and planned to investigate today.

But your finding sounds accurate and I am not surprised. The fix was actually for a completion bug and I have no idea why SelectionParser had to be changed. And there were no new selection tests either for us to verify. Only time I see that code being hit is for some tests in JavaSearchBugs17Tests#testBug573943_018(), but I don't the code or the lack of it having any impact on the outcome.

So, +1 for the removal of the code.

jjohnstn:
Thanks @jarthana

PR #1032, opened 2023-05-06
Improve the constructor expected type fix #979

### LINKED ISSUE #979: Expected type of the first parameter is not populated in context for constructors

gayanper:
## What it does
The original fix didn't exit when there are no constructors found which results in suggesting visible symbols in current enclosing scope.

Now the fix is improved so that if we don't find constructors we will not continue to look for visible symbols that could match the argument type at the current position.

## How to test
Unit Test in jdt.core and jdt.ui.text

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests18.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests18.java
@@ -6955,4 +6955,31 @@ public void testGH979_on1stConstructorArgument_expectCompletionsMatchinType() th
 					+ (R_DEFAULT + R_RESOLVED + R_INTERESTING + R_CASE + R_UNQUALIFIED + R_NON_RESTRICTED) + "}",
 			result);
 }
+
+public void testGH979_onAnonClassConstructorWith_expectOnlyAnonClassCompletion() throws JavaModelException {
+	this.workingCopies = new ICompilationUnit[1];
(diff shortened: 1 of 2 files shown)


org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests18.java
```diff
@@ -6955,4 +6955,31 @@ public static GH979List<String> newInstance() {
...
+
+public void testGH979_onAnonClassConstructorWith_expectOnlyAnonClassCompletion() throws JavaModelException {
+	this.workingCopies = new ICompilationUnit[1];
+	this.workingCopies[0] = getWorkingCopy("/Completion/src/GH979.java", """
+		import java.io.Serializable;
+		public class GH979 {
+				public GH979() {}
+
+				public void foo() {
+					Serializable run= new Serializable() {
```
trancexpress, line 6967: Is it possible to use a local type and not a JVM type?

gayanper: Just copied and merge the snippet for jdt.ui.text test. But good suggestion, now i changed it.

trancexpress (approved):

PR #1038, opened 2023-05-09
Fix expected type on constructor first arg. Fix #1037

### LINKED ISSUE #1037: Expected types for constructor is not calculated when it is nested in chain of invocations
```
			public class GH979List<T> {

			}

			public class GH979 {
				public GH979(GH979List<String> names, int age, boolean valid) {}

				public static void foo() {
					GH979.instance().boo(new GH979(, 20, false));
				}

				public GH979 instance() {
					return null;
				}

				public void boo(GH979 g) {

				}

				public static GH979List<String> newInstance() {
					return null;
				}
```

Trying to complete behind `GH979.instance().boo(new GH979(` will not identify GH979List as a expected because the completion node is identified as `"<CompleteOnAllocationExpression:new GH979(<CompleteOnName:>, 20, false)>"`

gayanper:
## What it does
This fix handles situation where the constructor in inside a method invocation which is part of chain method invocations. The fix tries to handle the <CompletionOnName:> which is found in this scenario and interpret it as no arguments.

## How to test
Unit Tests and snippet provided in #1037 

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
+++ b/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
@@ -68,6 +68,7 @@ import org.eclipse.jdt.core.search.SearchPattern;
 import org.eclipse.jdt.core.search.SearchRequestor;
 import org.eclipse.jdt.internal.codeassist.complete.AssistNodeParentAnnotationArrayInitializer;
 import org.eclipse.jdt.internal.codeassist.complete.CompletionJavadoc;
+import org.eclipse.jdt.internal.codeassist.complete.CompletionNode;
 import org.eclipse.jdt.internal.codeassist.complete.CompletionNodeDetector;
 import org.eclipse.jdt.internal.codeassist.complete.CompletionNodeFound;
 import org.eclipse.jdt.internal.codeassist.complete.CompletionOnAnnotationOfType;
@@ -3335,7 +3336,23 @@ public final class CompletionEngine
 				-1);
 		}
 
-		findCompletionsForArgumentPosition(methodsFound, argTypes != null ? argTypes.length : 0, scope);
+		findCompletionsForArgumentPosition(methodsFound, findCompletionArgIndex(messageSend.arguments), scope);
+	}
+
+	private int findCompletionArgIndex(Expression[] arguments) {
+		if (arguments == null) {
+			return 0;
+		}
+		for (int index = 0; index < arguments.length; index++) {
+			if ((arguments[index] instanceof CompletionNode)
+					// the following handles when the cursor is at the receiver in a completed expression like
+					// new PersonDetails(GH969.emptyList(), 0) here the cursor is after new PersonDetails(
+					|| (arguments[index] instanceof MessageSend ms && ms.receiver instanceof CompletionNode)
+					|| (arguments[index] instanceof FieldReference fr && fr.receiver instanceof CompletionNode)) {
+				return index;
+			}
+		}
+		return arguments.length;
 	}
 
 	private void findCompletionsForArgumentPosition(ObjectVector methodsFound, int completedArgumentLength, Scope scope) {
@@ -3690,13 +3707,7 @@ public final class CompletionEngine
 					false);
 			}
 		}
-		// here we assume that if first type is null we have a <CompleteOnName:>
-		int completedArgLen = 0;
-		if (argTypes != null) {
-			completedArgLen = argTypes[0] == null ? 0 : argTypes.length;
-		}
-
-		findCompletionsForArgumentPosition(constructorsFound, completedArgLen, scope);
+		findCompletionsForArgumentPosition(constructorsFound, findCompletionArgIndex(allocExpression.arguments), scope);
 	}
 
 	private void completionOnQualifiedNameReference(ASTNode astNode, ASTNode enclosingNode, Binding qualifiedBinding,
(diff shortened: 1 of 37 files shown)


org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
```diff
@@ -3689,7 +3689,13 @@ private void completionOnQualifiedAllocationExpression(ASTNode astNode, Binding
 					false);
 			}
 		}
-		findCompletionsForArgumentPosition(constructorsFound, argTypes != null ? argTypes.length : 0, scope);
+		// here we assume that if first type is null we have a <CompleteOnName:>
```
trancexpress, line 3692: Sorry, I'm not knowledgeable of the completion engine. Can you explain what is `CompleteOnName:`?

stephan-herrmann: This seems to work OK, but I wonder, why we should use such indirect reasoning. Could this break if a provided first argument is not resolvable? What would be the `resolvedType` in that case, `null` or some `Problem(Reference)Binding`?

Alternatively we could explicitly check if `allocExpression.arguments[0]` is the completion node. I'm just surprised we don't seem to have a simple query that would answer this, but that could of course be added right now.

gayanper: @stephan-herrmann This is how the parser.assistNode looks like

`CompletionOnQualifiedAllocationExpression@103 "<CompleteOnAllocationExpression:new PersonDetails(<CompleteOnMessageSendName:()>, 0)>"`

I also checked when the first argument is not resolvable, then it will be null as well. But we only go through here for the first argument of the constructor since we want to also suggest constructor signature as completion items which is used for signature help in jdt ui and jdt.ls.

We cannot use `allocExpression.arguments[0]` since it will not be the actual completion node as I have shown above. 

Now since we don't hit this method other than for the first parameter, I wonder why we really need to pass the index based on arguments, we could just pass 0. WDYT ?

gayanper: @stephan-herrmann could you have a look at my last comment and let me know if we need to change the fix or if this is ok for merging ?

stephan-herrmann: Relying on `argTypes[0] == null` depends on how `SelectionOnSingleNameReference.resolveType()` at all are implemented. Incidentally, in #1077 I'm working on a change that _could_ indeed cause `resolveType()` to invoke super and thus set a non-null `resolvedType`. I haven't checked if that is a real or only theoretical conflict between both changes.

> We cannot use allocExpression.arguments[0] since it will not be the actual completion node as I have shown above.


For completeness, here's a sketch of what I was thinking of:
```
	if (arguments != null && arguments.length > 0 && arguments[0] instanceof CompletionOnSingleNameReference) {
		completedArgLen = 0; // just confirming :)
	} else if (argTypes != null) {
		completedArgLen = argTypes.length;
	}
```
Tests pass, but of course the instanceof check is incomplete, might be improved e.g., by a marker interface AnyCompletionNode or similar approach.

This would formalize the case were indeed we are completing a first argument.

But maybe all this is overkill.

> Now since we don't hit this method other than for the first parameter, I wonder why we really need to pass the index based on arguments, we could just pass 0. WDYT ?

That adds another indirect reasoning. Why should we be sure that it's only for the first parameter? Class header comment in `CompletionOnQualifiedAllocationExpression` says differently, no?

stephan-herrmann:
> Tests / Build — failed: 2, passed: 75391 

did my suggestions cause this?

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/complete/CompletionOnReferenceExpressionName.java
```diff
@@ -44,10 +44,14 @@ public TypeBinding resolveType(BlockScope scope) {
...
-				if ((this.resolvedTypeArguments[i] = typeReference.resolveType(scope, true /* check bounds*/)) == null) {
+				if ((this.resolvedTypeArguments[i] = typeReference.resolveType(scope,
+						true /* check bounds */)) == null) {
 					typeArgumentsHaveErrors = true;
 				}
-				if (typeArgumentsHaveErrors && typeReference instanceof Wildcard) { // resolveType on wildcard always return null above, resolveTypeArgument is the real workhorse.
+				if (typeArgumentsHaveErrors && typeReference instanceof Wildcard) { // resolveType on wildcard always
+																					// return null above,
+																					// resolveTypeArgument is the real
+																					// workhorse.
```
stephan-herrmann, line 54: who's auto-formatting unchanged code?

gayanper:
> > Tests / Build — failed: 2, passed: 75391
> 
> did my suggestions cause this?

The first failure is highly unlikely. But the second failure in CompletionTests3 is due to now getting more completions resolved with this improvement. I have updated the test.

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/complete/CompletionOnReferenceExpressionName.java
gayanper: I will fix it, I'm manually formatting the code since the auto formatter sometimes change unwanted lines which are int he edge of a modification. And i time to time use vscode with jdt.ls for development with the eclipse default formatter.
I will be vigilant in future to make sure with white space issues since I have been make bit a lot last few PRs :(.

gayanper:
@stephan-herrmann the issues are fixed now. Let me know what you think ?

stephan-herrmann (approved):
All looks well to me, thanks.

Issue #1039, opened 2023-05-06
Junit4 test finder fails to find tests


vgcpge:
My tests had worked fine in 2022-12, but do not work in any later Eclipse release.

### Test steps
Import project https://github.com/basilevs/jstackfilter/tree/0.0.2/library
Create a new launch configuration that executes all tests in the project.
Run it

### Expected result
Some tests are found and executed, just like in 2022-12

### Actual result
No tests found.
No errors in error log.

The project in question is a bit strange - it is a Java modular project, that also plays a role of Eclipse plugin. Hence, it includes the tests folder (explicitly marked so).
Before the disaster I could run these tests in both pure JDT and in PDE.

trancexpress:
The `@Test` annotation type is apparently not found. A `ProblemReferenceBinding` is created here:

```
"ModalContext" #159 prio=6 os_prio=0 cpu=75.42ms elapsed=3081.66s tid=0x00007f42a2b35590 nid=0x11856 at breakpoint [0x00007f41679fd000]
   java.lang.Thread.State: RUNNABLE
        at org.eclipse.jdt.internal.compiler.lookup.ProblemReferenceBinding.<init>(ProblemReferenceBinding.java:30)
        at org.eclipse.jdt.internal.compiler.lookup.LookupEnvironment.getInaccessibleBinding(LookupEnvironment.java:2318)
        at org.eclipse.jdt.internal.compiler.lookup.LookupEnvironment.getInaccessibleBinding(LookupEnvironment.java:2289)
        at org.eclipse.jdt.internal.compiler.lookup.CompilationUnitScope.findImport(CompilationUnitScope.java:607)
        at org.eclipse.jdt.internal.compiler.lookup.CompilationUnitScope.findSingleImport(CompilationUnitScope.java:650)
        at org.eclipse.jdt.internal.compiler.lookup.CompilationUnitScope.faultInImports(CompilationUnitScope.java:479)
        at org.eclipse.jdt.internal.compiler.lookup.CompilationUnitScope.faultInTypes(CompilationUnitScope.java:552)
        at org.eclipse.jdt.internal.core.search.matching.MatchLocator.process(MatchLocator.java:2032)
        at org.eclipse.jdt.internal.core.search.matching.MatchLocator.locateMatches(MatchLocator.java:1364)
        at org.eclipse.jdt.internal.core.search.matching.MatchLocator.locateMatches(MatchLocator.java:1401)
[...]

PR #1040, opened 2023-05-09
Junit4 test finder fails to find tests in modular Java project

trancexpress:
The fix for #675 introduces a regression, since the JUnit container is not marked as modular. The search for JUnit types (e.g. `org.junit.Test`) however depends on the modular info attached for the `PackageFragmentRoot` of the JUnit container.

This change adjusts the fix for #675 to assume a modular container, if the Java project is modular.

Fixes: #1039

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/matching/JavaSearchNameEnvironment.java
+++ b/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/matching/JavaSearchNameEnvironment.java
@@ -594,29 +594,30 @@ private static boolean isComplianceJava9OrHigher(IJavaProject javaProject) {
 }
 
 private static boolean isOnModulePath(JavaProject javaProject, PackageFragmentRoot root) {
-	boolean isOnModulePath;
+	if (hasSystemModule(root)) {
+		/*
+		 * The JRE 9+ container is on the module path without the 'module' classpath attribute being set.
+		 * We detected the JRE container by checking the container for system modules.
+		 */
+		return true;
+	}
 	try {
 		IClasspathEntry classpathEntry = root.getRawClasspathEntry();
-		if (classpathEntry.getEntryKind() == IClasspathEntry.CPE_SOURCE) {
+		if (ClasspathEntry.isModular(classpathEntry)) {
+			return true;
+		}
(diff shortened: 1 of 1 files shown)


trancexpress:
I really don't know about this. From my POV its correct to assume JUnit is not on the module path, in the given situation. What the JDT search does with this fix (and before changes for #675), is add everything from the JUnit container to the module of the Java project... that can't be right, since the JUnit container itself has no modules...

But I don't find a simple way to add the contents of the container to the unnamed module. With the bugs we have fixed so far in this area, we only cared about source folders and jar libraries. If we run into search problems for containers (which is not unlikely) I guess we'll have to revisit this fix.

iloveeclipse:
> From my POV its correct to assume JUnit is not on the module path, in the given situation. What the JDT search does with this fix (and before changes for #675), is add everything from the JUnit container to the module of the Java project... that can't be right, since everything in the JUnit container doesn't define modules...

This is my opinion too. If we move JUnit to the module path (and update module dependency to junit), the test is found.

> But I don't find a simple way to add the contents of the container to the unnamed module

Then we probably need a complicated way :-)

iloveeclipse:
Converted PR to draft for now.

trancexpress:
> Converted PR to draft for now.

From my POV better to merge as is, to fix the regression from #675. We can add a TODO to try to find a better way to fix the original bug (adding the container on the module path, even though it doesn't have modular libraries).

iloveeclipse:
> > Converted PR to draft for now.
> 
> From my POV better to merge as is, to fix the regression from #675. We can add a TODO to try to find a better way to fix the original bug (adding the container on the module path, even though it doesn't have modular libraries).

I'm not convinced that the fix here is correct. The JUnit container **is** on classpath but **not** on module path. The fix puts it to the later one (or does something similar answering "yes" from `isOnModulePath()`), which is questionable and will most likely break other cases again.

trancexpress:
> The fix puts it to the later one (or does something similar answering "yes" from `isOnModulePath()`), which is questionable and will most likely break other cases again.

The only thing the change does is disable the fix for #675 in the case of containers, in case of a modular project. Do you actually care for call hierarchy working (w.r.t. #675) those cases?

> I'm not convinced that the fix here is correct.

I don't imagine another fix is coming soon.

basilevs:
Eclipse PDE uses a container for plug-in  dependencies. Call-hierarchy being glitchy there is quite annoying. Will PDE work?

iloveeclipse:
Superceeded by https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1066

Issue #1035, opened 2023-05-08
SourceField.isRecordComponent() returns false for record component


trancexpress:
Seen with the fix for https://github.com/eclipse-jdt/eclipse.jdt.ui/issue/571, https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/572.

To reproduce, use that fix in JDT UI. Create 2 source files:

```
public record Car(int wheels) {}
```
```
public class Foo {
    void foo(Car c) {
        var wheels = c.wheels();
    }
}
```

Select `wheels` in `Foo.java`, compute call hierarchy, observe that the code added for https://github.com/eclipse-jdt/eclipse.jdt.ui/issue/571 evaluates `SourceField.isRecordComponent()` to `false`, while in `CallerMethodWrapper.getAccurateSearchScope()`. So the next block (that checks if the parent type is a record) is required to ensure a correct call hierarchy.

Selecting `wheels` in `Car.java` doesn't run into this problem, `isRecordCompoent()` returns the expected `true`.

The member that returns false for `isRecordComponent()` is of type `SourceField`. The member that returns the expected true is of an anonymous type and is created here: `CompilationUnitStructureRequestor.createRecordComponent()`

iloveeclipse:
This seem to be regression from [bug 565211](https://bugs.eclipse.org/bugs/show_bug.cgi?id=565211) / commit 7f0b7a275edf8d5988fd058f6d1144e100d714e0 that changed code added for [bug 562637](https://bugs.eclipse.org/bugs/show_bug.cgi?id=562637) / commit aa930daf6545d757b7e55d1cfef67c6ba4fa3e4f.

The 7f0b7a275edf8d5988fd058f6d1144e100d714e0 removed `notifySourceElementRequestor(RecordComponent)` that called `requestor.exitRecordComponent` where `SourceFieldElementInfo.isRecordComponent = true;` was set, and the code uses now only `notifySourceElementRequestor(FieldDeclaration)`. In this one, the `SourceFieldElementInfo.isRecordComponent` is not changed.

I will push a fix in a moment.

iloveeclipse:
Note: same issue happens also with BinaryField elements, see #1059 tests.

PR #1059, opened 2023-05-12
*Field.isRecordComponent() returns false for record component (#1035)

iloveeclipse:
**First commit**
Cleanup after https://github.com/eclipse-jdt/eclipse.jdt.core/commit/7f0b7a275edf8d5988fd058f6d1144e100d714e0

`ISourceElementRequestor.exitRecordComponent()` is not called anymore,
hence this is all dead code and can be removed without affecting
functionality.

**Second commit**
This seem to be regression from bug 565211 / commit 7f0b7a2 that changed
code added for bug 562637 / commit aa930da.

The 7f0b7a2 removed `notifySourceElementRequestor(RecordComponent)` that
called `requestor.exitRecordComponent` where
`SourceFieldElementInfo.isRecordComponent = true;` was set, and the code
uses now only `notifySourceElementRequestor(FieldDeclaration)`. In this
one, the `SourceFieldElementInfo.isRecordComponent` is not changed.

Fix is to update exitField() to transfer also
`fieldInfo.isRecordComponent` value to `SourceFieldElementInfo`.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1035

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/BinaryType.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/BinaryType.java
@@ -337,7 +337,12 @@ private IField[] getFieldsOrComponents(boolean component) throws JavaModelExcept
 public IField getRecordComponent(String compName) {
 	try {
 		if (isRecord())
-			return new BinaryField(this, compName);
+			return new BinaryField(this, compName) {
+				@Override
+				public boolean isRecordComponent() throws JavaModelException {
+					return true;
+				}
+			};
 	} catch (JavaModelException e) {
 		//
 	}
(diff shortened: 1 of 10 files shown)


org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/CompilationUnitStructureRequestor.java
```diff
@@ -697,6 +697,7 @@ public void exitField(int initializationStart, int declarationEnd, int declarati
 	info.setFlags(fieldInfo.modifiers);
 	char[] typeName = JavaModelManager.getJavaModelManager().intern(fieldInfo.type);
 	info.setTypeName(typeName);
+	info.isRecordComponent = fieldInfo.isRecordComponent;
```
iloveeclipse, line 700: Actual fix is here

iloveeclipse:
I've updated tests locally to show the issue but found yet another bug, in binary types too. Will update later.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/BinaryType.java
```diff
@@ -337,7 +337,12 @@ private IField[] getFieldsOrComponents(boolean component) throws JavaModelExcept
 public IField getRecordComponent(String compName) {
 	try {
 		if (isRecord())
-			return new BinaryField(this, compName);
+			return new BinaryField(this, compName) {
+				@Override
+				public boolean isRecordComponent() throws JavaModelException {
+					return true;
+				}
```
iloveeclipse, line 344: This is the second fix for BinaryField

PR #1066, opened 2023-05-16
Junit4 test finder fails to find tests in modular Java project

iloveeclipse:
The fix for #675 introduces a regression, if the JUnit container is placed on classpath and project is modular. In this case, the container was not added to any module and the org.junit.Test annotation can not be resolved / not found by exact search (JUnit4TestFinder accepts only resolved matches with SearchMatch.A_ACCURATE).

This change refactors the code once again, trying to simplify module to package fragment mapping.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1039


trancexpress:
Seems fine, I guess we can proceed with merging.

Issue #986, opened 2023-04-19
NullPointerException: Cannot invoke "TypeBinding.kind()" because "type" is null


liefke:
I just migrated from Eclipse 2022-12 to 2023-03.

Now I get the following exception when building or editing some of the classes in my project:

```
java.lang.NullPointerException: Cannot invoke "org.eclipse.jdt.internal.compiler.lookup.TypeBinding.kind()" because "type" is null
	at org.eclipse.jdt.internal.compiler.lookup.LookupEnvironment.convertToRawType(LookupEnvironment.java:844)
	at org.eclipse.jdt.internal.compiler.lookup.RawTypeBinding.initializeArguments(RawTypeBinding.java:238)
	at org.eclipse.jdt.internal.compiler.lookup.ParameterizedTypeBinding.substitute(ParameterizedTypeBinding.java:1432)
	at org.eclipse.jdt.internal.compiler.lookup.Scope$Substitutor.substitute(Scope.java:586)
	at org.eclipse.jdt.internal.compiler.lookup.Scope.substitute(Scope.java:530)
	at org.eclipse.jdt.internal.compiler.lookup.ParameterizedFieldBinding.<init>(ParameterizedFieldBinding.java:36)
	at org.eclipse.jdt.internal.compiler.lookup.ParameterizedTypeBinding.fields(ParameterizedTypeBinding.java:609)
	at org.eclipse.jdt.internal.compiler.lookup.ParameterizedTypeBinding.getField(ParameterizedTypeBinding.java:803)
	at org.eclipse.jdt.internal.compiler.lookup.Scope.findField(Scope.java:1437)
	at org.eclipse.jdt.internal.compiler.lookup.Scope.findField(Scope.java:1383)
	at org.eclipse.jdt.internal.compiler.lookup.MethodScope.findField(MethodScope.java:473)
	at org.eclipse.jdt.internal.compiler.lookup.BlockScope.getBinding(BlockScope.java:623)
	at org.eclipse.jdt.internal.compiler.ast.QualifiedNameReference.resolveType(QualifiedNameReference.java:1029)
	at org.eclipse.jdt.internal.compiler.ast.MemberValuePair.resolveTypeExpecting(MemberValuePair.java:93)
	at org.eclipse.jdt.internal.compiler.ast.Annotation.resolveType(Annotation.java:932)
	at org.eclipse.jdt.internal.compiler.ast.ASTNode.resolveAnnotations(ASTNode.java:958)
[...]

Issue #854, opened 2023-03-17
Error marker when null analysis enabled


snjeza:
A related issue - https://github.com/redhat-developer/vscode-java/issues/3002

Steps to reproduce:
- import[ test project](https://github.com/snjeza/vscode-test/raw/master/testissue3002.zip)
- you will get this error
```
Type mismatch: cannot convert from Class<App.MyInit> to Class<? extends Init<? extends Configuration>>
```

stephan-herrmann:
This problem is covered by #1067

stephan-herrmann:
New test has been disabled via #1079 to be re-fixed via #1077

stephan-herrmann:
Resolved via #1077

Issue #875, opened 2023-03-21
Annotation Type mismatch: cannot convert from Class<...> to Class<? extends AbstractDeclarativeValidator>


rubenporras:
After upgrading to Eclipse 2023-03, on usages of the following Xtext annotation (org.eclipse.xtext.validation.ComposedChecks)
```
@Retention(RetentionPolicy.RUNTIME)
@Target({ ElementType.TYPE})
public @interface ComposedChecks {

	Class<? extends AbstractDeclarativeValidator>[] validators();

}
```

I get the error "Type mismatch: cannot convert from Class<AvqScriptJavaValidator> to Class<? extends AbstractDeclarativeValidator>". However, the type `AvqScriptJavaValidator` extends `AbstractDeclarativeValidator`, see the screenshot:

![image](https://user-images.githubusercontent.com/43636626/226542945-ba126ba2-6d8e-473c-a77a-0f815dcec2da.png)

rubenporras:
As additional information, the code was compiling without errors in Eclipse 2022-12

iloveeclipse:
Please try to provide a self containing reproducer project.

srikanth-sankaran:
Do you perhaps have annotation based null analysis turned on ? I am wondering if this is a duplicate of https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1007

rubenporras:
Yes, I do have them enabled. I will disable them and try again.

rubenporras:
Hi @srikanth-sankaran , I can confirm that disabling the JDT annotation base analysis by using 
`org.eclipse.jdt.core.compiler.annotation.nullanalysis=disabled`
fixes the problem.

Also, if instead of 
```
@CompilerCatalog(compilers = {XCheck.class})
public class AvqBaseCheckValidator extends AcfCheckValidator {

  @Override
  protected List<EPackage> getEPackages() {
    return null;
  }
}
```
I just have

```
@CompilerCatalog(compilers = {XCheck.class})
public class AvqBaseCheckValidator {

  protected List<EPackage> getEPackages() {
    return null;
  }
}
```

then the error marker also disappear. Do you think you still need a self contained reproducer?

rubenporras:
@srikanth-sankaran , also removing the brackets as commented in ttps://github.com/eclipse-jdt/eclipse.jdt.core/issues/1007#issuecomment-1534834099 removes the marker. Just to make it clear, the following has also no marker:

```
[...]

PR #1067, opened 2023-05-16
False type mismatch on array initializer with null annotations (regression 2023-03)

### LINKED ISSUE #1007: False type mismatch on array initializer with null annotations (regression 2023-03)
The attached example compiles in 2022-12, but not in 2023-03 or jdt.core head:
grafik
SubClass is a sub class of SuperClass, but null annotations are given for some of the involved declarations.

[reproducer.zip](https://github.com/eclipse-jdt/eclipse.jdt.core/files/11322052/reproducer.zip)

stephan-herrmann:
fixes #1007

Here I removed one code block from #630 and made resolving on one particular call path deferrable.

This breaks the following unfortunate chain:

1. A type's hierarchy was being connected
2. ClassScope.findSupertype() triggered initializing the null default 
(since [#630](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/630), still needed and retained)
4. This triggered resolving annotations
5. An annotation had member value pairs, where values were type checked against the annotation's attributes
6. The value type contained a type that did not yet have super types connected
7. This triggered ahead of time  resolving
( [#630](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/630), now removed)
9. That ahead of time resolving caused havoc in  [#1007](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1007)

Starting at (4) different test cases had different ways to cause havoc (different ways to trigger undesired ahead-of-time side-effects). So that point offered itself as a suitable location for severing the unfortunate chain and start deferring instead.

The complex part is making the set of deferred tasks as narrow as possible (see new method `Annotation.shouldDeferResolvingDetails()`).

I merged the solution into prior art around `ClassScope#deferredBoundChecks`, which works fine. Only the name `ClassScope.checkParameterizedTypeBounds()` is no longer accurate (I simply forgot to change it).

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Annotation.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Annotation.java
@@ -903,6 +903,55 @@ public abstract class Annotation extends Expression {
 			return null;
 		}
 
+		if (!shouldDeferResolvingDetails(scope, (ReferenceBinding) typeBinding)
+			|| !scope.classScope().deferCheck(() -> resolveDetails(scope)))
+		{
+			resolveDetails(scope);
+		}
+		return this.resolvedType;
+	}
+	/** check if resolving annotation details right now risks undesired (ahead-of-time) side effects. */
+	private boolean shouldDeferResolvingDetails(BlockScope scope, ReferenceBinding annotationType) {
+		if (this instanceof MarkerAnnotation)
+			return false; // no details, no need to defer
+		ReferenceBinding enclosing = scope.enclosingReceiverType();
+		if (enclosing == null)
+			return false; // happens in module scope
+		// risk may exist when resolving annotation on the enclosing type:
+		if (this.recipient != enclosing.original())
+			return false;
+		// risk exist only if that enclosing type or one of its members isn't yet ready:
+		if (isReadyIncludingMembers(enclosing))
+			return false;
+		// exclude @NonNullByDefault, because
+		// * caller may need detail information
+		// * member-value-pairs cannot point back into the unconnected type
+		if (annotationType.hasNullBit(TypeIds.BitNonNullByDefaultAnnotation))
+			return false;
+		// exclude annotations on @NonNull/@Nullable itself because
+		// * tagBits needs to be set right now
+		// * the current (meta) annotation 'annotationType' cannot possibly point back into 'enclosing':
+		if (enclosing.hasNullBit(TypeIds.BitNonNullAnnotation|TypeIds.BitNullableAnnotation))
+			return false;
+		return true;
+	}
+
+	protected boolean isReadyIncludingMembers(ReferenceBinding referenceType) {
+		if ((referenceType.tagBits & TagBits.EndHierarchyCheck) == 0)
+			return false;
+		if (referenceType instanceof SourceTypeBinding) {
+			ReferenceBinding[] members = ((SourceTypeBinding) referenceType).memberTypes;
+			if (members != null) {
+				for (ReferenceBinding member : members) {
+					if (!isReadyIncludingMembers(member))
+						return false;
+				}
+			}
+		}
+		return true;
+	}
+
+	private void resolveDetails(BlockScope scope) {
 		ReferenceBinding annotationType = (ReferenceBinding) this.resolvedType;
 		MethodBinding[] methods = annotationType.methods();
 		// clone valuePairs to keep track of unused ones
@@ -1112,7 +1161,6 @@ public abstract class Annotation extends Expression {
 			}
 			checkAnnotationTarget(this, scope, annotationType, kind, this.recipient, tagBits & TagBits.AnnotationNullMASK);
 		}
-		return this.resolvedType;
 	}
 
 	public long handleNonNullByDefault(BlockScope scope) {
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
@@ -60,7 +60,6 @@ import org.eclipse.jdt.core.compiler.InvalidInputException;
 import org.eclipse.jdt.internal.compiler.ast.LambdaExpression;
 import org.eclipse.jdt.internal.compiler.ast.MethodDeclaration;
 import org.eclipse.jdt.internal.compiler.ast.NullAnnotationMatching;
-import org.eclipse.jdt.internal.compiler.ast.TypeReference;
 import org.eclipse.jdt.internal.compiler.classfmt.ClassFileConstants;
 import org.eclipse.jdt.internal.compiler.impl.CompilerOptions;
 import org.eclipse.jdt.internal.compiler.impl.JavaFeature;
@@ -1353,27 +1352,6 @@ public boolean implementsInterface(ReferenceBinding anInterface, boolean searchH
 				for (int b = 0; b < nextPosition; b++)
 					if (TypeBinding.equalsEquals(next, interfacesToVisit[b])) continue nextInterface;
 				interfacesToVisit[nextPosition++] = next;
-			}
-		}
-	}
-	// see https://github.com/eclipse-jdt/eclipse.jdt.core/issues/629
-	if (nextPosition == 0 && this instanceof SourceTypeBinding) {
-		SourceTypeBinding sourceType = (SourceTypeBinding) this;
-		if (sourceType.scope != null && sourceType.scope.referenceContext != null && sourceType.scope.compilerOptions().isAnnotationBasedNullAnalysisEnabled) {
-			TypeReference[] references = sourceType.scope.referenceContext.superInterfaces;
-			if (references == null || references.length == 0) {
-				return false;
-			}
-			for (TypeReference reference: references) {
-				if (!(reference.resolvedType instanceof ReferenceBinding)) {
-					reference.resolveType(sourceType.scope);
-				}
-				if (reference.resolvedType instanceof ReferenceBinding) {
-					ReferenceBinding binding = (ReferenceBinding) reference.resolvedType;
-					if (binding.isEquivalentTo(anInterface)) {
-						return true;
-					}
-				}
 			}
 		}
 	}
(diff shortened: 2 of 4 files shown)


stephan-herrmann:
@srikanth-sankaran , @iloveeclipse I plan to have a look at other issues linked from #1007 on Thursday, with a goal to have this merged before RC1.

Until then feel free to have a look, test against other reports, improve ... whatever may be needed to get this into good shape in time.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ParameterizedTypeBinding.java
```diff
@@ -1062,6 +1062,24 @@ public boolean isHierarchyConnected() {
...
 
+	@Override
+	public boolean hasUnconnectedSourceHierarchy() {
+		if (this.type.hasUnconnectedSourceHierarchy())
+			return true;
+		if (this.arguments != null) {
+			enterRecursiveFunction();
+			try {
+				for (TypeBinding argument : this.arguments) {
+					if (argument == null || argument.hasUnconnectedSourceHierarchy())
```
stephan-herrmann, line 1073: For this to fully make sense, I may have to override `hasUnconnectedSourceHierarchy` also in other types like `TypeVariableBinding`, `WildcardBinding` etc.

Until then `enterRecursiveFunction` is more of a scarecrow than actually doing work, I suppose.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeBinding.java
```diff
@@ -1765,6 +1765,14 @@ public void exitRecursiveFunction() {
 public boolean isFunctionalType() {
 	return false;
 }
+
+/**
+ * Returns true if the type or one of its constituents is a SourceTypeBinding that has not yet connected its hierarchy
+ */
+public boolean hasUnconnectedSourceHierarchy() {
```
stephan-herrmann, line 1772: I was tempted to use (and extend) the existing `isHierarchyConnected()`, but specifically regarding `BinaryTypeBinding` the requirements seem to be incompatible.

snjeza:
The PR fixes #1007, tests pass.
It also fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/854
The issue described at https://github.com/redhat-developer/vscode-java/issues/3076#issuecomment-1542691617 isn't fixed.
It is caused by
> ClassScope.findSupertype() triggered initializing the null default
(since https://github.com/eclipse-jdt/eclipse.jdt.core/pull/630, still needed and retained)

snjeza:
> The issue described at https://github.com/redhat-developer/vscode-java/issues/3076#issuecomment-1542691617

A simple project example: 
[test3076.zip](https://github.com/eclipse-jdt/eclipse.jdt.core/files/11493231/test3076.zip)

A potential patch:
```
diff --git a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
index e45721be10..c6b68c2cb2 100644
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
@@ -1539,7 +1539,7 @@ public abstract class Scope {
 
        // Internal use only
        public ReferenceBinding findMemberType(char[] typeName, ReferenceBinding enclosingType) {
-               if ((enclosingType.tagBits & TagBits.HasNoMemberTypes) != 0)
+               if ((enclosingType.tagBits & TagBits.HasNoMemberTypes) != 0 && enclosingType.hasUnconnectedSourceHierarchy())
                        return null; // know it has no member types (nor inherited member types)
 
                ReferenceBinding enclosingSourceType = enclosingSourceType();

```

stephan-herrmann:
Thanks, @snjeza  I can reproduce.

stephan-herrmann:
> A potential patch:

Thanks, this helps to understand the chain of causality.

Still I prefer to avoid undesired side effects (ahead of time) rather than repairing broken results after the fact. I have a patch-in-progress, that fixes this issue, just need to resolve some hiccups in existing tests :)

stephan-herrmann:
The example from  redhat-developer/vscode-java#3076  led me to revise the implementation strategy.

Before updating the PR description, here's the old version for posterity:

> Here I removed one code block from https://github.com/eclipse-jdt/eclipse.jdt.core/pull/630 and made type checking on one particular call path deferrable.
>
> This breaks the following unfortunate chain:
>
> * A type's hierarchy was being connected
> * ClassScope.findSupertype() triggered initializing the null default
>    (since [#630](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/630), still needed and retained)
> * This triggered resolving annotations
> * An annotation had member value pairs, where values were type checked against the annotation's attributes
> * The value type contained a type that did not yet have super types connected
> * This triggered ahead of time resolving
>    ([#630](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/630), now removed)
> * That ahead of time resolving caused havoc in [#1007](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1007)
> 
> That chain is severed at the point where we went from resolving (OK) to type checking (NOK, because super type information was needed).
> 
> At first I thought I could merge the solution with prior art around ClassScope#deferredBoundChecks but that mechanism provides deferring only until the end of the current connectTypeHierarchy() but here we may have to wait until some other type has its hierarchy connected. For that reason I added a new list of DeferrableChecks to the central LookupEnvirnment, each equipped with its own enablement condition.
> 
> I wondered if iterating all deferred check tasks after each connectTypeHierarchy() could have an impact on performance, but for now the condition for deferring is so rare that I see no motivation for optimization yet (may revisit when other situations join in that mechanism).

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ParameterizedTypeBinding.java
stephan-herrmann: this method no longer exists

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeBinding.java
stephan-herrmann: this method no longer exists

stephan-herrmann:
>> We have a cluster of defects and need a fix that will address all of them. This is what I know at the moment:
>
> https://github.com/eclipse-jdt/eclipse.jdt.core/issues/854
> https://github.com/eclipse/eclipse.jdt.ls/issues/2386
> https://github.com/eclipse-jdt/eclipse.jdt.core/issues/629
> https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1007
> https://github.com/eclipse-jdt/eclipse.jdt.core/issues/875

Of these the following have been fixed with tests to witness:
* #854
* #629 (test from [#630](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/630))
* #1007 

eclipse/eclipse.jdt.ls#2386 is just the downstream of #629.

I don't (yet) have a test case for #875.

snjeza:
> I don't (yet) have a test case for https://github.com/eclipse-jdt/eclipse.jdt.core/issues/875.

@stephan-herrmann You may want to take a look at https://github.com/eclipse-jdt/eclipse.jdt.core/issues/875#issuecomment-1538496370

stephan-herrmann:
> > I don't (yet) have a test case for #875.
> 
> @stephan-herrmann You may want to take a look at [#875 (comment)](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/875#issuecomment-1538496370)

Thanks, I saw it. But I saw no proof that it's really the same issue :smile:

stephan-herrmann:
With just a tiny improvement of my fix I could also resolve #986

snjeza:
> Thanks, I saw it. But I saw no proof that it's really the same issue

I can reproduce the issue described at https://github.com/eclipse-jdt/eclipse.jdt.core/issues/875#issue-1633370080 using the jdt.core master - https://github.com/eclipse/xtext/blob/main/org.eclipse.xtext.tests/src/org/eclipse/xtext/validation/AbstractCompositeValidatorTest.java#L140
Your PR fixes it.

stephan-herrmann:
> I can reproduce the issue described at [#875 (comment)](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/875#issue-1633370080) using the jdt.core master - https://github.com/eclipse/xtext/blob/main/org.eclipse.xtext.tests/src/org/eclipse/xtext/validation/AbstractCompositeValidatorTest.java#L140 Your PR fixes it.

Thanks for testing!

Looking at that test class, yes, the class, whose literal is used as an annotation argument will not have been connected yet, because it sits in the same compilation unit as the type being processed. It will be processed later based on source order.

Additionally, this declaration in `ComposedChecks` requires the subtype relationship which was broken due to #630:
`Class<? extends AbstractDeclarativeValidator>[] validators();`

With this I'm convinced it's a duplicate indeed, and forgo writing another test.

stephan-herrmann:
Now that tests passed:

@srikanth-sankaran may I get your review for inclusion in RC1. I promise I'll stop force-pushing more changes now :)

@iloveeclipse can I get your project lead's approval?

Thanks.

stephan-herrmann:
I'm in the middle of preparing an alternative, even safer PR. I hope to be able to complete it today.

stephan-herrmann:
While it was fairly straight-forward to get this approach to pass all tests, there's a known weakness, for which I don't know what symptoms it might cause, but from looking at the code this approach creates the following issue:

When `Annotation.resolveType()` is invoked from `ASTNode.resolveAnnotations()`, the work in `Annotation.resolveType()` itself can be safely split into two chunks performed at different points in time, but when deferring, the tail of `ASTNode.resolveAnnotations()` will work with incomplete information and I found no easy way to catch up in this regard after the deferred task has been performed.

For that reason I created an alternative, more drastic approach in PR #1077.

stephan-herrmann:
Resolved indeed via #1077

Issue #1096, opened 2023-05-26
ECJ out of sync with JLS 9.6.4.1


srikanth-sankaran:
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=552082

It appears that JDT went ahead and implemented the spec change https://bugs.openjdk.org/browse/JDK-8231435 eagerly.
It also appears that javac never implemented this change  as can be seen from this snippet

```
@interface Simple {}
class Test { <@Simple T> void m(T arg1) {} }
```

being rejected by javac 14-20 with the diagnostic

```
X.java:2: error: annotation @Simple not applicable in this type context
class Test { <@Simple T> void m(T arg1) {} }
              ^
1 error
```

Now JLS has changed course with https://bugs.openjdk.org/browse/JDK-8261610  and javac also is implementing https://bugs.openjdk.org/browse/JDK-8303784

Current reading of JLS 9.6.4.1 :
```
...
If an annotation of type java.lang.annotation.Target is not present on the
declaration of an annotation interface A, then A is applicable in all declaration
contexts and in no type contexts.
...
```

srikanth-sankaran:
See also https://bugs.eclipse.org/bugs/show_bug.cgi?id=568240

srikanth-sankaran:
See also https://bugs.eclipse.org/bugs/show_bug.cgi?id=566803

Issue #1092, opened 2023-05-25
Duplicate Annotation Error for Records


vladoiliev02:
The following code compiles successfully using the eclipse compiler and throws a ```java.lang.annotation.AnnotationFormatError: Duplicate annotation...``` exception on runtime. The same code compiled with a different compiler has no issues. This leads me to believe that this is an issue with the eclipse implementation of the java compiler. Using a decompiler on the '.class' file shows that really the ```@NotBlank``` annotation has been repeated inside the type parameter of the ```List``` post compilation.

```
@JsonInclude(JsonInclude.Include.NON_NULL)
public record Record (
        @Valid
        Record2 record2,

        @Valid
        Record3 record3,

        @Valid
        Record1 record1
) {
}
```

```
public record Record1(
        @NotEmpty(message = "Not empty Record1 strings")
        List<@NotBlank(message = "Not blank Record1 string") String> strings,

        @NotBlank(message = "Not blank Record1 str")
        String str
) {
}
```

```
public record Record2(
        @NotEmpty(message = "Not empty Record2 strings")
        List<@NotBlank(message = "Not blank Record2 string") String> strings,

        @NotBlank(message = "Not blank Record2 str")
        String str
) {
}
```

```
public record Record3(
        @NotEmpty(message = "Not empty Record3 strings")
        List<@NotBlank(message = "Not blank Record3 string") String> strings,

        @NotBlank(message = "Not blank Record3 str")
        String str
) {
}
```

```
@RestController
public class TestController {

    @PatchMapping("/request")
    public ResponseEntity<?> patch(@RequestBody @Valid Record dto) {
        return ResponseEntity.accepted().build();
    }
}
```

The problem occurs when calling the patch endpoint in the controller above. The exception occurs during validation of the dto. 

Exception:
```
java.lang.annotation.AnnotationFormatError: Duplicate annotation for class: interface javax.validation.constraints.NotBlank: @javax.validation.constraints.NotBlank(message="Not blank Record2 string", payload={}, groups={})
	at java.base/sun.reflect.annotation.TypeAnnotationParser.mapTypeAnnotations(TypeAnnotationParser.java:387) ~[na:na]
	at java.base/sun.reflect.annotation.AnnotatedTypeFactory$AnnotatedTypeBaseImpl.<init>(AnnotatedTypeFactory.java:152) ~[na:na]
[...]

PR #1097, opened 2023-05-26
ECJ out of sync with JLS 9.6.4.1

srikanth-sankaran:
## What it does
Reverse the changes made for https://bugs.eclipse.org/bugs/show_bug.cgi?id=552082 as a first step
To sync up with JLS  9.6.4.1 by a subsequent commit to fix https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1096

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Annotation.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Annotation.java
@@ -43,7 +43,6 @@ import java.util.Stack;
 import org.eclipse.jdt.core.compiler.CharOperation;
 import org.eclipse.jdt.internal.compiler.ASTVisitor;
 import org.eclipse.jdt.internal.compiler.classfmt.ClassFileConstants;
-import org.eclipse.jdt.internal.compiler.codegen.AnnotationTargetTypeConstants;
 import org.eclipse.jdt.internal.compiler.env.EnumConstantSignature;
 import org.eclipse.jdt.internal.compiler.impl.BooleanConstant;
 import org.eclipse.jdt.internal.compiler.impl.CompilerOptions;
@@ -769,7 +768,7 @@ public abstract class Annotation extends Expression {
 		return (metaTagBits & TagBits.AnnotationRetentionMASK) == TagBits.AnnotationClassRetention;
 	}
 
-	public boolean isRuntimeTypeInvisible(int targetType) {
+	public boolean isRuntimeTypeInvisible(boolean targetingTypeParameter) {
 		final TypeBinding annotationBinding = this.resolvedType;
 		if (annotationBinding == null) {
 			return false;
@@ -777,7 +776,7 @@ public abstract class Annotation extends Expression {
 		long metaTagBits = annotationBinding.getAnnotationTagBits(); // could be forward reference
 
 		if ((metaTagBits & (TagBits.AnnotationTargetMASK)) == 0) { // In the absence of explicit target, applicable only to declaration sites
-			if (targetType != AnnotationTargetTypeConstants.CLASS_TYPE_PARAMETER && targetType != AnnotationTargetTypeConstants.METHOD_TYPE_PARAMETER)
+			if (!targetingTypeParameter)
 				return false;
 		} else if ((metaTagBits & (TagBits.AnnotationForTypeParameter | TagBits.AnnotationForTypeUse)) == 0) {
 			return false;
@@ -789,7 +788,7 @@ public abstract class Annotation extends Expression {
 		return (metaTagBits & TagBits.AnnotationRetentionMASK) == TagBits.AnnotationClassRetention;
 	}
 
-	public boolean isRuntimeTypeVisible(int targetType) {
+	public boolean isRuntimeTypeVisible(boolean targetingTypeParameter) {
 		final TypeBinding annotationBinding = this.resolvedType;
 		if (annotationBinding == null) {
 			return false;
@@ -797,7 +796,7 @@ public abstract class Annotation extends Expression {
 		long metaTagBits = annotationBinding.getAnnotationTagBits();
 
 		if ((metaTagBits & (TagBits.AnnotationTargetMASK)) == 0) { // In the absence of explicit target, applicable only to declaration sites
-			if (targetType != AnnotationTargetTypeConstants.CLASS_TYPE_PARAMETER && targetType != AnnotationTargetTypeConstants.METHOD_TYPE_PARAMETER)
+			if (!targetingTypeParameter)
 				return false;
 		} else if ((metaTagBits & (TagBits.AnnotationForTypeParameter | TagBits.AnnotationForTypeUse)) == 0) {
 			return false;
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Annotation.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Annotation.java
@@ -775,11 +775,12 @@ public abstract class Annotation extends Expression {
 		}
 		long metaTagBits = annotationBinding.getAnnotationTagBits(); // could be forward reference
 
-		if ((metaTagBits & (TagBits.AnnotationTargetMASK)) != 0) {
-			if ((metaTagBits & (TagBits.AnnotationForTypeParameter | TagBits.AnnotationForTypeUse)) == 0) {
-				return false;
-			}
-		} // else: no-@Target always applicable
+		if ((metaTagBits & (TagBits.AnnotationTargetMASK)) == 0) { // explicit target required for JSR308 style annotations.
+			return false;
+		}
+		if ((metaTagBits & (TagBits.AnnotationForTypeParameter | TagBits.AnnotationForTypeUse)) == 0) {
+			return false;
+		}
 
 		if ((metaTagBits & TagBits.AnnotationRetentionMASK) == 0)
 			return true; // by default the retention is CLASS
@@ -794,11 +795,12 @@ public abstract class Annotation extends Expression {
 		}
 		long metaTagBits = annotationBinding.getAnnotationTagBits();
 
-		if ((metaTagBits & (TagBits.AnnotationTargetMASK)) != 0) {
-			if ((metaTagBits & (TagBits.AnnotationForTypeParameter | TagBits.AnnotationForTypeUse)) == 0) {
-				return false;
-			}
-		} // else: no-@Target always applicable
+		if ((metaTagBits & (TagBits.AnnotationTargetMASK)) == 0) { // explicit target required for JSR308 style annotations.
+			return false;
+		}
+		if ((metaTagBits & (TagBits.AnnotationForTypeParameter | TagBits.AnnotationForTypeUse)) == 0) {
+			return false;
+		}
 		if ((metaTagBits & TagBits.AnnotationRetentionMASK) == 0)
 			return false; // by default the retention is CLASS
 
@@ -1300,7 +1302,12 @@ public abstract class Annotation extends Expression {
 
 		long metaTagBits = annotationType.getAnnotationTagBits(); // could be forward reference
 		if ((metaTagBits & TagBits.AnnotationTargetMASK) == 0) {
-			// does not specify any target restriction - all locations are possible
+			// does not specify any target restriction - all locations supported in Java 7 and before are possible
+			// TBD - revisit for modules - as per 9.6.4.1, annotation without target is applicable for module declaration
+			// which is listed as a declaration context, but javac does not allow this
+			if (kind == Binding.TYPE_PARAMETER || kind == Binding.TYPE_USE) {
+				scope.problemReporter().explitAnnotationTargetRequired(annotation);
+			}
 			return AnnotationTargetAllowed.YES;
 		}
 
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Annotation.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Annotation.java
@@ -1165,7 +1165,7 @@ public abstract class Annotation extends Expression {
 	}
 
 	public enum AnnotationTargetAllowed {
-		YES, TYPE_ANNOTATION_ON_QUALIFIED_NAME, NO;
+		YES, NO_DUE_TO_LACKING_TARGET, TYPE_ANNOTATION_ON_QUALIFIED_NAME, NO/*_DUE_TO_MISMATCHED_TARGET*/;
 	}
 
 	private static AnnotationTargetAllowed isAnnotationTargetAllowed(Binding recipient, BlockScope scope, TypeBinding annotationType, int kind, long metaTagBits) {
@@ -1310,10 +1310,7 @@ public abstract class Annotation extends Expression {
 			   declaration of an annotation interface A, then A is applicable in all declaration
 			   contexts and in no type contexts.
 			*/
-			if (kind == Binding.TYPE_USE) {
-				scope.problemReporter().explitAnnotationTargetRequired(annotation);
-			}
-			return AnnotationTargetAllowed.YES;
+			return kind == Binding.TYPE_USE ?  AnnotationTargetAllowed.NO_DUE_TO_LACKING_TARGET : AnnotationTargetAllowed.YES;
 		}
 
 		// https://bugs.eclipse.org/bugs/show_bug.cgi?id=391201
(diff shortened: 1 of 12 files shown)


stephan-herrmann:
> @stephan-herrmann, this backs out your code and also advances it forward - May I request you to glance through the changes ? TIA.
> 
> See comments beginning with [#1092 (comment)](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1092#issuecomment-1563735170) for background although it turned out #1092 is completely orthogonal

I was a bit confused by the network of JDK bugs, and about how to figure out which solution replaces which previous solutions, so I collected the changes directly from JLS:
* JLS 8: "... T is applicable in all declaration contexts except type parameter declarations, and in no type contexts. "
* JLS 14: "... T is applicable in all nine declaration contexts and in all 16 type contexts."
* JLS 17+: "... A is applicable in all declaration contexts and in no type contexts."

Going from 8 to 17+ would have been easy. With the detour via 14 (which javac skipped?) more effort is needed, so from an outside look, backing out changes done for 14 sounds like the natural thing to do.

My next question would be: should ecj follow any of this back and forth based on compliance settings, or should we assume the 17+ variant is simply the best and should be supported for all compliance levels?

I'll take an inside look, too.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Annotation.java
```diff
@@ -768,37 +769,39 @@ public boolean isRuntimeInvisible() {
...
 		final TypeBinding annotationBinding = this.resolvedType;
 		if (annotationBinding == null) {
 			return false;
 		}
 		long metaTagBits = annotationBinding.getAnnotationTagBits(); // could be forward reference
 
-		if ((metaTagBits & (TagBits.AnnotationTargetMASK)) != 0) {
-			if ((metaTagBits & (TagBits.AnnotationForTypeParameter | TagBits.AnnotationForTypeUse)) == 0) {
+		if ((metaTagBits & (TagBits.AnnotationTargetMASK)) == 0) { // In the absence of explicit target, applicable only to declaration sites
+			if (targetType != AnnotationTargetTypeConstants.CLASS_TYPE_PARAMETER && targetType != AnnotationTargetTypeConstants.METHOD_TYPE_PARAMETER)
```
stephan-herrmann, line 780: In my naive understanding, methods `isRuntimeTypeVisible` and `isRuntimeTypeInvisible` would only cover codegen for **type annotations** (declaration scenarii being covered by `isRuntimeVisible`/`isRuntimeInvisible`).
If that were true then parameter `targetType` would be unnecessary, as in absence of a `@Target`  we can directly answer `false`, can't we?

The only test that disagrees is the new `testGH1096()`, challenging a no-target annotation on a type parameter.

Does this mean, annotations on type parameters are handled through the TYPE_USE-related code, rather than like "normal" declaration annotations?

Should we schedule a follow-up issue, to reconsider this?

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/RecordsRestrictedClassTest.java
```diff
@@ -3746,22 +3746,18 @@ public void testBug562439_019() throws IOException, ClassFormatException {
 			"    RuntimeInvisibleAnnotations: \n" +
 			"      #8 @Annot(\n" +
 			"      )\n" +
-			"    RuntimeInvisibleTypeAnnotations: \n" +
-			"      #8 @Annot(\n" +
-			"        target type = 0x13 FIELD\n" +
-			"      )\n" +
 			"  \n";
```
stephan-herrmann, line 3749: How do changes in this test suite relate to #1092? You wrote that one is orthogonal, yet here you are fixing duplication. Just curious.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/codegen/AnnotationTargetTypeConstants.java
```diff
@@ -19,6 +19,8 @@
 
 public interface AnnotationTargetTypeConstants {
 
+	int NonJVMS4_7_20_TargetLocation         = -1;
```
stephan-herrmann, line 22: I was a bit at loss, where in JVMS exactly I should look for an explanation...

stephan-herrmann (approved):
No objections, just some questions for further clarification, opportunities for (future) cleanup etc.

srikanth-sankaran:
> I was a bit confused by the network of JDK bugs, and about how to figure out which solution replaces which previous solutions, so I collected the changes directly from JLS:
> 
> * JLS 8: "... T is applicable in all declaration contexts except type parameter declarations, and in no type contexts. "
> * JLS 14: "... T is applicable in all nine declaration contexts and in all 16 type contexts."
> * JLS 17+: "... A is applicable in all declaration contexts and in no type contexts."
> 
> Going from 8 to 17+ would have been easy. With the detour via 14 (which javac skipped?) more effort is needed, so from an outside look, backing out changes done for 14 sounds like the natural thing to do.

Yes, javac never resolved https://bugs.openjdk.org/browse/JDK-8231436 which is still open that corresponds to the spec bug https://bugs.openjdk.org/browse/JDK-8231435 in JDK14 (which was superseded subsequently in JDK17)

> 
> My next question would be: should ecj follow any of this back and forth based on compliance settings, or should we assume the 17+ variant is simply the best and should be supported for all compliance levels?

IMO, it is best to skip supporting past behavior - it is not the case that a valid program ceases to be valid or changes meaning in some subtle manner - which would call for preserving old behavior with a compliance check.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Annotation.java
```diff
@@ -1300,7 +1306,13 @@ static AnnotationTargetAllowed isAnnotationTargetAllowed(Annotation annotation,
...
 		if ((metaTagBits & TagBits.AnnotationTargetMASK) == 0) {
-			// does not specify any target restriction - all locations are possible
+			/* JLS 9.6.4.1: If an annotation of type java.lang.annotation.Target is not present on the
+			   declaration of an annotation interface A, then A is applicable in all declaration
+			   contexts and in no type contexts.
+			*/
+			if (kind == Binding.TYPE_USE) {
+				scope.problemReporter().explitAnnotationTargetRequired(annotation);
+			}
 			return AnnotationTargetAllowed.YES;
```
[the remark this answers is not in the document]
srikanth-sankaran: > Answering YES after detecting an `explitAnnotationTargetRequired` error looks funny, but I can see that this has been inherited from 10 years ago. Would you see value in inventing an answer `AnnotationTargetAllowed.TYPE_USE_CONTEXT_LACKING_TARGET` and letting the two callers of `isAnnotationTargetAllowed` report the issue along with similar issues? Would this change whether or not we report follow-up errors (like when `checkAnnotationTarget` resets `tagBitsToRevert`)?

I opted for the name `AnnotationTargetAllowed.NO_DUE_TO_LACKING_TARGET` and implemented the change suggested above. 

It results in better error messages in the case of repeating annotations without explicit target being applied to TYPE_USE locations. See org.eclipse.jdt.core.tests.compiler.regression.RepeatableAnnotationTest.testRepeatingAnnotationsWithoutTarget()

Rather than complain 

```
----------\n
1. ERROR in X.java (at line 1)\n
	public class X<@Foo @Foo T> extends @Foo @Foo Object {\n
	                                    ^^^^\n
Annotation types that do not specify explicit target element types cannot be applied here\n
----------\n
2. ERROR in X.java (at line 1)\n
	public class X<@Foo @Foo T> extends @Foo @Foo Object {\n
	                                    ^^^^\n
Annotation types that do not specify explicit target element types cannot be applied here\n
----------\n
3. ERROR in X.java (at line 1)\n
	public class X<@Foo @Foo T> extends @Foo @Foo Object {\n
	                                         ^^^^\n
Annotation types that do not specify explicit target element types cannot be applied here\n
----------\n
```

we now complain:

```
----------\n
1. ERROR in X.java (at line 1)\n
	public class X<@Foo @Foo T> extends @Foo @Foo Object {\n
	                                    ^^^^\n
Annotation types that do not specify explicit target element types cannot be applied here\n
----------\n
2. ERROR in X.java (at line 1)\n
	public class X<@Foo @Foo T> extends @Foo @Foo Object {\n
	                                    ^^^^\n
The annotation @Foo cannot be repeated at this location since its container annotation type @FooContainer is disallowed at this location\n
----------\n
3. ERROR in X.java (at line 1)\n
	public class X<@Foo @Foo T> extends @Foo @Foo Object {\n
	                                         ^^^^\n
Annotation types that do not specify explicit target element types cannot be applied here\n
----------\n
```

As for `tagBitsToRevert` I think we should see what the effect is and take suitable follow up action. Returning Yes from there is bogus - yes we have done that for 10 years. I'll initiate a test run and see what happens

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Annotation.java
srikanth-sankaran: > In my naive understanding, methods `isRuntimeTypeVisible` and `isRuntimeTypeInvisible` would only cover codegen for **type annotations** (declaration scenarii being covered by `isRuntimeVisible`/`isRuntimeInvisible`). If that were true then parameter `targetType` would be unnecessary, as in absence of a `@Target` we can directly answer `false`, can't we?
> 

Consider this program:

```
import java.lang.annotation.Target;

@Target(ElementType.TYPE_USE)
@interface TA {
}

@TA
public class X {
} 
```

which reaches 

```
SingleMemberAnnotation(Annotation).isRuntimeTypeInvisible(int) line: 777	
ClassFile.generateRuntimeAnnotations(Annotation[], long) line: 4581	
ClassFile.addAttributes() line: 423	
TypeDeclaration.generateCode(ClassFile) line: 772	
TypeDeclaration.generateCode(CompilationUnitScope) line: 831	
```

and org.eclipse.jdt.core.tests.compiler.regression.TypeAnnotationTest.testGH1096()

which reaches the call stack:

```
MarkerAnnotation(Annotation).isRuntimeTypeInvisible(int) line: 777	
TypeReference$AnnotationCollector.internalVisit(Annotation) line: 197	
TypeReference$AnnotationCollector.visit(MarkerAnnotation, BlockScope) line: 244	
MarkerAnnotation.traverse(ASTVisitor, BlockScope) line: 43	
TypeParameter.getAllAnnotationContexts(int, int, List<AnnotationContext>) line: 72	
ClassFile.completeMethodInfo(MethodBinding, int, int) line: 2455	
MethodDeclaration(AbstractMethodDeclaration).generateCode(ClassFile) line: 377	
MethodDeclaration(AbstractMethodDeclaration).generateCode(ClassScope, ClassFile) line: 292	
TypeDeclaration.generateCode(ClassFile) line: 761	
```
In both scenarios the method is supposed to return true.

What is a suitable value for the parameter targetType of the method
org.eclipse.jdt.internal.compiler.ast.Annotation.isRuntimeTypeInvisible(int) for the first call (stack) ??

(Despite it not being relevant - the parameter gets "touched" only when no target is specified) I simply invented a value that is outside of all possible type annotation target locations enumerated by table 4.7.20 of the JVMS. 

There are only three meaningful values that influence the method's behavior:
targetType == AnnotationTargetTypeConstants.CLASS_TYPE_PARAMETER
targetType == AnnotationTargetTypeConstants.METHOD_TYPE_PARAMETER and
targetType being *anything else* - For that -1  given the symbolic name NonJVMS4_7_20_TargetLocation was my choice.

Better ideas welcome. 


> The only test that disagrees is the new `testGH1096()`, challenging a no-target annotation on a type parameter.
> 
> Does this mean, annotations on type parameters are handled through the TYPE_USE-related code, rather than like "normal" declaration annotations?
> 
> Should we schedule a follow-up issue, to reconsider this?

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/RecordsRestrictedClassTest.java
srikanth-sankaran: > How do changes in this test suite relate to #1092? You wrote that one is orthogonal, yet here you are fixing duplication. Just curious.

This is reversing the effect of https://bugs.eclipse.org/bugs/show_bug.cgi?id=552082 - In the absence of explicit target @Annot is both a type annotation and a SE7 declaration annotation and gets emitted once each as a declarative annotation and as a type annotation.

In https://bugs.eclipse.org/bugs/show_bug.cgi?id=552082#c1, you asserted 
```
This change made me look also at this in existing JLS §9.7.4:
   "If the annotation's type is applicable in the declaration context corresponding to the declaration and in type contexts, then the annotation is deemed to apply to both the declaration and the type which is closest to the annotation."

I don't see this dual applicability represented in the class files generated by javac nor ecj. Just mentioning as am NOT going to fix this, i.e., I will not invest efforts to become incompatible with javac in this case.
```

That is not so! We do implement the dual applicability and that needs to adapt to the latest change since in the absence of a target @Annot is not a type annotation *anymore*

The duplication we are talking about in https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1092 is indeed orthogonal brought about by (incorrect) double traversal of the relevant AST nodes.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/RepeatableAnnotationTest.java
```diff
@@ -1472,12 +1464,15 @@ public void testDeprecation2() { // verify that deprecation warning does not sho
 				"@interface T {\n" +
 				"  public int value() default -1;\n" +
 				"}\n" +
-				"interface I<@T(1) @T(2) K> {\n" +
+				"interface I extends @T(1) Runnable {\n" +
```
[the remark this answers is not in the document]
srikanth-sankaran: Just to retain it as a negative test. Otherwise the code would compile fine with the new behavior implemented

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/codegen/AnnotationTargetTypeConstants.java
srikanth-sankaran: > I was a bit at loss, where in JVMS exactly I should look for an explanation...

Per above, I need a sentinel value for the targetType parameter of org.eclipse.jdt.internal.compiler.ast.Annotation.isRuntimeTypeInvisible(int) method that is outside the legal range of values for type annotation target locations enumerated by JVMS Table 4.7.20  - 

Better suggestions welcome

stephan-herrmann: Rather than a lot of hand waving (where I couldn't be sure whether or not I was talking bullshit), I experimented with disentangling how we handle type annotations and declaration annotations as much as possible. Please find my result in PR #1117 - which is created on top of this PR but pushed separately as not to interfere with your work here.

The key connection to this constant is: the invocations where the new constant was used could be _removed_ altogether :)

WDYT?

stephan-herrmann:
> > My next question would be: should ecj follow any of this back and forth based on compliance settings, or should we assume the 17+ variant is simply the best and should be supported for all compliance levels?
> 
> IMO, it is best to skip supporting past behavior - it is not the case that a valid program ceases to be valid or changes meaning in some subtle manner - which would call for preserving old behavior with a compliance check.

sounds good to me.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/codegen/AnnotationTargetTypeConstants.java
srikanth-sankaran: > I was a bit at loss, where in JVMS exactly I should look for an explanation...

I have in the latest additional push, gotten rid of this head-scratch producing enumeration.

srikanth-sankaran: > Rather than a lot of hand waving (where I couldn't be sure whether or not I was talking bullshit), I experimented with disentangling how we handle type annotations and declaration annotations as much as possible. Please find my result in PR #1117 - which is created on top of this PR but pushed separately as not to interfere with your work here.
> 
> The key connection to this constant is: the invocations where the new constant was used could be _removed_ altogether :)
> 
> WDYT?

I will review this - I suggest let us do this in a follow up ticket - (I'll raise one)

PR #1117, opened 2023-06-01
ECJ out of sync with JLS 9.6.4.1

stephan-herrmann:
This PR adds some experimental clean-up on top of PR #1097:

As per the commit message:

* reduce mixed handling of type/declaration annotations
  * when generating decl annos don't care about visibility as type anno
* remove redundant filtering of decl annos
* remove interim `NonJVMS4_7_20_TargetLocation`
* replace `SE7AnnotationTargetMASK` with `AnnotationForDeclarationMASK`
* clarify that `AnnotationCollector` does not traverse type parameter
* a few clarifying code comments
* update two tests in `RepeatableAnnotationTest`:
  * `test026` extend the error message
  * `testUnspecifiedTarget2` don't expect conflict, `@T` applies to all decls

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ClassFile.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ClassFile.java
@@ -419,7 +419,7 @@ public class ClassFile implements TypeConstants, TypeIds {
 					else if (this.referenceBinding.isAnnotationType())
 						targetMask = TagBits.AnnotationForType | TagBits.AnnotationForAnnotationType;
 					else
-						targetMask = TagBits.AnnotationForType | TagBits.AnnotationForTypeUse;
+						targetMask = TagBits.AnnotationForType | TagBits.AnnotationForTypeUse; // 9.7.4 ... applicable to type declarations or in type contexts
 					attributesNumber += generateRuntimeAnnotations(annotations, targetMask);
 				}
 			}
@@ -2431,6 +2431,7 @@ public class ClassFile implements TypeConstants, TypeIds {
 				}
 				Annotation[] annotations = methodDeclaration.annotations;
 				if (annotations != null && !methodDeclaration.isClinit() && (methodDeclaration.isConstructor() || binding.returnType.id != T_void)) {
+					// at source level type annotations have not been moved from declaration to type use position, collect them now:
 					methodDeclaration.getAllAnnotationContexts(AnnotationTargetTypeConstants.METHOD_RETURN, allTypeAnnotationContexts);
 				}
 				if (!methodDeclaration.isConstructor() && !methodDeclaration.isClinit() && binding.returnType.id != T_void) {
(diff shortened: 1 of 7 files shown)


stephan-herrmann:
Let me try to disambiguate different kinds of ambiguity:

* The JLS change clarifying the semantics of no-target annotations actually _removes_ one ambiguity for those annotations: they can no longer swing both ways  as type or declaration annotation :+1: 
* Of course, annotations that declare target `TYPE_USE` and one or more declaration location remain ambiguous in a few syntactic locations:
  * Annotations on fields or methods can still apply to both the declaration and its (return) type.
  * Even annotations syntactically placed on type _declarations_, can be type annotations, when they have  target `TYPE_USE` (_short hand for future interpretation as "all uses of this type should be seen as having this type annotation"_).
* Looking from such syntactically ambiguous locations, a non-ambiguous target will resolve the ambiguity, that's why code gen needs to filter annotations based on their target `tagBits`.
* Technically, at the bytecode level type annotations (runtime visible or invisible) comprise different things:
  * true type annotations (`TYPE_USE`)
  * annotations on type parameters (which in JLS are specified as declaration annotations, but are encoded as type annotations in byte code - go figure).
 
In the compiler, method `ASTNode.copySE8AnnotationsToType()` moves some annotations from the syntactic location on a declaration to their semantic location on a type. This, however, doesn't help disentangling the same for code generation, as it doesn't move the source `Annotation` but only manipulates `AnnotationBinding`s, which is not sufficient for code generation.

srikanth-sankaran:
@stephan-herrmann I have brought the branch up to date with origin/master. Can you take a quick look to confirm this is the set of all changes you had in mind ? I will review it early next week and we can take it forward.

I assume there is a further cleanup in removing the boolean parameter of org.eclipse.jdt.internal.compiler.ast.Annotation.isRuntimeTypeInvisible(boolean)

I'll add a commit for that

srikanth-sankaran:
> I assume there is a further cleanup in removing the boolean parameter of org.eclipse.jdt.internal.compiler.ast.Annotation.isRuntimeTypeInvisible(boolean)
> 

Or should it stay ?? Need to check

stephan-herrmann:
> I have brought the branch up to date with origin/master. Can you take a quick look to confirm this is the set of all changes you had in mind ?

It took me a little while to sift out what difference resulted from rebasing and force-pushing, vs. comparing the payload of previous & current branch. I mentioned elsewhere that I'm not a fan of force push as soon as more than one person is working on a PR. May I suggest for future collaboration that we try to separate merging or rebasing from any further changes?

But, yes, I can find all my changes plus your simplification from `int targetType` to `boolean targetingTypeParameter` looking good :)

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Annotation.java
```diff
@@ -1313,7 +1299,7 @@ static AnnotationTargetAllowed isAnnotationTargetAllowed(Annotation annotation,
 		}
 
 		// https://bugs.eclipse.org/bugs/show_bug.cgi?id=391201
-		if ((metaTagBits & TagBits.SE7AnnotationTargetMASK) == 0
+		if ((metaTagBits & TagBits.AnnotationForDeclarationMASK) == 0
 				&& (metaTagBits & (TagBits.AnnotationForTypeUse | TagBits.AnnotationForTypeParameter)) != 0) {
 			if (scope.compilerOptions().sourceLevel < ClassFileConstants.JDK1_8) {
```
srikanth-sankaran, line 1304: Should this be:
```
if ((metaTagBits & TagBits.AnnotationForDeclarationMASK) == 0
				&& (metaTagBits & TagBits.AnnotationForTypeUse) != 0) {
```

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ClassFile.java
```diff
@@ -4578,9 +4579,9 @@ private int generateRuntimeAnnotations(final Annotation[] annotations, final lon
 			if (annotationMask != 0 && (annotationMask & targetMask) == 0) {
 				if (!jdk16packageInfoAnnotation(annotationMask, targetMask)) continue;
 			}
```
srikanth-sankaran, line 4581: Is this jdk16packageInfoAnnotation stuff still relevant ??

srikanth-sankaran (approved):
Overall looks good, some minor comments/questions here and there.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ClassFile.java
stephan-herrmann: very good question :)  I had no idea about this, but it leads us back to https://bugs.eclipse.org/bugs/show_bug.cgi?id=449330 where Shankha wrote:

> If a annotation without the annotation target being set to PACKAGE is used on package-info.java we should allow that in 1.6 and below and generate the annotation in the class file based on the runtime policy. 

So, apart from the question whether supporting 1.6 is still relevant (unfortunately I lost any reference to where we discussed what can/should be dropped - it was mentioned in the not so distant past), perhaps the new JLS rule implemented here supersedes that special casing for 1.6-. 

But no, `jdk16packageInfoAnnotation` seems to admit annotations on package-info.java if they do have a `@Target` meta annotation but no mention of `PACKAGE` in it. That's different from what we find in current JLS.

Although the compiler regression suite run at -Dcompliance=1.6 does not need this code, I'm reluctant to remove it. Feel free to file a follow-up for clean-up.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Annotation.java
stephan-herrmann: Good catch, I only modified one line without looking into the next ... 

Still I don't think it's possible to construct a case where it would make a difference, as we'd need to provide "no declaration target but target type parameter", which however is a complicated way of saying "false" :)

I'll push a fix, just to avoid future puzzlement.

Issue #1112, opened 2023-05-30
jakarta.inject.Inject not treated properly with annotation-based null analysis


sithmein:
Fields annotated with `@javax.inject.Inject` are treated specially by the annotation-based null analysis in that they don't raise an error when they are not explicitly initialized (because the value gets injected during object construction). This does not work if fields are annotated with the new `@jakarta.inject.Inject` annotation.
This makes it impossible to use annotation-based null analysis together with recent Jakarta EE frameworks such as Quarkus 3.

PR #1121, opened 2023-06-02
Issue eclipse-jdt#1112 - jakarta.inject.Inject is now recognized

sithmein:
...during annotation-based null analysis.

## What it does
Fixed issue #1112 

## How to test
The PR contains a testcase.

## Author checklist


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
```diff
@@ -513,10 +513,11 @@ else if (CharOperation.equals(TypeConstants.JDK_INTERNAL_PREVIEW_FEATURE, this.c
 					if (!CharOperation.equals(TypeConstants.JAVA, packageName))
 						return;
 					break; // continue below ...
-				case 5:
+				case 5: // javax
 					switch (packageName[1]) {
 						case 'a':
-							if (CharOperation.equals(TypeConstants.JAVAX_ANNOTATION_INJECT_INJECT, this.compoundName))
+							if (CharOperation.equals(TypeConstants.JAVAX_ANNOTATION_INJECT_INJECT, this.compoundName)
+									|| CharOperation.equals(TypeConstants.JAKARTA_ANNOTATION_INJECT_INJECT, this.compoundName))
```
stephan-herrmann, line 520: the second part can never become true when `packageName.length == 5`, can it?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
```diff
@@ -525,6 +526,14 @@ else if (CharOperation.equals(TypeConstants.JDK_INTERNAL_PREVIEW_FEATURE, this.c
 							return;
 					}
 					return;
+				case 7: // jakarta
+					switch (packageName[1]) {
+						case 'a':
+							if (CharOperation.equals(TypeConstants.JAKARTA_ANNOTATION_INJECT_INJECT, this.compoundName))
+								this.id = TypeIds.T_JavaxInjectInject;
```
stephan-herrmann, line 533: Can you confirm that `javax.annotation.inject.Inject` and `jakarta.annotation.inject.Inject` are indeed *identical* from a semantic p.o.v.? Otherwise, a new constant in `TypeIds` might be better.

stephan-herrmann (changes requested):
Thanks for your PR.

Here is the main question I keep asking when hard coding special semantics of library types: How sure can we be that the _intended_ semantic rule is actually _enforced_? Specifically, does `jakarta.annotation.inject.Inject` support some notion of optionality, or will absence of a value to inject invariably raise an exception?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
sithmein: `jakarta.inject.Inject` is a 1:1 replacement of `javax.inject.Inject`. Therefore the semantics are identical.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
sithmein: That's actually true.

sithmein:
While `com.google.inject.Inject` has the concept of optionality, `jakarta.inject.Inject` does not. You will get an exception if no injectable object is available.

stephan-herrmann (approved):
Thanks for clarification and code polish.

Issue #1126, opened 2023-06-06
[17] compiler.ast.CaseStatement.ResolvedCase.intValue not initialized for Strings


jukzi:
@jarthana 
The return value of ` c.stringValue().hashCode();` is not used. Was it meant to be assigned to this.intValue ?

https://github.com/eclipse-jdt/eclipse.jdt.core/blame/28cc628fdc35aef58a2321cf603f7ccc63e0ee56/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/CaseStatement.java#LL207C3-L207C31

jarthana:
At a glance, your observation looks to be correct. Let me take a closer look soon.

jarthana:
OK, I see that I meant to capture that intValue but didn't. The reason it works is because, down the line, in SwitchStatement (line 981), we do a String comparison anyway instead of intValue().

					`return c2.stringValue().equals(con.stringValue());`

I think it makes sense to avoid the String comparison and compare the intValue() and if we do, then tests will fail.

And good catch. I will shortly propose a fix.

iloveeclipse:
> And good catch. I will shortly propose a fix.

Jay, I see a commit but don't see a PR. Missed to create one?

jarthana:
> > And good catch. I will shortly propose a fix.
> 
> Jay, I see a commit but don't see a PR. Missed to create one?

It's already here - https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1128

iloveeclipse:
> It's already here - #1128

Jay, for github automation / better traceability it is needed to refer to the issue in the commit message with the keyword `Fixes` following by the issue hash or (better) full issue url, like:
```
[17] compiler.ast.CaseStatement.ResolvedCase.intValue not initialized for Strings


Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1126
```

This way 
- the PR will be linked to issue automatically
- if PR is merged, issue will be automatically fixed
- From the Git history in IDE you can directly jump to the issue

jarthana:
> > It's already here - #1128
> 
> Jay, for github automation / better traceability it is needed to refer to the issue in the commit message with the keyword `Fixes` following by the issue hash or (better) full issue url, like:
> 
> ```
> [17] compiler.ast.CaseStatement.ResolvedCase.intValue not initialized for Strings
> 
> 
[...]

PR #1128, opened 2023-06-07
[17] compiler.ast.CaseStatement.ResolvedCase.intValue not initialized for Strings #1126

jarthana:
## What it does
Fixed https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1126

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/SwitchStatement.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/SwitchStatement.java
@@ -978,7 +978,7 @@ public class SwitchStatement extends Expression {
 								IntPredicate check = (idx) -> {
 									Constant c2 = this.otherConstants[idx].c;
 									if (con.typeID() == TypeIds.T_JavaLangString) {
-										return c.intValue() == this.otherConstants[idx].intValue();
+										return c2.stringValue().equals(con.stringValue());
 									} else {
 										if (c2.typeID() == TypeIds.T_JavaLangString)
 											return false;
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/SwitchStatement.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/SwitchStatement.java
@@ -978,7 +978,7 @@ public class SwitchStatement extends Expression {
 								IntPredicate check = (idx) -> {
 									Constant c2 = this.otherConstants[idx].c;
 									if (con.typeID() == TypeIds.T_JavaLangString) {
-										return c2.stringValue().equals(con.stringValue());
+										return c.intValue() == this.otherConstants[idx].intValue();
 									} else {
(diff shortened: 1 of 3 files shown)


org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/SwitchPatternTest.java
```diff
@@ -6182,4 +6182,25 @@ public void testIssue919() {
...
 			"----------\n");
 	}
+	public void testIssue1126() {
+		runConformTest(
+				new String[] {
+					"X.java",
+					"public class X {\n"+
+					" 	static int foo(String i) {\n"+
+					"		return switch (i) {\n"+
+					"	     case \"abc\" -> 0;\n"+
```
jukzi, line 6192:
```suggestion
					"	     case \"FB\" -> 0;\n"+
```

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/SwitchPatternTest.java
```diff
@@ -6182,4 +6182,25 @@ public void testIssue919() {
...
+					"	     case \"abcd\" -> 1;\n"+
```
jukzi, line 6193:
```suggestion
					"	     case \"Ea\" -> 1;\n"+
```

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/SwitchStatement.java
```diff
@@ -978,7 +978,7 @@ public void resolve(BlockScope upperScope) {
 								IntPredicate check = (idx) -> {
 									Constant c2 = this.otherConstants[idx].c;
 									if (con.typeID() == TypeIds.T_JavaLangString) {
-										return c2.stringValue().equals(con.stringValue());
+										return c.intValue() == this.otherConstants[idx].intValue();
```
jukzi, line 981: That shows an Error if two Strings have the same hashCode. For exampe "FB" and "Ea" gives "Duplicate case".

jukzi (changes requested):

jarthana:
Yes, we have some failures due to the same reason. Will look into this.

jarthana:
So, it's not useful to do the int comparison. I thought of doing some optimization of falling back to String comparison when the int comparison returns true.  But I don't see much value as the Strings are probably compared for reference and interned as well. So, have to live with the fact that the intValue() is not much useful for String constants in case.

jukzi:
But did you try the test with hash-conflicting strings? it did still fail with revert for me.

jarthana:
> But did you try the test with hash-conflicting strings? it did still fail with revert for me.

Yes, now with us doing a simple String comparison, works fine. Will capture this in a new test case for the record.

jarthana:
> But did you try the test with hash-conflicting strings? it did still fail with revert for me.

@jukzi can you confirm that this no longer an issue?

jukzi (approved):
SwitchPatternTest now passes locally for me
image

PR #1134, opened 2023-06-08
Fix spurious dead code warning after catching NPE (561565)

btj:
See https://bugs.eclipse.org/bugs/show_bug.cgi?id=561565 .

## Author checklist


jukzi:
"[Bug 561565](https://bugs.eclipse.org/bugs/show_bug.cgi?id=561565) - Non-null analysis incorrectly reports dead code; ignores catching NullPointerException" @stephan-herrmann looks like this is something you could review
@btj is it possible that you add a junit test for it?

stephan-herrmann:
Hi @btj long time no see! :smile:

stephan-herrmann:
@btj can you tell me, which test case this is supposed to fix? When I create a unit test from https://bugs.eclipse.org/bugs/show_bug.cgi?id=561565#c0 I get the following unexpected problems:
```
----------\n
1. ERROR in Foo.java (at line 9)\n
	if (foo == null)\n
	    ^^^\n
Null comparison always yields false: The variable foo cannot be null at this location\n
----------\n
2. WARNING in Foo.java (at line 10)\n
	throw new IllegalArgumentException();\n
	^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n
Dead code\n
----------\n
```

Perhaps some other change in JDT broke your fix, or did you look at a different example?

stephan-herrmann:
To ensure we're on the same page, this is the test I added to NullReferenceTest:
```
public void testGH1134() {
	runConformTest(
		new String[] {
			"Foo.java",
			"""
			class Foo {

				int x;

				static void foo(Foo foo) {
					try {
						if (foo.x == 0);
					} catch (Throwable t) {}
					if (foo == null)
						throw new IllegalArgumentException();
				}

			}
			"""
		});
```

btj:
Hmm. Yes, this test case was supposed to be fixed by this PR. I guess something else broke it again. Never mind, then! I guess I should close the PR.

jukzi:
> Hmm. Yes, this test case was supposed to be fixed by this PR. I guess something else broke it again. Never mind, then! I guess I should close the PR.

or better add the test to the PR and fix it :-)

PR #1139, opened 2023-06-12
Bug 527899 [9] Implement JEP 280: Indify String Concatenation

jarthana:
## What it does
Implements https://openjdk.org/jeps/280 for ECJ

## Author checklist


jarthana:
There are still known issues that should be addressed before merging this PR.

- [ ] Couple of tests `inXLargeTest `failing with new/different error "The type generates a string that requires more than 65535 bytes to encode in Utf8 format in the constant pool"
- [ ] The storage mechanism used for the new bootstrap entry (using `Expression `as a key) and storing the `concatenatedLiteral `in `Expression `need to be revisited
- [ ] Need to polish up `ClassFile#addBootStrapStringConcatEntry()` a bit

jarthana:
Attention JDT committers, leads and experts (@mpalat  @iloveeclipse @stephan-herrmann @srikanth-sankaran ) (others are welcome to chime in too.
This feature meant for Java 9 but we chose not to do this at that time. But with String Templates on the horizon, now is a good time to implement this.

This patch implements the feature for all code >= Java 9. But this approach poses a major problem.
The problem is, all Eclipse platform projects now have already moved to Java 9+ and will see a different binaries. Basically, we are are going to be flooded by comparator errors and will be forced to touch all the bundles.

If we want to avoid this, we can limit this to compliance 20+ or even 21+. 

What does everyone think? I wanted to discuss this here before taking to the wider forum, like the jdt mailing list.

iloveeclipse:
> Basically, we are are going to be flooded by comparator errors and will be forced to touch all the bundles.

OK for me, especially in M1 time frame, I can help touching bundles.

> If we want to avoid this, we can limit this to compliance 20+ or even 21+.

I would not do that, simply to get higher test coverage (by "eating our own dog food"). *Assuming* the generated bytecode is "faster" executed by the modern JVM, why should ve let users wait for another LTS update to get it into their applications built with ECJ?

Where I have more concerns is the resulting binaries performance - considering that this affects almost every compiled class, that should be carefully tested / verified. Have you tried to run some performance tests (there were some hints in https://openjdk.org/jeps/280)? Can you outline chosen strategy / possible differences to the javac implementation?

jarthana:
> > Basically, we are are going to be flooded by comparator errors and will be forced to touch all the bundles.
> 
> OK for me, especially in M1 time frame, I can help touching bundles.

Thanks!

> I would not do that, simply to get higher test coverage (by "eating our own dog food"). _Assuming_ the generated bytecode is "faster" executed by the modern JVM, why should ve let users wait for another LTS update to get it into their applications built with ECJ?

Two additional points. ECJ is not obligated to implement this. This is not particularly relavent to the quoted point, but just wanted to say that. 
The second point is, if we make this as a new feature as part of our Java 21 support, then nobody will see their .class files change, except when they move to Java 21. That will give us a nice cover to avoid this bytecode change.

> Where I have more concerns is the resulting binaries performance - considering that this affects almost every compiled class, that should be carefully tested / verified. Have you tried to run some performance tests (there were some hints in https://openjdk.org/jeps/280)? Can you outline chosen strategy / possible differences to the javac implementation?

I ran into some failures in XLargeTest and had a chance to study some of the tests involving large string concat and didn't find anything, particularly in terms of performance. Let me run the entire JDT tests with and without the fix and see if that brings out something. 

Comparing with Javac:

- As far as produced output is concerned, except a corner case (where we avoid pushing null into operand stack), we pretty much produce what Javac does. 
- And because we let the JVM do the concatenation for us, now some of the code that ran before won't run because of the breach of the 65535 size for a String constant. Earlier, we avoided this by invoking a series of StringBuilder.append().

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/XLargeTest.java
```diff
@@ -3773,7 +3785,7 @@ public void test012() {
 			"public class X {\n" +
 			"  void foo(String a, String b, String c, String d, String e) {\n" +
 			"    String s = a + (\n");
-	for (int i = 0; i < 1000; i++) {
+	for (int i = 0; i < 500; i++) {
```
szarnekow, line 3788: Does this test change indicate that the new compile strategy fails with shorter strings?

mpalat:
> Attention JDT committers, leads and experts (@mpalat @iloveeclipse @stephan-herrmann @srikanth-sankaran ) (others are welcome to chime in too. This feature meant for Java 9 but we chose not to do this at that time. But with String Templates on the horizon, now is a good time to implement this.
> 
I strongly agree that now is a good time to implement this since String templates will preview in 21. However, instead of clubbing with Java 21, this should be implemented in master as a separate feature since it is a separate feature for one, and second, this is not a preview feature and should be available for every one. 4.29 master would be the way to go - M1 being ideal.
> 
> If we want to avoid this, we can limit this to compliance 20+ or even 21+.

Continuing from above, no - we should not be restricting this since its already available the minimum required jdk level as of now.

srikanth-sankaran:
I don't have much to add to what @iloveeclipse and @mpalat have said - substantially agree with their observations.

I will add myself as a reviewer.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/XLargeTest.java
jarthana: Yes, but it's not all bad news. If that does happen, then that probably means the generated method size will be smaller the current error we throw in that scenario will be avoided.

srikanth-sankaran:
Looks like the last commit [Handle cases of a String literal containing '\u0001' or '\u0002'](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1139/commits/eab96b9a6aef615d1aca8448b1c919ff79f7e9f5) touching several files can be greatly simplified with with a single line change in org.eclipse.jdt.internal.compiler.ast.Expression.buildStringForConcatation(BlockScope, CodeStream, int, StringBuilder, List<TypeBinding>)

changing

```
if ((this.bits & ASTNode.HasStringConcatMarkers) != 0) {
```
with
```
if (this.constant.stringValue().indexOf('\u0001') != -1 || this.constant.stringValue().indexOf('\u0002') != -1)
```

srikanth-sankaran:
> Looks like the last commit [Handle cases of a String literal containing '\u0001' or '\u0002'](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1139/commits/eab96b9a6aef615d1aca8448b1c919ff79f7e9f5) touching several files can be greatly simplified with with a single line change in org.eclipse.jdt.internal.compiler.ast.Expression.buildStringForConcatation(BlockScope, CodeStream, int, StringBuilder, List)
> 
> changing
> 
> ```
> if ((this.bits & ASTNode.HasStringConcatMarkers) != 0) {
> ```
> 
> with
> 
> ```
> if (this.constant.stringValue().indexOf('\u0001') != -1 || this.constant.stringValue().indexOf('\u0002') != -1)
> ```

If you agree, cleaning that up plus rebasing on top of all the recent changes will help simplify and speed up the review. Thanks!

jarthana:
OK, I usually don't do a force-push, but since the last one is only a partial reversal of previous change, I rewrote that previous commit. 

Anyway, I was going to argue that the String comparison (indexOf) might cause some performance impact, especially with sources like our tests that include lot of string concatenation. I did run some performance tests using our performance suite. Although I see some difference, they are not convincing enough to complicate the code. So, went with the proposed change.

srikanth-sankaran:
> 

Aside from the standard observation about premature optimization and humans being poor judges of where time is likely spent, string concatenation in the style of our tests would not be a problem anyway since they would be constant folded at compile time. 

Anyway it is good that you studied it to make sure there is no discernible difference

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Expression.java
```diff
@@ -68,7 +69,7 @@
 public abstract class Expression extends Statement {
 
 	public Constant constant;
-
+	public StringBuilder concatenatedLiteral;
```
szarnekow, line 72: Could this already be the String that is to be made from the StringBuilder? I cannot really spot code that would append to the StringBuilder after it was set as a field here.

szarnekow (approved):

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Expression.java
```diff
@@ -931,7 +932,45 @@ public void generateOptimizedStringConcatenationCreation(BlockScope blockScope,
...
+			case TypeIds.T_boolean :
+				generateCode(blockScope, codeStream, true);
+				argTypes.add(this.resolvedType);
+				builder.append(STRING_CONCAT_MARKER_1);
+				break;
+			default :
+				if (this.resolvedType.id == TypeIds.T_null) {
+					// Optimize it, avoid aconst_null, simply append the String Literal
+					builder.append(new String(NullLiteral.source));
+				} else {
```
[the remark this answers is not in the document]
jarthana: Not without a cast.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Expression.java
jarthana: I am getting rid of this field altogether as @srikanth-sankaran suggested.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ClassFile.java
```diff
@@ -3657,6 +3658,8 @@ private int generateBootstrapMethods(List<ASTNode> bootStrapMethodsList) {
 				} else {
 					localContentsOffset = addBootStrapTypeSwitchEntry(localContentsOffset, stmt, fPtr);
 				}
+			} else if (o instanceof Expression) {
+				localContentsOffset = addBootStrapStringConcatEntry(localContentsOffset, (Expression) o, fPtr);
 			}
```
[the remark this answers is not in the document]
jarthana: Great suggestion, thanks! This was one of my open questions at the start of the discussion. Will mark that done as well.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeConstants.java
```diff
@@ -117,6 +117,8 @@ public interface TypeConstants {
     char[] UPPER_MODULE = "MODULE".toCharArray(); //$NON-NLS-1$
     char[] UPPER_RECORD_COMPONENT = "RECORD_COMPONENT".toCharArray(); //$NON-NLS-1$
     char[] YIELD = "yield".toCharArray(); //$NON-NLS-1$
+    // Duplicated since java.lang.invoke.StringConcatFactory.TAG_ARG isn't public
+    char[] STRING_CONCAT_MARKER_1 = new char[] {'\u0001'};
```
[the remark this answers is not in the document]
jarthana: I don't see anything changing in terms of behavior. But clearly I see that the String argument we pass is likely to surpass the 65K limit more often.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
```diff
@@ -2933,6 +2933,11 @@ public final ReferenceBinding getJavaLangRuntimeSwitchBootstraps() {
 		unitScope.recordQualifiedReference(TypeConstants.JAVA_LANG_RUNTIME_SWITCHBOOTSTRAPS);
 		return unitScope.environment.getResolvedJavaBaseType(TypeConstants.JAVA_LANG_RUNTIME_SWITCHBOOTSTRAPS, this);
 	}
+	public final ReferenceBinding getJavaLangInvokeStringConcatFactory() {
+		CompilationUnitScope unitScope = compilationUnitScope();
+		unitScope.recordQualifiedReference(TypeConstants.JAVA_LANG_INVOKE_STRING_CONCAT_FACTORY);
+		return unitScope.environment.getResolvedJavaBaseType(TypeConstants.JAVA_LANG_INVOKE_STRING_CONCAT_FACTORY, this);
+	}
```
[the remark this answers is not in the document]
jarthana: Raised #https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1164 for follow up.

jarthana:
I resolved most of the points raised. However, I did some more testing with String concatenations with lot of operands and special marker characters. It reveals that the code we produce is lot more naive and very soon breach the 255 method arguments mark. OTOH, Javac is smarter and splits the operation into multiple ones. We should do that too.

srikanth-sankaran:
> I resolved most of the points raised. However, I did some more testing with String concatenations with lot of operands and special marker characters. It reveals that the code we produce is lot more naive and very soon breach the 255 method arguments mark. OTOH, Javac is smarter and splits the operation into multiple ones. We should do that too.
> 

I would suggest we keep this PR for the basic implementation and follow up on a separate ticket for improvements

jarthana:
> > I resolved most of the points raised. However, I did some more testing with String concatenations with lot of operands and special marker characters. It reveals that the code we produce is lot more naive and very soon breach the 255 method arguments mark. OTOH, Javac is smarter and splits the operation into multiple ones. We should do that too.
> 
> I would suggest we keep this PR for the basic implementation and follow up on a separate ticket for improvements

If it makes it easier to get this PR merged, sure. But I will fix this sooner than later. I will start working on it right after this.

jarthana:
@srikanth-sankaran I have a follow-up PR in #1172. Fairly simple and takes care of the cases when the concat gets too long with too many arguments. It undoes some of the test changes I made here. Just FYI.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/impl/CompilerOptions.java
```diff
@@ -552,6 +553,9 @@ public class CompilerOptions {
 	/** Should the compiler ignore the unnamed module when a package is defined in both a named module and the unnamed module? */
```
srikanth-sankaran, line 553: org.eclipse.jdt.internal.compiler.impl.CompilerOptions.toString() needs update ??

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/ConstantTest.java
```diff
@@ -427,9 +432,9 @@ public void test009() throws Exception {
 			"        [pc: 63, pc: 87] local: s index: 2 type: java.lang.String\n" +
 			"        [pc: 78, pc: 87] local: s2 index: 3 type: java.lang.String\n";
 
-	String expectedOutput15 =
+	String expectedOutputPost9 =
 			"  // Method descriptor #15 ([Ljava/lang/String;)V\n" +
```
srikanth-sankaran, line 436: should be `expectedOutput9OrLater`

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericTypeTest.java
```diff
@@ -56,7 +56,7 @@ public GenericTypeTest(String name) {
 	// Static initializer to specify tests subset using TESTS_* static variables
 	// All specified tests which does not belong to the class are skipped...
 	static {
-//		TESTS_NAMES = new String[] { "test0593" };
+		TESTS_NAMES = new String[] { "test1066" };
 //		TESTS_NUMBERS = new int[] { 470, 627 };
```
srikanth-sankaran, line 60: comment is removed ??

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/TryStatementTest.java
```diff
@@ -1499,14 +1499,15 @@ public void test033() throws Exception {
...
+	String builder = this.complianceLevel >= ClassFileConstants.JDK1_5 ? "StringBuilder" : "StringBuffer";
 	this.runConformTest(
 			new String[] {
 				"X.java",
 				"public class X {\n" +
 				"	private static int scenario(){\n" +
 				"		try {\n" +
 				"			int i = 1;\n" +
-				"			System.out.print(\"[i: \" + i+\"]\");\n" +
+				"			System.out.print(new " + builder + "(\"[i: \").append(i).append(\"]\").toString());\n" +
```
srikanth-sankaran, line 1510: Surprised by this change - what brought this about ? Is this a way of short circuiting the new indy based concat ?

srikanth-sankaran:
OK, I have made one more pass over the changes - The test changes I only glanced through. 

I suggest let us set up a "structured walk through" for these files:

```
BinaryExpression.java
CombinedBinaryExpression.java
Expression.java
CodeStream.java
```

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/TryStatementTest.java
jarthana: Yes, I didn't want to change all the tests' output that involved a String concat. I kept few to validate the new generated code, though.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/CombinedBinaryExpression.java
```diff
@@ -236,6 +238,50 @@ public void generateOptimizedStringConcatenation(BlockScope blockScope,
 	}
 }
 
+@Override
+public void buildStringForConcatation(BlockScope blockScope, CodeStream codeStream, int typeID, StringBuilder recipe, List<TypeBinding> argTypes) {
+	if (this.referencesTable == null) {
+		super.buildStringForConcatation(blockScope, codeStream,	typeID, recipe, argTypes);
+	} else {
+		// copied from below method
+		if ((((this.bits & ASTNode.OperatorMASK) >> ASTNode.OperatorSHIFT) ==
```
srikanth-sankaran, line 247: Better to mention method name rather than just say below.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/CombinedBinaryExpression.java
```diff
@@ -236,6 +238,50 @@ public void generateOptimizedStringConcatenation(BlockScope blockScope,
 	}
 }
 
+@Override
+public void buildStringForConcatation(BlockScope blockScope, CodeStream codeStream, int typeID, StringBuilder recipe, List<TypeBinding> argTypes) {
+	if (this.referencesTable == null) {
+		super.buildStringForConcatation(blockScope, codeStream,	typeID, recipe, argTypes);
```
srikanth-sankaran, line 244: Can you please check to make sure that we have 100% branch coverage for the new code in Expression, BinaryExpression and CombinedBinaryExpression

srikanth-sankaran (approved):

jarthana:
@srikanth-sankaran It turns out the `makeConcatWithConstants()` is useful when we pass static bootstrap arguments. But I think we can continue to use this for what we do now (pass everything as dynamic argument). This gives us the flexibility when/if we decide to pass static arguments. WDYT?

iloveeclipse:
@jarthana, @srikanth-sankaran: next week Friday is M1 planned.
Usually we declare M1 and **after** M1 switch the compiler used to build SDK to M1 version.

Should we keep with that, or do we want to promote & switch the ecj compiler used to build SDK to use this change **before** M1?

iloveeclipse:
@akurtakov : FYI. The updated compiler would generate different code now for lot of bundles, which would mean bumping versions. Better before M1 or after?

akurtakov:
The earlier the better IMHO.

srikanth-sankaran:
> The earlier the better IMHO.

I agree. The other java compiler bootstraps the entire JDK (compiler, VM and JDK) for every single line of change that serves as a excellent test bed and safety net. Sooner the better.

iloveeclipse:
I've created https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1184

@akurtakov : do you know how one can "promote" ecj build to make it available for the next SDK build? If yes, please comment in https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1184.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/CombinedBinaryExpression.java
srikanth-sankaran: > Can you please check to make sure that we have 100% branch coverage for the new code in Expression, BinaryExpression and CombinedBinaryExpression

@jarthana : this review comment perhaps fell through the cracks. Without the new test added by @iloveeclipse viz `org.eclipse.jdt.core.tests.compiler.regression.StringConcatTest.test005()` in the context of https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1202 there is no test in the full run of `RunJDTCoreTests` that exercises that branch :-(

Here is what I suggest/would do: In `Expression`, `BinaryExpression` and `CombinedBinaryExpression`, in all the new code, I would insert a `throw new UnsupportedOperationException();` and comment out the code there. Now `RunJDTCoreTests` to make sure that every one of those exception actually triggers.

stephan-herrmann: I'm running the suite in HEAD/master with coverage / EclEmma enabled. Will report the results here.

stephan-herrmann: results:
* Expression.buildStringForConcatation() not covered: lines 965-7 (block starting codeStream.ldc)
* BinaryExpression.buildStringForConcatation() not covered: line 1572 (super call)
   * first compound if condition didn't see all combinations
* CombinedBinaryExpression.buildStringForConcatation() not covered:
  * lines 261/2, this was [#1202](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1202) (tested before that one got merged)
  * lines 265-7  (inner else block)
  * line 281 (final super call)

stephan-herrmann: BTW, do you care to fix the spelling of concatenation in the method name?

iloveeclipse:
@srikanth-sankaran, @mpalat, @jarthana, @stephan-herrmann, @sravanlakkimsetti : I've created https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1203 as follow up on our discussion with @srikanth-sankaran about how we could improve ecj validation & get faster feedback cycle.

@sravanlakkimsetti, @laeubi, @akurtakov : I hope you could give some guidance & help regarding required maven SDK / ECJ build changes.

Issue #1031, opened 2023-05-06
LambdaConversionException


Aleksey2093:
**code**
```java
package com.example.demo;

import com.vaadin.flow.component.HasValue;
import com.vaadin.flow.component.orderedlayout.VerticalLayout;
import com.vaadin.flow.component.textfield.TextField;
import com.vaadin.flow.router.Route;

@Route(value = "test")
public class TestPage extends VerticalLayout {

    public TestPage() {
        setWidthFull();
        setHeightFull();
        TextField textField = new TextField();
        add(textField);
        textField.setWidthFull();
        HasValue<?, ?> hasValue = textField;
        hasValue.addValueChangeListener(event -> {
            System.out.println(event.getValue());
        });
    }

}
```

This code does not work if it was compiled using vscode.
If you build the code using mvn clean package -DskipTests=true -Pproduction, the code will work.

**main**
```
package com.example.demo;

import org.springframework.boot.SpringApplication;
import org.springframework.boot.autoconfigure.SpringBootApplication;

@SpringBootApplication
public class DemoApplication {

	public static void main(String[] args) {
		SpringApplication.run(DemoApplication.class, args);
	}

}

```

**pom.xml**:
```
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
	<modelVersion>4.0.0</modelVersion>
	<parent>
		<groupId>org.springframework.boot</groupId>
		<artifactId>spring-boot-starter-parent</artifactId>
		<version>2.7.11</version>
		<relativePath/> <!-- lookup parent from repository -->
	</parent>
	<groupId>com.example</groupId>
	<artifactId>demo</artifactId>
	<version>0.0.1-SNAPSHOT</version>
	<name>demo</name>
	<description>Demo project for Spring Boot</description>
	<properties>
		<java.version>17</java.version>
		<vaadin.version>23.3.11</vaadin.version>
	</properties>
	<dependencies>

		<dependency>
			<groupId>org.springframework.boot</groupId>
			<artifactId>spring-boot-starter-actuator</artifactId>
		</dependency>

		<dependency>
			<groupId>com.vaadin</groupId>
[...]

PR #1082, opened 2023-05-22
LambdaConversionException due to invalid instantiated method

snjeza:
LambdaConversionException due to invalid instantiated method type argument to LambdaMetafactory::metafactory

Fixes #1031 

Steps to reproduce:

- create the following class
```
public class Main {    
    public static void main(String[] args) {
		getHasValue().addValueChangeListener(evt -> {System.out.println("hello");});		
    }

    public static HasValue<?, ?> getHasValue() { 
        return new HasValue<HasValue.ValueChangeEvent<String>, String>() { 
			@Override
			public void addValueChangeListener(
					HasValue.ValueChangeListener<? super HasValue.ValueChangeEvent<String>> listener) {
				listener.valueChanged(null);
			}
		};

    }    
}

interface HasValue<E extends HasValue.ValueChangeEvent<V>,V> {    
    public static interface ValueChangeEvent<V> {}    
    public static interface ValueChangeListener<E extends HasValue.ValueChangeEvent<?>> {
        void valueChanged(E event);
    }    
    void addValueChangeListener(HasValue.ValueChangeListener<? super E> listener);
}
```
- try to run it
The issue can't be reproduced with JDK.

Related issues:
- https://bugs.eclipse.org/bugs/show_bug.cgi?id=546161
- https://github.com/eclipse/eclipse.jdt.ls/issues/2631

PR #1145, opened 2023-06-14
LambdaConversionException due to invalid instantiated method 

srikanth-sankaran:
## What it does

Fix CaptureBinding#signature()'s incorrect fall back on its super implementation.
Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1031

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding.java
@@ -263,6 +263,12 @@ public class CaptureBinding extends TypeVariableBinding {
 			case Wildcard.UNBOUND :
 				this.setSuperClass(substitutedVariableSuperclass);
 				this.setSuperInterfaces(substitutedVariableInterfaces);
+				if (substitutedVariableSuperclass != null && substitutedVariableInterfaces != null) {
+					if (substitutedVariableSuperclass.id == TypeIds.T_JavaLangObject
+							&& substitutedVariableInterfaces.length > 0) {
+						this.setFirstBound(substitutedVariableInterfaces[0]);
+					}
+				}
 				this.tagBits &= ~TagBits.HasTypeVariable;
 				break;
 			case Wildcard.SUPER :
@@ -587,19 +593,6 @@ public class CaptureBinding extends TypeVariableBinding {
 		return super.toString();
 	}
 
-	@Override
-	public char[] constantPoolName() {
-		// What is the right thing here? Can we do better than defaulting to
-		// org.eclipse.jdt.internal.compiler.lookup.TypeVariableBinding.constantPoolName()
-		return super.constantPoolName();
-	}
-
-	@Override
-	public TypeBinding erasure() {
-		// What is the right thing here? Can we do better than defaulting to
-		// org.eclipse.jdt.internal.compiler.lookup.TypeVariableBinding.erasure()??
-		return super.erasure();
-	}
 
 	@Override
 	public char[] signature() /* Ljava/lang/Object; */ {
@@ -609,8 +602,7 @@ public class CaptureBinding extends TypeVariableBinding {
 		if (this.firstBound instanceof ArrayBinding) {
 			this.signature = constantPoolName();
 		} else {
-			char[] poolName = this.wildcard != null ? this.wildcard.constantPoolName() : constantPoolName();
-			this.signature = CharOperation.concat('L', poolName, ';');
+			this.signature = CharOperation.concat('L', constantPoolName(), ';');
 		}
 		return this.signature;
 
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding.java
@@ -587,6 +587,19 @@ public class CaptureBinding extends TypeVariableBinding {
 		return super.toString();
 	}
 
+	@Override
+	public char[] constantPoolName() {
+		// What is the right thing here? Can we do better than defaulting to
+		// org.eclipse.jdt.internal.compiler.lookup.TypeVariableBinding.constantPoolName()
+		return super.constantPoolName();
+	}
+
+	@Override
+	public TypeBinding erasure() {
+		// What is the right thing here? Can we do better than defaulting to
+		// org.eclipse.jdt.internal.compiler.lookup.TypeVariableBinding.erasure()??
+		return super.erasure();
+	}
 
 	@Override
 	public char[] signature() /* Ljava/lang/Object; */ {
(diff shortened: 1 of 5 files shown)


srikanth-sankaran:
@stephan-herrmann , can you please share your opinion on this? 

Essentially: 

I think the defect arises from the following methods incorrectly falling back on the TVB implementation rather than specializing the implementation by overriding:

org.eclipse.jdt.internal.compiler.lookup.CaptureBinding.signature()
org.eclipse.jdt.internal.compiler.lookup.CaptureBinding.erasure()
org.eclipse.jdt.internal.compiler.lookup.CaptureBinding.constantPoolName()

(Actually just tweaking org.eclipse.jdt.internal.compiler.lookup.CaptureBinding.signature() appears to solve the problems with all 4 tickets, but this would be a good occasion to ask questions about whether erasure() and constantPoolName() should fall back on super TVB's implementation.)

We may need to ask of CaptureBinding18 too as to whether it is entirely appropriate to inherit behavior around the methods called out.

What is the right thing here ?

srikanth-sankaran:
@snjeza please take a look and let us discuss

srikanth-sankaran:
> (Actually just tweaking org.eclipse.jdt.internal.compiler.lookup.CaptureBinding.signature() appears to solve the problems with all 4 tickets, but this would be a good occasion to ask questions about whether erasure() and constantPoolName() should fall back on super TVB's implementation.)

The 4 tickets referred to are:

https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1082 (the snippet from top)
https://bugs.eclipse.org/bugs/show_bug.cgi?id=574269
https://bugs.eclipse.org/bugs/show_bug.cgi?id=570511
https://bugs.eclipse.org/bugs/show_bug.cgi?id=546161

snjeza:
> @snjeza please take a look and let us discuss

@srikanth-sankaran It works fine for me.

srikanth-sankaran:
> Here are all the existing tests that would see a change in behavior due to this fix:
> 
> ```
> org.eclipse.jdt.core.tests.compiler.regression.LambdaExpressionsTest.test432625()
> org.eclipse.jdt.core.tests.compiler.regression.NegativeLambdaExpressionsTest.test428177()
> org.eclipse.jdt.core.tests.compiler.regression.GenericsRegressionTest_1_8.testBug424710()
> org.eclipse.jdt.core.tests.compiler.regression.GenericsRegressionTest_1_8.testBug512156_10()
> org.eclipse.jdt.core.tests.compiler.regression.GenericsRegressionTest_1_8.testBug512156_3()
> org.eclipse.jdt.core.tests.compiler.regression.GenericsRegressionTest_1_8.testBug559951()
> org.eclipse.jdt.core.tests.compiler.regression.NullTypeAnnotationTest.testBug513855lambda()
> org.eclipse.jdt.core.tests.model.ResolveTests18.test424198a()
> org.eclipse.jdt.core.tests.model.ResolveTests18.test424198b()
> ```
> 
> Next step is to see if these tests see a welcome change or not.

OK, I have studied all these tests that reach CaptureBinding.signature(): Net observation is there is no material change due to the fix here. OIOW, in all these tests constantPoolName() == this.wildcard.constantPoolName(), so the code change has no effects ill or otherwise.

So this fix looks good to go - We will tackle with whether erasure() and constantPoolName() should inherit TVB's behavior or not in a separate follow up ticket.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding.java
```diff
@@ -596,7 +609,8 @@ public String toString() {
 		if (this.firstBound instanceof ArrayBinding) {
 			this.signature = constantPoolName();
 		} else {
-			this.signature = CharOperation.concat('L', constantPoolName(), ';');
+			char[] poolName = this.wildcard != null ? this.wildcard.constantPoolName() : constantPoolName();
+			this.signature = CharOperation.concat('L', poolName, ';');
```
stephan-herrmann, line 613: So, this change is needed in a situation where `CaptureBinding.firstBound == null` but the `wildcard` has a relevant upper bound.

Given "The erasure of a type variable ([§4.4](https://docs.oracle.com/javase/specs/jls/se20/html/jls-4.html#jls-4.4)) is the erasure of its leftmost bound." (JLS 4.6) we seem to be inconsistent as to what is a capture's leftmost bound. The super implementation assumes, that's either `firstBound` or `superclass`, but doesn't consider `superinterfaces`. Assuming this might be correct for `TypeVariableBinding`, perhaps `CaptureBinding.initializeBounds()` should be blamed? Why don't we set `firstBound`?

It happens in the 2nd `case Wildcard.UNBOUND`, where the following addition seems to render the change in `signature` unnecessary (line 266):

```
	if (substitutedVariableSuperclass.id == TypeIds.T_JavaLangObject && substitutedVariableInterfaces.length > 0) {
		this.setFirstBound(substitutedVariableInterfaces[0]);
	}
```

To me this looks like a fix closer to the root cause. WDYT?

stephan-herrmann (commented):
no clear yes or no, yet, but food for discussion :)

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding.java
srikanth-sankaran: I ran the full test suite with the proposed change and see two extra failures:

1. org.eclipse.jdt.core.tests.model.NullAnnotationModelTests.testBug495635() and
2. org.eclipse.jdt.core.tests.model.CompletionTests_1_5.test0356()

The first one is an NPE at 
```
	if (substitutedVariableSuperclass.id == TypeIds.T_JavaLangObject && substitutedVariableInterfaces.length > 0) {
		this.setFirstBound(substitutedVariableInterfaces[0]);
	}
```

which I can circumvent by null checks - Although I am not (yet at least) sure if there is some thing wrong in seeing substitutedVariableSuperclass against the TypeVariable `O` in the test case.

The other test seems to fails with
```
junit.framework.ComparisonFailure: 
getClass[METHOD_REF]{getClass(), Ljava.lang.Object;, ()Ljava.lang.Class<+Ljava.lang.Object;>;, getClass, null, 60}\n
get[METHOD_REF]{get(), Ltest.util.List<Ljava.lang.Object;>;, (I)Ljava.lang.Object;, get, (i), 64}
getClass[METHOD_REF]{getClass(), Ljava.lang.Object;, ()Ljava.lang.Class<+Ljava.lang.Object;>;, getClass, null, 60}\n
get[METHOD_REF]{get(), Ltest.util.List<Ltest.util.List<TT;>;>;, (I)Ltest.util.List<TT;>;, get, (i), 64}
```

Looking into both

srikanth-sankaran: I'll also do some wider impact analysis of the proposed change.

srikanth-sankaran: > To me this looks like a fix closer to the root cause. WDYT?

Not having any objections, I will adopt the suggested change - Thanks!

Issue #1154, opened 2023-06-17
An internal error occurred during: "Java Search"


laeubi:
I encountered this when trying to perform a search for a reference on

https://github.com/eclipse-equinox/p2/blob/98e45d63f909561b8d85c21a533fe3e51bfc70a5/bundles/org.eclipse.equinox.p2.core/src/org/eclipse/equinox/internal/p2/core/Activator.java#L55

in the Eclipse-SDK-IDE workspace:
https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#create-an-eclipse-development-environment

```
eclipse.buildId=4.29.0.I20230615-1800
java.version=17.0.2
java.vendor=Oracle Corporation
BootLoader constants: OS=linux, ARCH=x86_64, WS=gtk, NL=de_DE
Command-line arguments:  -os linux -ws gtk -arch x86_64

org.eclipse.core.jobs
Error
Sat Jun 17 07:53:09 CEST 2023
An internal error occurred during: "Java Search".

java.lang.NullPointerException: Cannot invoke "org.eclipse.jdt.internal.compiler.util.JrtFileSystem.walkModuleImage(org.eclipse.jdt.internal.compiler.util.JRTUtil$JrtFileVisitor, int)" because the return value of "org.eclipse.jdt.internal.compiler.util.JRTUtil.getJrtSystem(java.io.File, String)" is null
	at org.eclipse.jdt.internal.compiler.util.JRTUtil.walkModuleImage(JRTUtil.java:241)
	at org.eclipse.jdt.internal.core.JavaProject.loadModulesInJimage(JavaProject.java:983)
	at org.eclipse.jdt.internal.core.JavaProject.computePackageFragmentRoots(JavaProject.java:782)
	at org.eclipse.jdt.internal.core.JavaProject.computePackageFragmentRoots(JavaProject.java:1103)
	at org.eclipse.jdt.internal.core.JavaProject.computePackageFragmentRoots(JavaProject.java:1048)
	at org.eclipse.jdt.internal.core.JavaProject.computePackageFragmentRoots(JavaProject.java:1025)
	at org.eclipse.jdt.internal.core.JavaProject.buildStructure(JavaProject.java:538)
	at org.eclipse.jdt.internal.core.Openable.generateInfos(Openable.java:266)
	at org.eclipse.jdt.internal.core.JavaElement.openWhenClosed(JavaElement.java:597)
[...]

Issue #1150, opened 2023-06-16
Test fails with I20230615-1800


iloveeclipse:
We have new **reproducible** test fails on **all** platform in JDT with I20230615-1800 build.

**jdt.apt.tests**
- https://download.eclipse.org/eclipse/downloads/drops4/I20230615-1800/testresults/html/org.eclipse.jdt.apt.tests_ep429I-unit-cen64-gtk3-java17_linux.gtk.x86_64_17.html
- `AptReconcileTests | testGeneratedFile` complains about workspace settings: `Workspace options should be back to their default.`

**jdt.core.tests.model**
- https://download.eclipse.org/eclipse/downloads/drops4/I20230615-1800/testresults/html/org.eclipse.jdt.core.tests.model_ep429I-unit-cen64-gtk3-java17_linux.gtk.x86_64_17.html
- `JavaProjectTests.testAddExternalLibFolder6` complains about workspace settings: `Workspace options should be back to their default.`
- `APIDocumentationTests.testJavaCoreAPI` : some strange fail

```
One or many discrepancies, including: . ----------- Expected ------------ ------------ but was ------------ *.launch --------- Difference is ---------- expected:<[]> but was:<[*.launch]>

junit.framework.ComparisonFailure: One or many discrepancies, including: .
----------- Expected ------------

------------ but was ------------
*.launch
--------- Difference is ----------
expected:<[]> but was:<[*.launch]>
at org.eclipse.jdt.core.tests.junit.extension.TestCase.assertStringEquals(TestCase.java:266)
at org.eclipse.jdt.core.tests.junit.extension.TestCase.assertEquals(TestCase.java:242)
at org.eclipse.jdt.core.tests.dom.APIDocumentationTests.testJavaCoreAPI(APIDocumentationTests.java:302)
```

**jdt.text.tests**

- https://download.eclipse.org/eclipse/downloads/drops4/I20230615-1800/testresults/html/org.eclipse.jdt.text.tests_ep429I-unit-cen64-gtk3-java17_linux.gtk.x86_64_17.html
- `PluginsNotLoadedTest | pluginsNotLoaded` some bundle loading issue
```
Wrong bundles loaded: - org.eclipse.jdt.apt.core - org.eclipse.jdt.apt.pluggable.core expected:<0> but was:<64>

java.lang.AssertionError: Wrong bundles loaded:
- org.eclipse.jdt.apt.core
- org.eclipse.jdt.apt.pluggable.core
[...]

PR #1155, opened 2023-06-17
Fix NPE in JRTUtil

laeubi:
## What it does

In some cases JRTUtil#getJrtSystem() can return null what results in a NPE a caller most likley do not exspect.

This adds required null checks and act accordingly.

Fix https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1154

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
@@ -16,6 +16,7 @@ package org.eclipse.jdt.internal.compiler.util;
 
 import java.io.ByteArrayInputStream;
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.FileReader;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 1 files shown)


laeubi:
@ jarthana @iloveeclipse can you take a look here, this is really annoying and currently makes working with the Java Search impossible in my Platform-IDE-SDK as it seems I have some project imported that triggers the NPE every time.
Test failure seems unrelated as I see it on other changes as well (e.g. here: https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-676/10/)

iloveeclipse:
Please rebase to get rid of #1150 issue.

Also I believe your issue could be caused by https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/231 and your fix is applied at wrong place.

I assume somehow the "autoimport" of JDK/JRE managed to setup invalid JDK/JRE.

Please check that: which is now the default one and how to prevent importing that one in first place.
Hint: I assume some project now uses JRE, not JDK as default, or something like this.

laeubi:
> Also I believe your issue could be caused by https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/231 and your fix is applied at wrong place.

I think this fix is good anyways, as obviously there can be `null` returned, so the code should handle this.

> I assume somehow the "autoimport" of JDK/JRE managed to setup invalid JDK/JRE.

I'll check that, how would I identify such "invalid JDK"? If i look at the dialog it all looks quite usual
grafik

I also selected each JDK and hit "edit" but no error was reported.

iloveeclipse:
> I'll check that, how would I identify such "invalid JDK"?

If you can always reproduce your issue, try to debug it and see which JDK causes the problem.
Do you have other errors in the log that could give a hint?

laeubi:
As it is on the batch-compiler it seems only system-out is printing something I started eclipse with a console now and see this error: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1154#issuecomment-1596509618

laeubi:
@jarthana @iloveeclipse I have rebased this now and all tests seem green. If you think it should be fixed at another place I'm fine with that but this currently makes my Eclipse-Ide somewhat unuseable...

iloveeclipse:
> If you think it should be fixed at another place I'm fine 

I think if possible we should check the presence of that file on adding such JRE's.

> with that but this currently makes my Eclipse-Ide somewhat unuseable...

Just remove or fix that Java 9 installation.

laeubi:
> Just remove or fix that Java 9 installation (/usr/lib/jvm/java-9-openjdk-amd64/release is missing).

If you can point to the spec that it requires a `release` file to be present for a JVM I can try to "fix" this, but unless then I fear I won't get much attention for that, I probably can even create an empty file, remove the JVM and so on, but to the core I think JDT can simply better handle the case, a NPE is from my point of view never acceptable here and the code obviously **can** return null, so what ever seems appropriate (ignore, throw an documented exception, ...) should be applied to JDT in parallel.

iloveeclipse:
> If you can point to the spec that it requires a `release` file to be present for a JVM I can try to "fix" this, but unless then I fear I won't get much attention for that

Well, almost all of JDK internal structure is not spec'd, even ct.sym is not - should we now rewrite all of JDT code that relies on the JDK internals? You've complained about, and I've offered a solution that would immediately help. You don't need to apply this solution. 

Note: the problem with "broken" JDK will appear surely later on in some different code in JDT, you may not even notice where it will be because the search or call hierarchy would return only a subset of matches.

laeubi:
> Well, almost all of JDK internal structure is not spec'd

And that's why I suggest to handle such cases... some of the methods are even throwing an exception so callers can handle / forward this, but the NPE bubbles up into an ugly error dialog I have to dismiss all the time.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
```diff
@@ -238,35 +239,67 @@ public static void reset() {
...
+		JrtFileSystem system = getJrtSystem(image, null);
+		if (system == null) {
+			return;
+		}
+		system.walkModuleImage(visitor, notify);
 	}
 
 	public static void walkModuleImage(File image, String release, final JRTUtil.JrtFileVisitor<java.nio.file.Path> visitor, int notify) throws IOException {
-		getJrtSystem(image, release).walkModuleImage(visitor, notify);
+		JrtFileSystem system = getJrtSystem(image, release);
```
iloveeclipse, line 250: Ideally getJrtSystem() should re-throw original IO exception where possible, not simply return null or false (the caller is supposed to properly handle this case as exception).

I will push a follow up PR to cleanup this.

iloveeclipse (approved):
The change itself is not complete, but can be merged. I will push a follow up PR.

iloveeclipse:
> I will push a follow up PR to cleanup this.

See https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1170

laeubi:
> deally getJrtSystem() should re-throw original IO exception where possible, not simply return null or false (the caller is supposed to properly handle this case as exception).

Thanks I also has thought about it, but was not sure about the consequences (especially as it seems one needs to enable exceptions explicitly). The followup looks like a good path forward!

PR #1160, opened 2023-06-19
fix "static field should be accessed in a static way"

jukzi:
reduce ecj warnings


org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/AbstractJavaModelTests.java
```diff
@@ -3708,6 +3708,8 @@ public void setUpJCLClasspathVariables(String compliance, boolean useFullJCL) th
 	@Override
 	public void setUpSuite() throws Exception {
 		super.setUpSuite();
+		Hashtable defaultOptions = getDefaultJavaCoreOptions();
```
iloveeclipse, line 3711: Please undo this change.

iloveeclipse (changes requested):

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/AbstractJavaModelTests.java
jukzi: upps. thanks for review!!

iloveeclipse (approved):

Issue #1162, opened 2023-06-19
Eclipse incorrectly requires catch for nested sneaky throws; OpenJDK compiles with no problem


garretwilson:
In Eclipse EE 2023-06 with Java 17 on Windows 10, my project is using the [sneaky throws](https://projectlombok.org/features/SneakyThrows) technique via the [Faux Pas](https://github.com/zalando/faux-pas) library. This technique is not new, and many projects have been using it with no problems, as have I for many years.

I'm using a method that looks like this:

```java
public void showEclipseProblem() throws IOException {
  final DocumentBuilder documentBuilder = null; //TODO
  try (final Stream<Path> barDirs = Files.list(Paths.get("foo"))) {
    barDirs.forEach(FauxPas.throwingConsumer(barDir -> {
      try (final Stream<Path> files = Files.list(barDir)) {
        files.forEach(FauxPas.throwingConsumer(file -> {
          final Document document;
          try (final InputStream inputStream = new BufferedInputStream(newInputStream(file))) {
//TODO uncomment this line to see the problem:             document = documentBuilder.parse(inputStream);
          }
        }));
      }
    }));
  }
}
```

This code compiles fine on Eclipse EE 2023-06 and on OpenJDK Runtime Environment Temurin-17.0.4+8 (build 17.0.4+8).

When I uncomment the line `document = documentBuilder.parse(inputStream)`, OpenJDK continues to compile the code with no problems. However Eclipse suddenly underlines `Files.list(barDir)` in red, indicating the following error:

> `Unhandled exception type IOException`

This error makes no logical sense, because `Files.list(barDir)` is inside `FauxPas.throwingConsumer(…)`, which should allow the `IOException` to be thrown with no declaration or `catch`. In fact to show how senseless this is, I can actually change the code to throw an exception manually, and Eclipse (correctly) finds no problem with it:

```java
public void showEclipseProblem() throws IOException {
  final DocumentBuilder documentBuilder = null; //TODO
  try (final Stream<Path> barDirs = Files.list(Paths.get("foo"))) {
    barDirs.forEach(throwingConsumer(barDir -> {
      try (final Stream<Path> files = Files.list(barDir)) {
        throw new IOException("Eclipse correctly doesn't think this needs to be caught.");
      }
    }));
  }
}
```

Furthermore it's odd that adding a completely unrelated line `document = documentBuilder.parse(inputStream)` would cause this error to appear on a completely other line altogether.

[...]

PR #1166, opened 2023-06-20
Fix problems in lambda exception analysis

srikanth-sankaran:
## What it does

* The method org.eclipse.jdt.internal.compiler.ast.LambdaExpression.analyzeExceptions() over-reaches into current lambda locals while wanting to flag all outer locals touched by the lambda as being definitely assigned and
   effectively final.

* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1162

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
@@ -560,9 +560,17 @@ public class LambdaExpression extends FunctionalExpression implements IPolyExpre
 		boolean oldAnalyseResources = compilerOptions.analyseResourceLeaks;
 		compilerOptions.analyseResourceLeaks = false;
 		try {
+			int firstLocalLocal = this.scope.outerMostMethodScope().analysisIndex;
+			LocalVariableBinding[] locals = this.scope.locals;
+			if (locals != null) {
+				for (LocalVariableBinding local : locals) {
+					if (local != null && local.isParameter())
+						firstLocalLocal = local.id + 1;
+				}
+			}
 			this.body.analyseCode(this.scope,
 									 ehfc = new ExceptionInferenceFlowContext(null, this, Binding.NO_EXCEPTIONS, null, this.scope, FlowInfo.DEAD_END),
-									 UnconditionalFlowInfo.fakeInitializedFlowInfo(this.scope.outerMostMethodScope().analysisIndex, this.scope.referenceType().maxFieldCount));
+									 UnconditionalFlowInfo.fakeInitializedFlowInfo(firstLocalLocal, this.scope.referenceType().maxFieldCount));
 			this.thrownExceptions = ehfc.extendedExceptions == null ? Collections.emptySet() : new HashSet<TypeBinding>(ehfc.extendedExceptions);
 		} catch (Exception e) {
 			// drop silently.
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
@@ -567,9 +567,6 @@ public class LambdaExpression extends FunctionalExpression implements IPolyExpre
 					if (local != null) {
 						if (local.isParameter()) {
 							firstLocalLocal = local.id + 1;
-						} else {
-							firstLocalLocal = local.id;
-							break;
 						}
 					}
 				}
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
@@ -564,8 +564,14 @@ public class LambdaExpression extends FunctionalExpression implements IPolyExpre
 			LocalVariableBinding[] locals = this.scope.locals;
 			if (locals != null) {
 				for (LocalVariableBinding local : locals) {
-					if (local != null && local.isParameter())
-						firstLocalLocal = local.id + 1;
+					if (local != null) {
+						if (local.isParameter()) {
+							firstLocalLocal = local.id + 1;
+						} else {
+							firstLocalLocal = local.id;
+							break;
+						}
+					}
 				}
 			}
 			this.body.analyseCode(this.scope,
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/LambdaExpressionsTest.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/LambdaExpressionsTest.java
@@ -7951,7 +7951,6 @@ public void testBug511958() {
 			);
 }
 
-
 // https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1060
 // NPE when inspecting scrapbook expression that uses Java 8 features
 public void testGH1060() {
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/LambdaExpressionsTest.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/LambdaExpressionsTest.java
@@ -7951,6 +7951,7 @@ public void testBug511958() {
 			);
 }
 
+<<<<<<< Upstream, based on branch 'master' of git@github.com:srikanth-sankaran/eclipse.jdt.core.git
 // https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1060
 // NPE when inspecting scrapbook expression that uses Java 8 features
 public void testGH1060() {
@@ -8250,6 +8251,8 @@ public void testBug577719_2() {
 			);
 }
 
+=======
+>>>>>>> b8be702 Handle lambdas with no parameters
 // https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1162
 // Eclipse incorrectly requires catch for nested sneaky throws; OpenJDK compiles with no problem
 public void testGH1162() {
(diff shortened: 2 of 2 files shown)


srikanth-sankaran:
@stephan-herrmann, Can you please take a look ? I have raised a separate ticket https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1169 for follow on investigation. 

`LE.tagAsHavingIgnoredMandatoryErrors` which is supposed to be a safety net seems to introduce some unforeseen correctness issues by short circuiting lambda exception analysis.

For the present ticket, let us just focus on whether the code as it exists today is implementing the stated intent of
org.eclipse.jdt.internal.compiler.flow.UnconditionalFlowInfo.fakeInitializedFlowInfo(int, int) 

```
fakeInitializedFlowInfo: For Lambda expressions tentative analysis during overload resolution.
   We presume that any and all outer locals touched by the lambda are definitely assigned and
   effectively final. Whether they are or not is immaterial for overload analysis (errors encountered
   in the body are not supposed to influence the resolution. It is pertinent only for the eventual
   resolution/analysis post overload resolution. For lambda's the problem is that we start the control/data
   flow analysis abruptly at the start of the lambda, so we need to present a cogent world view and hence
   all this charade.
``` 

There is also another follow up ticket https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1167

srikanth-sankaran:
It is likely https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1181 renders this fix technically unnecessary, but still is a good idea since it makes the code do what the comment claims.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
```diff
@@ -560,9 +560,23 @@ private void analyzeExceptions() {
 		boolean oldAnalyseResources = compilerOptions.analyseResourceLeaks;
 		compilerOptions.analyseResourceLeaks = false;
 		try {
+			int firstLocalLocal = this.scope.outerMostMethodScope().analysisIndex;
+			LocalVariableBinding[] locals = this.scope.locals;
+			if (locals != null) {
```
stephan-herrmann, line 565: I was going to flag this null check as unnecessary, until I found that CodeSnippetScope and ModuleScope could indeed be constructed without assigning `locals`.
That in itself looks a bit like a bad smell (but surely modules have no locals ...). Anyway, this doesn't affect the current PR. Just mentioning.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
```diff
@@ -560,9 +560,23 @@ private void analyzeExceptions() {
 		boolean oldAnalyseResources = compilerOptions.analyseResourceLeaks;
 		compilerOptions.analyseResourceLeaks = false;
 		try {
+			int firstLocalLocal = this.scope.outerMostMethodScope().analysisIndex;
```
stephan-herrmann, line 563: Strictly speaking `analysisIndex` points to the next _free_ slot in locals. But a variable name `firstLambdaLocalIfThereWereAny` doesn't look better, nor is `indexAfterTrulyOuterLocals` in for a beauty award :)

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
```diff
@@ -560,9 +560,23 @@ private void analyzeExceptions() {
...
+			LocalVariableBinding[] locals = this.scope.locals;
+			if (locals != null) {
+				for (LocalVariableBinding local : locals) {
+					if (local != null) {
+						if (local.isParameter()) {
+							firstLocalLocal = local.id + 1;
+						} else {
+							firstLocalLocal = local.id;
+							break;
```
stephan-herrmann, line 572: I couldn't find a test touching this block. Do you care to construct one which does?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
```diff
@@ -560,9 +560,23 @@ private void analyzeExceptions() {
 		boolean oldAnalyseResources = compilerOptions.analyseResourceLeaks;
 		compilerOptions.analyseResourceLeaks = false;
 		try {
+			int firstLocalLocal = this.scope.outerMostMethodScope().analysisIndex;
+			LocalVariableBinding[] locals = this.scope.locals;
+			if (locals != null) {
+				for (LocalVariableBinding local : locals) {
+					if (local != null) {
+						if (local.isParameter()) {
```
stephan-herrmann, line 568: Why is this distinction necessary? Is it because parameters are never unassigned? But treating them the same as regular locals wouldn't hurt, would it?

stephan-herrmann (approved):
The change basically looks good to go. Perhaps another test would be great.
The other remarks are nitpicks where any reaction is purely optional.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
srikanth-sankaran: For "regular" non-parameter locals of the current lambda - we don't want to mark them as definitely assigned. That was the bug here - we are assigning a blank final local as being already definitely assigned and when we see an assignment in the body of the lambda, report an error about double assignment to final variable - this inhibits exception analysis since we flag the lambda as having ignored mandatory errors.

So this code is computing the high water mark index value up to which we want to mark the locals (outer locals + parameters of the current lambda if any) as definitely assigned.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
srikanth-sankaran: I think you caught a truly unreachable piece of code. When I wrote this I conceptually viewed the lambda's parameters and the any locals of the outermost block body occupying the same scope. But that is now things are laid out. When the lambda has a block body, the block has its own scope - so we will not find any non-parameter true locals in the locals array being traversed. I'll remove this piece of code

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
stephan-herrmann: Sure, I just figured that parameters are always marked as definitely assigned anyway, and hence they don't _need_ the additional marking (which would simplify the code). In other words: only variables from truly outer scopes need the special treatment. Or am I missing smth?

srikanth-sankaran: > Sure, I just figured that parameters are always marked as definitely assigned anyway, and hence they don't _need_ the additional marking (which would simplify the code). In other words: only variables from truly outer scopes need the special treatment. Or am I missing smth?

**Here**, we **start** the analysis at the lambda **body** - 

c.f. 

`org.eclipse.jdt.internal.compiler.ast.LambdaExpression.analyseCode(BlockScope, FlowContext, FlowInfo)` which by calling `org.eclipse.jdt.internal.compiler.ast.AbstractMethodDeclaration.analyseArguments(LookupEnvironment, FlowInfo, Argument[], MethodBinding)` ensures that parameters are marked as definitely assigned. 

That part of the analysis dealing with the lambda header is skipped here, so this is not *additional* marking, but markiking *in lieu*

stephan-herrmann: > **Here**, we **start** the analysis at the lambda **body** -

I see, so we bypass `analyseArguments`. Thanks for explaining.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
stephan-herrmann: Now, that you explained this I found a way to bypass your fix (by removing the lambda parameter):
```
import java.io.IOException;
import java.nio.file.Path;
import java.util.function.Consumer;

public class X11 {

	public X11 parse() throws IOException {
		return null;
	}
	
	Stream<Path> list() throws IOException {
		return null;
	}

	public void foo() throws IOException {

		throwingConsumer(() -> {
			try (final Stream<Path> files = list()) {
				throwingConsumer(() -> {
					final X11 document;// = parse();
					document = parse();
				});
			}
		});
	}
	public static <T, X extends Throwable> ThrowingConsumer<T, X> throwingConsumer(
			final ThrowingConsumer<T, X> consumer) {
		return consumer;
	}
}


interface ThrowingConsumer<T, X extends Throwable> extends Consumer<T> {

	void tryAccept() throws X;

	default void accept(final T t) {
		tryAccept();
	}
}
```
yields
```
		"----------\n" + 
		"1. ERROR in X11.java (at line 19)\n" + 
		"	try (final Stream<Path> files = list()) {\n" + 
		"	                                ^^^^^^\n" + 
		"Unhandled exception type IOException\n" + 
		"----------\n" + 
		"2. ERROR in X11.java (at line 39)\n" + 
		"	tryAccept();\n" + 
		"	^^^^^^^^^^^\n" + 
		"Unhandled exception type X\n" + 
		"----------\n"
```

stephan-herrmann:
@srikanth-sankaran was my comment too late: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1166#discussion_r1246855662

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
srikanth-sankaran: Oops. I'll take a look and follow up. 

But there is already another PR waiting for your review that may completely short circuit the manifestation of a problem due to this.

srikanth-sankaran:
> @srikanth-sankaran was my comment too late: [#1166 (comment)](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1166#discussion_r1246855662)

@stephan-herrmann , entirely my fault! I was trying to clear this up last night so I can devote all day today to the review of null analysis changes. I am following up with https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1199 which also includes an additional change which can avoid head scratching on the part of passers by (RE was still using `lambdaScope.outerMostMethodScope().analysisIndex` for its invocation of `fakeInitializedFlowInfo` for the implicit lambda generation scenario (I think this is actually harmless since there can't be blank finals in the implicit lambda, but for consistency's sake better to harmonize)

Issue #1164, opened 2023-06-20
Handle the StringConcatFactory invocation with 200+ arguments


jarthana:
This is a follow up issue for issue #1139. We need to somehow handle or warn users when the arguments slots exceeds the max limit of 200. I don't yet have much insight into what needs to be done. This is to track the open issue.

PR #1172, opened 2023-06-21
Handle the StringConcatFactory invocation with 200+ arguments #1164

jarthana:
## What it does
Fixes issue #1164 

## Author checklist


jarthana:
This needs to be applied on top or PR #1139. We only need the last commit from here. Rest of them are already on the said PR.

jarthana:
@srikanth-sankaran In the interest of getting this tested enough, I will release this change (also expecting the new fail-fast approach help me here). 

You are welcome to review this whenever you find time.

jarthana:
This needs to wait. I see that the stack size goes up for all aload of the operands, but never removed when we consume them by a makeConcatWithConstants. Working on a fix.

jarthana:
When we have too many operands in a String concat expression, the StringConcatFactory hits the 200 arguments limit and crashes. To avoid this, we wrap every 190 arguments into a StringConcatFactory#makeConcatWithConstants() instead of keeping them in the stack. This way, we can accommodate thousands of operands (to be precise 190x199) before hitting the limit.

Issue #1179, opened 2023-06-22
Try to (automatically?) rebuild index on index format error


ewillink:
eclipse.buildId=4.27.0.I20230302-0300
java.version=17.0.6
java.vendor=Oracle Corporation
BootLoader constants: OS=win32, ARCH=x86_64, WS=win32, NL=en_GB
Command-line arguments:  -os win32 -ws win32 -arch x86_64 -clean -data E:\Development\Chital\Workspace

I have been observing frequent malfunction of searches and import quickfixes and finally noticed that at the root of this problem is the 'loss' of a plugin as consequence of the error entry below. (The lost plugin is ignored from all search results.)

java.io.IOException: Wrong format
	at org.eclipse.jdt.internal.core.index.DiskIndex.initialize(DiskIndex.java:410)
	at org.eclipse.jdt.internal.core.index.Index.<init>(Index.java:120)
	at org.eclipse.jdt.internal.core.search.indexing.IndexManager.loadMetaIndexIfNeeded(IndexManager.java:1657)
	at org.eclipse.jdt.internal.core.search.indexing.IndexManager.findMatchingIndexNames(IndexManager.java:1568)
	at org.eclipse.jdt.internal.core.search.IndexSelector.getIndexLocations(IndexSelector.java:319)
	at org.eclipse.jdt.internal.core.search.JavaSearchParticipant.selectIndexURLs(JavaSearchParticipant.java:160)
	at org.eclipse.jdt.internal.core.search.PatternSearchJob.getIndexes(PatternSearchJob.java:186)
	at org.eclipse.jdt.internal.core.search.PatternSearchJob.ensureReadyToRun(PatternSearchJob.java:99)
	at org.eclipse.jdt.internal.core.search.processing.JobManager.performConcurrentJob(JobManager.java:262)
	at org.eclipse.jdt.internal.core.search.BasicSearchEngine.findMatches(BasicSearchEngine.java:235)
	at org.eclipse.jdt.internal.core.search.BasicSearchEngine.search(BasicSearchEngine.java:602)
	at org.eclipse.jdt.core.search.SearchEngine.search(SearchEngine.java:670)
	at org.eclipse.jdt.internal.ui.search.JavaSearchQuery.run(JavaSearchQuery.java:178)
	at org.eclipse.search2.internal.ui.InternalSearchUI$InternalSearchJob.run(InternalSearchUI.java:94)
[...]

PR #1182, opened 2023-06-22
Try to automatically rebuild meta index on index reading errors

iloveeclipse:
- only done for meta index, because regular index is already rebuilt on error in IndexManager.getIndex()
- small fix for disc index stream closing on error
- improved error message on index format error
- synchronized deleteIndexFiles() and reset() called from rebuildIndex()

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1179


org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/IndexManager.java
```diff
@@ -1654,7 +1652,18 @@ private MetaIndex loadMetaIndexIfNeeded() throws IOException {
...
+					this.metaIndex = new MetaIndex(new Index(indexLocation, INDEX_META_CONTAINER, true));
+				} catch (IOException e) {
+					Util.log(e, "Failed to read saved meta index, re-creating"); //$NON-NLS-1$
+					if (VERBOSE) {
+						Util.verbose("-> failed to read saved meta-index: "+indexLocation+" path: "+INDEX_META_CONTAINER); //$NON-NLS-1$ //$NON-NLS-2$
+						Util.verbose("-> re-create meta-index: "+indexLocation+" path: "+INDEX_META_CONTAINER); //$NON-NLS-1$ //$NON-NLS-2$
+					}
+					indexLocation.delete();
+					this.metaIndex = new MetaIndex(new Index(indexLocation, INDEX_META_CONTAINER, false));
+					updateIndexState(indexLocation, REUSE_STATE);
```
iloveeclipse, line 1665: @gayanper : I'm not sure about this `REUSE_STATE`, I just copied code from the branch above where the new index is created.

gayanper: We used this this state to avoid normal index handling code not to delete this index and try to rebuild it.

gayanper (approved):
LGTM

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/IndexManager.java
iloveeclipse: OK, so if index is broken (or incompatible) it is OK to delete it, so REUSE_STATE would trigger rebuild after delete, right?

gayanper: If we are doing a full re-index, then that should be fine.

Issue #1184, opened 2023-06-23
IndexBasedHierarchyBuilder OOME


jukzi:
Recently i got OOME more often after working with eclipse for a long time.
Today i got heapdump which indicates IndexManager leaks memory - or at least holds to much (or is it nowadays supposed to consume 600MB for platform workspace?):
![image](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/51790620/1f01813f-21cf-4b57-9655-e4ed7fbe1185)
![image](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/51790620/ea3be2ce-866a-4133-b2ff-ff370b36f46d)
The most memory is hold in HierarchyBuilder$1.contents. Looks like an error that it keeps the content of thousends of files open:
![image](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/51790620/4e0111a2-0dee-4a62-b5cf-ff52bf322105)
My guess is that HierarchyBuilder$1 referrers to the anonymous ResourceCompilationUnit instance created in
org.eclipse.jdt.internal.core.hierarchy.HierarchyBuilder.createCompilationUnitFromPath(Openable, IFile, char[])

I don't see any code that releases org.eclipse.jdt.internal.core.hierarchy.IndexBasedHierarchyBuilder.cuToHandle.

iloveeclipse:
> I don't see any code that releases org.eclipse.jdt.internal.core.hierarchy.IndexBasedHierarchyBuilder.cuToHandle.

I don't see any code that **really** uses IndexBasedHierarchyBuilder.cuToHandle. It was removed / refactored around 2001 - 2003. PR is set.

iloveeclipse:
Jörg, please keep an eye on that and check with the next nightly if the problem just moved to some other place.

PR #1185, opened 2023-06-23
Remove unused IndexBasedHierarchyBuilder.cuToHandle map that leak memory

iloveeclipse:
This map isn't used since ~20 years but still leaks memory.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1184


jukzi:
nice, black hole removed.

PR #1103, opened 2023-05-29
Add completion boost for constructor parameters Fix: #1102

gayanper:
## What it does
Rank the completions on constructor parameters by considering the parameter type and name at the location based on what was introduced through [#1102](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/779)

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1102

## How to test
Unit test and sample in the issue

## Author checklist

- [x] I have thoroughly tested my changes
- [x] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [x] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

Issue #1102, opened 2023-05-29
Suggest attributes and method completions based on the target parameter type and name on constructors


gayanper:
Continuation of https://github.com/eclipse-jdt/eclipse.jdt.core/issues/778 for constructors.

```
public class  Execution {
   public Execution(String taskId, String taskName, Object payload) {}
}

public void runSomeTasks(Task task) {
   new Execution(|
}

public record Task(String taskId, String taskName, Object payload){}

```

Now when I complete on `Execution` parameters from Task parameter, I would expect the name to be considered after target type when ranking the completions.

gayanper:
Closing as this has been fixed.

PR #1192, opened 2023-06-27
Fix regression from #1102

gayanper:
## What it does
The fix the regression which added multiple constructor completions when completing constructor parameters which was introduced from #1103

## How to test
Unit test updated

## Author checklist


gayanper:
@testforstephen and @rgrunber Please let me know what you think, I didn't found a better way of collecting the constructor method binding other than adding a new flag not to collect it as a proposal even though its kind of different from other implementation.

gayanper:
@rgrunber shall we merge this ?

gayanper:
I would like to merge this since the original change cause a regression in parameter completion s on constructors, specially visible after second parameter.

Please object if anyone thinks i should hold this off.

iloveeclipse:
@gayanper : I have no objections, but I'm not familiar with the code in question / your changes at all. If it is a fix for regression, and you have a test that confirms that fix, and build is green - why not merge?

Issue #1100, opened 2023-05-29
ContentAssist / CompletionScanner running into deadlock


Kademlia:
Hey, 
I recently ran into a case that would freeze up Eclipse completely trough the ContentAssist-Popup.
Freeze seems to be originating from **AsyncCompletionProposalPopup** inner Lambda expression.

I was able to roughly reproduce this with a unedited version of eclipse-java-2023-03-R-win32-x86_64. The IDE freezes nearly the whole computer for about 20 seconds (5950x 16 core), content Assist fails with OutOfMemoryException in the end. Writing "boolean test" and pressing CTRL+Space as shown in the screenshot will cause CPU/Mem to spike.

openjdk 17.0.6 2023-01-17
OpenJDK Runtime Environment JBR-17.0.6+10-829.5-jcef (build 17.0.6+10-b829.5)

All of the load seems to come from created instances of **CompletionScanner** < **CompletionJavadocParser** < CompletionParser < AssistParser < CompletionParser

Timestamp of this happening: https://www.twitch.tv/videos/1831458336?t=05h40m22s

Minified example code freezing my instance:
```
package gg;

import java.lang.StackWalker.Option;

public class PlanetTravelActionLagTest {

	private void test() {
		Option opt = Option.RETAIN_CLASS_REFERENCE;

		switch (opt) { // <- Write 'boolean test' here and press CTRL+Space
			case RETAIN_CLASS_REFERENCE -> {
			}
			case SHOW_HIDDEN_FRAMES -> {
			}
			case SHOW_REFLECT_FRAMES -> throw new UnsupportedOperationException("Unimplemented case: ");
		}

	}

}
}
```

![002934_](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/152798/9626d60b-8e6d-4919-ad83-0fa3345f0b79)

![grafik](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/152798/688f16b8-e080-407b-ac73-206cee046667)

![grafik](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/152798/69936a36-1819-46ed-b9ba-b8a1322021c6)

![grafik](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/152798/6b0d7ef1-68aa-48b0-ab6e-68c3bf960b77)

![grafik](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/152798/247a722c-d912-439e-9850-b1394c114255)

[...]

PR #1194, opened 2023-06-28
ContentAssist / CompletionScanner running into deadlock

srikanth-sankaran:
## What it does

* Ensure that attempts to nudge recovery along don't stuck reducing empty statements

* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1100

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests17.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests17.java
@@ -654,4 +654,34 @@ public class CompletionTests17 extends AbstractJavaModelCompletionTests {
 				requestor.getResults());
 
 	}
+	// https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1100
+	// ContentAssist / CompletionScanner running into deadlock
+	public void testGH1100() throws JavaModelException {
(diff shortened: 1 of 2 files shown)


srikanth-sankaran:
In the method org.eclipse.jdt.internal.codeassist.impl.AssistParser.fallBackToSpringForward(Statement), there is a comment that reads:

`// OK, next token is no good to resume "in place", attempt some local repair. FIXME: need to make sure we don't get stuck keep reducing empty statements !!` - this was added by me in 2014 November - my last month in inning #1,  it has taken a long while to get to inning #2 and to the FIXME :-(

Anyways, code completion is an extremely complex area nothing short of dark magic as it stands - there is so little good quality documentation on it, there is so much global state that is scattered in bits and pieces each dealing with some corner case or other. Each successive maintainer has tried (and failed IMHO) to unearth the design and has ended up adding his own complex layers on top. 

This is not an intractable area in truth, but in practice very much so because most committers have no time to grasp the full picture, but do as well as the blind men in the fable with elephants.

The present ticket is even more tricky - CompletionParser ends up seeing lambdas where there are none! Things go downhill pretty quickly from there. I have added the smarts to detect where recovery is not progressing at all despite our attempts to nudge it along - by tracking the parser's and scanner's state to deduce deadlock and break free.

This would prevent eclipse from hanging. (There are no useful completion proposals at the site)

PR #1200, opened 2023-06-30
Touch bundles affected by the new ecj version

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1184


iloveeclipse:
There are 3 new fails in jenkins I also could reproduce locally, but also without my changes here.
Created https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1201 for investigation.
Since this PR doesn't touch any functionality that could affect tests and we need a stable SDK build, merging.

Issue #1201, opened 2023-06-30
CombinedBinaryExpression.buildStringForConcatation() produces unexpected bytecode


iloveeclipse:
We see 3 new fails on jenkins (and locally), but not in SDK build.
I couldn't find how to "fix" the tests, reverting jdt.core back to I20230623-1800 does not help.

[org.eclipse.jdt.core.tests.model.CompletionTests_1_5.test482775](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-1200/1/testReport/junit/org.eclipse.jdt.core.tests.model/CompletionTests_1_5/test482775/)
[org.eclipse.jdt.core.tests.model.CompletionTests18.testBug473654](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-1200/1/testReport/junit/org.eclipse.jdt.core.tests.model/CompletionTests18/testBug473654/)
[org.eclipse.jdt.core.tests.model.CompletionTests18.testBug537679](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-1200/1/testReport/junit/org.eclipse.jdt.core.tests.model/CompletionTests18/testBug537679/)

iloveeclipse:
OK, after we built SDK with latest ecj version, we see fails on SDK build too.
So this is direct side effect of https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1139

See
- https://download.eclipse.org/eclipse/downloads/drops4/I20230630-1800/testResults.php
- https://download.eclipse.org/eclipse/downloads/drops4/I20230630-1800/testresults/html/org.eclipse.jdt.core.tests.model_ep429I-unit-cen64-gtk3-java20_linux.gtk.x86_64_20.html

iloveeclipse:
Yep. One can grab the SDK build from https://download.eclipse.org/eclipse/downloads/drops4/I20230629-0130/ and the tests will be green if executed locally, on same (latest) commit on master.

Starting with the build https://download.eclipse.org/eclipse/downloads/drops4/I20230629-1200/ (where the https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1139 was merged), tests fail in the IDE.

[...]

PR #1202, opened 2023-07-01
CombinedBinaryExpression.buildStringForConcatation() produces unexpected

iloveeclipse:
bytecode

- additionally to the actual fix and test added StringConcatTest to the test suite (that was forgotten to be added before).

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1201


iloveeclipse:
Note: PR fails because of 3 failing tests from #1201, and they will be only fixed after we've built tests with the fixed compiler. Currently tests are built with broken one.

jarthana (approved):

jarthana:
I confirm that the fix is what the original fix should have been.

I have also verified that the tests fail locally on the latest build but fail when the build is patched with this PR.

PR #1219, opened 2023-07-08
Null constraint mismatch Regression in 4.28

### LINKED ISSUE #1152: Null constraint mismatch Regression in 4.28:
The following code compiles fine in 4.27

`	package bug;
	
	public class NonNullBug
	{
		interface Interface<@NonNull T>
		{
		}
	
		class Class<T> implements Interface<@NonNull T> // 4.28 Null constraint mismatch: The type 'T' is not a valid substitute for the type parameter '@NonNull T'
		{
		}
	}
`

but, as commented, reports a @NonNull mismatch in 4.28.

stephan-herrmann:
fixes #1152


stephan-herrmann:
We were updating a TVB instance with explicit null annotations from the underlying original TVB without annotations (just for the case that null defaults would have affected the original).
=> Just make sure that explicit null annotations are never dropped in this process.

PR #1220, opened 2023-07-08
--release option breaks fix for #88

### LINKED ISSUE #703: --release option breaks fix for #88
Since #88 the methods `java.util.Objects.requireNonNull(..)` should be synthetically decorated for a non-null return type.

The implementation relies on inspecting argument names from the .class file, which, however, does not work when instead the type is read from a `.sig` file inside `ct.sym`.

stephan-herrmann:
fixes #703


stephan-herrmann:
The fix switches from counting argumentNames (which are not available from ct.sym) to scanning the method descriptor for the same count.

The intermediate test failure indicated lack of test isolation: the existing test `NullAnnotationBatchCompilerTest.testGHTycho1641()` superimposes a JRE class with external annotations, which then influenced subsequent tests referring to the same class.

For test isolation I added a way to clear the relevant slice of `org.eclipse.jdt.internal.compiler.batch.ClasspathJrt.ModulesCache`. Right now this is used only by this one test, but this again brings up the issue of different real world projects (with different eea configuration) accessing shared classes from this JRT cache :(

Issue #1224, opened 2023-07-11
[20] Switch statement with record pattern illegally reported as non-exhaustive


jarthana:
The switch in the following code is rejected by ECJ as not exhaustive:

```
interface I {}
class Class implements I {}
record Record(I s) {}
public class X1 {
    @SuppressWarnings("preview")
	public void foo(Record exp) {
        switch (exp) {
            case Record(Class s) -> {break;}
        }
    }
}
```

This is because, as per RecordPattern#isTotalForType() , it is considered to be always false. This method needs consider the record pattern's type as well as the nested patterns' types.

jarthana:
Since this is still a preview feature in 20 and a standard feature in 21, I would rather we fix this for 21 directly.

Copying @mpalat

Issue #1099, opened 2023-05-29
[21] record-code gen issues - MatchException plus verification


mpalat:
Accessor access in  a record codegen should be with j.l.MatchException 
There are a few stack verify erros as well.

mpalat:
I have a WIP fix for this, but that's causing some regressions - investigating

srikanth-sankaran:
Sorry, what is ME ??

PR #1228, opened 2023-07-12
[20] Switch statement with record pattern illegally reported as non-exhaustive

jarthana:
## What it does
Fixes Issue https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1224

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/CaseStatement.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/CaseStatement.java
@@ -502,15 +502,18 @@ private Constant resolveConstantExpression(BlockScope scope,
 					return Constant.NotAConstant;
 				}
 			}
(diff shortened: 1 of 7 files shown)


jarthana:
Although the patch takes care of the incorrect error reporting, it falls short with respect to the generated code. Code generation part assumes that we can't have a total pattern and an explicit default case. While this is true for simple type patterns, record patterns can't be unconditional and therefore we need to generate an implicit default case (I presume at this point) and generate a MatchException. I will wait and see how #1099 shapes up and take this up and again.

jarthana:
> Although the patch takes care of the incorrect error reporting, it falls short with respect to the generated code. Code generation part assumes that we can't have a total pattern and an explicit default case. While this is true for simple type patterns, we can have explicit default with total record patterns. Where we lack an explicit default, we need to generate an implicit default case (I presume at this point) and generate a MatchException if required. I will wait and see how #1099 shapes up and take this up and again.

I have commented out the test that fails. We will take it up with a different issue soon.

PR #1246, opened 2023-07-25
[performance] ClasspathJar: query length & lastModified together

jukzi:


jukzi:
@rgrunber 
mind to measure?
Similar to https://github.com/eclipse-jdt/eclipse.jdt.core/pull/299#issuecomment-1646155376
may highly depend on the OS/Filesystem used. On Win with some virus scanner where file access is very slow it makes a difference to make a system call only once or twice.

rgrunber:
I checked the amount of time the "Java indexing" thread was running, and that was consistently around ~30s on a project where I made sure the workspace index was empty beforehand. `findPackageSet` was around 100-200ms and I couldn't detect `lastModified`. I think this just means the Java indexing is probably not the task that would benefit the most from this, but other workflows might. If we're reducing the amount of file access that should be a step in the right direction.

Issue #1235, opened 2023-07-14
ConcurrentModificationException in JavaModelManager.secondaryTypesRemoving


jukzi:
```
java.util.ConcurrentModificationException
	at java.base/java.util.HashMap$HashIterator.nextNode(HashMap.java:1597)
	at java.base/java.util.HashMap$EntryIterator.next(HashMap.java:1630)
	at java.base/java.util.HashMap$EntryIterator.next(HashMap.java:1628)
	at org.eclipse.jdt.internal.core.JavaModelManager.secondaryTypesRemoving(JavaModelManager.java:5120)
	at org.eclipse.jdt.internal.core.JavaModelManager.secondaryTypesMerging(JavaModelManager.java:4935)
	at org.eclipse.jdt.internal.core.JavaModelManager.secondaryTypes(JavaModelManager.java:4897)
	at org.eclipse.jdt.internal.core.NameLookup.findSecondaryType(NameLookup.java:712)
	at org.eclipse.jdt.internal.core.NameLookup.findType(NameLookup.java:1000)
	at org.eclipse.jdt.internal.core.NameLookup.findType(NameLookup.java:823)
	at org.eclipse.jdt.internal.core.NameLookup.findType(NameLookup.java:742)
	at org.eclipse.jdt.internal.core.SearchableEnvironment.find(SearchableEnvironment.java:188)
	at org.eclipse.jdt.internal.core.SearchableEnvironment.findType(SearchableEnvironment.java:542)
	at org.eclipse.jdt.internal.compiler.lookup.LookupEnvironment.fromSplitPackageOrOracle(LookupEnvironment.java:428)
	at org.eclipse.jdt.internal.compiler.lookup.LookupEnvironment.lambda$1(LookupEnvironment.java:301)
	at org.eclipse.jdt.internal.compiler.lookup.LookupEnvironment.askForTypeFromModules(LookupEnvironment.java:395)
	at org.eclipse.jdt.internal.compiler.lookup.LookupEnvironment.askForType(LookupEnvironment.java:300)
	at org.eclipse.jdt.internal.compiler.lookup.PackageBinding.getType(PackageBinding.java:195)
	at org.eclipse.jdt.internal.compiler.lookup.PackageBinding.isViewedAsDeprecated(PackageBinding.java:314)
	at org.eclipse.jdt.internal.compiler.lookup.ReferenceBinding.isViewedAsDeprecated(ReferenceBinding.java:1799)
[...]

PR #1249, opened 2023-07-26
SecondaryTypes cache refactoring and MT fix for JavaModelManager

iloveeclipse:
Some history research to explain how that code works / evolved.

JavaModelManager got a "secondary types" cache used to store data for the "secondary" types defined in a Java file.

See first implementation
- https://bugs.eclipse.org/bugs/show_bug.cgi?id=36032
- 2d9dc5098273e0e95832d81d735d2d77249738ef

This comment explains the theory of operations:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=36032#c35

Quote (still valid):

> Solution was to use a cache on JavaModelManager per project information. Secondary types are searched the first time using a specific BasicSearchEngine request and corresponding IType stored in the cache. Then, after the consumed time is only to look in maps to retrieve the key.

> Cache size is currently not considered as a potential problem due to the fact that secondary types number should not be too big. Cache is reset each time a secondary type entry is added to the index and type is removed from cache each time its file is deleted.

Big fix for stale cache issue
- https://bugs.eclipse.org/bugs/show_bug.cgi?id=118823
- 04c08732703d5f5e97b66e89468d2f87b7f84cff

Third main change adding secondary types to IJavaProject.findType()
- https://bugs.eclipse.org/bugs/show_bug.cgi?id=118789#c13
- 8aee5680d034dfec291b1e7b2abb4d9d081e8e31

Quote:

> NameLookup findType public internal methods have been modified to consider secondary types or not as well as to wait for indexes or not. For obvious performance potential issues, all JDT/Core tools (eg. code assist, reconciler, etc.) consider secondary types now but do _not_ wait for indexes...

Initially the code wasn't MT safe at all. On bug 118823 the cache was changed to Hashtable (from HashMap) and commented that there were MT issues considered and that depending on indexer state caller may see not all secondary types.

At this point the cache was used by multiple threads but the code wasn't fully MT safe. Later changes added a bit more complexity but the main issue with unsafe MT code remained.

The main problem is that the indexing, updating and querying the cache can happen in parallel, and that while we are searching clients may already query for types or remove types from cache.

The original design managed that by holding a main map for "known secondary types" and second map (originally inside the first map) for "types being re-indexed" and depending on the call entry either of maps (or both) could be used / updated by multiple clients. That made the code extremely complex / error prone.

The main map is known as "secondaryTypes" and temporary map used to update first one is "indexedSecondaryTypes".

The "secondaryTypes" is the main cache for secondary types in the project and is what the clients are supposed to query. The "secondaryTypes" is originally filled in secondaryTypesSearching() (but can be updated later). The "indexedSecondaryTypes" is internal map and is used only for the main cache updates coming from secondaryTypeAdding() and secondaryTypesRemoving().

The second cache is "merged" to the first one secondaryTypesMerging().

An example for MT problems with two maps: the second "indexedSecondaryTypes" map was accessed in parallel on deleting files and while merging latest search results, resulting in https://bugs.eclipse.org/bugs/show_bug.cgi?id=567512 and refactoring in e59811a5cf43566d5300a91fc0b632e557a64348.

The current issue #1235 is coming from another MT problem around secondaryTypesMerging().

secondaryTypes() call was supposed to wait for indexer and merge found new secondary types into the main cache but had a problem that it could be entered by multiple threads in parallel and so the secondaryTypesMerging() code would run amok.

This patch tries to minimize possible MT issues accessing two caches.

- We move cache management into a dedicated SecondaryTypes class and combine two maps in a SecondaryTypesCache structure. This allows to get atomic access to both caches and also to synchronize two caches updates
code.
- The code in secondaryTypesSearching() updates the caches / returns the "known" cache from inside synchronized block.
- The entire work in secondaryTypesMerging() is done inside synchronized block.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1235

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java
@@ -1244,6 +1244,73 @@ public class JavaModelManager implements ISaveParticipant, IContentTypeChangeLis
 	 */
 	protected WeakHashMap<AbstractSearchScope, ?> searchScopes = new WeakHashMap<>();
 
+	/**
+	 * Current secondary types cache and temporary indexing cache
+	 *
+	 * @param secondaryTypes
+	 *            the main cache containing known secondary types in a project, can be null to indicate no search were
+	 *            done. Note, the cache might be not complete if the search is still running.
+	 * @param indexingSecondaryCache
+	 *            the temporary structure used while indexing, previously known as INDEXED_SECONDARY_TYPES entry. If it
+	 *            is null, no cache updates are running now
+	 */
+	private static record SecondaryTypesCache(Hashtable<String, Map<String, IType>> secondaryTypes,
+			Map<IFile, Map<String, Map<String, IType>>> indexingSecondaryCache) {
+
+		boolean isIndexingDone() {
+			return this.secondaryTypes != null && this.indexingSecondaryCache == null;
+		}
+	}
+
+	/**
+	 * Secondary types cache management code for a {@link PerProjectInfo}
+	 */
+	private static class SecondaryTypes {
+		private volatile SecondaryTypesCache cache;
+
+		public SecondaryTypes() {
+			this.cache = new SecondaryTypesCache(null, null);
+		}
+
+		SecondaryTypesCache cache() {
+			return this.cache;
+		}
+
+		private synchronized SecondaryTypesCache getOrCreateCache() {
+			Hashtable<String, Map<String, IType>> secondaryTypes = this.cache.secondaryTypes;
+			Map<IFile, Map<String, Map<String, IType>>> indexingSecondaryCache = this.cache.indexingSecondaryCache;
+			if (secondaryTypes != null && indexingSecondaryCache != null) {
+				return this.cache;
+			}
+			if (secondaryTypes == null) {
+				secondaryTypes = new Hashtable<>(3);
+			}
+			if (indexingSecondaryCache == null) {
+				indexingSecondaryCache = Collections.synchronizedMap(new HashMap<>(3));
+			}
+			this.cache = new SecondaryTypesCache(secondaryTypes, indexingSecondaryCache);
+			return this.cache;
+		}
+
+		private synchronized SecondaryTypesCache startIndexing() {
+			this.cache = new SecondaryTypesCache(this.cache.secondaryTypes(), Collections.synchronizedMap(new HashMap<>(3)));
+			return this.cache;
+		}
+
+		private synchronized void indexingDone() {
+			this.cache = new SecondaryTypesCache(this.cache.secondaryTypes(), null);
+		}
+
+		private synchronized SecondaryTypesCache doneSearching(Hashtable<String, Map<String, IType>> newSecondaryTypes) {
+			this.cache = new SecondaryTypesCache(newSecondaryTypes, this.cache.indexingSecondaryCache());
+			return this.cache;
+		}
+
+		private synchronized void clearAllCaches() {
+			this.cache = new SecondaryTypesCache(null, null);
+		}
+	}
+
 	public static class PerProjectInfo {
 		private static final int JAVADOC_CACHE_INITIAL_SIZE = 10;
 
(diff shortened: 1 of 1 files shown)


iloveeclipse:
@szarnekow, @cdietrich: would be nice if you could review.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java
```diff
@@ -5090,19 +5158,12 @@ public void secondaryTypesRemoving(IFile file, boolean cleanIndexCache) {
...
-				Entry<String, Map<String, IType>> entry = entries.next();
-				String qualifiedName = entry.getKey();
-				buffer.append(qualifiedName+':'+ entry.getValue());
-			}
-			buffer.append(',');
-			buffer.append(file.getFullPath());
-			buffer.append(')');
+			buffer.append(',').append(file.getFullPath()).append(')');
 			Util.verbose(buffer.toString());
+			dumpSecondaryTypes(secondaryTypesMap);
```
cdietrich, line 5166: order change intentional?

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java
```diff
@@ -4831,22 +4902,20 @@ public void secondaryTypeAdding(String path, char[] typeName, char[] packageName
...
+			buffer.append(project.getElementName()).append(',').append(waitForIndexes).append(')');
 			Util.verbose(buffer.toString());
 		}
 
 		// Return cache if not empty and there's no new secondary types created during indexing
 		final PerProjectInfo projectInfo = getPerProjectInfoCheckExistence(project.getProject());
-		Map<IFile, Map<String, Map<String, IType>>> indexingSecondaryCache = projectInfo.secondaryTypes == null ? null : projectInfo.indexingSecondaryCache;
-		if (projectInfo.secondaryTypes != null && indexingSecondaryCache == null) {
-			return projectInfo.secondaryTypes;
+		SecondaryTypesCache secondaryTypesCache = projectInfo.secondaryTypes.cache();
```
cdietrich, line 4911: i assume this and the other 3 calls can not happen without getOrCreate be called before?

iloveeclipse: which other 3 calls do you mean?
The NP check happens in the `if(secondaryTypesCache.isIndexingDone())` and below.
So if the secondaryTypes map isn't created yet, we will start search.

cdietrich: no i mean cache() can never return null

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java
```diff
@@ -1244,6 +1244,73 @@ public static IJavaElement determineIfOnClasspath(IResource resource, IJavaProje
...
+	 * Current secondary types cache and temporary indexing cache
+	 *
+	 * @param secondaryTypes
+	 *            the main cache containing known secondary types in a project, can be null to indicate no search were
+	 *            done. Note, the cache might be not complete if the search is still running.
+	 * @param indexingSecondaryCache
+	 *            the temporary structure used while indexing, previously known as INDEXED_SECONDARY_TYPES entry. If it
+	 *            is null, no cache updates are running now
+	 */
+	private static record SecondaryTypesCache(Hashtable<String, Map<String, IType>> secondaryTypes,
```
szarnekow, line 1257: I cannot seem to find how the nested maps are guarded against concurrent reads and modifications, but I’m traveling these days and only see the code on my phone. I’ll review in depth in two weeks time.

iloveeclipse: They don't, but the manipulations on two map *references* is now atomic (which was not the case before, and most of the logic was based on "if map 1 is null, do something with map 2" etc).

PR #1234, opened 2023-07-13
[WIP] Use some javac classes

rgrunber:
See https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1230 . @mickaelistria wishes to test the provided Jenkinsfile changes.

This requires to start the VM with ```
--add-opens java.base/java.lang=ALL-UNNAMED
--add-opens jdk.compiler/com.sun.tools.javac.jvm=ALL-UNNAMED --add-opens jdk.compiler/com.sun.tools.javac.main=ALL-UNNAMED --add-opens jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED --add-opens jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED --add-opens jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED
``` and other packages
settings.

Support for newer Java constructs requires to have that Java version as
BREE.

<!--
Thank you for your Pull Request. Please provide a description and review
the requirements below.

Contributors guide: https://github.com/eclipse-jdt/.github/blob/main/CONTRIBUTING.md

Note: Security vulnerabilities should not be disclosed on GitHub, through a PR or any
other means. TODO: how to report security issues
-->

## What it does
<!-- Include relevant issues and describe how they are addressed. -->

## How to test
<!-- Explain how a reviewer can reproduce a bug, test new functionality or verify performance improvements. -->

## Author checklist

- [ ] I have thoroughly tested my changes
- [ ] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [ ] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

PR #1230, opened 2023-07-12
[WIP] Use some javac classes

mickaelistria:
This requires to start the VM with ```
--add-opens java.base/java.lang=ALL-UNNAMED
--add-opens jdk.compiler/com.sun.tools.javac.jvm=ALL-UNNAMED --add-opens jdk.compiler/com.sun.tools.javac.main=ALL-UNNAMED --add-opens jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED --add-opens jdk.compiler/com.sun.tools.javac.tree=ALL-UNNAMED --add-opens jdk.compiler/com.sun.tools.javac.api=ALL-UNNAMED
``` and other packages
settings.

<!--
Thank you for your Pull Request. Please provide a description and review
the requirements below.

Contributors guide: https://github.com/eclipse-jdt/.github/blob/main/CONTRIBUTING.md

Note: Security vulnerabilities should not be disclosed on GitHub, through a PR or any
other means. TODO: how to report security issues
-->

## What it does
<!-- Include relevant issues and describe how they are addressed. -->

## How to test
<!-- Explain how a reviewer can reproduce a bug, test new functionality or verify performance improvements. -->

## Author checklist

- [ ] I have thoroughly tested my changes
- [ ] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [ ] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

Issue #420, opened 2022-09-19
Random failing InclusionPatternsTests.testMoveFolderContainingPackage1


jukzi:
https://github.com/eclipse-jdt/eclipse.jdt.core/pull/69
https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-69/16/

```
Unexpected deltas.
----------- Expected ------------
P[*]: {CHILDREN | CONTENT}\n
 src[*]: {CHILDREN | CONTENT}\n
  p1.p2[-]: {}\n
  ResourceDelta(/P/src/p1)[-]\n
 ResourceDelta(/P/p1)[+]
------------ but was ------------
P[*]: {CHILDREN}\n
 src[*]: {CHILDREN | CONTENT}\n
  p1.p2[+]: {}\n
  ResourceDelta(/P/src/p1)[+]\n
\n
P[*]: {CHILDREN | CONTENT}\n
 src[*]: {CHILDREN | CONTENT}\n
  p1.p2[-]: {}\n
  ResourceDelta(/P/src/p1)[-]\n
 ResourceDelta(/P/p1)[+]
--------- Difference is ----------
 expected:<P[*]: {CHILDREN[] | CONTENT}\n
 src[*...> but was:<P[*]: {CHILDREN[}\n
 src[*]: {CHILDREN | CONTENT}\n
  p1.p2[+]: {}\n
  ResourceDelta(/P/src/p1)[+]\n
\n
P[*]: {CHILDREN] | CONTENT}\n
 src[*...>
```

stephan-herrmann:
I've seen similar failures in InclusionPatternsTests.testNestedSourceFolder2()

This is what I collected from debugging:

Given this sequence of calls in `testNestedSourceFolder2()`:
```
	setClasspath(..);
	createFolder(..);
	clearDeltas();
	createFile(..);
	assertDeltas(..);
```
* the assert expects the deltas from `createFile()`, only
* we randomly see the deltas from `setClasspath()` instead
* this indicates that the deltas created during `setClasspath()` arrive at the `deltaListener` only **after** `clearDeltas()`.
* recorded deltas are broadcast from `Workspace.endOperation()` -> `broadcastPostChange()`.
* in the call tree below `setClasspath()` broadcasting happens synchronuously, so if only thread `main` would be running, we'd have the guarantee, that deltas have been delivered before the end of `setClasspath()`.
[...]

PR #1254, opened 2023-07-28
Create ci.yml to build with GitHub Actions

mickaelistria:
## What it does

Adds GitHub action that validates branches and PRs.
Compared to Jenkinsfile, this workflow can easily be used to enable validation in forks, just by enabling GitHub actions on the forked repository. So it makes it almost free for contributors to get some CI validation, and thus to provide higher quality PRs more easily.

## Author checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,8 +6,6 @@ concurrency:
 on:
   push:
     branches: '**'
-  pull_request:
-    branches: [ master ]
 
 jobs:
   build:
@@ -30,7 +28,7 @@ jobs:
           JavaSE-20
         distribution: 'temurin'
     - name: Set up Maven
-      uses: stCarolas/setup-maven@v4.5
+      uses: stCarolas/setup-maven@07fbbe97d97ef44336b7382563d66743297e442f # v4.5
       with:
         maven-version: 3.9.3
     - name: Build with Maven 🏗️
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,38 @@
+name: Continuous Integration
+concurrency: 
+    group: ${{ github.workflow }}-${{ github.ref }}
+    cancel-in-progress: true
+
+on:
+  push:
+    branches: '**'
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: recursive
+        fetch-depth: 0
+    - name: Set up JDKs ☕
+      uses: actions/setup-java@v3
+      with:
+        java-version: |
+          8
+          17
+          20
+        mvn-toolchain-id: |
+          JavaSE-1.8
+          JavaSE-17
+          JavaSE-20
+        distribution: 'temurin'
+    - name: Set up Maven
+      uses: stCarolas/setup-maven@v4.5
+      with:
+        maven-version: 3.9.3
+    - name: Build with Maven 🏗️
+      run: |
+        mvn clean install -f org.eclipse.jdt.core.compiler.batch -DlocalEcjVersion=99.99 && mvn -U clean verify --batch-mode --fail-at-end -Ptest-on-javase-20 -Pbree-libs -Papi-check -Djava.io.tmpdir=$WORKSPACE/tmp -Dproject.build.sourceEncoding=UTF-8  -Dtycho.surefire.argLine="--add-modules ALL-SYSTEM -Dcompliance=1.8,11,17,20 -Djdt.performance.asserts=disabled" -Dcbi-ecj-version=99.99
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,4 +33,5 @@ jobs:
         maven-version: 3.9.3
     - name: Build with Maven 🏗️
       run: |
-        mvn clean install -f org.eclipse.jdt.core.compiler.batch -DlocalEcjVersion=99.99 && mvn -U clean verify --batch-mode --fail-at-end -Ptest-on-javase-20 -Pbree-libs -Papi-check -Djava.io.tmpdir=$WORKSPACE/tmp -Dproject.build.sourceEncoding=UTF-8  -Dtycho.surefire.argLine="--add-modules ALL-SYSTEM -Dcompliance=1.8,11,17,20 -Djdt.performance.asserts=disabled" -Dcbi-ecj-version=99.99
+        mvn clean install --batch-mode -f org.eclipse.jdt.core.compiler.batch -DlocalEcjVersion=99.99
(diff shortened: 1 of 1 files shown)


mickaelistria:
So the test indeed seems flaky, as it was already reported in https://github.com/eclipse-jdt/eclipse.jdt.core/issues/420#issuecomment-1613534520

mickaelistria:
@rgrunber If this is merged, then I think we can close #1234 and keep only #1230 and my local build on GitHub actions should be enough to demo what works or not.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,38 @@
...
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: recursive
+        fetch-depth: 0
+    - name: Set up JDKs ☕
+      uses: actions/setup-java@v3
+      with:
+        java-version: |
+          8
```
iloveeclipse, line 24: Does it mean, ecj will run on Java 8? That is not possible.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,38 @@
...
+          17
+          20
+        mvn-toolchain-id: |
+          JavaSE-1.8
```
iloveeclipse, line 28: Here also Java 8?

.github/workflows/ci.yml
mickaelistria: No, it just makes Java 8 will be available on the CI machine.

.github/workflows/ci.yml
mickaelistria: No, this is a specific Tycho/OSGi label here: tycho wants a toolchain that has the BREE name as id to properly resolve it, and for Java 8, it's JavaSE-1.8. See https://github.com/eclipse-tycho/tycho/issues/2670 .
Note that this particular workaround is necessary for the older org.eclipse.jdt.annotations bundle only.

.github/workflows/ci.yml
iloveeclipse: ... and why do we need Java 8 on CI machine?

mickaelistria: Tycho requires Java 8 to properly resolve Java 8 APIs. It can be turned off and the current Java runtime can be used instead (Java 17), but we've seen cases where some API disappear in newer versions and thus the code targetting Java 8 that cannot be compiled any longer by Java 17.
IIRC, Java 8 is also installed on Jenkins and configured in java toolchains.

iloveeclipse: > the code targetting Java 8 that cannot be compiled any longer by Java 17.

We do not have code targeting Java 8 AFAIK, and annotations don't use any Java 8 specific API's, so why all this overhead adding something we don't need?

mickaelistria: You need to ensure compatibility with Java 8 if you target Java 8 (in https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/org.eclipse.jdt.annotation_v1/META-INF/MANIFEST.MF#L8 ). Compatibility cannot be verified at build-time without a Java 8 installation being visible to Tycho.
The build config here is similar to Jenkins one. I wouldn't make it a requirement to chnage the build config to something too different from "main" build.
If you wish to get rid of the requirement of having Java 8 installed for build to succeed, please open another issue; it can be changed from pom.xml, but this would also affect the Jenkins build. It's another topic.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,38 @@
+name: Continuous Integration
+concurrency: 
+    group: ${{ github.workflow }}-${{ github.ref }}
+    cancel-in-progress: true
+
+on:
+  push:
+    branches: '**'
+  pull_request:
+    branches: [ master ]
```
rgrunber, line 10: So the idea is to have users verifying PRs on their own quota and not ours :stuck_out_tongue_winking_eye: Do we need the `pull_request` at all ? Are users really creating pull request against their own fork ? I would think the most common workflow is just to push on any branch in their repo.

This would be duplicating what's done on https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/master/ though right ?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,38 @@
...
+          8
+          17
+          20
+        mvn-toolchain-id: |
+          JavaSE-1.8
+          JavaSE-17
+          JavaSE-20
+        distribution: 'temurin'
+    - name: Set up Maven
+      uses: stCarolas/setup-maven@v4.5
```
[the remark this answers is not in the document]
mickaelistria: Ack, I will try to change it to the commit SHA.

.github/workflows/ci.yml
mickaelistria: > Do we need the pull_request at all ?

Indeed, I don't think we need it. I'll remove it.

> Are users really creating pull request against their own fork ?

If they want, they can change this file on their fork.

> This would be duplicating what's done on https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/master/ though right ?

Yes, but I think it's a good way to ensure this build works; and I hope it wouldn't be annoying.

rgrunber: If it does become annoying (or the verification ends up taking longer, or conflicting results), I think you might be able to use `$GITHUB_REPOSITORY_OWNER` (should be eclipse-jdt here) to distinguish between fork/non-fork. It'd have to be used after the job is already triggered so not the cleanest solution.

mickaelistria:
@rgrunber Anything left that prevents from merging here?

rgrunber (approved):
If there's no other opposition, I'll merge this ~~soon~~ after the freeze ?

@mickaelistria , wait I assume branch freeze policy still applies so we're waiting until 4.30..

akurtakov:
Releng improvements (like this one) are in general not under freeze together with docs/examples/tests changes.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,36 @@
...
+          JavaSE-17
+          JavaSE-20
+        distribution: 'temurin'
+    - name: Set up Maven
+      uses: stCarolas/setup-maven@07fbbe97d97ef44336b7382563d66743297e442f # v4.5
+      with:
+        maven-version: 3.9.3
+    - name: Build with Maven 🏗️
+      run: |
+        mvn clean install -f org.eclipse.jdt.core.compiler.batch -DlocalEcjVersion=99.99 && mvn -U clean verify --batch-mode --fail-at-end -Ptest-on-javase-20 -Pbree-libs -Papi-check -Djava.io.tmpdir=$WORKSPACE/tmp -Dproject.build.sourceEncoding=UTF-8  -Dtycho.surefire.argLine="--add-modules ALL-SYSTEM -Dcompliance=1.8,11,17,20 -Djdt.performance.asserts=disabled" -Dcbi-ecj-version=99.99
```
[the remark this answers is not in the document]
mickaelistria: Ack, I'll change all that.

mickaelistria:
Thank you! That will be pretty useful for people willing to experiment with JDT to have CI enabled out-of-the-box in their fork!

stephan-herrmann:
@mickaelistria Seems you're not done here, see https://www.eclipse.org/lists/jdt-dev/msg02287.html

mickaelistria:
Thanks for the heads-up and sorry for this inconvenience. I hope https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1389 fixes it.

Issue #316, opened 2022-08-16
IllegalArgumentException for ParenthesizedExpression on Assignment LHS


lucach:
The following exception is thrown with the latest version of `org.eclipse.jdt.core` (3.30.0) when requesting to create an AST containing a ParenthesizedExpression in the left-hand side of an Assignment.

```
Status ERROR: org.eclipse.jdt.core code=4 Exception occurred during compilation unit conversion:
----------------------------------- SOURCE BEGIN -------------------------------------
class C {
    void m() { (1+2) = 3; }
}

----------------------------------- SOURCE END ------------------------------------- java.lang.IllegalArgumentException
java.lang.IllegalArgumentException
	at org.eclipse.jdt.core.dom.InfixExpression.setOperator(InfixExpression.java:383)
	at org.eclipse.jdt.core.dom.ASTConverter.convert(ASTConverter.java:1287)
	at org.eclipse.jdt.core.dom.ASTConverter.convert(ASTConverter.java:2047)
	at org.eclipse.jdt.core.dom.ASTConverter.convertToParenthesizedExpression(ASTConverter.java:3853)
	at org.eclipse.jdt.core.dom.ASTConverter.convert(ASTConverter.java:1953)
	at org.eclipse.jdt.core.dom.ASTConverter.convert(ASTConverter.java:1204)
	at org.eclipse.jdt.core.dom.ASTConverter.convert(ASTConverter.java:1990)
	at org.eclipse.jdt.core.dom.ASTConverter.convert(ASTConverter.java:3095)
	at org.eclipse.jdt.core.dom.ASTConverter.convert(ASTConverter.java:751)
	at org.eclipse.jdt.core.dom.ASTConverter.buildBodyDeclarations(ASTConverter.java:208)
	at org.eclipse.jdt.core.dom.ASTConverter.convert(ASTConverter.java:3390)
	at org.eclipse.jdt.core.dom.ASTConverter.convert(ASTConverter.java:1534)
	at org.eclipse.jdt.core.dom.CompilationUnitResolver.resolve(CompilationUnitResolver.java:1136)
	at org.eclipse.jdt.core.dom.CompilationUnitResolver.resolve(CompilationUnitResolver.java:739)
	at org.eclipse.jdt.core.dom.ASTParser.createASTs(ASTParser.java:1049)
```

iloveeclipse:
Is this a recent  regression? Could you bisect to the commit caused this?

[...]

PR #1280, opened 2023-08-11
fix IllegalArgumentException in InfixExpression on LHS #316

jukzi:
A "+" Operator (OperatorId 14) was using
OperatorID 14+32 (+ the IsStrictlyAssigned Bit)

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/316

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/OperatorIds.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/OperatorIds.java
@@ -14,6 +14,7 @@
 package org.eclipse.jdt.internal.compiler.ast;
 
 public interface OperatorIds {
+	/**  org.eclipse.jdt.core.dom.InfixExpression */
 	public static final int AND_AND = 0;
 	public static final int OR_OR = 1;
 	public static final int AND = 2;
@@ -33,14 +34,14 @@ public interface OperatorIds {
 	public static final int REMAINDER = 16;
 	public static final int RIGHT_SHIFT = 17;
 	public static final int EQUAL_EQUAL = 18;
-	public static final int UNSIGNED_RIGHT_SHIFT= 19;
-	public static final int NumberOfTables = 20;
+	public static final int UNSIGNED_RIGHT_SHIFT= 19; // last org.eclipse.jdt.internal.compiler.ast.OperatorExpression
 
-	public static final int QUESTIONCOLON = 21;
-
-	public static final int NOT_EQUAL = 22;
-	public static final int EQUAL = 23;
-	public static final int INSTANCEOF = 24;
-	public static final int PLUS_PLUS = 25;
-	public static final int MINUS_MINUS = 26;
+	public static final int NOT_EQUAL = 20;
+	public static final int EQUAL = 21;
+	/** others */
+	public static final int QUESTIONCOLON = 22;
+	public static final int INSTANCEOF = 23;
+	/** postfix */
+	public static final int PLUS_PLUS = 24;
+	public static final int MINUS_MINUS = 25;
 }
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/OperatorIds.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/OperatorIds.java
@@ -36,11 +36,11 @@ public interface OperatorIds {
 	public static final int UNSIGNED_RIGHT_SHIFT= 19;
 	public static final int NumberOfTables = 20;
 
-	public static final int QUESTIONCOLON = 23;
+	public static final int QUESTIONCOLON = 21;
 
-	public static final int NOT_EQUAL = 29;
-	public static final int EQUAL = 30;
-	public static final int INSTANCEOF = 31;
-	public static final int PLUS_PLUS = 32;
-	public static final int MINUS_MINUS = 33;
+	public static final int NOT_EQUAL = 22;
+	public static final int EQUAL = 23;
+	public static final int INSTANCEOF = 24;
+	public static final int PLUS_PLUS = 25;
+	public static final int MINUS_MINUS = 26;
 }
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/BatchCompilerTest.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/BatchCompilerTest.java
@@ -13469,4 +13469,53 @@ public void testIssue89_5() {
 				"error: warnings found and -failOnWarning specified\n",
 				true);
 }
+
+public void testGitHub316(){
+	this.runNegativeTest(
+		new String[] {
+			"X.java",
+			"public class X {\n" +
+					"void m() { (1+2) = 3; }"+// was IllegalArgumentException in InfixExpression#setOperator
+			"}"},
+        "\"" + OUTPUT_DIR +  File.separator +
+        	"X.java\"",
+		"",
+		"----------\n"
+		+ "1. ERROR in ---OUTPUT_DIR_PLACEHOLDER---/X.java (at line 2)\n"
+		+ "	void m() { (1+2) = 3; }}\n"
+		+ "	           ^^^^^\n"
+		+ "The left-hand side of an assignment must be a variable\n"
+		+ "----------\n"
+		+ "1 problem (1 error)\n"
+		+ "",
+		true);
+}
+public void testGitHub1122(){
+	this.runNegativeTest(
+		new String[] {
+			"X.java",
+			"public class X {\n" +
+					"void m() {"
+					+ ".a() >0);"// was IllegalArgumentException in InfixExpression#setOperator
+					+ " }"+
+			"}"},
+        "\"" + OUTPUT_DIR +  File.separator +
+        	"X.java\"",
+		"",
+		"----------\n"
+		+ "1. ERROR in ---OUTPUT_DIR_PLACEHOLDER---/X.java (at line 2)\n"
+		+ "	void m() {.a() >0); }}\n"
+		+ "	          ^\n"
+		+ "Syntax error on token \".\", invalid (\n"
+		+ "----------\n"
+		+ "2. ERROR in ---OUTPUT_DIR_PLACEHOLDER---/X.java (at line 2)\n"
+		+ "	void m() {.a() >0); }}\n"
+		+ "	                 ^\n"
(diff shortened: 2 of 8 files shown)


iloveeclipse:
@srikanth-sankaran : is this an area where you could help reviewing?

srikanth-sankaran:
Yes, I will review this on Monday.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/OperatorIds.java
```diff
@@ -36,11 +36,11 @@ public interface OperatorIds {
...
-	public static final int EQUAL = 30;
-	public static final int INSTANCEOF = 31;
-	public static final int PLUS_PLUS = 32;
-	public static final int MINUS_MINUS = 33;
+	public static final int NOT_EQUAL = 22;
+	public static final int EQUAL = 23;
+	public static final int INSTANCEOF = 24;
+	public static final int PLUS_PLUS = 25;
+	public static final int MINUS_MINUS = 26;
 }
```
srikanth-sankaran, line 46: So, I am nervous here - These numerical assignments date back to almost two decades ago. Are you able to ascertain why the holes were originally left in ? Is that to maintain some kind of mapping ? If so is such mapping no longer relevant ?? What made it irrelevant ?

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/dom/StandAloneASTParserTest.java
```diff
@@ -1881,4 +1881,50 @@ public void testBug558517() throws IOException {
...
+			parser.setBindingsRecovery(true);
+
+			ASTNode node = parser.createAST(null);
+			assertTrue("Should be a compilation unit", node instanceof CompilationUnit);
+			CompilationUnit cu = (CompilationUnit) node;
+			assertEquals("Problems in compilation", 3,cu.getProblems().length);
+			assertEquals("Syntax error on token \".\", invalid (",cu.getProblems()[0].getMessage());
+			// XXX BatchCompiler instead reports 'Syntax error, insert "AssignmentOperator Expression" to complete Expression':
+			assertEquals("The left-hand side of an assignment must be a variable",cu.getProblems()[1].getMessage());
+	}
```
srikanth-sankaran, line 1929: Could you please also add the various tests from https://bugs.eclipse.org/bugs/show_bug.cgi?id=568629 to make sure those duplicates are covered properly by the fix ??

srikanth-sankaran (changes requested):

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/BatchCompilerTest.java
```diff
@@ -13469,4 +13469,53 @@ public void testIssue89_5() {
...
+		+ "	void m() {.a() >0); }}\n"
+		+ "	                 ^\n"
+		// XXX ASTParser instead reports "The left-hand side of an assignment must be a variable":
+		+ "Syntax error, insert \"AssignmentOperator Expression\" to complete Expression\n"
+		+ "----------\n"
+		+ "2 problems (2 errors)\n"
+		+ "",
+		true);
+}
+}
```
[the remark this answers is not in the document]
jukzi: I agree they are irrelevant for this fix, but they will prevent the same regression on the other compiler. Currently there is no test for these cases.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/dom/StandAloneASTParserTest.java
```diff
@@ -1881,4 +1881,50 @@ public void testBug558517() throws IOException {
 			packDir.delete();
 		}
 	}
+
+	public void testGitHub316() throws JavaModelException {
+		String contents =
+				"public class X {\n" +
+						"void m() {"
+//						+ " 1 = 3;"// did work
+						+ " (1+2) = 3;"// was IllegalArgumentException in InfixExpression#setOperator
```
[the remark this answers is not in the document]
jukzi: ok

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/OperatorIds.java
jukzi: I don't see any reason why there have been those gaps - everything seem to work without those.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/dom/StandAloneASTParserTest.java
jukzi: i added testBug568629

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/OperatorIds.java
szarnekow: Please note that there's also https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/org.eclipse.jdt.core.tests.model/workspace/Compiler/src/org/eclipse/jdt/internal/compiler/ast/OperatorIds.java
If you plan to do some perceived cleanup on the IDs, I suggest to do it in a separate changeset.

jukzi: @szarnekow i don't get what you are trying to say here. that ProfilingASTConvertionTest seems to be unused and totally unrelated to this topic.

srikanth-sankaran:
@jukzi I see a bunch of failures now in the latest run with unknown operator ... Could you please investigate them ?

jukzi:
> @jukzi I see a bunch of failures now in the latest run with unknown operator ... Could you please investigate them ?

It feels like that the Contstants are inlined in a bundle that is not updated during rebuild.  I push the exact same patch again to
https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1286 and guess that will suceeed. Maybe @iloveeclipse knows more why that happens.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/BatchCompilerTest.java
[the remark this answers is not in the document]
srikanth-sankaran: > I agree they are irrelevant for this fix, but they will prevent the same regression on the other compiler. Currently there is no test for these cases.

Actually, the problem is very much there on compiler AST side also although these tests don't bring it out to the fore. Anyways, I support retaining these tests

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/OperatorIds.java
srikanth-sankaran: > Please note that there's also https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/org.eclipse.jdt.core.tests.model/workspace/Compiler/src/org/eclipse/jdt/internal/compiler/ast/OperatorIds.java If you plan to do some perceived cleanup on the IDs, I suggest to do it in a separate changeset.

AFAICT, this is a snapshot - which need not be kept upto date - @szarnekow Do you think otherwise ?

srikanth-sankaran:
> > @jukzi I see a bunch of failures now in the latest run with unknown operator ... Could you please investigate them ?
> 
> It feels like that the Contstants are inlined in a bundle that is not updated during rebuild. I push the exact same patch again to #1286 and guess that will suceeed. Maybe @iloveeclipse knows more why that happens.

It is failing there too. Can you locally run/debug and see

jukzi:
> It is failing there too. Can you locally run/debug and see

locally everything is fine.

jukzi:
> On a first look I don't see any material difference

before "unsafe cast (cast expression)" bit was conflicting with operator-bits. I guess that was a problem too.

srikanth-sankaran:
> > On a first look I don't see any material difference
> 
> before "unsafe cast (cast expression)" bit was conflicting with operator-bits. I guess that was a problem too.

OK, That have had something to do with the bit reassignment. I see that if backout Jay's changes a test in GTT fails by emitting an extra diagnostic

```
4. ERROR in X.java (at line 11)\n
	} else 	if (t instanceof T) {\n
	       	    ^\n
Type T cannot be safely cast to T\n
----------\n

```

So let us ignore the hypothetical scenario of rolling back Jay's changes in ASTNode.java for Bug 562392

Coming to current patch:

That tests pass with the reassignments offers some degree of comfort but I would have felt a lot better if we could have unearthed the reason for those gaps. I don't think they were arbitrary. 

So, if the mystery of the 15 failing tests is explained, this is OK to go from my side, but would also prefer an OK from @jarthana

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/OperatorIds.java
szarnekow: Ok, understood. You're saying that the OperatorIds from the referenced test is only accidentally identical with OperatorIds from the production code. If that's the case, I'm fine with it being no longer the same.

srikanth-sankaran: > Ok, understood. You're saying that the OperatorIds from the referenced test is only accidentally identical with OperatorIds from the production code. If that's the case, I'm fine with it being no longer the same.

I don't think they are accidentally identical. I think it is a snapshot of the world as it was once upon a time but is not meant to be kept in sync. 

I don't have my desktop right now to confirm. But I would expect many things to be out of sync with main code base.

jukzi:
> So, if the mystery of the 15 failing tests is explained,

I added code in jdt.core (ASTConverter) and now those errors are gone: https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-1286/4/

iloveeclipse:
> I guess these are Random failing tests?
> InclusionPatternsTests.testMoveFolderContainingPackage1
> NullReferenceTest.testBug325229a
> PatternMatching16Test.testBug574906

InclusionPatternsTests.testMoveFolderContainingPackage1 yes. Others two: not sure.

srikanth-sankaran:
(Tuesday is a holiday in India. Jay make look at this likely only on Wednesday)

Issue #1261, opened 2023-08-02
Eclipse compiler not inferring lambda parameter type in Eclipse EE 2023-09 M1


garretwilson:
I'm using OpenJDK 17, although this code is set to use Java 8. I upgraded to Eclipse EE 2023-09 M1, and suddenly the Eclipse compiler doesn't like Java 8 code that has been working for a long time.

The code is in the https://github.com/globalmentor/globalmentor-base project. It relies on the https://github.com/globalmentor/globalmentor-root POM. You can install both of those with Maven simply by cloning them from the repository and running `mvn install`. (Install the `globalmentor-root` project first.)

The code that is breaking is [`MediaTypeTest`](https://github.com/globalmentor/globalmentor-base/blob/main/core/src/test/java/com/globalmentor/net/MediaTypeTest.java) in the method `testParameterAppendValueTo()`:

```java
/** @see MediaType.Parameter#appendValueTo(Appendable, String) */
@Test
public void testParameterAppendValueTo() throws IOException {
  PARAMETER_VALUE_STRINGS_BY_VALUE.forEach(throwingBiConsumer((value, valueString) -> {
    assertThat(MediaType.Parameter.appendValueTo(new StringBuilder(), value).toString(), is(valueString));
  }));
}
```

This tests the [`MediaType`](https://github.com/globalmentor/globalmentor-base/blob/main/core/src/main/java/com/globalmentor/net/MediaType.java) class.

`PARAMETER_VALUE_STRINGS_BY_VALUE` is a simple `Map<String, String>`. `throwingBiConsumer` is part of the [Faux Pas](https://github.com/zalando/faux-pas) library, which we discussed in #1162. It uses "sneaky throws" to allow exceptions to propagate even though they are not declared.

`MediaType.Parameter.appendValueTo()` looks like this:

```java
public static <A extends Appendable> A appendValueTo(@Nonnull final A appendable, @Nonnull CharSequence parameterValue) throws IOException
```

Thus in the code above, we know that `value` is the key in the `Map<String, String>`. We know that `new StringBuilder()` is of course an `Appendable`. Thus `(StringBuilder, String)` should be fine to pass to `(Appenable, CharSequence`). And in fact this worked fine in Eclipse EE 2023-06 and for many, many versions before that. But in Eclipse EE 2023-09 we get>

[...]

PR #1281, opened 2023-08-11
Eclipse compiler not inferring lambda parameter type in Eclipse EE 2023-09 M1 #1261

stephan-herrmann:
## What it does

Avoid re-using an inference context that was built with information from a premature overload resolution (i.e., overload resolution based on improper argument types).

## How to test

JUnit test is included.

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
@@ -988,6 +988,9 @@ public class LambdaExpression extends FunctionalExpression implements IPolyExpre
 				copy = copy();
 				if (copy == null)
 					throw new CopyFailureException();
+				if (InferenceContext18.DEBUG) {
+					System.out.println("Copy lambda "+this+" for target "+targetType.debugName()); //$NON-NLS-1$ //$NON-NLS-2$
+				}
 
 				copy.setExpressionContext(this.expressionContext);
 				copy.setExpectedType(targetType);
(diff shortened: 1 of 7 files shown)


stephan-herrmann:
The first commit only brings a bunch of debug prints, useful for delta debugging in case of a regression. This is controlled by `InferenceContext18.DEBUG`, which is constant `false` in production.

stephan-herrmann:
@srikanth-sankaran I'd like to get a second opinion here.

The core problem in #1261 is, that a first round of inference performs overload resolution for invocation `is(valueString)` **while the argument is still an ivar**. Contrary to my comment in #1261 this is not a problem per se (JLS 18.5.4 doesn't seem to have an issue with this), but we need to be sure that any type bounds based on this overload resolution will be discarded for the final round of inference.

Most of the code in this area (around `registerInferenceContext()`)  is _not_ backed by JLS:
* in [bug 511252 c17](https://bugs.eclipse.org/bugs/show_bug.cgi?id=511252#c17) we concluded that inference contexts should not be re-used if they were built based on improper argument types.  In fact we received a helpful hint from Maurizio for this one.
* in [#506](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/506) we found that discarding inference contexts broke other things and @testforstephen invented a different mechanism where instead of discarding context, we defer certain steps of inference until the final round (below `ASTNode.resolvePolyExpressionArguments()`).
* then here ([#1261](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1261)) we encounter another situation where context must not be re-used.

I readily admit, that the newest tweak here is an opportunistic attempt at fixing one particular test case. Is this opportunistic fix legitimate? What's even harder to see: can the situations of the three bugs listed above possibly interact, i.e., do we need to specifically handle combinations of reasons to register / not register an inference context for later re-use?

srikanth-sankaran:
@stephan-herrmann I have glanced through the changes - unfortunately I am still significantly out of touch with this part of the compiler and also the associated the spec to offer meaningful feedback.

We can ask @jarthana for some "stress testing" of this change set.

stephan-herrmann:
>  I have glanced through the changes

I appreciate ... 

> We can ask @jarthana for some "stress testing" of this change set.

Surely a good thing to do. Or should we just merge the changes to try to get community feedback before the next release?

srikanth-sankaran:
> > We can ask @jarthana for some "stress testing" of this change set.
> 
> Surely a good thing to do. Or should we just merge the changes to try to get community feedback before the next release?

Yes, these can happen in parallel

stephan-herrmann:
Hint to @jarthana for stress testing: replacing the invocation `is(valueString)` with other sub-kinds of `InvocationSite` with a similar overload situation might prove my fix to be incomplete.

PR #1298, opened 2023-08-17
Support for quickfix as of eclipse.jdt.ui#1294 when using TYPE_USE annotations

stephan-herrmann:
## What it does

distinguish unchecked conversion (null-related) toplevel vs. detail (use different problemId, but same problem message).

## How to test
will be consumed by https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/724

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/core/compiler/IProblem.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/core/compiler/IProblem.java
@@ -1938,7 +1938,16 @@ void setSourceStart(int sourceStart);
 	int AnnotatedTypeArgumentToUnannotatedSuperHint = Internal + 984;
 	/** @since 3.32 */
 	int NonNullArrayContentNotInitialized = Internal + 985;
-	/** @since 3.35 */
+	/**
+	 * Both {@link #NullityUncheckedTypeAnnotationDetail} and {@link #NullityUncheckedTypeAnnotation}
+	 * signal that unchecked conversion is needed to pass a value between annotated and un-annotated code.
+	 * In the case of {@link #NullityUncheckedTypeAnnotationDetail} the mismatch was observed only on some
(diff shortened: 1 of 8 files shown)


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/core/compiler/IProblem.java
```diff
@@ -1877,7 +1877,7 @@ public interface IProblem {
 	/** @since 3.10 */
 	int NullityMismatchingTypeAnnotationSuperHint = Internal + 954;
 	/** @since 3.10 */
-	int NullityUncheckedTypeAnnotationDetail = Internal + 955;
+	int NullityUncheckedTypeAnnotationDetail = Internal + 955; // see also NullityUncheckedTypeAnnotation
```
gayanper, line 1880: Could we add a comment explaining here the different between this `NullityUncheckedTypeAnnotationDetail` and `NullityUncheckedTypeAnnotation`  ?

stephan-herrmann:
@gayanper what did your force-push change?

stephan-herrmann:
> @gayanper what did your force-push change?

Ok, I manually checked that the diff before and after the force push is the same. 

Note that just yesterday I rebased the PR, which is different in several ways:
* I was convinced that I am the only person looking at the PR at that time
* I explicitly requested github to perform the rebase, which is visible to everybody looking at the PR

In all other situations force push will simply take away time from reviewing and fixing bugs, because I need to manually figure out what as changed, as github does a very poor job for this (perhaps they didn't anticipate such free use of force-push?)

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/core/compiler/IProblem.java
stephan-herrmann: sure, thanks.

Issue #1307, opened 2023-08-24
2623 test fails in jdt.core.tests.model in I20230823-1800


iloveeclipse:
https://download.eclipse.org/eclipse/downloads/drops4/I20230823-1800/testresults/html/org.eclipse.jdt.core.tests.model_ep429I-unit-mac64-java17_macosx.cocoa.x86_64_17.html

`org.eclipse.jdt.core.tests.rewrite.modifying.ASTRewritingModifyingOtherTest`

```
Cannot invoke "java.util.Map.get(Object)" because "defaultOptionsMap" is null

java.lang.NullPointerException: Cannot invoke "java.util.Map.get(Object)" because "defaultOptionsMap" is null
at org.eclipse.jdt.internal.core.util.Util.fixTaskTags(Util.java:3298)
at org.eclipse.jdt.internal.core.JavaModelManager.setOptions(JavaModelManager.java:5394)
at org.eclipse.jdt.core.JavaCore.setOptions(JavaCore.java:6264)
at org.eclipse.jdt.core.tests.rewrite.modifying.ASTRewritingModifyingTest.tearDownSuite(ASTRewritingModifyingTest.java:96)
at org.eclipse.jdt.core.tests.model.SuiteOfTestCases$Suite$1.protect(SuiteOfTestCases.java:90)
at junit.framework.TestResult.runProtected(TestResult.java:142)
at org.eclipse.jdt.core.tests.model.SuiteOfTestCases$Suite.run(SuiteOfTestCases.java:95)
at junit.framework.TestSuite.runTest(TestSuite.java:241)
at junit.framework.TestSuite.run(TestSuite.java:236)
at junit.framework.TestSuite.runTest(TestSuite.java:241)
at junit.framework.TestSuite.run(TestSuite.java:236)
```

After that most likely lot of tests failed because proper cleanup was not done:

```
Resource '/P' already exists.

org.eclipse.core.internal.resources.ResourceException(/P)[374]: java.lang.Exception: Resource '/P' already exists.
at org.eclipse.core.internal.resources.ResourceException.provideStackTrace(ResourceException.java:42)
at org.eclipse.core.internal.resources.ResourceException.<init>(ResourceException.java:38)
at org.eclipse.core.internal.resources.Resource.checkDoesNotExist(Resource.java:346)
at org.eclipse.core.internal.resources.Resource.checkDoesNotExist(Resource.java:333)
[...]

PR #1308, opened 2023-08-24
NPE in JavaModelManager.setOptions()

iloveeclipse:
- JavaModelManager.setOptions() is supposed to reset options on null argument (see org.eclipse.jdt.core.JavaCore.setOptions() javadoc), but instead it failed with NPE.
- Made JavaModelManager.optionsCache volatile

See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1307


iloveeclipse:
@stephan-herrmann : because you were interested/involved in the "options" management before, does this change make sense for you? I wonder it was possible to pass null (for years) with the "expected" NPE result and no one complained so far considering also the javadoc was saying the options should be reset to defaults.

Issue #1301, opened 2023-08-21
No code completion on string switch case


gayanper:
In the following code

```
public class Mapper {
  public Integer mapJobState(String value) {
    return switch (value) {
      case 
    };
  }
}
```

Code completions behind `case M` doesn't provide any completions.

gayanper:
This is a Java17 and above issue

gayanper:
@stephan-herrmann The PR was made under the following assumptions

When we write a switch block with a int variable
```
switch(i) {
   case Const
}
```
completions behind `Const` will provide proposals where user can choose the class where they want to select a constant. if they don't do that, the compiler will issue a error.

Now I would expect the same for a switch with a string variable like

```
switch(str) {
   case Const
}
```
But due to the code segment at https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java#L3967
which cause the scenario with str variable not to work. because the `else` condition at https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java#L3980 cause only to have switch block related keyword completions.

So the question is should we handle this scenario ? I feel we should since it is working for primitive and why not for string ? And the next question is why only limiting few keywords as completions when there is a non primitive resolved type in Java17 ? shouldn't the pattern specific errors must be handled by compiler ? And if we really want to support choosing right types in java17+ pattern matching, should we implement it based on completion relevance instead ?

gayanper:
Now I trace down the issue where this new code changes were introduced which is from https://github.com/eclipse-jdt/eclipse.jdt.core/commit/1f298921739ca05758c7a21eaeee917f2f801bba#diff-8469e0f05a7f350e7047101a3e20fe0bf4d0f65df01fa0a18eb6378128c5de5a

Could it be that we missed to add a default completion in the new if block ? @vik-chand

gayanper:
@vik-chand @stephan-herrmann any thoughts on above mentioned findings. 

I'm thinking of restoring pre Java 17 behaviour together with support for Java17 specific keywords to start with.

stephan-herrmann:
> @vik-chand @stephan-herrmann any thoughts on above mentioned findings.
> 
[...]

PR #1312, opened 2023-08-25
Fix completions in switch-case for type base switch for Java17 and above #1301

gayanper:
## What it does
Compute type completions for switch statements which are type base when the Java version is 17 or greater.

## How to test
Provided test case and unit tests.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
+++ b/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
@@ -3965,7 +3965,8 @@ public final class CompletionEngine
 			isSwitchEnumOrType = resolvedType.isEnum();
 			if(!isSwitchEnumOrType) {
 				if( this.compilerOptions.complianceLevel >= ClassFileConstants.JDK17)
-					isSwitchEnumOrType = resolvedType.isClass() || resolvedType.isInterface() || resolvedType.isRecord();
+					isSwitchEnumOrType = (resolvedType.id != TypeIds.T_JavaLangString)
+							&& (resolvedType.isClass() || resolvedType.isInterface() || resolvedType.isRecord());
 			}
 
 		}
@@ -3989,8 +3990,6 @@ public final class CompletionEngine
 					}
 					System.arraycopy(keywords, 0, keywords = new char[count][], 0, count);
 					findKeywords(this.completionToken, keywords, false, false);
-					// complete types as it is done for version < J17
-					findTypesAndPackages(this.completionToken, scope, false, false, new ObjectVector());
 				}
 
 		} else if (this.expectedTypesPtr > -1 && this.expectedTypes[0].isAnnotationType()) {
--- a/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
+++ b/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
@@ -3989,6 +3989,8 @@ public final class CompletionEngine
 					}
 					System.arraycopy(keywords, 0, keywords = new char[count][], 0, count);
 					findKeywords(this.completionToken, keywords, false, false);
+					// complete types as it is done for version < J17
+					findTypesAndPackages(this.completionToken, scope, false, false, new ObjectVector());
 				}
 
 		} else if (this.expectedTypesPtr > -1 && this.expectedTypes[0].isAnnotationType()) {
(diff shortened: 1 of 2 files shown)


gayanper:
@stephan-herrmann @trancexpress Will one of you have time to have a look at this PR ?

stephan-herrmann:
> @stephan-herrmann @trancexpress Will one of you have time to have a look at this PR ?

I'm looking at it now.

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
```diff
@@ -3989,6 +3989,8 @@ private void completionOnSingleNameReference(ASTNode astNode, ASTNode astNodePar
 					}
 					System.arraycopy(keywords, 0, keywords = new char[count][], 0, count);
 					findKeywords(this.completionToken, keywords, false, false);
+					// complete types as it is done for version < J17
+					findTypesAndPackages(this.completionToken, scope, false, false, new ObjectVector());
```
stephan-herrmann, line 3993: as all this is under the flag `isSwitchEnumOrType` we shouldn't reach this line for string switches, but we do.

One indication for a true type switch would be that the resolvedType is not `final`, because then we don't have any subtypes to propose. Trying this in the test, reveals that `jclMin17.jar` contains a stub for `String` that does not declare the class as `final` (should be corrected anyway).

Alternatively, we could explicitly exclude from `isSwitchEnumOrType` if `resolvedType.id == TypeIds.T_javaLangString`.

stephan-herrmann (changes requested):
I'm missing any tests with a switch over another reference type than `Object`.

E.g., I expect that for a switch over `Number` only types like `Integer`, `Double` ... are proposed. 

With this in mind the check for final-ness mentioned in another comment will become relevant.

stephan-herrmann:
> E.g., I expect that for a switch over `Number` only types like `Integer`, `Double` ... are proposed.

For obvious reasons, searching for all subtypes is not feasible for certain types like `Object` or `Serializable`.

See org.eclipse.jdt.internal.ui.text.java.ContentAssistHistory.UNCACHEABLE for a list of types with huge sub hierarchy. I'm surprised I couldn't find anything in core's search engine to deal with those. Perhaps I didn't look in the right places?

stephan-herrmann:
Another idea: for helping users new to this kind of switch, would it be prudent to insert not only a type, but also a name to which the value shall be bound? We could use existing heuristics for variable naming (`Integer i`, `Person person` ...).

gayanper:
Thanks for the feedback @stephan-herrmann I will look into and improve the PR.

gayanper:
@stephan-herrmann I did some changes as you suggest, Let me know if its go to merge

gayanper:
> I'm missing any tests with a switch over another reference type than `Object`.
> 
> E.g., I expect that for a switch over `Number` only types like `Integer`, `Double` ... are proposed.
> 
> With this in mind the check for final-ness mentioned in another comment will become relevant.

The main purpose of the PR was to allow code completions after `case`, To improve completions for pattern matching we cloud report another issue and capture the expected behavior and work on it. @stephan-herrmann WDYT ?

gayanper:
@stephan-herrmann The PR was made under the following assumptions

When we write a switch block with a int variable
```
switch(i) {
   case Const
}
```
completions behind `Const` will provide proposals where user can choose the class where they want to select a constant for example. if they don't do that, the compiler will issue a error.

Now I would see the same for a switch with a string variable like

```
switch(str) {
   case Const
}
```
But due to the code segment at https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java#L3967
which cause the scenario with str variable not to work. because the `else` condition at https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java#L3980 cause only to have switch block related keyword completions.

So the question is should we handle this scenario ? I feel we should since it is working for primitive and why not for string ? And the next question is why only limiting few keywords as completions when there is a non primitive resolved type in Java17 ? shouldn't the pattern specific errors must be handled by compiler ? And if we really want to support choosing right types in java17+ pattern matching, should we implement it based on completion relevance instead ?

stephan-herrmann:
Thanks, @gayanper . I see now where you're coming from.
Indeed, the fix for [bug 575340](https://bugs.eclipse.org/bugs/show_bug.cgi?id=575340)  leaves some room for improvement, in that it addresses two keywords (`null` and `default`) which have different context conditions.

Indeed also, this PR improves the situation for one particular use case, but I'm afraid that this process will converge only very slowly, unless we ask the major questions first:
* what are the different kinds of switch statements?
* how can we detect them?
* what proposals should be offered for each kind?
  * keywords?
  * search based names? (what search strategy?)
  * more than names (entire patterns)?
* do different kinds of switch statements have commonalities that can be leveraged in the code (this question probably to be asked last)?

I predict that a condition like `isSwitchEnumOrType` will not stand the test of time, because it tries to generalize two situations that have very little in common.

I don't expect to solve all details at once, but tweaking corner cases within a sub-optimal overall design looks like an unprofitable investment of time.

stephan-herrmann:
> And the next question is why only limiting few keywords as completions ...

what other keywords do you have in mind (in addition to `null` and `default`)?

gayanper:
> > And the next question is why only limiting few keywords as completions ...
> 
> what other keywords do you have in mind (in addition to `null` and `default`)?

My question was not about keywords, But rather on other completions such as Type completions. Now the following is on Java11 project

image


same code with Java17 project

image

So as you can see regardless If its primitive or not (in this case String) we get Type completions in Java11 but not in Java17.

This is why I suggest we start with providing Type completion on Java17+ as well as we get in Java11. And then if needed we can fine grain the completions base on Patterns.

stephan-herrmann:
Clickable link to the issue: #1301

Issue #1336, opened 2023-09-01
[21][record patterns] type inference with wildcard


mpalat:
Given the following, ecj gives an error - investigate
```
record R<T> ( T t) {}
public final class X { 
    private static boolean foo(R<?> r) {
        if (r instanceof R(String s)) { // ecj gives failure "Pattern of type ? is not compatible with type java.lang.String"
         return true;
        }
        return false;
    }
 
    public static void main(String argv[]) {
        System.out.println(foo(new R<>(new String("hello"))));
    }
}
```

Record Type inference as per - https://cr.openjdk.org/~gbierman/jep440+441/jep440+441-20230612/specs/patterns-switch-record-patterns-jls.html#jls-18.5.5

stephan-herrmann:
@mpalat  aside from a missing `<T>`  (and bad formatting) I could find no problem with your test. I wrapped it in a proper junit and uploaded it as PR #1338 to make sure we're on the same page. On my machine it passes, let's see what jenkins thinks about it.

stephan-herrmann:
Oops, inference isn't even triggered, because `RecordPattern.shouldInitiateRecordTypeInference()` answers `false`. If I force it to say `true`, then I see the mentioned error. 

Is `shouldInitiateRecordTypeInference()` confused about `PatternVariablesVisitor.hasPatternVar` vs. `PatternVariablesVisitor.typeElidedVar`??

stephan-herrmann:
Next oops: if I tweak `shouldInitiateRecordTypeInference()` to answer `true` then the debugger tells me that `RecordPattern.inferRecordParameterization()` **never** looks at `RecordPattern.patterns`, i.e., the `String s` part is not considered at all. Hence there is no reason to "improve" from `R<?>` but during `setAccessorsPlusInfuseInferredType()` the inner pattern raises the error within `isPatternTypeCompatible()`.

This kind of pattern is called a deconstructor pattern, right? The linked spec, however, doesn't contain the word "deconstructor". So, where would I find specification how those are to be treated?

stephan-herrmann:
Inching along:
* the correct lingo as per the spec is "a record pattern with nested patterns"
* the nested pattern needs no inference, type of `s` is trivially `String`.
* From "_When [...] the pattern does not provide type arguments for R_" I infer that providing type arguments is a legit thing in record patterns:

```
			if (r instanceof R<String>(String s)) {
				return !s.isEmpty();
			}
```
[...]

PR #1338, opened 2023-09-01
[21][record patterns] type inference with wildcard

stephan-herrmann:
Test case for  #1336

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
@@ -32,9 +32,11 @@ import org.eclipse.jdt.internal.compiler.ast.Expression;
 import org.eclipse.jdt.internal.compiler.ast.FunctionalExpression;
 import org.eclipse.jdt.internal.compiler.ast.Invocation;
 import org.eclipse.jdt.internal.compiler.ast.LambdaExpression;
+import org.eclipse.jdt.internal.compiler.ast.Pattern;
 import org.eclipse.jdt.internal.compiler.ast.RecordPattern;
 import org.eclipse.jdt.internal.compiler.ast.ReferenceExpression;
 import org.eclipse.jdt.internal.compiler.ast.SwitchExpression;
+import org.eclipse.jdt.internal.compiler.ast.TypePattern;
 import org.eclipse.jdt.internal.compiler.ast.Wildcard;
 import org.eclipse.jdt.internal.compiler.lookup.TypeConstants.BoundCheckStatus;
 import org.eclipse.jdt.internal.compiler.util.Sorting;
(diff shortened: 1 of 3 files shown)


stephan-herrmann:
In the second commit I cheekily implemented what I guess to be missing from JLS 18.5.5.

This is a long shot needing a confirmation from a spec author or javac developer. For the simple test from #1336 it suffices, and it "suffices" to break test `testRecordPatternTypeInference_010()` which no fails to generate a necessary instanceof test, resulting in` java.lang.ClassCastException: class X$$Lambda/0x00000008000c0c00 cannot be cast to class java.lang.String`

Also the tweak in `shouldInitiateRecordTypeInference()` is likely illegal.

stephan-herrmann:
@mpalat let me know if you want me to translate the impl sketch back into spec language that could be used for discussion with the spec author.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
```diff
@@ -2005,6 +2010,25 @@ public ReferenceBinding inferRecordPatternParameterization(
...
+			Pattern[] patterns = recordPattern.patterns;
+			RecordComponentBinding[] components = ((ReferenceBinding) recordPattern.resolvedType.original()).components();
+			List<ConstraintFormula> extendedConstraints = new ArrayList<>();
+			if (this.initialConstraints != null)
+				extendedConstraints.addAll(Arrays.asList(this.initialConstraints));
+			for (int i = 0; i < patterns.length; i++) {
+				Pattern pattern = patterns[i];
+				if (pattern instanceof TypePattern && pattern.resolvedType != null && pattern.resolvedType.isValidBinding()) {
+					TypeBinding compTypeSubst = substitute(components[i].type);
+					if (!compTypeSubst.isProperType(true))
```
stephan-herrmann, line 2024: this check could be wrong, perhaps the constraint needs to be created always, to help detecting incompatible situations.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
```diff
@@ -2005,6 +2010,25 @@ public ReferenceBinding inferRecordPatternParameterization(
 		return getRecordPatternTypeFromUpwardsProjection(typeBinding, alphas, solution);
 	}
 
+	public void addConstraintsFromNestedPatterns(RecordPattern recordPattern) {
+		if (recordPattern.patterns != null) {
+			Pattern[] patterns = recordPattern.patterns;
+			RecordComponentBinding[] components = ((ReferenceBinding) recordPattern.resolvedType.original()).components();
```
stephan-herrmann, line 2016: calling `original()` probably is cheating. I guess we need to peel raw types but not parameterized types.

mpalat:
> @mpalat let me know if you want me to translate the impl sketch back into spec language that could be used for discussion with the spec author.

Yes, I would definitely require .. however, before that , I just want to confirm that it is not a bug from our side - ref https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1336#issuecomment-1703745394

stephan-herrmann:
Test case was copied to #1341--  the rest (second commit) is probably incorrect.

PR #1339, opened 2023-09-02
A investigative patch for

mpalat:
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1336 - WIP - DONOT MERGE

## Author checklist

- [ ] I have thoroughly tested my changes
- [ ] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [ ] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

PR #1341, opened 2023-09-02
[21][record patterns] type inference with wildcard

stephan-herrmann:
## What it does

Fresh PR with less guess work but containing:
* the test from #1338 
* an assumed fix by applying capture in one location
* update of impl from recent spec 18.5.5 item 3

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/CaseStatement.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/CaseStatement.java
@@ -506,7 +506,7 @@ private Constant resolveConstantExpression(BlockScope scope,
 			// The following code is copied from InstanceOfExpression#resolve()
 			// But there are enough differences to warrant a copy
 			if (!pb.isReifiable()) {
-				if (expressionType != TypeBinding.NULL) {
+				if (expressionType != TypeBinding.NULL && !(e instanceof RecordPattern)) {
 					boolean isLegal = e.checkCastTypesCompatibility(scope, pb, expressionType, e, false);
 					if (!isLegal || (e.bits & ASTNode.UnsafeCast) != 0) {
 						scope.problemReporter().unsafeCastInInstanceof(e, pb, expressionType);
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/RecordPattern.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/RecordPattern.java
@@ -190,7 +190,7 @@ public class RecordPattern extends TypePattern {
 	}
 	private void setAccessorsPlusInfuseInferredType(BlockScope scope) {
 		this.isTotalTypeNode = super.coversType(this.resolvedType);
-		RecordComponentBinding[] components = this.resolvedType.components();
+		RecordComponentBinding[] components = this.resolvedType.capture(scope, this.sourceStart, this.sourceEnd).components();
 		if (components.length != this.patterns.length) {
 			scope.problemReporter().recordPatternSignatureMismatch(this.resolvedType, this);
 		} else {
@@ -227,16 +227,7 @@ public class RecordPattern extends TypePattern {
 		}
 	}
 	private boolean shouldInitiateRecordTypeInference() {
-		ReferenceBinding binding = this.resolvedType != null ? this.resolvedType.actualType() : null;
-		if (binding == null || !binding.isGenericType())
-			return false;
-
-		if (this.resolvedType.isParameterizedType())
-			return false;
-		if (this.containsTypeElidedPatternVar == null) {
-			this.containsPatternVariable();
-		}
-		return this.containsTypeElidedPatternVar;
+		return this.resolvedType != null && this.resolvedType.isRawType();
 	}
 	private void infuseInferredType(TypePattern tp, RecordComponentBinding componentBinding) {
 		SingleTypeReference ref = new SingleTypeReference(tp.local.type.getTypeName()[0],
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
@@ -337,8 +337,11 @@ public class InferenceContext18 {
 		}
 	}
 
-	/** Add new inference variables for the given type arguments. */
-	public InferenceVariable[] addTypeVariableSubstitutions(TypeBinding[] typeArguments) {
+	/**
+	 * Add new inference variables for the given type arguments.
+	 * CAVEAT: when passing capturesOnly as true, then the result array may contain nulls!
+	 */
+	public InferenceVariable[] addTypeVariableSubstitutions(TypeBinding[] typeArguments, boolean capturesOnly) {
 		int len2 = typeArguments.length;
 		InferenceVariable[] newVariables = new InferenceVariable[len2];
 		InferenceVariable[] toAdd = new InferenceVariable[len2];
@@ -346,7 +349,7 @@ public class InferenceContext18 {
 		for (int i = 0; i < typeArguments.length; i++) {
 			if (typeArguments[i] instanceof InferenceVariable)
 				newVariables[i] = (InferenceVariable) typeArguments[i]; // prevent double substitution of an already-substituted inferenceVariable
-			else if (typeArguments[i] instanceof CaptureBinding)
+			else if (!capturesOnly || typeArguments[i] instanceof CaptureBinding)
 				toAdd[numToAdd++] =
 					newVariables[i] = InferenceVariable.get(typeArguments[i], i, this.currentInvocation, this.scope, this.object, false);
 		}
@@ -2092,7 +2095,7 @@ public class InferenceContext18 {
 			 * And hence the name notJust18_5_5_item_3_bullet_1Betas
 			 * TODO: a Just18_5_5_item_3_bullet_1Betas utility?
 			 */
-			InferenceVariable[] notJust18_5_5_item_3_bullet_1Betas = addTypeVariableSubstitutions(arguments);
+			InferenceVariable[] notJust18_5_5_item_3_bullet_1Betas = addTypeVariableSubstitutions(arguments, true);
 			TypeVariableBinding[] typeVariables = getTPrimeArgumentsAndCreateBounds(parameterizedType,
 					notJust18_5_5_item_3_bullet_1Betas);
 			tPrime = this.environment.createParameterizedType(
@@ -2129,7 +2132,7 @@ public class InferenceContext18 {
 		TypeBinding[] aArr = tPrime.typeArguments();
 		for (int i = 0, l = notJust18_5_5_item_3_bullet_1Betas.length; i < l; ++i) {
 			InferenceVariable beta = notJust18_5_5_item_3_bullet_1Betas[i];
-			if (!beta.equals(typeVariables[i])) continue; //not an expected inference variable.
+			if (beta == null || !beta.equals(typeVariables[i])) continue; //not an expected inference variable.
 
 			TypeBinding[] uArr = typeParams[i]!= null ? typeParams[i].allUpperBounds() : null;
 			if (uArr == null || uArr.length == 0) {
@@ -2169,7 +2172,7 @@ public class InferenceContext18 {
 		TypeBound bound;
 		for (int i = 0, l = arguments.length; i < l; ++i) {
 			bound = null;
-			if (arguments[i].kind() == Binding.WILDCARD_TYPE) {
+			if (arguments[i].kind() == Binding.WILDCARD_TYPE && beta[i] != null) {
 				WildcardBinding wildcard = (WildcardBinding) arguments[i];
 				switch(wildcard.boundKind) {
 					case Wildcard.EXTENDS :
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
@@ -337,18 +337,18 @@ public class InferenceContext18 {
 		}
 	}
 
-	/** Add new inference variables for the given type variables. */
-	public InferenceVariable[] addTypeVariableSubstitutions(TypeBinding[] typeVariables) {
-		int len2 = typeVariables.length;
+	/** Add new inference variables for the given type arguments. */
+	public InferenceVariable[] addTypeVariableSubstitutions(TypeBinding[] typeArguments) {
+		int len2 = typeArguments.length;
 		InferenceVariable[] newVariables = new InferenceVariable[len2];
 		InferenceVariable[] toAdd = new InferenceVariable[len2];
 		int numToAdd = 0;
-		for (int i = 0; i < typeVariables.length; i++) {
-			if (typeVariables[i] instanceof InferenceVariable)
-				newVariables[i] = (InferenceVariable) typeVariables[i]; // prevent double substitution of an already-substituted inferenceVariable
-			else
+		for (int i = 0; i < typeArguments.length; i++) {
+			if (typeArguments[i] instanceof InferenceVariable)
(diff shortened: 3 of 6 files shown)


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/RecordPattern.java
```diff
@@ -190,7 +190,7 @@ public TypeBinding resolveType(BlockScope scope, boolean isPatternVariable) {
 	}
 	private void setAccessorsPlusInfuseInferredType(BlockScope scope) {
 		this.isTotalTypeNode = super.coversType(this.resolvedType);
-		RecordComponentBinding[] components = this.resolvedType.components();
+		RecordComponentBinding[] components = this.resolvedType.capture(scope, this.sourceStart, this.sourceEnd).components();
```
stephan-herrmann, line 193: Rationale: accessing the components `r.ci()` would normally require to first evaluate `r` which includes capture conversion. This ensures that the components have real types, rather than wildcards (which are not types), given that the spec assumes that the components have types.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/RecordPattern.java
```diff
@@ -227,16 +227,7 @@ private ReferenceBinding inferRecordParameterization(BlockScope scope, Reference
...
-		if (binding == null || !binding.isGenericType())
-			return false;
-
-		if (this.resolvedType.isParameterizedType())
-			return false;
-		if (this.containsTypeElidedPatternVar == null) {
-			this.containsPatternVariable();
-		}
-		return this.containsTypeElidedPatternVar;
+		return this.resolvedType != null && this.resolvedType.isRawType();
```
stephan-herrmann, line 230: nothing more is stated in the spec

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
```diff
@@ -337,18 +337,18 @@ private void addInferenceVariables(InferenceVariable[] newVariables) {
...
 		InferenceVariable[] toAdd = new InferenceVariable[len2];
 		int numToAdd = 0;
-		for (int i = 0; i < typeVariables.length; i++) {
-			if (typeVariables[i] instanceof InferenceVariable)
-				newVariables[i] = (InferenceVariable) typeVariables[i]; // prevent double substitution of an already-substituted inferenceVariable
-			else
+		for (int i = 0; i < typeArguments.length; i++) {
+			if (typeArguments[i] instanceof InferenceVariable)
+				newVariables[i] = (InferenceVariable) typeArguments[i]; // prevent double substitution of an already-substituted inferenceVariable
+			else if (typeArguments[i] instanceof CaptureBinding)
```
stephan-herrmann, line 349: refresh from the spec: _the type variables produced by capture that are type arguments in Tc_

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
```diff
@@ -2078,14 +2078,14 @@ private boolean findRPrimeAndResultingBounds(TypeBinding typeBinding, InferenceV
 		} /* else part: Otherwise, B2 is the same as B1.*/
 		return true;
 	}
-	private TypeBinding deriveTPrime(TypeBinding candidateT, InferenceVariable[] alphas, TypeBinding typeBinding) {
+	private TypeBinding deriveTPrime(RecordPattern recordPattern, TypeBinding candidateT, InferenceVariable[] alphas, TypeBinding typeBinding) {
```
stephan-herrmann, line 2081: AST node needed for capture (needs a source position for uniqueness)

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
```diff
@@ -2078,14 +2078,14 @@ private boolean findRPrimeAndResultingBounds(TypeBinding typeBinding, InferenceV
...
 		ParameterizedTypeBinding parameterizedType = null;
 		TypeBinding tPrime = null;
 		if (candidateT.isParameterizedType()) {
 			parameterizedType = InferenceContext18.parameterizedWithWildcard(candidateT);
 		}
 		if (parameterizedType != null && parameterizedType.arguments != null) {
-			TypeBinding[] arguments = parameterizedType.arguments;
+			TypeBinding[] arguments = parameterizedType.capture(this.scope, recordPattern.sourceStart, recordPattern.sourceEnd).arguments;
```
stephan-herrmann, line 2088: refresh from the spec saying "_let Tc be the result of capture conversion (5.1.10) applied to T_"

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
```diff
@@ -2165,17 +2165,20 @@ private TypeVariableBinding[] getTPrimeArgumentsAndCreateBounds(
 			InferenceVariable[] beta) {
 		TypeBinding[] arguments = parameterizedType.typeArguments();
 		TypeVariableBinding[] typeVariables = new TypeVariableBinding[arguments.length];
+		InferenceSubstitution theta = new InferenceSubstitution(this.environment, beta, this.currentInvocation);
```
stephan-herrmann, line 2168: Here and usage of theta below: update from the spec where it says: "_let θ be the substitution[Z1:=β1, ..., Zk:=βk]. T' is Tc θ._" etc.

Note, that we don't seem to have any tests that actually **need** this.

stephan-herrmann:
> [NPE fix](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1341/commits/6bf8c30eedb8f6e03140c5847a6833e4c8e53f7b)

Cool, using the modified compiler to compile itself demonstrated that I had introduced a fresh NPE :)

stephan-herrmann:
I self-commented all changes to document the what and why. HTH.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/RecordPattern.java
mpalat: +1

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
```diff
@@ -337,18 +337,18 @@ private void addInferenceVariables(InferenceVariable[] newVariables) {
 		}
 	}
 
-	/** Add new inference variables for the given type variables. */
-	public InferenceVariable[] addTypeVariableSubstitutions(TypeBinding[] typeVariables) {
-		int len2 = typeVariables.length;
+	/** Add new inference variables for the given type arguments. */
```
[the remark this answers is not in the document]
mpalat: +1

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/RecordPattern.java
mpalat: Agree.. +1

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
mpalat: will take a look at the more recent code that has
`			else if (!capturesOnly || typeArguments[i] instanceof CaptureBinding)
`

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
mpalat: +1

mpalat:
> I self-commented all changes to document the what and why. HTH.

helps a lot.. thanks!

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
mpalat: > Here and usage of theta below: update from the spec where it says: "_let θ be the substitution[Z1:=β1, ..., Zk:=βk]. T' is Tc θ._" etc.
> 
> Note, that we don't seem to have any tests that actually **need** this.

`sealed interface I<T> {}
sealed interface J<T> {}
class A  {}
record R<T>(T t) implements I<T>, J<T>{}

public class X<T extends I<? extends A> & J<? extends A>> {

    public boolean foo(T t) {
    	return switch(t) {
    		case R(A a) -> true;
    		default -> false;
    	};
    }

       System.out.println(new X<R<? extends A>>().foo(new R(new A())));
    }
}`
the code above actually hits that line where theta is created.. and while javac passes ecj gives an error "Type T cannot be safely cast to R<? extends A>"

mpalat (approved):
@stephan-herrmann +1 for the code. 

Couple of things:
(1) There is one test issue at https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1341#discussion_r1314227342 - didn't check whether it was failing earlier. 
(2) you had pointed out an issue in https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1336#issuecomment-1703815933 - about parameterizedWithWildcard() in deriveTPrime() - I didn't see that change in this patch. so checking whether you think we need to keep it?

we can commit this patch and then take the above issues later. leaving to you to comment. 

Thanks a lot!

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
stephan-herrmann: I assume `I` and `J` should be generic (this time it wasn't github swallowing angle brackets).

But then I get
```
3. ERROR in X.java (at line 16)
	System.out.println(new X<R<? extends A>>().foo(new R(new A())));
	                         ^
Bound mismatch: The type R<? extends A> is not a valid substitute for the bounded parameter <T extends I<? extends A> & J<? extends A>> of the type X<T>
```

How should I complete the example?

stephan-herrmann:
> (2) you had pointed out an issue in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1341#discussion_r1314227342 - about parameterizedWithWildcard() in deriveTPrime() - I didn't see that change in this patch. so checking whether you think we need to keep it?

I wanted to check back with you, if you have a good "excuse" for that call :)

stephan-herrmann:
@mpalat  when I pull in `testRecordPatternTypeInference_015()` from #1339 I get
```
	if (r instanceof R(X x)) {
	                   ^^^
Pattern of type capture#2-of ? extends I is not compatible with type X
```
(message changes from `?` to `capture#2-of ?`, no surprise here).

I couldn't quite interpret the message: the problem is reported against the type pattern `X x`, so why does it say `Pattern of type capture#2-of ? extends I`? Should it say `Record component with type capture#2-of ? extends I` (assuming that `R` was inferred as `R<? extends I>`, so we already know that its component would have at least type `capture#2-of ? extends I`).

mpalat:
> 
> I wanted to check back with you, if you have a good "excuse" for that call :)

No I don't have a "good" excuse - that said, once I skip that part and make it just a typecast,

```
		if (candidateT.isParameterizedType()) {
			parameterizedType = (ParameterizedTypeBinding) candidateT;
		}

```
I do see a couple of test failures - testRecordPatternTypeInference_004 and testRecordPatternTypeInference_008 - am yet to analyse them though

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
mpalat: > I assume `I` and `J` should be generic (this time it wasn't github swallowing angle brackets).

yes, it should be generic - I don't know how this gobbling up happened again - I've corrected the same.  Could you please check whether you are getting the error? I am not seeing this error - noting that I've applied your patch and then checking, just to be sure we are on the same page.

stephan-herrmann: Thanks, now I got past those unintended errors. While I was at it, I took the liberty to use unique names for all type parameters in the test, because working with type `T` is quite cumbersome, if *every* type parameter is called `T` :)

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/CaseStatement.java
```diff
@@ -506,7 +506,7 @@ private Constant resolveConstantExpression(BlockScope scope,
 			// The following code is copied from InstanceOfExpression#resolve()
 			// But there are enough differences to warrant a copy
 			if (!pb.isReifiable()) {
-				if (expressionType != TypeBinding.NULL) {
+				if (expressionType != TypeBinding.NULL && !(e instanceof RecordPattern)) {
```
stephan-herrmann, line 509: I don't see why these checks should be appropriate for record patterns: casting to a generic record type will always be "unsafe", but this is compensated by separate instanceof check for the component(s).

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
stephan-herrmann: Test included in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1341/commits/0ace862918733f41e5f74520152b0575f79df142, but still I don't think we really exercised the substitution stuff ...

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/problem/messages.properties
```diff
@@ -1154,7 +1154,7 @@
 # Record patterns = Java 21
 1912 = Only record types are permitted in a record pattern
 1913 = Record pattern should match the signature of the record declaration
-1914 = Pattern of type {0} is not compatible with type {1}
+1914 = Record component with type {0} is not compatible with type {1}
```
stephan-herrmann, line 1157: An attempt at improving the message. @mpalat does this make sense? (see tests below for how it spells out in context).

stephan-herrmann:
Failure in InclusionPatternsTests.testNestedSourceFolder4 is most likely independent.

@manoj if you need this, I think it's a somewhat consistent state for merging, with the following tasks remaining (here or follow-up):
* your review of my change of error message
* validate the theta-substitution is doing the right thing, perhaps using examples for more complex inference situations
* investigate need/correctness of calling `parameterizedWithWildcard()` in `deriveTPrime()`

stephan-herrmann:
BTW, when merging I suggest to provide a custom commit message summarizing the changes :)

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/problem/messages.properties
mpalat: +1

mpalat:
> Failure in InclusionPatternsTests.testNestedSourceFolder4 is most likely independent.
yes, it is.
>  following tasks remaining (here or follow-up):
> 
> * your review of my change of error message 
done
> * validate the theta-substitution is doing the right thing, perhaps using examples for more complex inference situations
raised follow-up issue - https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1343
> * investigate need/correctness of calling `parameterizedWithWildcard()` in `deriveTPrime()`
raised  follow-up issue - https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1344

Issue #308, opened 2022-08-15
OOME during type reference search - same source multiple times in heap


jukzi:
I have a big workspace where i get a OOME when i do this search:
![image](https://user-images.githubusercontent.com/51790620/184623989-1de5da3c-7941-4de8-9934-808e8b30f464.png)
![image](https://user-images.githubusercontent.com/51790620/184626980-319c101e-7684-48d7-9809-86b6bf232e80.png)

I created a Heap dump when OOME occurs and see the sourcecode of  (big)  .java files is opened multiple times in the memory. (in ResourceCompilationUnit.contents).

![image](https://user-images.githubusercontent.com/51790620/184624308-89c073e1-6d86-4024-afdd-620446f36bc2.png)

The sourceEnd and sourceStart in QualifiedTypeRefence however are distinct:
![image](https://user-images.githubusercontent.com/51790620/184625328-fdd8ec11-8b19-4784-a7cc-080bb76648fd.png)

Would be nice if they could share the same sourcode instance to safe a lot memory!

May relate to https://bugs.eclipse.org/bugs/show_bug.cgi?id=571467 ".java files are read again for each lambda inside"

Also i noticed that memory is Garbage collected as soon as the search is finished (if i spend 9GB memory it does). So it would also help if there would be some memory released during search (for example after every project searched in).

jukzi:
Can be reproduced in Platform Workspace with java-search for references of `java.lang.Object`
![image](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/51790620/beae73f7-8b38-4cbd-9d66-9c1afe404ce1)
Search will go into OOME after only ~ 19.000 hits while normally search can handle millions of results (see https://github.com/eclipse-platform/eclipse.platform.text/issues/186)

srikanth-sankaran:
@jukzi https://bugs.eclipse.org/bugs/show_bug.cgi?id=571467 itself is resolved. Can you try if the problem you observed has gone away ??

jukzi:
> Can you try if the problem you observed has gone away ??
Still happens. search for reference to `java.lang.Object`
![image](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/51790620/cb414088-37e6-4435-8c53-1b01c3524cb0)
![image](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/51790620/afa4b277-b17b-477d-91f1-8e8157d38763)
[...]

PR #1358, opened 2023-09-06
Java Search: fix OOME during type reference search #308

jukzi:
* release more memory (TypeReferenceLocator.recordedResolutions) after each search bulk
* increase bulk size
* added some final modifiers.

memory consumption during search:
image


jukzi:
That's weird: setting MatchLocator.MAX_AT_ONCE to 1000 (any value >= 723) fails org.eclipse.jdt.core.tests.model.JavaSearchBugs16Tests.testAIOOBEForRecordClassGh790()

PR #1036, opened 2023-05-09
Bug 573552 - Missing warning for diamond operator

jarthana:
For method invocations, the allocation expression does not get its expected type soon enough. So, we should wait till it's available.

## Author checklist

- [ ] I have thoroughly tested my changes
- [ ] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [ ] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

PR #1378, opened 2023-09-14
False positive warning: redundant specification of type arguments

### LINKED ISSUE #1326: False positive warning: redundant specification of type arguments
Given the following classes:

```java
class Inner<T> {}
class Outer<T> {
   Outer(T t) {}
   Outer<T> self() { return this; }
}
```

This code generates a warning:

```java
Outer<Inner<String>> x = new Outer<>(new Inner<String>()).self();
//                                       ^^^^^
// Redundant specification of type arguments <String>
```

But omitting the type argument generates an error as type inference fails and falls-back to `Object`:

```java
Outer<Inner<String>> x = new Outer<>(new Inner<>()).self();
//                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
// Type mismatch: cannot convert from Outer<Inner<Object>> to Outer<Inner<String>>
```

This is a relatively recent change that I assume was introduced with #1036.

stephan-herrmann:
fixes #1326

## What it does
Creates a little worm-hole between disconnected pieces of resolving:
* at the end of outer inference, `forwardResult()` checks for the following situation:
  * an argument is an `AllocationExpression`
  * the corresponding parameter has been updated from the original method binding to the `ParameterizedGenericMethodBinding` as the result of inference
  * if so, this situation is remembered in a new flag `AllocationExpression.expectedTypeWasInferred`.
* now the code block from [#1036](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1036) can avoid false positives based on the above flag.

Rationale: the flag signals that the target type for a given inner allocation was indeed determined during outer inference. This _could_ indicate, that without type arguments of the inner, inference might have failed. In the end we _might_ err on the side of keeping silent, if that outer inference got its information from a different source than the inner allocation. The exact answer can only be given by modifying the code and running the compiler. But in such subtle cases, specifying a few potentially redundant type arguments is probably the least bad.

## How to test
JUnit tests included.

In particular `testGH1326_alt` demonstrates that an outer inference alone doesn't set `expectedTypeWasInferred` if the corresponding parameter was not influenced by inference. The test also shows, that in this situation type arguments can indeed safely be omitted.

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ASTNode.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ASTNode.java
@@ -358,7 +358,7 @@ public abstract class ASTNode implements TypeConstants, TypeIds {
 		if (argument instanceof AllocationExpression) {
 			AllocationExpression allocExp = (AllocationExpression) argument;
 			// we need this only when the error is not reported in AllocationExpression#checkTypeArgumentRedundancy()
-			if (allocExp.typeExpected == null) {
+			if (allocExp.typeExpected == null && !allocExp.expectedTypeWasInferred) {
 				final boolean isDiamond = allocExp.type != null && (allocExp.type.bits & ASTNode.IsDiamond) != 0;
 				allocExp.typeExpected = parameterType;
 				if (!isDiamond && allocExp.resolvedType.isParameterizedTypeWithActualArguments()) {
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/AllocationExpression.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/AllocationExpression.java
@@ -91,6 +91,11 @@ public class AllocationExpression extends Expression implements IPolyExpression,
 	public boolean argsContainCast;
 	public TypeBinding[] argumentTypes = Binding.NO_PARAMETERS;
 	public boolean argumentsHaveErrors = false;
+	/**
+	 * This flag avoids "Redundant specification of type arguments" if the target type was inferred in an outer context,
+	 * possibly based on the provided type arguments of this allocation:
+	 */
+	public boolean expectedTypeWasInferred;
 
 @Override
 public FlowInfo analyseCode(BlockScope currentScope, FlowContext flowContext, FlowInfo flowInfo) {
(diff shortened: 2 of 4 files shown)


jarthana (approved):
The fact that the changes from pull #1036 remain largely untouched makes me happy. The rest of the changes look good to me too.

stephan-herrmann:
> In particular testGH1326_alt demonstrates that an outer inference alone doesn't set expectedTypeWasInferred if the corresponding parameter was not influenced by inference. The test also shows, that in this situation type arguments can indeed safely be omitted.

This new test failed once and only on one platform: https://download.eclipse.org/eclipse/downloads/drops4/I20230916-1800/testresults/html/org.eclipse.jdt.core.tests.compiler_ep430I-unit-cen64-gtk3-java20_linux.gtk.x86_64_20.html#TestAll

```
Invalid problem log . ----------- Expected ------------ ----------\n 1. ERROR in Foo.java (at line 8)\n Outer<Inner<String>> x = new Outer<>(new Inner<String>(), inner).self();\n ^^^^^\n Redundant specification of type arguments <String>\n ----------\n ------------ but was ------------ ----------\n 1. ERROR in Foo.java (at line 8)\n Outer<Inner<String>> x = new Outer<>(new Inner<String>(), inner).self();\n ^^^^^\n Redundant specification of type arguments <String>\n ----------\n 2. ERROR in Foo.java (at line 9)\n Outer<Inner<String>> xok = new Outer<>(new Inner<>(), inner).self();\n ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n Cannot infer type arguments for Outer<>\n ----------\n --------- Difference is ---------- expected:<...ing>\n ----------\n []> but was:<...ing>\n ----------\n [2. ERROR in Foo.java (at line 9)\n Outer<Inner<String>> xok = new Outer<>(new Inner<>(), inner).self();\n ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n Cannot infer type arguments for Outer<>\n ----------\n ]>

junit.framework.ComparisonFailure: Invalid problem log .
----------\n
1. ERROR in Foo.java (at line 8)\n
Outer<Inner<String>> x = new Outer<>(new Inner<String>(), inner).self();\n
^^^^^\n
Redundant specification of type arguments <String>\n
----------\n

----------\n
1. ERROR in Foo.java (at line 8)\n
Outer<Inner<String>> x = new Outer<>(new Inner<String>(), inner).self();\n
^^^^^\n
Redundant specification of type arguments <String>\n
----------\n
2. ERROR in Foo.java (at line 9)\n
Outer<Inner<String>> xok = new Outer<>(new Inner<>(), inner).self();\n
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n
Cannot infer type arguments for Outer<>\n
----------\n

--------- Difference is ----------
expected:<...ing>\n
----------\n
[]> but was:<...ing>\n
----------\n
[2. ERROR in Foo.java (at line 9)\n
Outer<Inner<String>> xok = new Outer<>(new Inner<>(), inner).self();\n
^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n
Cannot infer type arguments for Outer<>\n
----------\n
]> 
```

Note that this is a genuine inference failure, *not* the redundant type arguments thing, which was changed in this issue.

Issue #1289, opened 2023-08-15
java.lang.IllegalArgumentException: "L;" at 0


jukzi:
logged during org.eclipse.jdt.debug.tests.eval.SyntheticVariableTests.testCompleteMethodParameter_DeepInTwoNestedClasses()

jukzi:
```
java.lang.IllegalArgumentException: "L;" at 0
	at org.eclipse.jdt.core.Signature.newIllegalArgumentException(Signature.java:3017)
	at org.eclipse.jdt.core.Signature.appendClassTypeSignature(Signature.java:608)
	at org.eclipse.jdt.core.Signature.appendTypeSignature(Signature.java:931)
	at org.eclipse.jdt.core.Signature.appendTypeSignature(Signature.java:875)
	at org.eclipse.jdt.core.Signature.toCharArray(Signature.java:2753)
	at org.eclipse.jdt.ui.text.java.CompletionProposalCollector.getDeclaringType(CompletionProposalCollector.java:639)
	at org.eclipse.jdt.ui.text.java.CompletionProposalCollector.accept(CompletionProposalCollector.java:281)
	at org.eclipse.jdt.internal.codeassist.CompletionEngine.findFields(CompletionEngine.java:7354)
	at org.eclipse.jdt.internal.codeassist.CompletionEngine.findFields(CompletionEngine.java:7491)
	at org.eclipse.jdt.internal.codeassist.CompletionEngine.findVariablesAndMethods(CompletionEngine.java:13457)
	at org.eclipse.jdt.internal.codeassist.CompletionEngine.completionOnSingleNameReference(CompletionEngine.java:4031)
	at org.eclipse.jdt.internal.codeassist.CompletionEngine.complete(CompletionEngine.java:2018)
	at org.eclipse.jdt.internal.codeassist.CompletionEngine.complete(CompletionEngine.java:2615)
	at org.eclipse.jdt.internal.core.SourceType.codeComplete(SourceType.java:142)
	at org.eclipse.jdt.internal.core.SourceType.codeComplete(SourceType.java:101)
	at org.eclipse.jdt.internal.core.SourceType.codeComplete(SourceType.java:87)
	at org.eclipse.jdt.internal.debug.ui.contentassist.JavaDebugContentAssistProcessor.computeCompletionProposals(JavaDebugContentAssistProcessor.java:164)
[...]

PR #1381, opened 2023-09-15
Completion: Do not create proposal for Anonymous Class #1289

jukzi:
fixes java.lang.IllegalArgumentException: "L;" in Signature
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1289


jukzi:
@jjohnstn mind to look on this change?

jjohnstn:
@jukzi Ok

jjohnstn (approved):
Seems like a reasonable solution.

PR #1389, opened 2023-09-19
Upload eventfile and unit test results

laeubi:
FYI @mickaelistria @stephan-herrmann

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,15 @@ on:
     branches: '**'
 
 jobs:
+  event_file:
+    name: "Upload Event File"
+    runs-on: ubuntu-latest
+    steps:
+    - name: Upload
+      uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
+      with:
+        name: Event File
+        path: ${{ github.event_path }}
   build:
     runs-on: ubuntu-latest
     steps:
@@ -35,3 +44,11 @@ jobs:
       run: |
         mvn clean install --batch-mode -f org.eclipse.jdt.core.compiler.batch -DlocalEcjVersion=99.99
         mvn -U clean verify --batch-mode --fail-at-end -Ptest-on-javase-20 -Pbree-libs -Papi-check -Djava.io.tmpdir=$WORKSPACE/tmp -Dproject.build.sourceEncoding=UTF-8 -Dtycho.surefire.argLine="--add-modules ALL-SYSTEM -Dcompliance=1.8,11,17,20 -Djdt.performance.asserts=disabled" -Dcbi-ecj-version=99.99
+    - name: Upload Test Results for Linux
+      if: always()
+      uses: actions/upload-artifact@a8a3f3ad30e3422c9c7b888a15615d19a852ae32 # v3.1.3
+      with:
+        name: test-results-linux
+        if-no-files-found: warn
+        path: |
+          ${{ github.workspace }}/**/target/surefire-reports/*.xml


mickaelistria:
Thank you @laeubi .
Sorry for the inconvenience, I indeed really forgot about that block and didn't spot the issue earlier.

.github/workflows/ci.yml
```diff
@@ -8,6 +8,15 @@ on:
     branches: '**'
```
laeubi, line 8: one *might* also want to trigger this workflow on PRs...
See for example https://github.com/eclipse-platform/eclipse.platform.ui/blob/04265dec58c0fa26b5fe7360c0a6d56cf8218517/.github/workflows/ci.yml#L6-L10

mickaelistria: > one might also want to trigger this workflow on PRs...

In a previous discussion with @rgrunber , we agreed that we don't want it at the moment, to avoid the risk of disturbing current activity on JDT, and the failure here proves it was wise to be cautious.
We may discuss enabling it for PRs later if there is a demand for it from JDT contributors; but at the moment Jenkins seems sufficient and reliable enough.

mickaelistria: Jenkins does already report problems well enough. The ci.yml here is more added so that contributors can have GitHub builds on their forks for free, to validate their changes, before pushing a PR (that would then be verified on Jenkins).

stephan-herrmann:
Two questions from a github-naive user:

> one might also want to trigger this workflow on PRs...

What issues would be detected by this workflow beyond what the jenkins job does?

Is the original intention to run additional verification *after* each PR merge? For what benefit?

mickaelistria:
The original intention is captured in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1254 : "an easily be used to enable validation in forks, just by enabling GitHub actions on the forked repository. So it makes it almost free for contributors to get some CI validation, and thus to provide higher quality PRs more easily."

stephan-herrmann:
> The original intention is captured in #1254 : "an easily be used to enable validation in forks, just by enabling GitHub actions on the forked repository. So it makes it almost free for contributors to get some CI validation, and thus to provide higher quality PRs more easily."

Sure, but what should be the result from all this **here** on the original repo? Should validations run **before and after** each PR merge??

mickaelistria:
> Sure, but what should be the result from all this here on the original repo?

The expected result here would be a GitHub build running only for branches (master and so on) when changes happen, and with a successful result. It's not meant to replace Jenkins build nor to add extra redundant checks to PR validations; nothing new happens for PRs.

rgrunber:
As mentioned in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1254#discussion_r1301879584 , we could have the job only run on forks (stay silent on the original repo until a migration away from Jenkins does happen) if having the additional job running becomes annoying.

laeubi:
> What issues would be detected by this workflow beyond what the jenkins job does?

In general having run the verification on different systems/configuration gives better confidence that we are not coding against a single configuration specific setup (e.g. Jenkins uses Redhat Linux, Github uses latest Ubuntu Linux), beside that Github offers runner for Windows+Mac while Jenkins only offers Linux by default.

Also from time to time we see the Jenkins instances where down, then its good to have at least the alternative verification at hand.

> Is the original intention to run additional verification _after_ each PR merge?

This is more a side-effect, without that users would have to do all the setup on their own fork with that one only needs to enable Action runs.

> For what benefit?

The only benefit is for those who don't know how to run a maven commandline.

laeubi:
@mickaelistria @akurtakov @rgrunber @stephan-herrmann any interest in improving github actions or should we close this?

.github/workflows/ci.yml
akurtakov: Why not upload the results from jenkins on pr as a comment?

stephan-herrmann:
Can someone with sufficient github knowledge please review this?

Otherwise #1254 is due for a revert, IMHO, as we keep getting bogus error mails upon every successful merge, which in the end means that no committer will pay any attention to such mails any more.

laeubi:
> If/when we get them not failing I will push them.

Everything is green...

akurtakov:
Let's have this one and see what(if anything) still annoys people after that. 
@stephan-herrmann I am looking to have https://github.com/eclipse-platform/eclipse.platform.ui/pull/1412#issuecomment-1855360972 like comments in jdt.

akurtakov:
I've rebased https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1698 to see the effects there.

PR #1395, opened 2023-09-19
Use StringBuilder instead of StringBuffer in package o.e.jdt.core

HannesWell:
## What it does
Replace all usages of  `StringBuffer` in package o.e.jdt.core by `StringBuilder` since the code seems to be purely single-threaded and StringBuilder is generally faster since it is not synchronized unlike `StringBuffer`.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/core/Signature.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/core/Signature.java
@@ -463,7 +463,7 @@ public final class Signature {
 	public static final int WILDCARD_TYPE_SIGNATURE = 5;
 
 // <x.y.z, a.b<c>.d<e.f>> --> <z,d<f>>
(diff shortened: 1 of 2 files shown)


HannesWell:
Thank you Andrey.
There are many more usages of `StringBuffer` in JDT, mainly in ASTNode.print*() methods.
Would you be interested to have them converted to StringBuilder too, at least for internal classes and methods?

HannesWell:
> As long as patches are small like this one, it is OK. But I personally do not see much value in this specific kind of changes, usually they don't provide much value from performance point of view.

It won't be this small, I estimate that it would affect about 100 classes, but each will only contain a few replacements.
It isn't possible to make this smaller since those print methods are implemented in many sub-classes.
The performance gain wouldn't be huge but probably a few percent, depending on the use case:
https://www.baeldung.com/java-string-builder-string-buffer

PR #1403, opened 2023-09-20
Use StringBuilder instead of StringBuffer for ASTNode.print*()

HannesWell:
## What it does
Replace all usages of `StringBuffer` in the signature and implementations of `ASTNode.print*()` by StringBuilder since the code seems to be purely single-threaded and `StringBuilder` is generally slightly faster since it is not synchronized unlike `StringBuffer`.

The changes were not made using a JDT clean-up, but by changing the root `ASTNode.print()` to use `StringBuilder` instead of `StringBuffer` and subsequently fixing all compiler errors.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
+++ b/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
@@ -16,7 +16,6 @@
 package org.eclipse.jdt.core.dom;
 
 import java.io.PrintWriter;
-import java.io.StringWriter;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
@@ -1136,11 +1135,8 @@ public class ASTParser {
 								}
 							} catch(JavaModelException e) {
 								// an error occured accessing the java element
-								StringWriter stringWriter = new StringWriter();
-								try (PrintWriter writer = new PrintWriter(stringWriter)) {
-									e.printStackTrace(writer);
-								}
-								throw new IllegalStateException(String.valueOf(stringWriter.getBuffer()));
+								CharSequence stackTrace = org.eclipse.jdt.internal.compiler.util.Util.getStackTrace(e);
+								throw new IllegalStateException(stackTrace.toString());
 							}
 						}
 					}
@@ -1206,11 +1202,8 @@ public class ASTParser {
 							sourceUnit = new BasicCompilationUnit(sourceString.toCharArray(), Util.toCharArrays(packageFragment.names), fileNameString, this.typeRoot);
 						} catch(JavaModelException e) {
 							// an error occured accessing the java element
-							StringWriter stringWriter = new StringWriter();
-							try (PrintWriter writer = new PrintWriter(stringWriter)) {
-								e.printStackTrace(writer);
-							}
-							throw new IllegalStateException(String.valueOf(stringWriter.getBuffer()));
+							CharSequence stackTrace = org.eclipse.jdt.internal.compiler.util.Util.getStackTrace(e);
+							throw new IllegalStateException(stackTrace.toString());
 						}
 					} else if (this.rawSource != null) {
 						needToResolveBindings =
--- a/org.eclipse.jdt.apt.core/META-INF/MANIFEST.MF
+++ b/org.eclipse.jdt.apt.core/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jdt.apt.core; singleton:=true
-Bundle-Version: 3.8.200.qualifier
+Bundle-Version: 3.8.300.qualifier
 Bundle-Localization: plugin
 Export-Package: com.sun.mirror.apt,
  com.sun.mirror.declaration,
--- a/org.eclipse.jdt.apt.core/pom.xml
+++ b/org.eclipse.jdt.apt.core/pom.xml
@@ -17,6 +17,6 @@
     <version>4.31.0-SNAPSHOT</version>
   </parent>
   <artifactId>org.eclipse.jdt.apt.core</artifactId>
-  <version>3.8.200-SNAPSHOT</version>
+  <version>3.8.300-SNAPSHOT</version>
   <packaging>eclipse-plugin</packaging>
 </project>
--- a/org.eclipse.jdt.apt.core/src/org/eclipse/jdt/apt/core/internal/AptPlugin.java
+++ b/org.eclipse.jdt.apt.core/src/org/eclipse/jdt/apt/core/internal/AptPlugin.java
@@ -195,7 +195,7 @@ public class AptPlugin extends Plugin implements DebugOptionsListener {
 
 	public static void trace(final String msg){
 		if (DEBUG) {
-			StringBuffer sb = new StringBuffer();
+			StringBuilder sb = new StringBuilder();
 			sb.append('[');
 			// SimpleDateFormat is not thread-safe, according to javadoc
 			synchronized(TRACE_DATE_FORMAT) {
--- a/org.eclipse.jdt.apt.core/src/org/eclipse/jdt/apt/core/internal/declaration/AnnotationValueImpl.java
+++ b/org.eclipse.jdt.apt.core/src/org/eclipse/jdt/apt/core/internal/declaration/AnnotationValueImpl.java
@@ -206,7 +206,7 @@ public class AnnotationValueImpl implements EclipseMirrorObject, AnnotationValue
 			return "null"; //$NON-NLS-1$
 		} else if (_value instanceof String) {
 			String value = (String) _value;
-			StringBuffer sb = new StringBuffer();
+			StringBuilder sb = new StringBuilder();
 			sb.append('"');
 			for (int i = 0; i < value.length(); i++) {
 				Util.appendEscapedChar(sb, value.charAt(i), true);
@@ -214,7 +214,7 @@ public class AnnotationValueImpl implements EclipseMirrorObject, AnnotationValue
 			sb.append('"');
 			return sb.toString();
 		} else if (_value instanceof Character) {
-			StringBuffer sb = new StringBuffer();
+			StringBuilder sb = new StringBuilder();
 			sb.append('\'');
 			Util.appendEscapedChar(sb, ((Character) _value).charValue(), false);
 			sb.append('\'');
--- a/org.eclipse.jdt.apt.pluggable.core/META-INF/MANIFEST.MF
+++ b/org.eclipse.jdt.apt.pluggable.core/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jdt.apt.pluggable.core;singleton:=true
-Bundle-Version: 1.4.200.qualifier
+Bundle-Version: 1.4.300.qualifier
 Bundle-Activator: org.eclipse.jdt.internal.apt.pluggable.core.Apt6Plugin
 Bundle-Vendor: %providerName
 Require-Bundle: org.eclipse.core.runtime,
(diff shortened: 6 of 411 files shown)


vogella:
@jukzi could you review?

jukzi:
The commit LGTM. Since the only reason seems to be for performance: please document a use case and measurement where the change is an actual performance improvement. That would also be good as an indication if such optimization would be worth on other places too.  For example https://alblue.bandlem.com/2016/04/jmh-stringbuffer-stringbuilder.html indicates that the difference can be in general below measurement error:
image

vogella:
Javadoc says that StringBuilder should used in case of StringBuffer in used
in thread-safe code. This should be sufficient to justify such changes.
Also the existing JDT clean-up action for this should be an indication that
jdt should use the new version.

Jörg Kubitz ***@***.***> schrieb am Fr., 20. Okt. 2023, 08:22:

jukzi:
I am sure it's not bad but the question is if it's worth to have a heavy blame change and further investigations.

vogella:
This was clarified long time ago in the pmc. Code clean-ups are more
important than Git blame.

Please do not open this can of worms again, the discussions in the past of
every single clean-up (Is it really worth it? Please prove that this is
better, Should you not focus on something important? Why should I waste my
time in reviewing it, What if this clean-up breaks something, Why change
working code without good reason, Is this really helping in getting new
contributors) let to huge frustration in the community in the past on both
sides (those who dislike it and those who think Eclipse should be open to
these changes).

So the rule from PMC is clear: clean-ups are welcome.

Jörg Kubitz ***@***.***> schrieb am Fr., 20. Okt. 2023, 11:43:

iloveeclipse:
> This was clarified long time ago in the pmc

Nice for pmc. But as long as pmc respects committer rights, it's at the project committers to decide about every code contribution. Or will pmc now review, approve and merge every "cleanup" PR in JDT, along with fixing bugs later on?

> Please do not open this can of worms again, the discussions in the past of
every single clean-up 

So that is now the "welcome" words from pmc, right? "Welcome to Eclipse, but please don't discuss here". Good to know.

> (Is it really worth it? Please prove that this is
better, Should you not focus on something important? Why should I waste my
time in reviewing it, What if this clean-up breaks something, Why change
working code without good reason, Is this really helping in getting new
contributors) let to huge frustration in the community in the past on both
sides (those who dislike it and those who think Eclipse should be open to
these changes).

1) All questions above are still valid for **every** pr.
2) There is a difference between a simple pet project where git blame is irrelevant and a project with very complex code that is used by many where code changes are hard to understand and so git blame is very important.
3) pmc can recommend but decisions are made by committers. So if a committer has valid concerns, he is free to decide against merging a pr.

vogella:
In this case, the change was reviewed by a JDT committers and Jörg only
raised concerns about Git blame. This is what was clarified a long time ago
by the pmc.

Please note that one of the pmc leaders is also a JDT project lead.

Andrey Loskutov ***@***.***> schrieb am Fr., 20. Okt. 2023,
12:59:

jukzi:
> Please note that one of the pmc leaders is also a JDT project lead.

Then he might submit? i don't raise a -1, i only do not feel confident to submit this particular PR. It's not an automated cleanup but a manual one. Automated JDT cleanups would have another justification: to check if the clean up works. Please bear in mind that JDT is much more sensitive about blame and performance then other IDE projects for good reasons. I would be really nice to just show any performance test to end this PR with a fruitable result. Would be odd to celebrate a performance improvement if there is none.
Another way would be to use the JDT cleanup instead - next M1. I can not remember that JDT uses StringBuffer anywhere across Threads.

jukzi:
I wonder why this code was not part of https://github.com/eclipse-jdt/eclipse.jdt.core/commit/9caccf56f3e879b209fa461aabf64754fac603ca
see https://bugs.eclipse.org/bugs/show_bug.cgi?id=574532#c8 ff

HannesWell:
> The commit LGTM. Since the only reason seems to be for performance: please document a use case and measurement where the change is an actual performance improvement.

Actually my main motivation was also to apply the recommendation from the `StringBuffer` Java-Doc that Lars mentioned, hoping for a little bit better performance. Therefore I don't have a specific use-case that I can measure.
But as mentioned in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1395#issuecomment-1728301012, the performance gain will probably quite small and it will also depend on the specific JVM.


> Another way would be to use the JDT cleanup instead - next M1. I can not remember that JDT uses StringBuffer anywhere across Threads.

Understand.
I re-did the change by applying the `Convert StringBuffer to StringBuilder` clean-up (btw. it is listed under `Performance`) to all projects in the `jdt-core` and `jdt-core-tests` working set, which includes all previous changes (according to git), so at the places considered before the result is the same.
Just applying the clean-up to all those projects lead to a compilation error in `org.eclipse.jdt.internal.compiler.util.Util` since in
```
StringWriter stringWriter = new StringWriter();
exception.printStackTrace(new PrintWriter(stringWriter));
StringBuffer buffer = stringWriter.getBuffer();
```
only `StringBuffer buffer`  is replaced by `StringBuilder buffer`, but of course `StringWriter` still returns a `StringBuffer`.

In order fix that error I created another commit (re-order as first one) that unifies printing exception stack-traces to String without relying on `StringBuffer` and applied the newly created Util method in the jdt.core wherever I found similar patterns.

jjohnstn:
@HannesWell I have created: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/869 to fix the bug you mention about the snippet using StringWriter.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
```diff
@@ -1136,11 +1135,8 @@ private ASTNode internalCreateAST(IProgressMonitor monitor) {
 								}
 							} catch(JavaModelException e) {
 								// an error occured accessing the java element
-								StringWriter stringWriter = new StringWriter();
-								try (PrintWriter writer = new PrintWriter(stringWriter)) {
-									e.printStackTrace(writer);
-								}
-								throw new IllegalStateException(String.valueOf(stringWriter.getBuffer()));
+								CharSequence stackTrace = org.eclipse.jdt.internal.compiler.util.Util.getStackTrace(e);
```
jukzi, line 1138: was this automated cleanup? looks odd to not use an import here.

HannesWell:
> @HannesWell I have created: [eclipse-jdt/eclipse.jdt.ui#869](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/869) to fix the bug you mention about the snippet using StringWriter.

Thank you.

I just pushed a little enhancement to the `Util.getStackTrace()` method and override write(String, int, int), which is the only method called by `PrintWriter` and overriding it saves the intermediate charbuffer creation and copying to it.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
HannesWell: No all in the first commit was done by hand. But in this class there is already
 `import org.eclipse.jdt.internal.core.util.Util;`
which is why `org.eclipse.jdt.internal.compiler.util.Util` cannot be imported.

HannesWell:
@jukzi can you tell what's the plan for this one?
Since it is now a clean-up application do you want to keep this open until after the next release?

jukzi:
> Since it is now a clean-up application do you want to keep this open until after the next release?

yes, it's better to wait with commit for M1, since its untested - just in case.
Note: normally we go for single commit per PR, but LGTM.

HannesWell:
> > Since it is now a clean-up application do you want to keep this open until after the next release?
> 
> yes, it's better to wait with commit for M1, since its untested - just in case.

Acknowledge.

> Note: normally we go for single commit per PR, but LGTM.

If you prefer I can also split out the first commit, just let me know.

jukzi:
org.eclipse.jdt.internal.core.JavaElement.escapeMementoName() uses StringBuffer and is heavy used when pressing strg+t on java.lang.Object
image
and drops from 1 sec to 300ms when using StringBuilder in that case

vogella:
> org.eclipse.jdt.internal.core.JavaElement.escapeMementoName() uses StringBuffer and is heavy used when pressing strg+t on java.lang.Object image and drops from 1 sec to 300ms when using StringBuilder in that case

Great to see this confirmation that StringBuilder is faster. General remark: Would be nice go see JDT apply there existing clean-ups to leverage performance improvements in more areas. A few milliseconds here and there could add up and would improve Eclipse responsiveness. Also trying to use the JDT clean-ups in JDT would help to improve the clean-ups.

jukzi (approved):
lgtm

jukzi:
Please add a version bump commit
https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-1403/7/console
```
13:52:22  [ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.4:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.jdt.core: Only qualifier changed for (org.eclipse.jdt.core/3.36.0.v20231128-1847). Expected to have bigger x.y.z than what is available in baseline (3.36.0.v20231115-1055) -> [Help 1]
13:52:22  [ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.4:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.jdt.compiler.apt.tests: Only qualifier changed for (org.eclipse.jdt.compiler.apt.tests/1.3.200.v20231128-1847). Expected to have bigger x.y.z than what is available in baseline (1.3.200.v20231114-0937) -> [Help 1]
```

HannesWell:
> Please add a version bump commit
> https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-1403/7/console

Done. The build is finally green.

jukzi:
Thanks for improving JDT, please check next integration builds for regression if any.

vogella:
Great to see this finally happening, the first attempt from 2017 to do this was unfortunately unsuccessful, see https://bugs.eclipse.org/bugs/show_bug.cgi?id=516844

Big thanks to @HannesWell and @jukzi

jukzi:
As stated in https://bugs.eclipse.org/bugs/show_bug.cgi?id=516844#c28 such `should be done with proper measurements first`

vogella:
@jukzi which you did which made the difference, thanks for that,  instead of blocking until someone else fulfilled your requirement.

rgrunber:
@gayanper mentioned this PR breaks lombok support for him, https://github.com/eclipse-jdtls/eclipse.jdt.ls/issues/2985 . Looking at https://github.com/projectlombok/lombok/blob/master/src/core/lombok/eclipse/handlers/EclipseHandlerUtil.java#L1333-L1336 that seems to be the case.

If the performance benefit is really worth it, and given it's internal API, maybe lombok can update, but this basically prevents JDT-LS from updating until addressed.

akurtakov:
@rgrunber what's the relation between jdt.ls and lombok?

rgrunber:
Not JDT-LS, but clients such as vscode-java have Lombok support via. https://github.com/redhat-developer/vscode-java/blob/master/src/lombokSupport.ts by adding it to `-agentlib`.

jukzi:
When lombook uses JDT internals it's not our problem
```
java.lang.NoSuchMethodError: 'java.lang.StringBuffer org.eclipse.jdt.internal.compiler.ast.Expression.print(int, java.lang.StringBuffer)'
	at lombok.eclipse.handlers.EclipseHandlerUtil.createAnnotation(EclipseHandlerUtil.java:1335)
```

vogella:
@rgrunber thanks for opening the bug in Lombok.

HannesWell:
> Not JDT-LS, but clients such as vscode-java have Lombok support via. https://github.com/redhat-developer/vscode-java/blob/master/src/lombokSupport.ts by adding it to `-agentlib`. Basically we could update JDT-LS, but that would leave Lombok support broken in vscode-java until they updated.

I agree with Jörg, we cannot consider every external project that is using internals. 
It is of course undesirable to break other projects, but if we would consider all external that use internals, we would have to stop development because nothing could be changed anymore, even a bug-fix could break someone external project.
Others should not use internals and if they do and it breaks it is their fault. If they really need something stable, they should propose and discuss how to build a good API for for it.

But I proposed a solution that should be even backwards compatible in https://github.com/projectlombok/lombok/issues/3564#issuecomment-1841230003.

vogella:
@HannesWell thanks, very impressive that you got involved in solving the problem of Lombok. I think you set a new standards for "taking responsibility".

jukzi:
> Another way would be to use the JDT cleanup instead - next M1. I can not remember that JDT uses StringBuffer anywhere across Threads.

i was wrong: org.eclipse.jdt.core.tests.util.TestVerifier.outputBuffer, errorBuffer are used concurrently (read in another thread then written), which is not guaranteed to work with StringBuilder, i will fix that

Issue #1330, opened 2023-08-30
AIOOBE in ReferenceExpression.sIsMoreSpecific when saving incomplete statement


HeikoKlare:
### Issue
Saving/compiling a Java file with the following content leads to an `ArrayIndexOutOfBoundsException`:
```java
import java.io.File;

public class JDTIssue {
	private  void indexOfOfBounds() {
		new File("").listFiles(File::isDirectory)
	}
}
```
Note that the statement is incomplete, as the semicolon is missing.

### Initial Analysis
`File::listFiles()` is polymorphic and has two 1-argument implementations that accept either a `FilenameFilter` or a `FileFilter`.
These interfaces look as follows:
```java
public interface FilenameFilter {
    boolean accept(File dir, String name);
}
public interface FileFilter {
    boolean accept(File pathname);
}
```

While searching for the correct method binding (which in this case is the method accepting a `FileFilter`), the  method `ReferenceExpression::sIsMoreSpecific()` is executed with type bindings for the two `accept` methods of `FileNameFilter` and `FileFilter`. It compares the parameter lists of both `accept` methods and fails with an AIOOBE due to different sizes of the parameter lists, although the code should only be reached in case the parameter lists have equal length:
https://github.com/eclipse-jdt/eclipse.jdt.core/blob/816bc73d6b044d55043601b9e3a4e00f08551df4/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ReferenceExpression.java#L1302-L1306

The stack when the exception occurs looks as follows:
```
ReferenceExpression.sIsMoreSpecific(TypeBinding, TypeBinding, Scope) line: 1304	
PolyTypeBinding.sIsMoreSpecific(TypeBinding, TypeBinding, Scope) line: 84	
MethodScope(Scope).mostSpecificMethodBinding(MethodBinding[], int, TypeBinding[], InvocationSite, ReferenceBinding) line: 4659	
MethodScope(Scope).findDefaultAbstractMethod(ReferenceBinding, char[], TypeBinding[], InvocationSite, ReferenceBinding, ObjectVector, MethodBinding[]) line: 1310	
MethodScope(Scope).findMethod0(ReferenceBinding, char[], TypeBinding[], InvocationSite, boolean) line: 1938	
MethodScope(Scope).findMethod(ReferenceBinding, char[], TypeBinding[], InvocationSite, boolean) line: 1668	
[...]

PR #1419, opened 2023-09-22
AIOOBE in ReferenceExpression.sIsMoreSpecific when saving incomplete …

ShahzaibIbrahim:
…statement #1330

While searching for the correct method binding, the method ReferenceExpression::sIsMoreSpecific() is executed with type bindings for the two accept methods of FileNameFilter and FileFilter.

Fix: Added a check before iterating over the arguments for both message to avoid AIOOBE.

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1330

## Author checklist


iloveeclipse:
Thanks for the patch! Could you provide a regression test too?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ReferenceExpression.java
```diff
@@ -1300,9 +1300,13 @@ public boolean sIsMoreSpecific(TypeBinding s, TypeBinding t, Scope scope) {
 		TypeBinding[] sParams = sSam.parameters;
 		TypeBinding[] tParams = tSam.parameters;
 		// Both must have the same number of parameters if we got this far
-		for (int i = 0; i < sParams.length; i++) {
-			if (TypeBinding.notEquals(sParams[i], tParams[i]))
-				return false;
+		if(sParams.length == tParams.length) {
```
stephan-herrmann, line 1303: When Sasikanth wrote this loop, he was convinced "Both must have the same number of parameters if we got this far". Unfortunately he is no longer on the team, so we can't ask him why he thought so.

At least we should have an example that tells us, how Sasi's assumption can be violated.

A brief look at the spec lets me think if perhaps one of the methods is varargs, in which case returning `false` might be wrong, as the spec has specific treatment for varargs methods.

Additionally, at first glance the corresponding method in `LambdaExpression` might have the same issue.


@srikanth-sankaran  fyi.

stephan-herrmann: JLS §18.5.4. refers to §15.12.2.5 for this part.

In JLS version 8 (from which this method was first implemented) I found: 
> (where U1 ... Uk and R1 are the parameter types and return type of the function type of the capture of S, and V1 ... Vk and R2 are the parameter types and return type of the function type of T)

To me this seems to say that also JLS 8 assumed both function types to have the same number of parameters (k).

Interestingly, this parenthesis can no longer be found in JLS 20. But still in version 20, the definitions of ` P1, ..., Pn` and `P1', ..., Pn'` use the same `n`.

Looking at the part specific to method reference expressions, JLS 8 defined `for all i (1 ≤ i ≤ k), Ui is the same as Vi,`. Very interestingly, I cannot find anything like that in JLS 20. If that was an intentional deletion in JLS, then also we should deleted the code comparing parameter types.

For clarity, I'm referring to this:
[JLS 8:](https://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.12.2.5)
> If e is an exact method reference expression ([§15.13.1](https://docs.oracle.com/javase/specs/jls/se8/html/jls-15.html#jls-15.13.1)), then i) for all i (1 ≤ i ≤ k), Ui is the same as Vi, and ii) one of the following is true: ...

[JLS 20:](https://docs.oracle.com/javase/specs/jls/se20/html/jls-15.html#jls-15.12.2.5)
> e is an exact method reference expression ([§15.13.1](https://docs.oracle.com/javase/specs/jls/se20/html/jls-15.html#jls-15.13.1)), and one of the following is true: ...

ShahzaibIbrahim:
> Thanks for the patch! Could you provide a regression test too?

Do we really need the regression test since we are deleting this part of code?

iloveeclipse:
> Do we really need the regression test since we are deleting this part of code?

Well, to confirm that someone would not break the use case again (by adding same code again for example, or just by any other change), regression test is needed.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ReferenceExpression.java
ShahzaibIbrahim: Is removing the code right solution? As it sill works for other versions like 1.8 and I checked the tests are failing.

srikanth-sankaran: > Is removing the code right solution? As it sill works for other versions like 1.8 and I checked the tests are failing.

Look at the uses of `org.eclipse.jdt.internal.compiler.impl.CompilerOptions.complianceLevel` in the compiler. By checking the compliance levels, it is possible to conditionally alter the behavior. Eventually, the behavior should match corresponding javac versions (ideally unless it is accepted to a defective unhelpful behavior) and should come along with a regression test that establishes appropriate behavior at different JDK levels

ShahzaibIbrahim:
image
> Well, to confirm that someone would not break the use case again (by adding same code again for example, or just by any other change), regression test is needed.

I am trying to write a test case to generate this error (pop-window). Has anyone any idea how to capture this pop-up window ( I can't find any similar test in project )
The case is simple upon compilation of this particular code block, this pop-up window shouldn't be shown. If it shows, test fails.
image.png
 
image

srikanth-sankaran:
> > I am trying to write a test case to generate this error (pop-window).
> 
> This is "too late" for the test, as the fix is in jdt.core, the test should be there too (not in UI layer).

Take a look at org.eclipse.jdt.core.tests.compiler.regression.LambdaExpressionsTest and 
org.eclipse.jdt.core.tests.compiler.regression.NegativeLambdaExpressionsTest .

ShahzaibIbrahim:
> > > I am trying to write a test case to generate this error (pop-window).
> > 
> > 
> > This is "too late" for the test, as the fix is in jdt.core, the test should be there too (not in UI layer).
> 
> Take a look at org.eclipse.jdt.core.tests.compiler.regression.LambdaExpressionsTest and org.eclipse.jdt.core.tests.compiler.regression.NegativeLambdaExpressionsTest . You want to write a test that compiles with no incident while without your fix, it crashes with AIOOB.

Without my fix, the error is resolved into this popup and there is nothing in the console. I tried to find a similar example in both the files you mentioned. The AIOOBE is resolved and its shown on a pop-up dialogue box not in console. All I get is 

`"----------\n" +
	"1. ERROR in X.java (at line 4)\n" +
	"	new File(\"\").listFiles(File::isDirectory)\n" +
	"	                                        ^\n" +
	"Syntax error, insert \";\" to complete BlockStatemens\n" +
	"----------\n");`

srikanth-sankaran:
> > > > I am trying to write a test case to generate this error (pop-window).
> > > 
> > > 
> > > This is "too late" for the test, as the fix is in jdt.core, the test should be there too (not in UI layer).
> > 
> > 
> > Take a look at org.eclipse.jdt.core.tests.compiler.regression.LambdaExpressionsTest and org.eclipse.jdt.core.tests.compiler.regression.NegativeLambdaExpressionsTest . You want to write a test that compiles with no incident while without your fix, it crashes with AIOOB.
> 
> Without my fix, the error is resolved into this popup and there is nothing in the console. I tried to find a similar example in both the files you mentioned. The AIOOBE is resolved and its shown on a pop-up dialogue box not in console. All I get is
> 
> `"----------\n" + "1. ERROR in X.java (at line 4)\n" + " new File(\"\").listFiles(File::isDirectory)\n" + " ^\n" + "Syntax error, insert \";\" to complete BlockStatemens\n" + "----------\n");`

The first order business is to write a test even before attempting a fix. JDT/Core tests are run in "headless" mode i.e the UI is not running when the automated tests are running (mostly). 

There is some mystery here - I haven't read through all the comments here and elsewhere - but the missing semicolon seems necessary to reproduce the problem - Why ?? Did you find out ?

Also if I add a regression test in LET with the snippet from above, with or without the missing semicolon I don't see any AIOOB - this needs to be analyzed (debugged) to fully understand the problem landscape. 

So we do have one interactive way of reproducing the problem - why does the problem vanish when adding a junit ?? Can you study the difference by debugging and comparing ?

srikanth-sankaran:
(If after trying, you are still not able to make headway, let me know - I can spend some cycles on unblocking you next week)

ShahzaibIbrahim:
> > > > > I am trying to write a test case to generate this error (pop-window).
> > > > 
> > > > 
> > > > This is "too late" for the test, as the fix is in jdt.core, the test should be there too (not in UI layer).
> > > 
> > > 
> > > Take a look at org.eclipse.jdt.core.tests.compiler.regression.LambdaExpressionsTest and org.eclipse.jdt.core.tests.compiler.regression.NegativeLambdaExpressionsTest . You want to write a test that compiles with no incident while without your fix, it crashes with AIOOB.
> > 
> > 
> > Without my fix, the error is resolved into this popup and there is nothing in the console. I tried to find a similar example in both the files you mentioned. The AIOOBE is resolved and its shown on a pop-up dialogue box not in console. All I get is
> > `"----------\n" + "1. ERROR in X.java (at line 4)\n" + " new File(\"\").listFiles(File::isDirectory)\n" + " ^\n" + "Syntax error, insert \";\" to complete BlockStatemens\n" + "----------\n");`
> 
> The first order business is to write a test even before attempting a fix. JDT/Core tests are run in "headless" mode i.e the UI is not running when the automated tests are running (mostly).
> 
> There is some mystery here - I haven't read through all the comments here and elsewhere - but the missing semicolon seems necessary to reproduce the problem - Why ?? Did you find out ?
> 
> Also if I add a regression test in LET with the snippet from above, with or without the missing semicolon I don't see any AIOOB - this needs to be analyzed (debugged) to fully understand the problem landscape.
> 
> So we do have one interactive way of reproducing the problem - why does the problem vanish when adding a junit ?? Can you study the difference by debugging and comparing ?

Yes I tried to debug both the cases. 
Complete statement with semi-colon added: It doesn't consider FileNameFilter Candidate for the closest match
Without semi-colon: It consider both FileNameFilter and FileFilter candidates for closest match and that's why sIsMoreSpecificMethod was called and exception was thrown. 

I am not entirely sure why it is not resolved earlier (before sIsMoreSpecific Method is called.

stephan-herrmann:
@srikanth-sankaran  I would be surprised if the change in JLS consisted only in deleting this one sentence. Shouldn't we check the vicinity and refresh our implementation from current JLS?
BTW, are you aware of any channel how we could get informed about every change in JLS? Experience tells me, that not all changes go the route via an active JEP. Diffing the public spec after the fact is no fun.

srikanth-sankaran:
> > > > > > I am trying to write a test case to generate this error (pop-window).
> > > > > 
> > > > > 
> > > > > This is "too late" for the test, as the fix is in jdt.core, the test should be there too (not in UI layer).
> > > > 
> > > > 
> > > > Take a look at org.eclipse.jdt.core.tests.compiler.regression.LambdaExpressionsTest and org.eclipse.jdt.core.tests.compiler.regression.NegativeLambdaExpressionsTest . You want to write a test that compiles with no incident while without your fix, it crashes with AIOOB.
> > > 
> > > 
> > > Without my fix, the error is resolved into this popup and there is nothing in the console. I tried to find a similar example in both the files you mentioned. The AIOOBE is resolved and its shown on a pop-up dialogue box not in console. All I get is
> > > `"----------\n" + "1. ERROR in X.java (at line 4)\n" + " new File(\"\").listFiles(File::isDirectory)\n" + " ^\n" + "Syntax error, insert \";\" to complete BlockStatemens\n" + "----------\n");`
> > 
> > 
> > The first order business is to write a test even before attempting a fix. JDT/Core tests are run in "headless" mode i.e the UI is not running when the automated tests are running (mostly).
> > There is some mystery here - I haven't read through all the comments here and elsewhere - but the missing semicolon seems necessary to reproduce the problem - Why ?? Did you find out ?
> > Also if I add a regression test in LET with the snippet from above, with or without the missing semicolon I don't see any AIOOB - this needs to be analyzed (debugged) to fully understand the problem landscape.
> > So we do have one interactive way of reproducing the problem - why does the problem vanish when adding a junit ?? Can you study the difference by debugging and comparing ?
> 
> Yes I tried to debug both the cases. Complete statement with semi-colon added: It doesn't consider FileNameFilter Candidate for the closest match Without semi-colon: It consider both FileNameFilter and FileFilter candidates for closest match and that's why sIsMoreSpecificMethod was called and exception was thrown.
> 
> I am not entirely sure why it is not resolved earlier (before sIsMoreSpecific Method is called.

You need to peel the next layer and investigate why in one case it doesn't consider a particular candidate while in the other it does

HeikoKlare:
> You need to peel the next layer and investigate why in one case it doesn't consider a particular candidate while in the other it does

Yes, finding the root cause for the issue is essential to properly solve it. I just wanted to let you know that we (@ShahzaibIbrahim, @fedejeanne, @HeikoKlare) had to push the issue a bit down in our backlog since it seems to require quite some additional effort (at least for us not being so deep into JDT yet). But we did not drop the topic and will proceed work as soon as possible.
In case someone else wants to proceed meanwhile, feel free to do so.

jukzi:
This PR still misses a Junit Test

stephan-herrmann:
> Without my fix, the error is resolved into this popup and there is nothing in the console. I tried to find a similar example in both the files you mentioned. The AIOOBE is resolved and its shown on a pop-up dialogue box not in console. All I get is
> 
> `"----------\n" + "1. ERROR in X.java (at line 4)\n" + " new File(\"\").listFiles(File::isDirectory)\n" + " ^\n" + "Syntax error, insert \";\" to complete BlockStatemens\n" + "----------\n");`

Using the example from #1330 I cannot reproduce using the batch compiler, so regular compilation is not sufficient.

The stacktrace in the issue, however, indicates that the exception was raised while requesting DOM AST.
Thus to reproduce the problem the test should probably be a DOM test, like those you can find in ASTConverter_17Test or similar.

PR #1446, opened 2023-09-28
Eclipse reports Type mismatch for @Deprecated class

### LINKED ISSUE #1412: Eclipse reports Type mismatch for @Deprecated class
We have a situation that can be simplified to the following case:

1.) An abstract class with a generic type:
```
public abstract class AbstractClass<T> {}
```

2.) An annotation interface that references that class:
```
public @interface AnnotationWithClassValue {
	Class<? extends AbstractClass<?>> value();
}
```

3.) Two classes referencing each other:
3.a) A concrete class extending the abstract class:
```
//Adding @Deprecated here fixes the bug
//@Deprecated
public class ConcreteClass extends AbstractClass<AnnotatedClass> {}
```

3.b) A class annotated with the annotation that is marked as deprecated:
```
@Deprecated
@AnnotationWithClassValue(ConcreteClass.class) //Type mismatch: cannot convert from Class<ConcreteClass> to Class<? extends AbstractClass<?>>
public class AnnotatedClass {}
```

Editing any of the classes clears the error till the next clean build, which restores it. Marking the `ConcreteClass` as `@Deprecated` also resolves the issue.

The problem appeared when we moved a project from Java 8 to Java 11, remains with Java 17 and vanishes when returning the project to Java 8.

I can reproduce the error in 4.28 and 4.29.

I assume this might be related to  #1399 and/or #189

stephan-herrmann:
fixes #1412

## What it does

Defer problem reporting if it would require ahead-of-time resolving of annotations (tasks stored in `CompilationResult`).
Materialize such problems just at the beginning of `CompilationUnitDeclaration.finalizeProblems()`.

## How to test

JUnit included

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/problem/ProblemReporter.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/problem/ProblemReporter.java
@@ -1957,6 +1957,8 @@ public void deprecatedType(TypeBinding type, ASTNode location) {
 public void deprecatedType(TypeBinding type, ASTNode location, int index) {
 	if (location == null) return; // 1G828DN - no type ref for synthetic arguments
 	final TypeBinding leafType = type.leafComponentType();
+	if (!leafType.isReadyForAnnotations() && scheduleProblemForContext(() -> deprecatedType(type, location, index)))
+		return;
 	int sourceStart = -1;
 	if (location instanceof QualifiedTypeReference) { // https://bugs.eclipse.org/bugs/show_bug.cgi?id=300031
 		QualifiedTypeReference ref = (QualifiedTypeReference) location;
@@ -12452,5 +12454,22 @@ public void falseLiteralInGuard(Expression exp) {
 			NoArgument,
 			exp.sourceStart,
 			exp.sourceEnd);
+}
+public boolean scheduleProblemForContext(Runnable problemComputation) {
+	if (this.referenceContext != null) {
+		CompilationResult result = this.referenceContext.compilationResult();
+		if (result != null) {
+			result.scheduleProblem(() -> {
+				ReferenceContext save = this.referenceContext;
+				try {
+					problemComputation.run();
+				} finally {
+					this.referenceContext = save;
+				}
+			});
+			return true;
+		}
+	}
+	return false;
 }
 }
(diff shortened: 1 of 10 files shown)


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/problem/ProblemReporter.java
```diff
@@ -12453,4 +12458,40 @@ public void falseLiteralInGuard(Expression exp) {
...
+ * <li>individual methods should ensure that {@link #referenceContext} is reset to {@code null} afterwards. This may
+ * happen in two ways:
+ * <ul><li>by calling any of the {@code handle(..)} methods on all paths, or failing that
+ * <li>wrap their body in {@code try(this) { ... }}.
+ * </ul></ul>
+ */
+@Override
+public void close() {
+	this.referenceContext = null;
+}
```
stephan-herrmann, line 12496: @srikanth-sankaran @jarthana One more change that tries to make an implicit strategy more explicit. Do you want to comment on changes outlined in the above javadoc?

I briefly thought of actually pulling all this into a new method in `Scope`, to replace all those `scope.getProblemReporter().reportXYZ(foo,bar)` calls with smth like this:
```
scope.withContext(pr -> pr.reportXYZ(foo,bar));
```
This would encapsulate that strategy even better, but it would require a million of edits all over the place ;-P -- so it may be no more than a theoratical exercise ...

stephan-herrmann: I should mention that this change was motivated by the observation that (a) my new method `scheduleProblemForContext()` was implemented incompletely, and (b) nobody noticed, because we accidentally benefit from other methods being sloppy wrt `ProblemReporter#referenceContext`, notably all those calls to `problemReporter().validateRestrictedKeywords()`, which normally never reset the context.

stephan-herrmann: > @srikanth-sankaran @jarthana One more change that tries to make an implicit strategy more explicit. Do you want to comment on changes outlined in the above javadoc?

Of course I will accept your comments any time, despite having merged this PR.

PR #1449, opened 2023-09-29
Clean Up: Add final modifier to private fields

jukzi:


jarthana:
This introduced a new warning in `TypeDeclarationPattern`. The field `allowModuleRegex `is final and set to false at definition and never changed again.
@jukzi Can you take care of this please?

jukzi:
> This introduced a new warning in `TypeDeclarationPattern`

see https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1544

PR #1455, opened 2023-10-01
Add null guard for scope in SourceTypeBinding.storeAnnotations

kriegaex:
As discussed with @laeubi in https://github.com/eclipse-jdt/eclipse.jdt.core/commit/8d512be7b8594e8c6dfefff06b36a0b95d7250d7#r128469112, when using JDT Core from the AspectJ compiler, we ran into NPEs in this place. This is difficult to reproduce without AspectJ, but when looking on how and when `this.scope` is initialised, we see that it is not checked for null in the constructor. Furthermore, it is set to null in the public `cleanUp()` method. I.e., chances are, that it can be null in some JDT Core situations, too, even without AspectJ in the mix.

## Author checklist

- [ ] I have thoroughly tested my changes → No, I made the changes in a web editor. I did not want to build locally. Please either review and test this change by yourself or discard it. I just mean to help a little bit. I made the same change in the AspectJ fork of JDT Core already, using it in the AspectJ compiler.

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/SourceTypeBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/SourceTypeBinding.java
@@ -3343,7 +3343,8 @@ SimpleLookupTable storedAnnotations(boolean forceInitialize, boolean forceStore)
 @Override
 void storeAnnotations(Binding binding, AnnotationBinding[] annotations, boolean forceStore) {
 	super.storeAnnotations(binding, annotations, forceStore);
-	this.scope.referenceCompilationUnit().compilationResult.annotations.add(annotations);
+	if (this.scope != null)
+		this.scope.referenceCompilationUnit().compilationResult.annotations.add(annotations);
 }
 
 @Override


stephan-herrmann:
This guard doesn't look like a good idea to me, for the following reasons:

1. When no scope is present, the effect of `storeAnnotations()` would be crippled, i.e., the feature from https://github.com/eclipse-jdt/eclipse.jdt.core/pull/676 would be broken in such cases. I'm not a fan of letting the compiler silently skip some of its work.
2. The problem "shouldn't occur" based on ecj's design of compilation phases (see also https://github.com/eclipse-jdt/eclipse.jdt.core/wiki/ECJ#phases)

To add a few more details to (2): 

Instantiation of `SourceTypeBinding` is always initiated from a scope passing `this` as the scope. There hardly is a more reliable non-null reference than `this` :)

Calls to `storeAnnotations()` happen during `completeTypeBindings` or `resolve_`. During those phases we assume non-null scopes all over the place, we cannot compile without them, so adding null checks doesn't help.

We are aware of situations where scopes may go missing (due to calls to `cleanUp()`): when using ASTParser and requesting type bindings that haven't been created during the initial invocation.  If that is the context of your problem, then you probably want https://bugs.eclipse.org/bugs/show_bug.cgi?id=150657 implemented :)

Other than that a `null` scope signals a bug elsewhere in the (AspectJ) compiler.

kriegaex:
I think, I know why the scope is null: The problem occurs in `BinaryTypeBinding` instances, which someone, long before around 2021 I started helping to maintain AspectJ, hackily changed into extending `SourceTypeBinding` instead of the original `ReferenceBinding`. This is all quite contrived, and I am not confident to rewrite all of that code. Anyway, in `BinaryTypeBinding`, the scope is never initialised, therefore it came as a surprise to our code that suddenly in `storeAnnotations` the scope is being used.

I think, I am going to close this PR and try to add something hacky like this to the forked code:

```java
	// AspectJ extension: Add null guard for BinaryTypeBinding which for AJ was hackily changed into extending
	// SourceTypeBinding instead of the original ReferenceBinding.
	if (this instanceof BinaryTypeBinding && this.scope == null)
		return;
	// End AspectJ extension
	this.scope.referenceCompilationUnit().compilationResult.annotations.add(annotations);
```

I think, that makes sense for AspectJ purposes, given the fact that a `BinaryTypeBinding` is not really a `SourceTypeBinding`.

laeubi:
@stephan-herrmann @kriegaex maybe there should be a `Objects.requireNonNull` check in the constructor then?

kriegaex:
@laeubi, that would not solve the problem and still generate an exception in a place where I am not expecting any.

laeubi:
But iw ould have made you (as the user of the API) maybe made you aware that a `scope` is mandatory.

stephan-herrmann:
> But iw ould have made you (as the user of the API) maybe made you aware that a `scope` is mandatory.

Regarding API I would probably agree, but this is not API :)

laeubi:
> > But iw ould have made you (as the user of the API) maybe made you aware that a `scope` is mandatory.
> 
> Regarding API I would probably agree, but this is not API :)

True, still it would help consumers / extenders of this non-api class then ;-)

_Disclaimer_: I did not know much about what aspectJ changes but
@kriegaex have you ever considered to add the required extensibility/information so AspectJ can use JDT out of the box?

stephan-herrmann:
> _Disclaimer_: I did not know much about what aspectJ changes but @kriegaex have you ever considered to add the required extensibility/information so AspectJ can use JDT out of the box?

Long time ago, when the original authors of AJDT where still active, an attempt was made for a round table of all stakeholders wishing to extend JDT in ways beyond existing API and extension points. We even created a wiki page for this initiative: https://wiki.eclipse.org/Extending_JDT_For_Java_Like_Languages

As you can see, not much came of it. I think the consensus was, that the changes required to extend JDT for another language are just several orders of magnitude too big and complex to ever come near any kind of black-box re-use. For this kind of project, forking and byte-code weaving are the only approaches that have yielded any viable results.

Let me repeat that with Object Teams I'm in the same boat. This actually paved the way for my participation in JDT - which in turn made my maintenance of Object Teams much smoother... :)

Issue #1106, opened 2023-05-30
[21] JEP 445: Unnamed Classes and Instance Main Methods (Preview)


mpalat:
Ref : https://openjdk.org/jeps/445
Summary
Evolve the Java language so that students can write their first programs without needing to understand language features designed for large programs. Far from using a separate dialect of Java, students can write streamlined declarations for single-class programs and then seamlessly expand their programs to use more advanced features as their skills grow. This is a [preview language feature](https://openjdk.org/jeps/12).

mpalat:
retargeting out of BETA_JAVA21

akurtakov:
After reading up and discussing the topic I think the work here could be split the following way:
* Unnamed Classes - requires extensive grammar change and is most likely the most expensive part in this JEP
* Main methods - itself could be split into:
* * static main() - @mpalat hinted me that whether the method has args or not is not part of the grammar (I hope I got that correct) thus support for this could be added with least effort. I plan to look into this one as an entry step.
* * non-static main() 

As there is a very tight schedule I hope that we can go the path of adding support gradually rather and thus deliver more incrementally so as much as possible makes it into December release.

mpalat:
> * Main methods -  @mpalat hinted me that whether the method has args or not is not part of the grammar (I hope I got that correct) thus support for this could be added with least effort. I plan to look into this one as an entry step.
@akurtakov - to make it clear - what I meant is there is no special grammar for "main" method and hence there is no "special" grammar to check whether "main" method being the main method has args or not in the grammar

mickaelistria:
We will probably need something like #1470 to be copied or moved upstream in org.eclipse.jdt.compiler.batch/org.eclipse.jdt.internal.compiler.ast.MethodDeclaration` because the JLS specifies that "it is a compile-time error if this [unnamed] class does not declare a candidate main method" so we need to be able to list candidate `main` in the compiler directly.

As for the unnamed class, I've been only focusing on grammar and parsing so far (and am making some progress), but I realized that it's only half of the job, the other part being the codegen. So I think what would be best is to first focus on how we'd like the AST to be built for that case, merge the provisional API so we can work on parallel on parsing and on codegen.
My first idea would be to create a new `UnnamedClass` ASTNode which would contain `MethodDeclaration`s, `FieldDeclaration`s and `TypeDeclaration`s make the CompilationUnitDeclaration have a field `Optional<UnnamedClass>`. Does it sound like a good starting point?

[...]

PR #1470, opened 2023-10-06
Recognize instance main methods as main methods.

akurtakov:
## What it does
Implements the JEP 445 definition of main methods on Java 21 with --enable-preview.
This is the jdt.core model part which makes "Run As" recognize the new main methods although it fails after that due to changes needed in jdt.debug or jdt.ui

## How to test
Run the newly created Java21ElementTests in your workspace with Java 21.
Note: this test case is not added in any suite as I didn't manage to figure Java 21 specific testing yet thus information where to add the test case would be much appreciated.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/Java21ElementTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/Java21ElementTests.java
@@ -1,0 +1,370 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Red Hat Inc. and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Red Hat Inc. - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.jdt.core.tests.model;
+
+import org.eclipse.jdt.core.ICompilationUnit;
+import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.jdt.core.JavaCore;
+import org.eclipse.jdt.core.tests.util.AbstractCompilerTest;
+
+import junit.framework.Test;
+
+public class Java21ElementTests extends AbstractJavaModelTests {
+
+	private IJavaProject project;
+	public Java21ElementTests(String name) {
+		super(name);
+	}
+
+	public static Test suite() {
+		return buildModelTestSuite(AbstractCompilerTest.F_21, Java21ElementTests.class);
+	}
+
+	@Override
+	protected void setUp() throws Exception {
+		super.setUp();
+		this.project = createJavaProject("Java21Elements", new String[] { "src" },
+				new String[] { "JCL21_LIB" }, "bin", "21");
+	}
+
+	@Override
+	protected void tearDown() throws Exception {
+		deleteProject("Java21Elements");
+		super.tearDown();
+	}
+
+	public void test001() throws Exception {
+		this.project.setOption(JavaCore.COMPILER_PB_ENABLE_PREVIEW_FEATURES, JavaCore.ENABLED);
+		this.project.open(null);
+		String fileContent = """
+				public interface Test {
+					public static void main(String[] args) {
+					}
+				}
+					""";
+		createFile("/Java21Elements/src/Test.java", fileContent);
+
+		ICompilationUnit unit = getCompilationUnit("/Java21Elements/src/Test.java");
+		assertTrue(unit.getTypes()[0].getMethods()[0].isMainMethodCandidate());
+	}
+
+	public void test002() throws Exception {
+		this.project.setOption(JavaCore.COMPILER_PB_ENABLE_PREVIEW_FEATURES, JavaCore.ENABLED);
+		this.project.open(null);
+		String fileContent = """
+				public class Test {
+					public static void main() {
+					}
+				}
+					""";
+		createFile("/Java21Elements/src/Test.java", fileContent);
+
+		ICompilationUnit unit = getCompilationUnit("/Java21Elements/src/Test.java");
+		assertTrue(unit.getTypes()[0].getMethods()[0].isMainMethodCandidate());
+	}
+
+	public void test003() throws Exception {
+		this.project.setOption(JavaCore.COMPILER_PB_ENABLE_PREVIEW_FEATURES, JavaCore.ENABLED);
+		this.project.open(null);
+		String fileContent = """
+				public class Test {
+					public void main() {
+					}
+				}
+					""";
+		createFile("/Java21Elements/src/Test.java", fileContent);
+
+		ICompilationUnit unit = getCompilationUnit("/Java21Elements/src/Test.java");
+		assertTrue(unit.getTypes()[0].getMethods()[0].isMainMethodCandidate());
+	}
+
+	public void test004() throws Exception {
+		this.project.setOption(JavaCore.COMPILER_PB_ENABLE_PREVIEW_FEATURES, JavaCore.ENABLED);
+		this.project.open(null);
+		String fileContent = """
+				public class Test {
+					protected void main() {
+					}
+				}
+					""";
+		createFile("/Java21Elements/src/Test.java", fileContent);
+
+		ICompilationUnit unit = getCompilationUnit("/Java21Elements/src/Test.java");
+		assertTrue(unit.getTypes()[0].getMethods()[0].isMainMethodCandidate());
+	}
+
+	public void test005() throws Exception {
+		this.project.setOption(JavaCore.COMPILER_PB_ENABLE_PREVIEW_FEATURES, JavaCore.ENABLED);
+		this.project.open(null);
+		String fileContent = """
+				public class Test {
+					void main() {
+					}
+				}
+					""";
+		createFile("/Java21Elements/src/Test.java", fileContent);
+
+		ICompilationUnit unit = getCompilationUnit("/Java21Elements/src/Test.java");
+		assertTrue(unit.getTypes()[0].getMethods()[0].isMainMethodCandidate());
+	}
+
+	public void test006() throws Exception {
+		this.project.setOption(JavaCore.COMPILER_PB_ENABLE_PREVIEW_FEATURES, JavaCore.ENABLED);
+		this.project.open(null);
+		String fileContent = """
+				public class Test {
+					private void main() {
+					}
+				}
+					""";
+		createFile("/Java21Elements/src/Test.java", fileContent);
+
+		ICompilationUnit unit = getCompilationUnit("/Java21Elements/src/Test.java");
+		assertFalse(unit.getTypes()[0].getMethods()[0].isMainMethodCandidate());
+	}
+
+	public void test007() throws Exception {
+		this.project.setOption(JavaCore.COMPILER_PB_ENABLE_PREVIEW_FEATURES, JavaCore.ENABLED);
+		this.project.open(null);
+		String fileContent = """
+				public class Test {
+					protected void main(String[] args) {
+					}
+				}
+					""";
+		createFile("/Java21Elements/src/Test.java", fileContent);
+
+		ICompilationUnit unit = getCompilationUnit("/Java21Elements/src/Test.java");
+		assertTrue(unit.getTypes()[0].getMethods()[0].isMainMethodCandidate());
+	}
+
+	public void test008() throws Exception {
+		this.project.setOption(JavaCore.COMPILER_PB_ENABLE_PREVIEW_FEATURES, JavaCore.ENABLED);
+		this.project.open(null);
+		String fileContent = """
+				public class Test {
+					void main(String[] args) {
+					}
+				}
+					""";
+		createFile("/Java21Elements/src/Test.java", fileContent);
+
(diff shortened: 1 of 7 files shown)


akurtakov:
I would appreciate any feedback on this one as I might not be doing things in the most idiomatic JDT way.

akurtakov:
This is first part of implementing https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1106

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/Java21ElementTests.java
```diff
@@ -0,0 +1,204 @@
...
+package org.eclipse.jdt.core.tests.model;
+
+import org.eclipse.jdt.core.ICompilationUnit;
+import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.jdt.core.JavaCore;
+import org.eclipse.jdt.core.tests.util.AbstractCompilerTest;
+
+import junit.framework.Test;
+
+public class Java21ElementTests extends AbstractJavaModelTests {
```
iloveeclipse, line 23: The tests need to be added to the testsuite. Note, there are different levels at which tests run (depending on the JLS level they suppose to test on)

akurtakov: Would you please guide me in identifying where exactly to add these?

iloveeclipse: AllJavaModelTests

akurtakov:
@iloveeclipse So what's the story with these build failures?
```
14:17:04  3 API ERRORS
14:17:04  [API ERROR] File MANIFEST.MF at line 5: The major version should be incremented in version 3.36.0, since API breakage occurred since version 3.35.0 - The super interfaces set has been reduced (java.lang.constant.Constable) for type org.eclipse.jdt.core.util.ExternalAnnotationUtil.MergeStrategy
14:17:04  - The super interfaces set has been reduced (java.lang.constant.Constable) for type org.eclipse.jdt.core.dom.rewrite.ImportRewrite.TypeLocation (location: /home/jenkins/agent/workspace/eclipse.jdt.core-Github_PR-1470/org.eclipse.jdt.core/META-INF/MANIFEST.MF)
14:17:04  [API ERROR] File ImportRewrite.java at line 116: The super interfaces set has been reduced (java.lang.constant.Constable) for type org.eclipse.jdt.core.dom.rewrite.ImportRewrite.TypeLocation (location: /home/jenkins/agent/workspace/eclipse.jdt.core-Github_PR-1470/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java)
14:17:04  [API ERROR] File ExternalAnnotationUtil.java at line 68: The super interfaces set has been reduced (java.lang.constant.Constable) for type org.eclipse.jdt.core.util.ExternalAnnotationUtil.MergeStrategy (location: /home/jenkins/agent/workspace/eclipse.jdt.core-Github_PR-1470/org.eclipse.jdt.core/model/org/eclipse/jdt/core/util/ExternalAnnotationUtil.java)
```

iloveeclipse:
> So what's the story with these build failures?

https://github.com/eclipse-pde/eclipse.pde/issues/771

They come and go, depending on the moon position. If one retriggers build often enough, they go away... maybe.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/Java21ElementTests.java
akurtakov: Added in the last revision.

akurtakov:
So I think the comments are addressed, tests are running and verification build succeeds. Is there anything else I have to do?

iloveeclipse:
> So I think the comments are addressed, tests are running and verification build succeeds. Is there anything else I have to do?

No, thanks. @jarthana : could you please review?

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/Member.java
```diff
@@ -394,14 +395,31 @@ public boolean isBinary() {
...
+		int flags = method.getFlags();
+		ICompilationUnit icu = method.getCompilationUnit();
+		if (icu != null) {
+			Map<String, String> options = icu.getOptions(true);
+			if (JavaCore.ENABLED.equals(options.get(JavaCore.COMPILER_PB_ENABLE_PREVIEW_FEATURES))
+					&& JavaCore.VERSION_21.equals(options.get(JavaCore.COMPILER_CODEGEN_TARGET_PLATFORM))) {
+				if (!Flags.isPrivate(flags)) {
+					String[] paramTypes = method.getParameterTypes();
+					if (paramTypes.length == 1) {
+						String typeSignature = Signature.toString(paramTypes[0]);
```
gayanper, line 407: Could we extract this duplicate statements to a seriate method which checks weather the parameter is a String array ?

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/Member.java
```diff
@@ -394,14 +395,31 @@ public boolean isBinary() {
 }
 protected boolean isMainMethod(IMethod method) throws JavaModelException {
 	if ("main".equals(method.getElementName()) && Signature.SIG_VOID.equals(method.getReturnType())) { //$NON-NLS-1$
-		int flags= method.getFlags();
+		int flags = method.getFlags();
+		ICompilationUnit icu = method.getCompilationUnit();
```
gayanper, line 399: Was there any specific reason for using compilation unit over method's JavaProject for checking the options ?

akurtakov: This is the way I found first. I will look into doing it the other way if it's the prefered way, is it?

jarthana:
Sorry, was on leave for the past few days. I am back and looking at it now.

akurtakov:
> Sorry, was on leave for the past few days. I am back and looking at it now. I have some points/questions/thoughts:
> 
>     1. We generally add an `@noreference` for anything that is preview related, so that clients don't run into trouble if we later remove them. This is a peculiar, because this is an existing method, but the behavior gets changed, which may change again. But good news is, `IMethod `already gets an `@noimplement`. So, we may actually get away with it. But just not sure whether the method should still get the `@noreference`.

If IMethod itself has @noimplement I don't see a need  to add it for the method specifically but will do if you request it.

> 
>     2. In the following case, `isMainMethod()` returns true for both methods.
> 
> 
> ```
> public static void main(String ... args) {}
> public static void main() {}
> ```
> 
> However, Java recognizes (correctly) only the first as 'the' main method. Clients using this API might get confused with this behavior. 

As per https://cr.openjdk.org/~gbierman/jep445/jep445-20230524/specs/unnamed-classes-instance-main-methods-jls.html#jls-12.1.1 there are many "invoking ~~the~~ a method main " which changes the spec in a way that it no longer claims there is only one main method. Thus it is correct for the isMainMethod() to return true for both. 
It is another question whethere a new API like **isTheMainMethod()** is needed which would determine whether this would be the called method based on https://cr.openjdk.org/~gbierman/jep445/jep445-20230524/specs/unnamed-classes-instance-main-methods-jls.html#jls-12.1.4 . IMHO it's too early to add such API until this becomes standard feature.

>3. It will be better and future-proof if we use `JavaFeature.isSupported()` instead of checking for compliance and preview flag. We need to add a new `JavaFeature`, first, though. 

I can work on that if you explain in more details (preferably with link to this done before) what is needed.

>4. Can we have a testcase that checks the qualified name (`java.lang.String[]`) so that the `getSimpleName()` get tested?

Done in latest version of the patch.

jarthana:
> If IMethod itself has @noimplement I don't see a need to add it for the method specifically but will do if you request it.
> 
I was in two minds. But I guess we can live with it.
> 
> As per https://cr.openjdk.org/~gbierman/jep445/jep445-20230524/specs/unnamed-classes-instance-main-methods-jls.html#jls-12.1.1 there are many "invoking ~the~ a method main " which changes the spec in a way that it no longer claims there is only one main method. Thus it is correct for the isMainMethod() to return true for both. It is another question whethere a new API like **isTheMainMethod()** is needed which would determine whether this would be the called method based on https://cr.openjdk.org/~gbierman/jep445/jep445-20230524/specs/unnamed-classes-instance-main-methods-jls.html#jls-12.1.4 . IMHO it's too early to add such API until this becomes standard feature.
> 

There will be more cases where it will be impossible not to add new APIS. That's why we have started tagging the preview APIs with `@noreference` and such.
How about adding a new method, something like, `isAMainMethod()`? That way, we make it explicit that the clients, when using the new method, may have an additional responsibility to find out 'the' main method. Added bonus is, we leave the old one untouched/unbroken.

> > 3. It will be better and future-proof if we use `JavaFeature.isSupported()` instead of checking for compliance and preview flag. We need to add a new `JavaFeature`, first, though.
> 
> I can work on that if you explain in more details (preferably with link to this done before) what is needed.

A new enum constant for this feature needs to be added in `JavaFeature` and the clients just use that like below:
`JavaFeature.UNNAMED_CLASSES.isSupported(options)`

Just look for references of `isSupported()`; there are several examples.

jarthana:
> This would mean JDT Debug, JDT LS and etc. will have to replace almost all usages of isMainMethod with isAMainMethod when quering whether a class contains main method. It's extra churn over adding isTheMainMethod, which should be needed if one wants to determine which main method would be called and this choice is done transparently by JVM itself. If you still think isAMainMethod is the way to do I'll do the isAMainMethod.

Yes, I understand it's not convenient. But the spec changes, the behavior changes and the clients better look at each use-case and take appropriate action. In fact, the more I think, we should not have thought about changing the existing API at all, because the API the clients relied on for getting THE main method is no longer present. We must respect the contract more than anything else.

org.eclipse.jdt.core/model/org/eclipse/jdt/core/IMethod.java
```diff
@@ -255,6 +255,14 @@ public interface IMethod extends IMember, IAnnotatable {
  * <li>it is <code>static</code> and <code>public</code></li>
  * <li>it defines one parameter whose type's simple name is <code>String[]</code></li>
  * </ul>
+ * Starting with Java 21 there is "preview feature" that allows <a href="https://openjdk.org/jeps/445">instance main methods</a>.
```
jarthana, line 258: We should keep this API/contract untouched and create a new API with @noreference, so that existing clients are not broken.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/Member.java
```diff
@@ -394,20 +395,38 @@ public boolean isBinary() {
 }
 protected boolean isMainMethod(IMethod method) throws JavaModelException {
 	if ("main".equals(method.getElementName()) && Signature.SIG_VOID.equals(method.getReturnType())) { //$NON-NLS-1$
-		int flags= method.getFlags();
+		int flags = method.getFlags();
+		Map<String, String> options = method.getJavaProject().getOptions(true);
+		if (JavaCore.ENABLED.equals(options.get(JavaCore.COMPILER_PB_ENABLE_PREVIEW_FEATURES))
```
jarthana, line 400: This is not ideal and will need to be touched again when the preview status or version change. Add a new feature to JavaFeature and call its isSupported(options).

jarthana (changes requested):

mickaelistria:
So `isMainMethod` contract should be specified to return true for *the* main method (ie which include a resolution of which one is main) ? I think it would make sense, since after all, other methods named `main` are not real main when there is a higher priority `main` existing; they're just regular/non-main methods, which happen to be called `main`.

akurtakov:
> So `isMainMethod` contract should be specified to return true for _the_ main method (ie which include a resolution of which one is main) ? I think it would make sense, since after all, other methods named `main` are not real main when there is a higher priority `main` existing; they're just regular/non-main methods, which happen to be called `main`.

Being called main is not enough - it has to be non-private and it should either has zero args or only String[]. So let's call there is the "main" and a "main" candidate.

org.eclipse.jdt.core/model/org/eclipse/jdt/core/IMethod.java
```diff
@@ -263,6 +263,31 @@ public interface IMethod extends IMember, IAnnotatable {
...
+ * <li>it is non-<code>private</code></li>
+ * <li>it defines one parameter whose type's simple name is <code>String[]</code> or no parameter at all</li>
+ * </ul>
+ *
+ * @exception JavaModelException if this element does not exist or if an
+ *      exception occurs while accessing its corresponding resource.
+ * @since 3.36
+ * @return true if this method is a main method, false otherwise
+ */
+boolean isAMainMethod() throws JavaModelException;
```
mickaelistria, line 289: What about `isMainCandidate` ?

jarthana: > isMainMethodCandidate might even be clearer. WDYT, @jarthana ?

Yes, this makes it unambiguous.

akurtakov: Done with latest revision.

akurtakov:
@jarthana If I submit enough times I'll get a clean build but I don't think this is good use of my time.

jarthana (approved):

akurtakov:
Thanks for the review, Jay!

PR #1486, opened 2023-10-13
Jenkins file changes to debug random API analysis errors

iloveeclipse:
- Enable debug for API tooling on Jenkins
Should help to understand strange API analysis errors.
- Disable DetectVMInstallationsJob in Jenkins
That should reduce possible side effects on API analysis.

See https://github.com/eclipse-pde/eclipse.pde/issues/782


iloveeclipse:
See also https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1161 that was done for **tests**. This PR is for API analysis application that also could be affected.

Issue #814, opened 2023-03-08
Organize Imports produces impossible imports


laeubi:
Assume the following code: [test.access.staticprotected.zip](https://github.com/eclipse-jdt/eclipse.jdt.core/files/10888888/test.access.staticprotected.zip)

Now I use "Organize Imports" on the class `TestMe` what results in a compile error:
![grafik](https://user-images.githubusercontent.com/1331477/223667125-1a3b7342-37da-444e-9a1a-c7e5dd62c8d1.png)

The correct import would be `import static test.access.staticprotected.ThisIsPublic.thisIsPublic`

![grafik](https://user-images.githubusercontent.com/1331477/223667424-6599c4b3-3ff2-4e7e-99ef-7ac83c8a94e8.png)

erik-brangs:
Is this the same as https://bugs.eclipse.org/bugs/show_bug.cgi?id=565766 ? I'm asking because I investigated whether I could fix that and had to give up. I still have some notes about that. I'm not sure if they're useful to anyone else. IIRC I also prepared a test case.

laeubi:
@erik-brangs yes this seems the same problem!

erik-brangs:
[testcase-for-organize-imports-jdt-ui.zip](https://github.com/eclipse-jdt/eclipse.jdt.core/files/10920063/testcase-for-organize-imports-jdt-ui.zip)

My notes are from last year, so the situation may have changed in the mean time. I may also be wrong. Nevertheless, here's an extract of my notes in case they're helpful to anyone:
- static imports are added in ``ImportRewriteContext`` in ``addStaticImport`` which is in jdt.core
- the ``declaringType`` is simply used for both methods and fields without checking visibility
- I didn't find any convenient way to check for visibility or scope. There's a ``ScopyAnalyzer`` but it's in ``org.eclipse.jdt.core.manipulation`` which is in jdt.ui and AFAIK isn't accessible from jdt.core

laeubi:
@erik-brangs AFAIK there are some attempts to merge generic parts from "manipulation" to "jdt-core" so if you think you can take a second look at this ... probabbly one needs the original `ScopyAnalyzer` extnds the moved class for backward compatibility, but maybe @iloveeclipse or @stephan-herrmann can give some advice here.

stephan-herrmann:
In compiler-land, there's a method `ReferenceBinding.canBeSeenBy(PackageBinding)`, which would solve the problem.

[...]

PR #1498, opened 2023-10-19
Import static with protected base class yiels wrong class reference compiled in resulting classfile

### LINKED ISSUE #809: Import static with protected base class yiels wrong class reference compiled in resulting classfile
I have a quite uncommon class structure (but I can't change it as the code is generated), where I have a **package protected class** that has a **public static method**, then there is another **public class** that extends the first one:

grafik

Now I have a class in another package that uses this public class and calls the static method and has a static import with wildcard:

grafik

in the IDE this compiles fine, no warnings or errors:

grafik

But if I now run the class it fails:

> Exception in thread "main" java.lang.IllegalAccessError: failed to access class test.access.staticprotected.ThisIsProtected from class test.another.pkg.TestMe (test.access.staticprotected.ThisIsProtected and test.another.pkg.TestMe are in module test.access.staticprotected of loader 'app')

The important part here is that **test.access.staticprotected.ThisIsProtected** is used, what is forbidden!

Interesting enough, this can be also seen in the IDE if i organize the imports what results in:
grafik

If I now change the code to use the public class like this
grafik

Then the IDE is fine **but** running it results in the same runtime error.

Next I change it to use full qualified name like this
grafik

Now it works in the IDE **and** if I run the program.

The sourceode for this example can be found here: [test.access.staticprotected.zip](https://github.com/eclipse-jdt/eclipse.jdt.core/files/10888888/test.access.staticprotected.zip)

@stephan-herrmann I think we can need your compiler expertise here, I can't even imagine if this is a bug or probably expected what one is seeing here, but whatever this is there seems a discrepancy between what the IDE thinks about this code and what the runtime is assuming.

stephan-herrmann:
fixes #809

## What it does
Remember the receiver type through which in import resolved a field or method, and if necessary use that "source-level" declaring class during code generation.

## How to test
JUnit test cases included.
@jukzi  can you think of any other failing situations not covered by my junits?

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
@@ -2246,7 +2246,12 @@ public abstract class Scope {
 										if (importReference != null && needResolve) {
 											importReference.bits |= ASTNode.Used;
 										}
-										invocationSite.setActualReceiverType(foundField.declaringClass);
+										TypeBinding importedType = this.getType(importBinding.compoundName, importBinding.compoundName.length-1);
+										if (importedType instanceof ReferenceBinding && importBinding.isValidBinding()) {
+											invocationSite.setActualReceiverType((ReferenceBinding) importedType);
+										} else {
+											invocationSite.setActualReceiverType(foundField.declaringClass);
+										}
 										if (foundField.isValidBinding()) {
 											return foundField;
 										}
@@ -2258,12 +2263,14 @@ public abstract class Scope {
 						}
 						// check on demand imports
 						boolean foundInImport = false;
+						ReferenceBinding sourceCodeReceiver = null;
 						for (int i = 0, length = imports.length; i < length; i++) {
 							ImportBinding importBinding = imports[i];
 							if (importBinding.isStatic() && importBinding.onDemand) {
 								Binding resolvedImport = importBinding.resolvedImport;
 								if (resolvedImport instanceof ReferenceBinding) {
-									FieldBinding temp = findField((ReferenceBinding) resolvedImport, name, invocationSite, needResolve);
+									ReferenceBinding importedReferenceBinding = (ReferenceBinding) resolvedImport;
+									FieldBinding temp = findField(importedReferenceBinding, name, invocationSite, needResolve);
 									if (temp != null) {
 										if (!temp.isValidBinding()) {
 											if (problemField == null)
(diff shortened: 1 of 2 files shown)


jukzi:
I am not familiar with this topic. Maybe the bug reports can contribute more ideas?  @laeubi @HannesWell

stephan-herrmann:
> I am not familiar with this topic. Maybe the bug reports can contribute more ideas? @laeubi @HannesWell

Ups, of course! Sorry for the mix-up! 

(_Perhaps creating a PR while aboard a train is prone to lots of distractions_ :smile:  )

laeubi:
@stephan-herrmann if you like some more complex examples you can take a look here:

https://github.com/openjdk/jextract/tree/master/samples

especially this one was the failing thing I derived the reproducer from:
https://github.com/openjdk/jextract/blob/master/samples/opengl/Teapot.java

stephan-herrmann:
> @stephan-herrmann if you like some more complex examples you can take a look here:
> 
> https://github.com/openjdk/jextract/tree/master/samples
> 
> especially this one was the failing thing I derived the reproducer from: https://github.com/openjdk/jextract/blob/master/samples/opengl/Teapot.java https://github.com/openjdk/jextract/blob/master/samples/opengl/README.md

Thanks. From a quick glance the Teapot example uses exactly one pattern: star-import plus simple name reference to a field. This should be covered OK. Am I missing other patterns? Additionally, tests already include importing the class and using `MyClass.FIELD` syntax, explicit import of fields, plus all the same for methods.

laeubi:
> From a quick glance the Teapot example uses exactly one pattern: star-import plus simple name reference to a field. This should be covered OK. Am I missing other patterns?

I usually have enabled "Organize imports" as a save action, so for me on save this converts to the ones actually used (what currently fails due to https://github.com/eclipse-jdt/eclipse.jdt.core/issues/814), so having "non stared" static imports is important, and another common pattern would be of having something like `glut_h.glClearColor(...)` so from your description and as far as I understand your test-cases this should be covered as well.

laeubi:
@stephan-herrmann test seem al green so can't wait to see this in the next release :-)

Does this also fixes
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/814

as it seem to suffer from a similar problem?

stephan-herrmann:
> Does this also fixes
>     * [Organize Imports produces impossible imports #814](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/814)
> 
> as it seem to suffer from a similar problem?

No, this current PR fixes the compiler, whereas #814 is a problem in or near org.eclipse.jdt.core.dom.rewrite.ImportRewrite.

PR #1513, opened 2023-10-26
Jep 430 String template - WIP patch

### LINKED ISSUE #544: [21] JEP 430 String Templates (First Preview)
JEP 430 at https://openjdk.org/jeps/430
Ref https://cr.openjdk.java.net/~gbierman/jep430/latest for more details on LS changes

At the time of creating this issue, this is not yet targeted but it *may* be targeted - keeping this in the radar for 20

jarthana:
Issue #544 

## What it does
Fixes #544 

## Author checklist


jarthana:
The failures in DietRecoveryTest are recovery related and I think we can simply adjust the tests here. However, the ones in NullAnnotationTest need attention. They arise because we are triggering initialization of null analysis and thus resolution of annotations much earlier. I..e at the time of computing default imports itself. Either we need to 
1. ensure the null analysis is not prematurely triggered or 
2. instead of adding an import, replace "STR" or "RAW" with their qualified names. (I haven't tried this though)

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
```diff
@@ -2032,99 +1862,129 @@ protected int processSingleQuotes(boolean checkIfUnicode) throws InvalidInputExc
...
-						this.withoutUnicodePtr--;
-						// consume next character
-						this.unicodeAsBackSlash = false;
-						if (((this.currentCharacter = this.source[this.currentPosition++]) == '\\') && (this.source[this.currentPosition] == 'u')) {
-							getNextUnicodeChar();
-							isUnicode = true;
-							this.withoutUnicodePtr--;
+
+						if (templatesAllowed && this.currentCharacter == '{' && openBraces++ == 0) {
+							pushFragmentPosition(this.withoutUnicodePtr, this.currentPosition - (isUnicode ? 7 : 2));
```
srikanth-sankaran, line 1974: I think using jumpOverMethodBody can significantly simplify the implementation. Let us discuss

srikanth-sankaran (commented):
First batch of comments.

srikanth-sankaran:
A major recommendation is to:

1. Come up with a method `jumpOverEmbeddedExpression` modelled after `jumpOverMethodBody` (best not to completely duplicate this method, but define a customizable method `jumpToMatchingBrace` that can be called by both
2. This method should continue to maintain withoutUnicodeBuffer and withoutUnicodePtr as needed.
3. Maintain co-ordinates of fragments and expression relative to string start rather than relative to source, this way if we find unicodes and switch to alternate buffering, the already recorded offsets will still be valid.
4. The recorded offsets need to be string relative if `withoutUnicodePtr` == 0 and will simply be` withoutUnicodePtr` itself if the string has any unicodes

jarthana:
> A major recommendation is to:
> 
>     1. Come up with a method `jumpOverEmbeddedExpression` modelled after `jumpOverMethodBody` (best not to completely duplicate this method, but define a customizable method `jumpToMatchingBrace` that can be called by both

I played around a bit. While it works well for most cases, it is confused with cases like this:
`String s = STR."\{ STR."\{ STR."}abc{" }" }";`

To make it work, we should make it aware of the nested \{} too, that might make it more complicated than it is now.

srikanth-sankaran:
> > A major recommendation is to:
> > ```
> > 1. Come up with a method `jumpOverEmbeddedExpression` modelled after `jumpOverMethodBody` (best not to completely duplicate this method, but define a customizable method `jumpToMatchingBrace` that can be called by both
> > ```
> 
> I played around a bit. While it works well for most cases, it is confused with cases like this: `String s = STR."\{ STR."\{ STR."}abc{" }" }";`
> 
> To make it work, we should make it aware of the nested {} too, that might make it more complicated than it is now.

Can you pick up the changes from https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1571 - this passes all tests and in this change `jumpOverMethodBody` calls `scanForStringLiteral` which should automatically handle the case you cite

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
```diff
@@ -114,6 +115,16 @@ public class Scanner implements TerminalTokens {
...
+	// Positions in the source[] where \{ and } appear
+	public int[] embeddedExpressionPositions = new int[20];
+	// Contains a set of 4 positions
+	// 0: start pos of String fragment in this.withoutUnicodeBuffer, i.e. before /{
+	// 1: start pos of String fragment in this.source[], i.e. before /{
+	// 2: end pos of String fragment in this.withoutUnicodeBuffer i.e. after }
+	// 3: end pos of String fragment in this.source[], i.e. after }
+	public int[] fragmentPositions = new int[20];
+	public int embeddedExpressionPtr = 0;
+	public int fragmentPtr = 0;
```
srikanth-sankaran, line 128: It ought to be sufficient to just maintain embeddedExpression start offset and end offset. If these offsets are relative to the string. At any point, we need to maintain these offsets only for the "current" "top level" string template. When we call parse expression on an embedded expression if that contains an embedded string template, that will become the "current" "top level" as we descend into it. Thus we will peel it layer by layer

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
```diff
@@ -2032,99 +1862,128 @@ protected int processSingleQuotes(boolean checkIfUnicode) throws InvalidInputExc
...
+						Scanner newScanner = cloneScanner(this.currentPosition, this.eofPosition);
+						newScanner.scanForStringLiteral();
+						// transfer the updated values that we need to the original scanner
+						this.currentPosition = newScanner.currentPosition;
+						this.linePtr = newScanner.linePtr;
+					}
+					if (!templatesAllowed || openBraces == 0) {
+						break;
+					}
 				}
```
srikanth-sankaran, line 1897: This clone approach should not be needed if we jump past embedded expressions.

jarthana:
I noticed we report an explicit import of STR as unused import. This import should be marked as used or ignored.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
```diff
@@ -517,6 +518,30 @@ public final String getCurrentTokenString() {
 		this.startPosition,
 		this.currentPosition - this.startPosition);
 }
+public char[] getCurrentTokenInRange(int start, int end) {
+	char[] result;
+	if (this.withoutUnicodePtr != 0) {
+		if (this.textBlockOffset > 0) {
+			System.arraycopy(this.withoutUnicodeBuffer, this.textBlockOffset + 1,
+			//2 is 1 (real start) + 1 (to jump over the ")
+			result = new char[this.withoutUnicodePtr - this.textBlockOffset], 0, this.withoutUnicodePtr - this.textBlockOffset);
```
srikanth-sankaran, line 527: Line 526 comment needs to be deleted as it is stale

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
```diff
@@ -575,230 +600,23 @@ protected final boolean scanForTextBlockClose() throws InvalidInputException {
 	return false;
 }
 public char[] getCurrentTextBlock() {
-	// 1. Normalize, i.e. convert all CR CRLF to LF
+	char[][] lines = getCurrentTextBlockAsLines();
+	int indent = TextBlock.getTextBlockIndent(lines);
+	//	get rid of all the cached values
+	this.textBlockOffset = -1;
+	return TextBlock.formatTextBlock(lines, indent);
```
srikanth-sankaran, line 607: Clearing of cached values is better handled elsewhere. After getNextToken() returns a template, it should be possible to call getCurrentTextBlock() multiple times for whatever reason

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
```diff
@@ -2030,16 +1848,48 @@ protected int processSingleQuotes(boolean checkIfUnicode) throws InvalidInputExc
...
+public record EmbeddedExpression (int start, int end) implements IStringTemplateComponent {
+
+}
+List<IStringTemplateComponent> stringTemplateComponents = null;
+void addStringTemplateComponent(IStringTemplateComponent stc) {
+	if (this.stringTemplateComponents == null) {
+		this.stringTemplateComponents = new ArrayList<>();
+	}
+	// collapse these - introduced for inspecting values while debugging
+	this.stringTemplateComponents.add(stc);
```
srikanth-sankaran, line 1868: Comment needs to be deleted

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
```diff
@@ -2030,16 +1848,48 @@ protected int processSingleQuotes(boolean checkIfUnicode) throws InvalidInputExc
...
+}
+List<IStringTemplateComponent> getCurrentStringTemplateComponents() {
+	return this.stringTemplateComponents;
+}
```
srikanth-sankaran, line 1872: These API methods names can be shortened by just saying Template

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
```diff
@@ -2146,20 +2000,39 @@ protected int scanForStringLiteral() throws InvalidInputException {
...
 				if (scanForTextBlockClose()) {
-					this.currentPosition += 2;
+					if (this.stringTemplateComponents != null) {
+						// mark the ending of the last fragment in the string template (before the closing quote)
+			            if (this.jumpingOverEmbeddedExpression == 0) {
+						    addStringTemplateComponent(new TextFragment(textFragmentStart, this.currentPosition + 1,
+						    		getCurrentTokenInRange(textFragmentStart - this.startPosition, this.currentPosition - this.startPosition - 2)));
+			            }
+			            this.currentPosition += 2; // move it past the triple quote
+						return TokenNameTextBlockTemplate;
```
srikanth-sankaran, line 2033: Doubtful if this will work for unicode usage for """

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
```diff
@@ -2243,13 +2126,33 @@ protected int scanForTextBlock() throws InvalidInputException {
...
 	}
 }
+
+private int getAndStoreStringFragment(int lastFragmentStart, int textFragmentEnd) {
+	if (this.jumpingOverEmbeddedExpression == 0) {
+		addStringTemplateComponent(new TextFragment(lastFragmentStart, textFragmentEnd - 1, getCurrentTokenInRange(lastFragmentStart - this.startPosition, textFragmentEnd - this.startPosition)));
+		int eeStart = this.currentPosition;
+		jumpOverEmbeddedExpression();
+		// verify that we are at } or handle error
+		int eeEnd;
```
srikanth-sankaran, line 2143: Do we handle the case of missing } properly ?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
```diff
@@ -4571,6 +4574,40 @@ protected void consumeInsideCastExpressionLL1WithBounds() {
 protected void consumeInsideCastExpressionWithQualifiedGenerics() {
 	// InsideCastExpressionWithQualifiedGenerics ::= $empty
 }
+protected void consumeTemplateExpressionWithPrimary() {
+	Expression expression = this.expressionStack[this.expressionPtr--];
+	StringTemplate template;
+	if (expression instanceof StringLiteral) {
+		template = new StringTemplate(new StringLiteral[] {(StringLiteral) expression},
+									new Expression[0], expression.sourceStart, expression.sourceEnd);
```
srikanth-sankaran, line 4582: Is there a test that verifies we do the right thing for a text block without any embedded expressions.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
```diff
@@ -4571,6 +4574,40 @@ protected void consumeInsideCastExpressionLL1WithBounds() {
...
+	} else {
+		template = (StringTemplate) expression;
+	}
+	TemplateExpression exp = new TemplateExpression(
+								ref,
+								template);
+	exp.sourceStart = ref.sourceStart;
+	exp.sourceEnd = expression.sourceEnd;
+	pushOnExpressionStack(exp);
```
srikanth-sankaran, line 4609: Lots of duplication with previous method. Given NameReference is an Expression, perhaps extracting common code may be a possibility

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
```diff
@@ -13748,6 +13832,9 @@ public char[][] parsePackageDeclaration(char[] source, CompilationResult result)
 
 }
 public Expression parseExpression(char[] source, int offset, int length, CompilationUnitDeclaration unit, boolean recordLineSeparators) {
+	return parseExpression(source, offset, length, unit, recordLineSeparators, false);
+}
+private Expression parseExpression(char[] source, int offset, int length, CompilationUnitDeclaration unit, boolean recordLineSeparators, boolean stopAtError) {
```
srikanth-sankaran, line 13838: stopAtError is better named embeddedExpression

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
```diff
@@ -13759,6 +13846,19 @@ public Expression parseExpression(char[] source, int offset, int length, Compila
 	this.scanner.setSource(source);
 	this.scanner.resetTo(offset, offset + length - 1);
 	try {
+		if (stopAtError) {
+			int nextToken;
+			try {
+				nextToken = this.scanner.getNextToken();
+			} catch(InvalidInputException iie) {
+				return null;
+			}
```
srikanth-sankaran, line 13855: May be we can return NullLiteral

srikanth-sankaran:
Latest commit seems to have added GenerateParserScript.class - is this supposed to be a part of the commit ??

srikanth-sankaran (commented):
Next batch of comments

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ClassFile.java
```diff
@@ -4073,6 +4078,40 @@ private int addBootStrapStringConcatEntry(int localContentsOffset, String recipe
...
 	}
+	private int addBootStrapTemplateRuntimeEntry(int localContentsOffset, StringTemplate template, Map<String, Integer> fPtr) {
+		final int contentsEntries = 10;
+		int indexForProcess = fPtr.get(ClassFile.NEW_STRING_TEMPLATE);
+		if (contentsEntries + localContentsOffset >= this.contents.length) {
+			resizeContents(contentsEntries);
+		}
+		if (indexForProcess == 0) {
+			ReferenceBinding javaLangRuntimeSwitchBootstraps = this.referenceBinding.scope.getJavaLangRuntimeTemplateRuntimeBootstraps();
+			indexForProcess = this.constantPool.literalIndexForMethodHandle(ClassFileConstants.MethodHandleRefKindInvokeStatic, javaLangRuntimeSwitchBootstraps,
```
srikanth-sankaran, line 4089: copy pasted name switch*

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
```diff
@@ -10001,33 +10046,68 @@ protected void consumeStaticOnly() {
...
+	List<IStringTemplateComponent> components = this.scanner.getCurrentStringTemplateComponents();
+	int size = components.size();
+	StringLiteral[] fragments = new StringLiteral[size/2 + 1];
+	char[][] textBlockFraments = new char[size/2 + 1][];
+	Expression[] expressions = new Expression[size/2];
+	for (int i = 0, fi = 0, ei = 0; i < size; i++) {
+		IStringTemplateComponent component = components.get(i);
+		if (component instanceof TextFragment tf) {
+			StringLiteral s = new StringLiteral(tf.text(), tf.start(), tf.end(), 0 /* fixme */);
+			fragments[fi] = s;
```
srikanth-sankaran, line 10059: This fixme needs to be handled

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/StringTemplate.java
```diff
@@ -0,0 +1,122 @@
...
+		return this.fragments;
+    }
+    public Expression[] values() {
+		return this.values;
+    }
+	@Override
+	public void resolve(BlockScope scope) {
+		if (this.constant != Constant.NotAConstant) {
+			this.constant = Constant.NotAConstant;
+		}
```
srikanth-sankaran, line 46: This looks suspicious -

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/codegen/ConstantPool.java
```diff
@@ -326,6 +326,10 @@ public class ConstantPool implements ClassFileConstants, TypeIds {
 			"(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/Class;Ljava/lang/invoke/MethodHandle;[Ljava/lang/Object;)Ljava/lang/Object;".toCharArray(); //$NON-NLS-1$
 	public static final char[] JAVA_LANG_ENUMDESC_OF_SIGNATURE = "(Ljava/lang/constant/ClassDesc;Ljava/lang/String;)Ljava/lang/Enum$EnumDesc;".toCharArray(); //$NON-NLS-1$
 	public static final char[] JAVA_LANG_CLASSDESC_OF_SIGNATURE = "(Ljava/lang/String;)Ljava/lang/constant/ClassDesc;".toCharArray(); //$NON-NLS-1$
+	public static final char[] JAVA_LANG_RUNTIME_STRING_TEMPLATE_SIGNATURE =
+			"(Ljava/lang/invoke/MethodHandles$Lookup;Ljava/lang/String;Ljava/lang/invoke/MethodType;[Ljava/lang/String;)Ljava/lang/invoke/CallSite;".toCharArray(); //$NON-NLS-1$
+
+	public static final char[] PROCESS = "process".toCharArray(); //$NON-NLS-1$
```
srikanth-sankaran, line 332: On both sides (in ClassFile and here) there is a method/ctor call to obtain the other form

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
```diff
@@ -2980,6 +2980,11 @@ public final ReferenceBinding getJavaLangInvokeStringConcatFactory() {
 		CompilationUnitScope unitScope = compilationUnitScope();
 		unitScope.recordQualifiedReference(TypeConstants.JAVA_LANG_INVOKE_STRING_CONCAT_FACTORY);
 		return unitScope.environment.getResolvedJavaBaseType(TypeConstants.JAVA_LANG_INVOKE_STRING_CONCAT_FACTORY, this);
+	}
+		public final ReferenceBinding getJavaLangRuntimeTemplateRuntimeBootstraps() {
+		CompilationUnitScope unitScope = compilationUnitScope();
```
srikanth-sankaran, line 2985: Indentation not proper

srikanth-sankaran, line 2985: Name appears mangled with two Runtime's

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/complete/CompletionScanner.java
```diff
@@ -904,7 +904,7 @@ protected int getNextNotFakedToken() throws InvalidInputException {
 protected int scanForTextBlock() throws InvalidInputException {
 	int lastQuotePos = 0;
```
srikanth-sankaran, line 905: Since this method overrides the super method completely and the super method has various changes specifically for templates, it is likely that completion inside text blocks is broken. We need tests to establish what works and what doesn't and raise follow up tickets

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/StringTemplateExpression.java
```diff
@@ -0,0 +1,229 @@
...
+	 * A list of property descriptors (element type:
+	 * {@link StructuralPropertyDescriptor}),
+	 * or null if uninitialized.
+	 */
+	private static final List PROPERTY_DESCRIPTORS;
+
+	/**
+	 * The "patterns" structural property of this node type (child type: {@link Pattern}).
+	 */
+	public static final ChildPropertyDescriptor TEMPLATE_PROCESSOR =
```
srikanth-sankaran, line 41: Comment is broken due to copy + paste

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/StringTemplateExpression.java
```diff
@@ -0,0 +1,229 @@
...
+	public static List propertyDescriptors(int apiLevel) {
+		return PROPERTY_DESCRIPTORS;
+	}
+
+	public static List propertyDescriptors(int apiLevel, boolean previewEnabled) {
+		if (DOMASTUtil.isStringTemplateSupported(apiLevel, previewEnabled)) {
+			return PROPERTY_DESCRIPTORS;
+		}
+		return null;
+	}
```
srikanth-sankaran, line 154: Is it reasonable that only one checks for isStringTemplateSupported ??

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/StringTemplateExpression.java
```diff
@@ -0,0 +1,229 @@
...
+	 * @param apiLevel the API level; one of the
+	 * <code>AST.JLS*</code> constants
+
+	 * @return a list of property descriptors (element type:
+	 * {@link StructuralPropertyDescriptor})
+	 * @since 3.24
+	 */
+	public static List propertyDescriptors(int apiLevel) {
+		return PROPERTY_DESCRIPTORS;
+	}
```
srikanth-sankaran, line 147: I think this is supposed to return null ??

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/StringTemplate.java
```diff
@@ -0,0 +1,167 @@
...
+package org.eclipse.jdt.core.dom;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.eclipse.jdt.internal.core.dom.util.DOMASTUtil;
+
+/**
+ * StringTemplateExpression  AST node type.
+ *
```
srikanth-sankaran, line 24: Wrong class name

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/StringTemplate.java
```diff
@@ -0,0 +1,167 @@
...
+	 * A list of property descriptors (element type:
+	 * {@link StructuralPropertyDescriptor}),
+	 * or null if uninitialized.
+	 */
+	private static final List PROPERTY_DESCRIPTORS;
+	/**
+	 * The "patterns" structural property of this node type (child type: {@link Pattern}).
+	 */
+	public static final ChildListPropertyDescriptor STRING_TEMPLATE_FRAGMENTS =
+			new ChildListPropertyDescriptor(StringTemplate.class, "fragments", Expression.class, CYCLE_RISK); //$NON-NLS-1$
```
srikanth-sankaran, line 40: Please fix all javadoc copy paste issues

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/StringTemplate.java
```diff
@@ -0,0 +1,167 @@
...
+	 * @param apiLevel the API level; one of the
+	 * <code>AST.JLS*</code> constants
+
+	 * @return a list of property descriptors (element type:
+	 * {@link StructuralPropertyDescriptor})
+	 * @since 3.24
+	 */
+	public static List propertyDescriptors(int apiLevel) {
+		return PROPERTY_DESCRIPTORS;
+	}
```
srikanth-sankaran, line 83: Probably should return null

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTConverter.java
```diff
@@ -2073,9 +2073,52 @@ public Expression convert(org.eclipse.jdt.internal.compiler.ast.Expression expre
...
+		if (expression instanceof org.eclipse.jdt.internal.compiler.ast.TemplateExpression) {
+			return convert((org.eclipse.jdt.internal.compiler.ast.TemplateExpression) expression);
+		}
+		if (expression instanceof org.eclipse.jdt.internal.compiler.ast.StringTemplate) {
+			return convert((org.eclipse.jdt.internal.compiler.ast.StringTemplate) expression);
+		}
 		return null;
 	}
+	public Expression convert(org.eclipse.jdt.internal.compiler.ast.TemplateExpression expression) {
+		StringTemplateExpression templateExpr = new StringTemplateExpression(this.ast);
```
srikanth-sankaran, line 2085: Return type can be sharper

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTConverter.java
```diff
@@ -2073,9 +2073,52 @@ public Expression convert(org.eclipse.jdt.internal.compiler.ast.Expression expre
...
+				fragments.add(convert((org.eclipse.jdt.internal.compiler.ast.TextBlock) fragment));
+			} else {
+				fragments.add(convert(fragment));
+			}
+		}
+		List<Expression> expressions = strTemplate.expressions();
+		for (org.eclipse.jdt.internal.compiler.ast.Expression exp : template.values()) {
+			if (exp == null)
+				continue;
+			expressions.add(convert(exp));
```
srikanth-sankaran, line 2114: This check for null and continue could result in DOM expressions being not the right size. NullLiteral if constructed by Parser would work better here

srikanth-sankaran (commented):
Final batch of comments

srikanth-sankaran:
I still have to review CompliationUnitScope and TextBlock changes. I will share comments on these tomorrow.

srikanth-sankaran:
My recommendation is to defer this until after 4.30 ships and release it as early as possible in 4.31 M1 cycle.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
jarthana: OK, I just checked with Javac and it appears quotes as unicode are not valid delimiters for text blocks. So, I think we are safe here.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
jarthana: I will add some validation and new error reporting. But doubt it will be of much use as it's not common to not have a closing } after a template expression. The scanner usually ends up with one of the closing } of  the next block/method and subsequently reports an unterminated text block error.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/StringTemplate.java
jarthana: Perhaps, not required for the StringTemplate. But we need this for the TemplateExpression. The check was overly cautious; I will remove it.

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/complete/CompletionScanner.java
jarthana: Completion after a template works after a minor fix. Completion inside a template expression doesn't. I have raised #1601 for this.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/StringTemplateExpression.java
jarthana: As discussed offline, the other one will return null anyway as this is preview only. When we promote this as standard, we should fix them all.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
jarthana: I think the only case where we want to create NullLiteral is when nextToken == TokenNameEOF. For the other cases, it's very likely that we will have some other syntax errors already reported.

jarthana:
With the latest patch, the compiler is good shape. But there are still two major issues that affect the editor:
1. Code selection doesn't work inside embedded expressions. This is due to the fact that we parse the embedded expressions in a separate parser and perhaps the state must be shared acrosss parsers

mpalat:
> 2. The DOM nodes need more thoughts on handling the string literals and text blocks that will not have the delimiters such as " or """.
> 
capturing the essence of an offline discussion with @jarthana here [also to show to management that I am not whiling away  time :) ]
(1) A list of nodes without regard to correctness is one option what we discussed. This is probably the simplest solution - leave the correctness questions [eg: is there an expression immediately after String etc]  to the compiler.

(2) Alternatively, I feel, something akin to VariableDeclarationFragment may be suited for this. Essentially, this would be a StringDeclarationFragment having two properties - one for the string fragment and the other for the expression - the embedded expression - the { and } will have to be handled. There is a clear demarkation of what is String and What is Expression, so that setting can be done separately. Agree that there will be an odd man out, but this is not an issue -think of a variable without an initialiser.  It would also mean continuity in source range [satisfying the source range verifier] which was one of the issues Jay raised.
- 2 cents.

jarthana:
In the interest of getting this out to users and have it tested enough, I plan to push these changes tomorrow. I have taken out the DOM changes and plan to take it via #1513.
Other than the DOM part, what needs work is the code assist. I have tested that assist (including completion) works around string templates (but just not inside them).

mpalat:
>  I have taken out the DOM changes and plan to take it via #1513. 
@jarthana I guess you meant the PR number of the DOM changes?

stephan-herrmann:
> The failures in DietRecoveryTest are recovery related and I think we can simply adjust the tests here. However, the ones in NullAnnotationTest need attention. They arise because we are triggering initialization of null analysis and thus resolution of annotations much earlier. I..e at the time of computing default imports itself. Either we need to
> 1. ensure the null analysis is not prematurely triggered or
> 2. instead of adding an import, replace "STR" or "RAW" with their qualified names. (I haven't tried this though)

@jarthana  I happen to see this just now. Does early triggering initialization of null analysis still happen with the released change? If so, could you point me to the code change causing this? "Doing things too early" is a constant concern in the implementation of null analysis ...

jarthana:
> NullAnnotationTest

I have added a kludge in CUScope, line number 772:
this.environment.globalOptions.isAnnotationBasedNullAnalysisEnabled = false;

If you comment out this code and run the NullAnnotationTest, you will see some tests failing.

stephan-herrmann:
> > NullAnnotationTest
> 
> I have added a kludge in CUScope, line number 772: this.environment.globalOptions.isAnnotationBasedNullAnalysisEnabled = false;
> 
> If you comment out this code and run the NullAnnotationTest, you will see some tests failing.

@jarthana as this now hits in #2178 I took a brief look and see the failures in NullAnnotationTest at 21+ when the kludge is removed. Just the other day we launched #2209, because the order of compilation steps is still brittle after quite some efforts in this field. By resolving a field right during the very first step CHECK_AND_SET_IMPORTS, we are certainly not in a position to safely resolve things like annotations, but class `StringTemplate` *is* annotated. So that approach is asking for trouble indeed.

Since we don't even know if this implicit import `StringTemplate#STR` will survive beyond the current round of preview, I wonder how much we should invest her. **Would it be fair to put the magic import under the preview flag for now?**

If that's not good, my next idea would be to encapsulate all fields of ImportBinding, so that we can use lazy initialization, but that would affect all locations accessing this class, so not a tiny change.

Third option: add the magic import much later. This might be OK because we'll never need a resolved field for resolving supertypes :)  -- when resolving annotations we could in theory need to see the field from smth like `@Process(tmpl=STR)` but that will hardly be a realistic case ... so *perhaps* we could find a safe point in time to set the import after the `LookupEnvironment` has been bootstrapped with core types and annotations (incl. meta annotations).

Speaking of it: @srikanth-sankaran , do you think an explicit phase for resolving some primordial things like `@Target`, `@Retention`, `Object`, `Enum`, `String`, `StringTemplate` in a well defined order could help?

stephan-herrmann:
> If that's not good, my next idea would be to encapsulate all fields of ImportBinding, so that we can use lazy initialization, but that would affect all locations accessing this class, so not a tiny change.

That's what I did in #2226. This makes resolving of `StringTemplate.STR` behave more like normal fields.

srikanth-sankaran:
> 
> Since we don't even know if this implicit import `StringTemplate#STR` will survive beyond the current round of preview, I wonder how much we should invest her. **Would it be fair to put the magic import under the preview flag for now?**

Not unreasonable, but we may be postponing the problem.

> Third option: add the magic import much later. This might be OK because we'll never need a resolved field for resolving supertypes :) -- when resolving annotations we could in theory need to see the field from smth like `@Process(tmpl=STR)` but that will hardly be a realistic case ... so _perhaps_ we could find a safe point in time to set the import after the `LookupEnvironment` has been bootstrapped with core types and annotations (incl. meta annotations).

This sounds promising. We could defer it almost to the point where an error would otherwise surface.

> 
> Speaking of it: @srikanth-sankaran , do you think an explicit phase for resolving some primordial things like `@Target`, `@Retention`, `Object`, `Enum`, `String`, `StringTemplate` in a well defined order could help?

Perhaps, but needs deeper study.

Issue #1364, opened 2023-09-07
SelectionParser behavior erratic wrt to live pattern variables upon loop exit


srikanth-sankaran:
```
public class X {
	String xx = "Hello";
	void foo(Object o) {
		do {
			
		} while (!(o instanceof X xxx));
		xxx.foo(o); // F3 on xxx fails
		while (!(o instanceof X yyy)) {
			
		}
		yyy.foo(o); // F3 on yyy works ok
	}
}
```

CompletionParser recognizes both cases of live pattern variables correctly

srikanth-sankaran:
The PR https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1349/ fixes this problem and includes a regression test from here and I will close this as a duplicate

srikanth-sankaran:
Duplicate of #1195

srikanth-sankaran:
Duplicate of https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1195

srikanth-sankaran:
Duplicate of https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1195

Issue #1288, opened 2023-08-14
Open Declaration (F3) sometimes not working for "Pattern Matching for instanceof"


youngroman:
simple code reproducing the issue in Eclipse 2023-06 on Windows 11:
```
	public void test(Object s) {
		if(s instanceof String x) {
			x.length(); // fails
		}
	}
```
when you F3 on the length method you only get a "Current text selection cannot be opened in an editor" error message in the StatusBar. no exception is thrown, which makes this different to all the other reported issues when Pattern Matching is used.

if you remove the curly braces it works:
```
	public void test(Object s) {
		if(s instanceof String x)
			x.length(); // works
	}
```
some more test cases, where it is not always obvious why it works or fails:
```
		if(s instanceof String x) {
			x.length(); x.length(); // 1st fails, 2nd works
		}

		if(s instanceof String x) {
			x.length(); x.length(); x.length(); // 1st fails, 2nd + 3rd works
		}

		if(s instanceof String x) {
			int i; x.length(); // works
		}

		if(s instanceof String x) {
			int i; x.length(); int j; // fails
		}

		if(s instanceof String x) {
			x.length(); int j; // fails
		}
```

iloveeclipse:
Looks related to https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1278.

@srikanth-sankaran :  could you also please take a look here? If the fix is "simple", would be nice to have in 4.29.

srikanth-sankaran:
> Looks related to #1278.
> 
> @srikanth-sankaran : could you also please take a look here? If the fix is "simple", would be nice to have in 4.29.

I will follow up. Prior to my 4 weeks time away from work, I was investigating a cluster of problems related to pattern matching instanceof - Symptoms are varied: CCE during code selection, hover, hyperlink, completion etc. 

The crux of the problem is that, so far, there has not been a construct that JDT had to handle that happens to be an expression but introduces a new variable into the scope. This is a new kind of a scenario and while the main Parser seems to have been upgraded to handle this adequately by all indications, AssistParsers have not been suitably enhanced. 

There have been a series of band-aid fixes that have been slapped on top of each other to mask the problem, but this is completely failing as can be witnessed by the several variants of bug reports mentioned at https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1195#issuecomment-1612277800

A full solution would have to back out half a dozen defect fixes and refix those problems first. Grammar changes are going to be needed too. 

[...]

Issue #1263, opened 2023-08-03
CCE: LocalDeclaration cannot be cast to class ForeachStatement


iloveeclipse:
Using latest build I20230802-1800.
Open `org.eclipse.jdt.internal.launching.DetectVMInstallationsJob` class in the IDE.
Hold "Ctrl" and hover over the "arrows" in lambda expressions in the `run()` method.

Exception reported:
```
java.lang.ClassCastException: class org.eclipse.jdt.internal.compiler.ast.LocalDeclaration cannot be cast to class org.eclipse.jdt.internal.compiler.ast.ForeachStatement (org.eclipse.jdt.internal.compiler.ast.LocalDeclaration and org.eclipse.jdt.internal.compiler.ast.ForeachStatement are in unnamed module of loader org.eclipse.osgi.internal.loader.EquinoxClassLoader @2f9236d8)
	at org.eclipse.jdt.internal.compiler.parser.Parser.consumeEnhancedForStatement(Parser.java:3435)
	at org.eclipse.jdt.internal.compiler.parser.Parser.consumeRule(Parser.java:8471)
	at org.eclipse.jdt.internal.compiler.parser.Parser.parse(Parser.java:13145)
	at org.eclipse.jdt.internal.codeassist.impl.AssistParser.parseBlockStatements(AssistParser.java:2139)
	at org.eclipse.jdt.internal.codeassist.impl.AssistParser.parseBlockStatements(AssistParser.java:1979)
	at org.eclipse.jdt.internal.codeassist.impl.Engine.parseBlockStatements(Engine.java:349)
	at org.eclipse.jdt.internal.codeassist.impl.Engine.parseBlockStatements(Engine.java:312)
	at org.eclipse.jdt.internal.codeassist.SelectionEngine.select(SelectionEngine.java:1081)
	at org.eclipse.jdt.internal.core.Openable.codeSelect(Openable.java:167)
	at org.eclipse.jdt.internal.core.CompilationUnit.codeSelect(CompilationUnit.java:389)
	at org.eclipse.jdt.internal.core.CompilationUnit.codeSelect(CompilationUnit.java:382)
	at org.eclipse.jdt.internal.ui.javaeditor.JavaElementHyperlinkDetector.detectHyperlinks(JavaElementHyperlinkDetector.java:124)
[...]

Issue #1197, opened 2023-06-29
Get rid of separate stacks for managing patterns during parsing


srikanth-sankaran:
`org.eclipse.jdt.internal.compiler.parser.Parser` defines two stacks `patternStack` and `patternLengthStack` - these seem completely redundant and are a distraction. `Parser.astStack` may very well be used to hold patterns during parsing.

PR #1517, opened 2023-10-26
Parser changes for JEP 443

datho7561:
## What it does
Support parsing unnamed variables and unnamed patterns.

## How to test
Create a Java 21 project with preview features enabled. Ensure that the parser can parse unnamed variables and unnamed patterns as described in https://openjdk.org/jeps/443.

## Author checklist


robstryker:
@mpalat  Hi Manoj:  We were looking for feedback on this draft PR while we work to get all the tests green. 

Thanks in advance.

datho7561:
waiting on https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1523 to be merged for many of the tests to be fixed.

mickaelistria:
This change looks pretty good to me.
I see a couple of tests are failing. Those tests may require an update to turn green and consider the PR for thorough review and merge. There are several ways to turn the test green; which one to pick is up to you to chose the one that work best.
1. Do not change the behavior for those tests: new rules and code don't introduce deviation in current behavior so the same rules remain used (it's often very hard as far as I can tell)
2. change the test case: sometimes, it's possible to ensure we go through the same rules as before by tweaking a bit the test case; for example adding 1 import, 1 statement, 1 parameter....)
3. change the test assertion: if the rule cannot be avoided, then unless the message becomes more confusing, it is fine to change the assertion to mention the new message/rule. As those messages are presented to users, I think you can play a bit with the `readableName` to improve those when the default rule name is hard to understand.

HTH

akurtakov:
@mpalat Would you please review this one?

mpalat:
> @mpalat Would you please review this one?

sure.will take a look

mpalat:
> Hi Manoj. Thanks for the review. I think I found a way to reduce the number of changes necessary in `java.g`, but it still requires modifying the Scanner to add a new `UNDERSCORE` token for `_` in order for the grammar to remain LALR(1). Are you okay with this?

@datho7561 - That should be fine - 
I was experimenting with something similar here
"	BeginRecordPattern
	UnderScore
".
However, I would need to see those changes to comment - please resubmit.

datho7561:
@mpalat Alright, I've pushed the changes so that less work is done in `Parser.java`. The tests should be green now.

mpalat:
> @mpalat Alright, I've pushed the changes so that less work is done in `Parser.java`. The tests should be green now.

Noting that I am away from work (or supposed to) for the last couple of days not being in the best of health, am just looking at this as health permits. Also my time is limited in the next few weeks - in short the reviews may not happen fast. 

That said, to have more number of (very good) eyes look at it and to continue the review, have discussed with @srikanth-sankaran and he has agreed to help out with the new patches once he is finished reviewing String templates this is being implemented by Jay with inputs from Srikanth. 

And ideally it would be good to take it one by one since both Unnammed Classes and this have java.g changes and that we need to really concentrate on one problem of grammar change at a time to evolve a minimal, maintainable change. I  would suggest to wait for @mickaelistria unnamed class and then take this patch unnammed patterns . Both of us (@srikanth-sankaran  and I ) felt that the grammar changes should be very minimal in the unnammed patterns - much less than the java.g changes listed in the patch, but we will have to scrutinize/try it out to be sure.

datho7561:
@mpalat Hello Manoj, is it possible for you to review this PR sometime soon? Mickael and I are still working on getting the tests for the Unnamed Classes grammar change PR to pass, whereas on this PR they are passing.

mickaelistria:
@mpalat Thanks for considering the unnamed class as higher priority, but at the moment, I think the patch is not yet ready for review (as long as tests are stuck or some are failing), so I would recommend for the best interest of the project that you remove the unnamed class patch for your current monitoring and instead consider the unnamed variable patch which is assumed complete already.

datho7561:
@srikanth-sankaran when you have time, could you please review this PR?

srikanth-sankaran:
> @srikanth-sankaran when you have time, could you please review this PR?

Yes, I will, thanks.

mpalat:
> > @srikanth-sankaran when you have time, could you please review this PR?
> 
> Yes, I will, thanks.

Thanks @srikanth-sankaran. @datho7561 as informed earlier, am tied up next few days, however will work with Srikanth as time permits

mpalat:
> 
> I was experimenting with something similar here as shown below: " BeginRecordPattern UnderScore ".
> 

@srikanth-sankaran - in here: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1604, I have been experimenting with a grammar change which is very minimal. Haven't taken this to completion, but hopefully this would help - just created a PR - not for merging though, just java.g changes to compare. thx.

datho7561:
> I have been experimenting with a grammar change which is very minimal.

Looking at your change, it seems very similar to what we're doing in this PR, although there are two cases that will require you to make further changes to the grammar:
- Lambdas with multiple underscore parameters (eg. `(_, _) -> 1000`)
- Record pattern matching allows underscores without types: `foo instanceof MyRecordA(_, MyRecordB(_, int myNestedValue))`

srikanth-sankaran:
I second Manoj's observation that a good part of the grammar changes are unnecessary and can be handled otherwise. Mind you, I am not saying that grammar changes proposed are incorrect - The unwritten rule and practice has been to minimize the changes to the grammar - with each evolving version of the Java language and with each shiny new construct we are adding, it has been increasingly difficult to keep the grammar LALR(1) - while a particular set of changes may still leave it LALR(1), there is no telling how it will play with Java 23 features or Java 37 features about which we know next to nothing today.

Let us start with this observation: '_' **was** a valid identifier at some point. At Java 8 we started issuing warnings and at Java 9 they ceased to be valid identifiers. Now the proposed JLS change from here [https://cr.openjdk.org/~abimpoudis/unnamed/jep456-20231017/specs/unnamed-jls.html](url) says "The keyword _ (underscore) may be used in declarations in place of an identifier ([6.1](https://cr.openjdk.org/~abimpoudis/unnamed/jep456-20231017/specs/unnamed-jls.html#jls-6.1)).

I think it is possible to ignore the pedanticism between the reserved keyword '_' and the identifier '_' and return simply a TokenNameIdentifier when we see an '_'.   At the master branch this is what the scanner is doing. Even though '_' is a reserved keyword, it is returning TokenNameIdentifier when it encounters a lone underscore at compliance level 20 and this then results in the diagnostic `org.eclipse.jdt.internal.compiler.problem.ProblemReporter.illegalUseOfUnderscoreAsAnIdentifier(int, int, boolean)` being issued.

I think all that is needed (except for unnamed **and** type elided record components which likely require grammar changes - I'll check), is to adjust what we do when we see '_' as an identifier. Today we if compliance level is 8+ we issue a diagnostic in this call stack:

```
ProblemReporter(ProblemHandler).handle(int, String[], int, String[], int, int, int, ReferenceContext, CompilationResult) line: 129	
ProblemReporter(ProblemHandler).handle(int, String[], String[], int, int, ReferenceContext, CompilationResult) line: 228	
ProblemReporter.handle(int, String[], String[], int, int) line: 2693	
ProblemReporter.illegalUseOfUnderscoreAsAnIdentifier(int, int, boolean) line: 9807	
Parser.pushIdentifier(char[], long) line: 13904	
Parser.pushIdentifier() line: 13912	
Parser.consumeToken(int) line: 10284	
Parser.parse() line: 13139	

```

In Java 21, it is too premature to complain here, We just need to handle the '_' as identifier in other more suitable places - in some cases rejecting them (fields, method parameters ...) and allowing them in others for 21+ + preview enabled mode.

srikanth-sankaran:
From Java 9 till Java 20 when '_' was a reserved keyword, we didn't tokenize it as a reserved keyword, but only as an identifier. We may be able to continue to do that without complicating the grammar is the summation.

srikanth-sankaran:
> > I have been experimenting with a grammar change which is very minimal.
> 
> Looking at your change, it seems very similar to what we're doing in this PR, although there are two cases that will require you to make further changes to the grammar:
> 
> * Lambdas with multiple underscore parameters (eg. `(_, _) -> 1000`)

I don't think this is a problem. The compiler on master parses this code:

```
public class X  {
    interface I {
        void foo(int x, int y);
    }

    I i = (_, _) -> {};
} 
```

and issues the following messages:

```
'_' is a keyword from source level 9 onwards, cannot be used as identifier
'_' is a keyword from source level 9 onwards, cannot be used as identifier
Duplicate parameter _
```

Am I overlooking something ??

> * Record pattern matching allows underscores without types: `foo instanceof MyRecordA(_, MyRecordB(_, int myNestedValue))`

Right, this will likely require changes in the grammar - I will confirm

srikanth-sankaran:
Just to make sure I am not missing some nuance about lambda parameters - can you share a test case that will not parsed correctly on master that involves lambdas and multiple unnamed parameters ? The one you cited above viz `(_, _) -> 1000` is already parsed properly at master
`

srikanth-sankaran:
@datho7561 Your grammar also doesn't handle the changes under 14.11.1 Switch Blocks in [https://cr.openjdk.org/~abimpoudis/unnamed/jep443-20230322/specs/unnamed-jls.html](url)

This is perhaps intentional, but I don't see that being called out in comments on this PR

srikanth-sankaran:
Here as an experiment I have consolidated all the snippets/fragments from [https://openjdk.org/jeps/456](url) that we already parse correctly 

```
import java.util.Queue;
import java.util.stream.Collectors;

record Point(int x, int y) { }
enum Color { RED, GREEN, BLUE }
record ColoredPoint(Point p, Color c) { }
record Box<T extends Ball>(T content) { }

sealed abstract class Ball permits RedBall, BlueBall, GreenBall { }
final  class RedBall   extends Ball { }
final  class BlueBall  extends Ball { }
final  class GreenBall extends Ball { }

public class X {
	
	
	
		ColoredPoint r = new ColoredPoint(new Point(3,4), Color.GREEN);
		if (r instanceof ColoredPoint(Point(int x, int y), Color _)) {
		    
		}
		
		if (r instanceof ColoredPoint(Point(int x, int y), var _)) {
		    
		}
		
		for (int i = 0, _ = sideEffect(); i < 10; i++) {
		}
		
		Queue<Integer> q = null;
		while (q.size() >= 3) {
		    var x = q.remove();
		    var _ = q.remove();       // Unnamed variable
		    var _ = q.remove();       // Unnamed variable
		}
		
		String s = "";
		try {
		    int i = Integer.parseInt(s);
		} catch (NumberFormatException _) {        // Unnamed variable
		    System.out.println("Bad number: " + s);
		} catch (NullPointerException _) {
			// 
		}
		
		class ScopedContext implements AutoCloseable {

			public static ScopedContext acquire() {
				return null;
			}

			@Override
			public void close() throws Exception {
			}
			
		}
		
		try (var _ = ScopedContext.acquire()) {    // Unnamed variable
		    
		}
		Stream<String> stream = null;
		stream.collect(Collectors.toMap(String::toUpperCase, _ -> "NODATA")) ;
		
		
		

		Ball ball = null;
		switch (ball) {
		    case RedBall   _   -> process(ball);
		    case BlueBall  _  -> process(ball);
		    case GreenBall _ -> stopProcessing();
		}
		
		

		Box<? extends Ball> box = null;
		switch (box) {
		    case Box(RedBall   red)     -> processBox(box);
		    case Box(BlueBall  blue)    -> processBox(box);
		    case Box(GreenBall green)   -> stopProcessing();
		    case Box(var       itsNull) -> pickAnotherBox();
		    default -> throw new IllegalArgumentException("Unexpected value: " + box);
		}
		
		switch (box) {
		    case Box(RedBall _)   -> processBox(box);   // Unnamed pattern variable
		    case Box(BlueBall _)  -> processBox(box);   // Unnamed pattern variable
		    case Box(GreenBall _) -> stopProcessing();  // Unnamed pattern variable
		    case Box(var _)       -> pickAnotherBox();  // Unnamed pattern variable
			default -> throw new IllegalArgumentException("Unexpected value: " + box);
		}
	}
	
	private static Object pickAnotherBox() {
		// TODO Auto-generated method stub
		return null;
	}

	private static Object processBox(Box<? extends Ball> box) {
		// TODO Auto-generated method stub
		return null;
	}

	private static Object stopProcessing() {
		return null;
	}

	private static Object process(Ball ball) {
		return null;
	}

	static int sideEffect() {
		return 0;
	}
	
	class Order {}
	
	static int count(Iterable<Order> orders) {
	    int total = 0;
	    for (Order _ : orders)    // Unnamed variable
	        total++;
	    return total;
	}
}
```

srikanth-sankaran:
Here is a collection of snippets we don't parse on master:

```
import java.util.Queue;
import java.util.stream.Collectors;

record Point(int x, int y) { }
enum Color { RED, GREEN, BLUE }
record ColoredPoint(Point p, Color c) { }
record Box<T extends Ball>(T content) { }

sealed abstract class Ball permits RedBall, BlueBall, GreenBall { }
final  class RedBall   extends Ball { }
final  class BlueBall  extends Ball { }
final  class GreenBall extends Ball { }

public class X {
	
	
	
		Box<? extends Ball> box = null;
		switch (box) {
		    case Box(RedBall _), Box(BlueBall _) -> processBox(box);
		    case Box(GreenBall _)                -> stopProcessing();
		    case Box(_)                          -> pickAnotherBox();
		}
		
		if (r instanceof ColoredPoint(Point(int x, _), _)) {
			
		}
	}
}
```

We get:

```
Syntax error, insert "Identifier" to complete TypePattern
Syntax error, insert "Identifier" to complete TypePattern
Syntax error on token ")", -> expected after this token
Syntax error on token(s), misplaced construct(s)
```

srikanth-sankaran:
> Here is a collection of snippets we don't parse on master:
> 

This snippet doesn't compile on using PR too and results in:

```
Syntax error, insert ":" to complete SwitchLabel
Syntax error, insert ")" to complete MethodInvocation
Syntax error on tokens, delete these tokens
```

presumably due to grammar support not being there for multiple case pattern labels

srikanth-sankaran:
All that is valid - but I realize there may be a complication in always scanning '_' as an identifier. In unnamed patterns for which new productions are needed the easiest way to write the new rules is by using '_' expressly, That would introduce a twist. I am experimenting ... stay tuned.

srikanth-sankaran:
(I'll spend some time on this over the weekend, but most of next week except Tuesday I expect to be away on personal time off.)

srikanth-sankaran:
I am attaching an experimental and untested version of a patch that has 3 lines of changes in the grammar file and 6 lines of change in the scanner that should scan and parse everything in this JEP except for cases with multiple record patterns. This is based on the fact that it is axiomatic that the pushdown automaton that implements the LALR parser will **never ever** shift an erroneous token (it may reduce but never shift), The scanner can use this to discriminate whether an '_' is used in place of an identifier (unnamed variables)  or an unnamed pattern. If the parser would shift the new terminal symbol UnnamedPattern, then the '_' we are seeing now is surely for an unnamed pattern and if it would not shift then it must be an unnamed variable. 

(Care should be exercised - the grammar may be over permissive by design, but in this case not so)

```
diff --git a/org.eclipse.jdt.core.compiler.batch/grammar/java.g b/org.eclipse.jdt.core.compiler.batch/grammar/java.g
index 40ed20b..dd347c5 100644
--- a/org.eclipse.jdt.core.compiler.batch/grammar/java.g
+++ b/org.eclipse.jdt.core.compiler.batch/grammar/java.g
@@ -122,6 +122,7 @@
 	BeginCaseElement
 	RestrictedIdentifierWhen
 	BeginRecordPattern
+    UnnamedPattern
 
 --    BodyMarker
 
@@ -1284,6 +1285,8 @@
 /:$compliance 20:/
 
 PatternList -> Pattern
+PatternList -> UnnamedPattern
+PatternList ::= PatternList ',' UnnamedPattern
 PatternList ::= PatternList ',' Pattern
 /.$putCase consumePatternList();  $break ./
 /:$readableName PatternList:/
@@ -3176,3 +3179,4 @@
 
 $end
 -- need a carriage return after the $end
+
diff --git a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
index 0c285df..124b6ee 100644
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
@@ -1435,6 +1435,12 @@
 		this.consumingEllipsisAnnotations = false;
 	} else if (mayBeAtCasePattern(token)) {
 		token = disambiguateCasePattern(token, this);
+	} else if (token == TokenNameIdentifier) {
+		char[] currentIdentifierSource = getCurrentIdentifierSource(); // inefficient , just for proof of concept
+		if (currentIdentifierSource.length == 1 && currentIdentifierSource[0] == '_') {
+			if (this.activeParser.automatonWillShift(TokenNameUnnamedPattern))
+				token = TokenNameUnnamedPattern;
+		}
 	}
 	addTokenToLookBack(token);
 	this.multiCaseLabelComma = false;

```

srikanth-sankaran:
Not saying necessarily (at least yet) we need to course correct - I'll continue to the review, Could you look into the missing grammar changes for multiple record patterns in case please - TIA

org.eclipse.jdt.core.compiler.batch/grammar/java.g
```diff
@@ -1901,6 +1928,11 @@ TypeElidedFormalParameter ::= Modifiersopt Identifier
 /:$readableName TypeElidedFormalParameter:/
 /:$compliance 1.8:/
 
+TypeElidedFormalParameter ::= Modifiersopt '_'
+/.$putCase consumeTypeElidedLambdaParameter(true); $break ./
```
srikanth-sankaran, line 1932: `Modifiersopt` is superfluous in the new rule,

srikanth-sankaran:
I think the grammar changes I have proposed above are good, but don't interact well with the many hacks injecting synthetic tokens - it can be made to work by adjusting those hacks but it will in the process grow and won't be just less than 10 lines anymore.

As things stand, the alternate grammar I proposed parses this program incorrectly:

```
import java.util.Queue;
import java.util.stream.Collectors;

record Point(int x, int y) { }
enum Color { RED, GREEN, BLUE }
record ColoredPoint(Point p, Color c) { }
record Box<T extends Ball>(T content) { }

sealed abstract class Ball permits RedBall, BlueBall, GreenBall { }
final  class RedBall   extends Ball { }
final  class BlueBall  extends Ball { }
final  class GreenBall extends Ball { }

public class X {
	
		Box<? extends Ball> box = null;
		switch (box) {
		    case Box(_)                          -> pickAnotherBox();
		    default -> throw new IllegalArgumentException("Unexpected value: " + box);
		}
	}

	private static Object pickAnotherBox() {
		return null;
	}    
}
```

In `case Box(\_)'  Box is parsed as a message invocation rather than as a Record pattern due to the non-injection of the synthetic symbol `BeginCaseElement` - this happens because my changes haven't updated `org.eclipse.jdt.internal.compiler.parser.Scanner.Goal.PatternRules` 

At this point, I am inclined to go with the original proposed grammar changes (with bug fixes and enhancements for missing parts) rather than advocate an alternate approach.  I'll continue the review on that basis

org.eclipse.jdt.core.compiler.batch/grammar/java.g
```diff
@@ -1269,30 +1273,48 @@ TypePattern ::= Modifiersopt Type 'Identifier'
...
+RecordPattern ::= Modifiersopt ReferenceType PushLPAREN ComponentPatternListopt PushRPAREN
 /.$putCase consumeRecordPattern(); $break ./
 /:$readableName RecordPattern:/
 /:$compliance 20:/
 
-PatternListopt ::=  $empty
+ComponentPatternListopt ::=  $empty
 /.$putCase consumePatternListopt(); $break ./
 /:$readableName PatternListopt:/
 /:$compliance 20:/
```
srikanth-sankaran, line 1284: Why would we allow the readable name to deviate from the non-terminal's name ?

org.eclipse.jdt.core.compiler.batch/grammar/java.g
```diff
@@ -1269,30 +1273,48 @@ TypePattern ::= Modifiersopt Type 'Identifier'
...
+/.$putCase consumeTypePattern(); $break ./
+/:$readableName TypePattern:/
+/:$compliance 21:/
+
+ComponentPattern -> UnnamedPattern
+/:$compliance 21:/
+
+UnnamedPattern ::= '_'
+/.$putCase consumeUnnamedPattern(); $break ./
+/:$readableName UnnamedPattern:/
```
srikanth-sankaran, line 1313: Check the uses of `org.eclipse.jdt.internal.compiler.parser.Scanner.Goal.PatternRules` to see if UnnamedPattern must be added there

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
```diff
@@ -3254,6 +3254,9 @@ public int scanIdentifierOrKeyword() {
 		//have a length which is <= 12...but there are lots of identifier with
 		//only one char....
 		if ((length = this.currentPosition - this.startPosition) == 1) {
+			if (this.source[this.startPosition] == '_') {
+				return TokenNameUNDERSCORE;
+			}
 			return TokenNameIdentifier;
```
srikanth-sankaran, line 3260: Technically you need similar change in the else branch to handle underscores in unicode

srikanth-sankaran:
@datho7561 Could you also please create regression tests with the snippets from https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1517#issuecomment-1816033939 and https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1517#issuecomment-1816054518 ? I always find it a useful exercise to make sure that every snippet from the JEP is tested - this is a form of coverage testing so to speak and would have caught the grammar not handling cases with multiple record patterns for example

org.eclipse.jdt.core.compiler.batch/grammar/java.g
```diff
@@ -1269,30 +1273,48 @@ TypePattern ::= Modifiersopt Type 'Identifier'
...
 -----------------------------------------------
 
+---
+TypePattern ::= Modifiersopt Type '_'
+/.$putCase consumeTypePattern(); $break ./
+/:$readableName TypePattern:/
+/:$compliance 21:/
+
+ComponentPattern -> UnnamedPattern
+/:$compliance 21:/
```
srikanth-sankaran, line 1309: Please group all rules related to the same nonterminal together

org.eclipse.jdt.core.compiler.batch/scripts/.gitignore
```diff
@@ -0,0 +1 @@
+*.class
```
srikanth-sankaran, line 1: Not sure if this is supposed to be merged in -

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/core/compiler/IProblem.java
```diff
@@ -2564,4 +2564,10 @@ public interface IProblem {
 	 * @since 3.35
 	 */
 	int SyntheticAccessorNotEnclosingMethod = MethodRelated + 1990;
+
+	/**
+	 * @since 3.36
+	 * @noreference preview feature
+	 */
+	int FieldNameCannotBeUnderscore = PreviewRelated + 2000;
```
srikanth-sankaran, line 2572: This program:

```
public class X {
	int _;
	
	X(int _) {
		try {
			
		} catch (NullPointerException _) {
			
		}
	}
	
}
```

triggers two errors from javac but only one from ECJ - method/ctor parameters may not be unnamed. Please also check if any of the existing IProblem constants may be reused. 

Javac error message is much more informational - we may want to emulate

srikanth-sankaran:
The downside of recognizing underscores as separate token distinct from identifiers is that error messages lack clarity.
Given:

```
public class X {
	
	X(int _) {
		try {
			_ = 10;
		} catch (NullPointerException _) {
			System.out.println(_);
		}
	}
	
}
```

we report:

```
Syntax error, insert "-> LambdaBody" to complete ArgumentList
Syntax error, insert "-> LambdaBody" to complete Expression
```

which is bizarre.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/FieldDeclaration.java
```diff
@@ -70,6 +70,9 @@ public FieldDeclaration(	char[] name, int sourceStart, int sourceEnd) {
 }
 
 public FlowInfo analyseCode(MethodScope initializationScope, FlowContext flowContext, FlowInfo flowInfo) {
+	if (this.isUnnamed(initializationScope)) {
+		initializationScope.problemReporter().fieldNameCannotBeUnderscore(this);
+	}
```
srikanth-sankaran, line 75: This should be done in resolve phase not analyze phase.

srikanth-sankaran:
I'll continue the review, unfortunately comments will come piecemeal as I am busy this week with intermittent time off

mpalat:
> 		switch (box) {
> 		    case Box(_)                          -> pickAnotherBox();
> 		    default -> throw new IllegalArgumentException("Unexpected value: " + box);
> 		}

@srikanth-sankaran : As far as the type-elided Record Pattern is concerned, the following grammar change takes care of this in java.g

TypePattern ::= Modifiersopt Type UnderScore
/.$putCase consumeTypePattern(true); $break ./
/:$readableName TypePattern:/

TypePattern ::= UnderScore
/.$putCase consumeTypePattern(false); $break ./
/:$readableName TypePattern:/

And in parser, this can push a Var Type internally for type reference. I've updated https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1604 with this change

srikanth-sankaran:
> > ```
> >   switch (box) {
> >       case Box(_)                          -> pickAnotherBox();
> >       default -> throw new IllegalArgumentException("Unexpected value: " + box);
> >   }
> > ```
> 
> @srikanth-sankaran : As far as the type-elided Record Pattern is concerned, the following grammar change takes care of this in java.g
> 
> TypePattern ::= Modifiersopt Type UnderScore /.$putCase consumeTypePattern(true); $break ./ /:$readableName TypePattern:/
> 
> TypePattern ::= UnderScore /.$putCase consumeTypePattern(false); $break ./ /:$readableName TypePattern:/
> 
> And in parser, this can push a Var Type internally for type reference. I've updated #1604 with this change

@mpalat I don't see any substantive difference between this grammar and the one proposed by @datho7561

The differences I see are:

1. Your grammar has not spelled out what underscore means and leaves it to scanner to manufacture it. While th e original proposed grammar makes it clear what `UNDERSCORE` is
2. @datho7561 proposed grammar renames some productions (`PatternList -> ComponentPatternList` etc) - while these are not essential, they make the grammar more readable I would say and can be accepted on that basis.
3. Will your grammar handle lambdas of the form `(\_)->{}` I don't see the equivalent of @datho7561's change of the form
```
TypeElidedFormalParameter ::= Modifiersopt '_'
/.$putCase consumeTypeElidedLambdaParameter(true); $break ./
```
WIthout this rule the scanner will have to return different tokens when the Scanner sees an \_ and I don't see how that disambiguation happenbs

datho7561:
I figured out the grammar change that's needed to get this case working:

```
case Box(RedBall _), Box(BlueBall _) -> processBox(box);
```

I haven't pushed the change yet as I'm trying to write some tests for this case.

In response to [this comment](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1517#discussion_r1398641829), I started looking into improving error markers when underscore is used somewhere it's not valid (parameter name, method name, class name, ...). With the changes in this PR, the parser will fail to parse a method named `_`, so the error message it generates is not helpful. The workaround for this it to modify the parser further to accept either 'Identifier' or 'Underscore' as a method name, then generate the error message when creating the MethodDeclaration node. If I do this for all the places where you aren't allowed to use `_` anymore (since Java 9), that would result in a lot of grammar changes. @srikanth-sankaran I'm starting to think your change (along with any more code needed) would be better, since less work would need to be done in the grammar to generate the error messages nicely.

I'll focus on adding the changes needed to parse a list of patterns in a case statement, adding tests for this, and addressing your comments for now.

srikanth-sankaran:
>If I do this for all the places where you aren't allowed to use `_` anymore (since Java 9), that would result in a lot of grammar changes. @srikanth-sankaran I'm starting to think your change (along with any more code needed) would be better, since less work would need to be done in the grammar to generate the error messages nicely.

While emitting as clear an error message as possible is a goal, this is also an unlikely scenario - a class or method named \_ - I won't bend over backwards to handle it in the grammar. Or opt for a grammar that has other complications in the scanner. So just on this account, we don't have to look at abandoning the grammar you have imo.

datho7561:
Handling a list of patterns in a case statement is taking longer than expected. The original change I had wasn't correct, so I spent most of last week figuring out the changes I needed to make to get it to parse correctly.

I have the parsing aspect working now. However, I need to rewrite part of the AST so that a Guard can apply to a list of patterns. This probably means rewriting the code generation. I also still need to detect when multiple pattern variables are declared, and provide an appropriate error message.

srikanth-sankaran:
I am back at work and will continue with my review this week. A plausible approach is to split the work into two PRs - unnamed patterns and variables can go in one merge. The multiple case patterns is an orthogonal feature not organically related to the rest of JEP and can come in a separate follow up PR/ticket.

datho7561:
> The multiple case patterns is an orthogonal feature not organically related to the rest of JEP and can come in a separate follow up PR/ticket

Okay, I'll work on that aspect of the feature in a different PR. I'll focus on addressing the other comments you had on this PR first.

org.eclipse.jdt.core.compiler.batch/scripts/.gitignore
datho7561: this is intentional; I would like to avoid staging `GenerateParserScript.class`

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/core/compiler/IProblem.java
datho7561: I adjusted the new error message I added to be similar to the javac one, and use this new error message consistently when `_` is used in the wrong spot when using Java 21 with preview enabled.

org.eclipse.jdt.core.compiler.batch/grammar/java.g
datho7561: I don't think it needs to be added to the list of PatternRules, since you can't use an unnamed pattern by itself as a case label, and if I understand correctly, `PatternRules` is for all the possible starting states of parsing a case label with a pattern in it.

datho7561:
@srikanth-sankaran alright, I think I've addressed all the feedback. I've added some more tests as well.

Just to confirm:
- Due to the grammar changes, the error messages for classes and methods that are named `_` is "unexpected token" instead of something nicer. It's possible to improve this error message, but will require more grammar changes.
- I'm going to work on lists of patterns in a case label as a separate PR, I created https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1643 to track this issue

srikanth-sankaran:
@datho7561 Great, I will pull these commits and resume the review. Hope to conclude in a couple of days and provide consolidated feedback

srikanth-sankaran:
> @datho7561 Could you also please create regression tests with the snippets from [#1517 (comment)](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1517#issuecomment-1816033939) and [#1517 (comment)](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1517#issuecomment-1816054518) ? I always find it a useful exercise to make sure that every snippet from the JEP is tested - this is a form of coverage testing so to speak and would have caught the grammar not handling cases with multiple record patterns for example

@datho7561 Looks like this review feedback was not incorporated ? I get two compile errors with this snippet which javac 21 compiles fine:

```
import java.util.Queue;
import java.util.stream.Collectors;

record Point(int x, int y) { }
enum Color { RED, GREEN, BLUE }
record ColoredPoint(Point p, Color c) { }
record Box<T extends Ball>(T content) { }

sealed abstract class Ball permits RedBall, BlueBall, GreenBall { }
final  class RedBall   extends Ball { }
final  class BlueBall  extends Ball { }
final  class GreenBall extends Ball { }

public class X {
	
	
	
	public static void main(String[] args) throws Exception {
		ColoredPoint r = new ColoredPoint(new Point(3,4), Color.GREEN);
		if (r instanceof ColoredPoint(Point(int x, int y), Color _)) {
		    
		}
		
		if (r instanceof ColoredPoint(Point(int x, int y), var _)) {
		    
		}
		
		for (int i = 0, _ = sideEffect(); i < 10; i++) {
		}
		
		Queue<Integer> q = null;
		while (q.size() >= 3) {
		    var x = q.remove();
		    var _ = q.remove();       // Unnamed variable
		    var _ = q.remove();       // Unnamed variable
		}
		
		String s = "";
		try {
		    int i = Integer.parseInt(s);
		} catch (NumberFormatException _) {        // Unnamed variable
		    System.out.println("Bad number: " + s);
		} catch (NullPointerException _) {
			// 
		}
		
		class ScopedContext implements AutoCloseable {

			public static ScopedContext acquire() {
				return null;
			}

			@Override
			public void close() throws Exception {
			}
			
		}
		
		try (var _ = ScopedContext.acquire()) {    // Unnamed variable
		    
		}
		Stream<String> stream = null;
		stream.collect(Collectors.toMap(String::toUpperCase, _ -> "NODATA")) ;
		
		
		

		Ball ball = null;
		switch (ball) {
		    case RedBall   _   -> process(ball);
		    case BlueBall  _  -> process(ball);
		    case GreenBall _ -> stopProcessing();
		}
		
		

		Box<? extends Ball> box = null;
		switch (box) {
		    case Box(RedBall   red)     -> processBox(box);
		    case Box(BlueBall  blue)    -> processBox(box);
		    case Box(GreenBall green)   -> stopProcessing();
		    case Box(var       itsNull) -> pickAnotherBox();
		    default -> throw new IllegalArgumentException("Unexpected value: " + box);
		}
		
		switch (box) {
		    case Box(RedBall _)   -> processBox(box);   // Unnamed pattern variable
		    case Box(BlueBall _)  -> processBox(box);   // Unnamed pattern variable
		    case Box(GreenBall _) -> stopProcessing();  // Unnamed pattern variable
		    case Box(var _)       -> pickAnotherBox();  // Unnamed pattern variable
			default -> throw new IllegalArgumentException("Unexpected value: " + box);
		}
	}
	
	private static Object pickAnotherBox() {
		// TODO Auto-generated method stub
		return null;
	}

	private static Object processBox(Box<? extends Ball> box) {
		// TODO Auto-generated method stub
		return null;
	}

	private static Object stopProcessing() {
		return null;
	}

	private static Object process(Ball ball) {
		return null;
	}

	static int sideEffect() {
		return 0;
	}
	
	class Order {}
	
	static int count(Iterable<Order> orders) {
	    int total = 0;
	    for (Order _ : orders)    // Unnamed variable
	        total++;
	    return total;
	}
}
```

Taking a look at the grammar, it looks like enhanced for loop related production need updates too and currently lacking in java.g

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Argument.java
```diff
@@ -116,19 +116,25 @@ private Annotation[] getCorrespondingRecordComponentAnnotationsIfApplicable(Abst
 		return null;
 	}
 	public TypeBinding bind(MethodScope scope, TypeBinding typeBinding, boolean used) {
+		if (this.isUnnamed(scope)) {
+			scope.problemReporter().illegalUseOfUnderscoreAsAnIdentifier(this.sourceStart, this.sourceEnd, scope.compilerOptions().sourceLevel > ClassFileConstants.JDK1_8, true);
+		}
+
 		TypeBinding newTypeBinding = createBinding(scope, typeBinding); // basically a no-op if createBinding() was called before
```
srikanth-sankaran, line 123: We need to check if `this` is an argument of a lambda expression ?? Otherwise we reject lambda parameters from being unnamed and report an error `As of release 21, '_' is only allowed to declare unnamed patterns, local variables, exception parameters or lambda parameters`

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/problem/ProblemReporter.java
```diff
@@ -12455,6 +12456,14 @@ public void falseLiteralInGuard(Expression exp) {
 			exp.sourceStart,
 			exp.sourceEnd);
 }
+public void underscoreCannotBeUsedHere(FieldDeclaration fieldDeclaration) {
+	this.handle(
```
srikanth-sankaran, line 12460: This method appears unused ??

srikanth-sankaran:
This code compiles with ECJ but javac rejects it correctly:

```
public class X {
		void foo(int x, int y) {
			int _;
			int _;
		}
}
```

JEP 456 states:

```
Declaring an unnamed variable does not place a name in scope, so the variable cannot be written or read after it is initialized. An initializer must be provided for an unnamed variable declared in a local variable declaration statement or in the resource specification of a try-with-resources statement.
```

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Pattern.java
```diff
@@ -39,6 +39,8 @@ public abstract class Pattern extends Expression {
 
 	public int nestingLevel = 0;
 
+	public int index = -1; // denoting position
+
 	@Override
```
srikanth-sankaran, line 44: Comment can be more descriptive, As I look at it now, I am unable to conceive what this is for.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/RecordPattern.java
```diff
@@ -192,11 +192,11 @@ private void setAccessorsPlusInfuseInferredType(BlockScope scope) {
 		} else {
 			for (int i = 0; i < components.length; i++) {
 				Pattern p = this.patterns[i];
-				if (!(p instanceof TypePattern))
+				if (!(p instanceof VariablePattern))
 					continue;
-				TypePattern tp = (TypePattern) p;
```
srikanth-sankaran, line 197: Can you explain what exactly is achieved by introducing a new layer in the class hierarchy here ? I am not sure `VariablePattern` is a suitable name - but can say for sure only after I understand the purpose

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/parser/JEP443UnnamedVariableTest.java
```diff
@@ -0,0 +1,424 @@
...
+import org.eclipse.jdt.internal.compiler.classfmt.ClassFileConstants;
+import org.eclipse.jdt.internal.compiler.env.ICompilationUnit;
+import org.eclipse.jdt.internal.compiler.impl.CompilerOptions;
+import org.eclipse.jdt.internal.compiler.parser.Parser;
+import org.eclipse.jdt.internal.compiler.problem.DefaultProblemFactory;
+import org.eclipse.jdt.internal.compiler.problem.ProblemReporter;
+import org.junit.Test;
+
+public class JEP443UnnamedVariableTest  extends AbstractCompilerTest {
+	public static boolean optimizeStringLiterals = false;
```
srikanth-sankaran, line 29: Kust checking: Have you verified the behavior of all these tests with javac ??

srikanth-sankaran (changes requested):

srikanth-sankaran:
@datho7561 I have made a pass over all the changes and shared my comments. Next steps:

1. Review the review comments and incorporate the changes requested.
2. I am requesting @jarthana for a deeper review of the following files:
```
org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Pattern.java
org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/RecordPattern.java
org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypePattern.java
org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/UnnamedPattern.java
org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/VariablePattern.java

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/parser/JEP443UnnamedVariableTest.java
org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/UnnamedPatternsAndVariablesTest.java
```
3. I am requesting @jarthana for a "stress test" of this implementation. Jay, please note that this PR does not implement one part of the JEP having to do with multiple patterns in a case. So you may discount any failures you encounter in that part.
4. @datho7561 It would be better to add some tests that actually deconstruct a record field and do something with it rather than of the form: `r = null; if (r instanceof ...)` 
5. All snippets from the JEP should be included in regression tests. 

Thanks

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/problem/ProblemReporter.java
```diff
@@ -9790,9 +9790,10 @@ public void useEnumAsAnIdentifier(int sourceStart, int sourceEnd) {
 		sourceStart,
 		sourceEnd);
 }
-public void illegalUseOfUnderscoreAsAnIdentifier(int sourceStart, int sourceEnd, boolean reportError) {
+public void illegalUseOfUnderscoreAsAnIdentifier(int sourceStart, int sourceEnd, boolean reportError, boolean unusedVariablesSupported) {
 	this.underScoreIsError = reportError;
 	int problemId = (reportError) ? IProblem.ErrorUseOfUnderscoreAsAnIdentifier : IProblem.IllegalUseOfUnderscoreAsAnIdentifier;
```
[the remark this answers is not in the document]
datho7561: I checked the usages using Eclipse, and it seems like `IProblem.ErrorUseOfUnderscoreAsAnIdentifier` and `IProblem.IllegalUseOfUnderscoreAsAnIdentifier` are only referenced in this method and the method that determines problem severity. I also did a `grep` of eclipse.jdt.ui and didn't see any references to these problems there at all. I'll perform the simplification.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/parser/JEP443UnnamedVariableTest.java
datho7561: I'll migrate these tests (minus the two that you pointed out aren't helpful) to the regression test suite.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/RecordPattern.java
datho7561: The main reason I avoided using or subclassing `TypePattern` was that many of the methods expect that the type has a source range, even when it's an elided type. I ended up using many hacks in order to get unused variables working. I'll take another look to see if I can avoid creating a separate class.

srikanth-sankaran: > The main reason I avoided using or subclassing `TypePattern` was that many of the methods expect that the type has a source range, even when it's an elided type. I ended up using many hacks in order to get unused variables working. I'll take another look to see if I can avoid creating a separate class.

Let us wait for feedback from @jarthana in this. I am not saying what you have done is wrong. Jay let us have your opinion on this. TIA.

jarthana: It's interesting to note that the TypePattern gets its TypeReference from this.local.type, which comes from VariablePattern. So, I am not sure why we can't make it work. If you can tell me what breaks without this change, that'll help.

jarthana: > It's interesting to note that the TypePattern gets its TypeReference from this.local.type, which comes from VariablePattern. So, I am not sure why we can't make it work. If you can tell me what breaks without this change, that'll help. Edit: I see that we deal with a similar case in lambda arguments where the argument's type is set to null when explicit type is provided. We might be able to make do with null?

I played around and was able to make new tests pass without the VariablePattern. Losing the FakeTypeReference, however, might be a bit more involving. So, I am not particular about the latter change.

Other point is, do we need to do the analyseCode() for the unnamed pattern? Currently, UnnamedPattern delegates to the one in TypePattern but not sure we even need to do that.

datho7561: I deleted the UnnamedPattern class entirely and am just using TypePattern. I also deleted VariablePattern and FakeTypeReference.

rgrunber:
@srikanth-sankaran , I've just heard (from Andrey) that Jay will be on vacation for the rest of the month.

srikanth-sankaran:
> @srikanth-sankaran , I've just heard (from Alex) that Jay will be on vacation for the rest of the month.

I heard that myself a couple of days ago. @datho7561 : Can you clarify where we are wrt to this work ? Have all review comments been incorporated ? Has the parallel PR for multiple patterns in case been worked on ? While the latter is independent for gatekeeping reasons @jarthana may _prefer_ to test them together. Based on where we are, we can plan the next steps

srikanth-sankaran:
> > @srikanth-sankaran , I've just heard (from Alex) that Jay will be on vacation for the rest of the month.
> 
> I heard that myself a couple of days ago. @datho7561 : Can you clarify where we are wrt to this work ? Have all review comments been incorporated ? Has the parallel PR for multiple patterns in case been worked on ? While the latter is independent for gatekeeping reasons @jarthana may _prefer_ to test them together. Based on where we are, we can plan the next steps

(Jay is the best person for reviewing half a dozen files I have called out - It will require high investment from me to familiarize myself with that part of the code and review myself and I am trying to wrap up some unrelated work for 4.31 M1. Due to some reasons, Jay is also in a better position to stress test this and issue a go than me )

datho7561:
> Can you clarify where we are wrt to this work ? Have all review comments been incorporated ?

Yes, I have incorporated all feedback on this PR.

> Has the parallel PR for multiple patterns in case been worked on ?

I have a branch in progress, but I haven't opened a PR yet. From what I recall, it's working but the tests are not passing. I have been focusing on unnamed classes first.

jarthana:
OK, I found some time to spend on this today and may be tomorrow. I reviewed the half a dozen pattern related AST files files again and fine with the current changes. I also did some testing and found the following cases having issues. 
1. The following code is rejected by ECJ, but allowed by Javac:
```
    	for (final int _ : new int[1]) {}
```
2.
```
public class X {
	record Paper(int color) {}
	record Box<T>(T a) {}
		Box<?> b = new Box<>(null);
		boolean res = b instanceof Box(Paper _);
		assert (res == false);
	}
}
```
3. The following code when compiled with ecj and run, throws a VerifyError:
```
public class X {
	record Paper(int color) {}
	record Box<T>(T a) {}
		foo(null, null);
	}
	public static void foo(String abc, String def) {
		Box<?> p = new Box<>(new Paper(0));
		boolean b = false;
		switch (p) {
			case Box(Paper _) -> {
				b = true;
				break;
			}
			default -> {
				b = false;
				break;
			}
		}
		System.out.println(b);
	}
}
```

datho7561:
It seems like 2. is an existing bug, since I can replicate it if I replace the `_` with `a`.

datho7561:
3. is also an existing bug, maybe related to 2.

datho7561:
@jarthana Okay, I fixed 1. I'll fix 2. and 3. in a separate PR (https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1816)

datho7561:
I can't replicate the test failure

srikanth-sankaran:
> I am Okay with the changes in the specific set of files I looked at. I still see some requested from Srikanth pending. Once they are resolved, we are good to go.

@datho7561, please go over all the past comments to make sure everything is addressed. Once you confirm, I'll make a sanity check and help close this

datho7561:
The main thing that isn't addressed is this bug in code generation that I created a separate PR for: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1816

Other than that it should be good.

srikanth-sankaran:
A heads up that fundamental restructuring of pattern matching support is happening via https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1885 which should get merged in a day or two. I'll see what if any changes would be needed to harmonize this work with that PR and help align them.

srikanth-sankaran:
> A heads up that fundamental restructuring of pattern matching support is happening via #1885 which should get merged in a day or two. I'll see what if any changes would be needed to harmonize this work with that PR and help align them.

This is done and if you are update your changes, I can take it forward. Thanks

srikanth-sankaran:
OK, I am starting a quick pass over this and will try to dedicate the next 2-3 days to this and see if this can be merged.
My plan is to raise follow up tickets for any issues discovered and not block this PR unless something absolutely in the nature of a show stopper is uncovered.

srikanth-sankaran:
I am having a tough time rebasing to master - I know this has been a recurring exercise and I wanted to take it off of your hands. But having a tough time, hitting various parser generator bugs ... I will continue to look into this, in the meantime if you see this message and can spare the cycles to update against master and force-push, that would be super

srikanth-sankaran:
> I am having a tough time rebasing to master - I know this has been a recurring exercise and I wanted to take it off of your hands. But having a tough time, hitting various parser generator bugs ... I will continue to look into this, in the meantime if you see this message and can spare the cycles to update against master and force-push, that would be super

OK, I have figured out the parser generator breakage. But merge is still proving to be unintutive. I think I'll squash all commit into one and then rebase. This may result in a simpler workflow

srikanth-sankaran:
> > I am having a tough time rebasing to master - I know this has been a recurring exercise and I wanted to take it off of your hands. But having a tough time, hitting various parser generator bugs ... I will continue to look into this, in the meantime if you see this message and can spare the cycles to update against master and force-push, that would be super
> 
> OK, I have figured out the parser generator breakage. But merge is still proving to be unintutive. I think I'll squash all commit into one and then rebase. This may result in a simpler workflow

No, spoke too soon it seems. I get various errors while running the parser generator scripts:

```
  [java] parser6.rsc creation complete
     [java] parser7.rsc creation complete
     [java] parser8.rsc creation complete
     [java] parser9.rsc creation complete
     [java] java.lang.ArrayIndexOutOfBoundsException: Index 61296 out of bounds for length 61296
     [java] 	at org.apache.tools.ant.taskdefs.ExecuteJava.execute(ExecuteJava.java:194)
     [java] 	at org.apache.tools.ant.taskdefs.Java.run(Java.java:892)
     [java] 	at org.apache.tools.ant.taskdefs.Java.executeJava(Java.java:232)
     [java] 	at org.apache.tools.ant.taskdefs.Java.executeJava(Java.java:136)
     [java] 	at org.apache.tools.ant.taskdefs.Java.execute(Java.java:109)
     [java] 	at org.apache.tools.ant.UnknownElement.execute(UnknownElement.java:299)
     [java] 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
     [java] 	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
     [java] 	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
     [java] 	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
     [java] 	at org.apache.tools.ant.dispatch.DispatchUtils.execute(DispatchUtils.java:99)
     [java] 	at org.apache.tools.ant.Task.perform(Task.java:350)
     [java] 	at org.apache.tools.ant.Target.execute(Target.java:449)
     [java] 	at org.apache.tools.ant.Target.performTasks(Target.java:470)
     [java] 	at org.apache.tools.ant.Project.executeSortedTargets(Project.java:1401)
     [java] 	at org.apache.tools.ant.Project.executeTarget(Project.java:1374)
     [java] 	at org.apache.tools.ant.helper.DefaultExecutor.executeTargets(DefaultExecutor.java:41)
     [java] 	at org.eclipse.ant.internal.launching.remote.EclipseDefaultExecutor.executeTargets(EclipseDefaultExecutor.java:34)
     [java] 	at org.apache.tools.ant.Project.executeTargets(Project.java:1264)
```

srikanth-sankaran:
OK, I am giving up on the rebase/merge conflicts/ approach. I will review the PR as is. Before merging it needs to be brought to uptodate state anyway. This you can handle.

srikanth-sankaran:
> OK, I am giving up on the rebase/merge conflicts/ approach. I will review the PR as is. Before merging it needs to be brought to uptodate state anyway. This you can handle.

Summary: I am not blocked. I am reviewing by comparing your branch HEAD against master as opposed to rebase and then compare with previous version. It should all wash out

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/AbstractVariableDeclaration.java
```diff
@@ -144,4 +145,15 @@ public void setDepth(int depth) {
...
+
+	/**
+	 * Returns true if this variable is an unnamed variable (_) and false otherwise.
+	 *
+	 * @param scope used to determine source level
+	 */
+	public boolean isUnnamed(BlockScope scope) {
+		return ((this.name.length == 1 && this.name[0] == '_') || ("\\u005F".equals(this.name.toString()))) //$NON-NLS-1$
+				&& scope.compilerOptions().sourceLevel >= ClassFileConstants.JDK21
+				&& scope.compilerOptions().enablePreviewFeatures;
```
srikanth-sankaran, line 157: Curious - is the ` ("\\u005F".equals(this.name.toString())))` really required ? Won't the scanner internalize the unicode and return an '_' ??

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
```diff
@@ -3082,13 +3082,20 @@ public int scanIdentifierOrKeyword() {
...
 		}
 		data = this.source;
 		index = this.startPosition;
 	} else {
-		if ((length = this.withoutUnicodePtr) == 1)
+		if ((length = this.withoutUnicodePtr) == 1) {
+			if (this.withoutUnicodeBuffer[0] == '_') {
+				return TokenNameUNDERSCORE;
+			}
```
srikanth-sankaran, line 3096: This is what I meant. When '_' occurs as unicode, the scanner converts that to '_' - No ??

srikanth-sankaran (approved):
Ok, I have made a full pass over the changes and exhausted all questions and comments. Please proceed to

1. Rebase against master - there are some more pattern related changes on master that need to be carefully picked up.
2. I have provided a comment about unicode - it is more a nit and academic curiosity actually. Please take a look.
3. Do you swear under penalty of perjury that you have tested every snippet in `org.eclipse.jdt.core.tests.compiler.regression.UnnamedPatternsAndVariablesTest` and in general every modified test to compare behavior with `JDK21`  ?? :) 

If all this is addressed, we are ready to merge this into master.

srikanth-sankaran:
FYI, I have been drastically overhauling our patterns implementation to make it simpler to understand and maintain:

See:

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1884 - [Patterns] Alternate, simpler pattern binding management implementation
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1897 - [Patterns] Cleanup/refactor patterns implementation
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1887 - [Patterns] ECJ tolerates erroneous redeclaration of pattern bindings in some cases. 
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1856 - [switch][record patterns] NPE: Cannot invoke "org.eclipse.jdt.internal.compiler.lookup.MethodBinding.isStatic()
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1853 - [switch][pattern] Scope of pattern binding extends illegally resulting in wrong diagnostic 
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1835 - [Switch][patterns][record] AssertionError at org.eclipse.jdt.internal.compiler.ast.YieldStatement.addSecretYieldResultValue(YieldStatement.java
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1788 - Inference issue between the diamond syntax and pattern matching (switch on objects) 
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1759 - Pattern variable is not recognized in AND_AND_Expression
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1726 - [21] Pattern-matching in instanceof doesn't work
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1485- ECJ hangs when pattern matching code is used in a nested conditional expression
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1364 - SelectionParser behavior erratic wrt to live pattern variables upon loop exit
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1288 - Open Declaration (F3) sometimes not working for "Pattern Matching for instanceof
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1263 - CCE: LocalDeclaration cannot be cast to class ForeachStatement
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1278- Wrong method redirect
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1197 - Get rid of separate stacks for managing patterns during parsing
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1195- Open declaration results in ClassCastException: LocalDeclaration cannot be cast to LambdaExpression
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1076- ECJ accepts invalid Java code instanceof final Type 
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/968 - JavaElementHyperlinkDetector: Parser runs into NegativeArraySizeException in some cases
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/769- Open Declaration(F3) broken in pattern instanceof
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/567 - [Patterns] Report unused variable for variables declared in instanceof pattern
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1907 - [Patterns] Withdraw support for record pattern element variable in enhanced for loop 
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/456 - Syntax error on token "RestrictedIdentifierWhen" 


An alert that your work will also come under this comprehensive overhaul. If you wish to be notified, I can CC you on the various tickets.

srikanth-sankaran:
This one is ready to merge - let us move forward please. I am aware that you have had to repeatedly rebase and resolve conflicts - what can look like a thanksless exercise sometimes. Really appreciate your patience @datho7561 !

srikanth-sankaran:
So everybody is on same page: I am listing the next steps for the 4 PRs @datho7561 has been working on:

1. https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1517 - this one - is ready for merge subjecting to https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1517#pullrequestreview-1845277099
2. https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1742 - I have proposed that David and I will jointly own and close this as part of cleanup work going on patterns.
3. https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1816 -  I have proposed that David and I will jointly own and close this as part of cleanup work going on patterns.
4. https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1521 - I can review the parser/grammar changes next week. Other areas we need reviewers. @mpalat please help with delegating so we can divide and conquer

iloveeclipse:
Please rebase **and squash** all intermediate commits to one before merging.

datho7561:
Alright, I've rebased the branch, squashed the commits, addressed the comment about Unicode support (you're right, I didn't need that check), and double checked I have all the snippets from the JEP (excluding those dealing with lists of patterns).

Issue #1526, opened 2023-10-28
 JRTUtil.getJrtSystem(...) permanently consumes a large amount of memory


laeubi:
If one calls `JRTUtil.getJrtSystem(...)` it consumes about 180MB of persistent memory.
Even worse, if one passes the release flag, it consumes another 180MB, so calling the method with multiple release options can easily exhaust the memory of the JVM.

I think the following should be investigated:

1. Check if techniques like _Memory-Mapped File_ could be used to access things instead of holding everything in memory.
2. Check if loading for another release data can be shared with the previously loaded one
3. Check if maybe a Soft or at least Weak References can be used to allow the JVM to clear out some of the data in question

This was discovered here: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/3885#note_1266412 but can easily affect the IDE as well.

laeubi:
Another source of high memory usage seems

```
17:35:37  Exception in thread "Java indexing" java.lang.OutOfMemoryError: Java heap space
17:35:37  	at java.base/java.nio.file.Files.read(Files.java:3238)
17:35:37  	at java.base/java.nio.file.Files.readAllBytes(Files.java:3295)
17:35:37  	at org.eclipse.jdt.internal.compiler.util.JRTUtil.safeReadBytes(JRTUtil.java:323)
17:35:37  	at org.eclipse.jdt.internal.compiler.util.JrtFileSystem.lambda$1(JRTUtil.java:669)
17:35:37  	at org.eclipse.jdt.internal.compiler.util.JrtFileSystem$$Lambda$3646/0x00007f4c44c26000.apply(Unknown Source)
17:35:37  	at java.base/java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1708)
17:35:37  	at org.eclipse.jdt.internal.compiler.util.JrtFileSystem.getFileBytes(JRTUtil.java:667)
17:35:37  	at org.eclipse.jdt.internal.compiler.util.JrtFileSystem.getClassfileContent(JRTUtil.java:648)
17:35:37  	at org.eclipse.jdt.internal.compiler.util.JRTUtil.getClassfileContent(JRTUtil.java:268)
17:35:37  	at org.eclipse.jdt.internal.core.search.indexing.AddJrtToIndex$JrtIndexer.visitFile(AddJrtToIndex.java:143)
[...]

PR #1528, opened 2023-10-28
Only hold soft reference to the JrtFileSystem

laeubi:
Currently all acquired JrtFileSystem are hold in a strong cache and never can be cleared out. As these objects can become quite large this results in a JrtFileSystem only transitively used can fill up the memory.

This changes the cache in a way that it only holds a very small reference object that internally holds the big object in a soft cache so the JVM can clear it out when under memory pressure.

See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1526

@iloveeclipse @stephan-herrmann please review as this currently creates large memory consumption on the build-nodes due to many different JrtFileSystem queried. This also slightly improves error handling as exception do no longer need to be wrapped/unwrapped.

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
@@ -20,6 +20,7 @@ import java.io.FileNotFoundException;
 import java.io.FileReader;
 import java.io.IOException;
 import java.io.InputStream;
+import java.lang.ref.SoftReference;
 import java.net.URI;
 import java.nio.channels.ClosedByInterruptException;
 import java.nio.file.DirectoryStream;
@@ -70,7 +71,7 @@ public class JRTUtil {
 	public static final int NOTIFY_ALL = NOTIFY_FILES | NOTIFY_PACKAGES | NOTIFY_MODULES;
 
 	// TODO: Java 9 Think about clearing the cache too.
-	private static Map<String, JrtFileSystem> images = new ConcurrentHashMap<>();
+	private static Map<String, JrtFileSystemReference> images = new ConcurrentHashMap<>();
 
 	/**
 	 * Map from JDK home path to ct.sym file (located in /lib in the JDK)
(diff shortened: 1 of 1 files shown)


laeubi:
I think instead of a `SoftReference` maybe a `WeakReference` can be used here instead as it seems to be called multiple times. Beside that it seems the most space is consumed by the classfile cache, I'll provide another PR to address that.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
```diff
@@ -147,21 +148,9 @@ public static JrtFileSystem getJrtSystem(File image, String release) throws IOEx
...
-					// a dummy JrtFileSystem is not created.
-					String errorMessage = "Error: failed to create JrtFileSystem from " + image; //$NON-NLS-1$
-					throw new RuntimeIOException(errorMessage, e);
-				}
-			});
-			return system;
-		} catch (RuntimeIOException e) {
-				throw e.getCause();
-		}
+		JrtFileSystemReference reference = images.computeIfAbsent(key, x -> new JrtFileSystemReference(jdk,release));
```
szarnekow, line 151: I think the key should be made from the `Jdk` itself since a normalized file name (toJdkHome) is stored in there.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
```diff
@@ -70,7 +71,7 @@ public class JRTUtil {
 	public static final int NOTIFY_ALL = NOTIFY_FILES | NOTIFY_PACKAGES | NOTIFY_MODULES;
 
 	// TODO: Java 9 Think about clearing the cache too.
```
szarnekow, line 73: This comment is obsolete if a SoftReference is used. We could even consider to use a smarter cache with a time-based LRU eviction policy.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
```diff
@@ -348,6 +337,31 @@ public static String getJdkRelease(File image) throws IOException {
 	}
 }
 
+class JrtFileSystemReference {
+
+	private Jdk jdk;
```
szarnekow, line 342: Nit: please mark the first two fields final

szarnekow:
> I think instead of a `SoftReference` maybe a `WeakReference` can be used here

I don't think that'll fly. The file system does not seem to be referenced from the outside of JrtUtil and is merely a mechanism to cache the modules. If we'd use a WeakReference, it'll be discarded almost immediately.

laeubi:
With my two recent PRs merged I think we can close this for now and see how it behaves.

- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1534
- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1532

I agree that a smarter caching would still be helpful, its a bit sad that JDK does not offer anything out of the box (LinkeHashMap has some LRU support but is not thread safe) also regarding how "important" a reference is, e.g I wish one could give a hint to the GC if it should clear out:
- always
- only if nothing else helps
- based on a LRU principle
- ...

PR #1529, opened 2023-10-29
Add some API filters

fedejeanne:
Get rid of some unnecessary warnings saying that some classes illegally use or instantiate `InvalidInputException` and `CategorizedProblem`, the classes are all part of the same plugin group so it's safe to assume that they are allowed to do it.


fedejeanne:
There's been no activity on this one. Could someone please review it?

jukzi:
Which warnings  are supposed to reduced by this change? screenshot please.
Could those warnings also be reduced by declaring x-friends instead?

fedejeanne:
> Which warnings are supposed to reduced by this change? screenshot please. 

At this point only this one is left:

image

All the others have been removed by https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1545

> Could those warnings also be reduced by declaring x-friends instead?

Should they? It's not a problem of visibility/access, it's the fact that they implement an interface marked with `@noimplement`. In this particular case, both classes are in the same plugin so I don't think `x-friends` is even correct.

fedejeanne:
Solved by a8bd1f815fe822ec591fc6af1090f2e145e910cd

PR #1532, opened 2023-10-29
Use the computed JDK home as the cache key

laeubi:
The Jdk performs internal normalization of a passed in image but the cache uses the unnormalized value as a key. This can result in the same Jrt indexed multiple times.

This changes the key to always use the normalized value to prevent duplicate caching.

@iloveeclipse @stephan-herrmann this seems as a trivial improvement towards
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1526

what currently can even have things cached multiple times.


iloveeclipse:
Thanks.

PR #1534, opened 2023-10-29
Only soft reference class bytes and share them across releases

laeubi:
Currently each `JrtFileSystem` has its own cache for class files, this produces duplicate items to be loaded (and retained) even if only the release flag is changing. Even worse, the cache is hold via strong reference what means the space can afterwards never reclaimed again, what easily can pile up to a few gigabyte of data.

This introduces a global (soft) cache for class bytes read by `JrtFileSystem` that can be shared across releases and allows cache eviction if the JVM is under high memory pressure.

FYI @iloveeclipse this should fix https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1526 or mitigate most of the concerns identified there and result in a much lower memory footprint both in IDE and Tycho in combination with API tools.

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/SoftClassCache.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/SoftClassCache.java
@@ -1,0 +1,86 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Christoph Läubrich.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Christoph Läubrich - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.jdt.internal.compiler.util;
+
+import java.io.IOException;
+import java.lang.ref.SoftReference;
+import java.nio.file.Path;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+/**
+ * Implements a soft cache for reading class files from disk, as these caches can grow quite large but data can be
+ * recovered afterwards we only hold a soft reference to the bytes itself.
+ */
+class SoftClassCache {
+
+	private final ConcurrentMap<String, JdkClasses> jdks = new ConcurrentHashMap<>();
+
+	void clear() {
+		this.jdks.clear();
+	}
+
+	public byte[] getClassBytes(Jdk jdk, Path path) throws IOException {
+		return this.jdks.computeIfAbsent(jdk.path, JdkClasses::new).get(path);
+	}
+
+	private static final class JdkClasses {
+		private final ConcurrentMap<Path, ClassBytes> classes = new ConcurrentHashMap<>(10007);
+		private final String jdkPath;
+
+		public JdkClasses(String jdkPath) {
+			this.jdkPath = jdkPath;
+		}
+
+		public byte[] get(Path path) throws IOException {
+			return this.classes.computeIfAbsent(path, ClassBytes::new).getBytes();
+		}
+
+		@Override
+		public String toString() {
+			return "Class Cache for " + this.jdkPath; //$NON-NLS-1$
+		}
+	}
+
+	private static final class ClassBytes {
+		private final Path path;
+		private volatile boolean empty;
+		private volatile SoftReference<byte[]> bytes;
+
+		public ClassBytes(Path path) {
+			this.path = path;
+		}
+
+		public byte[] getBytes() throws IOException {
+			if (this.empty) {
+				return null;
+			}
+			SoftReference<byte[]> reference = this.bytes;
+			if (reference != null) {
+				byte[] bs = reference.get();
+				if (bs != null) {
+					return bs;
+				}
+			}
+			byte[] readBytes = JRTUtil.safeReadBytes(this.path);
+			if (readBytes == null) {
+				this.empty = true;
+				return null;
+			}
+			this.bytes = new SoftReference<byte[]>(readBytes);
+			return readBytes;
+		}
+	}
+
+}
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
@@ -40,7 +40,6 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Optional;
 import java.util.Properties;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Predicate;
@@ -76,6 +75,8 @@ public class JRTUtil {
 	 * Map from JDK home path to ct.sym file (located in /lib in the JDK)
 	 */
 	private static final Map<Path, CtSym> ctSymFiles = new ConcurrentHashMap<>();
+
+	static final SoftClassCache classCache = new SoftClassCache();
 
 	public interface JrtFileVisitor<T> {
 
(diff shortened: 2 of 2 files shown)


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/SoftClassCache.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+/**
+ * Implements a soft cache for reading class files from disk, as these caches can grow quite large but data can be
+ * recovered afterwards we only hold a soft reference to the bytes itself.
+ */
+class SoftClassCache {
+
+	private ConcurrentMap<String, JdkClasses> jdks = new ConcurrentHashMap<>(2 ^ 4);
```
iloveeclipse, line 28: `2 ^ 4` == 6. Did you meant `2 << 4` (32)? Why not keep default (16), as it was before?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/SoftClassCache.java
```diff
@@ -0,0 +1,85 @@
...
+	void clear() {
+		this.jdks.clear();
+	}
+
+	public byte[] getClassBytes(Jdk jdk, Path path) throws IOException {
+		return this.jdks.computeIfAbsent(jdk.path, JdkClasses::new).get(path);
+	}
+
+	private static final class JdkClasses {
+		private ConcurrentMap<Path, ClassBytes> classes = new ConcurrentHashMap<>(2 ^ 15);
```
iloveeclipse, line 39: `2 ^ 15` == 13. Did you meant `2 << 15` (65536)? Why not just keep as it was (10007)?

I would prefer not to address different issues in a single commit. If you've observed issues with performance due different cache size, please report an issue.

iloveeclipse (changes requested):

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/SoftClassCache.java
laeubi: > 2 ^ 15 == 13. Did you meant 2 << 15 (65536)? Why not just keep as it was (10007)?

2 ^ 15 = 32768 and I choose that because in the debugger I see that at most there are about 25.000 classes cached, and I have not found any explanation why `10007` should be a good default, even if the JVM might be smart enough it seem to not very well align on usual byte boundaries that are multiple of 2 ...

So currently the cache grows from 10007 > 20014 > 40028 with the internal data structure, now it currently never resizes, but again here compared to the size of the bytes from the classes it should not matter much.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/SoftClassCache.java
laeubi: > 2 ^ 4 == 6.

2^4 == 16 what seems a reasonable default for the number of different JDKs a user might use before it is resized, if you like we can use the JDK default as well I think it does not matter much here...

laeubi:
I have now reverted to the previous defaults.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/SoftClassCache.java
szarnekow: @laeubi  for future reference:
```
jshell> 2^4
$1 ==> 6

jshell> 2 ^15
$2 ==> 13

jshell> 2 << 4
$3 ==> 32

jshell> 2 << 15
$4 ==> 65536
```

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/SoftClassCache.java
```diff
@@ -0,0 +1,85 @@
...
+			return "Class Cache for " + this.jdkPath; //$NON-NLS-1$
+		}
+	}
+
+	private static final class ClassBytes {
+		private final Path path;
+		private boolean empty;
+		private SoftReference<byte[]> bytes;
+
+		private ClassBytes(Path path) {
```
iloveeclipse, line 61: Please remove `private` from constructor to avoid extra synthetic method that need to be generated to create instance of ClassBytes in JdkClasses

iloveeclipse:
Also please rebase next time on master.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/SoftClassCache.java
laeubi: Should it be package or public?

iloveeclipse: I would make it public, to be consistent with other class. But actually this doesn't matter here, so choose what you want except private

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/SoftClassCache.java
```diff
@@ -0,0 +1,85 @@
...
+	private static final class ClassBytes {
+		private final Path path;
+		private boolean empty;
+		private SoftReference<byte[]> bytes;
+
+		public ClassBytes(Path path) {
+			this.path = path;
+		}
+
+		public synchronized byte[] getBytes() throws IOException {
```
iloveeclipse, line 65: Please remove `synchronized` here, as the data we are supposed to read shouldn't change, we don't care if that is read only once or may be two times in parallel **initially** - for the cost of **permanent** locking on data retrieval. Note, the original code wasn't synchronized for same reason.

iloveeclipse (changes requested):
LGTM for me, except synchronized comment below.

@szarnekow : would be nice if you could take a look at this PR too.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/SoftClassCache.java
laeubi: The previous code *was* synchronized by the concurrent hash map therefore I found it might be a good idea here as well. if that's not desired I can change it but wanted to note it here just in case...

iloveeclipse: > I can change it but wanted to note it here just in case...

Thanks, noted, please remove synchronized.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/SoftClassCache.java
```diff
@@ -0,0 +1,85 @@
...
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+/**
+ * Implements a soft cache for reading class files from disk, as these caches can grow quite large but data can be
+ * recovered afterwards we only hold a soft reference to the bytes itself.
+ */
+class SoftClassCache {
+
+	private ConcurrentMap<String, JdkClasses> jdks = new ConcurrentHashMap<>();
```
szarnekow, line 28: Small remark: I'd mark this field final. Same for the fields in JdkClasses

szarnekow:
Looks all good and sane to me. IIUIC deduplicating the cached file system instances is part of another MR?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/SoftClassCache.java
laeubi: Done.

laeubi:
> IIUIC deduplicating the cached file system instances is part of another MR?

yes as this is a bit of an independent issue.

szarnekow (approved):
Nice find!

iloveeclipse (approved):

iloveeclipse:
Thanks.

PR #1539, opened 2023-10-30
Eliminate need to update JavadocTagConstants when no new tags introduced.

### LINKED ISSUE #1538: Pain point on Javadoc test suite when new java versions are added
https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1523

When a new java version is added, jdt devs must go add an empty array to a static model. This should not be required if at all possible. Only when new features or tags are added should this step be required.

robstryker:
- Fixes #1538 

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/JavadocTagConstants.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/JavadocTagConstants.java
@@ -204,7 +204,7 @@ public interface JavadocTagConstants {
 			ret[i] = new char[][] {};
 		}
 		for( int i = 0; i < input.length; i++ ) {
-			int nextLevel = input[i].level - ClassFileConstants.MAJOR_VERSION_0;;
+			int nextLevel = input[i].level - ClassFileConstants.MAJOR_VERSION_0;
 			ret[nextLevel] = input[i].tags;
 		}
 		return ret;
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/JavadocTagConstants.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/JavadocTagConstants.java
@@ -183,32 +183,19 @@ public interface JavadocTagConstants {
 	public final static char[] HREF_TAG = {'h', 'r', 'e', 'f'};
 
 	/**
-	 * A class representing additional tags to be made available at specific
-	 * java version levels.
+	 * A record class representing additional tags to be made
+	 * available at specific java version levels.
 	 *
-	 * This class is intended to minimize the necessary metadata updates
-	 * that are currently required when adding new versions of Java.
+	 * This class is intended to prevent maintainers from having to add
+	 * empty arrays for each new java version that is released.
 	 */
-	public static class LevelTags {
-		int level;
-		char[][] tags;
-		/**
-		 * Constructor representing a level paired with new tags
-		 * @param level The level of the class version
-		 * @param tags  The tags to be added at that level
-		 */
-		public LevelTags(int level, char[][] tags) {
-			this.level = level;
-			this.tags = tags;
-		}
-	}
-
+	record LevelTags(int level, char[][] tags) { }
 	/**
 	 * Convert an array of LevelTags into the char[][][] structure that is currently used to
 	 * discover the additional tags available at a given java version.
 	 *
-	 * @param input A LevelTags array representing tag additions for each java version
-	 * @return a char[][][] array representing newly available tags for each java version
+	 * @param input A LevelTags array representing additional tags available for each java version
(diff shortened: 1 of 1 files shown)


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/JavadocTagConstants.java
```diff
@@ -180,101 +181,54 @@ public interface JavadocTagConstants {
 
 	// href tag
 	public final static char[] HREF_TAG = {'h', 'r', 'e', 'f'};
+
+	public static class LevelTags {
```
iloveeclipse, line 185: A javadoc explaining
1) purpose
2) structure
3) update policy

is needed here.
The old code didn't provided that, but this shouldn't be an excuse.

iloveeclipse (changes requested):

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/JavadocTagConstants.java
iloveeclipse: Nothing to sorry. Just imagine someone would need to update this code in few years to make Java 42 compatible.

Write docs/instructions for this case, assuming it would be not you doing the change. That's all what I expect to be added. 

Note, I also have no clue about this code, and wondered few times before what all these arrays are supposed to do.

robstryker:
So I'm not sure if my comments in this PR are clear enough.  The explanation for what this code does is:

I added a new lower-level model that looks like below. It's a list of only new tags to be added or be made available at each java level, and only when new additions actually exist:
```
[
   { javaVersion1, [tag1 tag2]},
   { javaVersion3, [tag3 tag4]},
]
```

We then use this to generate the (before-this-PR) version of the model, which is a char[][][] and includes a list of new tags to be made available at each and EVERY level. As shown below. You'll notice that version 2 has an empty array, whereas the above code does not. 
```
[
   // version 1
   [tag1, tag2],
   // version 2
   [],
   // version 3
   [tag3, tag4],
]
```

This is (later) changed by consumers to look like what's below, a list of tags that are actually available at each level. This is basically accomplished by adding any new tags at the current level to the list of tags available at the previous java level.  I have not touched any of the consumers that generate this code because it looks messier than I'd like to touch. But this is what the end goal looks like: 
```
[
   // version 1
   [tag1, tag2],
   // version 2
   [tag1, tag2],
   // version 3
   [tag1, tag2, tag3, tag4],
]
```

As you can see above, the goal is that tag1 and tag2 are made available at java 1, are still available at java 2, and then the new tags made available at level 3 are appended, leading to all 4 tags being made available at java 4. 

I'm not sure if I should go this verbose in the code / javadoc, or not. If you think I should, I will.

akurtakov:
Is there smth still blocking this one?

rgrunber:
Change seems reasonable. Resulting block / inline tag 3d-arrays seem identical. As for updating this, I think the process seems fairly intuitive, and there's more documentation now than before.

Also, if I had to change this I'd probably trust previous commits to steer me in the right direction :

- https://github.com/eclipse-jdt/eclipse.jdt.core/commit/c38f7ec99f890cc544604535ff82d16992ad90c9#diff-701a12ba0a6f37ed1961246f469e33b9e097c0bc1d8f0fb186a6f91de360bec0 ( no longer a change to be made) which will likely be the majority of cases.
- https://github.com/eclipse-jdt/eclipse.jdt.core/commit/e0901a379870329bd90648ee6ea1637a862a5467#diff-21affdf2d42d5a4f604ea9893f8c5bf84a7fe7e12c4d86e5065b1bca6133aa12 .

So in the worst case, once (might be some time) a new Java version introduces a new block/inline tag, and the first contribution using the new format is done, it really should be smooth going forward.

iloveeclipse:
> Is there smth still blocking this one?

I've initially requested the change, but I honestly have no free time to review the result. 
So a review from the committer is needed.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/JavadocTagConstants.java
```diff
@@ -180,101 +181,69 @@ public interface JavadocTagConstants {
...
+	 * @return a char[][][] array representing newly available tags for all java class version
+	 */
+	public static char[][][] levelTagsToChar3d(LevelTags[] input) {
+		int expectedLength = ClassFileConstants.MAJOR_LATEST_VERSION - ClassFileConstants.MAJOR_VERSION_0 + 1;
+		char[][][] ret = new char[expectedLength][][];
+		for( int i = 0; i < ret.length; i++ ) {
+			ret[i] = new char[][] {};
+		}
+		for( int i = 0; i < input.length; i++ ) {
+			int nextLevel = input[i].level - ClassFileConstants.MAJOR_VERSION_0;;
```
rgrunber, line 207: Unnecessary semicolon at end here.

PR #1544, opened 2023-11-02
TypeDeclarationPattern.allowModuleRegex: remove dead code

jukzi:


jarthana:
I see the [original bug](https://bugs.eclipse.org/bugs/show_bug.cgi?id=519151) that introduced this code is still open. So, I guess we can just remove the dead code and let that one take care of it if and when required. +1 for the patch. Please go ahead and release this.

@mpalat  I see that the above mentioned bug is assigned to you. So, you might want to take a look at this PR too when you can.

jarthana (approved):

PR #1545, opened 2023-11-02
InvalidInputException: fix "illegally instantiates" warnings

jukzi:
InvalidInputException is marked @noinstantiate and should only be constructed by Scannner

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/IntersectionTypeBinding18.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/IntersectionTypeBinding18.java
@@ -25,7 +25,6 @@ package org.eclipse.jdt.internal.compiler.lookup;
 
 import java.util.Set;
 
-import org.eclipse.jdt.core.compiler.InvalidInputException;
 import org.eclipse.jdt.internal.compiler.ast.Wildcard;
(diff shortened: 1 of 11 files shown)


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/core/compiler/InvalidInputException.java
```diff
@@ -28,7 +28,7 @@ public class InvalidInputException extends Exception {
  * Creates a new exception with no detail message.
  */
 public InvalidInputException() {
-	super();
+	org.eclipse.jdt.internal.compiler.parser.Scanner.invalidInput();
```
iloveeclipse, line 31: Not sure I understood why `Scanner.invalidInput()` is added here?

jukzi: thanks. left over from refactoring

iloveeclipse:
> InvalidInputException is marked @noinstantiate and should only be constructed by Scannner

I believe this is a misunderstanding. The code was originally in the jdt.core bundle, and `@noinstantiate` was for the "all other bundles" outside JDT core. Now the code moved from jdt core to batch compiler, jdt core itself became "other" client (which is not the case). Instead of this refactoring probably better to silence the warning in jdt code.

jukzi:
> Instead of this refactoring probably better to silence the warning in jdt code.

I prefer solutions (here: moving back the code to jdt core) over disabling warnings, which could be helpful in other circumstances

PR #1548, opened 2023-11-02
jdt.apt: Suppress ecj "Discouraged access" warnings of jdt.core

jukzi:


stephan-herrmann:
Shouldn't we regulate intra-jdt dependencies using `x-friends`, rather than `@SuppressWarnings`?

I'd like to use `@SuppressWarnings` only as a last resort, if no reasonable option exist to *avoid* the warning.

Secondly, you are giving a carte blanche to violate *any* access rules within the marked classes, i.e., you are granting more permissions than desired.

PR #1551, opened 2023-11-02
DefaultCodeFormatterConstants#getJavaConventionsSettings: use Generics

jukzi:
so that caller needs no typecast


jukzi:
needed for https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/894

PR #1564, opened 2023-11-06
TestVerifier: never wait 100ms

jukzi:
a try to improve test time


jukzi:
This build's junit time was 26min instead of 32min of master
master:
image
this PR:
image

time was reduced for org.eclipse.jdt.core.tests.compiler.regression

PR #1567, opened 2023-11-07
Suggest 'var' type variables as auto-completion method arguments. Fix #228

### LINKED ISSUE #228: PR #4173 needs a regression test
Edit: The problem described below is addressed by https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4173 and this ticket is repurposed to follow up with a regression test for the problem which is not included in that PR.

// -------------------------------------- original problem report below ---------------------------------------

Using this code example

	void d() {
		int i2 = 3;
		var i3 = 3;
		Integer.valueOf
	}

Pressing ctrl+space when the caret is after 'Integer.valueOf' will bring up a menu for auto-completing the method. Select the 'int' flavor. Eclipse will add the code so that it's now

Integer.valueOf(i)

with a popup menu to select from proposals for the argument of the method. It will show a generic 'i', the default '0', and only 'i2' out of the 2 local variables. 'i3' is not shown because it is declared as 'var' even though the correct 'int' type is inferred. Changing the declaration to

int i3 = 3;

and repeating the process will indeed also show 'i3'.

'var' declaration should also be added to the arguments suggestions menu.

hotspotoq:
## What it does

- Fixes #228

#### Uses existing logic which has been commented out.
The condition `local.declaration.initialization != null` matches with variables being initialized but also matches var declarations. Then var variables get out of completion candidates. Fortunately, there was [existing logic in code's comments](https://github.com/eclipse-jdt/eclipse.jdt.core/commit/cb53eff5db7d4f76eb24c91a2abecee1441f0bbc#diff-0e781fe5c69959e9da651b4c13022062c34e898f147a412c5cf6a8807a3f7f90R752-R758) that can deal with this problem.

This simple changes also fix this issue but are not perfect here. (The problem is that var variables being initialized will appear in proposals, but can this change be better choice?)
[78adf91f002a1a9b7d1838c24de08c2d5f608e25](https://github.com/eclipse-jdt/eclipse.jdt.core/commit/78adf91f002a1a9b7d1838c24de08c2d5f608e25#diff-8469e0f05a7f350e7047101a3e20fe0bf4d0f65df01fa0a18eb6378128c5de5aR12338)

#### Provides resolved types of var declarations.
When calculating auto-boxing scores, var is considered qualified class type whose name is var. Without the change, primitive type variables declared with var get bad scores; the auto-completion suggests 0 of literal instead of var variables. This change is same as: [3b1eade6257846b964d26359bdcfbc40d6c4bec4](https://github.com/eclipse-jdt/eclipse.jdt.core/commit/3b1eade6257846b964d26359bdcfbc40d6c4bec4#diff-2a561a84bd9c4008e33ec162133f1f0734636eada70d3367792b70a12d4d7b98R470-R483)

#### Uses the method with null checks and adds an argument.
I believe it should have null checks. `local.declaringScope` is used to tell whether var keyword is available at the version.

## How to test

[eclipse.jdt.ui/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/contentassist/MethodParameterGuessingCompletionTest10.java](https://github.com/eclipse-jdt/eclipse.jdt.core/files/13282131/MethodParameterGuessingCompletionTest10.java.txt)
[eclipse.jdt.ui/org.eclipse.jdt.text.tests/src/org/eclipse/jdt/text/tests/contentassist/CompletionTestSetup10.java](https://github.com/eclipse-jdt/eclipse.jdt.core/files/13282132/CompletionTestSetup10.java.txt)
These tests will have to be situated in eclipse.jdt.ui, what can I do?

## Author checklist

I'm open to changing expressions since I'm not good at English.

### DIFF
--- a/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
+++ b/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
@@ -13220,7 +13220,7 @@ public final class CompletionEngine
 								continue next;
 
 							// https://bugs.eclipse.org/bugs/show_bug.cgi?id=328674
-							if (local.declaration.initialization != null && !local.declaration.type.isTypeNameVar(null)) {
+							if (local.declaration.initialization != null && !local.declaration.isTypeNameVar(local.declaringScope)) {
 								// proposal being asked inside field's initialization. Don't propose this field.
 								continue next;
 							}
--- a/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/InternalExtendedCompletionContext.java
+++ b/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/InternalExtendedCompletionContext.java
@@ -251,6 +251,13 @@ public class InternalExtendedCompletionContext {
 		}
 		if (parent == null) return null;
 
+		String typeSig;
+ 		if (local.type == null || local.type.isTypeNameVar(binding.declaringScope)) {
+ 			typeSig = Signature.createTypeSignature(binding.type.signableName(), true);
+ 		} else {
+ 			typeSig = Util.typeSignature(local.type);
+ 		}
+
 		return new LocalVariable(
 				parent,
 				DeduplicationUtil.toString(local.name),
@@ -258,7 +265,7 @@ public class InternalExtendedCompletionContext {
 				local.declarationSourceEnd,
 				local.sourceStart,
 				local.sourceEnd,
-				local.type == null ? Signature.createTypeSignature(binding.type.signableName(), true) : Util.typeSignature(local.type),
+				typeSig,
 				binding.declaration.annotations,
 				local.modifiers,
 				local.getKind() == AbstractVariableDeclaration.PARAMETER);
--- a/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/InternalExtendedCompletionContext.java
+++ b/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/InternalExtendedCompletionContext.java
@@ -742,7 +742,10 @@ public class InternalExtendedCompletionContext {
 						// then don't propose the local variable
 						if (local.declaration.initialization != null) {
 							/*(use this if-else block if it is found that local.declaration.initialization != null is not sufficient to
-							  guarantee that proposal is being asked inside a local variable declaration's initializer)
+							  guarantee that proposal is being asked inside a local variable declaration's initializer)*/
+
+							// use this to avoid matching with var declarations.
+							// https://github.com/eclipse-jdt/eclipse.jdt.core/issues/228
 							 if(local.declaration.initialization.sourceEnd > 0) {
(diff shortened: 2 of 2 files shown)


hotspotoq:
I forgot updating header years but I don't know how to add extra commits...

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
```diff
@@ -13220,7 +13220,7 @@ private void findVariablesAndMethods(
 								continue next;
 
 							// https://bugs.eclipse.org/bugs/show_bug.cgi?id=328674
-							if (local.declaration.initialization != null && !local.declaration.type.isTypeNameVar(null)) {
+							if (local.declaration.initialization != null && !local.declaration.isTypeNameVar(local.declaringScope)) {
```
hotspotoq, line 13223:
```suggestion
if (local.declaration.initialization != null && !local.declaration.type.isTypeNameVar(null)) {
```

srikanth-sankaran:
@hotspotoq - This PR is missing a regression test but otherwise looks to be solidly on the right track with respect to diagnosis and remedy - Why did you close the PR without merging ??

I am reviving it in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4173 - FYI

hotspotoq:
I’ve closed this because it’s been quite a while, and I’m no longer confident that the fix is appropriate.

> Do not just on account of the initialization not being nulled out, conclude that completion is happening inside the initialized and excluded the declared variable. Use co-ordinates to be sure.

Also, the comment you added to the source was something I originally had to clarify when submitting the PR.
Thank you for your understanding and time.

> This PR is missing a regression test ...

> How to test
...
These tests will have to be situated in eclipse.jdt.ui, what can I do?

I think these tests would be unit tests, but does "regression test" mean a specific unit test or the Jenkins test? For my future reference, could you please tell me that? I’m sorry, I’m a beginner.

srikanth-sankaran:
> I’m no longer confident that the fix is appropriate.

Your diagnosis and fix were indeed good!

> > How to test
> > ...
> > These tests will have to be situated in eclipse.jdt.ui, what can I do?

No, you write tests to work in a "headless" mode - see PR https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4181 for an example

> I think these tests would be unit tests, but does "regression test" mean a specific unit test or the Jenkins test? For my future reference, could you please tell me that? I’m sorry, I’m a beginner.

> but does "regression test" mean a specific unit test

Basically a test that establishes that with the fix things are green and without red. As does https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4181



> I’m a beginner.

Good effort 👍

srikanth-sankaran:
As mentioned in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4173#issuecomment-3048302763 I meant to include you as the co-author in the commit message, but forgot. Sorry about that!

hotspotoq:
> No, you write tests to work in a "headless" mode - see PR https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4181 for an example

I understand! Thank you.

> I meant to include you as the co-author in the commit message, but forgot.

No problem! Thank you for moving this forward.

PR #1569, opened 2023-11-07
Null analysis - Record synthetic accessor methods cannot implement interfaces with null constraints

### LINKED ISSUE #1302: Null analysis - Record synthetic accessor methods cannot implement interfaces with null constraints
@stephan-herrmann 

# Behavior

Assume we have `@NonNullByDefault` on a package. The package contains:

```java
public interface Parent {
  Map<String, String> model();
}
```

*Now a separate file (this important):*

```java
public record Child(Map<String, String> model) implements Parent {
}
```

We get an error on `Child`:

```
The method @NonNull Map<String,String> model() from Child cannot implement the corresponding method from Parent due to incompatible nullness constraints
```

# Work around

```java
public record Child(Map<String, String> model) implements Parent {
  Map<String,String> model() { return this.model }
}
```

**The reason this was probably not found earlier is if the interface and record are in the same file this issue disappears.**

For example this is fine:

```java
public interface Parent {
  Map<String, String> model();
  public record Child(Map<String, String> model) implements Parent {
  }
}
```

stephan-herrmann:
fixes #1302

## What it does
Revert one change from https://bugs.eclipse.org/bugs/show_bug.cgi?id=563181 to ensure that TYPE_USE annotations on a type of a record component are verbatim seen on accessors.

Modify a test from https://bugs.eclipse.org/bugs/show_bug.cgi?id=571919 where absence of such TYPE_USE was extected.


stephan-herrmann:
@jarthana  I'm afraid I need your help to see if this change breaks your prior art from the mentioned bugs.

In all linked specification I found a lot about annotations *on a record component* and *on a record accessor*, but TYPE_USE annotations (possibly on nested type arguments etc) should IMHO not count as annotations on the accessor. Even if they would, I see the spec explicitly admitting those in the context of delegating to accessors.

Ergo, I'm questioning the `type.unannotated()` call from the former bug and the expectation (0 annotations) in two test situations of the latter bug. Removing that call resolves #1302 

Yes, I saw that you observed javac 15 to behave differently.  Oops, the very test case in question (`testRecords4a`) has its javac-variant disabled. And yes, enabling it would let it fail *after* my changes of test expectation here. javac-bug, or am I reading the spec wrongly?

stephan-herrmann:
From [JLS 8.10.3](https://docs.oracle.com/javase/specs/jls/se21/html/jls-8.html#jls-8.10.3): 

> If a record class has a record component for which an accessor method is not declared explicitly, then an accessor method for that record component is declared implicitly, with the following properties:
> ...
> * It is annotated with the annotations, if any, that appear on the corresponding record component and whose annotation interfaces are applicable in the method declaration context, **or in type contexts**, or both ([§9.7.4](https://docs.oracle.com/javase/specs/jls/se21/html/jls-9.html#jls-9.7.4)).

(my emphasis)

jarthana (approved):

iloveeclipse:
@stephan-herrmann : we are at RC1 phase now. I assume this PR was safe enough to be merged that late?

stephan-herrmann:
> @stephan-herrmann : we are at RC1 phase now. I assume this PR was safe enough to be merged that late?

I consulted with @jarthana who confirmed that his existing approval is good for RC1.

PR #1570, opened 2023-11-07
Fix selection in for each loop nested in if statement

### LINKED ISSUE #1568: Current text selection cannot be opened in an editor
```java
import java.util.ArrayList;
import java.util.List;

public class RowRenderData {

    private List<Object> cells = new ArrayList<>();

    public List<Object> getCells() {
        return cells;
    }

    public void setCells(List<Object> cells) {
        this.cells = cells;
    }

        List<RowRenderData> rows = new ArrayList<>();
        if (true) {
            for (RowRenderData row : rows) {
                row.getCells();
            }
        }
    }
}
```
image

jjohnstn:
- fixes #1568

## What it does

Fixes selecting a variable inside a for each loop that is nested inside another statement such as if or while.

## How to test

See issue or new test.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/select/SelectionParser.java
+++ b/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/select/SelectionParser.java
@@ -126,6 +126,8 @@ public class SelectionParser extends AssistParser {
 	public static final char[] SUPER = "super".toCharArray(); //$NON-NLS-1$
 	public static final char[] THIS = "this".toCharArray(); //$NON-NLS-1$
 
+	public boolean lastIfHasPattern = false;
+
 public SelectionParser(ProblemReporter problemReporter) {
 	super(problemReporter);
 	this.javadocParser.checkDocComment = true;
@@ -882,8 +884,10 @@ protected void consumeGuard() {
 protected void consumePostExpressionInIf() {
 	super.consumePostExpressionInIf();
 	if (this.expressionStack[this.expressionPtr].containsPatternVariable()) {
+		this.lastIfHasPattern = true;
 		pushOnElementStack(K_INSIDE_THEN, this.expressionPtr, this.astPtr);
 	} else {
+		this.lastIfHasPattern = false;
 		popUntilElement(K_INSIDE_IF);
 		popElement(K_INSIDE_IF);
 	}
@@ -1539,7 +1543,9 @@ protected void consumeToken(int token) {
 				}
 				break;
 			case TokenNameelse:
-				pushOnElementStack(K_INSIDE_ELSE, this.expressionPtr, this.astPtr);
+				if (this.lastIfHasPattern) {
+					pushOnElementStack(K_INSIDE_ELSE, this.expressionPtr, this.astPtr);
+				}
 				break;
 			case TokenNameAND_AND:
 				pushOnElementStack(K_POST_AND_AND, this.expressionPtr);
--- a/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/select/SelectionParser.java
+++ b/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/select/SelectionParser.java
@@ -43,6 +43,7 @@ import org.eclipse.jdt.internal.compiler.ast.EmptyStatement;
 import org.eclipse.jdt.internal.compiler.ast.ExplicitConstructorCall;
 import org.eclipse.jdt.internal.compiler.ast.Expression;
 import org.eclipse.jdt.internal.compiler.ast.FieldReference;
+import org.eclipse.jdt.internal.compiler.ast.ForeachStatement;
 import org.eclipse.jdt.internal.compiler.ast.GuardedPattern;
 import org.eclipse.jdt.internal.compiler.ast.IfStatement;
 import org.eclipse.jdt.internal.compiler.ast.ImportReference;
@@ -103,6 +104,8 @@ public class SelectionParser extends AssistParser {
 	protected static final int K_INSIDE_STATEMENT_SWITCH = SELECTION_PARSER + 11; // whether we are in a switch statement
 	protected static final int K_INSIDE_EXPRESSION_SWITCH = SELECTION_PARSER + 12; // whether we are in an expression switch
 	protected static final int K_INSIDE_WHEN = SELECTION_PARSER + 13; // whether we are in the guard
+
+	protected static final int K_INSIDE_FOR_EACH = SELECTION_OR_ASSIST_PARSER + 14; // whether we are in a for each statement
 
 
 
@@ -213,6 +216,7 @@ private void buildMoreCompletionContext(Expression expression) {
 			case K_INSIDE_ELSE:
 			case K_INSIDE_STATEMENT_SWITCH:
 			case K_INSIDE_EXPRESSION_SWITCH:
+			case K_INSIDE_FOR_EACH:
 				int newAstPtr = (int) this.elementObjectInfoStack[i];
 				int length = this.astPtr - newAstPtr;
 				Statement[] statements = new Statement[length + (orphan != null ? 1 : 0)];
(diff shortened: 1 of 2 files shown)


jjohnstn:
Hi @srikanth-sankaran There was a regression introduced from 8cbe978792d366f6f792053f082b9fa2179056d4.  What happens is that the new logic for if statements will end up consuming statements inside a for each loop which may include a reference to the element variable.  When the user selects the element variable reference and hits F3, the user sees the error: "Current text selection cannot be opened in an editor".

I took a stab at fixing it by adding a new K flag for inside a for each loop so that the statements could be placed into the for each loop and not the outside if/while etc.. statements.  I also added a test but basically matched the output I saw.  In testing, I used the following test.  I am able to select and use F3 on both x variables, the row variable inside the for loop, rows, and j inside the nested if statement and they go to the correct declarations.

If you would have done it differently, feel free to change the patch.

```
import java.util.ArrayList;
import java.util.List;

public class RowRenderData {

	public static Object z;
	
    private List<Object> cells = new ArrayList<>();

    public List<Object> getCells() {
        return cells;
    }

    public void setCells(List<Object> cells) {
        this.cells = cells;
    }

        List<RowRenderData> rows = new ArrayList<>();
        if (z instanceof String x) {
        	x.concat("abc"); //$NON-NLS-1$
            for (RowRenderData row : rows) {
            	int f = rows.size();
                row.getCells();
            }
            if (z instanceof String j) {
            	String s = j;
            }
            x.concat("def"); //$NON-NLS-1$
        }
    }
}
```

srikanth-sankaran:
@jjohnstn , Thanks, I will take a look next week, Ok ?  Right now I am neck deep buried in String templates related work and would prefer not to context switch.

jjohnstn:
@srikanth-sankaran No problem.  I got the tests all working and F3 works in the editor for both element variables and record pattern variables.  As mentioned, you might have a different solution.

iloveeclipse:
@srikanth-sankaran : could you please check if you can review this PR till tomorrow, so it can be merged (or not) for RC1?

srikanth-sankaran:
Yes, I will look at it first thing tomorrow Wednesday. Sorry I was very busy with string templates implementation review for the past week and half - as it involved substantial discussion and rewrite of key parts.

srikanth-sankaran:
This patch has a bunch of problems and it recovers the parse tree incorrectly. The maintenance of the state variable `lastIfHasPattern` is tricky and is not handled correctly - when I (re)implemented selection support for pattern matching, I did consider not stacking if statements (and other control structures) that don't introduce a pattern variable but gave that up in preference for a simpler approach of always stacking because the alternate was a bit messier.

With the proposed patch, given 

```
public class X {
	Integer abcdef = 10;
    public void main(String argv[]) {
        Object o = argv;
        if (!(o instanceof String [] abcdef)) {
        	if (abcdef == null) {
        		
        	}
        } else {
        	if (abcdef.length > 0) { // control clicking on abcdef in this line take you to wrong declaration
        		
        	}
        }
    }
}
```
control clicking on `abcdef` in the else block takes you to the wrong declaration because the recovered parse tree is 

```
public class X {
  Integer abcdef;
  public X() {
  }
  public void main(String[] argv) {
    Object o;
    {
      if (<SelectOnName:abcdef>)
          ;
    }
  }
}
```

srikanth-sankaran:
Here is another problem. In this snippet selecting the indicated text and pressing F3 results in an error: `"current text selection cannot be opened in editor"`

```
public class X {
	

	public void main(String argv[]) {
		Object o = argv;
		if (o instanceof String[] abcdef) {
			if (o != null) {
				if (argv[0] instanceof String str) {
				}
			} else {
				if (abcdef.length > 0) {
					// selecting abcdef in the line above and F3ing results in "current text selection cannot be opened in editor"

				}
			}
		}
	}
}
```

This is because the recovered parse tree is attaching the `else` to the wrong `if`

```
public class X {
  public X() {
  }
  public void main(String[] argv) {
    Object o;
    {
      {
        if ((o instanceof String[] abcdef))
            {
              {
                if ((argv[0] instanceof String str))
                    {
                    }
              }
            }
        else
            {
              if (<SelectOnName:abcdef>)
                  ;
            }
      }
    }
  }
}

```

I would suggest we don't try to optimize the stacking of statements based on whether they contain pattern variables or not - as it is tricky and we can end up reconstructing wrong parse trees rather readily - our test coverage is poor in these areas and don't offer nearly a safety net :-(

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/select/SelectionParser.java
```diff
@@ -104,6 +105,8 @@ public class SelectionParser extends AssistParser {
 	protected static final int K_INSIDE_EXPRESSION_SWITCH = SELECTION_PARSER + 12; // whether we are in an expression switch
 	protected static final int K_INSIDE_WHEN = SELECTION_PARSER + 13; // whether we are in the guard
 
+	protected static final int K_INSIDE_FOR_EACH = SELECTION_OR_ASSIST_PARSER + 14; // whether we are in a for each statement
+
```
srikanth-sankaran, line 109: I think this constant needs to be defined only for the SELECTION_PARSER namespace - cf earlier constants in the same file.

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/select/SelectionParser.java
```diff
@@ -871,7 +883,21 @@ protected void consumeGuard() {
...
+		popUntilElement(K_INSIDE_IF);
+		popElement(K_INSIDE_IF);
+	}
+}
+
+@Override
+protected void consumeEnhancedForStatementHeader() {
+	// TODO Auto-generated method stub
+	super.consumeEnhancedForStatementHeader();
+	pushOnElementStack(K_INSIDE_FOR_EACH, this.expressionPtr, this.astPtr);
```
srikanth-sankaran, line 900: It looks like this delimiter `K_INSIDE_FOR_EACH` is not popped at all ?

srikanth-sankaran (changes requested):

srikanth-sankaran:
> @srikanth-sankaran No problem. I got the tests all working and F3 works in the editor for both element variables and record pattern variables. As mentioned, you might have a different solution.

I will take a stab at it later today.

srikanth-sankaran:
Original problem arises from the incorrect parse tree recovered:

```
public @SuppressWarnings("preview") class X {\n
  public static Object k_;\n
  public int val_;\n
  <clinit>() {\n
  }\n
  public X() {\n
  }\n
  public static void foo(X[] ar_ray) {\n
    {\n
      {\n
        if ((k instanceof String z_))\n
            {\n
              for (X x_ : ar_ray) ;\n   //  <<<<<--------------------------  See the semicolon
              int;\n
              int per;\n
              <SelectOnName:x_>;\n
            }\n
      }\n
    }\n
  }\n
}\n

```

srikanth-sankaran:
> 
> I took a stab at fixing it by adding a new K flag for inside a for each loop so that the statements could be placed into the for >each loop and not the outside if/while etc.. statements. I also added a test but basically matched the output I saw.

As you can see, there are some (1) extra partial statements and (2) the completion node repeats in your tests, but the first one is an artifact of where the recovery triggers - the `int` type node is **actually** there in the astStack as a partial statement, If recovery were delayed until the full local declaration is reduced this problem will go away but that is extra work. Completion node repeating is no aesthetically pleasing but is harmless as we recursively descend the parse tree, the first completion node upon being visited would throw an exception (`CompletionNodeFound`) so the repeated node is harmless. 

Cleaning up these oddities is possible, but not mandatory

srikanth-sankaran:
Alternate PR https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1602 under test - Passes RunModelTests locally.

srikanth-sankaran:
https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1602 suprecedes this one and this can be closed without action

jjohnstn:
Thanks @srikanth-sankaran for looking at it

PR #1579, opened 2023-11-09
[test] remove latestBREE

jukzi:
fixes warning `Execution environment is lower than one of the plug-in's dependencies (org.eclipse.jdt.core.tests.compiler) which has an execution environment of JavaSE-17.`


jukzi:
@jarthana  Whats the purpose of the latestBREE project?  an old java version seem wrong - should be even 20? 21?

akurtakov:
Out of curiosity - what is this used for?

jarthana:
Now that JDT core has moved to Java 17, this is not that important. We can even remove this. Or, developers can use this to try and write code with latest Java language features after moving the bundle to the latest BREE. Or if we think it's not relevant, I am okay with deleting this project.

jukzi:
I prepared a removal of the project, copying additional test methods into the regular TextBlockTest test

Issue #1578, opened 2023-11-08
ecj spams the console with irrelevant zip exceptions


stephan-herrmann:
Caused by [this change](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/239/files#diff-c0b06f7f0bde6fd19dd40c4dfdb8cf81916f21156eae057b0d2f54865fa1f66f)  from #239:

When compiling on JRE 1.8, the compiler spits out tons of exceptions like this:
```
Failed to init Classpath for jar file /somepath/j2sdk/linux/1.8.0_111/jre/lib/fontconfig.SuSE.10.properties.src
java.util.zip.ZipException: zip END header not found
	at
java.base/java.util.zip.ZipFile$Source.findEND(ZipFile.java:1469)
	at
java.base/java.util.zip.ZipFile$Source.initCEN(ZipFile.java:1477)
	at
java.base/java.util.zip.ZipFile$Source.<init>(ZipFile.java:1315)
	at
java.base/java.util.zip.ZipFile$Source.get(ZipFile.java:1277)
	at
java.base/java.util.zip.ZipFile$CleanableResource.<init>(ZipFile.java:709)
	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:243)
	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:172)
	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:186)
	at
org.eclipse.jdt.internal.compiler.batch.ClasspathJar.initialize(ClasspathJar.java:204)
	at
org.eclipse.jdt.internal.compiler.batch.FileSystem.<init>(FileSystem.java:235)
	at
org.eclipse.jdt.internal.compiler.batch.Main.getLibraryAccess(Main.java:3504)
	at
org.eclipse.jdt.internal.compiler.batch.Main.performCompilation(Main.java:4750)
	at
org.eclipse.jdt.internal.compiler.batch.Main.compile(Main.java:1802)
	at
org.eclipse.tycho.compiler.jdt.JDTCompiler.compileInProcess(JDTCompiler.java:370)
```

It does so for every file in that `jre/lib` directory.

I admit, that the strategy to detect zip files is a bit crude, but it looks like no reliable positive specification of zip file names was available when that code was written.

Hence the `ZipException` is perfectly legal in this situation, and silently catching it with an `// ignore` comment was by full intention.

Can we please get rid of this particular console logging?

[...]

PR #1585, opened 2023-11-10
Check if linked classpath entry is actually a jar

laeubi:
Possibly fix https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1578

@stephan-herrmann what do you think?


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathJar.java
```diff
@@ -89,7 +89,7 @@ public List<Classpath> fetchLinkedJars(FileSystem.ClasspathSectionProblemReporte
 				directoryPath = directoryPath.substring(0, lastSeparator + 1); // potentially empty (see bug 214731)
 				while (calledFilesIterator.hasNext()) {
 					File linkedFile = new File(directoryPath + (String) calledFilesIterator.next());
-					if (linkedFile.isFile()) {
+					if (linkedFile.isFile() && linkedFile.getName().toLowerCase().endsWith(".jar")) { //$NON-NLS-1$
```
stephan-herrmann, line 92: If s.o. on this planet every put a jar on their classpath that doesn't have the `.jar` file extension, then this change will be a breaking change for them.
I don't recall JLS requiring that file extension.

Issue #1593, opened 2023-11-13
[21] Missing Java 20, 21 in Parser class


kriegaex:
Here, `else if` branches are missing for Java 20 and 21:

https://github.com/eclipse-jdt/eclipse.jdt.core/blob/ef1e06636f4f6b7832d3c53c0d9df7fe3d741dcb/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java#L249-L255

`ClassFileConstants.JDK20` and `JDK21` already exist, they are simply not being used there.

Sorry for not creating a trivial PR, but I have checked out an AspectJ branch and am in the middle of something. Before I forget it, I am creating this issue for this drive-by find.

PR #1598, opened 2023-11-13
Handle missing JDK version compliance cases

srikanth-sankaran:
## What it does

Handles missing JDK version compliance discrimination.
Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1593

## Author checklist


akurtakov:
Do we have any code that was supposed to test this?

iloveeclipse:
@srikanth-sankaran: I assume this is supposed to fix https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1593.
Please don't forget to add this to the commit message like:

```
Handle missing JDK version compliance cases

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1593
```

so github can properly **automatically** link PR & ticket.

iloveeclipse:
Another question is: is this something we *must* take into 4.30, or just some internal problem that has no effect on downstream consumers?

srikanth-sankaran:
Tests can be hard to construct. Basically, the data computed at `org.eclipse.jdt.internal.compiler.parser.Parser.buildFileForCompliance(String, int, String[])` gets written out to a file by name `parser21.rsc` which is read by `org.eclipse.jdt.internal.compiler.parser.Parser.initTables()` to initialize `org.eclipse.jdt.internal.compiler.parser.Parser.rules_compliance`. 

This array is then consulted by `org.eclipse.jdt.internal.compiler.parser.diagnose.DiagnoseParser.parseCheck(int[], int, int, int)` and `org.eclipse.jdt.internal.compiler.parser.diagnose.DiagnoseParser.scopeTrialCheck(int[], int, PrimaryRepairInfo, int)` to determine if a certain production that appears a candidate to try at a certain point in the parse should even be in the reckoning given the current source level. 

This is hard to reconstruct in a test

srikanth-sankaran:
> Another question is: is this something we **must** take into 4.30, or just some internal problem that has no effect on downstream consumers?

No, this is not must fix. Although it is trivial enough that we may as well fix it right now. Basically when we have a syntax error in the program, we invoke a special parser called `DiagnoseParser` which makes educated guesses about what is syntactically wrong with the program and attempts various repairs to make the program syntactically valid - one such repair could be "insertion of some tokens", or "deletion of some tokens" or "replacement of some tokens with some others" etc. Multiple such repairs may yield valid programs and where there is a choice of repairs  `DiagnoseParser` uses various weightages to pick one.

In attempting these repairs, if a program is being compiled at say source level 19,  `DiagnoseParser` should not consider grammar rules that are specific for JDK levels 20+. If it does, then it could end up recommending a repair strategy that is not valid for source level 19: say by spitting out an error message like "Add } to close the embedded expression in the string template"

So the rules_compliance array is used to prune rules unsuitable for the current source level. 

I say it is not a must fix because the probability of such a recommendation surfacing is pretty low.

iloveeclipse:
Yet another question: if I see it right, the change here will have no (visible) effect until one regenerates `*.rsc` files via `GenerateParserScript`? Should this PR then contain regenerated files too - otherwise we have a mismatch between what we've supposed to generate and what we are using?

srikanth-sankaran:
> Yet another question: if I see it right, the change here will have no (visible) effect until one regenerates `*.rsc` files via `GenerateParserScript`? Should this PR then contain regenerated files too - otherwise we have a mismatch between what we've supposed to generate and what we are using?

Grrrr. That logically follows from what I outlined as the use model for `org.eclipse.jdt.internal.compiler.parser.Parser.rules_compliance` two comments above - but I overlooked that very same point.

Thanks for raising the question. I'll follow up

jarthana (approved):

iloveeclipse:
Note: there are four new test fails, see
[org.eclipse.jdt.core.tests.compiler.regression.InstanceofPrimaryPatternTest.test002 - 17](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-1598/3/testReport/junit/org.eclipse.jdt.core.tests.compiler.regression/InstanceofPrimaryPatternTest/test002___17/)
[org.eclipse.jdt.core.tests.compiler.regression.InstanceofPrimaryPatternTest.test003 - 17](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-1598/3/testReport/junit/org.eclipse.jdt.core.tests.compiler.regression/InstanceofPrimaryPatternTest/test003___17/)
[org.eclipse.jdt.core.tests.compiler.regression.InstanceofPrimaryPatternTest.test002 - 19](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-1598/3/testReport/junit/org.eclipse.jdt.core.tests.compiler.regression/InstanceofPrimaryPatternTest/test002___19/)
[org.eclipse.jdt.core.tests.compiler.regression.InstanceofPrimaryPatternTest.test003 - 19](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-1598/3/testReport/junit/org.eclipse.jdt.core.tests.compiler.regression/InstanceofPrimaryPatternTest/test003___19/)

iloveeclipse:
@srikanth-sankaran, @jarthana : are these only **test related** fails based on wrong test assumptions, or do they also affect "real life" compiler behavior? If later: is this something we **must** fix & release for 4.30?

srikanth-sankaran:
> @srikanth-sankaran, @jarthana : are these only **test related** fails based on wrong test assumptions, or do they also affect "real life" compiler behavior? If later: is this something we **must** fix & release for 4.30?

These are "failures" in the sense that the expected and actual diagnostics differ neither being invalid.

Let us start with `org.eclipse.jdt.core.tests.compiler.regression.InstanceofPrimaryPatternTest.test002()` 

The construct `if (obj instanceof (String s))` has a syntax error - the `'` surrounding `String s` is illegal.

Earlier on the compiler (actually `DiagnoseParser` issued the message:

```
1. ERROR in X.java (at line 3)\n
 if (obj instanceof (String s)) {\n
         ^^^^^^^^^^\n
Syntax error on token "instanceof", ReferenceType expected after this token\n
```

Now it does
```
1. ERROR in X.java (at line 3)\n
 if (obj instanceof (String s)) {\n
                    ^\n
Syntax error on token "(", delete this token\n
----------\n
2. ERROR in X.java (at line 3)\n
 if (obj instanceof (String s)) {\n
                              ^\n
Syntax error on token ")", delete this token\n
----------\n
```

both are valid repairs, but I would say the present new one is better.

Moving onto `org.eclipse.jdt.core.tests.compiler.regression.InstanceofPrimaryPatternTest.test003()`  this is more of a complicated case.

The construct `if (obj instanceof ((String s)))` is ill formed - old error message 1 is spot on while old error 2 is not particularly helpful.

The new diagnostics 1 and 2 are ok, but 3, 4 and 5 are of low quality - as often happens with DiagnoseParser. Recovering the first error - if it doesn't repair the program fully, further issued diagnostics are of degrading quality. This is an inherent problem with how `DiagnoseParser` works.

To answer you question about real life compiler behavior - yes, the diagnostics can differ. Repair suggestions can change. Sometimes resulting in much better diagnostics - some times degrading the understandability of messages. 

In the past our approach has been to eyeball the changes to make sure that they are reasonable or at least not patently wrong and to remaster the test's expectations

srikanth-sankaran:
For org.eclipse.jdt.core.tests.compiler.regression.InstanceofPrimaryPatternTest.test003() both the new diagnostic 1 and 2 say insert but they imply different insertion point. This is confusing.

(i.e. `if (obj instanceof Type)` and  `if (obj) Statement` use different insertion points for what is recommended to be inserted)

New messages 3,4,5 have degraded to the point of being undecipherable I would say. For example, for the new message 3,  It is not at all obvious how `if (obj instanceof ((String (` may lead to a repair.

This sort of diagnostics changes are not new in the presence of syntax errors

srikanth-sankaran:
Hoping 4th push is the charm! @jarthana after the tests finish green which they should this time, please look through the change once more and approve. TIA.

jarthana (approved):

PR #1613, opened 2023-11-22
Javadoc: fix unclosed <code>

jukzi:


akurtakov:
This PR changes test content to have code tag closed there too that's what breaks tests. Either revert that part or fix the tests with the new content.

PR #1615, opened 2023-11-22
Performance: Add public API to cache Zip Files in UI - closes #1614

### LINKED ISSUE #1614: API for faster UI bulk Operations
Currently performing UI Operations like a "fast" type hierarchy can be kinda slow. By example for `java.lang.Object`.
Sampling i see most time is spend on opening binary JARs. Debugging that i see the same Zip files are opened again and again. JDT CORE already has an internal mechanism to prevent that: `org.eclipse.jdt.internal.core.JavaModelManager.cacheZipFiles(Object)` .
However there is no way to use that in jdt.ui. Example stacktrace:
image
The loop is in the UI and there is not CORE code around it - only jdt ui code.
As a proof of concept i added the internal cacheZipFiles() call around
`org.eclipse.jdt.internal.ui.typehierarchy.HierarchyInformationControl.setInput(Object) ` and
`org.eclipse.jdt.internal.core.SearchableEnvironment.SearchableEnvironment(JavaProject, ICompilationUnit[], boolean)`
Which results in the time spend in org.eclipse.jdt.internal.core.JavaModelManager.getZipFile () is reduced from 48seconds to 40ms - **more then thousand times faster**! (SWT then becomes the bottle neck)
Any objection that JDT provides public API like `org.eclipse.jdt.core.IJavaModel.runBatch(Runnable)` ? 
Other suggestions for the naming? @iloveeclipse @jjohnstn

jukzi:
Also adds trace warnings when caching is missing.
And uses zip cache during some core actions that need it.

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1614


org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
```diff
@@ -5971,6 +5971,75 @@ public static void run(IWorkspaceRunnable action, ISchedulingRule rule, IProgres
...
+			manager.flushZipFiles(instance);
+		}
+	}
+
+	/**
+	 * Runs the argument and will forward it's Exception. To be used on read-only batch operations to cache
+	 * data.
+	 *
+	 * @param runnable
+	 *            A JavaCallable that can throw an Exception
```
szarnekow, line 6027:
```suggestion
	 *            A JavaRunnable that can throw an Exception
```

org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
```diff
@@ -5971,6 +5971,75 @@ public static void run(IWorkspaceRunnable action, ISchedulingRule rule, IProgres
...
+		/**
+		 * Runs or throws an exception.
+		 *
+		 * @throws E the Exception of given type
+		 */
+		void run() throws E;
+	}
+
+	/**
+	 * Calls the argument and returns it's result or its Exception. To be used on read-only batch operations to cache
```
szarnekow, line 6001:
```suggestion
	 * Calls the argument and returns its result or its Exception. To be used on read-only batch operations to cache
```

org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
```diff
@@ -5971,6 +5971,75 @@ public static void run(IWorkspaceRunnable action, ISchedulingRule rule, IProgres
...
+		try {
+			manager.cacheZipFiles(instance);
+			return callable.call();
+		} finally {
+			manager.flushZipFiles(instance);
+		}
+	}
+
+	/**
+	 * Runs the argument and will forward it's Exception. To be used on read-only batch operations to cache
```
szarnekow, line 6023:
```suggestion
	 * Runs the argument and will forward its Exception. To be used on read-only batch operations to cache
```

jukzi: need to learn english :-)

jukzi:
I also added the use of JavaModelManager's temporaryCache to the call() - that further doubles the overall speed of typehierarchy of java.lang.Object in my example workspace

org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
```diff
@@ -5971,6 +5971,75 @@ public static void run(IWorkspaceRunnable action, ISchedulingRule rule, IProgres
...
+			manager.getTemporaryCache();
+			manager.cacheZipFiles(instance);
+			return callable.call();
+		} finally {
+			manager.flushZipFiles(instance);
+			if (!hadTemporaryCache) {
+				manager.resetTemporaryCache();
+			}
+		}
+	}
```
fedejeanne, line 6025: Looking at this, it looks like a `JavaCore` knows too much about the internal workings of `JavaModelManager` (it needs to call 3 methods in the right order in order to make it work properly). I think it would be better to try and hide such details in order to avoid even more unnecessary coupling between these two (already pretty coupled) classes.
A "quick and dirty" solution would be to move everything around `callable.call()` to a new method inside the `JavaModelManager`, something like:

```suggestion
	public static <T, E extends Exception> T call(JavaCallable<T, E> callable) throws E {
		JavaModelManager manager = JavaModelManager.getJavaModelManager();
		Object instance = new Object();
		boolean hadTemporaryCache = manager.hasTemporaryCache();
		try {
			manager.getTemporaryCache();
			manager.cacheZipFiles(instance);
			return callable.call();
		} finally {
			manager.flushZipFiles(instance);
			if (!hadTemporaryCache) {
				manager.resetTemporaryCache();
			}
		}
	}
```

... and you could move `JavaCallable` & Co. down to the `JavaModelManager` too instead of exposing them as public API (they would be `internal`).

fedejeanne: > AFAICT every usage of this new method can get away with a Callable since there are no checked exceptions flying around, am I right?

I just saw that you added a few more calls to `run` / `call` in https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/933 so my last comment is probably already outdated (wrong). You may have to expose `JavaCallable` / `JavaRunnable` but the good thing is that is currently only needed in `eclipse.jdt.ui` which means you could still expose them as `internal` and add `e,jdt.ui` as a "friend of" `e.jdt.core`.

jukzi: I want public api so that others can use it too. for example PDE may useit.  I also saw  minor usecases where jdt-debug would also benefit.
JavaCore  was selected instead of JavaModelManager  as requested in https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1614#issuecomment-1822597670

iloveeclipse (dismissed):

iloveeclipse:
I've requested javdoc change in the review. 

However, looking on the existing `JavaCore.run()` methods contract and implementation, plus looking on `JavaModelOperation`, I wonder if the provided implementation is sufficient for long term.

The main problem I see is in this assumption from the javadoc: "The given callable is supposed to query Java model and do not modify it."

Similar to `IProgressMonitor.beginTask()`, where "contract" was just saying "This must only be called once on a given progress monitor instance" and no one checked if that is actually done, we may run later into problems with this new API if we expect the clients to behave "nice" - in reality, no one cares to read javadoc and without checks on our side this would just cause bugs and frustration in long term.

What I imagine, why don't we create a dedicated `CacheOptimizedOperation extends JavaModelOperation` class, similar to `CreateTypeHierarchyOperation` (with `canModifyRoots()` returning false and `isReadOnly()` returning true) and then 
 run internally `new CacheOptimizedOperation(callable).run(monitor);` .

Unfortunately as of today, if I see it right, this above would not add yet checks if anything was modified during the operation, but that can be added in `JavaModelOperation.run(IProgressMonitor)` code that walks over the java deltas in the  finally block. So I can imagine we can check there if any delta was created during "read only" operation and report that as an error.

WDYT?

jukzi:
I will adjust javadoc.

> What I imagine, why don't we create a dedicated `CacheOptimizedOperation extends JavaModelOperation` class

The decision Class vs static method is independent from the other topics you addressed. I prefer the static method as it is very easy to use for the client (see examples in UI). To prevent confusion with write-operation the method could be either renamed or put into another class - for example IJavaModel, or a dedicated class "JavaModelRead". None of clients i have seen is aware of the workspace.

> if I see it right, this above would not add yet checks if anything was modified during the operation

That's true. However i don't see a need for it. And i don't want the implementation to be a WorkspaceRunnable that needs to care about locking. I am afraid such checks may introduce memory or performance penalties and potentially deadlocks.
Nevertheless the proposed API would not prevent the implementation to check for writes, if somebody wants to contribute such checks.

org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
iloveeclipse: Not sure why my comment is unclear.

I meant: To be consistent, new public "run something on Java model" API methods should be added on JavaCore, next to existing API. It makes sense to keep similar API in same class.

fedejeanne: That's what I thought 👍 , which is good because then @jukzi can use my suggestion and keep the current API consistent without adding more coupling between `JavaCore` and `JavaModelManager`:
* The entry point is still a `run`-method in `JavaCore`
* `JavaCore` delegates to `JavaModelManager`, which knows its own internals and knows how to properly handle caching

org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
```diff
@@ -5971,6 +5971,75 @@ public static void run(IWorkspaceRunnable action, ISchedulingRule rule, IProgres
...
+		/**
+		 * Runs or throws an exception.
+		 *
+		 * @throws E the Exception of given type
+		 */
+		void run() throws E;
+	}
+
+	/**
+	 * Calls the argument and returns its result or its Exception. To be used on read-only batch operations to cache
```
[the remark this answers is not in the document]
jukzi: done

jukzi:
i changed to names to runReadOnly so that it is obvious that argument should read only

org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
jukzi: JavaCore now delegates to JavaModelManager

fedejeanne:
I find the discussion here very interesting but I have to admit I'm lacking some concepts and I can't follow. Would someone please provide an example of how the new methods `call-` and `runReadOnly` could be misused? It sounds like it would be very easy to get this new API wrong and pass some invalid callable/runnable into them.

If that's the case then the new API should not only make it obvious that the operation is `read-only`, it should also state that there will be no checks and that it's the sole responsibility of the user to check what he/she passes on: one could have 2 sets of "read-only" methods:
1. `call`-/`runReadOnlyUnsafe`: does not check if the callable/runnable performed some writing (fast but dangerous)
2. `call`-/`runReadOnly`: checks if the callable/runnable performed some writing (slow but safe)

@jukzi I'm not saying you need to provide both versions of the methods, just leave the door open so whoever comes along knows that there's a gap to be filled _i.e._:
* Rename `call-`/`runReadOnly`: add "`Unsafe`" at the end
* (Optional) Create the "safe" versions (`call-`/`runReadOnly`) and throw an `UnsupportedOperationException` or similar.

What do you guys think?

jukzi:
> `call-` and `runReadOnly` could be misused

litterally all current usages of WorkbenchRunnableAdapter / JavaCore.run. For example all refactorings are write operations. 

It does not make sense to have a separation between runReadOnlyUnsafe and runReadOnly if nobody is going to invest the time to make a foolproof version.

fedejeanne:
> > `call-` and `runReadOnly` could be misused
> 
> literally all current usages of WorkbenchRunnableAdapter / JavaCore.run. For example all refactorings are write operations.

So for example doing along these lines (compilation problems aside) ... 

```java
IProgressMonitor pm = ...;
CreateFieldOperation createField = ...; // write operation
JavaCore.runReadOnly(() -> createField.runOperation(pm));
```

.... would be wrong but it would be allowed by the new API, right? 
What would the result be: some exception being thrown or some weird (possibly unnoticed) behavior?
 
> It does not make sense to have a separation between runReadOnlyUnsafe and runReadOnly if nobody is going to invest the time to make a foolproof version.

If I got the gist of this conversation right, you're opening the door to some serious bugs and maintenance efforts in the future and all that's preventing that from happening is the documentation. Even if the "safe" version is not provided in this PR (which I'm OK with), it's worth making it **really** obvious for the users of this API that passing some **write** operations might have undesired (even nondeterministic?) behavior by renaming the current method to `...Unsafe` and by explicitly mentioning this in the documentation. Something like this in the JavaDocs should help: `The argument's {@code call()/run()} is supposed to query the Java model and must not modify it. Failing to do so i.e. passing a write operation will have undesired/non-deterministic effects since this method doesn't do any kind of validation checks`

jukzi:
> What would the result be: some exception being thrown or some weird (possibly unnoticed) behavior?

i don't even know, by chance it will even work. It's forbidden - you should not do that. period. i am not opening any door, its already possible to write wrong code - we can't protect against that.

jukzi:
i added a  `IllegalStateException("Its not allow to modify JavaModel during ReadOnly action.")` to JavaCore.run()

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelOperation.java
```diff
@@ -805,12 +805,13 @@ public void runOperation(IProgressMonitor monitor) throws JavaModelException {
 		}
 		try {
 			if (isReadOnly()) {
-				run(monitor);
+				JavaCore.runReadOnly(() -> run(monitor));
```
jukzi, line 808: Thats interesting: the failing tests - like JavaSearchMultipleProjectsTests.testFieldOccurencesInWorkingCopies() - show that isReadOnly() in this context only means no resources are modified - but still the javamodel is modified.
I will remove the runReadOnly() here.

fedejeanne: hm, that's also a design problem because if `isReadOnly()` can be wrong (e.g. if a new subclass forgot to override it and set it to return `true`) then the verification mechanism in `JavaCore::runReadOnly` would not throw an exception. Isn't there a way to check for the Deltas of the `JavaModelManager` instead of relying on `isReadOnly()`?

fedejeanne:
> i added a `IllegalStateException("Its not allow to modify JavaModel during ReadOnly action.")` to JavaCore.run()

I tried this (I placed it in the plugin `org.eclipse.jdt.core.tests.model`) but the test failed:

```java
@Test
public void testRunReadOnly_writeOperation_IllegalStateExceptionIsThrown() throws JavaModelException {

	JavaModelOperation writeOperation = new JavaModelOperation() {
		@Override
		public boolean isReadOnly() {
			return false;
		}

		@Override
		protected IJavaModelStatus verify() {
			return JavaModelStatus.VERIFIED_OK;
		}

		@Override
		protected void executeOperation() throws JavaModelException {
			// do some writing
			JavaElementDelta delta = new JavaElementDelta(JavaModelManager.getJavaModelManager().getJavaModel());
			addDelta(delta);
		}
	};

	assertThrows(IllegalStateException.class, //
			() -> JavaCore.runReadOnly(() -> writeOperation.runOperation(null)));
}
```

The code doesn't even walk into `org.eclipse.jdt.core.JavaCore.run(IWorkspaceRunnable, ISchedulingRule, IProgressMonitor)` so I don't think the current verification works :-\

jukzi:
> The code doesn't even walk into `org.eclipse.jdt.core.JavaCore.run(IWorkspaceRunnable, ISchedulingRule, IProgressMonitor)` so I don't think the current verification works :-\

The check is not all-inclusive, i will not work on a complete check thats by far out of scope, but it would have detected https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/933#discussion_r1407680536 - if it would have been a problem

iloveeclipse:
@szarnekow, @stephan-herrmann, @trancexpress : I'm pretty busy this week.
Could you please check the proposed API / contract / javadoc / client side changes?

org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
```diff
@@ -5971,6 +5972,68 @@ public static void run(IWorkspaceRunnable action, ISchedulingRule rule, IProgres
...
+		 * @return the result
+		 * @throws E the Exception of given type
+		 */
+		V call() throws E;
+	}
+	/**
+	 * @since 3.37
+	 */
+	@FunctionalInterface
+	public static interface JavaRunnable<E extends Exception> {
```
trancexpress, line 5992: `JavaCoreRunnable`? `JDTRunnable`? `JavaModelRunnable`?

`JavaCore` is not much better than `JavaRunnable` as a name, so I'm fine with the name here. But maybe some better name can be invented for the new code here? Same for the one above.

org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
```diff
@@ -5971,6 +5972,68 @@ public static void run(IWorkspaceRunnable action, ISchedulingRule rule, IProgres
...
+	 * Calls the argument and returns its result or its Exception. The argument's {@code call()} is supposed to query
+	 * Java model and must not modify it. This method will try to run Java Model queries in optimized way (caching
+	 * things during the operation). It is safe to nest multiple calls - but not necessary.
+	 *
+	 *
+	 * @param callable
+	 *            A JavaCallable that can throw an Exception
+	 * @return the result
+	 * @exception E
+	 *                An Exception that is thrown by the argument.
```
trancexpress, line 6012: > An {@link Exception} that could be thrown by the {@code callable}?

org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
```diff
@@ -5971,6 +5972,68 @@ public static void run(IWorkspaceRunnable action, ISchedulingRule rule, IProgres
...
+		 * @throws E the Exception of given type
+		 */
+		void run() throws E;
+	}
+
+
+	/**
+	 * Calls the argument and returns its result or its Exception. The argument's {@code call()} is supposed to query
+	 * Java model and must not modify it. This method will try to run Java Model queries in optimized way (caching
+	 * things during the operation). It is safe to nest multiple calls - but not necessary.
```
trancexpress, line 6005: > things

Is there some better word here instead of "things"?

org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
jukzi: Since these are @FunctionalInterface their names will unlikely be ever written. Typically it will be just a lambda. so name does not matter here

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java
```diff
@@ -2976,19 +2983,50 @@ public ZipFile getZipFile(IPath path) throws CoreException {
...
 		ZipFile zipFile;
-		if ((zipCache = this.zipFiles.get()) != null
-				&& (zipFile = zipCache.getCache(path)) != null) {
+		if ((zipCache = this.zipFiles.get()) != null && (zipFile = zipCache.getCache(path)) != null) {
 			return zipFile;
 		}
 		File localFile = getLocalFile(path);
-
 		try {
+			if (ZIP_ACCESS_WARNING) {
```
trancexpress, line 3002: Can this whole block be moved to a method?

trancexpress:
Not a big fan of adding new APIs, without proposals on how other code will be improved by the APIs.

Other than that, the change looks fine. Would be great if the thread locals can be avoided, but I imagine that is too much work.

jukzi:
> Not a big fan of adding new APIs, without proposals on how other code will be improved by the APIs.

see https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/933 for an example

> Other than that, the change looks fine. Would be great if the thread locals can be avoided, but I imagine that is too much work.

I don't know how that could be archived, Also i don't know why this should be done: The Boolean is light-weight and the other is only used on tracing

trancexpress:
> see [eclipse-jdt/eclipse.jdt.ui#933](https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/933) for an example

Alright, thanks, that looks good.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java
jukzi: done

PR #1620, opened 2023-11-23
core.compiler.batch: declare org.eclipse.jdt.apt as x-friends

jukzi:
https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1548#issuecomment-1814661318


stephan-herrmann (approved):
LGTM, thanks.

akurtakov:
@jukzi When a new stream opens would you please rebase on top of master before pushing? That would catch missing version bump here. Note that this breaks verification build for aggregator build now https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1604 .
I'll fix it now but I would really appreciate if this burden is shared a bit more.

jukzi:
> I'll fix it now but I would really appreciate if this burden is shared a bit more.

thanks, i am on it.

PR #1623, opened 2023-11-24
[performance] ClassFileReader: use already open Zip File

jukzi:
instead of creating new ZipFile instance.

also:
* use Files.readAllBytes
* removed unused code
* faster toUri avoiding isDirectory() check for the JARs which are known
to be no directory

improves performance of read() on windows by factor 2
tested with java reference search to java.lang.Object

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/classfmt/ClassFileReader.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/classfmt/ClassFileReader.java
@@ -23,6 +23,7 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.Arrays;
 import java.util.function.Predicate;
 
@@ -156,19 +157,38 @@ public static ClassFileReader readFromModule(
 		throws ClassFormatException, java.io.IOException {
 		return JRTUtil.getClassfile(jrt, filename, moduleName, moduleNameFilter);
 }
-public static ClassFileReader read(
-	java.util.zip.ZipFile zip,
-	String filename,
-	boolean fullyInitialize)
-	throws ClassFormatException, java.io.IOException {
+
+public static ClassFileReader read(java.util.zip.ZipFile zip, String filename, boolean fullyInitialize)
+		throws ClassFormatException, java.io.IOException {
 	java.util.zip.ZipEntry ze = zip.getEntry(filename);
-	if (ze == null)
+	if (ze == null) {
 		return null;
-	ClassFileReader classFileReader = Util.getZipEntryClassFile(zip.getName(), filename);
-	if (fullyInitialize) {
-		classFileReader.initialize();
 	}
-	return classFileReader;
+	try (InputStream stream = zip.getInputStream(ze)) {
+		URI uri =  URI.create("jar:file://" + toUri(zip.getName()).getRawPath() + "!/" + filename); //$NON-NLS-1$ //$NON-NLS-2$
+		ClassFileReader classFileReader = new ClassFileReader(uri, Util.getInputStreamAsByteArray(stream),
+				filename.toCharArray());
+		if (fullyInitialize) {
+			classFileReader.initialize();
+		}
+		return classFileReader;
+	}
+}
+
+/**
+ * same as <code>new java.io.File(absoluteNormalFilePath).toURI()</code> if absoluteNormalFilePath is not a directory
+ * but faster because it avoid IO for the isDirectory check.
+ **/
+private static URI toUri(final String absoluteNormalFilePath) {
+	String p = absoluteNormalFilePath.replace(File.separatorChar, '/');
+	if (!p.startsWith("/")) { //$NON-NLS-1$
+		p = "/" + p; //$NON-NLS-1$
+	}
+	try {
+		return new URI("file", null, p, null); //$NON-NLS-1$
+	} catch (URISyntaxException x) {
+		throw new RuntimeException(x);
+	}
 }
 
 public static ClassFileReader read(String fileName) throws ClassFormatException, java.io.IOException {
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/tool/Util.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/tool/Util.java
@@ -14,7 +14,6 @@
 package org.eclipse.jdt.internal.compiler.tool;
 
 import java.io.IOException;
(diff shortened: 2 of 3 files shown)


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/classfmt/ClassFileReader.java
```diff
@@ -156,19 +159,24 @@ public static ClassFileReader readFromModule(
...
 	java.util.zip.ZipEntry ze = zip.getEntry(filename);
-	if (ze == null)
+	if (ze == null) {
 		return null;
-	ClassFileReader classFileReader = Util.getZipEntryClassFile(zip.getName(), filename);
-	if (fullyInitialize) {
-		classFileReader.initialize();
 	}
-	return classFileReader;
+	try (BufferedInputStream stream = new BufferedInputStream(zip.getInputStream(ze))) {
```
szarnekow, line 169: Since `Util.getInputStreamAsByteArray` uses `stream.readAllBytes`, the additional buffering could/should be unnecessary.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/classfmt/ClassFileReader.java
```diff
@@ -156,19 +159,24 @@ public static ClassFileReader readFromModule(
...
+		Path zipfile = Paths.get(zip.getName());
+		URI uri1 = URI.create("jar:file://" + zipfile.toUri().getRawPath() + "!/" + filename); //$NON-NLS-1$ //$NON-NLS-2$
```
szarnekow, line 171: inline `uri1`

szarnekow: Also we might want to move toJarUri to this class.

szarnekow: There might even be room for improvements.
A ZipEntry knows its uncompressed size in bytes. We could allocate the byte array with the correct size and call `read(byte[], offset, len)` in a loop until it's fully populated. This should spare us all the intermediate allocations.

jukzi: You should report that to JDK. Probably InflaterInputStream.readAllBytes() could return InflaterInputStream.buf for example

jukzi: > Also we might want to move toJarUri to this class.

its not used anywhere else. there is still a org.eclipse.core.runtime.URIUtil.toJarURI(URI, IPath)

jukzi: > inline `uri1`

done

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/classfmt/ClassFileReader.java
jukzi: done.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/classfmt/ClassFileReader.java
szarnekow: > You should report that to JDK. Probably InflaterInputStream.readAllBytes() could return InflaterInputStream.buf for example

I don't understand what you're suggesting.

ZipEntry offers `getSize` which returns the uncompressed size

```
byte[] bytes = new byte[ze.getSize()];
stream.readNBytes(bytes, 0, bytes.length);
```
*Should* be the fastest way to read the bytes in this particular case.

jukzi: stream.readAllBytes() should be the fastest way to read all bytes. if it is not the stream implementation should be improved. JDK already did that for Files, with massive performance boost

szarnekow: stream.readAllBytes() allocates a number of intermediate byte arrays that will be concatenated in the end to a new one. I don't see how that can be faster than allocating a single array and reading into that one directly.

szarnekow: Note that it is not the stream implementation's "fault" or responsibility. The stream can/should not make any assumptions about the offset within the stream when readAllBytes is called. Here, we want to read all bytes from start to end. Therefore a targeted optimization is possible, which would not be possible for the inflater stream without breaking encapsulation or maintaining additional state.

jukzi: A stream provided by ZipFile does know its size itself. see for example https://bugs.openjdk.org/browse/JDK-8264777 how good JDK solved such reports in the past.

szarnekow: Ok, fair enough. I'd argue optimize locally for the time being until a general fix is available, but I understand your point.

jukzi:
Example measurement:
image
before:
image

after:
image

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/classfmt/ClassFileReader.java
laeubi: @szarnekow can you explain why buffering is not needed with `stream.readAllBytes`

API says
> Note that this method is intended for simple cases where it is
> convenient to read all bytes into a byte array. **It is not intended for
> reading input streams with large amounts of data**

while one might argue that "large" is not the case here, buffering can still be good even if `stream.readAllBytes` is invoked? Or is Zip using some data from the archive to know the size in advance?

szarnekow: readAllBytes uses internal buffers to perform bulk reads on the stream. The buffer themselves. The difference between buffered reads and direct reads is most notable if bytes are read one-by-one. That’s not the case here

PR #1625, opened 2023-11-24
fix node detector to detect correct parent node. Fix #1561

### LINKED ISSUE #1561: Content assist does not activate, inside if() within a switch
_I'm migrating [bugzilla 576934](https://bugs.eclipse.org/bugs/show_bug.cgi?id=576934) here, as requested in (the closely-related) github issue #341_

I had upgraded from Eclipse 4.20 > 4.21 to pick up fixes for content assist regressions, but I ran into this still-failing case:

    public class Snippet {
      final String name = "test";
      enum Type { A }

      private void go(Type a) {
        switch (a) {
        case A:
          if (nam)
          break;
        }
      }
    }

Trying ctrl+space after 'nam' in the incomplete 'if' statement still shows nothing, similarly to [bugzilla 573632](https://bugs.eclipse.org/bugs/show_bug.cgi?id=573632)

Also per a workaround mentioned in there, adding another symbol after the identifier gets completion working again:

  `if (nam <)`

Deleting the closing ')' also kicks it into working:

  `if (nam`

Note, however, that Eclipse by default inserts that closing ')' for us. 

I have confirmed this is still present in a recent nightly - Eclipse 2023-12 (4.30) Build id: I20231024-1800

gayanper:
## What it does
Fix the issue with detecting the parent node when If condition is nested inside a switch, the reason is IfStatment are not visited while leaving which cause the node detection the while the closest parent to be the switch.

## How to test
Provide sample code and added Unit Tests

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests.java
@@ -2066,7 +2066,8 @@ public void testCompletionAfterInstanceof01() throws JavaModelException {
 	int cursorLocation = str.lastIndexOf(completeBehind) + completeBehind.length();
 	this.workingCopies[0].codeComplete(cursorLocation, requestor, this.wcOwner);
 
-	int relevance1 = R_DEFAULT + R_RESOLVED + R_INTERESTING + R_CASE + R_NON_STATIC + R_NON_RESTRICTED;
+	int relevance1 = R_DEFAULT + R_RESOLVED + R_INTERESTING + R_CASE + R_NON_STATIC + R_NON_RESTRICTED + R_VOID;
+	int relevance2 = R_DEFAULT + R_RESOLVED + R_INTERESTING + R_CASE + R_NON_STATIC + R_NON_RESTRICTED + R_EXACT_EXPECTED_TYPE;
 	int start1 = str.lastIndexOf("equal") + "".length();
 	int end1 = start1 + "equal".length();
 	int start2 = str.lastIndexOf("a.equal");
@@ -2074,8 +2075,8 @@ public void testCompletionAfterInstanceof01() throws JavaModelException {
 	int start3 = str.lastIndexOf("a.");
 	int end3 = start3 + "a".length();
 	assertResults(
-			"equals[METHOD_REF]{equals(), Ljava.lang.Object;, (Ljava.lang.Object;)Z, equals, (obj), replace["+start1+", "+end1+"], token["+start1+", "+end1+"], " + (relevance1) + "}\n" +
-			"equalsFoo[METHOD_REF_WITH_CASTED_RECEIVER]{((CompletionAfterInstanceOf)a).equalsFoo(), Ltest.CompletionAfterInstanceOf;, ()V, Ltest.CompletionAfterInstanceOf;, equalsFoo, null, replace["+start2+", "+end2+"], token["+start1+", "+end1+"], receiver["+start3+", "+end3+"], " + (relevance1) + "}",
+			"equalsFoo[METHOD_REF_WITH_CASTED_RECEIVER]{((CompletionAfterInstanceOf)a).equalsFoo(), Ltest.CompletionAfterInstanceOf;, ()V, Ltest.CompletionAfterInstanceOf;, equalsFoo, null, replace["+start2+", "+end2+"], token["+start1+", "+end1+"], receiver["+start3+", "+end3+"], " + (relevance1) + "}\n" +
+			"equals[METHOD_REF]{equals(), Ljava.lang.Object;, (Ljava.lang.Object;)Z, equals, (obj), replace["+start1+", "+end1+"], token["+start1+", "+end1+"], " + (relevance2) + "}",
 			requestor.getResults());
 }
 public void testCompletionAfterInstanceof01_02() throws JavaModelException {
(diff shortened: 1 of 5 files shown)


org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests.java
```diff
@@ -2066,16 +2066,17 @@ public void testCompletionAfterInstanceof01() throws JavaModelException {
 	int cursorLocation = str.lastIndexOf(completeBehind) + completeBehind.length();
 	this.workingCopies[0].codeComplete(cursorLocation, requestor, this.wcOwner);
 
-	int relevance1 = R_DEFAULT + R_RESOLVED + R_INTERESTING + R_CASE + R_NON_STATIC + R_NON_RESTRICTED;
+	int relevance1 = R_DEFAULT + R_RESOLVED + R_INTERESTING + R_CASE + R_NON_STATIC + R_NON_RESTRICTED + R_VOID;
+	int relevance2 = R_DEFAULT + R_RESOLVED + R_INTERESTING + R_CASE + R_NON_STATIC + R_NON_RESTRICTED + R_EXACT_EXPECTED_TYPE;
```
stephan-herrmann, line 2070: I'm afraid all these additions of `R_EXACT_EXPECTED_TYPE` might be wrong.
Inside `computeExpectedTypes()` we now trigger this block:
```
		} else if (parent instanceof IfStatement) {
			addExpectedType(TypeBinding.BOOLEAN, scope);
```

That's what makes `boolean equals()` appear as a perfect match, even though the boolean result is ignored!!

So the engine seems to believe we are seeing the condition of an if-statement, which is not the case.

stephan-herrmann (changes requested):
The actual change in `CompletionNodeDetector` seem to make sense, but ...

Tests are really hard to read due to the format, but more importantly, I think we're not done in what to read from `astNodeParent instanceof IfStatement`.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests14.java
```diff
@@ -881,4 +881,75 @@ public void testGH697_CompletionOnSwitchExpressionAsFirstParameter_OnInitializer
...
+				requestor.getResults());
+	}
+
+	public void testGH1561_CompletionInWhileConditionInsideASwitchStatement() throws JavaModelException {
+		this.workingCopies = new ICompilationUnit[1];
+		this.workingCopies[0] = getWorkingCopy("/Completion/src/SwitchIf.java",
+				"public class SwitchIf {\n" + "		final String name = \"test\";\n" + "		enum Type { A }\n"
+						+ "		private void foo(Type input) {\n" + "			switch (input) {\n"
+						+ "			case A:\n" + "				while (nam)\n" + "				break;\n"
+						+ "			}\n" + "		}\n" + "}\n");
```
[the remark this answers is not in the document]
gayanper: This was a u fortunate autoformat on save. But yes i used text blocks in some of the new tests. But here i used what was used in other tests. But i think i will switch to textblocks. It's easy for future devs.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests.java
gayanper: Since we are completing inside

`if (nam)` isn't it correct that the expected type is a boolean ?

Because for non qualified member completions this might help to suggest methods and variables from current scope which has return value of boolean right ?

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
```diff
@@ -1954,7 +1955,7 @@ private void buildTokenLocationContext(InternalCompletionContext context, Scope
 				if (expression.body().sourceStart <= astNode.sourceStart &&
 						astNode.sourceEnd <= expression.body().sourceEnd) {
 					// completion is inside a method body
-					if ((astNodeParent == null || astNodeParent == expression || (astNodeParent instanceof Block)) &&
+					if ((astNodeParent == null || astNodeParent == expression || (astNodeParent instanceof Block) || inParentControlStatement(astNodeParent)) &&
```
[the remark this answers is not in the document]
gayanper: Yes it will, I'm not sure if we have another `START` constant for this, but there were some tests which was expecting this while completion in side a if condition. Let me try to find those tests.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests.java
stephan-herrmann: > Since we are completing inside
> `if (nam)` isn't it correct that the expected type is a boolean ?

No.
Tons of test in CompletionEngine now signal R_EXACT_EXPECTED_TYPE for boolean from code like
```
if (a instanceof CompletionAfterInstanceOf)
    a.equal|
```
At the '|' location expecting boolean is plain wrong.

With your change in CompletionNodeDetector every location in CompletionEngine that checks `assistNodeParent instanceof IfExpression` must also check _where_ in the if statement the assistNode sits: condition or then statement or else statement.

stephan-herrmann:
> @[gayanper](https://github.com/gayanper) gayanper [force-pushed](https://github.com/eclipse-jdt/eclipse.jdt.core/compare/d661bd3f870ffc441e686d3a10484a3bf67cc04d..34e8ed29ef4e4d66b4968733c9536c7f6d789d34) the fix_if_inside_switch branch from[ d661bd3 ](https://github.com/eclipse-jdt/eclipse.jdt.core/commit/d661bd3f870ffc441e686d3a10484a3bf67cc04d)to 34e8ed2

This looks like a rebase on master? If this force-push contains any changes of your own I have to withdraw from reviewing, because for continued reviewing I _need_ to see the differences between rounds of reviews (yes, I sound like a broken record, maybe I am).

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests.java
gayanper: Right got it, i think we should make the parent null in such scenarios rather than wrong fall to what is above, like the switch in this particular scenario. 

I will try to amend the changes and see if my idea will work.

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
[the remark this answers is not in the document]
gayanper: With latest changes this change is not needed.

gayanper:
@stephan-herrmann what do you think about the latest changes ?

gayanper:
@stephan-herrmann could you check the latest changes since this PR was open for long time now ?

stephan-herrmann:
Clickable link to the issue: #1561

stephan-herrmann (commented):
I have a bunch of questions & comments, but before answering please wait a bit, as I'm preparing an alternative solution.

stephan-herrmann:
Superseded by #1953

PR #1630, opened 2023-11-27
TestVerifier: never wait 100ms

jukzi:
improve test time, accurate timeout


jukzi:
See https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1564#issuecomment-1795176613
But with accurate timeout to prevent failing tests.

PR #1690, opened 2023-12-05
CompilationUnitResolver: Name the CU that causes Exception

jukzi:
for example during Cleanup
https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/950

image

### DIFF
--- a/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolver.java
+++ b/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolver.java
@@ -1004,9 +1004,12 @@ class CompilationUnitResolver extends Compiler {
 						ast.setDefaultNodeFlag(0);
 						ast.setOriginalModificationCount(ast.modificationCount());
 
-						// pass it to requestor
-						astRequestor.acceptAST(source, compilationUnit);
-
+						try {
+							// pass it to requestor
+							astRequestor.acceptAST(source, compilationUnit);
+						} catch (RuntimeException e) {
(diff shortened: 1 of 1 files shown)


jukzi:
before it was just: 
image

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolver.java
```diff
@@ -1004,9 +1004,12 @@ private void resolve(
...
 						ast.setOriginalModificationCount(ast.modificationCount());
 
-						// pass it to requestor
-						astRequestor.acceptAST(source, compilationUnit);
-
+						try {
+							// pass it to requestor
+							astRequestor.acceptAST(source, compilationUnit);
+						} catch (RuntimeException e) {
+							throw new RuntimeException("Error on " + source.getPath(), e); //$NON-NLS-1$
```
jjohnstn, line 1011: Shouldn't this be a translated string since it will appear with translated dialog text in your example?

jukzi: > Shouldn't this be a translated string since it will appear with translated dialog text in your example?

Typically we don't have translated Strings in Exceptions. We won't like to get Issues reported with exceptions in arbitrary user language. 
Priority is that those Exceptions just do not occur.

jjohnstn: Ok

Issue #421, opened 2022-09-19
Random failing org.eclipse.jdt.core.tests.model.JavaSearchBugsTests2.testBug483650_0002


jukzi:
https://github.com/eclipse-jdt/eclipse.jdt.core/pull/304
https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-304/4/

```
Unexpected search results.
----------- Expected ------------
/P/src/p478042/AllMethodDeclarations01.java char p478042.AllMethodDeclarations01.foo03(Object o,String s)\n
/P/src/p478042/AllMethodDeclarations01.java int p478042.AllMethodDeclarations01.foo02(Object o)\n
/P/src/p478042/AllMethodDeclarations01.java void p478042.AllMethodDeclarations01.foo01()\n
/P/src/p478042/AllMethodDeclarations01b.java Integer p478042.AllMethodDeclarations01b.fooInt()
------------ but was ------------
/P/src/p478042/AllMethodDeclarations01.java char p478042.AllMethodDeclarations01.foo03(Object o,String s)\n
/P/src/p478042/AllMethodDeclarations01.java int p478042.AllMethodDeclarations01.foo02(Object o)\n
/P/src/p478042/AllMethodDeclarations01.java void p478042.AllMethodDeclarations01.foo01()
--------- Difference is ----------
 expected:<...clarations01.foo01()[\n
/P/src/p478042/AllMethodDeclarations01b.java Integer p478042.AllMethodDeclarations01b.fooInt()]> but was:<...clarations01.foo01()[]>
```

jukzi:
it's an AbstractJavaModelTests - may be fixed with https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1612. reopen if it happens again

jukzi:
happend again
https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-1700/2/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugsTests2/testBug483650_0001/
```

/P/src/p478042/AllMethodDeclarations01b.java Integer p478042.AllMethodDeclarations01b.fooInt()]> but was:<...clarations01.foo01()[]>
	at org.eclipse.jdt.core.tests.model.JavaSearchBugsTests2.testBug483650_0001(JavaSearchBugsTests2.java:2397)
```

srikanth-sankaran:
[...]

PR #1700, opened 2023-12-06
tests: enable discouragedReference=warning, declare x-friends

jukzi:
to get rid of warnings during build (does not respect the jdt preferences)
image


jukzi:
ignoring random fail https://github.com/eclipse-jdt/eclipse.jdt.core/issues/421

org.eclipse.jdt.core.tests.model/.settings/org.eclipse.jdt.ui.prefs
```diff
@@ -10,47 +10,137 @@ sp_cleanup.add_missing_nls_tags=false
...
 sp_cleanup.never_use_parentheses_in_expressions=true
+sp_cleanup.no_string_creation=false
+sp_cleanup.no_super=false
+sp_cleanup.number_suffix=false
+sp_cleanup.objects_equals=false
 sp_cleanup.on_save_use_additional_actions=true
-sp_cleanup.organize_imports=false
+sp_cleanup.one_if_rather_than_duplicate_blocks_that_fall_through=false
+sp_cleanup.operand_factorization=false
+sp_cleanup.organize_imports=true
```
stephan-herrmann, line 66: If this change was intentional: would have been good to discuss this with the team, at least: mention.

If not intentional, I suggest to be more careful about changes not necessary for the current fix.

jukzi: Hi Stephan, i am sorry that this slipped through.  It's on my TODO list to ask if we could please enable organize imports on all jdt projects since i lose much time to manually fix imports.  Since you are mentioning this topic, can you please elaborate why you dislike organize imports (that is how i interpret your comment)?

stephan-herrmann: >  i am sorry that this slipped through.

We don't want to introduce mandatory reviews for all PRs, do we? :wink: 

I found this, because I in turn lost time, trying to avoid unrelated changes in one of my PRs.

If you start a discussion and mention me, I will add more details there. Hint: if the change of strategy is done well, I'm not strictly against it.

PR #1716, opened 2023-12-08
Enhance resource leak analysis with additional annotation(s)

### LINKED ISSUE #1530: Enhance resource leak analysis with additional annotation(s)
In https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1447#issuecomment-1760403838 I was reminded that resource leak analysis is inherently incomplete when a resource is passed to a method invocation (either as receiver or argument).

This could be improved by supporting one or more additional annotations.

Some references re annotations for resource leak analysis:
* https://stackoverflow.com/questions/75871642/can-jsr-305s-willclose-et-al-annotation-or-one-to-the-same-effect-be-imp
* The above links to https://checkerframework.org/manual/#resource-leak-checker

Their pair of annotations @MustCall and @CalledMethods is actually more general then the issue of resource analysis: in our case we are interested in calls to exactly one method only: `close()`.

In an attempt to generalize in a different direction, I'm thinking of adding a variant of `@Owning`, plus possible `@NotOwning` with the following semantics:
* ~~I'd design these annotations as `TYPE_USE` annotations.~~ No, that would be overselling the idea. This annotation doesn't play well with generics (_also the folks of the Checker Framework came to the same conclusion_).
* passing a resource into a method invocation where the corresponding parameter has an `@Owning` type delegates the responsibility to close the argument resource to the body of the target method.
* invoking a method on a resource instance, where the method has an `@Owning` `this`-parameter delegates the responsibility to close the receiver to the body of the target method.
* returning a value from a method that has an `@Owning` return type, delegates responsibility to close the returned resource to the caller.

In https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1447#issuecomment-1760426031 I imagined a pair of annotations to support the parameter and the receiver scenario, which can be avoided by the above approach as `TYPE_USE` annotation, which will be applicable also to a `this`-parameter.

While we should always be careful with introducing our own annotations with JDT-specific semantics, I think the above approach should be future proof (e.g., in case some day jspecify will specify similar annotations):
* We would again apply the dual strategy of supplying a default annotation type via `org.eclipse.jdt.annotation` (v2) plus an option to specify a custom annotation type via its FQN.
* We would start with minimal semantic evaluation as outlined above, but hopefully other ownership-related evaluation could follow, should such become mainstream in a (distant) future.

Even only an `@Owning` annotation will help to significantly improve analysis, where currently the analysis applies a lot of hand-waiving as to say "may", "potentially" etc.:
* avoid false positives at the origin of a data flow
* strengthen potential problems to errors at the receiving end

Additionally supported `@NotOwning` would improve in the inverse direction:
* avoid false positives at the receiving end
* strengthen some potential problems at the origin of a data flow to error.

stephan-herrmann:
fixes #1530

## What it does

Demonstrate how evaluating one more annotations `@Owning` can enhance our resource leak analysis.

Most of the change is infra structure work to define new options, let the compiler detect that annotation and propagate that information to where it is needed.

Few semantically interesting code changes suffice to tell the existing analysis:
1. a method parameter of a closeable type and marked `@Owning` **must** be closed in, or on behalf of this method.
2. passing an unclosed closeable value into a method where the receiving parameter is marked `@Owning` delegates responsibility of closing to the receiving method.
3. returning an unclosing closeable value from a method annotated as `@Owning` delegates responsibility to the calling method.

The above can be mapped as equivalent to existing constructs:
1. corresponds to acquiring a fresh resource, e.g., by allocating a new instance.
2. corresponds to calling `close()` on the resource being sent to the other method
3. corresponds to calling `close()` on the resource being returned from the current method

## How to test

1. Ensure you have `org.eclipse.jdt.annotation` bundle on you buildpath (with addition in this PR)
2. Enter the following lines in the project's `.settings/org.eclipse.jdt.core.prefs`:
```
org.eclipse.jdt.core.compiler.annotation.resourceanalysis=enabled
org.eclipse.jdt.core.compiler.problem.potentiallyUnclosedCloseable=warning
org.eclipse.jdt.core.compiler.problem.unclosedCloseable=error
```

In particular the first option doesn't yet have UI. By specifying `org.eclipse.jdt.core.compiler.annotation.owning=your.Annot` you could optionally roll you own equivalent of `@org.eclipse.jdt.annotation.Owning`.

With this setup interested parties can play with the new analysis. Have a look at new tests or the following example for inspiration:

```
public class Producer {
	@Owning InputStream get() throws FileNotFoundException {
		return new FileInputStream("foo");
	}
}
```
```
public class Consumer {
	@Owning InputStream inspect(@Owning InputStream input) {
		if (input != null)
			System.out.println("ok");
		return input;
	}
	public void consume(@Owning AutoCloseable rc) {
		try (rc) {
			System.out.println("reading");
		} catch (Exception e) {
			e.printStackTrace();
		}
	}
}
```
```
public class Client {
	void ok(Producer p, Consumer c) throws IOException {
		InputStream is = p.get(); // *1*
		InputStream is2 = c.inspect(is); // *2* 
		c.consume(is2);  // *3*
	}
}
```
By commenting line \*3* and then \*2* you will see errors appear on the previous line. By removing the annotation from `Producer.get()` the error will disappear again. Etc. pp.

### DIFF
--- a/org.eclipse.jdt.annotation/src/org/eclipse/jdt/annotation/Owning.java
+++ b/org.eclipse.jdt.annotation/src/org/eclipse/jdt/annotation/Owning.java
@@ -1,0 +1,66 @@
+/*******************************************************************************
+ * Copyright (c) 2023 GK Software SE and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     Stephan Herrmann - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.jdt.annotation;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * This annotation allows to specify responsibility for certain objects as they are passed from one method to another.
+ * <p>
+ * Based on this annotation, flow oriented analysis tools can be made more precise without the need to perform
+ * full-system analysis, as the {@code @Owning} annotation describes a contract, consisting of two sides that can be
+ * checked independently.
+ * </p>
+ * <p>
+ * The Eclipse Compiler for Java&trade; (version 3.37 and greater) interprets this annotation when placed on a value of
+ * type {@link AutoCloseable} or any subtype. This annotation is interpreted specifically in these locations:
+ * </p>
+ * <dl>
+ * <dt>Method parameter ({@link ElementType#PARAMETER}):</dt>
+ * <dd>Here {@link Owning} denotes that responsibility to close a resource passed into this parameter will lie with the
+ * receiving method, on this particular flow.
+ * </dd>
+ * <dt>Method ({@link ElementType#METHOD}) - as a way to refer to the method return value:</dt>
+ * <dd>Here {@link Owning} denotes that responsibility to close a resource received from a call to this method will lie
+ * with the receiving code.
+ * </dd></dd></dt>
+ * <p>Responsibility to close a resource may <strong>initially arise</strong> from these situations:</p>
+ * <ul>
+ * <li>Instantiating a class that is a subtype of {@link AutoCloseable}.</li>
+ * <li>Receiving a method argument via a parameter of type {@link AutoCloseable} (or subtype) that is marked as {@code @Owning}.</li>
+ * <li>Receiving a result from a call to method that is marked as {@code @Owning} and returns {@link AutoCloseable} (or a subtype).
+ * 	<ul><li>If the {@code @Owning} annotation is absent in this situation, the receiving method is <em>potentially responsible</em>.</li></ul>
+ * </li>
+ * <li>
+ * <p>Responsibility to close a resource may be <strong>fulfilled</strong> by ensuring any of these measures on every possible path:</p>
+ * <ul>
+ * <li>Invoking {@link AutoCloseable#close()}.
+ * <li>Passing the resource into a method where the receiving parameter has type {@link AutoCloseable} (or subtype)
+ *  and is marked as {@code @Owning}.</li>
+ * <li>Returning the resource to the caller, provided that the current method is marked as {@code @Owning} and has a declared return type
+ * 	of {@link AutoCloseable} (or a subtype).</li>
+ * </ul>
+ *
+ * @since 2.3
+ */
+@Retention(RetentionPolicy.CLASS)
+@Documented
+@Target({ ElementType.PARAMETER, ElementType.METHOD })
+public @interface Owning {
+	// no details
+}
--- a/org.eclipse.jdt.annotation/src/org/eclipse/jdt/annotation/Owning.java
+++ b/org.eclipse.jdt.annotation/src/org/eclipse/jdt/annotation/Owning.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2023 GK Software SE and others.
+ * Copyright (c) 2024 GK Software SE and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -32,35 +32,46 @@ import java.lang.annotation.Target;
  * </p>
  * <dl>
  * <dt>Method parameter ({@link ElementType#PARAMETER}):</dt>
- * <dd>Here {@link Owning} denotes that responsibility to close a resource passed into this parameter will lie with the
+ * <dd>Here {@code @Owning} denotes that the responsibility to close a resource passed into this parameter will lie with the
  * receiving method, on this particular flow.
  * </dd>
  * <dt>Method ({@link ElementType#METHOD}) - as a way to refer to the method return value:</dt>
- * <dd>Here {@link Owning} denotes that responsibility to close a resource received from a call to this method will lie
+ * <dd>Here {@code @Owning} denotes that responsibility to close a resource received from a call to this method will lie
  * with the receiving code.
- * </dd></dd></dt>
- * <p>Responsibility to close a resource may <strong>initially arise</strong> from these situations:</p>
+ * </dd>
+ * <dt>Field ({@link ElementType#FIELD})</dt>
+ * <dd>This annotation marks that the lifecycle of a resource stored in the field is tied to the lifecycle of the
+ *  enclosing object. In order to allow for precise analysis in this situation, it is recommended that the enclosing
+ *  class of such a field implements {@link AutoCloseable}. In that case, it will be the responsibility of the class's
+ *  {@code close()} implementation to also close all resources stored in {@code @Owning} fields.
+ * </dl>
+ * <p><strong>Responsibility</strong> to close a resource may <strong>initially arise</strong> from these situations:</p>
  * <ul>
  * <li>Instantiating a class that is a subtype of {@link AutoCloseable}.</li>
  * <li>Receiving a method argument via a parameter of type {@link AutoCloseable} (or subtype) that is marked as {@code @Owning}.</li>
  * <li>Receiving a result from a call to method that is marked as {@code @Owning} and returns {@link AutoCloseable} (or a subtype).
  * 	<ul><li>If the {@code @Owning} annotation is absent in this situation, the receiving method is <em>potentially responsible</em>.</li></ul>
  * </li>
- * <li>
- * <p>Responsibility to close a resource may be <strong>fulfilled</strong> by ensuring any of these measures on every possible path:</p>
+ * <li>Within the {@code close()} method of a class implementing {@link AutoCloseable}, each resource field annotations as {@code @Owning}
+ * must be closed.</li>
+ * </ul>
+ * <p><strong>Responsibility</strong> to close a resource may be <strong>fulfilled</strong> by ensuring any of these measures on every possible path:</p>
  * <ul>
  * <li>Invoking {@link AutoCloseable#close()}.
  * <li>Passing the resource into a method where the receiving parameter has type {@link AutoCloseable} (or subtype)
  *  and is marked as {@code @Owning}.</li>
  * <li>Returning the resource to the caller, provided that the current method is marked as {@code @Owning} and has a declared return type
  * 	of {@link AutoCloseable} (or a subtype).</li>
+ * <li>Assigning the resource to a field annotated as {@code @Owning}.
+ * <li>Within the {@code close()} method of a class implementing {@link AutoCloseable} (see above) closing the resource held by a field
+ *  tagged as {@code @Owning} can happen either directly, or for inherited fields by invoking {@code super.close()}.
  * </ul>
  *
  * @since 2.3
  */
 @Retention(RetentionPolicy.CLASS)
 @Documented
-@Target({ ElementType.PARAMETER, ElementType.METHOD })
+@Target({ ElementType.PARAMETER, ElementType.METHOD, ElementType.FIELD })
 public @interface Owning {
 	// no details
 }
--- a/org.eclipse.jdt.annotation/src/org/eclipse/jdt/annotation/Owning.java
+++ b/org.eclipse.jdt.annotation/src/org/eclipse/jdt/annotation/Owning.java
@@ -44,6 +44,10 @@ import java.lang.annotation.Target;
  *  enclosing object. In order to allow for precise analysis in this situation, it is recommended that the enclosing
  *  class of such a field implements {@link AutoCloseable}. In that case, it will be the responsibility of the class's
  *  {@code close()} implementation to also close all resources stored in {@code @Owning} fields.
+ * <dt>Method receiver (via {@link ElementType#TYPE_USE})</dt>
+ * <dd>Annotating a method receiver (explicit {@code this} parameter) as owning signals that the method is responsible
+ *  for closing all contained resources. The method will be treated as a "custom close method".<br/>
+ *  The annotation is not evaluated in any other TYPE_USE locations.
  * </dl>
  * <p><strong>Responsibility</strong> to close a resource may <strong>initially arise</strong> from these situations:</p>
  * <ul>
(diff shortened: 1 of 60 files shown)


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/MethodBinding.java
```diff
@@ -73,6 +73,8 @@ public class MethodBinding extends Binding {
 	public Boolean[] parameterNonNullness;  // TRUE means @NonNull declared, FALSE means @Nullable declared, null means nothing declared
 	public int defaultNullness; // for null *type* annotations
 
+	public Boolean[] parameterOwning; // TRUE means @Owning declared, ...
```
szarnekow, line 76: I can understand why `parameterNonNullness` benefits from the boxed Boolean. For `parameterOwning` I don't see why we cannot use `boolean[]`

Does it make sense to bit-encode the information of both arrays into a single long[]?

org.eclipse.jdt.annotation/src/org/eclipse/jdt/annotation/Owning.java
```diff
@@ -0,0 +1,66 @@
...
+ *  and is marked as {@code @Owning}.</li>
+ * <li>Returning the resource to the caller, provided that the current method is marked as {@code @Owning} and has a declared return type
+ * 	of {@link AutoCloseable} (or a subtype).</li>
+ * </ul>
+ *
+ * @since 2.3
+ */
+@Retention(RetentionPolicy.CLASS)
+@Documented
+@Target({ ElementType.PARAMETER, ElementType.METHOD })
```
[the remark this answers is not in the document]
stephan-herrmann: Thanks, this looks like a valuable extension of the concept.

Would you expect that a class with an `@Owning` field implements `AutoCloseable` itself, as to declare `close()` in a canonical way? If so, this would develop into a generalization of what currently we whitelist as "wrapper closeables" (like Buffered\*). Not sure if the implementation could leverage this generalization, but it would ensure conceptual uniformity.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/MethodBinding.java
stephan-herrmann: I haven't yet dismissed the idea to add the dual annotation `@NotOwning`, in particular because having that additional option would allow us to make `@Owning` the default for method returns.

But while consolidating the implementation, merging all this into a array of bitsets makes perfect sense. Assuming that you didn't only consider beauty of the code, but at least partially performance, too, I went one step further to choose `byte[]` for now. This means we cannot directly use `TagBits.AnnotationOwning` and friends as index, but we are converting between various formats anyway, and conversions look pretty succinct to me.

stephan-herrmann:
> @stephan-herrmann [+ introduce the dual annotation @NotOwning](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1716/commits/9fcd205c3bfa0e6598dbc983212a3064864ef711)
[9fcd205](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1716/commits/9fcd205c3bfa0e6598dbc983212a3064864ef711)

This commit is mostly boring infra structure to detect the dual annotation `@NotOwning`. Not much logic yet.

stephan-herrmann:
> @stephan-herrmann [Feature work:](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1716/commits/0e49d67ac8b0994cbd2b2bd98d9ccb1cce50e0b9)  [0e49d67](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1716/commits/0e49d67ac8b0994cbd2b2bd98d9ccb1cce50e0b9)

This is a major update. 

**User visible changes** include:
+ evaluate `@NotOwning`
+ include `@Owning` on fields
  + enclosing class should implement `AutoCloseable`
  + assignment to such field delegates the responsibility to the `close()` method
  + `close()` method is checked for closing all `@Owning` fields, also considering super calls.
  + simple pattern `if (this.resource != null) this.resource.close()` is understood, implicitly using the machinery of syntactic null analysis for fields
+ JavaCore now lists the new problem options with corresponding javadoc
+ generally improved precision of the analysis particularly in try-finally contexts
+ when resource annotations are enabled most of the existing "heuristic guesswork" for data flows across method boundaries is replaced by more precise analysis based on the new annotations (i.e., analysis operates in two distinct modes)

Behind the scenes some **ad-hoc coding has been generalized & unified.**

Additionally, a **new test** class has been added, which *inherits from ResourceLeakTests*, so that all existing tests can be re-run in the new, annotation-based mode. Only a few test expectations are overridden in the new mode.

Some infra-structure work was needed to **run test methods from the super class**, and since these changes where made in the `suite()` method, I had to ensure that `AbstractCompilerTest` actually uses the `suite()` method during `buildComplianceTestSuite()`. It looks like all existing test classes have suitable `suite()` methods, so nothing seems to have changed for those. @jarthana @mpalat @srikanth-sankaran  fyi.

More detailed description can be found in the commit message.

stephan-herrmann:
Remaining TODOs:
* test compiling Eclipse SDK with this option enabled
* add a few more warnings for unsuitable use of annotations (e.g., when type is not AutoCloseable)
* decide if methods other than `close()` should be able to take responsibility for closing `@Owning` fields

For the last bullet I see two options:

(1) Add a new annotation like "`@WillCloseAllOwnedResourceFields`" - name tbd.

(2) Admit `@Owning` in one more location:

```
    public void processAllResources1(@Owning MyClass this) { // <<<--- here
        // may (of course) read from resource:
        this.resource.read();
        // must close them:
        this.resource.close(); // should this trigger suggestion to use t-w-r?
    }
```

This syntax `@Owning MyClass this` is less-known but it would allow us to get away with fewer annotations, and precisely expresses what we need: the method annotating `this` with `@Owning` will be responsible for closing everything owned by `this`.

Instead of directly closing resource fields, such a method can of course simply delegate to a `close()` method.

Should it be OK to *only* provide consuming methods with `@Owning MyClass this` and forgo implementing `AutoCloseable`?

szarnekow:
IIUIC, option (2) `@Owning MyClass this` is generalizable enough to assume that `AutoClosable.close` can be considered a special case with dedicated compiler support. Still, apart from that, is it semantically the same as mandating the `@Owning` annotation on `AutoClosable.close`, too?

stephan-herrmann:
> IIUIC, option (2) `@Owning MyClass this` is generalizable enough to assume that `AutoClosable.close` can be considered a special case with dedicated compiler support.

I agree. Indeed one of my goals is to empower users to add more classes and methods to existing dedicated compiler support.

For a start the compiler would just implicitly regard `AutoCloseable.close()` as if it were declared with `@Owning AutoCloseable this` in the parameter list.

> Still, apart from that, is it semantically the same as mandating the `@Owning` annotation on `AutoClosable.close`, too?

I'm not quite sure what you are asking here. If this is about `@Owning AutoCloseable this`, you have my answer above :) - but one detail is still missing: the compiler needs to ensure that every override adheres to the contract of the super, i.e., generally overriding a method with `@Owning` somewhere in the parameter list obliges the override to repeat that annotation. (Conversely a `@NotOwning` on the method return should prevent overrides from sneaking in a new `@Owning` obligation - plain contravariance that is).

The next thing to (easily) generalize is the concept of "resource-free closeables", like `StringReader`: We could assume that `AutoCloseable` is implicitly annotated with `@Owning` **on the interface declaration**. Now we could say, that this annotation is normally inherited (all implementors of `AutoCloseable` *need* closing), but our existing white-list could be mapped to adding `@NotOwning` on any subtype that should be exempted from leak analysis:
```
public @NotOwning class StringReader extends Reader { ...
```

Even the concept of wrapper closeables could be mapped to annotations like this:
```
    @Owning public class BufferedInputStream extends FilteredInputStream {
            public BufferedInputStream(@Owning InputStream in) { ...
```
And voila: passing an `InputStream` to that constructor means the client no longer has to worry about closing the `InputStream`, but instead they are now responsible for closing the wrapper `BufferedInputStream`.

I will add observations from toying with this analysis on a full SDK-workspace, shortly. Spoiler: I believe we want integration with .eea *really soon*.

stephan-herrmann:
Globally enabling the new analysis on an Eclipse SDK workspace and sprinkling a few annotations, let's me discover these problems:

Definite resource leaks:
* platform.ant:
  * [AbstractAntUITest.getReaderContentAsString()](https://github.com/eclipse-platform/eclipse.platform/blob/9aa3ddbffcdf831d4e014f290bd14d056dc67bec/ant/org.eclipse.ant.tests.ui/test%20plugin/org/eclipse/ant/tests/ui/testplugin/AbstractAntUITest.java#L246): `bufferedReader` (should actually be `@Owning BufferedReader bufferedReader`).
  * [OptionTests.testXmlLoggerListener()](https://github.com/eclipse-platform/eclipse.platform/blob/9aa3ddbffcdf831d4e014f290bd14d056dc67bec/ant/org.eclipse.ant.tests.core/tests/org/eclipse/ant/tests/core/tests/OptionTests.java#L237): `stream`
  * [InternalAntRunner.getAntVersionNumber()](https://github.com/eclipse-platform/eclipse.platform/blob/9aa3ddbffcdf831d4e014f290bd14d056dc67bec/ant/org.eclipse.ant.launching/remote/org/eclipse/ant/internal/launching/remote/InternalAntRunner.java#L676): `in` -- `Properties.load()` explicitly states that it does not close the stream (we should see the signature as `load(@NotOwning InputStream inputStream)`).
* equinox:
  * [ExtensionRegistry.addContribution()](https://github.com/eclipse-equinox/equinox/blob/5ba80301a491b414cd3297eda0d811e0fa7ed9b9/bundles/org.eclipse.equinox.registry/src/org/eclipse/core/internal/registry/ExtensionRegistry.java#L1139): `is` is closed in a finally block below, but here it may escape due to `throw` outside the try-finally structure.

Understanding exceptional flows remains tricky:
* [AbstractAntUITest.getStreamContentAsString()](https://github.com/eclipse-platform/eclipse.platform/blob/9aa3ddbffcdf831d4e014f290bd14d056dc67bec/ant/org.eclipse.ant.tests.ui/test%20plugin/org/eclipse/ant/tests/ui/testplugin/AbstractAntUITest.java#L234): reports that at this return `reader` is not closed.
  * Strictly speaking we should assume that `reader` remains null when the exception occurs
  * OTOH the error message implies that actually `inputStream` is not closed, which is true (I marked `inputStream` as `@Owning`, because the only caller does not call `close()`).
* [CheatSheetSaveHelper.readXMLFile(URL)](https://github.com/eclipse-platform/eclipse.platform/blob/9aa3ddbffcdf831d4e014f290bd14d056dc67bec/ua/org.eclipse.ui.cheatsheets/src/org/eclipse/ui/internal/cheatsheets/data/CheatSheetSaveHelper.java#L165)
  * here analysis correctly sees that return within the catch block is OK
  * but then we complain at `if (source==null) return null;` -- it requires some indirect reasoning that we don't have a resource at this location - it's the category of "correlation analysis" which ecj is unable of, also during null analysis
  * still a t-w-r with conditional initialization would be cleaner, I guess :)

Unnecessary code:
* p2:
  * [DiscoveryRegistryStrategy.processBundle(IExtensionRegistry, Entry, File)](https://github.com/eclipse-equinox/p2/blob/dc136d37b54601d3cbaade68504a85718ee2d676/bundles/org.eclipse.equinox.p2.discovery.compatibility/src/org/eclipse/equinox/internal/p2/discovery/compatibility/DiscoveryRegistryStrategy.java#L85C20-L85C20)
    * we see a cautious try-with-resources, but:
      * `addContribution()` should say `@Owning InputStream is` (with the exception bug mentioned above), after which callers are no longer responsible
    * the same in `processDiscoveryCoreBundle()`

Design questions
 * Class org.w3c.css.sac.InputSource is not nice to analysis. Holding closable fields, but not being AutoCloseable, so: what should be the contract??
 * usage of `ZipFile.getInputStream(ZipEntry)` could be made safe in different ways:
   * it would be OKish to mark this method as `@NotOwning` because closing the zipfile will suffice to also close the dependent stream
   * OTOH, some clients think it safer to immediately close the input stream regardless of the zip file. Those clients would be better helped with `@Owning` on `getInputStream(ZipEntry)`.
* ZipLeveledStructureProvider: this class could well be declared as a "resource wrapper" (mark field and contructor arg as owning)
    * unfortunately, `close()` is implemented in a way, which the analysis cannot understand:
      * delegation to `closeArchive()` could be address using a `@Owning ZipLeveledStructureProvider this` parameter
      * but then reading the field via method `getZipFile()` is just one too many levels of (unnecessary) indirection.
      * replacing `getZipFile()` with direct field read makes this class analyzable as desired.
   * still `getContents(Object)` is ambiguous just like `ZipFile.getInputStream()`.
* long-lived resources may be difficult to annotate
  * org.eclipse.ui.testing.dumps.TimeoutDumpTimer.StreamForwarder ??
  * org.eclipse.ui.internal.browser.browsers.StreamConsumer ??
 * I couldn't quite make heads or tails of `LazyReader` specifically when cast from an `ILazySource` which is not AutoCloseable.

Library classes lacking annotations
* I see more classes that are actually wrapper resources (which the compiler could easily learn using .eea)
* `java.nio.file.FileSystems.getDefault()` creates no responsibility for closing
* when subclassing a library class known as wrapper resource (hardcoded), we should imply that `super(rc)` in a constructor will take care of the passed resources (i.e., join into the wrapping contract). Example: `TarOutputStream.TarOutputStream(OutputStream)`.
* System.setIn(), System.setOut(), System.setErr() should have owning parameters
* `Process.getInputStream()` (_Output_, _Error_) should probably return notowning streams

Standard warnings:
* I see lots of places where the tool recommends to
  * mark method returns as Owning
  * mark resource fields as Owning
    * but then not all classes with resource fields are AutoCloseable ...

Nice: single annotations helping lots of location for precise analysis:
* `closeZipFile(@Owning ZipFile file, Shell shell)`
* `IFile.setContents(@Owning InputStream source...)`

 
Plus a few possible bugs in my implementation.

Bottom line:
* new warnings/errors do make sense
* annotations give plenty of options to make the design explicit in a way that can be analyzed
* for full blown usage .eea would be very cool
* annotating a `this` parameter will also help in existing code.
* generally, experimentally placing a few annotations greatly helped understanding the underlying design
  * if annotations don't match the design, plenty of new errors/warnings will pop up :)

stephan-herrmann:
> when subclassing a library class known as wrapper resource (hardcoded), we should imply that super(rc) in a constructor will take care of the passed resources (i.e., join into the wrapping contract).

Done in [a98d151](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1716/commits/a98d1510f3034490fd847d6d6987cda168c73745)

stephan-herrmann:
> (2) Admit `@Owning` in one more location:

Done in [384b83c](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1716/commits/384b83c8c9f18237b6fd4576d7a24e30c0b648b5) see [testConsumingMethod_nok()](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1716/commits/384b83c8c9f18237b6fd4576d7a24e30c0b648b5#diff-7db07bbabae06eff59ad5dab81081499549376d2e21b3c866a4623b210bebdc7) onward

stephan-herrmann:
In preparation of a merge I'm looking for any left-overs:

* I mentioned a change in `AbstractCompilerTest`, which, however, broke the build on jenkins. A better fix fine tunes `buildComplianceTestSuite()` by reflecting on a static field `INHERITED_DEPTH` if present.
* I already outsourced some follow-up:
    *  #1849 (should)
    *  #1848 (might).
* Field tests in a full Eclipse SDK workspace gave good results (after a few iterations :) )
* I don't (yet) support resource classes annotated `@Owning` but not implementing `AutoCloseable` - no plans currently, I don't feel like supporting such style, since it defeats the purpose of try-with-resources (where all this started)
* I haven't yet filed an issue for integration with .eea

iloveeclipse:
@stephan-herrmann : FYI, I had to fix javadoc you've added to fix SDK build (that failed on javadoc errors), please double check if the final result is what you've meant https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1852

jukzi:
@stephan-herrmann we have now some more warnings in AbstractCommentParser, i guess they are related to this change
image

Issue #1720, opened 2023-12-10
Regression: 4.31 no longer supports standalone JavaCore.getOptions()


ewillink:
Version: 2024-03 (4.31)
Build id: I20231207-1800

org.eclipse.emf.codegen.ecore.genmodel.util.GenModelUtil uses JavaCore.getOptions() when there is no running platform. This makes use of the JavaModelManager whose constructor still has Platform.isRunning() conditionalization demonstrating that this is (still) a valid use case. Unfortunately the initialization of touchJob has moved from a lazy guarded call from touchProjects() to an eager unguarded constructor initialization.

Fix. Revert the unjustified eager initailization and add a JUnit test demonstrating that JavaCore.getOptions() works standalone. (This is the second time I have had to report a standalone regression.)

iloveeclipse:
> and add a JUnit test demonstrating that JavaCore.getOptions() works standalone.
> (This is the second time I have had to report a standalone regression.)

@ewillink : if your use case is important, consider to contribute this unit test, especially you've hit that already before. Could you please add a reference to the original bug report?

> org.eclipse.emf.codegen.ecore.genmodel.util.GenModelUtil uses JavaCore.getOptions() when there is no running platform.

Could you please add a link to the `GenModelUtil` source, so one can see how the code is supposed to be used in that case?

> Unfortunately the initialization of touchJob has moved from a lazy guarded call from touchProjects() to an eager unguarded constructor initialization.

This is a regression from 948410121f536540c27ef777540740cf5ac7df05. Javadoc from `JavaCore.getOptions()` says it is also supposed to be used without running platform. 

@jukzi : could you please fix that? Note:  I believe `WorkspaceJob touchJob` field should be not only lazy, but change the type to `Job touchJob` to avoid early init of resources bundle, but I haven't checked that.

ewillink:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=540122

A smoke test, or possibly a single property access is a really trivial test. Except that perhaps JDT has no standalone test support so there is a non-trivial releng activity revamping the pom.xml to support both maven and tycho surefire. I have no familiarity with JDT's releng and consequently any revamping by me wil almost certainly not be done in an acceptable fashion.

Sorry. It's your code. I have more than enough to do on my code. When my users embarrass me, I avoid my future embarrassment by writing regression tests myself.

  /**
   * @since 2.9,
   */
[...]

PR #1724, opened 2023-12-11
JavaModelManager: lazy initialize TouchJob #1720

jukzi:
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1720


iloveeclipse:
Jörg, I wonder if you could add a "simple" (usual) test that would run a trivial Java program accessing JavaCore.getOptions() by using current bundle jar in classpath? The test would just read the output and expect no "crash".

jukzi:
> Jörg, I wonder if you could add a "simple" (usual) test that would run a trivial Java program accessing JavaCore.getOptions() by using current bundle jar in classpath? The test would just read the output and expect no "crash".

Would be nice, but i don't want to invest time in standalone scenario and don't know how todo so without investigation.

PR #1748, opened 2023-12-14
Performance : getMainTypeName in CompilationUnitDeclaration is called very often Fix #1747

### LINKED ISSUE #1747: Performance : getMainTypeName in CompilationUnitDeclaration is called very often
getMainTypeName in CompilationUnitDeclaration is called very often
add a cash to increase performance

image

RoiSoleil:
add a cash to increase performance

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/CompilationResult.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/CompilationResult.java
@@ -64,7 +64,7 @@ public class CompilationResult {
 	public CategorizedProblem tasks[];
 	public int problemCount;
 	public int taskCount;
-	public ICompilationUnit compilationUnit;
+	public final ICompilationUnit compilationUnit;
 	private Map<CategorizedProblem, ReferenceContext> problemsMap;
 	private Set firstErrors;
 	private final int maxProblemPerUnit;
@@ -98,6 +98,7 @@ public class CompilationResult {
 
 public CompilationResult(char[] fileName, int unitIndex, int totalUnitsKnown, int maxProblemPerUnit){
 	this.fileName = fileName;
+	this.compilationUnit = null;
 	this.unitIndex = unitIndex;
 	this.totalUnitsKnown = totalUnitsKnown;
 	this.maxProblemPerUnit = maxProblemPerUnit;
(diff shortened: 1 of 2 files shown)


jukzi (dismissed):
When you want to cache results you have to
1. declare all used values final
2. show a benchmark how much time is saved
3. show how much memory it uses
then justify 3. vs 2.

RoiSoleil:
1. declare all used values final ==> Done
2. show a benchmark how much time is saved ==> [getMainTypeName_#1747.xlsx](https://github.com/eclipse-jdt/eclipse.jdt.core/files/13685455/getMainTypeName_.1747.xlsx)
Gain of 4% in time
Gain of 18% in standard deviation due to less GC

4. show how much memory it uses
then justify 3. vs 2. ==> I think that the cost of caching the mainTypeName is really negligible compare to the cost of recomputing it every type (increasing the number of GC you can see it in the Excel)

jukzi:
> 2\. getMainTypeName_#1747.xlsx

thanks, please describe what you measured and what are the units.

> show how much memory it uses

missing. take for example heapdumps from open workspace after code was executed

RoiSoleil:
> > 2. getMainTypeName_#1747.xlsx
> 
> thanks, please describe what you measured and what are the units.
This a time to compile one of my project in milliseconds. I did 20 loop on the same VM  one time with the modification and one time without and with 10 build of the project for each loop. 
> 
> > show how much memory it uses
> 
> missing. take for example heapdumps from open workspace after code was executed

The heapdump is 800MB. How can you analyse it ? What do you want to see ?

jukzi:
> The heapdump is 800MB. How can you analyse it ? What do you want to see ?

with for example VisualVM you could show the number of instances of `CompilationUnitDeclaration`. that count times average name length of 15 times 2 (for char) should give you the additional memory in bytes.

jukzi:
according to your screenshot in the issue a better (less memory) solution would be to avoid getMainTypeName() in isModuleInfo(). The actual name does not matter. you could cheaper cache the boolean isModuleInfo or isParsingModuleDeclaration() or even isInModuleDeclaration()
image

jukzi:
@RoiSoleil i guess we have found a better solution with https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1831
Please test if that solves your performance problem. That should render this PR needless.

RoiSoleil:
That's good for me.

jukzi:
Thanks for pointing out the problem!!

Issue #1743, opened 2023-12-13
OOME during rename


jukzi:
Today i tried to refactor a class by renaming the class name. It ran in an (catched) OOME
![image](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/51790620/eefb03a5-bc92-40ae-8c45-d62c7905e20c)
Two observations: 
1. There where many duplicate char[] in memory:
![image](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/51790620/ab8b726a-6247-4c6d-ba7d-e5d91107f4fe)
For example for "eclipse" in `ast.*.tokens` and `lookup.*.compoundName`:
![image](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/51790620/224ef2f8-28ac-40c4-a5a5-c026aa8e9709)

2. There where thousands of strong references to the source of .java files in org.eclipse.jdt.internal.core.util.ResourceCompilationUnit.contents:
![image](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/51790620/062f5d15-23b1-4013-8101-925b7f59dc40)
![image](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/51790620/30504a3d-28a4-41a9-8800-ce8617c4c577)

jukzi:
can be reproduced by renaming a class in Platform workspace to IPath

PR #1749, opened 2023-12-14
deduplicate "eclipse" #1743

jukzi:
CharDeduplication was not designed to deduplicate tokens with length 7+ which could lead to high memory consumption. With this change tokens of all sizes can be deduplicated.

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1743

A benchmark implemented in CharDeduplicationTest.main(String[]) shows the new deduplication is performed at similar speed (.21s instead of .16s) but deduplicates much more tokens (99% instead of 36%).

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/CharDeduplication.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/CharDeduplication.java
@@ -23,19 +23,17 @@ public class CharDeduplication {
 
 	// ----- immutable static part (thread safe): ----
 
+	private final static char[] CHAR_ARRAY0 = new char[0];
 	static final char[] ASCII_CHARS[] = new char[128][];
 	static {
 		for (int i = 0; i < ASCII_CHARS.length; i++) {
 			ASCII_CHARS[i] = new char[] { (char) i };
 		}
 	}
-	public static final int TABLE_SIZE = 30; // XXX thats not a prime -> bad for hashing, nor a power of 2 -> expensive
-												// modulo computation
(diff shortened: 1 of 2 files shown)


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/CharDeduplication.java
```diff
@@ -118,167 +92,60 @@ public void reset() {
...
-		char c0, c1 = src[start + 1], c2, c3 = src[start + 3], c4, c5 = src[start + 5];
-		int hash = (((c0 = src[start]) << 12) + ((c2 = src[start + 2]) << 6) + (c4 = src[start + 4])) % TABLE_SIZE;
-		char[][] table = this.charArray_length[4][hash];
-		int i = this.newEntry6;
-		while (++i < INTERNAL_TABLE_SIZE) {
-			char[] charArray = table[i];
-			if ((c0 == charArray[0]) && (c1 == charArray[1]) && (c2 == charArray[2]) && (c3 == charArray[3])
-					&& (c4 == charArray[4]) && (c5 == charArray[5]))
-				return charArray;
+	private boolean equals(char[] source, int from, int to, char[] charArray) {
```
szarnekow, line 140: Small thing: I suggest to use `java.util.Arrays.equals(char[], int, int, char[], int, int)` instead. It delegates to the vectorized ArraySupport.mismatch.

jukzi: i did not know that function, but it's slower - probably because tokens are typically small and i use several conditions, that don't need to be checked.
try
`Arrays.equals(source, from, to, charArray, 0,  charArray.length)`
with CharDeduplicationTest

PR #1754, opened 2023-12-16
Test uniqueness of problem IDs

stephan-herrmann:
## What it does

Whenever we add a new constant to `IProblem` it's getting harder by the day, to find a definitely unused ID (since new IDs are allocated in different ranges, and constants are not strictly sorted).

So rather than breaking clients by renumbering, I simply wrote a reflective test that ensures that any new constant doesn't share the ID with an existing one.

This was first requested in https://bugs.eclipse.org/bugs/show_bug.cgi?id=509643.

Initially, my test did detect some existing duplicates, where it turned out, that one in each pair was `@deprecated` (in javadoc). In order to make this visible via reflection, I'm also adding the `@Deprecated` annotation (which was not available in JDT/Core at the time of deprecating pre-1.5 problems).

## How to test
Add a duplicate to see the test fail :)

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/core/compiler/IProblem.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/core/compiler/IProblem.java
@@ -835,25 +835,35 @@ void setSourceStart(int sourceStart);
 	int IllegalModifierCombinationForType = TypeRelated + 334;
 
 	/** @deprecated - problem is no longer generated, use {@link #UndefinedType} instead */
+	@Deprecated
 	int SuperclassNotFound =  TypeRelated + 329 + ProblemReasons.NotFound; // TypeRelated + 330
 	/** @deprecated - problem is no longer generated, use {@link #NotVisibleType} instead */
+	@Deprecated
 	int SuperclassNotVisible =  TypeRelated + 329 + ProblemReasons.NotVisible; // TypeRelated + 331
 	/** @deprecated - problem is no longer generated, use {@link #AmbiguousType} instead */
+	@Deprecated
 	int SuperclassAmbiguous =  TypeRelated + 329 + ProblemReasons.Ambiguous; // TypeRelated + 332
 	/** @deprecated - problem is no longer generated, use {@link #InternalTypeNameProvided} instead */
+	@Deprecated
 	int SuperclassInternalNameProvided =  TypeRelated + 329 + ProblemReasons.InternalNameProvided; // TypeRelated + 333
 	/** @deprecated - problem is no longer generated, use {@link #InheritedTypeHidesEnclosingName} instead */
+	@Deprecated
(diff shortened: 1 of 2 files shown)


stephan-herrmann:
@srikanth-sankaran @jarthana @iloveeclipse  FYI.

In particular let me know if you see problems with:
* adding `@Deprecated` to many constants in IProblem (previously tagged only in javadoc)
* excluding these deprecated fields from the analysis.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/core/compiler/IProblem.java
```diff
@@ -829,25 +829,35 @@ public interface IProblem {
 	int AbstractMethodsInConcreteClass = TypeRelated + 333;
 
 	/** @deprecated - problem is no longer generated, use {@link #UndefinedType} instead */
+	@Deprecated
```
jukzi, line 832: when those constants are not used anymore, can you please add "forRemoval"?
like
`@Deprecated(forRemoval = true, since = "3.37.0 2024-03")`

stephan-herrmann: The deprecation (in javadoc) is actually much older (older than the annotation type? likely older than "forRemoval" :) ) - should we still use the current version in since?

stephan-herrmann: > the since refers to the forRemoval

No, 
> [String](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/String.html) since
Returns the version in which the annotated element became deprecated.

And
> boolean forRemoval
Indicates whether the annotated element is subject to removal in a future version.

There is no direct connection between both properties.

stephan-herrmann: @jukzi once more I scanned the constants in question, and found that those have been deprecated some time in the range 2005 - 2011. So proper marking would require to look up the exact JDT release for each of those constants to provide a proper `since` value. Setting `forRemoval` is a separate strategic decision altogether.

In this PR I only want to make the existing deprecations (javadoc) visible during a test (using the annotation). Can't we think of this as orthogonal to the issue of improving the details of deprecation? This PR should serve only one purpose: avoid ambiguous IDs of "active" problems, ignoring those that are no longer raised.

jukzi: My desire is to actually have unused code to be removed in the future. That should only be done 2 years after announcing the removal. For that purpose it does not matter when exactly deprecation happened, but it should have the forRemoval flag 2 years in front. If you do not feel confident about adding a rough date it is fine if you just ignore my request to add a date (or the forRemoval at all) and leave that burden for someone else in the future.  In doubt we just live with the unused constants forever. Wont be of any problem. 
Still the easiest and widely accepted solution (yes, slightly dirty, but most helpful) would be to temporarily just use the current's release version/date and delete it after 2 years.

PR #1761, opened 2023-12-18
Pattern variable is not recognized in AND_AND_Expression

### LINKED ISSUE #1759: Pattern variable is not recognized in AND_AND_Expression
The related issue - https://github.com/eclipse-jdtls/eclipse.jdt.ls/issues/2992
Steps to reproduce:

- create the following class
```
import java.util.ArrayList;
import java.util.List;
public class X {
    public void test() {
        Object obj = "test"; 
        List<String> values = new ArrayList<>();
        if (obj instanceof String str && values.stream().anyMatch(str::equalsIgnoreCase)) {
            System.out.println(str);
        }
    }
}
```
- JDT returns
```
The method anyMatch(java.util.function.Predicate<? super java.lang.String>) in the type Stream<String> is not applicable for the arguments (java.util.function.Predicate<? super java.lang.String>)
```
The following code works fine
```
import java.util.ArrayList;
import java.util.List;
public class X {
    public void test() {
        Object obj = "test"; 
        List<String> values = new ArrayList<>();
        if (obj instanceof String str) {
        	if (values.stream().anyMatch(str::equalsIgnoreCase)) {
        		System.out.println(str);
        	}
        }
    }
}
```

snjeza:
Fixes #1759 

str is not recongized in
 `if (obj instanceof String str && values.stream().anyMatch(str::equalsIgnoreCase))`

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ReferenceExpression.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ReferenceExpression.java
@@ -998,7 +998,20 @@ public class ReferenceExpression extends FunctionalExpression implements IPolyEx
 			}
 			copy.setExpressionContext(this.expressionContext);
 			copy.setExpectedType(targetType);
+			// https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1759
+			LocalVariableBinding[] locals = this.enclosingScope.locals;
+			if (locals != null) {
+				LocalVariableBinding[] copyLocals = new LocalVariableBinding[locals.length];
+				System.arraycopy(locals, 0, copyLocals, 0, locals.length);
+				this.enclosingScope.locals = copyLocals;
+				for (LocalVariableBinding local: copyLocals) {
+					if (local != null) {
+						local.modifiers = local.modifiers & (1 << ExtraCompilerModifiers.AccPatternVariable);
+					}
+				}
+			}
 			copy.resolveType(this.enclosingScope);
+			this.enclosingScope.locals = locals;
 
 			if (this.copiesPerTargetType == null)
 				this.copiesPerTargetType = new HashMap<>();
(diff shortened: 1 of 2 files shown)


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ReferenceExpression.java
```diff
@@ -998,7 +998,20 @@ private ReferenceExpression cachedResolvedCopy(TypeBinding targetType) {
...
+			// https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1759
+			LocalVariableBinding[] locals = this.enclosingScope.locals;
+			if (locals != null) {
+				LocalVariableBinding[] copyLocals = new LocalVariableBinding[locals.length];
+				System.arraycopy(locals, 0, copyLocals, 0, locals.length);
+				this.enclosingScope.locals = copyLocals;
+				for (LocalVariableBinding local: copyLocals) {
+					if (local != null) {
+						local.modifiers = local.modifiers & (1 << ExtraCompilerModifiers.AccPatternVariable);
+					}
```
srikanth-sankaran, line 1010: @snjeza This seems to be a side effect of unknown nature. What exactly is attempted here ?? Why would you want to adjust the modifiers thus ??

srikanth-sankaran:
I think the real problem is `org.eclipse.jdt.internal.compiler.ast.ReferenceExpression.copy()` not propagating the pattern variables when true/false to the copy from the original

srikanth-sankaran:
I have created an alternate PR here: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1768 - let us see what the test results look like

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ReferenceExpression.java
snjeza: I agree. #1768 is better.

srikanth-sankaran:
Closed in favor of https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1768

PR #1787, opened 2023-12-23
Start position and length of Dimension node lost when cloned

### LINKED ISSUE #1786: Start position and length of Dimension node lost when cloned
We recently raised the api level of the JDT parser we use for the WindowBuilder project from JSL3 to JSL21 and as a result, we're currently having roughly 130 test cases that now fail because of the check `assertTrue(begin >= parent_begin)`:

```
@Override
public void preVisit(ASTNode node) {
	assertSame(editorAST, node.getAST());
	// check positions for parent/child
	{	ASTNode parent = node.getParent();
		if (parent != null) {
			int begin = AstNodeUtils.getSourceBegin(node);
			int end = AstNodeUtils.getSourceEnd(node);
			int parent_begin = AstNodeUtils.getSourceBegin(parent);
			int parent_end = AstNodeUtils.getSourceEnd(parent);
			assertTrue(begin >= parent_begin);
			assertTrue(end <= parent_end);
		}
	}
}
```

Instead of the actual position, both the start position and length of the node are set to `-1`.
Note that this problem only shows up when arrays are used.  For example in this test class:

```
public class Test extends JPanel {
	public Test() {
		addButtons(new JButton[]{});
		JButton button = new JButton();
		addButtons( new JButton[] { button, new JButton() } );
	}

	public void addButtons(JButton[] buttons) {
	}
}
```

The reason it didn't happen previously is because with JSL8, JDT switched from a ComponentType to an ElementType when dealing with arrays. ElementTypes have an additional Dimension attribute, which is what's causing problems now.

Normally, the position and length of the Dimension node is set while parsing. However, this information gets lots when it is cloned via `ASTNode.copySubtree(...)`. So I believe this information simply has to be set when `clone0()` is called.

ptziegler:
## What it does

When a Dimension node is cloned, the start position and length of the node needs to be explicitly set. Otherwise this information is lost.

## How to test

Execute ASTTest#testDimension().

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTTest.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTTest.java
@@ -9575,5 +9575,19 @@ public class ASTTest extends org.eclipse.jdt.core.tests.junit.extension.TestCase
 		AST a = new AST(options);
 		assertEquals("Incorrect ast mapping", a.apiLevel(), AST.getJLSLatest());
 	}
+
+	@SuppressWarnings("deprecation")
+	public void testDimension() {
+		// Introduced with JSL8
+		if (this.ast.apiLevel() < AST.JLS8) {
+			return;
+		}
+		Dimension dimension = this.ast.newDimension();
+		dimension.setSourceRange(1, 2);
+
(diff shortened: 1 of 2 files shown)


ptziegler:
Tested it locally and while there are still 4 test failures, those are unrelated to the Dimension node but rather because we still call some deprecated JDT methods somewhere else.

srikanth-sankaran:
Patch looks good. Thanks @ptziegler

PR #1821, opened 2024-01-08
[performance] less calls to isParsingModuleDeclaration() #1747

jukzi:
keep Scanner.insideModuleInfo in sync with isParsingModuleDeclaration()

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1747

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
@@ -2567,8 +2567,7 @@ public final boolean jumpOverUnicodeWhiteSpace() throws InvalidInputException {
 }
 
 public boolean isInModuleDeclaration() {
-	return this.fakeInModule || this.insideModuleInfo ||
-			(this.activeParser != null ? this.activeParser.isParsingModuleDeclaration() : false);
+	return this.fakeInModule || this.insideModuleInfo;
 }
 protected boolean areRestrictedModuleKeywordsActive() {
 	return this.scanContext != null && this.scanContext != ScanContext.INACTIVE;
@@ -2887,7 +2886,7 @@ private ScanContext getScanContext(int begin) {
 	CompilerOptions options = new CompilerOptions();
 	options.complianceLevel = this.complianceLevel;
 	options.sourceLevel = this.sourceLevel;
-	ScanContextDetector parser = new ScanContextDetector(options);
+	ModuleScanContextDetector parser = new ModuleScanContextDetector(options);
 	return parser.getScanContext(this.source, begin - 1);
 }
 protected final void scanEscapeCharacter() throws InvalidInputException {
@@ -4825,8 +4824,8 @@ private static class VanguardParser extends Parser {
 	}
 }
 
-private class ScanContextDetector extends VanguardParser {
-	ScanContextDetector(CompilerOptions options) {
+private class ModuleScanContextDetector extends VanguardParser {
+	ModuleScanContextDetector(CompilerOptions options) {
 		super(new ProblemReporter(
 					DefaultErrorHandlingPolicies.ignoreAllProblems(),
 					options,
@@ -4862,12 +4861,12 @@ private class ScanContextDetector extends VanguardParser {
 
 	@Override
 	public boolean isParsingModuleDeclaration() {
-		return true;
+		return true; // for Modules only
 	}
 
 	public ScanContext getScanContext(char[] src, int begin) {
 		this.scanner.setSource(src);
-		this.scanner.resetTo(0, begin);
+		this.scanner.resetTo(0, begin, true);
 		goForCompilationUnit();
 		Goal goal = new Goal(TokenNamePLUS_PLUS, null, 0) {
 			@Override


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
```diff
@@ -2567,8 +2567,8 @@ public final boolean jumpOverUnicodeWhiteSpace() throws InvalidInputException {
 }
 
 public boolean isInModuleDeclaration() {
-	return this.fakeInModule || this.insideModuleInfo ||
-			(this.activeParser != null ? this.activeParser.isParsingModuleDeclaration() : false);
+	// this.insideModuleInfo == (this.activeParser != null ? this.activeParser.isParsingModuleDeclaration() : false)
```
iloveeclipse, line 2570: Can you please remove comment here?

iloveeclipse:
@srikanth-sankaran : probably straight forward fix, could you please check?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
```diff
@@ -4861,12 +4860,12 @@ void updateScanContext(int token) {
 
 	@Override
 	public boolean isParsingModuleDeclaration() {
-		return true;
+		return Scanner.this.insideModuleInfo;
```
jukzi, line 4863: Note that the only test where this made a difference is org.eclipse.jdt.core.tests.compiler.regression.SwitchExpressionsYieldTest.testBug577220_001() (illegal java code in a module-info.java)

srikanth-sankaran:
I see two problems: 

(1) A more orthodox fix would be to change 

```
protected final boolean mayBeAtCasePattern(int token) {
	return (!isInModuleDeclaration() && JavaFeature.PATTERN_MATCHING_IN_SWITCH.isSupported(this.complianceLevel, this.previewEnabled))
			&& (token == TokenNamecase || this.multiCaseLabelComma);
}
```

to 

```
protected final boolean mayBeAtCasePattern(int token) {
	return  (token == TokenNamecase || this.multiCaseLabelComma) && (!isInModuleDeclaration() && JavaFeature.PATTERN_MATCHING_IN_SWITCH.isSupported(this.complianceLevel, this.previewEnabled));
}
```

presently `mayBeAtCasePattern` gets called for a majority of scanned tokens. This can be simply fixed by changing the singleton call site from being:

```
} else if (mayBeAtCasePattern(token)) {
		token = disambiguateCasePattern(token, this);
	}
```

to be 

```
} else if ( (token == TokenNamecase || this.multiCaseLabelComma) && mayBeAtCasePattern(token)) {
		token = disambiguateCasePattern(token, this);
	}
```

(and remove the checks for `(token == TokenNamecase || this.multiCaseLabelComma)` from inside `mayBeAtCasePattern`)

See that every other `mayBe*` method call is preceded by a token check - `mayBeAtCasePattern` is an exception.

(2) The check for `!isInModuleDeclaration()` is wrong - It is _not_ illegal for module-info.java to have code. For example this code compiles with javac when stored in module-info.java - it has always been the case that `package-info.java` and `module-info-java` may have code. 

```
class X {
 void main(Integer i) {
   Object a = switch (i) {
   default -> {
     yield i.toString();
   }
   };
 }
}
```

srikanth-sankaran:
Please note : The changes I propose are not tested. But should stand to reason

srikanth-sankaran:
I would be fine with burying our heads in sand on module-info having extraneous code and have this PR just attempt to cut back the calls to `mayBeAtCasePattern` along the proposed lines.

@jukzi WDYT ?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
srikanth-sankaran: Comments provided on the PR

srikanth-sankaran (changes requested):

iloveeclipse:
> It is _not_ illegal for module-info.java to have code.

OMG. Why on earth specs allow this?

srikanth-sankaran:
> > It is _not_ illegal for module-info.java to have code.
> 
> OMG. Why on earth specs allow this?

JLS 7.4.1 has this to say about package-info.java:

```
Typically package-info.java contains only a package declaration, preceded immediately by the
annotations on the package. While the file could technically contain the source code for
one or more classes with package access, it would be very bad form.
```

I am unable to spot a similar wording for module-info, but here is what my experiments show. Javac accepts code in module-info.java if it doesn't actually contain module description. I am unable to combine code **with** module description.

So in summary I would say, the point I raised about module-info checks in `mayBeAtCasePattern` is a super corner case and the code can be left as is wrt to that observation.

Rest of review comments stand

srikanth-sankaran:
> > > It is _not_ illegal for module-info.java to have code.
> > 
> > 
> > OMG. Why on earth specs allow this?
> 
> JLS 7.4.1 has this to say about package-info.java:
> 
> ```
> Typically package-info.java contains only a package declaration, preceded immediately by the
> annotations on the package. While the file could technically contain the source code for
> one or more classes with package access, it would be very bad form.
> ```
> 
> I am unable to spot a similar wording for module-info, but here is what my experiments show. Javac accepts code in module-info.java if it doesn't actually contain module description. I am unable to combine code **with** module description.
> 
> So in summary I would say, the point I raised about module-info checks in `mayBeAtCasePattern` is a super corner case and the code can be left as is wrt to that observation.
> 
> Rest of review comments stand

Related but distinct rant: 

module-info scanning and parsing could have been a lot simpler by design by making all the directives, keywords iff they occur in module-info.java rather than context sensitive. IMO the language designers opted for unnecessary generality. For example, `provides`, `requires`, `opens`, `module`, `uses` etc when they occur in module-info may be just plain identifiers or may be based on context special directives. 

Why is this generality necessary ? module-info is a brand new artifact from Java 9 and we may as well have mandated they are always keywords if they occur in module-info.java.

This is a valid module-info:

```
module module {
   requires requires;
   exports to to exports;
   uses module;
   provides uses with to;
}
```

and so is this:
```
import module.pack1.exports.pack2;
import module.open.pack1.opens.pack2;
@open @module(true)
open module module.module.module {
   requires static transitive requires;
   requires transitive static transitive;
   exports to to exports;
   opens module.to.pack1 to to.exports;
   uses module;
   provides uses with to;
}
```
The same textual token becomes a directive or a plain identifier based on location. 

See https://bugs.eclipse.org/bugs/show_bug.cgi?id=488541 - @stephan-herrmann may have some views about this

jukzi:
@srikanth-sankaran thanks for your detailed review
1) I also thought about reducing calls to isInModuleDeclaration by reordering the other conditions. However that will still call isParsingModuleDeclaration() more then once when the other conditions apply. So my thought was to finally fix the performance problem by using the already exiting attempt to cache it's value and to just fix it. As far as i understand we could also apply both improvements - if my current PR is also correct.
2) this issue is only about the real world performance impact. Please lets separate the functional requirement to a dedicated issue. I personally have no preference if code in package-info.java is supported. To me that is no business-case as in 100% of productive code the filename equals the classname and "-" is not valid part of name. Nevertheless i think that any condition based on the .java's filename sounds wrong at the first place.

srikanth-sankaran:
> 2. this issue is only about the real world performance impact. Please lets separate the functional requirement to a dedicated issue. I personally have no preference if code in package-info.java is supported. To me that is no business-case as in 100% of productive code the filename equals the classname and "-" is not valid part of name. Nevertheless i think that any condition based on the .java's filename sounds wrong at the first place.

Did you see my earlier comments _I would be fine with burying our heads in sand on module-info having extraneous code and have this PR just attempt to cut back the calls to mayBeAtCasePattern along the proposed lines_ and _So in summary I would say, the point I raised about module-info checks in mayBeAtCasePattern is a super corner case and the code can be left as is wrt to that observation_ - I am not asking for a separate dedicated issue. I am saying we can bury our heads in sands and pretend the problem doesn't exist :-)

srikanth-sankaran:
> 1. I also thought about reducing calls to isInModuleDeclaration by reordering the other conditions. However that will still call isParsingModuleDeclaration() more then once when the other conditions apply. So my thought was to finally fix the performance problem by using the already exiting attempt to cache it's value and to just fix it. As far as i understand we could also apply both improvements - if my current PR is also correct.

If what you say here: **the issue is a performance regression from
Scanner.mayBeAtCasePattern(int) calling slow isInModuleDeclaration() since https://github.com/eclipse-jdt/eclipse.jdt.core/commit/cb1a0dd5d12bbdadf948163de0016f3b1956addf** is indeed the case, then the proper fix IMHO is to cut down on the hugely wasteful number of calls to `mayBeAtCasePattern` - This should be called only in tiny numbers but presently it gets called whenever the current token is anything other than `TokenNameLPAREN`, `TokenNameLESS`, `TokenNameAT`, `TokenNameARROW` and `TokenNameELLIPSIS`

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
srikanth-sankaran: So what difference did it make ?? In Jenkins the tests seem to be passing and I don't see any changes to the tests themselves ??

This change and the next in `org.eclipse.jdt.internal.compiler.parser.Scanner.ScanContextDetector.getScanContext(char[], int)` are confusing to me. Why would you replace true with some state in scanner. We get into the method `org.eclipse.jdt.internal.compiler.parser.Scanner.ScanContextDetector.getScanContext(char[], int)` only if we are in module-info. If you wanted to switch the call from `org.eclipse.jdt.internal.compiler.parser.Scanner.resetTo(int, int)` to `org.eclipse.jdt.internal.compiler.parser.Scanner.resetTo(int, int, boolean)` why won't you simply pass true.

srikanth-sankaran (commented):
Additional comment

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
jukzi: When debugging the test it was visible, that during isInModuleDeclaration() this.insideModuleInfo was still false even though it was within a module. so the point is that resetTo(0, begin) did not setup this.insideModuleInfo correctly. If this code is called only in a module-info it is equivalent to use `true `or Scanner.this.insideModuleInfo. But its not obvious that that the code should only be used that way.

jukzi:
@srikanth-sankaran i took sampling measurement with the latest build and i still see getMainTypeName() unnecessary high rated in the hotspots - when called from getVanguardParser(). Can we consider to apply this patch on top of the other improvement?

image

srikanth-sankaran:
I am back from a 4 day long official holiday break. I'll take a look at this within the next 2-3 days and help close it for M2. Thanks for your patience @jukzi

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
srikanth-sankaran: > When debugging the test it was visible, that during isInModuleDeclaration() this.insideModuleInfo was still false even though it was within a module. so the point is that resetTo(0, begin) did not setup this.insideModuleInfo correctly. If this code is called only in a module-info it is equivalent to use `true `or Scanner.this.insideModuleInfo. But its not obvious that that the code should only be used that way.

Please take a look at the first line of `org.eclipse.jdt.internal.compiler.parser.Scanner.getScanContext(int)` which reads:

```
if (!isInModuleDeclaration())
		return ScanContext.INACTIVE;
```

The only instantiation of `org.eclipse.jdt.internal.compiler.parser.Scanner.ScanContextDetector` happens below this. So it is guaranteed that it is created only for module-info files. In fact its very purpose is to handle module-info files. So this change is obscuring matters. Leave it as `return true;` please

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
```diff
@@ -4861,12 +4860,12 @@ void updateScanContext(int token) {
...
 	}
 
 	public ScanContext getScanContext(char[] src, int begin) {
 		this.scanner.setSource(src);
-		this.scanner.resetTo(0, begin);
+		this.scanner.resetTo(0, begin, isParsingModuleDeclaration());
 		goForCompilationUnit();
```
srikanth-sankaran, line 4869: Change the third parameter to `true` if you need the version with the third parameter please. Again at this place we are inside module-info

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
```diff
@@ -2567,8 +2567,7 @@ public final boolean jumpOverUnicodeWhiteSpace() throws InvalidInputException {
 }
 
 public boolean isInModuleDeclaration() {
-	return this.fakeInModule || this.insideModuleInfo ||
-			(this.activeParser != null ? this.activeParser.isParsingModuleDeclaration() : false);
+	return this.fakeInModule || this.insideModuleInfo;
 }
```
srikanth-sankaran, line 2571: I think this change is the most crucial and looks fine. If I look at the write operations to `org.eclipse.jdt.internal.compiler.parser.Scanner.activeParser` there is ONLY one in `org.eclipse.jdt.internal.compiler.parser.Scanner.setActiveParser(ConflictedParser)` - This one already caches whether we are in module declaration.

So the code `this.activeParser != null ? this.activeParser.isParsingModuleDeclaration() : false` is redundant - it is already cached in Scanner#isModuleInfo

srikanth-sankaran (commented):
Of the three chunks of code - I suggest you get rid of one (change of true to some computation), change one to pass true if needed rather than a method call which is guaranteed to return true.

Change to `org.eclipse.jdt.internal.compiler.parser.Scanner.isInModuleDeclaration()` looks good

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
jukzi: understood. I'll rename the class to make it more obvious, OK?

srikanth-sankaran:
Sorry, couldn't get to it today. Will review this tomorrow. Thanks for your patience.

srikanth-sankaran (approved):
LGTM. +1

PR #1830, opened 2024-01-10
Fix files that have been checked in with \r\n to use only \n

merks:
## What it does

Fixes files that have been committed with \r\n to use only \n.

## How to test

This is a whitespace only change, so nothing to test.

## Author checklist

- [x] I have thoroughly tested my changes (there is nothing to test though.


merks:
If I read the history correctly these appear to be coming from @srikanth-sankaran's commits.

image

Please use autocrlf true or autocrlf input when working on Windows:

https://stackoverflow.com/questions/3206843/how-line-ending-conversions-work-with-git-core-autocrlf-between-different-operat

merks:
FYI, I notice such problems because one sees dirty files that have no changes and cannot be made clean.

image

And then one eventually does a pull and can't update such a file because of "conflicts."  The only workaround is to  use "assume unchanged" but that works only temporarily until the next time.

akurtakov:
Looks good.

srikanth-sankaran:
> If I read the history correctly these appear to be coming from @srikanth-sankaran's commits.

Thanks @merks - I seem to have autocrlf=true in my git system configuration file and  autocrlf=false in my git user configuration file. Don't know how it came about - Now both are set to true.

merks:
FYI, it's not really much trouble.  I have an oomph utility that fix such problems so it only takes a few minutes.  Of course best to avoid the problem in the first place. 😄

PR #1831, opened 2024-01-10
[performance] less calls to isInModuleDeclaration() #1747

jukzi:
Reorder costly conditions

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1747

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
@@ -4902,8 +4902,9 @@ private VanguardScanner getNewVanguardScanner() {
 	return vs;
 }
 protected final boolean mayBeAtCasePattern(int token) {
-	return (!isInModuleDeclaration() && JavaFeature.PATTERN_MATCHING_IN_SWITCH.isSupported(this.complianceLevel, this.previewEnabled))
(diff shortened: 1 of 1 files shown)


srikanth-sankaran (changes requested):
Why don't we pull up the token check to the call site ?? As pointed out earlier every other maybe* method is called after the token check. 

The way it stands now we will still make gazillion trips into maybeAtCasePattern although each call is faster now. 

I am making more of a consistency argument than a performance argument here. The patch may have solved the performance problem.

jukzi:
> Why don't we pull up the token check to the call site ??

i did it to avoid code duplication:
image

jukzi:
> The way it stands now we will still make gazillion trips into maybeAtCasePattern although each call is faster now.

if that is a hotspot typically the hotspot compiler would inline the method. I would not make assumptions that it is a problem just by looking at the code.

srikanth-sankaran:
Ok. I would have preferred the checks being hoisted even if into two places but I can live with what you have proposed. Please proceed.

srikanth-sankaran:
> > The way it stands now we will still make gazillion trips into maybeAtCasePattern although each call is faster now.
> 
> 
> 
> if that is a hotspot typically the hotspot compiler would inline the method. I would not make assumptions that it is a problem just by looking at the code.

Let us say hypothetically that it is a hotspot. The code being inlined into the call sites would be the wrong optimization imo but we don't have to argue that 😊

srikanth-sankaran (approved):

Issue #1835, opened 2024-01-10
[Switch][patterns][record] AssertionError at org.eclipse.jdt.internal.compiler.ast.YieldStatement.addSecretYieldResultValue(YieldStatement.java:120)


KostiantynDenysiuk:
It seems that expression String s = switch(...) should take a value in any case unless the control flow is not broken. 
When we throw exception in default block, we got an error when parse this example. 

[error]: Unexpected exception: java.lang.AssertionError
java.lang.AssertionError: null
at org.eclipse.jdt.internal.compiler.ast.YieldStatement.addSecretYieldResultValue(YieldStatement.java:120) ~[ecj-3.36.0-20231107.jar:?]....

Reproducer: 

public class Reproducer {

   public class DataX {
         String data;
    }

   public class DataY {
        String data;
   }

   record X (DataX data) {}
   record Y (DataY data) {}

   Reproducer() {
        DataX dataX = new DataX();
        DataY dataY = new DataY();

        X x = new X(dataX);
        Y y = new Y(dataY);

        foo(x);
    }

    void foo(Object obj) {
           String s = switch(obj) {
                  case X(var x) -> x.data;
                  case Y(var y) -> y.data;
        
                 default -> {
                          try {
                                   throw new Exception();
                               } catch (Exception e) {
                                     throw new RuntimeException();
                               }
                 }
          };
    }

}

iloveeclipse:
Two things: 
1) something in your snippet is missing.
2) please try with latest nightly SDK build from https://download.eclipse.org/eclipse/downloads/drops4/I20240109-1800/ , there were few fixes in switch area.

KostiantynDenysiuk:
I've just tried new fixes but nothing has changed.

iloveeclipse:
> I've just tried new fixes but nothing has changed.

OK. Please then provide full self contained java file showing the problem.

KostiantynDenysiuk:
[Reproducer.zip](https://github.com/eclipse-jdt/eclipse.jdt.core/files/13888571/Reproducer.zip)

srikanth-sankaran:
I'll take a look.

srikanth-sankaran:
Hello @KostiantynDenysiuk - On master, I don't see the problem. This program:

```
public class Reproducer {

	public class DataX {
		String data = "DataX";
	}

	public class DataY {
		String data1 = "DataY";
	}

	record X(DataX data) {
	}

	record Y(DataY data) {
[...]

PR #1840, opened 2024-01-11
 AssertionError at org.eclipse.jdt.internal.compiler.ast.YieldStatement.addSecretYieldResultValue(YieldStatement.java:120) 

srikanth-sankaran:
## What it does

Resolve the yield statement in its enclosing scope.
Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1835

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/YieldStatement.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/YieldStatement.java
@@ -31,7 +31,6 @@ public class YieldStatement extends BranchStatement {
 	public boolean isImplicit;
 	static final char[] SECRET_YIELD_RESULT_VALUE_NAME = " secretYieldValue".toCharArray(); //$NON-NLS-1$
 	private LocalVariableBinding secretYieldResultValue = null;
-	public BlockScope scope;
 
 public YieldStatement(Expression exp, int sourceStart, int sourceEnd) {
 	super(null, sourceStart, sourceEnd);
@@ -103,7 +102,7 @@ public FlowInfo analyseCode(BlockScope currentScope, FlowContext flowContext, Fl
 protected void setSubroutineSwitchExpression(SubRoutineStatement sub) {
 	sub.setSwitchExpression(this.switchExpression);
 }
-protected void addSecretYieldResultValue(BlockScope scope1) {
+protected void addSecretYieldResultValue(BlockScope scope) {
 	SwitchExpression se = this.switchExpression;
 	if (se == null || !se.containsTry)
 		return;
@@ -117,8 +116,8 @@ protected void addSecretYieldResultValue(BlockScope scope1) {
 	local.declaration = new LocalDeclaration(YieldStatement.SECRET_YIELD_RESULT_VALUE_NAME, 0, 0);
 	assert se.yieldResolvedPosition >= 0;
 	local.resolvedPosition = se.yieldResolvedPosition;
-	assert local.resolvedPosition < this.scope.maxOffset;
-	this.scope.addLocalVariable(local);
+	assert local.resolvedPosition < scope.maxOffset;
+	scope.addLocalVariable(local);
 	this.secretYieldResultValue = local;
 }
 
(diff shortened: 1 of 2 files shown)


srikanth-sankaran:
The concerned assert got introduced in the fix for https://bugs.eclipse.org/bugs/show_bug.cgi?id=545567 and appears suspect.

Here is the relevant code in `org.eclipse.jdt.internal.compiler.ast.YieldStatement.addSecretYieldResultValue(BlockScope)`

```
         SwitchExpression se = this.switchExpression;
                             ...
                             ...
         assert se.yieldResolvedPosition >= 0;
	local.resolvedPosition = se.yieldResolvedPosition;
	assert local.resolvedPosition < this.scope.maxOffset;
```

The problem is that `org.eclipse.jdt.internal.compiler.ast.SwitchExpression.yieldResolvedPosition` gets assigned in `org.eclipse.jdt.internal.compiler.ast.SwitchExpression.processTypesBindingsOnStack(CodeStream)` which works with the scope associated with the `SwitchExpression`.

But a `YieldStatement` gets its own subscope in `org.eclipse.jdt.internal.compiler.ast.YieldStatement.resolve(BlockScope)` thusly:

```
private boolean isInsideTry() {
	return this.switchExpression != null && this.switchExpression.containsTry;
}
@Override
public void resolve(BlockScope skope) {
	this.scope = isInsideTry() ? new BlockScope(skope) : skope;
```

In the present test case, `isInsideTry()` is true and so YieldStatements are assigned a subscope of `SwitchExpression`'s scope.

But the assert is asserting that a resolved position computed wrt to a parent scope is within a subscope of the parent.

Something amiss there.

I see three possible fixes:

1. Is the assert really adding value ? If not get rid of it.
2. Compare the proper scopes by also checking for  `isInsideTry()` inside `YieldStatement.addSecretYieldResultValue` 
3. Resolve YieldStatements in the same enclosing scope instead of creating a subscope.

In the PR, I am taking approach 3. All tests are green - So it is unclear what the motivation for creating a subscope for yield statements is - @mpalat can you shed some light on this ??

srikanth-sankaran:
@mpalat - Can you review this please ?? Is getting rid of the subscope creation violates some design intent ??

Could you also take a look at the code generated for the minimal test case and confirm it meets expectations ? 

TIA

mpalat:
> @mpalat - Can you review this please ?? Is getting rid of the subscope creation violates some design intent ??

@srikanth-sankaran will try to recollect (its been sometime) but I believe this is documented in detail in the bug - please take a look at https://bugs.eclipse.org/bugs/show_bug.cgi?id=545567#c20 - specifically this line "currently in the patch it is upper scope of switch expression - ideally I think it should be in the switch expression scope since multiple sibling switch expressions in a block will have conflict" - I believe this was the design intent to have subscopes - I don't have a ready-made example for this though.

srikanth-sankaran:
> > @mpalat - Can you review this please ?? Is getting rid of the subscope creation violates some design intent ??
> 
> @srikanth-sankaran will try to recollect (its been sometime) but I believe this is documented in the bug - please take a look at https://bugs.eclipse.org/bugs/show_bug.cgi?id=545567#c20 - specifically this line "currently in the patch it is upper scope of switch expression - ideally I think it should be in the switch expression scope since multiple sibling switch expressions in a block will have conflict" - I believe this was the design intent to have subscopes - no I don't have an example demonstrating this issue. Tried out a few scenarios, but they are working with your PR. So I believe, this subscoping was not required - +1 for removing the subscope.

@mpalat on master the secret variable is added to the switch expression's scope only and not to its upperscope. I say this because in `org.eclipse.jdt.internal.compiler.ast.SwitchStatement.resolve(BlockScope)` I see the following: 

```
			addSecretPatternSwitchVariables(upperScope);
			if (this.statements != null) {
				if (this.scope == null)
					this.scope = new BlockScope(upperScope);
```

So, if the switch expression has a body at all, it gets its own scope. _It has to!_ Anything declared (named non-secret variables) inside switch's {} _cannot_ be added to its upper scope

And `org.eclipse.jdt.internal.compiler.ast.SwitchExpression.yieldResolvedPosition` is computed wrt to the scope of the switch expression in `org.eclipse.jdt.internal.compiler.ast.SwitchExpression.processTypesBindingsOnStack(CodeStream)` and it is only right that the secret variable should be added to the scope of the switch.

I have also pushed a commit, getting rid of the state `scope` in `YieldStatement` which was introduced for this express purpose.

mpalat (approved):
+1, Looks good - Thanks @srikanth-sankaran

srikanth-sankaran:
> +1, Looks good - Thanks @srikanth-sankaran

Thank you!

Issue #1717, opened 2023-12-09
ECJ batch compiler logs INFO different then WARNING


jukzi:
from https://ci.eclipse.org/jdt/job/eclipse.jdt.ui-github/job/PR-997/1/consoleFull
```
23:22:02  [INFO] --- tycho-compiler:4.0.5-SNAPSHOT:compile (default-compile) @ org.eclipse.jdt.ui ---
23:22:02  [INFO] Compiling 1783 source files to /home/jenkins/agent/workspace/eclipse.jdt.ui-github_PR-997/org.eclipse.jdt.ui/target/classes using Eclipse Compiler for Java(TM) 3.36.0.v20231114-0937
23:22:23  [WARNING] /home/jenkins/agent/workspace/eclipse.jdt.ui-github_PR-997/org.eclipse.jdt.ui/ui/org/eclipse/jdt/internal/ui/text/correction/SuppressWarningsSubProcessor.java:[65] 
23:22:23  	static final String ADD_SUPPRESSWARNINGS_ID= SuppressWarningsSubProcessorCore.ADD_SUPPRESSWARNINGS_ID;
23:22:23  	                    ^^^^^^^^^^^^^^^^^^^^^^^
23:22:23  The field SuppressWarningsSubProcessor.ADD_SUPPRESSWARNINGS_ID is hiding a field from type SuppressWarningsSubProcessorCore
23:22:23  3. INFO in /home/jenkins/agent/workspace/eclipse.jdt.ui-github_PR-997/org.eclipse.jdt.ui/ui/org/eclipse/jdt/ui/text/java/correction/ChangeCorrectionProposal.java (at line 195)
23:22:23  	ITextOperationTarget textOperationTarget= activeEditor.getAdapter(ITextOperationTarget.class);
23:22:23  	                                          ^^^^^^^^^^^^
23:22:23  Potential null pointer access: The variable activeEditor may be null at this location
23:22:23  [WARNING] /home/jenkins/agent/workspace/eclipse.jdt.ui-github_PR-997/org.eclipse.jdt.ui/ui refactoring/org/eclipse/jdt/internal/ui/refactoring/actions/RefactoringActions.java:[53] 
23:22:23  	private static IType convertToEnclosingOrPrimaryType(IJavaElement element) throws JavaModelException {
23:22:23  	                                                                                  ^^^^^^^^^^^^^^^^^^
[...]

PR #1846, opened 2024-01-11
Update Jenkinsfile

jukzi:
skip (not working) Info marker

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1717


jukzi:
that NORMAL marker is also not rendered well :-(:
image

PR #1862, opened 2024-01-13
Allow resolving compilation unit (DOM) with Javac

mickaelistria:
Extracts an interface ICompilationUnitResolver out of CompilationUnitResolver. This interface is used by ASTParser to create DOM instance from a Java source. The default implementation remains CompilationUnitResolver and uses ECJ parser/resolver to create a DOM. A system property `ASTParser.compilationUnitResolver` can be used to force usage of alternative implementations. At the moment, those alternative implementations need to be in the same classloader (eg in a fragment).

A new fragment org.eclipse.jdt.core.javac is introduced and provides an implementation of ICompilationUnitResolver which uses Javac "API"s to create a JDT DOM. It can be enabled by setting
`-DASTParser.compilationUnitResolver=org.eclipse.jdt.core.dom.JavacCompilationUnitResolver`.

### DIFF
--- a/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
+++ b/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
@@ -33,7 +33,6 @@ import org.eclipse.jdt.core.JavaModelException;
 import org.eclipse.jdt.core.WorkingCopyOwner;
 import org.eclipse.jdt.core.compiler.CategorizedProblem;
 import org.eclipse.jdt.core.compiler.CharOperation;
-import org.eclipse.jdt.internal.compiler.ast.CompilationUnitDeclaration;
 import org.eclipse.jdt.internal.compiler.ast.ConstructorDeclaration;
 import org.eclipse.jdt.internal.compiler.ast.ExplicitConstructorCall;
 import org.eclipse.jdt.internal.compiler.batch.FileSystem.Classpath;
@@ -221,6 +220,8 @@ public class ASTParser {
 	 */
 	private int bits;
 
+	private final ICompilationUnitResolver unitResolver;
+
 	/**
 	 * Creates a new AST parser for the given API level.
 	 * <p>
@@ -233,6 +234,7 @@ public class ASTParser {
 	ASTParser(int level) {
 		DOMASTUtil.checkASTLevel(level);
 		this.apiLevel = level;
+		this.unitResolver = ICompilationUnitResolver.getInstance();
 		initializeDefaults();
 	}
 
@@ -245,17 +247,23 @@ public class ASTParser {
 			}
 			if (this.sourcepaths != null) {
 				for (int i = 0, max = this.sourcepaths.length; i < max; i++) {
-					String encoding = this.sourcepathsEncodings == null ? null : this.sourcepathsEncodings[i];
-					main.processPathEntries(
-							Main.DEFAULT_SIZE_CLASSPATH,
-							allClasspaths, this.sourcepaths[i], encoding, true, false);
+					String sourcePath = this.sourcepaths[i];
+					if (sourcePath != null) {
+						String encoding = this.sourcepathsEncodings == null ? null : this.sourcepathsEncodings[i];
+						main.processPathEntries(
+								Main.DEFAULT_SIZE_CLASSPATH,
+								allClasspaths, sourcePath, encoding, true, false);
+					}
 				}
 			}
 			if (this.classpaths != null) {
 				for (int i = 0, max = this.classpaths.length; i < max; i++) {
-					main.processPathEntries(
-							Main.DEFAULT_SIZE_CLASSPATH,
-							allClasspaths, this.classpaths[i], null, false, false);
+					String classpath = this.classpaths[i];
+					if (classpath != null) {
+						main.processPathEntries(
+								Main.DEFAULT_SIZE_CLASSPATH,
+								allClasspaths, classpath, null, false, false);
+					}
 				}
 			}
 			ArrayList pendingErrors = main.pendingErrors;
@@ -951,9 +959,9 @@ public class ASTParser {
 				if ((this.bits & CompilationUnitResolver.BINDING_RECOVERY) != 0) {
 					flags |= ICompilationUnit.ENABLE_BINDINGS_RECOVERY;
 				}
-				CompilationUnitResolver.resolve(compilationUnits, bindingKeys, requestor, this.apiLevel, this.compilerOptions, this.project, this.workingCopyOwner, flags, monitor);
+				this.unitResolver.resolve(compilationUnits, bindingKeys, requestor, this.apiLevel, this.compilerOptions, this.project, this.workingCopyOwner, flags, monitor);
 			} else {
-				CompilationUnitResolver.parse(compilationUnits, requestor, this.apiLevel, this.compilerOptions, flags, monitor);
+				this.unitResolver.parse(compilationUnits, requestor, this.apiLevel, this.compilerOptions, flags, monitor);
 			}
 		} finally {
 			// reset to defaults to allow reuse (and avoid leaking)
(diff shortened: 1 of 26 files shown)


README_JAVAC_EXPERIMENT.md
```diff
@@ -0,0 +1,57 @@
+# JDT over Javac
```
jukzi, line 1: move the file to o.e.j.c.javac/README.md please

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
```diff
@@ -233,9 +235,24 @@ public static ASTParser newParser(int level) {
 	ASTParser(int level) {
 		DOMASTUtil.checkASTLevel(level);
 		this.apiLevel = level;
+		this.unitResolver = getCompilationUnitResolverInstance();
 		initializeDefaults();
 	}
 
+	@SuppressWarnings("unchecked")
+	private static ICompilationUnitResolver getCompilationUnitResolverInstance() {
```
jukzi, line 243: move to ICompilationUnitResolver.getInstance()

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
```diff
@@ -1362,6 +1338,7 @@ private ASTNode internalCreateASTCached(IProgressMonitor monitor) {
 	 * @see ASTNode#getLength()
 	 */
 	private ASTNode internalCreateASTForKind() {
+		// TODO make it independent from ECJ
```
jukzi, line 1341: no TODOS in master please

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolver.java
```diff
@@ -81,6 +81,100 @@
 
 @SuppressWarnings({ "rawtypes", "unchecked" })
 class CompilationUnitResolver extends Compiler {
+	
+	public static final ICompilationUnitResolver FACADE = new ICompilationUnitResolver() {
```
jukzi, line 85: better a named Class to get more meaningfull stacktraces

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ICompilationUnitResolver.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.Map;
+
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.jdt.core.ICompilationUnit;
+import org.eclipse.jdt.core.IJavaElement;
+import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.jdt.core.WorkingCopyOwner;
+import org.eclipse.jdt.internal.compiler.batch.FileSystem.Classpath;
+
+interface ICompilationUnitResolver {
```
jukzi, line 23: is this going to be public API? then it needs tons of javadoc

jukzi (changes requested):
https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-1862/2/java/new/ states the commit contains 31 new warnings. JDT core tries to keep warnings down to zero

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ICompilationUnitResolver.java
mickaelistria: No, internal only so far.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
mickaelistria: I can just remove the TODO for the convention, but that would be kind of lying: there would still be things to do here, things that won't be done very soon. So how to track such "will need more work" spots in JDT?
To be honest, I think a convention against TODOs is kind of missing the idea that software development can be incremental.

jukzi: The problem with TODOS in JDT Core is that they tend to silently stay forever - currently already 260 matches :-(
Until the TODOs are solved its better to stay on a branch, like the compiler guys do it with new Java features.

mickaelistria: > The problem with TODOS in JDT Core is that they tend to silently stay forever

That's not only in JDT, it's the nature of TODOs.
TODOs rarely become highest priority -they're not at the time they're written, otherwise they would be fixed. Yet they're valuable hints for people who come to this code as they point to hotspots where to look at. I've seen some TODOs here and there actually describing a problem that was not so important for a while, and suddenly important enough to consider, and the TODO pointing to the right location and action.

But I'm curious here, is the "no TODO" rule a rule that is backed by the whole project or just your opinion?
ie would it be an objective blocker for a future merge, or are we just discussing it "casually" because why not?

> Until the TODOs are solved its better to stay on a branch, like the compiler guys do it with new Java features.

I strongly disagree with that. This approach of keeping upcoming critical value on a branch is much harder for new contributor to join development and testing. It's IMO a bad practice, and -without willing to be disrespectful of the legacy workflows that were relevant at some point- they're also 1 reason why JDT is facing difficulties coping with new Java releases.
I'd rather avoid replicating this bad practice.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolver.java
mickaelistria: I'm thinking about renaming the ICompilationUnitResolver interface to ICompilationUnitResolverFacade, and the particular implementation here as `ECJCompilationUnitResolverFacade`. What do you think?

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
mickaelistria: Good idea, I'll do that.

README_JAVAC_EXPERIMENT.md
mickaelistria: Ack

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolver.java
jukzi: I don't care about the actual name much. Only find it hard to read annonymous classes. Personally i would skip the postfix "Fasade", as it should be an implementation details if it's implemented as Fasace.

jukzi:
@mickaelistria i suggest that you also search for feedback from other JDT committers, as i can only give some formal hints and won't be able to give enough substantial feedback to submit

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ICompilationUnitResolver.java
```diff
@@ -0,0 +1,60 @@
...
+
+	@SuppressWarnings("unchecked")
+	static ICompilationUnitResolver getInstance() {
+		String compilationUnitResolverClass = System.getProperty(ICompilationUnitResolver.class.getSimpleName());
+		if (compilationUnitResolverClass != null) {
+			try {
+				Class<? extends ICompilationUnitResolver> clazz = (Class<? extends ICompilationUnitResolver>) Class.forName(compilationUnitResolverClass);
+				return clazz.getDeclaredConstructor().newInstance();
+			} catch (Exception e) {
+				ILog.get().error(e.getMessage(), e);
```
jukzi, line 54: e.getMessage() may be null. please add a unique error text, including the offending "compilationUnitResolverClass"

jukzi (dismissed):
my overall impression is that the many "TODOs" and uncommented code lines render this changes far from production ready on master.

iloveeclipse:
Can we limit the changes that would be merged to master to the changes that need to be done in existing JDT code to support new fragment, plus tests for these changes / API added, and keep "incubating" fragment code in a separated repository (original one from Mickael)?

The changes in existing code must work without fragment, and I assume they should be more or less minimal - the main work (and most of changes) should happen anyway in the fragment.

If that would be OK, we would have "small/slow changing" integration layer in JDT core and "big/fast changing" fragment. That would also match resources we have available for the review.

For this PR it would mean we basically keep only changes on the files below: 
- org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
- org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolver.java
- org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ICompilationUnitResolver.java

plus a test that validates that org.eclipse.jdt.core.dom.ICompilationUnitResolver.getInstance() can return non-default ICompilationUnitResolver if configured.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
robstryker: Regarding TODOs,  it would appear almost all of the TODO listed here are in what is a new bundle. This seems very reasonable to me, as newer bundles, whose logic only gets enabled by proactively setting some new setting, are usually under a fast-paced development and are not expected to have all of their implementation functional immediately. 

If @mickaelistria  had peppered TODOs throughout the existing jdt.core code, I would agree that this is not good practice and could leave the codebase in a messy and incomplete state. 

But in this case, I disagree.

mickaelistria:
Replaced with https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1984 as source repo/branch has changed.

PR #1866, opened 2024-01-15
Update Jenkinsfile: eclipse warnings

jukzi:
reject new ECJ warnings


iloveeclipse:
> do you remember what was the exact problem in [eclipse-jdt/eclipse.jdt.ui#1026](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1026) ? I would like to get those warnings stable and reject new.

No. 
I given up trying to understand quality gate results as they always were weird in some way, so it was always wasted time.

PR #1878, opened 2024-01-17
Demonstrate "Make Static" refactoring on AbstractRegressionTest

HeikoKlare:
This is a dogfooding change demonstrating the "Make Static" refactoring recently added to JDT UI (eclipse-jdt/eclipse.jdt.ui#689). It is only supposed to be used for demonstration and validation purposes and is **not to be merged**.

The change consists of the applications of the refactoring to all methods within `org.eclipse.jdt.core.tests.compiler.regression.AbstractRegressionTest` that fulfill the refactoring's preconditions (in particular not being overwriting or overwritten methods).

The refactoring was used in the state with the to-be-integrated fix in eclipse-jdt/eclipse.jdt.ui#1085

### DIFF
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/AbstractRegressionTest.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/AbstractRegressionTest.java
@@ -139,7 +139,7 @@ public abstract class AbstractRegressionTest extends AbstractCompilerTest implem
 		ASTVisitor visitor;
 
 		protected void runConformTest() {
-			runTest(this.shouldFlushOutputDirectory,
+			AbstractRegressionTest.runTest(AbstractRegressionTest.this, this.shouldFlushOutputDirectory,
 					this.testFiles,
 					this.dependantFiles != null ? this.dependantFiles : new String[] {},
 					this.classLibraries,
@@ -164,7 +164,7 @@ public abstract class AbstractRegressionTest extends AbstractCompilerTest implem
 		}
 
 		protected void runNegativeTest() {
-			runTest(this.shouldFlushOutputDirectory,
+			AbstractRegressionTest.runTest(AbstractRegressionTest.this, this.shouldFlushOutputDirectory,
 					this.testFiles,
 					this.dependantFiles != null ? this.dependantFiles : new String[] {},
 					this.classLibraries,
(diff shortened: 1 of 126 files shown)


org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/AbstractRegressionTest.java
```diff
@@ -1402,14 +1402,14 @@ protected ClassFileReader getInternalClassFile(String directoryName, String clas
...
 		} finally {
-			removeTempClass(className);
+			AbstractRegressionTest.removeTempClass(className);
 		}
 	}
 
-	protected void checkDisassembledClassFile(String fileName, String className, String expectedOutput) throws Exception {
-		this.checkDisassembledClassFile(fileName, className, expectedOutput, ClassFileBytesDisassembler.DETAILED);
+	protected static void checkDisassembledClassFile(AbstractRegressionTest abstractRegressionTest, String fileName, String className, String expectedOutput) throws Exception {
+		AbstractRegressionTest.checkDisassembledClassFile(fileName, className, expectedOutput, ClassFileBytesDisassembler.DETAILED);
```
jukzi, line 1410: the prefix "AbstractRegressionTest." feels unnecessary

HeikoKlare: It is syntactically unnecessary, that's true. You could remove it here and it is just a matter of preference whether you want to have this class name for calls of static methods within the same class.
We have actively chosen to implement it this way and also reflected this in test cases for the refactoring (e.g. `testThisInDeclaration`). The decision was guided by the fact that, e.g., IntelliJ does it the same way.
We can, of course, also implement it differently. If you see good reasons to change that, we can open and address an according issue.

jukzi: personally idk, since i probably will never use that refactoring. i don't get the value of it. I had hoped it would be a useful cleanup that would just add "static" wherever "this" is not needed.

HeikoKlare: We have also thought about reusing the refactoring functionality to provide a quick fix based on it that shows up for methods that can be made static without adding a parameter (as they do not rely on `this`). That would a follow-up task. Value of the refactoring of course depends on your use cases. We sometimes have them and probably others have them as well (otherwise competitors like IntelliJ would have probably not have them implemented).

HeikoKlare:
Closing as this was supposed to be a demonstration-only dogfooding PR for the "Make Static" refactoring.

Issue #1894, opened 2024-01-22
Slow JavaModelManager.getInfo(IJavaElement)


jukzi:
I sampled a full build of Platform Workspace while i had 90 Java Editors open. Beside FileSystem IO the top hotspot is JavaModelManager.getInfo
The only slow thing about is is it's "synchronized" which is unnecessary when only the threadlocal temporaryCache is used
![image](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/51790620/ce4232c9-1a4d-46a7-94fe-0257faf92ec2)

jukzi:
I was not able to reproduce the problem. Today JavaModelManager.getInfo did not show in hotspots:

![image](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/51790620/6fe1ac0e-b2a5-41d9-97c5-d2eef114803c)

There must be some special condition that i am missing why the java editors synchronized with the build

jukzi:
closing, since i could not reproduce

PR #1899, opened 2024-01-23
performance: do not synchronize JavaModelManager.temporaryCache #1894

jukzi:
used in JavaModelManager.getInfo(IJavaElement)
it's a ThreadLocal anyway.

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1894

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java
@@ -124,7 +124,6 @@ import org.eclipse.jdt.core.IPackageFragmentRoot;
 import org.eclipse.jdt.core.IParent;
 import org.eclipse.jdt.core.IProblemRequestor;
 import org.eclipse.jdt.core.IType;
-import org.eclipse.jdt.core.ITypeRoot;
 import org.eclipse.jdt.core.JavaConventions;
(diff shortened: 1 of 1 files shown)


org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java
```diff
@@ -1258,7 +1257,7 @@ public static IJavaElement determineIfOnClasspath(IResource resource, IJavaProje
 	/**
 	 * Infos cache.
 	 */
-	private JavaModelCache cache;
+	private final JavaModelCache cache = new JavaModelCache();
```
iloveeclipse, line 1260: Interesting, this is actually "back to the roots". Original code had this until commit af33da7216734bcb2ae0e5dbadaa836c42812db8 that moved init after init of debug options.

iloveeclipse (approved):
Looks safe. 

Do you see a measurable performance improvement with this?

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java
jukzi: The JavaModelCache constructor already uses the debug options, therefore the debug options have been initialized before.

jukzi:
closing because i could not reproduce + issue with debug tracing options

PR #1916, opened 2024-01-26
final JavaElement elements

jukzi:
[performance] cache org.eclipse.jdt.internal.core.JavaElement.hashCode


jukzi:
Example Measurement: empty workspace, show type hierarchy of java.lang.Object
before JavaElement.hashCode() often called Arrays.hashCode taking ~80ms: (hotspotview)
image

after: JavaElement.hashCode() takes only 5ms. Arrays.hashCode not visible in hotspots anymore:
image

the reverse call stack before was:
image

Issue #1921, opened 2024-01-29
Non deterministic compile results with lamda numbering


laeubi:
In PDE we are experiencing a build instability with the class [BundleLauncherHelper](https://github.com/eclipse-pde/eclipse.pde/blob/master/ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/BundleLauncherHelper.java) according to the logs the numbering of lamdas is different compared to those on the baseline, I have captured a compile log that shows the differences and **has compile logs with exact classpath and compiler options** here: [compare.zip](https://github.com/eclipse-jdt/eclipse.jdt.core/files/14088236/compare.zip)

The textcompare shows that only the numbering is different here, please note that the baseline seem not contain any `lambda$0` but starting with `lambda$1` immediately:
![grafik](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/1331477/aa8be345-9adb-4f3c-aa73-c9f4a217996b)

I can't reproduce this locally and even on the CI instance the behavior varies between builds where none are actually changing that bundle.

laeubi:
I can't say I understand anything about how lambda handling works, but can't the numbering use the order in wich it encounters the lamda in the source code?
Can we do something to help the compiler, e.g. sort source files by name? So basically the order of this list influences lamda numbers:

```
<argument value="/home/jenkins/agent/workspace/eclipse.pde_master/ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/BundleLauncherHelper.java"/>
<argument value="/home/jenkins/agent/workspace/eclipse.pde_master/ui/org.eclipse.pde.launching/src/org/eclipse/pde/launching/JUnitLaunchConfigurationDelegate.java"/>
<argument value="/home/jenkins/agent/workspace/eclipse.pde_master/ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/ILaunchingPreferenceConstants.java"/>
<argument value="/home/jenkins/agent/workspace/eclipse.pde_master/ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/LaunchConfigurationListener.java"/>
<argument value="/home/jenkins/agent/workspace/eclipse.pde_master/ui/org.eclipse.pde.launching/src/org/eclipse/pde/internal/launching/launcher/RequirementHelper.java"/>
[...]

PR #1922, opened 2024-01-30
Sort filename (if given) by name for consistent output

laeubi:
Currently the generation of lamdas can change depending on when the source file is processed.

## What it does

This sorts the array of files (if given) to always have a predictable order independent of order given on the commandline / arguments.

Relates to https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1921

## How to test
Use this project:
- https://github.com/eclipse-pde/eclipse.pde/tree/master/ui/org.eclipse.pde.launching
- pass the source files in different order to the batch compiler
- see that depending on the order you get different lamda namings for the same code
- with this patch the ordering of source file arguments do not matter

See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1921 for detailed compiler arguments, logs and example class files showing the difference.

## Author checklist

- [x] I have tested my changes in a debugger with a Tycho build by manipulating the passed source files order

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/Main.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/Main.java
@@ -72,6 +72,7 @@ import java.util.ResourceBundle;
 import java.util.Set;
 import java.util.StringTokenizer;
 import java.util.function.Function;
+import java.util.stream.IntStream;
 
 import org.eclipse.jdt.core.compiler.CategorizedProblem;
 import org.eclipse.jdt.core.compiler.CharOperation;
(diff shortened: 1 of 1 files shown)


iloveeclipse (changes requested):
The problem is "per output destination" feature, see https://bugs.eclipse.org/bugs/show_bug.cgi?id=146554

`Main.filenames` & `Main.destinationPaths` should be in-sync, so change must be done not here but in `configure()`

iloveeclipse:
> I applied a workaround to Tycho to simulate what this patch does

I would keep that in Tycho, as the patch here is incomplete and needs to be fixed.

laeubi:
> The problem is "per output destination" feature, see https://bugs.eclipse.org/bugs/show_bug.cgi?id=146554
> 
> `Main.filenames` & `Main.destinationPaths` should be in-sync, so change must be done not here but in `configure()`

You are right, I also noticed that there are even `public fields` in this class so now I only sort the index here... another approach would be to do the ordering in the compiler itself I'll take a look at this as well.

laeubi:
This now seems to pass the testsuite and already shows some classes that are different in JDT, I have createad a PR here:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1758

so get a better overview what would happen in the large (requires snapshot Tycho).

> I would keep that in Tycho, as the patch here is incomplete and needs to be fixed.

I leave the decisions to you (or JDT team) if this change is useful so it can benefit everyone not only latest Tycho users...

iloveeclipse:
> if this change is useful so it can benefit everyone not only latest Tycho users.

What I've meant is: as long as patch here is incomplete and breaks ecj functionality, Tycho workaround has to be used.
If the patch is OK, we will for sure like to have it in the compiler, as stable compilation behavior is important for everyone.

laeubi:
If I where to fix it on the compiler level it seems worth to sort the CUs by name in the method `org.eclipse.jdt.internal.compiler.Compiler.internalBeginToCompile(ICompilationUnit[], int)` I'll prepare a PR for this to see if it gives similar results but that's beyond my understanding if compiler internals to judge what is a better approach.

laeubi:
This is done on the Compiler /CU level
- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1933

laeubi:
Looking at the build results this has not produced any baseline problem eve though it would be expected it seems the change has had not the same effect as desired...

iloveeclipse:
> Looking at the build results this has not produced any baseline problem eve though it would be expected it seems the change has had not the same effect as desired...

This change is needed to see new compiler used for SDK: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1760

PR #1924, opened 2024-01-30
fix HandleFactory.packageHandles reuse

jukzi:
With resourcePath like "lib\jrt-fs.jar|java.base|com/sun/Class.class" the packageHandles cache was never reused since the length of the lastPkgFragmentRootPath and jarPath never matched.

With this change PackageFragment (JarPackageFragment , JrtPackageFragmentRoot) instances are reused as long as package does not change


jukzi:
example usecase: create Type Hierarchy of java.lang Object, take heapdump:
before:
image
after:
image

PR #1933, opened 2024-01-31
Sort by (CU) filename (if given) by for consistent output

laeubi:
Currently the generation of lamdas can change depending on when the source file is processed.

This sorts the array of CUs by file name (if given) to always have a predictable order independent of order given on the arguments.

Relates to https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1921


jukzi:
jenkins reports a lot of failed tests, please fix.

iloveeclipse:
With https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1922 merged I think we don't need this one.

Issue #1915, opened 2024-01-26
ASTParser.createAST fails at handling missing types


mickaelistria:
While working on DOM based operation ( #1863  ), I noticed that a DOM created with `ASTParser.createAST` is not always complete and can fail at resolving some bindings.

Inspired from ASTConverterTest15.test0238_2
```java
	public void test0238_2() throws JavaModelException, CoreException {
		this.workingCopy = getWorkingCopy("/Converter15/src/test0238/X.java",
			"""
			package test0238;
			public class X extends A {
			}
			""", true /*resolve*/);
		ASTParser parser = ASTParser.newParser(AST.getJLSLatest());
		parser.setBindingsRecovery(true);
		parser.setResolveBindings(true);
		parser.setStatementsRecovery(true);
		parser.setSource(this.workingCopy);
		parser.setProject((IJavaProject)ResourcesPlugin.getWorkspace().getRoot().getProject("Converter15").getNature(JavaCore.NATURE_ID));
		CompilationUnit unit = (CompilationUnit)  parser.createAST(null);
		TypeDeclaration typeDeclaration = (TypeDeclaration) unit.types().get(0);
		ITypeBinding superTypeBinding = typeDeclaration.resolveBinding().getSuperclass();
		IMethodBinding[] methodBindings = superTypeBinding.getDeclaredMethods();
		assertEquals("wrong size", 2, methodBindings.length);
	}
```
fails because the method referencing the unresolved type isn't included.

The expectation here would be that this piece of code returns a CompilationUnit DOM and bindings fully equivalent to what `CompilationUnit.buildStructure()` would process with the same settings (recovery).

`CompilationUnit.buildStructure` uses a combination of parse/resolve which is similar to what `createAST` does, however, it looks like `createAST` doesn't properly set the resolution context/environment before computing the bindings.

PR #1942, opened 2024-01-31
ASTParser.createAST looses ability to resolve some bindings

mickaelistria:
Keeping the lookupEnvironment intact as it can be necessary for further binding resolution.
The former "paranoia" seems not to be worried about as the referenced element in the environment is the unit, that does exist and is the root of the scope/environment. So this shouldn't consume more memory, and should cause a risk of memory leak.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1915

## What it does

Retain the necessary information in the CompilationUnitDeclaration.scope when building with ASTParser, so that all further bindings can resolve.

## How to test

Try the new `ASTConverter15Test.test0238_ASTParser` with and without the fix to see the fix turns the test ✅

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverter15Test.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverter15Test.java
@@ -27,6 +27,7 @@ import junit.framework.Test;
 
 import org.eclipse.core.resources.IFolder;
 import org.eclipse.core.resources.IResource;
+import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.jdt.core.BindingKey;
 import org.eclipse.jdt.core.IAnnotation;
@@ -7888,6 +7889,29 @@ public class ASTConverter15Test extends ConverterTestSetup {
 		typeBinding = typeBinding.getSuperclass();
 		IMethodBinding[] methodBindings = typeBinding.getDeclaredMethods();
 		assertNotNull("No method bindings", methodBindings);
+		assertEquals("wrong size", 2, methodBindings.length);
+	}
+
+	/*
+	 * https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1915
+	 */
+	public void test0238_ASTParser() throws JavaModelException, CoreException {
+		this.workingCopy = getWorkingCopy("/Converter15/src/test0238/X.java",
+			"""
+			package test0238;
+			public class X extends A {
+			}
+			""", true /*resolve*/);
+		ASTParser parser = ASTParser.newParser(AST.getJLSLatest());
+		parser.setBindingsRecovery(true);
+		parser.setResolveBindings(true);
+		parser.setStatementsRecovery(true);
+		parser.setSource(this.workingCopy);
+		parser.setProject(JavaCore.create(ResourcesPlugin.getWorkspace().getRoot().getProject("Converter15")));
+		CompilationUnit unit = (CompilationUnit) parser.createAST(null);
+		TypeDeclaration typeDeclaration = (TypeDeclaration) unit.types().get(0);
+		ITypeBinding superTypeBinding = typeDeclaration.resolveBinding().getSuperclass();
+		IMethodBinding[] methodBindings = superTypeBinding.getDeclaredMethods();
 		assertEquals("wrong size", 2, methodBindings.length);
 	}
 
--- a/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolver.java
+++ b/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolver.java
@@ -816,6 +816,10 @@ class CompilationUnitResolver extends Compiler {
 				CancelableNameEnvironment cancelableNameEnvironment = (CancelableNameEnvironment) environment;
 				cancelableNameEnvironment.printTimeSpent();
 			}
+			if (unit != null && unit.scope != null && unit.scope.environment != null
+				&& unit.scope.environment.unitBeingCompleted == null) {
+				unit.scope.environment.unitBeingCompleted = unit;
+			}
 			return unit;
 		} finally {
 			if (environment != null) {


mickaelistria:
@stephan-herrmann This changes a single line you've added with a comment about it being "paranoid" ;) Would you please have a look at the proposal here and re-discuss whether keeping the reference is fine or error-prone? I've put in the initial comment some arguments to make the change feel safe.

stephan-herrmann:
Hi @mickaelistria , have you seen https://bugs.eclipse.org/bugs/show_bug.cgi?id=150657 ? 
IIUC, that would be the systematic approach to address your root problem here. In short, ASTParser makes no guarantees about ability to resolve bindings that have not been created during the initial invocation, but by enabling clients to hold on to some kind of session/environment, the necessary context can be preserved - passing the responsibility to discard that session to the client, I guess.

Perhaps you can help Alex Boyko to make progress with his existing patch? Luckily he still seemed to be interested last time I asked.

stephan-herrmann:
And now for the specific question:

> @stephan-herrmann This changes a single line you've added with a comment about it being "paranoid" ;) Would you please have a look at the proposal here and re-discuss whether keeping the reference is fine or error-prone?

If ever `previousUnit` is non-null at this location, this would likely indicate that we are still in the middle of one of the `completTypeBindings()` methods. Changing `LookupEnvironment.unitBeingCompleted` under foot (as we are doing here) without finally restoring to the previous value would definitely break the ongoing execution of `completTypeBindings()`.

By saying "paranoia" I only expressed that perhaps this line never has any effect. But when it has an effect, that effect is direly needed.

> I've put in the initial comment some arguments to make the change feel safe.

I'm not sure what "arguments" you are referring to.

stephan-herrmann:
Finally, as you found me as the author of that line, you probably also found where it was introduced: https://bugs.eclipse.org/bugs/show_bug.cgi?id=480462 right?

Unfortunately I didn't manage to capture the original problem in a unit test, so before reverting that change someone must create a test for that bug and provide an alternative fix.

mickaelistria:
@stephan-herrmann thanks for those insights. I will investigate this issue again according to those.
However, for some context, the oddity that lea thing is that if we get some unt built with `cu.reconcile()` still has the org.eclipse.jdt.core.dom.CompilationUnit.scope.environment.unitBeingCompleted still set, and this allows further bindings to resolve. So my proposal comes from the naive idea that if it's good to keep it for `cu.reconcile()`, it should be OK to keep it for ASTParser. But I don't think those are exclusive, I hope I can rework the patch to set the `previousUnit` before going through resolution in case only 1 unit is being processed, so that it wouldn't break the fix you implemented, while allowing to fix #1915 . I hope I'll be able to create a test for the case you fixed in https://git.eclipse.org/c/jdt/eclipse.jdt.core.git/commit/?id=d6d10602861f99a2e87d2ce2ba051c7b3da388a8 to make everything feel safer.

mickaelistria:
While I didn't manage to fully understand the erroneous case of https://bugs.eclipse.org/bugs/show_bug.cgi?id=150657 to capture it in a unit test, I've reworked this patch so that it alters less the related fix, and would hopefully wouldn't re-introduce the bug.

stephan-herrmann:
> @stephan-herrmann thanks for those insights. I will investigate this issue again according to those. However, for some context, the oddity that lea thing is that if we get some unt built with `cu.reconcile()` still has the org.eclipse.jdt.core.dom.CompilationUnit.scope.environment.unitBeingCompleted still set, and this allows further bindings to resolve. 

That's an interesting observation. It seems that https://bugs.eclipse.org/bugs/show_bug.cgi?id=162478 made reconcile() behave this way for reasons that are similar to your case.

After debugging your test case, let's also state what it is you are trying to fix:
* you request AST for **exactly one** CU
* then you request `getDeclaredMethods()` from the superclass binding
* this needs to resolve signatures in those methods (where types are represented so far by `UnresolvedReferenceBinding`, which is specific to the binary case)
* one **type** in a method signature **cannot be found**
* compiler calls `isClasspathCorrect()` (aka cannot resolve _ which is indirectly referenced _)
* reporting this error doesn't have a direct source to report against (recall we are resolving a binary type)
* if by any chance, a `unitBeingCompleted` is still present then we know that this is where the error should be reported
* otherwise no location to report against and hence throwing `AbortCompilation` is the last resort.

First of all this is a very narrow situation (a situation that should better be avoided in the first place!), and fixing this will not come close to the goal of the real feature request https://bugs.eclipse.org/bugs/show_bug.cgi?id=150657 which IMHO deserves some love, still.

For a moment I was thinking the change could be tricked to use that unit completely out of scope, but as you restrict this to the situation of parsing exactly one cu, that part should be safe, considering that only type bindings from the same universe will have back pointers to the lookup environment, which keeps things well-scoped.

The change is also more restrictive than necessary, as it applies only in one branch of `resolve()`. Therefor I suggest that you move your change to a location outside those branches, perhaps close to the end of resolving the original compiler AST, and shortly before it will be converted to DOM AST. That way we'll be sure that the change won't interfere in any degree with the current compiler invocation (CompilationUnitResolver is a variant of the Compiler), and signal that the unit is saved only for *subsequent* requests.

I recall that the original JDT developers where quite concerned to free as much memory as possible before returning the unit to the caller, but at closer look the unit is already kept alive by DefaultBindingResolver, so keeping one more reference doesn't hurt.

mickaelistria:
Thank you.

> The change is also more restrictive than necessary, as it applies only in one branch of resolve(). Therefor I suggest that you move your change to a location outside those branches, perhaps close to the end of resolving the original compiler AST, and shortly before it will be converted to DOM AST. That way we'll be sure that the change won't interfere in any degree with the current compiler invocation (CompilationUnitResolver is a variant of the Compiler), and signal that the unit is saved only for subsequent requests.

I resolve that hint to around line 1000, before converter.convert(...) is called. Does that seem correct?

akurtakov:
@stephan-herrmann What do you think of the new approach from @mickaelistria ?

stephan-herrmann:
> @stephan-herrmann What do you think of the new approach from @mickaelistria ?

Yes, as a gradual improvement this looks OK. 

Just to mention the inherent limitation: at this point we are only able to "survive" the detection of a missing type, the return type of method `foo()` will be a missing type (`isRecovered() == true`). This happens _even_ when type `Zork` is actually available in the project. When the initial call `parser.createAST()` completes we still haven't touch class `Zork` and lateron we no longer have the name environment needed for discovering this class. This fundamental issue can only be solved by something in the lines of https://bugs.eclipse.org/bugs/show_bug.cgi?id=150657

PR #1944, opened 2024-02-01
Capture all information of a source file into one SourceInfo record

laeubi:
## What it does
To get a consistent compilation order that is independent of the order of arguments one needs to sort the files to process by name. As the name might be related to to other information all of these are first captured inside a SourceInfo record and are the sort this array by name, in the following code always this sorted array is used and possibly updated with additional information.

## How to test

Relates to:
- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1921

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/Main.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/Main.java
@@ -71,7 +71,6 @@ import java.util.ResourceBundle;
 import java.util.Set;
 import java.util.StringTokenizer;
 import java.util.function.Function;
-import java.util.stream.IntStream;
 
 import org.eclipse.jdt.core.compiler.CategorizedProblem;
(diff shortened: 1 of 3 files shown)


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/Main.java
```diff
@@ -1741,13 +1742,38 @@ private boolean checkVMVersion(long minimalSupportedVersion) {
...
+}
+
 /*
  *  Low-level API performing the actual compilation
  */
 public boolean compile(String[] argv) {
 	// decode command line arguments
 	try {
 		configure(argv);
+		// sort array by file names so we have a consistent order of compiling / handling them
```
laeubi, line 1767: this might even be pushed to to perform compile but as the EclipseCompiler extending this class uses some own indexed data structure this need to be carefully done in another PR...

laeubi:
@iloveeclipse can you review/merge? All checks are green now.

Issue #1946, opened 2024-02-01
Potential resource leak warning on streams from java.nio.file.Files;


jukzi:
false positives `Potential resource leak: '<unassigned Closeable value>' may not be closed` when for streams returned from java.nio.file.Files even though the stream are not resources but light weight String/Path:
```
import java.io.IOException;
import java.nio.file.Files;
import java.util.Arrays;

class A {
	A() throws Exception {
		java.util.Arrays.stream(new String[] {}); // no warning :-)
		lines(null).toList(); // no warning :-)
		java.nio.file.Files.lines(null).toList(); // stream of string -> warning :-(		
		java.nio.file.Files.walk(null, null).toList(); // stream of Path -> warning :-(
	}
	// same declaration as java.nio.file.Files.lines:
    public static java.util.stream.Stream<String> lines(java.nio.file.Path path) throws java.io.IOException {
        return null;
    }
}
```
if Potential resource leaks are enabled:
![image](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/51790620/dcc0eb68-597f-4bab-a04e-489a5109faf8)

@stephan-herrmann can you please fix?

stephan-herrmann:
In 2013 I had direct communication with Brian Goetz on this topic, when they discussed if `Stream` should implement `AutoCloseable` and how Eclipse could distinguish closeables that do / don't *require* closing ("GC-resistant resources"?). That communication included a discussion specifically concerning class `Files`.

In that discussion we destilled the following rules (mostly using Brian's words):
* Implementing `AutoCloseable` is a tag indicating that *perhaps* closing is required to prevent leaks
  * `Stream` extends `AutoCloseable`, but generally such streams do NOT need closing
      * The only streams that do need closing are the static methods in `Files` that produce streams (`lines()`, `walk()`, etc.)

So it seems we have confirmation by Brian Goetz that those Streams do require closing, not because the String elements are GC-resistant, but because the Stream internally holds a GC-resistant resource, from which the strings are computed. That part "works as designed". 

_What we failed to discuss back then_: how are those streams actually closed. We'll have to check the actual implementation of `Files` to be certain, but my guess is that any **terminal operations** on such streams (like `toList()`) should include closing the underlying resource.

Side note: Brian mentioned the idea that JSR-308 (type annotations) could provide the right tool for specifying which types need to be checked for leaks.

[...]

PR #1947, opened 2024-02-01
Enable potential resource leak analysis, fix unmanaged streams

jukzi:
fixes
"At least one of the problems in category 'resource' is not analysed due
to a compiler option being ignored"

"Potential resource leak: '<unassigned Closeable value>' may not be
closed"

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1946


stephan-herrmann:
@jukzi I saw your review request and I would have loved to discuss this, but I simply didn't find the time at short notice. Please remember that I work on Eclipse entirely in my spare time.

So you killed the idea to enforce a closing-protocol on ProblemReporter, huh? I admit this wasn't quite mature yet, but why the haste? At least the javadoc of ProblemReporter.close() should be synced with the code, either way.

jukzi:
Silence is no clear indicator - When you don't find the time to review, but plan to review later, please leave a short notice. I removed the autoclosable since 1. it did not cover a system resource 2. autocloseable enforced nothing but only gave a warning 3. it introduced so many warnings that you disabled the warnings which also removed unrelated relevant warnings. 
Please adjust javadoc as you want, or even re-add autoclosable if you can do it without disabling the warnings.

stephan-herrmann:
> Please adjust javadoc as you want

It was consistent before your change, suggesting t-w-r is wrong due to your change.

jukzi:
> suggesting t-w-r is wrong due to your change.

ok, i will fix that. thanks for the hint

jukzi:
see https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1988

PR #1948, opened 2024-02-01
suppress "CategorizedProblem implements non-API interface IProblem"

jukzi:
IProblem is API with @noimplement ... by CLIENTs only


jukzi:
image

PR #1950, opened 2024-02-01
pom.xml: remove code.ignoredWarnings

jukzi:
to get same warnings on jenkins as in workspace


jukzi:
There is a diff of 5:
image

jukzi:
those are `@SuppressWarnings("restriction")` which are now analyzed.

PR #1953, opened 2024-02-01
Content assist does not activate, inside if() within a switch

### LINKED ISSUE #1561: Content assist does not activate, inside if() within a switch

stephan-herrmann:
fixes #1561

Simpler approach, using new tests from #1625 

@gayanper I tried to explain with minimal code comments. LMK if this is sufficient to understand the approach.

This change is not quite complete yet:
* no explanation for changed relevance in three tests
* may need to cover more control statements: only implemented for those that are covered by tests from [#1625](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1625)

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests18.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests18.java
@@ -5950,8 +5950,10 @@ public void testBug574823_completeOn_methodInvocationWithParams_inIfConidtion_in
 	this.workingCopies[0].codeComplete(cursorLocation, requestor, this.wcOwner);
 
 	String result = requestor.getResults();
-    assertTrue(String.format("Result doesn't contain method forEach (%s)", result),
-    		result.contains("forEach[METHOD_REF]{forEach(), Ljava.lang.Iterable<Ljava.lang.String;>;, (Ljava.util.function.Consumer<-Ljava.lang.String;>;)V, null, null, forEach, (arg0), replace[149, 149], token[149, 149], 55}"));
+	int relevance = R_DEFAULT + R_RESOLVED + R_INTERESTING + R_CASE + R_VOID + R_NON_STATIC + R_NON_RESTRICTED;
+	assertTrue(String.format("Result doesn't contain method forEach (%s)", result),
+			result.contains("forEach[METHOD_REF]{forEach(), Ljava.lang.Iterable<Ljava.lang.String;>;, (Ljava.util.function.Consumer<-Ljava.lang.String;>;)V," +
+					" null, null, forEach, (arg0), replace[149, 149], token[149, 149], "+relevance+"}"));
 }
 public void testBug574823_completeOn_methodInvocationWithParams_inIfConidtion_insideIf_followedByChainedStatment() throws Exception {
 	this.workingCopies = new ICompilationUnit[1];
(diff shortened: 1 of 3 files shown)


gayanper:
@stephan-herrmann This approach is much elegant. I think we should go ahead with this.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests18.java
```diff
@@ -5951,7 +5951,7 @@ public void testBug574823_completeOn_methodInvocationWithParams_inIfConidtion_in
 
 	String result = requestor.getResults();
     assertTrue(String.format("Result doesn't contain method forEach (%s)", result),
-    		result.contains("forEach[METHOD_REF]{forEach(), Ljava.lang.Iterable<Ljava.lang.String;>;, (Ljava.util.function.Consumer<-Ljava.lang.String;>;)V, null, null, forEach, (arg0), replace[149, 149], token[149, 149], 60}"));
+    		result.contains("forEach[METHOD_REF]{forEach(), Ljava.lang.Iterable<Ljava.lang.String;>;, (Ljava.util.function.Consumer<-Ljava.lang.String;>;)V, null, null, forEach, (arg0), replace[149, 149], token[149, 149], 55}"));
```
gayanper, line 5954: @stephan-herrmann  Could this be due to poping out the `interestingEnclosings` ?

gayanper:
I tried to see if we are missing any control conditions, I see we don't have a endVisit for ForStatement, But even without it our unit test pass with following test snippet

```
switch (input) {
case A:
	for (int i = 0; i < nam)
	break;
}
```
Do you think we are missing any ?

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests18.java
stephan-herrmann: > @stephan-herrmann Could this be due to poping out the `interestingEnclosings` ?

To assess the change in relevance I'd first check which ingredient to this number got lost.

stephan-herrmann:
> I tried to see if we are missing any control conditions, I see we don't have a endVisit for ForStatement, But even without it our unit test pass with following test snippet
> 
> ```
> switch (input) {
> case A:
> 	for (int i = 0; i < nam)
> 	break;
> }
> ```
> 
> Do you think we are missing any ?

To parallel the original test case I'd try smth like this:
```switch (input) {
case A:
	for (int i = 0; nam|)
	    break;
}
```
i.e., completion on a _direct_ child in the ForStatement.condition location (currently you have an intermediate binary expression).

Another check for completeness: if we have a method in scope that returns boolean and has a name starting with 'nam': do we propose it with relevance `R_EXACT_EXPECTED_TYPE`? I.e. is completion correctly leveraging the information from the .condition location? I'm not sure though, if CompletionEngine is complete in this regard, yet.

gayanper:
@stephan-herrmann I added a additional test for testing the scenarios you mentioned in above comment https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1953#issuecomment-1937844419

PR #1960, opened 2024-02-05
Unnecessary generic / lamda in CharDeduplication#CacheReference

laeubi:
I'm currently analyzing high memory pressure and OOM problem in builds like seen for example here:

- https://github.com/eclipse-equinox/p2/pull/441#issuecomment-1920761676

a heapdump shows that `CharDeduplication` is a possible candidate where memory is retained, while analyzing this I noticed that there are some indirection in this class that make it hard to follow references and its unclear where a reference actually is retained to some live memory.

## What it does
CharDeduplication#CacheReference uses generics with a lamda supplier to handle a SoftReference to the cache, but this generalization/abstraction is actually not needed or used anywhere.

This removes the CacheReference and implements its functionality directly in the CharDeduplication#getThreadLocalInstance making it easier to follow the flow and not creating lamda instance indirections.

## How to test
Run Batch Compiler / JDT

## Author checklist


laeubi:
@iloveeclipse @jukzi can anyone review or suggest a reviewer? All tests are passing...

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/CharDeduplication.java
```diff
@@ -35,31 +34,17 @@ public class CharDeduplication {
...
-	}
-
-	private final static ThreadLocal<CacheReference<CharDeduplication>> mutableCache = ThreadLocal.withInitial(()->new CacheReference<>(CharDeduplication::new));
+	private final static ThreadLocal<SoftReference<CharDeduplication>> mutableCache = ThreadLocal
+			.withInitial(() -> new SoftReference<>(new CharDeduplication()));
 
 	/** @return an instance that is *not* thread safe. To be used in a single thread only. **/
 	public static CharDeduplication getThreadLocalInstance() {
-		return mutableCache.get().get();
+		CharDeduplication local = mutableCache.get().get();
```
jukzi, line 42: this looks not atomic. i got the feeling it does not need to be atomic because of threadlocal. but the charm of the former implementation was that i did not have to think about it.

jukzi (commented):
I don't see any obvious pros or cons with this implementation. can you please somehow proof/show with screenshot whats the benefit of changed implementation

jukzi:
> a heapdump shows that `CharDeduplication` is a possible candidate where memory is retained,

A Softreference can not contirbute to a OOME. i guess you are on the wrong track.

jukzi:
Can you please create a Issue for the OOME sharing a Heapdump?

laeubi:
> A Softreference can not contribute to OOME. i guess you are on the wrong track.

A softreference itself not, but it won't be cleared if its value is held directly or indirectly, e.g you use it inside a lambda expression, but that's not the issue with this PR, it is whole to remove many indirection that makes it hard to follow references to the object itself.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/CharDeduplication.java
laeubi: What do you mean by atomic? Atomic means that an action either completes successful or not happen at all including all side effects. the former implementation did exactly the same thing.

If you mean threadsafe instead, then yes this is threadsafe because it uses a Threadlocal (that previously was just wrapped in CacheReference ...

laeubi:
> I don't see any obvious pros or cons with this implementation. can you please somehow proof/show with screenshot whats the benefit of changed implementation

Benefits are:
- less code
- less indirection
- easier to debug and understand
- equal behavior

laeubi:
> As this PR seems to have no drawbacks found and makes the code easier to follow up I'm merging this one.

Thanks, with this change now in the I-Build from today I can go down to roughly `700mb` max memory without an OOM error in the P2 repository and `CharDeduplication` is not showing up any longer to retaining considerable amount of memory (only a few bytes).

PR #1863, opened 2024-01-13
Switch to make some operations rely on DOM instead of ECJ

mickaelistria:
Introduced 2 system property switches to control whether some IDE operations are using ECJ parser (legacy/default) or whether to make those operations powered by a full DOM.

* CompilationUnit.DOM_BASED_OPERATIONS will make codeSelect (hover, link to definition...), buildStructure (JDT Project element model), reconciler (code diagnostics feedback) and indexer
* CompilationUnit.DOM_BASED_COMPLETION controls completion (unlike other operations, completion based on DOM is far from being complete or straightforward)

The DOM-based operation can then allow to plug alternative compiler then ECJ and still get JDT functionalities working.

PR #1983, opened 2024-02-08
Switch to make some operations rely on DOM instead of ECJ

mickaelistria:
(continuation of https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1863 as the source repo/branch changed)

Switch to make some operations rely on DOM instead of ECJ
    
Introduced 2 system property switches to control whether some IDE
operations are using ECJ parser (legacy/default) or whether to make
those operations powered by a full DOM.
    
* CompilationUnit.DOM_BASED_OPERATIONS will make codeSelect (hover, link to definition...), buildStructure (JDT Project element model), reconciler (code diagnostics feedback); this one seems currently working ✔
* CompilationUnit.codeComplete.DOM_BASED_OPERATIONS controls completion (unlike other operations, completion based on DOM is far from being complete or straightforward) 🏗️
* SourceIndexer.DOM_BASED_INDEXER controls whether the indexation of a source document should first build/use a full DOM. This one is currently incomplete 🏗️

The DOM-based operation can then allow to plug alternative compiler then ECJ and still get JDT functionalities working.

### DIFF
--- a/.github/workflows/ci-dom-javac.yml
+++ b/.github/workflows/ci-dom-javac.yml
@@ -1,0 +1,53 @@
+name: Continuous Integration with DOM/Javac
+concurrency: 
+    group: ${{ github.workflow }}-${{ github.ref }}-dom
+    cancel-in-progress: true
+
+on:
+  push:
+    branches: [ 'dom-based-operations', 'dom-with-javac' ]
+  pull_request:
+    branches: [ 'dom-based-operations', 'dom-with-javac' ]
+
+jobs:
+  build-dom-javac:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Install xmllint
+      shell: bash
+      run: |
+        sudo apt update
+        sudo apt install -y libxml2-utils
+    - uses: actions/checkout@v4
+      with:
+        submodules: recursive
+        fetch-depth: 0
+    - name: Enable DOM-first and Javac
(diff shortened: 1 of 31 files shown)


mickaelistria:
We won't work further on making JDT's DOM-first approach fully compatible with ECJ. It would still be a nice-to-have, but the remaining issues are hard to fix in ECJ and are not a requirement to fix them in order to enable alternative compiler backends, so no further work in planned on this topic. DOM-first is now fully included with DOM with Javac branch.
If anyone still believes this DOM-first work makes sense independently of Javac, please start the discussion and we'd be happy to support anyone willing to consolidate DOM-first with ECJ, it's more or less a matter of making ~30 tests pass (but they are tricky ones).

mickaelistria:
FWIW, I'll probably rewrite this patch to not include the implementation, but just have an extension point allowing to define alternative startegies for
* model population/reconcile
* codeSelect
* codeComplete
* populateIndex

mickaelistria:
I have splitted this in 3 distinct PRs:
https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2699 , https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2700 an https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2701 so far. Completion and Search are not yet well enough covered with a DOM-first approach and will probably be contributed later.

PR #1984, opened 2024-02-08
Allow resolving compilation unit (DOM) with Javac

mickaelistria:
(continuation of https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1862 as source repo/branch changed)

Allow resolving compilation unit (DOM) with Javac
    
Extracts an interface ICompilationUnitResolver out of
CompilationUnitResolver. This interface is used by ASTParser to create
DOM instance from a Java source. The default implementation remains
CompilationUnitResolver and uses ECJ parser/resolver to create a DOM.
A system property `ICompilationUnitResolver` can be used to
force usage of alternative implementations. At the moment, those
alternative implementations need to be in the same classloader (eg in a
fragment).
    
A new fragment org.eclipse.jdt.core.javac is introduced and provides an
implementation of ICompilationUnitResolver which uses Javac "API"s to
create a JDT DOM. It can be enabled by setting
`-DICompilationUnitResolver=org.eclipse.jdt.core.dom.JavacCompilationUnitResolver`.

Javac can also be used to generate .class files while running the builder, setting `-DAbstractImageBuilder.compiler=org.eclipse.jdt.internal.javac.JavacCompiler`


.github/workflows/ci.yml
```diff
@@ -31,10 +31,12 @@ jobs:
           8
           17
           20
+          21
```
akurtakov, line 34: Would you please split out the JavaSE-21ci change in to separate PR so we can land it immediately? Please remove Java 20 while on it too.

mickaelistria: Done in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1994

mickaelistria:
Let's close this one as the result of discussion is that the JDT project isn't willing to host such ongoing development. We're now fully working in our fork at https://github.com/eclipse-jdtls/eclipse-jdt-core-incubator/tree/dom-with-javac and might consider contribute things back to JDT later. Anyone interested can consider contributing to the fork.

Issue #1955, opened 2024-02-02
[Patterns] Redesign resolution of patterns to follow natural visitation 


srikanth-sankaran:
Presently, resolution of patterns is implemented in a somewhat twisted/contorted tree visitation style. 
Some nodes are visited more than once, in weird order. Actions that would "naturally" happen during 
a node visit get pulled up ahead of time into a separate visit to a node etc.

We should clean this up so the code is easier to understand and maintain

srikanth-sankaran:
The following cleanups are to be included in this ticket:

1. Withdraw the method `org.eclipse.jdt.internal.compiler.ast.Pattern.resolveWithExpression(BlockScope, Expression)` and its overrides. Modify use sites to follow the code-style:

```
        pattern.setExpressionContext(ExpressionContext.INSTANCEOF_CONTEXT);
	pattern.setExpectedType(expressionType);
        pattern.resolveType(...)
```
2. Withdraw the method `org.eclipse.jdt.internal.compiler.ast.InstanceOfExpression.resolvePatternVariable(BlockScope)` - Pattern variables will get resolved automatically, when the holding nodes are visited in the normal course of business.
3. Merge `org.eclipse.jdt.internal.compiler.ast.RecordPattern.resolveWithExpression(BlockScope, Expression)` and `org.eclipse.jdt.internal.compiler.ast.RecordPattern.resolveTypeWithBindings(LocalVariableBinding[], BlockScope)` into one namely `org.eclipse.jdt.internal.compiler.ast.RecordPattern.resolveType(BlockScope)`
4. Eliminate component patterns (re) resolution code `org.eclipse.jdt.internal.compiler.ast.RecordPattern.setAccessorsPlusInfuseInferredType(BlockScope)`
5. Make `org.eclipse.jdt.internal.compiler.ast.RecordPattern.infuseInferredType(Scope, TypePattern, RecordComponentBinding)` produce resolved types rather than provide a resolveType method

srikanth-sankaran:
Clean ups of the same nature that are deferred to a later pass: 

1. Withdraw `org.eclipse.jdt.internal.compiler.ast.Expression.addPatternVariables(BlockScope, CodeStream)` - This should happen naturally as declaring nodes get visited.

2. Withdraw the methods `org.eclipse.jdt.internal.compiler.ast.Pattern.resolveAtType(BlockScope, TypeBinding)` and its overrides. Replace use sites with:
```
[...]

PR #2005, opened 2024-02-12
Address gratuitous behavior difference introduced by #1955

srikanth-sankaran:
## What it does

Preserve behavior from the multiple resolution world in the unified resolution world.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/RecordPattern.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/RecordPattern.java
@@ -125,6 +125,15 @@ public class RecordPattern extends TypePattern {
 			scope.problemReporter().recordPatternSignatureMismatch(this.resolvedType, this);
 			return this.resolvedType = null;
 		}
+
+		LocalVariableBinding [] bindings = NO_VARIABLES;
+		for (Pattern p : this.patterns) {
+			p.resolveTypeWithBindings(bindings, scope);
+			bindings = LocalVariableBinding.merge(bindings, p.bindingsWhenTrue());
+		}
+		for (LocalVariableBinding binding : bindings)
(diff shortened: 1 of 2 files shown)


srikanth-sankaran:
There are some interesting behaviors in compiler code base prior to commit https://github.com/eclipse-jdt/eclipse.jdt.core/commit/4447841085fc4ba5b0f248a48598f32dac612d45

To see this, follow these steps:

1. Create a local branch with R4_30 tag as the source
2. Apply this patch:
```
diff --git a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypePattern.java b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypePattern.java
index 03de844..69d62f4 100644
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypePattern.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypePattern.java
@@ -210,7 +210,7 @@
 							RecordComponentBinding rcb = components[this.index];
 							TypeVariableBinding[] mentionedTypeVariables = findSyntheticTypeVariables(rcb.type);
 							if  (mentionedTypeVariables != null && mentionedTypeVariables.length > 0) {
-								this.local.type.resolvedType = recType.upwardsProjection(scope, mentionedTypeVariables);
+								throw new RuntimeException("Not expected to reach this"); //$NON-NLS-1$
 							} else {
 								this.local.type.resolvedType = rcb.type;
 							}

```
3. Run RecordPatternTest.java

You will see that exception is never thrown!!!

Basically, resolveType gets called multiple times, all but the first call being short circuited by

```
if (this.resolvedType != null)
			return this.resolvedType;
```

In the case of record type inference, the component patterns are resolved once against the raw enclosing type which doesn't have any synthetic type variables. After inference, attempts to resolve component patterns are short circuited.

With the new unified resolution model, this flow is altered and surprising results show up

mpalat (approved):
+1

PR #2009, opened 2024-02-12
Quick Type Hierarchy:fix Progress

jukzi:
before the progress was stuck at 5%/80% almost all the time.

Testable with big type hierarchies like:
 java.lang.Object
 java.lang.Throwable
 org.eclipse.core.runtime.IAdaptable


jukzi:
before:
image

PR #1742, opened 2023-12-13
List of patterns in case statements

datho7561:
- Parse multiple Patterns in a case statement
- AST & generator changes for list of patterns
- Fix error reporting for a pattern variable + multiple patterns

Closes #1643

## What it does

Parse case statements with multiple patterns, as per JEP 443.

## How to test

Confirm the following code snippet compiles and executes as expected:

```java
Object o = "Hello, World";
switch (o) {
case String _, List _ when 1 == 1: System.out.println("it's working"); break;
default: System.out.println("it's not working");
}
```

Confirm that errors are placed on `a` and `b` in the following snippet:

```java
Object o = "Hello, World";
switch (o) {
case String a, List b when 1 == 1: System.out.println("it's working"); break;
default: System.out.println("it's not working");
}
```

## Author checklist

- [x] I have thoroughly tested my changes
- [x] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [x] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

PR #2011, opened 2024-02-13
[Patterns] Cumulative fixes for various problems

### LINKED ISSUE #92: [Patterns] Suggest identifier completion for `obj instanceof Dog`
Similarly to how identifier completion (usually derived from type) is available when declaring a variable, it would be nice to get similar completion after the `obj instanceof Dog` to completion (when using recent enough Java), would suggest `if (obj isntanceof Dog dog)`.

### LINKED ISSUE #773: [20][pattern switch] unnecessary code generated
check:
public class X {
	
		Object o = null;
		foo(new R());
	}
	@SuppressWarnings("preview")
	public static void foo(Object o) {
		switch (o) {
	    case R():                         // Multiple case labels!
	        System.out.println("R Only");
	    case S():                         // Multiple case labels!
	        System.out.println("Either S or an R");
	        break;
	    default:
	}      
	}
} 
 
record R() {}  
record S() {}              

see :
  // Method descriptor #22 (Ljava/lang/Object;)V
  // Stack: 2, Locals: 3
  public static void foo(java.lang.Object o);
     0  aload_0 [o]
     1  dup
     2  invokestatic java.util.Objects.requireNonNull(java.lang.Object) : java.lang.Object [27]
     5  pop
     6  astore_1
     7  iconst_0
     8  istore_2
     9  aload_1
    10  iload_2
    11  invokedynamic 0 typeSwitch(java.lang.Object, int) : int [33]
    16  tableswitch default: 72
          case 0: 40
          case 1: 56
    40  goto 48
    43  iconst_1
    44  istore_2
    45  goto 9
    48  getstatic java.lang.System.out : java.io.PrintStream [37]
    51  ldc <String "R Only"> [43]
    53  invokevirtual java.io.PrintStream.println(java.lang.String) : void [45]
    56  goto 64
    59  iconst_2
    60  istore_2
    61  goto 9
    64  getstatic java.lang.System.out : java.io.PrintStream [37]
    67  ldc <String "Either S or an R"> [51]
    69  invokevirtual java.io.PrintStream.println(java.lang.String) : void [45]
    72  return

Line numbers 43-45 [and consequently goto at 40) and similarly 56-61 are unnecessary. To check

### LINKED ISSUE #1171: StackOverflowException in XLargeTest#test011()
To reproduce this, you need to select the single given test name (by setting TESTS_NAMES at the top).

The SOE happens at BinaryExpression#containsPatternVariable(). The reason it happens is, this method and possibly few others that were added to BinaryExpression were not overridden in CombinedBinaryExpression and therefore fell into the SOE. Just as analyseCode(), we should have for the other methods too that propagate to left and right.

Another note is, perhaps we should limit the call containsPatternVariable() only for compliance >= 16.

### LINKED ISSUE #1595: [22] JEP 456: Unnamed Variables & Patterns
Ref: [JEP 456](https://openjdk.org/jeps/456)
Earlier JEP: [JEP 443](https://openjdk.org/jeps/443)

earlier issue: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/893 

Latest Spec Changes : [https://cr.openjdk.org/~abimpoudis/unnamed/latest](https://cr.openjdk.org/~abimpoudis/unnamed/latest)

no changes from earlier  - https://cr.openjdk.org/~abimpoudis/unnamed/jep443-20230322/specs/unnamed-jls.html

### LINKED ISSUE #1643: JEP 443: lists of patterns as a case label
i.e.

```java
record Box<T extends Ball>(T content) { }

sealed abstract class Ball permits RedBall, BlueBall, GreenBall { }
final  class RedBall   extends Ball { }
final  class BlueBall  extends Ball { }
final  class GreenBall extends Ball { }

public class A {
        Box box = null;
        switch (box) {
        case Box(RedBall _), Box(BlueBall _) when args.length == 4:
            System.out.println("hello");
            break;
        default:
            System.out.println("goodbye");
        }
    }
}
```

see the discussion here: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1517#issuecomment-1829038759

### LINKED ISSUE #1889: [Patterns] [Whitebox] Redundant calls to org.eclipse.jdt.internal.compiler.codegen.CodeStream.addVisibleLocalVariable(LocalVariableBinding)
Plant a breakpoint in `org.eclipse.jdt.internal.compiler.codegen.CodeStream.addVisibleLocalVariable(LocalVariableBinding)` and compile the following code either via a test or in inner eclipse instance launched in debug mode.

Given

```
public class X {
		Object o = new Object();
		while (o instanceof String s && o instanceof X x) {
			System.out.println(s);
			System.out.println(x); 
		}
	}
} 
```

you will see that the pattern bindings `s` and `x` are added more than once:

First time in this call stack:

```
TypeAnnotationCodeStream(CodeStream).addVisibleLocalVariable(LocalVariableBinding) line: 299	
InstanceOfExpression.addPatternVariables(BlockScope, CodeStream) line: 262	
AND_AND_Expression(BinaryExpression).addPatternVariables(BlockScope, CodeStream) line: 1831	
WhileStatement.generateCode(BlockScope, CodeStream) line: 196	
MethodDeclaration(AbstractMethodDeclaration).generateCode(ClassFile) line: 393	
MethodDeclaration(AbstractMethodDeclaration).generateCode(ClassScope, ClassFile) line: 327	
TypeDeclaration.generateCode(ClassFile) line: 759	
TypeDeclaration.generateCode(CompilationUnitScope) line: 829	
CompilationUnitDeclaration.generateCode() line: 414	

```

and then again in this call stack:

```
TypeAnnotationCodeStream(CodeStream).addVisibleLocalVariable(LocalVariableBinding) line: 299	
InstanceOfExpression.addPatternVariables(BlockScope, CodeStream) line: 262	
InstanceOfExpression.generateOptimizedBoolean(BlockScope, CodeStream, BranchLabel, BranchLabel, boolean) line: 158	
AND_AND_Expression.generateOptimizedBoolean(BlockScope, CodeStream, BranchLabel, BranchLabel, boolean) line: 226	
WhileStatement.generateCode(BlockScope, CodeStream) line: 259	
MethodDeclaration(AbstractMethodDeclaration).generateCode(ClassFile) line: 393	
MethodDeclaration(AbstractMethodDeclaration).generateCode(ClassScope, ClassFile) line: 327	
TypeDeclaration.generateCode(ClassFile) line: 759	
TypeDeclaration.generateCode(CompilationUnitScope) line: 829	
CompilationUnitDeclaration.generateCode() line: 414	

```

This is wasteful and twisted control flow.

Problem arises basically because of calls to `org.eclipse.jdt.internal.compiler.ast.Expression.addPatternVariables(BlockScope, CodeStream)` at various places. This is wrong. The recursive descent as it reaches the local declaration should add the variable.

### LINKED ISSUE #1986: [Patterns] Design smell: Pattern classes are far from object oriented
[Description to come]

### LINKED ISSUE #1999: [Patterns][records] Instanceof with record deconstruction patterns should never be flagged as unnecessary
```
interface I {
}

final class A implements I {
}

final class B implements I {
}

record R(I x, I y) {
}

public class X {
	public static boolean foo(R r) {
		if (r instanceof R(A a1, A a2))  // wrong warning here: The expression of type R is already an instance of type R 
			return true;
		return false;
	}

		System.out.println(X.foo(new R(new A(), new A())));
		System.out.println(X.foo(new R(new A(), new B())));
	}
}

```

### LINKED ISSUE #2001: [Patterns][records] ECJ fails to reject incompatible pattern types.
Javac rejects this:

```
public class X {
	record R(Long color) {
	}

		Object o = new R(10L);
		if (o instanceof R(long d)) {
			System.out.println(d);
		}
	}
}  
```

and this:

```
public class X {
	record R(long color) {
	}

		Object o = new R(10L);
		if (o instanceof R(Long d)) {
			System.out.println(d);
		}
	}
}  
```

while eclipse compiles the latter.

### LINKED ISSUE #2007: [Patterns][records][inference] ECJ never computes the upwards projection for var type inference in record patterns
This is a follow up from https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2004

JLS 14.30.1 reads:

```
The type of a pattern variable declared in a nested type pattern is determined as
follows:
• If the LocalVariableType is UnannType then the type of the pattern variable is
denoted by UnannType.
• If the LocalVariableType is var then the type pattern must appear directly in the
pattern list of a record pattern, or a compile-time error occurs.
Let R be the type of the record pattern, and let T be the type of the corresponding
component field in R (§8.10.3). The type of the pattern variable is the upward
projection of T with respect to all synthetic type variables mentioned by T.
```

If however you follow the steps in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2005#issuecomment-1938213236
you will see that the exception is never thrown - essentially the line with the call
`this.local.type.resolvedType = recType.upwardsProjection(scope, mentionedTypeVariables);` is never reached.

This was behavior as of **4.30** also of master as of date.

A few points to investigate:

1. `org.eclipse.jdt.internal.compiler.ast.RecordPattern.resolveType(BlockScope)` reads:

```
              if (this.resolvedType.isRawType()) {
			TypeBinding expressionType = expectedType();
			if (expressionType instanceof ReferenceBinding) {
				ReferenceBinding binding = inferRecordParameterization(scope, (ReferenceBinding) expressionType);
				if (binding == null || !binding.isValidBinding()) {
					scope.problemReporter().cannotInferRecordPatternTypes(this);
				    return this.resolvedType = null;
				}
				this.resolvedType = binding;
			}
		}
```

I suspect the last line should actually be `this.resolvedType = binding.capture(scope, this.sourceStart, this.sourceEnd);`

2. I think the capture call down below in the same method is misplaced.
3. There _could_ also be something wrong with `org.eclipse.jdt.internal.compiler.lookup.ReferenceBinding.upwardsProjection(Scope, TypeBinding[])`  - I found some strange results in experiments but didn't capture (!) (!!) them :)

srikanth-sankaran:
## What it does

* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1889
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1986
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2019
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2017
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2021
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2020
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2001
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1999
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/773
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2043
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/92
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2044
* Fixes https://bugs.eclipse.org/bugs/show_bug.cgi?id=575070
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2045
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2053
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2007
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1643
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1595
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2069
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2077
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1171

## Author checklist


mpalat (approved):
+1
merge please

mpalat (commented):
sorry - approved by mistake -
will wait until it is fully ready

srikanth-sankaran:
Hello David (@datho7561) - I am mostly done with cleaning up the implementation. This branch contains a bunch of fixes that were late for 4.31 and so are being accumulated here. Once master opens for 4.32 development, (March 2nd) I will merge these in.

The most recent two commits concern your past work - you are welcome to review them and report any problems. I invite you to review all the changes not just those two commit if time permits.

Where are you on https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1742 ?? If you are able to bring it up to date with this branch (not master), I can look into absorbing it. Is that PR fully functional ? What is missing if anything ?

datho7561:
I'll spend some time rebasing that PR onto this one. Most of my changes for the parsing aspect shouldn't cause conflicts. I'll need to add my code generation changes, and that's where the conflicts will be. I think your refactoring will make the code generation code simpler.

srikanth-sankaran:
Hi David (@datho7561) the most recent commit may impact you too - but should be in a good way overall.

srikanth-sankaran:
@mpalat & @jarthana - You are welcome to look through these changes - The commits are isolated one per issue. 

I plan to merge this by March 2nd when the master branch opens for 4.32

srikanth-sankaran:
Jay reported a behavior change in these snippets relative to master:

(1) ```
public class X {
	public Object a0 = new Boolean(false);

	public boolean testMethod() {
		Object ax = a0;
		boolean res = false;

		if (ax instanceof Boolean _) {
			res = true;
		}
		return res;
	}

		X c = new X();
		c.testMethod();
	}
}
```

(2) ```
class X {
	static public Object a0 = "a";

		String res = "";
		Object ax = a0;

		if ( (! (ax instanceof String a)) && (ax instanceof String a) ) {
			res += "t" + a; // after then
		} else {
			res += "e" + ""; // after else
		}
		if (!res.equals("e")) {
			System.out.println("Got: " + res + ", expected: e");
		}
	}
}
```

and 

(3) ```
class X {
	record CoffeeBreak() {}
	    public int recharge(CoffeeBreak c) {
	        int energyLevel = 0;
	        switch (c) {
	            case CoffeeBreak( ) -> {
	                energyLevel = 3; 
	            }
	            default->{
	                energyLevel = -3; 
	            }
	        }
	        return energyLevel;
	    }
	        X t = new X();
	        CoffeeBreak c = new CoffeeBreak();
	        if (t.recharge(c) == 3) {
	        }
	    }
}
```

Investigating ...

**Comment by srikanth-sankaran** [comment]
Manoj reported a problem with conditionals and negation that results in generated class file failing to verify. Investigating ...

```
public class X {
						  public static void main(String[] o) {
								foo("one");
								foo(new X());
							}
						  public static void foo(Object o) {
							  	boolean b  = (o instanceof String) ? (o instanceof String a) : (! (o instanceof String  a));
							  	System.out.println(b);
							}
						}
```

**Comment by srikanth-sankaran** [comment]
@datho7561 - Why did you feel it necessary to change the grammar ? At first glance, it looks like the existing grammar is perfectly capable of parsing multiple patterns in a case label ? Did I miss something ? 

In fact I would argue that the old grammar is significantly better by being permissive and tolerant of mixing illegal case labels combinations and allowing later phases to issue more meaningful messages rather than just complaining about syntax errors. The new grammar syntactically constrains illegal case label combinations - i.e. the permissiveness of earlier grammar goes away. 

Most of the changed tests in `SwitchPatternTest` illustrate this - quality of error messages has degraded significantly.

I think we should restore the grammar and adjust rest of the code accordingly.

**Comment by srikanth-sankaran** [comment]
> @datho7561 - Why did you feel it necessary to change the grammar ? At first glance, it looks like the existing grammar is perfectly capable of parsing multiple patterns in a case label ? Did I miss something ?
> 
> In fact I would argue that the old grammar is significantly better by being permissive and tolerant of mixing illegal case labels combinations and allowing later phases to issue more meaningful messages rather than just complaining about syntax errors. The new grammar syntactically constrains illegal case label combinations - i.e. the permissiveness of earlier grammar goes away.
> 
> Most of the changed tests in `SwitchPatternTest` illustrate this - quality of error messages has degraded significantly.
> 
> I think we should restore the grammar and adjust rest of the code accordingly.
> 
> Sometimes, it is better for the grammar to accept illegal programs and as long as they are weeded out in resolution phase.

I am working on this - have about 30 failures to address. Will push a commit once all issues are resolved. Stay tuned.

**Comment by srikanth-sankaran** [comment]
@datho7561, So sorry I gave you bad instructions - I am told by @mpalat and @jarthana that this needs to go into BETA_JAVA22 branch which means it is to be based on 4.31 RC1 work and not on top of all the accumulated changes here. So the earlier request to rebase against this branch is bad - apologies.

I will take your work from https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1742 and create a PR for BETA_JAVA22 branch. Hopefully not much work may be there.

**Comment by srikanth-sankaran** [comment]
> @datho7561, So sorry I gave you bad instructions - I am told by @mpalat and @jarthana that this needs to go into BETA_JAVA22 branch which means it is to be based on 4.31 RC1 work and not on top of all the accumulated changes here. So the earlier request to rebase against this branch is bad - apologies.
> 
> 
> 
> I will take your work from https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1742 and create a PR for BETA_JAVA22 branch. Hopefully not much work may be there. 

Discussing the best way forward on this. One way or other I'll help steer this to delivery

**Comment by srikanth-sankaran** [comment]
> > @datho7561, So sorry I gave you bad instructions - I am told by @mpalat and @jarthana that this needs to go into BETA_JAVA22 branch which means it is to be based on 4.31 RC1 work and not on top of all the accumulated changes here. So the earlier request to rebase against this branch is bad - apologies.
> > I will take your work from #1742 and create a PR for BETA_JAVA22 branch. Hopefully not much work may be there.
> 
> Discussing the best way forward on this. One way or other I'll help steer this to delivery

Latest plan is to first fix this in master (so here - this branch is to be merged to master as soon as 4.32 master opens) and pick up the changes for Java 22.

Accordingly, I have pushed the work from the PR https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1742 here as a separate commit and on top of it a commit to roll back the grammar, parser and scanner changes to experiment with existing grammar's capabilities themselves to handle multiple case patterns.

This fails about 40 tests in RVPT. Investigating ...

**Comment by srikanth-sankaran** [comment]
> > > @datho7561, So sorry I gave you bad instructions - I am told by @mpalat and @jarthana that this needs to go into BETA_JAVA22 branch which means it is to be based on 4.31 RC1 work and not on top of all the accumulated changes here. So the earlier request to rebase against this branch is bad - apologies.
> > > I will take your work from #1742 and create a PR for BETA_JAVA22 branch. Hopefully not much work may be there.
> > 
> > 
> > Discussing the best way forward on this. One way or other I'll help steer this to delivery
> 
> Latest plan is to first fix this in master (so here - this branch is to be merged to master as soon as 4.32 master opens) and pick up the changes for Java 22.
> 
> Accordingly, I have pushed the work from the PR #1742 here as a separate commit and on top of it a commit to roll back the grammar, parser and scanner changes to experiment with existing grammar's capabilities themselves to handle multiple case patterns.
> 
> This fails about 40 tests in RVPT. Investigating ...

Down to 10 test failures in RVPT

**Comment by srikanth-sankaran** [comment]
> > @datho7561 - Why did you feel it necessary to change the grammar ? At first glance, it looks like the existing grammar is perfectly capable of parsing multiple patterns in a case label ? Did I miss something ?
> > In fact I would argue that the old grammar is significantly better by being permissive and tolerant of mixing illegal case labels combinations and allowing later phases to issue more meaningful messages rather than just complaining about syntax errors. The new grammar syntactically constrains illegal case label combinations - i.e. the permissiveness of earlier grammar goes away.
> > Most of the changed tests in `SwitchPatternTest` illustrate this - quality of error messages has degraded significantly.
> > I think we should restore the grammar and adjust rest of the code accordingly.
> > Sometimes, it is better for the grammar to accept illegal programs and as long as they are weeded out in resolution phase.
> 
> I am working on this - have about 30 failures to address. Will push a commit once all issues are resolved. Stay tuned.
> 
> Also looks like in `org.eclipse.jdt.core.tests.compiler.regression.SwitchPatternTest21` something is amiss ? I am working with the old grammar, old Parser and old Scanner but all tests in `org.eclipse.jdt.core.tests.compiler.regression.SwitchPatternTest21` pass 😆 🤣

Turns out this is due to SwitchPatternTest21 not being hooked into `org.eclipse.jdt.core.tests.compiler.regression.TestAll`

**Comment by srikanth-sankaran** [comment]
@datho7561 - what exactly is `org.eclipse.jdt.core.tests.model.JavaSearchBugs21Tests.testListOfPatterns()` supposed to test ? 
That test case looks wrong to me: 

The wording of the draft spec viz:

```
For a case label with case patterns, it is a compile-time error if any of its case patterns declares one or more pattern variables.

If a case label had more than one case pattern that declared pattern variables, then it would not be clear which variable would be initialized if the case label were to apply. For example, in the code fragment switch (obj) { case Integer i, Boolean b -> { ... } ... } it is not clear in the block to the right of the -> which of the pattern variables i or b will be initialized, so this is a compile-time error. The fragment switch(obj) { case Integer i, Boolean _ -> { ... } ... } still results in a compile-time error, as it is still not clear if the pattern variable i will be initialized. The fragment switch (obj) { case Integer _, Boolean _ -> { ... } ... } does not result in a compile-time error.
```

should also apply to record patterns. Any reason why not ? 

Javac refuses to compile the program with: `illegal fall-through from a pattern`. With my changes we reject it with `Named pattern variables are not allowed here`

**Comment by srikanth-sankaran** [comment]
@datho7561 - I have pushed an alternate implementation that doesn't require any grammar changes. Tests are green. You are welcome to take a look, play around and report any issues. TIA

**Comment by srikanth-sankaran** [comment]
@datho7561 it may only be an academic curiosity exercise but the code you had CaseStatement.java that manages multiple patterns is problematic - it locally loops over candidates and does a checkcast as it advances through the patterns - this check cast may fail with a CCE. 

In the alternate implementation we have now, we do the more orthodox thing of beaching to the restart label for the type switch bootstrap. The table switch branch label will be entered only with the right instances so checkcast won't fail with CCE ever. 

FYI

File: org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/GuardedPattern.java [hunk 1]
```diff
@@ -27,121 +27,82 @@
 
 public class GuardedPattern extends Pattern {
 
-	public Pattern[] patterns;
+	public Pattern primaryPattern;
```
datho7561, line 30: I don't think having the guard connected to only one pattern is a good idea. When working with case statements, the user of the AST needs to know that they should check the latest pattern in the list of patterns for a guard, and will need to proceed differently based on that result.

Ideally, we should have the guard as a child of the case statement to match up with how the spec is written.

The reason I placed the list of patterns as a child of GuardedPattern instead of under CaseStatement was to avoid deprecating GuardedPattern and reduce the amount of changes I needed to do to code generation. Maybe this was a bad idea, since code generation needed to be rewritten anyways.

datho7561 (commented):
The changes are working well from my testing.

I don't feel great about not changing the AST, since the AST for guards in case statements will continue to be very different to how its described in the spec. This AST will not be intuitive to consume.

Is the plan to merge this PR or some part of this PR for RC2? From my understanding, this contains important fixes to code generation for the existing pattern features (i.e. for non preview features wrong/broken bytecode is generated), and I don't see why we should hold off on fixing it until releasing Java 22 support.

srikanth-sankaran:
I agree that there is an abstraction gap here in how multi case patterns are modeled. I was not happy with lumping things under GuardedPattern but what we have now is not ideal either. 

I initially introduced a new class EitherOrMultiPattern that can stand on its own when there is no declared guard or be the primaryPattern of a GuardedPattern but later in order to minimize the changes and more importantly to get it to work with the current implementation of CaseStatement (which badly needs a rewrite) opted for the present solution. 

I was worried about the work ballooning out of control. But now having ironed out all the wrinkles and gotten a fully working version I feel upto tackling that problem. 

This is very valuable feedback. I'll follow up. Thanks David! 😊

srikanth-sankaran:
> Is the plan to merge this PR or some part of this PR for RC2? From my understanding, this contains important fixes to code generation for the existing pattern features (i.e. for non preview features wrong/broken bytecode is generated), and I don't see why we should hold off on fixing it until releasing Java 22 support.

You ask an important question.

It would be a _very_ unorthodox move to push all this for RC2 - In the _normal scheme of things_, RC2 is for fixes for ship stopper defects : What we have here is not a ship stopper fix, but an altogether new SHIP - Patterns 2.0! 😆 (At least it resurrects the age old metaphysical question unanswered till date - see https://en.wikipedia.org/wiki/Ship_of_Theseus)

Anticipating code release guidelines tightening closer to the release cycle, I had compartmentalized the work and already significant portions were released for M2 and M3. So it is not like 4.31 won't benefit from this work. 

I'll defer to the elders of the project in this matter: @iloveeclipse @jarthana @mpalat @stephan-herrmann - FYI.

Personally I am OK with the current plan of releasing this on day 1 of 4.32 M1 and having it picked up for Java 22 branch work.

stephan-herrmann:
> It would be a _very_ unorthodox move to push all this for RC2

indeed so

I'm certain that these changes are of good quality, but there is no such thing as a risk-free change, at least not in a compiler. So it would set a bad example to rush significant changes such late in the game, as long as the product is viable without the fix :)

srikanth-sankaran:
> I was worried about the work ballooning out of control. But now having ironed out all the wrinkles and gotten a fully working version I feel upto tackling that problem.

Walking that path again, I realize why I backed out: the amount of work involved in preparing `CaseStatement` to work with `EitherOrMultiPattern` is non-trivial :) but it will leave us in good shape for Switches 2.0 😆 - so worth the investment

srikanth-sankaran:
> > I was worried about the work ballooning out of control. But now having ironed out all the wrinkles and gotten a fully working version I feel upto tackling that problem.
> 
> Walking that path again, I realize why I backed out: the amount of work involved in preparing `CaseStatement` to work with `EitherOrMultiPattern` is non-trivial :) but it will leave us in good shape for Switches 2.0 😆 - so worth the investment

The challenge seems to be in getting `CaseStatement` to work with `EitherOrMultiPattern` without peeling/peeking into the constituent patterns. It has no business to, but the way it is coded it expects to know each member pattern.

srikanth-sankaran:
I have pushed an implementation based on `EitherOrMultiPattern` - still many rough edges. I'll work on polishing these tomorrow.
I have tried to minimize the places CaseStatement/SwitchStatement look under the hood - It is impossible to completely eliminate these as there are too many low level tasks going on in laying out boostrap section entries, table switch layout etc. 

@datho7561 - do you have the cycles to work on the DOM/AST and ASTRewrite changes  for Java 22 ? We have limited time to get this done - March 19th is JDK GA.

At this point, I am not looking for low level feedback on the code - there is too much polish work that is pending. But high level comments welcome.

I am not 100% satisfied - but working with CaseStatement without completely rewriting calls for pragmatic compromises,

srikanth-sankaran:
I squashed the 3 commits for multiple patterns in case labels and pushed it for testing.

Files still needing review, verification and polish: 

```
org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/EitherOrMultiPattern.java
org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/GuardedPattern.java
org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Pattern.java
org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/RecordPattern.java
org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypePattern.java
org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/CaseStatement.java
org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/SwitchStatement.java
org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/SwitchPatternTest.java
org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/SwitchPatternTest21.java
```

srikanth-sankaran:
> There is still one test failing in formatter - but that is because there is no DOM/AST support yet. Not having heard interest there, I will take it up now.

This test is passing now after adjusting for white space - not happy to change the test, but have to halt this work somewhere.

srikanth-sankaran:
@jarthana Could you please revie this commit that adds a new DOM/AST node type EitherOrMultiPattern (compiler AST name is identical) : https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2011/commits/ef8a2a7231097117f91033ea5eed38c269d83bd9

It is mostly boiler plate but being API, I would like a second pair of eyes to look through

PR #1816, opened 2024-01-05
Fix nested pattern code generation

datho7561:
## What it does
- Fixes boolean assignments that use instanceof with nested patterns.
- Fixes cases labels that use nested patterns.

Fixes #1804

## How to test

See the sample snippets that reproduce the error in the linked issue.

## Author checklist

- [x] I have thoroughly tested my changes
- [x] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [x] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

Issue #1985, opened 2024-02-09
[Patterns][records] ECJ fails to generate code to deconstruct record in pattern


srikanth-sankaran:
The following program when compiled and run should trigger a `MatchException` caused by `java.lang.ArithmeticException` but fails to - because ECJ does not generate code to deconstruct the pattern:

```
public class X {

	record R(int x) {
		public int x() {
			return 100 / this.x;
		}
	}

	public static void main(String[] args) {
		try {
			boolean b = new R(0) instanceof R(int i);
		} catch (Throwable t) {
			System.out.println(t.getClass().getName());
		}
	}
}
```

javac handles it correctly.

But javac behavior on this is suspect:

```
public class X {

	boolean b = new R(0) instanceof R(int i);
	record R(int x) {
		public int x() {
			return 100 / this.x;
		}
	}

	public static void main(String[] args) {
            try {
        	new X();
            } catch (Throwable t) {
        	System.out.println(t.getClass().getName());
            }
	}
}



```

This with javac reports `java.lang.ArithmeticException` unwrapped by `MatchException` which I think is a bug in javac

PR #2016, opened 2024-02-13
Fix RecordPattern generation for standalone instanceof

### LINKED ISSUE #1804: Record pattern matching with `null` gives wrong result
Spotted by Jay in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1517#issuecomment-1859821498

Given the following code:
```java
public class BooleanVariable {
	record Paper(int color) {}
	record Box<T>(T a) {}
		Box<?> b = new Box<>(null);
		boolean res = b instanceof Box(Paper a);
		if (res) {
			System.out.println("res is true");
		} else {
			System.out.println("res is false");
		}
	}
}
```

ECJ prints "res is true" while javac prints "res is false".

datho7561:
Supercedes #1816

Signed-off-by: David Thompson <davthomp@redhat.com>

## What it does
Fixes #1804

## How to test

Try running the snippets in the linked issue.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/InstanceOfExpression.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/InstanceOfExpression.java
@@ -116,22 +116,30 @@ public void generateCode(BlockScope currentScope, CodeStream codeStream, boolean
 		codeStream.addVariable(this.secretExpressionValue);
 	}
 	codeStream.instance_of(this.type, this.type.resolvedType);
-	if (this.elementVariable != null) {
-		BranchLabel actionLabel = new BranchLabel(codeStream);
-		codeStream.dup();
-		codeStream.ifeq(actionLabel);
+	if (this.pattern != null) {
+		BranchLabel falseLabel = new BranchLabel(codeStream);
+		BranchLabel trueLabel = new BranchLabel(codeStream);
+		BranchLabel continueLabel = new BranchLabel(codeStream);
+		codeStream.ifeq(falseLabel);
+
 		if (this.secretExpressionValue != null) {
 			codeStream.load(this.secretExpressionValue);
 			codeStream.removeVariable(this.secretExpressionValue);
 		} else {
 			this.expression.generateCode(currentScope, codeStream, true);
 		}
-		codeStream.checkcast(this.type, this.type.resolvedType, codeStream.position);
-		this.elementVariable.binding.recordInitializationStartPC(codeStream.position);
-		codeStream.store(this.elementVariable.binding, false);
-		codeStream.removeVariable(this.elementVariable.binding);
-		codeStream.recordPositionsFrom(codeStream.position, this.sourceEnd);
-		actionLabel.place();
+		this.pattern.generateOptimizedBoolean(currentScope, codeStream, trueLabel, falseLabel);
+
+		for (LocalVariableBinding patternVariableBinding : this.pattern.bindingsWhenTrue()) {
+			codeStream.removeVariable(patternVariableBinding);
+		}
+		trueLabel.place();
+		codeStream.iconst_1();
+		codeStream.goto_(continueLabel);
+		falseLabel.place();
+		codeStream.iconst_0();
+		continueLabel.place();
+
 	}
 	if (valueRequired) {
 		codeStream.generateImplicitConversion(this.implicitConversion);
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/InstanceOfExpression.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/InstanceOfExpression.java
@@ -130,13 +130,13 @@ public void generateCode(BlockScope currentScope, CodeStream codeStream, boolean
 		}
 		this.pattern.generateOptimizedBoolean(currentScope, codeStream, trueLabel, falseLabel);
 
-		for (LocalVariableBinding patternVariableBinding : this.pattern.bindingsWhenTrue()) {
-			codeStream.removeVariable(patternVariableBinding);
-		}
 		trueLabel.place();
 		codeStream.iconst_1();
 		codeStream.goto_(continueLabel);
 		falseLabel.place();
+		for (LocalVariableBinding patternVariableBinding : this.pattern.bindingsWhenTrue()) {
+			codeStream.removeVariable(patternVariableBinding);
+		}
 		codeStream.iconst_0();
 		continueLabel.place();
 
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/InstanceOfExpression.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/InstanceOfExpression.java
@@ -134,8 +134,8 @@ public void generateCode(BlockScope currentScope, CodeStream codeStream, boolean
 		codeStream.iconst_1();
 		codeStream.goto_(continueLabel);
 		falseLabel.place();
-		for (LocalVariableBinding patternVariableBinding : this.pattern.bindingsWhenTrue()) {
-			codeStream.removeVariable(patternVariableBinding);
+		for (LocalVariableBinding binding : this.pattern.bindingsWhenTrue()) {
+			binding.recordInitializationEndPC(codeStream.position);
 		}
 		codeStream.iconst_0();
 		continueLabel.place();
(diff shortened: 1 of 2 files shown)


datho7561:
@srikanth-sankaran here is the updated fix for the bug

srikanth-sankaran:
Thanks David. These changes will be useful - I am still deciding whether this needs to go into 4.31 or be a part of the dozen or more cumulative fixes I expected to accumulate under the umbrella https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2011 and be merged in when master reopens for 4.32 M1 development (first week of March 2024)

The changes look in the right direction, but I am not done cleaning up the current implementation! In particular reference to the region you are touching this is my plan:

1. See if meat of `org.eclipse.jdt.internal.compiler.ast.InstanceOfExpression.generateCode(BlockScope, CodeStream, boolean)` and `org.eclipse.jdt.internal.compiler.ast.InstanceOfExpression.generateOptimizedBoolean(BlockScope, CodeStream, BranchLabel, BranchLabel, boolean)` can be extracted into a single method and called from bother places or alternately, see if one can call the other.
2. Remove this code from `org.eclipse.jdt.internal.compiler.ast.InstanceOfExpression.generateOptimizedBoolean`
```
if (this.elementVariable == null && this.pattern == null) {
		super.generateOptimizedBoolean(currentScope, codeStream, trueLabel, falseLabel, valueRequired);
		return;
}
```
and stop the buck right there.
3. Make the code more OO: (https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1986)
This fragment for example in  `org.eclipse.jdt.internal.compiler.ast.InstanceOfExpression.generateOptimizedBoolean` is an anti-pattern:
```
	if (this.pattern instanceof RecordPattern) {
		this.pattern.generateOptimizedBoolean(currentScope, codeStream, trueLabel, nextSibling);
	} else {
		codeStream.checkcast(this.type, this.type.resolvedType, codeStream.position);
		codeStream.store(this.elementVariable.binding, false);
	}
```

this should simply be:

```
this.pattern.generateOptimizedBoolean(currentScope, codeStream, trueLabel, nextSibling);
```
4. Get rid of the calls to `addPatternVariables` - that is due to wrong design (see https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1889)

Your changes will be useful, they just need to be adjusted for the other changes, I'll work on it.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/InstanceOfExpression.java
```diff
@@ -116,22 +116,30 @@ public void generateCode(BlockScope currentScope, CodeStream codeStream, boolean
...
-		this.elementVariable.binding.recordInitializationStartPC(codeStream.position);
-		codeStream.store(this.elementVariable.binding, false);
-		codeStream.removeVariable(this.elementVariable.binding);
-		codeStream.recordPositionsFrom(codeStream.position, this.sourceEnd);
-		actionLabel.place();
+		this.pattern.generateOptimizedBoolean(currentScope, codeStream, trueLabel, falseLabel);
+
+		for (LocalVariableBinding patternVariableBinding : this.pattern.bindingsWhenTrue()) {
+			codeStream.removeVariable(patternVariableBinding);
+		}
```
srikanth-sankaran, line 135: This is wrong code. You are perhaps following the problematic coding pattern prevalent in the code base that I am trying to weed out. In pattern matching implementation, variables are added and removed without much rhyme or reason. In this place, instanceof is true and the pattern binding is live - it should not be removed. The flow analysis should determine when it becomes not definitely assigned and close the ranges with a call to record to `org.eclipse.jdt.internal.compiler.lookup.LocalVariableBinding.recordInitializationEndPC(int)`

srikanth-sankaran: Note however the call to `codeStream.removeVariable(this.secretExpressionValue);` just above in line 127 is correct. We are done with the use of secret variable and it can be done away with. It being a compiler injected variable, its scope and live range is entirely upto us. We want to keep it alive for the minimum range required.

But pattern bindings are programmer defined variables and their scope and live ranges (pc intervals where are they are definitely assigned per flow scope) are spelled out by JLS. analyzeCode() phase computes the definite assignment ranges for locals.

srikanth-sankaran (changes requested):

srikanth-sankaran:
BTW, you may find this aggregate junit `org.eclipse.jdt.core.tests.RunVariousPatternsTests` useful - I use as a smoke test before using the Jenkins automation

srikanth-sankaran:
OK, I have made the change for live range adjustment and also added regression tests for https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1985 - which passes with this PR - and pushed. Let us see if we can get this included for M3, failing that for RC1.

However there is some infra issue with Jenkins it seems, let us see.

srikanth-sankaran:
@mpalat @jarthana @iloveeclipse - I would like this included for 4.31 RC1, if not M3 which may be a bit late I agree

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/InstanceOfExpression.java
```diff
@@ -116,22 +116,30 @@ public void generateCode(BlockScope currentScope, CodeStream codeStream, boolean
...
+		trueLabel.place();
+		codeStream.iconst_1();
+		codeStream.goto_(continueLabel);
+		falseLabel.place();
+		codeStream.iconst_0();
```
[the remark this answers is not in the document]
srikanth-sankaran: Not the for loop as is - but a modified version should be placed here. 

The variable should not be removed. Its definite assignment status must be adjusted and the live range closed.

Variable removable for user scoped variables should happen via `org.eclipse.jdt.internal.compiler.codegen.CodeStream.exitUserScope(BlockScope, Predicate<LocalVariableBinding>)` which will happen already at the right places

srikanth-sankaran (approved):
Changes look good to me. However since I also contributed, I'll ask Manoj for an additional review

srikanth-sankaran:
> Changes look good to me. However since I also contributed, I'll ask Manoj for an additional review

While the changes are correct per my review and testing, there is room for further cleanup and improvement. The original code prior to this PR itself has suffered from copy + paste and so has some strange behavior which this PR persists with:

1. It is odd to populate the operand stack with `true` or `false` with `iconst_1/iconst_0` and then `pop` it `if (!valueRequired)` 😆 
2. I think the call to `codeStream.generateImplicitConversion(this.implicitConversion);` towards the end is bogus. Already the operand stack has a boolean and we are supposed to produce a boolean - so what conversion is possible/required there ? See that the version of InstanceOfExpression.java prior to the start of pattern matching work (commit f6792492f110837cda3b566a10e36beea4e54b03)  has all of this for method body: 

```
@Override
public void generateCode(BlockScope currentScope, CodeStream codeStream, boolean valueRequired) {
	int pc = codeStream.position;
	this.expression.generateCode(currentScope, codeStream, true);
	codeStream.instance_of(this.type, this.type.resolvedType);
	if (valueRequired) {
		codeStream.generateImplicitConversion(this.implicitConversion);
	} else {
		codeStream.pop();
	}
	codeStream.recordPositionsFrom(pc, this.sourceStart);
}
```

I'll clean up this up later along with all the tasks mentioned in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2016#issuecomment-1942909619

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/InstanceOfExpression.java
```diff
@@ -116,22 +116,30 @@ public void generateCode(BlockScope currentScope, CodeStream codeStream, boolean
...
-		codeStream.checkcast(this.type, this.type.resolvedType, codeStream.position);
-		this.elementVariable.binding.recordInitializationStartPC(codeStream.position);
-		codeStream.store(this.elementVariable.binding, false);
-		codeStream.removeVariable(this.elementVariable.binding);
-		codeStream.recordPositionsFrom(codeStream.position, this.sourceEnd);
-		actionLabel.place();
+		this.pattern.generateOptimizedBoolean(currentScope, codeStream, trueLabel, falseLabel);
+
+		trueLabel.place();
+		codeStream.iconst_1();
```
mpalat, line 134: codeStream.dup() in the original code should have been kept instead of const_1() and const_0() [later]. However, not a critical change.

mpalat (approved):
+1
Am fine with this

mpalat:
> @mpalat @jarthana @iloveeclipse - I would like this included for 4.31 RC1, if not M3 which may be a bit late I agree

+1 for M3 -
(a) changes look fine
(b) in case there are issues, we can address in RC1

srikanth-sankaran:
Thanks @mpalat and @datho7561

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/InstanceOfExpression.java
srikanth-sankaran: > codeStream.dup() in the original code should have been kept instead of const_1() and const_0() [later]. However, not a critical change.

I will follow up on this along with various other cleanups I have called out in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2016#issuecomment-1942909619

Issue #2036, opened 2024-02-16
Type signatures returned through bindings don't match signatures returned through codeSelect


mickaelistria:
Add to TypeResolverTest
```java
// derived from test531046a, verifying we get same value from codeSelect or bindings
public void testJavaElementViaBindings() throws CoreException, IOException {
	if (!isJRE9) return;
	try {
		createJava10Project("P", new String[] {"src"});
		createFolder("/P/src/p");
		createFile("/P/src/p/X.java", "");
		waitForAutoBuild();

		org.eclipse.jdt.internal.core.CompilationUnit unit = (org.eclipse.jdt.internal.core.CompilationUnit)getCompilationUnit("/P/src/p/X.java");
		unit.becomeWorkingCopy(new ProblemRequestor(), null);
		unit.getBuffer().setContents("""
			package p;
			public class X {
			  public static void main(java.lang.String[] args) {
			    var s1 = args[0];
			    System.out.println(s1);
			  }
			}
			""");
		CompilationUnit dom = unit.makeConsistent(AST.getJLSLatest(), true, 0, null, null);
		VariableDeclarationFragment domVariable = (VariableDeclarationFragment)
			((VariableDeclarationStatement)
				((TypeDeclaration)dom.types().get(0)).getMethods()[0].getBody().statements().get(0))
			.fragments().get(0);
		ILocalVariable variable = (ILocalVariable)domVariable.resolveBinding().getJavaElement();
		assertEquals("incorrect type", "Ljava.lang.String;", variable.getTypeSignature());
	} finally {
		deleteProject("P");
	}
}
```
and try to run it:
```
junit.framework.ComparisonFailure: incorrect type.
----------- Expected ------------
Ljava.lang.String;
------------ but was ------------
Ljava/lang/String;
---------------------- ----------
 expected:<Ljava[.lang.]String;> but was:<Ljava[/lang/]String;>
	at org.eclipse.jdt.core.tests.junit.extension.TestCase.assertStringEquals(TestCase.java:267)
	at org.eclipse.jdt.core.tests.junit.extension.TestCase.assertEquals(TestCase.java:242)
	at org.eclipse.jdt.core.tests.model.TypeResolveTests.testJavaElementViaBindings(TypeResolveTests.java:1457)
...
```

[...]

Issue #1094, opened 2023-05-26
testCopyConstructorInDifferentProject failure


mpalat:
Intermittent failure
https://download.eclipse.org/eclipse/downloads/drops4/I20230525-0350/testresults/html/org.eclipse.jdt.core.tests.model_ep428I-unit-mac64-java17_macosx.cocoa.x86_64_17.html

No delta found for class Y [in Y.java [in <default> [in src [in P2]]]] Y(String), all deltas: [Java Model[*]: {CHILDREN} P2[*]: {CHILDREN} src[*]: {CHILDREN} <default>[*]: {CHILDREN} Y.java[*]: {CONTENT | PRIMARY RESOURCE}]

junit.framework.AssertionFailedError: No delta found for class Y [in Y.java [in <default> [in src [in P2]]]]
Y(String), all deltas: [Java Model[*]: {CHILDREN}
P2[*]: {CHILDREN}
src[*]: {CHILDREN}
<default>[*]: {CHILDREN}
Y.java[*]: {CONTENT | PRIMARY RESOURCE}]
at junit.framework.Assert.fail(Assert.java:57)
at junit.framework.TestCase.fail(TestCase.java:223)
at org.eclipse.jdt.core.tests.model.CopyMoveTests.copyPositive(CopyMoveTests.java:122)
at org.eclipse.jdt.core.tests.model.CopyMoveElementsTests.testCopyConstructorInDifferentProject(CopyMoveElementsTests.java:166)
at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
at java.base/java.lang.reflect.Method.invoke(Method.java:568)
at junit.framework.TestCase.runTest(TestCase.java:177)
at org.eclipse.jdt.core.tests.junit.extension.TestCase.runTest(TestCase.java:983)
at junit.framework.TestCase.runBare(TestCase.java:142)
at junit.framework.TestResult$1.protect(TestResult.java:122)
at junit.framework.TestResult.runProtected(TestResult.java:142)
at junit.framework.TestResult.run(TestResult.java:125)
at junit.framework.TestCase.run(TestCase.java:130)
[...]

PR #2037, opened 2024-02-16
Fix VariableBinding.getJavaElement() type signature

mickaelistria:
Make signature format returned by the binding match the usual signature format for  IJavaElement.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2036

## How to test

Run provided unit test with/without fix to check fixes makes test pass.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/TypeResolveTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/TypeResolveTests.java
@@ -35,6 +35,12 @@ import org.eclipse.jdt.core.IType;
 import org.eclipse.jdt.core.JavaCore;
 import org.eclipse.jdt.core.JavaModelException;
 import org.eclipse.jdt.core.Signature;
+import org.eclipse.jdt.core.dom.AST;
+import org.eclipse.jdt.core.dom.CompilationUnit;
+import org.eclipse.jdt.core.dom.SingleVariableDeclaration;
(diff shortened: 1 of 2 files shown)


mickaelistria:
The only test failure seems to be a recurring intermittent failure, not related to that patch: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1094 . So it looks like this fixes 1 (new) test and doesn't cause any regression.

mickaelistria:
@rgrunber @jjohnstn Can you please review/merge this fix proposal?

jjohnstn (changes requested):
Patch makes sense.  Since your logic treats arrays differently, you should add an array reference to your test.  Other than that, looks fine.

mickaelistria:
The test is already an array (`s1 = new Object[0]`). Am I missing something?

jjohnstn:
> The test is already an array (`s1 = new Object[0]`). Am I missing something?

Maybe I missed something.  In your test you have var s1 = args[0] so s1 is type Ljava.lang.String which you verify.  Line 271 of your patch does something extra if element is an array (e.g. var s2 = args).

mickaelistria:
I've added a check for arrays as suggested.

jjohnstn:
> I've added a check for arrays as suggested.

Thanks.  Check the test to see if the failure is due to misinterpreting what the result should be or whether it isn't doing what you expected.  When the test succeeds, I'll merge.

mickaelistria:
You were right wondering about those particular lines and asking for a test: adding the extra lines for arrays was a bug. I've removed them.

jjohnstn (approved):

PR #2066, opened 2024-02-26
[22] JEP 463: Unnamed Classes and Instance Main Methods (2nd Preview)

jarthana:
## What it does
Fixes issue #1106 

## Author checklist


jarthana:
Note to self: What should be the answer for ICompilationUnit#getMainTypeName()?

datho7561:
I think it might be nice to add a method to IType to tell if the type is an unnamed class. For instance, in jdt.ls, we use the IJavaElement model to generate Code Minings for references. It doesn't make sense to generate a reference code mining for unnamed classes, so we need some way to tell which classes are unnamed in order to skip them. I think similar cases might come up in jdt.core/jdt.ui

In https://github.com/eclipse-jdtls/eclipse-jdt-core-incubator/pull/102 I added an implementation of an API for this; please let me know what you think.

jarthana:
> I think it might be nice to add a method to IType to tell if the type is an unnamed class. For instance, in jdt.ls, we use the IJavaElement model to generate Code Minings for references. It doesn't make sense to generate a reference code mining for unnamed classes, so we need some way to tell which classes are unnamed in order to skip them. I think similar cases might come up in jdt.core/jdt.ui
> 
> In [eclipse-jdtls#102](https://github.com/eclipse-jdtls/eclipse-jdt-core-incubator/pull/102) I added an implementation of an API for this; please let me know what you think.

Sure, I will take a look at this right after releasing the compiler part. 

BTW, I just realized we need to generate static initializer for implicit types as well. A fix is on the way with more tests.

jarthana:
> Note to self: What should be the answer for ICompilationUnit#getMainTypeName()?

The recommendation from the spec is this:
`In simple implementations of the Java SE Platform, where compilation units are stored in files, the name of this implicitly declared class would typically be the name of the file containing the simple compilation unit ([7.3](https://cr.openjdk.org/~gbierman/jep463/jep463-20231030/specs/implicitly-declared-classes-instance-main-methods-jls.html#jls-7.3)) minus any extension (such as .java or .jav).`

So, ECJ will follow the same guidelines.

stephan-herrmann:
Is this the MR that resulted in https://github.com/eclipse-jdt/eclipse.jdt.core/commit/21f8969a3ad24c3c9aae08d799a7acf1cee769c7 ? The commit comment doesn't tell (neither by title nor using a link). Please _always_ include links to MR and Issue in the commit comment, thanks.

Reason for asking: that commit removed file jdtCompilerAdapter.jar and I wanted to check if any reasons were given or if it was an accident?

jarthana:
> Is this the MR that resulted in [21f8969](https://github.com/eclipse-jdt/eclipse.jdt.core/commit/21f8969a3ad24c3c9aae08d799a7acf1cee769c7) ? The commit comment doesn't tell (neither by title nor using a link). Please _always_ include links to MR and Issue in the commit comment, thanks.

Will do. Sorry about that. 

> Reason for asking: that commit removed file jdtCompilerAdapter.jar and I wanted to check if any reasons were given or if it was an accident?

Has to be an accident. Will restore it.

stephan-herrmann:
> Has to be an accident. Will restore it.

If you haven't yet done so: perhaps the file doesn't actually *need* to be stored in git? I can see that a maven build will create this file in phase prepare-package.

Can we rely on successful builds in the time when the file was "missing" to conclude that removal from git was actually harmless?

On a related note I see that @srikanth-sankaran  / @datho7561  added this very file to `.gitignore` in a commit titled "Support for multiple patterns in a case statement (JEP 456)" (no github link either :frowning_face: ). So there was a conflict between storing the file in git and automatic creation / update during maven builds.

Just trying to connect the dots.

iloveeclipse:
PS: it turned out,  either `org.eclipse.jdt.core.dom.CompilationUnit.types()` implementation **or** javadoc must be fixed, as it now can return also `ImplicitTypeDeclaration` elements - and that may affect clients expecting only `AbstractTypeDeclaration`.

stephan-herrmann:
@jarthana a naive question (I haven't looked at the change in detail): why couldn't `UnnamedClass` extend `AbstractTypeDeclaration`? Is that necessary to hide some features that are available only in regular classes?

datho7561:
> why couldn't `UnnamedClass` extend `AbstractTypeDeclaration`? Is that necessary to hide some features that are available only in regular classes?

Unnamed classes don't have names, whereas AbstractTypeDeclaration does have a name that's part of the API.

stephan-herrmann:
> > why couldn't `UnnamedClass` extend `AbstractTypeDeclaration`? Is that necessary to hide some features that are available only in regular classes?
> 
> Unnamed classes don't have names, whereas AbstractTypeDeclaration does have a name that's part of the API.

OK, but we have at least one precedent for this, `IPackageBinding`:
```
	/**
	 * Returns the name of the package represented by this binding. For named
	 * packages, this is the fully qualified package name (using "." for
	 * separators). For unnamed packages, this is an empty string.
	 *
	 * @return the name of the package represented by this binding, or
	 *    an empty string for an unnamed package
	 */
	@Override
	public String getName();
```

Does DOM have a representation for the unnamed module? That would be a second example.

datho7561:
In `AbstractTypeDeclaration` the name is lazily initialized. The implementation of `getName` creates a blank AST Node for the name if it's null. `setName` forbids you from setting the name to `null`. I think some methods callers have assumed that `getName` will never return null. We'd need to change them in order to accommodate this, and also come up with a way to lazily initialize the name while not creating a blank AST node in the case that it's unnamed.

iloveeclipse:
I've created https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2481.

PR #2107, opened 2024-03-06
completion returns no results with pattern matching

### LINKED ISSUE #2106: completion returns no results with pattern matching
The related issues:
- https://github.com/eclipse-jdtls/eclipse.jdt.ls/issues/3068
- https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/242

From https://github.com/eclipse-jdtls/eclipse.jdt.ls/issues/3068
Consider the following code:
```
import java.util.ArrayList;
import java.util.List;

public class SimpleJavaClass {
    Object unknown = new ArrayList<>();

    if (unknown instanceof List things) {
      things.is| // CA doesn't work
      Object nothing = null;
      things.is| // CA works
    }
  }
}
```
CA returns no results if requested for the first occurrence of things.is. It returns isEmpty(), however, for the second occurrence of thing.is

snjeza:
## What it does
Fixes #2106, https://bugs.eclipse.org/bugs/show_bug.cgi?id=568934#c8 and https://bugs.eclipse.org/bugs/show_bug.cgi?id=568934#c6

## How to test
try CA at |
```
import java.util.ArrayList;
import java.util.List;

public class SimpleJavaClass {
    Object unknown = new ArrayList<>();

    if (unknown instanceof List things) {
      things.is| // try CA here
      Object nothing = null;
      things.is|
    }
  }
}
```
Without the patch:
patternwithout

With the patch:
pattern

I will add a test.

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests17.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests17.java
@@ -524,4 +524,130 @@ public class CompletionTests17 extends AbstractJavaModelCompletionTests {
 				SqlStorage[TYPE_REF]{SqlStorage, , LSqlStorage;, null, null, 52}\
 					""", requestor.getResults());
 	}
+	// https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2106
+	public void testGH2106() throws JavaModelException {
+		this.workingCopies = new ICompilationUnit[1];
+		this.workingCopies[0] = getWorkingCopy(
+				"/Completion/src/X.java",
+				"""
+				import java.util.ArrayList;
+				import java.util.List;
+				public class X {
+				  public static void main(String[] args) {
+				    Object unknown = "abc";
+				    if (unknown instanceof String str) {
+				      str.le
+				      Object nothing = null;
+				      str.le
+				    }
+				  }
+				}
+				""");
+		CompletionTestsRequestor2 requestor = new CompletionTestsRequestor2(true);
+		requestor.allowAllRequiredProposals();
+		String str = this.workingCopies[0].getSource();
+		String completeBehind = "str.le";
+		int cursorLocation = str.indexOf(completeBehind) + completeBehind.length();
+		this.workingCopies[0].codeComplete(cursorLocation, requestor, this.wcOwner);
+		assertResults("length[METHOD_REF]{length(), Ljava.lang.String;, ()I, length, null, 60}",
+				requestor.getResults());
+	}
+	public void testGH2106a() throws JavaModelException {
+		this.workingCopies = new ICompilationUnit[1];
+		this.workingCopies[0] = getWorkingCopy(
+				"/Completion/src/X.java",
+				"""
+				import java.util.ArrayList;
+				import java.util.List;
+				public class X {
+				  public static void main(String[] args) {
+				    Object unknown = "abc";
+				    if (unknown instanceof String str) {
+				      str.
+				      Object nothing = null;
+				      str.
+				    }
+				  }
+				}
+				""");
+		CompletionTestsRequestor2 requestor = new CompletionTestsRequestor2(true);
+		requestor.allowAllRequiredProposals();
+		String str = this.workingCopies[0].getSource();
+		String completeBehind = "str.";
+		int cursorLocation = str.indexOf(completeBehind) + completeBehind.length();
+		this.workingCopies[0].codeComplete(cursorLocation, requestor, this.wcOwner);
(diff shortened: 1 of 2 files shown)


snjeza:
@srikanth-sankaran @iloveeclipse I have updated the PR.

> the fix proposed in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2107 is suspect. You should never have to fiddle with ExtraCompilerModifiers.AccOutOfFlowScope bits.

I left this part out. The PR doesn't touch the bits field.

> See also https://bugs.eclipse.org/bugs/show_bug.cgi?id=568934

It also fixes https://bugs.eclipse.org/bugs/show_bug.cgi?id=568934#c8 and https://bugs.eclipse.org/bugs/show_bug.cgi?id=568934#c6

If you agree, I can add a test.

srikanth-sankaran:
No, this fix is not correct. It is kludgy and may work but I can't predict what other problems it may cause down the road. 

Basically you are pairing a `LocalVariableBinding` where a TypeBinding would be the appropriate one. That is recipe for CCE. Even if it doesn't trigger CCE in testing, do we want to resort to such hacky fixes is the question.

The real problem as I outlined on the bug earlier is due to wrong parse tree recovery. Can you check why the scanner doesn't declare EOF upon seeing the completion at things.is| and why it returns the token Object at all ? 

As you can see the two statements get jammed into one 

things.is| Object 

so it appears as though you are completing on a nested type, while you are actually completing on a QNR.

snjeza:
> and the parse tree recovered in the non-working case is:

I have tried it and I get:

```
import java.util.ArrayList;
import java.util.List;
public class SimpleJavaClass {
  public SimpleJavaClass() {
  }
    Object unknown;
    {
      <CompleteOnType:things.is> Object;
    }
  }
}
```
There isn't `if ((unknown instanceof List things))`

> Completion is being attempted on a Name in both instances. Why do we think it is being attempted on a Type ? I can help with reviewing an alternate fix. Thanks
> The real problem as I outlined on the bug earlier is due to wrong parse tree recovery. Can you check why the scanner doesn't declare EOF upon seeing the completion at things.is| and why it returns the token Object at all ?
>     The parser returns CompletionOnQualifiedNameReference in line 10 (it works) and CompletionOnQualifiedTypeReference in line 8 (it doesn't work)
> Right, that is the bug that needs fixing.

I have checked the following code:

```
import java.util.ArrayList;
import java.util.List;
public class SimpleJavaClass2 {
    Object unknown = new ArrayList<>();
    List list = new ArrayList<>();
    if (unknown instanceof List things) {
      list.is|
      Object nothing = null;
      list.is
    }
    
  }
}
```
the parse tree recovered is
```
import java.util.ArrayList;
import java.util.List;
public class SimpleJavaClass2 {
  public SimpleJavaClass2() {
  }
    Object unknown;
    List list;
    {
      <CompleteOnType:list.is> Object;
    }
  }
}
```
The parser also returns CompletionOnQualifiedTypeReference. It works because the parser finds the `list` variable.

Based on it and the fix for https://bugs.eclipse.org/bugs/show_bug.cgi?id=287939 I have created a new PR

> Basically you are pairing a LocalVariableBinding where a TypeBinding would be the appropriate one. That is recipe for CCE. Even if it doesn't trigger CCE in testing, do we want to resort to such hacky fixes is the question.

This part is removed.

The recovered parse tree with the PR is:
```
import java.util.ArrayList;
import java.util.List;
public class SimpleJavaClass {
  public SimpleJavaClass() {
  }
    Object unknown;
    {
      List things;
      if ((unknown instanceof List things))
          <CompleteOnType:things.is> Object;
    }
  }
}
```

@srikanth-sankaran Could you, please, review.

snjeza:
The completion with the PR

pattern2

snjeza:
I have updated the PR. It works similar to CA in
```
import java.util.ArrayList;
import java.util.List;
public class SimpleJavaClass2 {
    Object unknown = new ArrayList<>();
    List list = new ArrayList<>();
    if (unknown instanceof List things) {
      list.is|
      Object nothing = null;
      list.is
    }
    
  }
}
```
I have also added some tests.

srikanth-sankaran:
I apologize - I don't have the time to do justice to this task - I have been super busy with redesigning and reimplementing ECJ support for Pattern matching, Sealed classes, Switch expressions and Enhanced switch statements (all delivered already) and Records (almost done)

Later this year, I plan to comprehensively look at code assist issues - all the way from lambdas to Java 24 features. If you think this is a show stopper for using Eclipse - (I have doubts but I could be wrong) - I am happy to take myself out of the path to integrating this.

stephan-herrmann:
@rgrunber for the future, could you please ensure that every commit merged to master has a reference to issue and/or pull request? Otherwise looking at a commit in eclipse does not give a clue where background information can be found.

See also https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md#commit-message-recommendations

Issue #2047, opened 2024-02-19
org.eclipse.jdt.core.dom.MethodBinding$LambdaMethod doesn't resolve to a org.eclipse.jdt.internal.core.LambdaMethod (err to ResolvedSourceMethod)


mickaelistria:
```java
// Inspired from ReslveTests18.test024
public void testBindingForLambdaMethod() throws JavaModelException {
	this.wc = getWorkingCopy(
			"/Resolve/src/X.java",
			"interface I {\n" +
			"	int foo(int a);\n" +
			"}\n" +
			"public class X {	\n" +
			"	I i = xyz -> xyz;\n" +
			"}\n");

	String str = this.wc.getSource();
	String selection = "xyz -> xyz";

	ASTParser parser = ASTParser.newParser(AST.getJLSLatest());
	parser.setSource(this.wc);
	parser.setProject(this.wc.getJavaProject());
	parser.setResolveBindings(true);
	CompilationUnit dom = (CompilationUnit)parser.createAST(null);

	org.eclipse.jdt.core.dom.LambdaExpression lambda = (org.eclipse.jdt.core.dom.LambdaExpression)new NodeFinder(dom, str.lastIndexOf(selection), selection.length()).getCoveredNode();
	IJavaElement javaElement = lambda.resolveMethodBinding().getJavaElement();
	assertTrue(javaElement.getClass().getName(), javaElement instanceof LambdaMethod);
}
```
fails with
```
junit.framework.AssertionFailedError: org.eclipse.jdt.internal.core.ResolvedSourceMethod
	at junit.framework.Assert.fail(Assert.java:57)
	at junit.framework.Assert.assertTrue(Assert.java:22)
	at junit.framework.TestCase.assertTrue(TestCase.java:192)
	at org.eclipse.jdt.core.tests.model.ResolveTests18.testBindingForLambdaMethod(ResolveTests18.java:3158)
...
```

Note that the output of `lambda.resolveMethodBinding()` is `org.eclipse.jdt.core.dom.MethodBinding$LambdaMethod` which means that the resolution happens perfectly. The translation to IJavaElement (in `getUnresolvedJavaElement(MethodBinding methodBinding, WorkingCopyOwner workingCopyOwner, BindingsToNodesMap bindingsToNodes)` most probably) seems to be the host the root cause of this issue.

PR #1617, opened 2023-11-23
Modernize code for DefaultBindingResolver

mickaelistria:
## Author checklist

- [x] I have thoroughly tested my changes
- [x] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [x] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

PR #2124, opened 2024-03-08
Resolve MethodBinding$LambdaMethod to LambdaMethod java element

mickaelistria:
Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2047

## What it does

Fixes the resolution of Binding->JavaElement for lambda methods: now returns a LambdaMethod Java Element as expected by eg `codeSelect` in test0024 instead of returning the binding of the overridden method.

## How to test

Run new test with or without fix.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ResolveTests18.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ResolveTests18.java
@@ -27,6 +27,11 @@ import org.eclipse.jdt.core.JavaCore;
 import org.eclipse.jdt.core.JavaModelException;
 import org.eclipse.jdt.core.Signature;
 import org.eclipse.jdt.core.WorkingCopyOwner;
+import org.eclipse.jdt.core.dom.NodeFinder;
+import org.eclipse.jdt.core.dom.CompilationUnit;
+import org.eclipse.jdt.core.dom.Name;
+import org.eclipse.jdt.core.dom.AST;
+import org.eclipse.jdt.core.dom.ASTParser;
 import org.eclipse.jdt.internal.core.LambdaExpression;
 import org.eclipse.jdt.internal.core.LambdaMethod;
 
@@ -3124,6 +3129,49 @@ public void testBug546563() throws Exception {
 		"Unexpected elements",
 		"getOptionalValue() [in Test [in [Working copy] Test.java [in <default> [in src [in Resolve]]]]]",
 		elements
+	);
+}
+
+// Inspired from ReslveTests18.test0027
+public void test0027_BindingForLambdaMethod() throws JavaModelException {
+	this.wc = getWorkingCopy(
+			"/Resolve/src/X.java",
+			"""
+			interface I {
+			    I doit(I xyz);
+			}
+			public class X {
+				public static void main(String[] args) {
+					I i = (pqr) -> {
+						return (xyz) -> {
+							return (abc) -> abc;
+						};
+					};
+				}
+			}
+			""");
+
+	String str = this.wc.getSource();
+	String selection = "abc";
+	int start = str.lastIndexOf(selection);
+	int length = selection.length();
+
+	ASTParser parser = ASTParser.newParser(AST.getJLSLatest());
+	parser.setSource(this.wc);
+	parser.setProject(this.wc.getJavaProject());
+	parser.setWorkingCopyOwner(this.wc.getOwner());
+	parser.setCompilerOptions(this.wc.getOptions(true));
+	parser.setResolveBindings(true);
+	parser.setBindingsRecovery(true);
+	parser.setStatementsRecovery(true);
+	CompilationUnit dom = (CompilationUnit)parser.createAST(null);
+	Name variable = (Name)new NodeFinder(dom, start, length).getCoveredNode();
+	IJavaElement javaElement = variable.resolveBinding().getJavaElement();	
+
+	assertElementsEqual(
+		"Unexpected elements",
+		"abc [in doit(I) [in <lambda #1> [in doit(I) [in <lambda #1> [in doit(I) [in <lambda #1> [in main(String[]) [in X [in [Working copy] X.java [in <default> [in src [in Resolve]]]]]]]]]]]]",
+		new IJavaElement[] { javaElement }
 	);
 }
 }
--- a/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/MethodBinding.java
+++ b/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/MethodBinding.java
@@ -31,6 +31,8 @@ import org.eclipse.jdt.internal.compiler.lookup.TypeBinding;
 import org.eclipse.jdt.internal.compiler.lookup.TypeVariableBinding;
 import org.eclipse.jdt.internal.compiler.problem.AbortCompilation;
 import org.eclipse.jdt.internal.core.JavaElement;
+import org.eclipse.jdt.internal.core.LambdaExpression;
+import org.eclipse.jdt.internal.core.LambdaFactory;
 import org.eclipse.jdt.internal.core.util.Util;
 
 /**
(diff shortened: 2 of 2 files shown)


mickaelistria:
Note that while it improves things, the case of eg test0027 is not resolved by that case.

mickaelistria:
I have looked at the case of test0027, and it looks to me that for this case, the issue is not really the Binding -> JavaElement resolution (which may be fixed with this patch), but more that the DefaultBindingResolver only inserts 1 of the LambdaMethod/LambdaExpression couple in the bindingToASTNodes map instead of the 3 ones existing in the code; line 941 `this.bindingsToAstNodes.put(methodBinding, lambda);` is only called once.
I'll keep investigating, but I would suggest that this patch would be considered for a review/merge as it, and we'll create new issues/PRs for the "other" symptom of the DefaultBindingResolver.

mickaelistria:
I could fix my patch to also address the concern of former comment.

akurtakov:
@srikanth-sankaran Can you help with reviewing this one?

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/MethodBinding.java
```diff
@@ -603,6 +605,32 @@ public IVariableBinding[] getSyntheticOuterLocals() {
...
+			}
+			if (!(this.resolver instanceof DefaultBindingResolver)) return null;
+	
+			DefaultBindingResolver defaultBindingResolver = (DefaultBindingResolver) this.resolver;
+			if (!defaultBindingResolver.fromJavaProject) return null;
+			ASTNode domNode = (ASTNode)defaultBindingResolver.bindingsToAstNodes.get(this);
+			// resolving all parent lambdas is necessary for proper resolution
+			// as it populates newAstToOldAst.
+			ASTNode current = domNode;
+			while (current != null) {
```
jjohnstn, line 622: Does this need to be done every time or can it be done once and then marked as completed or can the LambdaFactory LambdaMethod be cached to avoid this in the future?

jjohnstn (commented):
I am wondering if the test should be in ASTConverter18Tests or perhaps an additional test.  There you can call JavaElement.toStringWithAncestors(false) to verify the hierarchy.

srikanth-sankaran:
This looks very close/similar to https://bugs.eclipse.org/bugs/show_bug.cgi?id=573834 - Need to check why that solution is not covering this case. There also code selection is involved

mickaelistria:
>  Need to check why that solution is not covering this case. There also code selection is involved

The previous patch doesn't involve the DOM nor its bindings. It fixes the SelectionEngine and the SelectionParser , which are not involved when building a DOM.

srikanth-sankaran:
> > Need to check why that solution is not covering this case. There also code selection is involved
> 
> The previous patch doesn't involve the DOM nor its bindings. It fixes the SelectionEngine and the SelectionParser , which are not involved when building a DOM.

Correct, I realized that as started reviewing.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ResolveTests18.java
```diff
@@ -3126,4 +3131,47 @@ public void testBug546563() throws Exception {
...
+	parser.setStatementsRecovery(true);
+	CompilationUnit dom = (CompilationUnit)parser.createAST(null);
+	Name variable = (Name)new NodeFinder(dom, start, length).getCoveredNode();
+	IJavaElement javaElement = variable.resolveBinding().getJavaElement();	
+
+	assertElementsEqual(
+		"Unexpected elements",
+		"abc [in doit(I) [in <lambda #1> [in doit(I) [in <lambda #1> [in doit(I) [in <lambda #1> [in main(String[]) [in X [in [Working copy] X.java [in <default> [in src [in Resolve]]]]]]]]]]]]",
+		new IJavaElement[] { javaElement }
+	);
```
srikanth-sankaran, line 3175: At every level lambda ordinal is \#1 - is that right ??

srikanth-sankaran:
There appears to be a serious pre-existing defect in `org.eclipse.jdt.core.dom.DefaultBindingResolver.getBindingsToNodesMap()` - can we discuss that and understand if that somehow impacts the current behavior ?

What I am wondering is:

org.eclipse.jdt.core.dom.DefaultBindingResolver.getBindingsToNodesMap().new BindingsToNodesMap() {...}.get(Binding) takes a **compiler binding** as argument and does this with it:

```
return (org.eclipse.jdt.internal.compiler.ast.ASTNode)
					DefaultBindingResolver.this.newAstToOldAst.get(DefaultBindingResolver.this.bindingsToAstNodes.get(dBinding));
```
Now `DefaultBindingResolver.this.bindingsToAstNodes` maps from **DOM bindings** to DOM AST nodes. Using compiler bindings as key there is guaranteed to return null in this call stack:

```
Util.getUnresolvedJavaElement(MethodBinding, WorkingCopyOwner, Util$BindingsToNodesMap) line: 1413	
MethodBinding$LambdaMethod(MethodBinding).getUnresolvedJavaElement() line: 318	
MethodBinding$LambdaMethod.getUnresolvedJavaElement() line: 612	
MethodBinding$LambdaMethod(MethodBinding).getJavaElement() line: 304	
VariableBinding.getUnresolvedJavaElement() line: 292	
VariableBinding.getJavaElement() line: 160	
ResolveTests18.test0027_BindingForLambdaMethod() line: 3169	
```

I am wondering if the fix should really belong in the `then` clause of `if (node != null && !declaringType.isBinary()) {` after the above bug is corrected.

WDYT ?

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ResolveTests18.java
mickaelistria: It's a similar assertion that's implemented in test0027, with `1` being the ordinal on every level (unlike anonymous which have an ordinal based on the including type because it's used to generate the .class, lambdas do not generate a .class and thus ordinal seems only relative to the direct parent)

srikanth-sankaran:
OK, I am closing shop for the week, I'll continue to look at this early next week. Just fixing the above bug doesn't change much, but I would still like to discuss what is the impact of such glaringly defective code:

If I change it to:

```
	Util.BindingsToNodesMap getBindingsToNodesMap() {
		return new Util.BindingsToNodesMap() {
			@Override
			public org.eclipse.jdt.internal.compiler.ast.ASTNode get(Binding binding) {
				IMethodBinding dBinding = (IMethodBinding) DefaultBindingResolver.this.bindingTables.compilerBindingsToASTBindings.get(binding);
				return (org.eclipse.jdt.internal.compiler.ast.ASTNode)
					DefaultBindingResolver.this.newAstToOldAst.get(DefaultBindingResolver.this.bindingsToAstNodes.get(dBinding));
			}
		};
	}
```

it starts returning the mapping properly although in the context of this bug, it doesn't change much. Still I wonder if the fix should perhaps belong in the `then` clause. 

Will study more early next week

srikanth-sankaran:
Amazingly that code dates back 15 years. Someone please tell me I misunderstood it! 😆

mickaelistria:
I will try to investigate further later today or early next week. But 1 particular issue I've found is that the "parent" lambda bindings are not included in the maps unless we do explicitly call `resolveBinding` on related ASTNode. This seems to me a bit unusual compared to other types where the bindings seem to exist in the map.
It indeed looks like we might be facing a combination of bugs for which fixes are not so related.

Note that the bug about the requested type in the map was made visible with https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1617#discussion_r1410782257 some time ago.

srikanth-sankaran:
@stephan-herrmann Take a bow! 😊

srikanth-sankaran (approved):
Other than the pre-existing issue point out already, this looks ok. I haven't worked in these parts in almost 10 years - I must state that too :)

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/MethodBinding.java
mickaelistria: Usage of this method is relatively sporadic (happens only when user Ctrl+clicks it at the moment) and execution remains relatively fast. On the other hand, the DOM may be retained for a long time, so adding some cache here increase memory consumption.
Overall, I think it's best to not cache and keep more memory free for that case. We can easily revisit later if my impressions are proven wrong though.

akurtakov:
As the patch is approved and covered with test - merging.

PR #2139, opened 2024-03-12
[22] Implement Elements#getEnumConstantBody() #2135

### LINKED ISSUE #2135: [22] Implement Elements#getEnumConstantBody()
This is a new API added in Java 22 and the default implementation throws `UnsupportedOperationException`.

jarthana:
## What it does
Fixes #2135

## Author checklist


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/apt/model/ElementsImpl9.java
```diff
@@ -284,6 +288,37 @@ private JavaFileObject getSourceJavaFileObject(ReferenceContext referenceContext
...
+						return null;
+					if (sourceField.initialization instanceof QualifiedAllocationExpression qualAlloc) {
+						TypeDeclaration decl = qualAlloc.anonymousType;
+						if (decl == null)
+							return null;
+
+						SourceTypeBinding declaringClass = (SourceTypeBinding) field.declaringClass;
+						TypeDeclaration enumTypeDecl = declaringClass.scope.referenceContext;
+						// The following code is copied from BlockScope#addAnonymousType() to force
+						// resolution of the local type binding ahead of a full CU#resolve()
```
stephan-herrmann, line 307: At this point, could we already have `enumTypeDecl.binding` set due to a previous call to the current method?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/apt/model/ElementsImpl9.java
```diff
@@ -284,6 +288,37 @@ private JavaFileObject getSourceJavaFileObject(ReferenceContext referenceContext
...
+						BlockScope scope = field.isStatic() ? enumTypeDecl.staticInitializerScope : enumTypeDecl.initializerScope;
+						if (scope != null) {
+							ClassScope anonymousClassScope = new ClassScope(scope, decl);
+							anonymousClassScope.buildAnonymousTypeBinding(scope.enclosingSourceType(), declaringClass);
```
stephan-herrmann, line 311: `scope.enclosingSourceType()` looks a bit like going in a circle, like: isn't this the same as `declaringClass`?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/BlockScope.java
```diff
@@ -99,11 +99,14 @@ protected BlockScope(int kind, Scope parent) {
...
  */
 public final void addAnonymousType(TypeDeclaration anonymousType, ReferenceBinding superBinding) {
-	ClassScope anonymousClassScope = new ClassScope(this, anonymousType);
-	anonymousClassScope.buildAnonymousTypeBinding(
-		enclosingSourceType(),
-		superBinding);
-
+	// This may have been called from an annotation processor through Elements#getEnumConstantBody()
+	// and binding may have been set. If binding is already set, skip this
+	if (anonymousType.binding == null) {
```
stephan-herrmann, line 104: So the assumption is, that non-null will only occur if `superBinding` is an enum, right?
Should that be encoded as an assertion?

stephan-herrmann (commented):
Just from looking at the change (without testing it in debugger or otherwise) this looks morally OK.

If the conditional on-demand initialization inside apt gets troublesome I could even imagine letting the compiler eagerly create these bindings just for the case of enum constants. But then I'm not sure if that would improve anything.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/apt/model/ElementsImpl9.java
jarthana: It is. And admit, it was lazy coding although I did observe they are same :)

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/BlockScope.java
jarthana: Are you saying if it is non-null and super is an enum, we just don't need the code below and therefore return?

Issue #1125, opened 2023-06-06
jdt.core.dom uses wrong double-checked locking


jukzi:
SonarLint complains the used pattern used is wrong.
see https://wiki.sei.cmu.edu/confluence/display/java/LCK10-J.+Use+a+correct+form+of+the+double-checked+locking+idiom

Boils down to a "volatile" keyword on the members would solve it (but could make read access a bit slower).

![image](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/51790620/69dded10-1a78-4ae0-b06b-30faba066885)
...

jukzi:
the pattern was introduces with https://bugs.eclipse.org/bugs/show_bug.cgi?id=58314 `Make ASTs thread-safe for multiple readers`

jukzi:
according to https://wiki.sei.cmu.edu/confluence/display/java/LCK10-J.+Use+a+correct+form+of+the+double-checked+locking+idiom

- the member needs to be volatile OR 
- the member needs to Immutable AND the new instance needs to be returned

jukzi:
see also
https://rules.sonarsource.com/java/tag/multi-threading/RSPEC-2168/
https://www.cs.umd.edu/~pugh/java/memoryModel/DoubleCheckedLocking.html

PR #2152, opened 2024-03-12
fix jdt.core.dom wrong double-checked locking #1125

jukzi:
Make sure the newly created value is returned on all members that are lazy initialized

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1125

### DIFF
--- a/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/AbstractTypeDeclaration.java
+++ b/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/AbstractTypeDeclaration.java
@@ -35,7 +35,7 @@ public abstract class AbstractTypeDeclaration extends BodyDeclaration {
 	 * legal Java class identifier.
 	 * @since 2.0 (originally declared on {@link TypeDeclaration})
 	 */
-	SimpleName typeName = null;
+	volatile SimpleName typeName;
 
 	/**
 	 * The body declarations (element type: {@link BodyDeclaration}).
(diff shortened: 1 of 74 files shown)


org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/AbstractTypeDeclaration.java
```diff
@@ -35,7 +35,7 @@ public abstract class AbstractTypeDeclaration extends BodyDeclaration {
 	 * legal Java class identifier.
 	 * @since 2.0 (originally declared on {@link TypeDeclaration})
 	 */
-	SimpleName typeName = null;
+	volatile SimpleName typeName = null;
```
iloveeclipse, line 38: May I ask you to remove `= null` assignments for touched fields, because 1) it us useless code and 2) it would make debugging experience better (no jumping on object creation code from field to field)

jukzi: ok, a regexp fixed that.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/SwitchCase.java
```diff
@@ -122,7 +122,7 @@ public static List propertyDescriptors(int apiLevel) {
 	/**
 	 * Indicates whether <code>optionalExpression</code> has been initialized.
 	 */
-	private boolean expressionInitialized = false;
+	private volatile boolean expressionInitialized = false;
```
iloveeclipse, line 125: `= false` could be also removed

jukzi: isn't there a cleanup for remove unnecessary assignment of default value? If not maybe @jjohnstn could implement one?

Issue #2160, opened 2024-03-14
VerifyError with code with record patterns and Optional


pietrobraione:
The following Eclipse 2024-03 project [bug.zip](https://github.com/eclipse-jdt/eclipse.jdt.core/files/14605779/bug.zip), when compiled with the JDT compiler with Java 21 compatibility, produces a bug.Interpreter class that raises a VerifyError upon loading. The Interpreter class contains some nested "if" statements with record patterns and destructures some Optional values. A JUnit test is included that, when run, triggers the runtime error.

Code of the Interpreter class:

    public final class Interpreter {
        public void step(Configuration J) {
            var e = J.syExpression();
            if (e instanceof SyExpressionGetfield syExpressionGetfield) {
                if (syExpressionGetfield instanceof SyExpressionGetfield(SyExpressionValue(SyValueReferenceConstant(SyReferenceConstantSymbol Y)), String f)) {
                    final Optional<Objekt> maybeo = J.objectAt(Y);
                    maybeo.ifPresentOrElse(o -> {
                        //final Optional<SyValue> maybev = o.get(f); //works
                        final Optional<SyValue> maybev = o.get(syExpressionGetfield.fieldName()); //does not work
                        maybev.ifPresentOrElse(v -> {
                            freshSymbol(Y.sySymbol(), f);
                        }, () -> {});
                    }, () -> {});
                } 
            }
        }

        private SySymbolField freshSymbol(SySymbol sySymbol, String fieldName) {
            return null;
        }
    }

Exception stacktrace:

    java.lang.VerifyError: Operand stack overflow
    Exception Details:
      Location:
        bug/Interpreter.step(Lbug/Configuration;)V @101: aload
      Reason:
        Exceeded max stack size.
      Current Frame:
        bci: @101
        flags: { }
        locals: { 'bug/Interpreter', 'bug/Configuration', 'bug/syntax/SyExpression', 'bug/syntax/SyExpressionGetfield', 'bug/syntax/SyReferenceConstantSymbol', 'java/lang/String', 'java/util/Optional' }
        stack: { 'java/util/Optional', 'bug/Interpreter', 'bug/syntax/SyExpressionGetfield', 'bug/syntax/SyReferenceConstantSymbol' }
      Bytecode:
        0000000: 2bb6 0010 4d2c c100 1699 006b 2cc0 0016
        0000010: 4e2d c100 1699 005f 2dc0 0016 59b6 0018
[...]

PR #2100, opened 2024-03-06
Collect artifact comparison delta

laeubi:
This enables Tycho do produce comporator logs like in aggregators builds and make them available as downloads from the jenkins

This will only show up *after* a merge, if one likes to test this a committer needs to go to the job, copy the contents of my change Jenkins file and choose "Replay" then paste the change Jenkinsfile there.

FYI @iloveeclipse @srikanth-sankaran

PR #2171, opened 2024-03-15
ECJ fails to track operand stack size across branch edges

srikanth-sankaran:
## What it does

* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2160
* Revert fix for https://bugs.eclipse.org/bugs/show_bug.cgi?id=544298 and refix
* Revert fix for https://bugs.eclipse.org/bugs/show_bug.cgi?id=544556 and refix
* Revert fix for https://bugs.eclipse.org/bugs/show_bug.cgi?id=544204 and refix
* Fix arbitrary and incorrect stack manipulation code in string concatenation invoke dynamic call
* Fix incorrect stack manipulation in string templates invoke dynamic call
* Fix argument and receiver size calculation error for captured outer locals
* Replace various ad-hoc stack adjustment calls with structured stack size tracking

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/codegen/BranchLabel.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/codegen/BranchLabel.java
@@ -26,6 +26,7 @@ public class BranchLabel extends Label {
 
 	// Label tagbits
 	public int tagBits;
+	protected int targetStackDepth = -1;
 	public final static int WIDE = 1;
 	public final static int USED = 2;
 
@@ -116,7 +117,39 @@ public void becomeDelegateFor(BranchLabel otherLabel) {
 	this.forwardReferences = mergedForwardReferences;
 	this.forwardReferenceCount = indexInMerge;
 }
-
+private boolean shouldThrow = true;
+protected void trackStackDepth(boolean branch) {
+	/* Control can reach an instruction with a label in two ways: (1) via a branch using that label or (2) by falling through from the previous instruction.
+	   In both cases, we KNOW the stack depth at the instruction from which control flows to the instruction with the label.
+	   Control cannot reach the instruction with a label by falling through if the previous instruction completes abruptly via a goto/return/throw etc
+	   this.codeStream.lastAbruptCompletion == this.position ==> instruction prior to the one with the label is a goto/return/throw
+	*/
+	boolean sourceDepthKnown = branch || this.codeStream.lastAbruptCompletion != this.position;
+	if (this.targetStackDepth == -1) {
+		// First branch to this label || placement of this label
+		if (sourceDepthKnown) {
+			if (this.codeStream.stackDepth < 0) {
+				this.codeStream.classFile.referenceBinding.scope.problemReporter()
+						.operandStackSizeNegative(this.codeStream.classFile.referenceBinding.scope.referenceContext);
+				this.codeStream.stackDepth = 0; // fwiw
+				if (this.shouldThrow)
+					throw new AssertionError("Operand stack size negative!"); //$NON-NLS-1$
+			}
+			this.targetStackDepth = this.codeStream.stackDepth;
+		} // else: previous instruction completes abruptly via goto/return/throw: We can't use the stack depth there; Wait for a backward branch to be emitted.
+	} else {
+		// Stack depth known at label having encountered a previous branch and/or having fallen through to label
+		if (sourceDepthKnown) {
+			if (this.targetStackDepth != this.codeStream.stackDepth) {
+				this.codeStream.classFile.referenceBinding.scope.problemReporter().operandStackSizeInconsistent(
+						this.codeStream.classFile.referenceBinding.scope.referenceContext);
+				if (this.shouldThrow)
+					throw new AssertionError("Operand stack size inconsistent!"); //$NON-NLS-1$
+			}
+		}
+		this.codeStream.stackDepth = this.targetStackDepth;
+	}
+}
 /*
 * Put down  a reference to the array at the location in the codestream.
 */
@@ -137,6 +170,7 @@ void branch() {
 		 */
 		this.codeStream.writePosition(this);
 	}
+	trackStackDepth(true);
 }
 
 /*
@@ -157,6 +191,7 @@ void branchWide() {
 	} else { //Position is set. Write it!
 		this.codeStream.writeWidePosition(this);
 	}
+	trackStackDepth(true);
 }
 
 public int forwardReferenceCount() {
@@ -239,6 +274,7 @@ public void place() { // Currently lacking wide support.
 			this.codeStream.optimizeBranch(oldPosition, this);
 		}
 	}
+	trackStackDepth(false);
 }
 
 /**
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/codegen/BranchLabel.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/codegen/BranchLabel.java
@@ -117,7 +117,7 @@ public void becomeDelegateFor(BranchLabel otherLabel) {
 	this.forwardReferences = mergedForwardReferences;
 	this.forwardReferenceCount = indexInMerge;
 }
-private boolean shouldThrow = true;
+
 protected void trackStackDepth(boolean branch) {
 	/* Control can reach an instruction with a label in two ways: (1) via a branch using that label or (2) by falling through from the previous instruction.
 	   In both cases, we KNOW the stack depth at the instruction from which control flows to the instruction with the label.
@@ -130,23 +130,19 @@ protected void trackStackDepth(boolean branch) {
 		if (sourceDepthKnown) {
 			if (this.codeStream.stackDepth < 0) {
 				this.codeStream.classFile.referenceBinding.scope.problemReporter()
-						.operandStackSizeNegative(this.codeStream.classFile.referenceBinding.scope.referenceContext);
-				this.codeStream.stackDepth = 0; // fwiw
-				if (this.shouldThrow)
-					throw new AssertionError("Operand stack size negative!"); //$NON-NLS-1$
+						.operandStackSizeInappropriate(this.codeStream.classFile.referenceBinding.scope.referenceContext);
+				this.codeStream.stackDepth = 0; // FWIW
 			}
 			this.targetStackDepth = this.codeStream.stackDepth;
-		} // else: previous instruction completes abruptly via goto/return/throw: We can't use the stack depth there; Wait for a backward branch to be emitted.
+		} // else: previous instruction completes abruptly via goto/return/throw: Wait for a backward branch to be emitted.
 	} else {
 		// Stack depth known at label having encountered a previous branch and/or having fallen through to label
 		if (sourceDepthKnown) {
 			if (this.targetStackDepth != this.codeStream.stackDepth) {
-				this.codeStream.classFile.referenceBinding.scope.problemReporter().operandStackSizeInconsistent(
+				this.codeStream.classFile.referenceBinding.scope.problemReporter().operandStackSizeInappropriate(
 						this.codeStream.classFile.referenceBinding.scope.referenceContext);
 				if (this.targetStackDepth < this.codeStream.stackDepth)
-					this.targetStackDepth = this.codeStream.stackDepth; // fwiw, pick the higher water mark.
-				if (this.shouldThrow)
-					throw new AssertionError("Operand stack size inconsistent!"); //$NON-NLS-1$
+					this.targetStackDepth = this.codeStream.stackDepth; // FWIW, pick the higher water mark.
 			}
 		}
 		this.codeStream.stackDepth = this.targetStackDepth;
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/codegen/BranchLabel.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/codegen/BranchLabel.java
@@ -143,6 +143,8 @@ protected void trackStackDepth(boolean branch) {
 			if (this.targetStackDepth != this.codeStream.stackDepth) {
 				this.codeStream.classFile.referenceBinding.scope.problemReporter().operandStackSizeInconsistent(
 						this.codeStream.classFile.referenceBinding.scope.referenceContext);
+				if (this.targetStackDepth < this.codeStream.stackDepth)
+					this.targetStackDepth = this.codeStream.stackDepth; // fwiw, pick the higher water mark.
 				if (this.shouldThrow)
 					throw new AssertionError("Operand stack size inconsistent!"); //$NON-NLS-1$
 			}
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/core/compiler/IProblem.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/core/compiler/IProblem.java
@@ -505,9 +505,7 @@ void setSourceStart(int sourceStart);
 	/** @since 3.38 */
 	int OperandStackExceeds64KLimit = Internal + 87;
 	/** @since 3.38 */
-	int OperandStackSizeNegative = Internal + 88;
-	/** @since 3.38 */
-	int OperandStackSizeInconsistent = Internal + 89;
+	int OperandStackSizeInappropriate = Internal + 88;
 
 	// variable hiding
 	/** @since 3.0 */
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/core/compiler/IProblem.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/core/compiler/IProblem.java
@@ -504,6 +504,10 @@ void setSourceStart(int sourceStart);
 	int BytecodeExceeds64KLimitForSwitchTable = Internal + 86;
 	/** @since 3.38 */
 	int OperandStackExceeds64KLimit = Internal + 87;
+	/** @since 3.38 */
+	int OperandStackSizeNegative = Internal + 88;
+	/** @since 3.38 */
+	int OperandStackSizeInconsistent = Internal + 89;
 
 	// variable hiding
 	/** @since 3.0 */
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/AND_AND_Expression.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/AND_AND_Expression.java
@@ -171,7 +171,6 @@ public class AND_AND_Expression extends BinaryExpression {
 						codeStream.iconst_0();
 					} else {
 						codeStream.goto_(endLabel = new BranchLabel(codeStream));
-						codeStream.decrStackSize(1);
 						falseLabel.place();
 						codeStream.iconst_0();
 						endLabel.place();
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/AbstractMethodDeclaration.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/AbstractMethodDeclaration.java
@@ -388,7 +388,7 @@ public abstract class AbstractMethodDeclaration
 				for (Statement stmt : this.statements) {
 					stmt.generateCode(this.scope, codeStream);
 					if (!this.compilationResult.hasErrors() && codeStream.stackDepth != 0) {
-						throw new AssertionError("Unexpected stack size change after statement"); //$NON-NLS-1$
+						this.scope.problemReporter().operandStackSizeInappropriate(this);
 					}
 				}
 			}
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/AbstractMethodDeclaration.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/AbstractMethodDeclaration.java
@@ -387,6 +387,9 @@ public abstract class AbstractMethodDeclaration
 			if (this.statements != null) {
 				for (Statement stmt : this.statements) {
 					stmt.generateCode(this.scope, codeStream);
+					if (!this.compilationResult.hasErrors() && codeStream.stackDepth != 0) {
+						throw new AssertionError("Unexpected stack size change after statement"); //$NON-NLS-1$
+					}
 				}
 			}
 			// if a problem got reported during code gen, then trigger problem method creation
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/BinaryExpression.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/BinaryExpression.java
@@ -513,7 +513,6 @@ public void generateCode(BlockScope currentScope, CodeStream codeStream, boolean
 					codeStream.iconst_0();
 				} else {
 					codeStream.goto_(endLabel = new BranchLabel(codeStream));
-					codeStream.decrStackSize(1);
 					falseLabel.place();
 					codeStream.iconst_0();
 					endLabel.place();
@@ -536,7 +535,6 @@ public void generateCode(BlockScope currentScope, CodeStream codeStream, boolean
 					codeStream.iconst_0();
 				} else {
 					codeStream.goto_(endLabel = new BranchLabel(codeStream));
-					codeStream.decrStackSize(1);
 					falseLabel.place();
 					codeStream.iconst_0();
 					endLabel.place();
@@ -559,7 +557,6 @@ public void generateCode(BlockScope currentScope, CodeStream codeStream, boolean
 					codeStream.iconst_0();
 				} else {
 					codeStream.goto_(endLabel = new BranchLabel(codeStream));
-					codeStream.decrStackSize(1);
 					falseLabel.place();
 					codeStream.iconst_0();
 					endLabel.place();
@@ -582,7 +579,6 @@ public void generateCode(BlockScope currentScope, CodeStream codeStream, boolean
 					codeStream.iconst_0();
 				} else {
 					codeStream.goto_(endLabel = new BranchLabel(codeStream));
-					codeStream.decrStackSize(1);
 					falseLabel.place();
 					codeStream.iconst_0();
 					endLabel.place();
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Clinit.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Clinit.java
@@ -209,7 +209,6 @@ public class Clinit extends AbstractMethodDeclaration {
 			codeStream.ifne(falseLabel);
 			codeStream.iconst_1();
 			BranchLabel jumpLabel = new BranchLabel(codeStream);
-			codeStream.decrStackSize(1);
 			codeStream.goto_(jumpLabel);
 			falseLabel.place();
 			codeStream.iconst_0();
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ConditionalExpression.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ConditionalExpression.java
@@ -300,18 +300,6 @@ public FlowInfo analyseCode(BlockScope currentScope, FlowContext flowContext,
 				int position = codeStream.position;
 				codeStream.goto_(endifLabel);
 				codeStream.recordPositionsFrom(position, this.valueIfTrue.sourceEnd);
-				// Tune codestream stack size
-				if (valueRequired) {
-					switch(this.resolvedType.id) {
-						case TypeIds.T_long :
-						case TypeIds.T_double :
-							codeStream.decrStackSize(2);
-							break;
-						default :
-							codeStream.decrStackSize(1);
-							break;
-					}
-				}
 			}
 		}
 		if (needFalsePart) {
@@ -325,9 +313,6 @@ public FlowInfo analyseCode(BlockScope currentScope, FlowContext flowContext,
 				falseLabel.place();
 			}
 			this.valueIfFalse.generateCode(currentScope, codeStream, valueRequired);
-			if (valueRequired) {
-				codeStream.recordExpressionType(this.resolvedType);
-			}
 			if (needTruePart) {
 				// End of if statement
 				endifLabel.place();
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ConstructorDeclaration.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ConstructorDeclaration.java
@@ -451,6 +451,9 @@ private void internalGenerateCode(ClassScope classScope, ClassFile classFile) {
 		if (this.statements != null) {
 			for (Statement statement : this.statements) {
 				statement.generateCode(this.scope, codeStream);
+				if (!this.compilationResult.hasErrors() && codeStream.stackDepth != 0) {
+					throw new AssertionError("Unexpected stack size change after statement"); //$NON-NLS-1$
+				}
 			}
 		}
 		// if a problem got reported during code gen, then trigger problem method creation
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ConstructorDeclaration.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ConstructorDeclaration.java
@@ -452,7 +452,7 @@ private void internalGenerateCode(ClassScope classScope, ClassFile classFile) {
 			for (Statement statement : this.statements) {
 				statement.generateCode(this.scope, codeStream);
 				if (!this.compilationResult.hasErrors() && codeStream.stackDepth != 0) {
-					throw new AssertionError("Unexpected stack size change after statement"); //$NON-NLS-1$
+					this.scope.problemReporter().operandStackSizeInappropriate(this);
 				}
 			}
 		}
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/EqualExpression.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/EqualExpression.java
@@ -366,7 +366,6 @@ public class EqualExpression extends BinaryExpression {
 					} else {
 						BranchLabel endLabel = new BranchLabel(codeStream);
 						codeStream.goto_(endLabel);
-						codeStream.decrStackSize(1);
 						// comparison is FALSE
 						falseLabel.place();
 						codeStream.iconst_1();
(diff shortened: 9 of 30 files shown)


srikanth-sankaran:
@laeubi - is there a way I can trigger an I build without merging to get a **preview** of (a) any build issues due to new asserts added to the compiler (b) tests status (c) comparator errors ? ?

I have JDT specific data from the PR here : https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-2171/12/consoleText and that indicates massive changes - nearly 100+ class files change. Basically there is a bug fix to indy string concatenation - it was computing the stack size incorrectly and it is now rectified. Being a common operation string concatenation code generation change will result in large number of differences. 

I haven't yet sanity checked JDT comparator differences - I will before the merge,

laeubi:
This one shows how to collect the comparator data but it was not merged and now has conflicts, maybe you still want to enable these options in either a local build or for your PR:
- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2100

srikanth-sankaran:
> This one shows how to collect the comparator data but it was not merged and now has conflicts, maybe you still want to enable these options in either a local build or for your PR:
> 
> * [Collect artifact comparison delta #2100](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2100)

Hmm. I think for now, I will validate JDT class files changed + get code review before merge. Thanks.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/codegen/BranchLabel.java
```diff
@@ -116,7 +117,39 @@ public void becomeDelegateFor(BranchLabel otherLabel) {
...
+		// Stack depth known at label having encountered a previous branch and/or having fallen through to label
+		if (sourceDepthKnown) {
+			if (this.targetStackDepth != this.codeStream.stackDepth) {
+				this.codeStream.classFile.referenceBinding.scope.problemReporter().operandStackSizeInconsistent(
+						this.codeStream.classFile.referenceBinding.scope.referenceContext);
+				if (this.shouldThrow)
+					throw new AssertionError("Operand stack size inconsistent!"); //$NON-NLS-1$
+			}
+		}
+		this.codeStream.stackDepth = this.targetStackDepth;
```
mpalat, line 150: As a safety net, shouldn't this be max of the two values [assume sourceDepth is false and this.targetStackDepth != this.codeStream.stackDepth
[interim review comment]

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/codegen/BranchLabel.java
```diff
@@ -116,7 +117,39 @@ public void becomeDelegateFor(BranchLabel otherLabel) {
...
+	/* Control can reach an instruction with a label in two ways: (1) via a branch using that label or (2) by falling through from the previous instruction.
+	   In both cases, we KNOW the stack depth at the instruction from which control flows to the instruction with the label.
+	   Control cannot reach the instruction with a label by falling through if the previous instruction completes abruptly via a goto/return/throw etc
+	   this.codeStream.lastAbruptCompletion == this.position ==> instruction prior to the one with the label is a goto/return/throw
+	*/
+	boolean sourceDepthKnown = branch || this.codeStream.lastAbruptCompletion != this.position;
+	if (this.targetStackDepth == -1) {
+		// First branch to this label || placement of this label
+		if (sourceDepthKnown) {
+			if (this.codeStream.stackDepth < 0) {
```
mpalat, line 131: Now we are at it, just checking whether it makes sense to put a check at the decrement parts - yes, there are 193 references for accesses, and this incr/decr wrapping may slow-down - however, we would find the -ve at the time of being assigned. A check here is definitely better than current stage but it still finds the -ve only after it has happened and not at the time of it going -ve. can discuss

mpalat (commented):
please see the interim review comments

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/codegen/BranchLabel.java
srikanth-sankaran: Agreed, I plan to remove the assertion - have only warning and can pick the higher of any two differing values as a safety option,

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/codegen/BranchLabel.java
srikanth-sankaran: This is supposed to be extremely rare scenario. When required using conditional watchpoints can be an option rather than change all over the place.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/codegen/BranchLabel.java
```diff
@@ -116,7 +117,39 @@ public void becomeDelegateFor(BranchLabel otherLabel) {
...
 }
-
+private boolean shouldThrow = true;
+protected void trackStackDepth(boolean branch) {
+	/* Control can reach an instruction with a label in two ways: (1) via a branch using that label or (2) by falling through from the previous instruction.
+	   In both cases, we KNOW the stack depth at the instruction from which control flows to the instruction with the label.
+	   Control cannot reach the instruction with a label by falling through if the previous instruction completes abruptly via a goto/return/throw etc
+	   this.codeStream.lastAbruptCompletion == this.position ==> instruction prior to the one with the label is a goto/return/throw
+	*/
+	boolean sourceDepthKnown = branch || this.codeStream.lastAbruptCompletion != this.position;
```
mpalat, line 127: I think we should be resetting the fall-throught stack depth to zero
ie the code can modified as the one shown below:

`	if (this.codeStream.lastAbruptCompletion == this.position) { // fall-through
		this.codeStream.stackDepth = this.targetStackDepth > 0 ? this.targetStackDepth : 0;
		return;
	}
	boolean sourceDepthKnown = branch; // can just use branch now instead of sourceDepthKnown
`

mpalat (approved):
Rest of the code looks good

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/codegen/BranchLabel.java
srikanth-sankaran: Let us discuss this case:

```
public class X {
	static void foo(int a, int b, int c) {
		System.out.println(c);
	}
		args = new String [] { "Hello", "World" };
		
		foo(10, 20, switch(args.length) {
			default -> {
				int i = args.length;
				int j = 0;
				while (i > 0) {
					i--;
					j++;
				}
				yield j;
			}
		});
	}
}
```

which results in this bytecode:

```
  public static void main(java.lang.String[]);
    descriptor: ([Ljava/lang/String;)V
    flags: (0x0009) ACC_PUBLIC, ACC_STATIC
    Code:
      stack=4, locals=3, args_size=1
         0: iconst_2
         1: anewarray     #34                 // class java/lang/String
         4: dup
         5: iconst_0
         6: ldc           #36                 // String Hello
         8: aastore
         9: dup
        10: iconst_1
        11: ldc           #38                 // String World
        13: aastore
        14: astore_0
        15: bipush        10
        17: bipush        20
        19: aload_0
        20: arraylength
        21: pop
        22: aload_0
        23: arraylength
        24: istore_1
        25: iconst_0
        26: istore_2
        27: goto          36
        30: iinc          1, -1
        33: iinc          2, 1
        36: iload_1
        37: ifgt          30
        40: iload_2
        41: invokestatic  #40                 // Method foo:(III)V
        44: return
```

At bci 30, we come in with a `stackDepth` of 2 - this is the correct value - we should not reset it to 0.

srikanth-sankaran:
Question still remains what we should do when we see an instruction with a label which has never so far been branched into (i.e., no forward branches) and is preceded by a `goto` or `return` or `throw` - ATM we don't do anything and wait for a backward branch. It _**may not be unreasonable**_ to update `BranchLabel.targetStackDepth` with `CodeStream.stackDepth`

srikanth-sankaran:
The last commit introduces further asserts during code generation to ensure operand stack is empty after a full statement's translation. These asserts also help catch a problem in indy string concat code generation. There was a wrong decrement of stackDepth by 189.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/codegen/BranchLabel.java
srikanth-sankaran: > I think we should be resetting the stack depth of the instruction following the abrupt completion/goto to zero ie the code can modified as the one shown below:
> 
> ```
>         if (this.codeStream.lastAbruptCompletion == this.position) { 
> 
> 		this.codeStream.stackDepth = this.targetStackDepth > 0 ? this.targetStackDepth : 0;
> 
> 		return;
> 
> 	}
> 	boolean sourceDepthKnown = branch; // can just use branch now instead of sourceDepthKnown
> ```

What about the case `!branch && this.codeStream.lastAbruptCompletion != this.position` ??

laeubi:
@srikanth-sankaran this is now merged:
- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2100

so if you rebase your PR you should get artifact comparison zips for every project that has a baseline difference.

akurtakov:
I took the freedom to rebase it so we can see it in action.

srikanth-sankaran:
> @srikanth-sankaran this is now merged:
> 
> * [Collect artifact comparison delta #2100](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2100)
> 
> so if you rebase your PR you should get artifact comparison zips for every project that has a baseline difference.

When you say _every_ project - you mean every project in jdt.core ? In my conversation with @iloveeclipse yesterday I understood it to be the case that when a JDT PR is raised, the proposed code is used to build jdt core projects but rest of SDK will get built using the PR's changes only post merge to master

srikanth-sankaran:
> I took the freedom to rebase it so we can see it in action.

No problem. Is https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-2171/14/artifact/ the place one would look for files that differ ? In particular does the single download link labelled `"(all files in zip)"` suffice ?

laeubi:
You can download all or for example individual items like
[org.eclipse.jdt.compiler.tool.tests](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-2171/14/artifact/org.eclipse.jdt.compiler.tool.tests)/[target](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-2171/14/artifact/org.eclipse.jdt.compiler.tool.tests/target)/[artifactcomparison](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-2171/14/artifact/org.eclipse.jdt.compiler.tool.tests/target/artifactcomparison)/[no-classifier](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-2171/14/artifact/org.eclipse.jdt.compiler.tool.tests/target/artifactcomparison/no-classifier)

srikanth-sankaran:
> You can download all or for example individual items like https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-2171/14/artifact/org.eclipse.jdt.compiler.tool.tests/target/artifactcomparison/no-classifier/
> 
> or even individual class files:
> 
> https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-2171/14/artifact/org.eclipse.jdt.compiler.tool.tests/target/artifactcomparison/no-classifier/org/eclipse/jdt/compiler/tool/tests/

Thanks, very useful. Appreciate the effort to get this done! Thanks @laeubi

srikanth-sankaran:
I have inspected several dozens of different class files from jdt projects and they all seem reasonable. 

The only difference pattern is `MAXSTACK` estimate going down wrt to baseline; The new value can be lower by as many `makeConcatWithConstants` `invokedynamic` instructions needed in a statement in the method.

laeubi:
@srikanth-sankaran from now on (next build) these should also show up as warnings in the compile logs:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1914

if you are thinking it is fine, you should either bump the version (if not already bumped) or touch the forceQualifierUpdate.txt file as part of your PR.

trancexpress:
Opened issue for the comparator errors: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1923

@srikanth-sankaran , @sravanlakkimsetti , how do we continue here? Who checks if the changes are expected?

Ideally we would have had 2 builds between the changes from #2181 and #2171, unfortunately we have only 1 build for both changes. So we don't know which PR results in which comparator errors.

laeubi:
> Ideally we would have had 2 builds between the changes from #2181 and #2171, unfortunately we have only 1 build for both changes. So we don't know which PR results in which comparator errors.

This is only partially true. [If you look at the log](https://download.eclipse.org/eclipse/downloads/drops4/I20240320-0740/buildlogs/comparatorlogs/buildtimeComparatorUnanticipated.log.txt) then search for `eclipse.jdt.core` you will see that these comparator errors would have been detectable and analyzed as part of the PR itself.

Its a bit unfortunate that this is not done, I would have expsected that a PR is never merged with comparator errors in JDT as otherwise build failures are guaranteed!

srikanth-sankaran:
> > Ideally we would have had 2 builds between the changes from #2181 and #2171, unfortunately we have only 1 build for both changes. So we don't know which PR results in which comparator errors.
> 
> 
> 
> This is only partially true. [If you look at the log](https://download.eclipse.org/eclipse/downloads/drops4/I20240320-0740/buildlogs/comparatorlogs/buildtimeComparatorUnanticipated.log.txt) then search for `eclipse.jdt.core` you will see that these comparator errors would have been detectable and analyzed as part of the PR itself.
> 
> 
> 
> Its a bit unfortunate that this is not done, I would have expsected that a PR is never merged with comparator errors in JDT as otherwise build failures are guaranteed!

Quoting from an earlier comment in this thread:

"I have inspected several dozens of different class files from jdt projects and they all seem reasonable. Unfortunately this PR fixes an issue in a super-common operation: string concatenation with + that it is going to trigger gazillion comparator errors in the SDK build.

The only difference pattern is MAXSTACK estimate going down wrt to baseline; The new value can be lower by as many makeConcatWithConstants invokedynamic instructions needed in a statement in the method."

srikanth-sankaran:
@laeubi I am trying to comprehend what you mean by "PR is never merged with comparator errors in JDT" - in this case there are comparator errors and having inspected several dozens of differing class files I concluded these are along the expected lines. So no bug is suspected. What should I have done differently before merging ??

This is asked not confrontationally.
I am happy to adopt a different methodology if prescribed. At the moment I don't know what alternatives exist. 

Thanks for educating/enlightening me.

srikanth-sankaran:
> Opened issue for the comparator errors: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1923
> 
> 
> 
> @srikanth-sankaran , @sravanlakkimsetti , how do we continue here? Who checks if the changes are expected?
> 
> 
> 
> Ideally we would have had 2 builds between the changes from #2181 and #2171, unfortunately we have only 1 build for both changes. So we don't know which PR results in which comparator errors.

Thanks @trancexpress : I'll follow up by inspecting the class files. I'll rope in help from @mpalat and @jarthana  as needed. 

Help from any one else is welcome too. 

The only change in class files I expect to see due to my code changes are: 

MAXSTACK value going down in methods that use Indy string concatenation operation.

srikanth-sankaran:
And @mpalat and @jarthana can confirm. I don't expect Java22 work to result in any comparator errors. 


To clarify, my recent PR corrects problems in code generation in miscomputing stack size. So it WILL produce class files that WILL DIFFER. So the comparator errors are not avoidable.

laeubi:
@srikanth-sankaran if you have inspected a difference and think its fine there are two options:

1. The affected bundle was not bumped, then increment the version by +100
2. It already was, in such case you usually add the bug number to the forceQualifierUpdate.txt

That way you should have a "clean" PR (of course other non jdt bundles later would maybe need version increments).

@trancexpress yes I mean comparator errors in jdt.core, even if they are only a few, they should be analyzed and fixed as part of the PR otherwise it might result in "unexpected" changes that the put pressure on fixing them when they are already in.

Of course there is no guarantee for that we not discover anything later, but what could be analyzed and fixed locally should be done at JDT, at best one run the aggregator build and check a wider range of bundle like described here:

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/?tab=readme-ov-file#build-with-custom-compiler

if you pass in `-Dcompare-version-with-baselines.skip=false` and `-Dtycho.debug.artifactcomparator` you can discover any changes (and even prepare PRs if you want) before running the full ibuild.

srikanth-sankaran:
> @srikanth-sankaran if you have inspected a difference and think its fine there are two options:
> 
> 
> 
> 1. The affected bundle was not bumped, then increment the version by +100
> 
> 2. It already was, in such case you usually add the bug number to the forceQualifierUpdate.txt
> 
> 
> 
> That way you should have a "clean" PR (of course other non jdt bundles later would maybe need version increments).


Let me understand. By a "clean" PR you mean a PR with no comparator errors against jdt core itself. Is that right ?

But a clean PR can produce comparator error in 200 other bundles legitimately, because a super basic super ubiquitous operation like string concatenation gets a bug fix ??

If so, the steps you suggest above don't sound like not a big win to me. Or have I misunderstood ??

I am with you that before merging in a PR, it is useful and necessary to see if there are comparator errors in JDT itself and validate them. 

In the last round for PR 2011 I didn't do it - it resulted in comparator errors in 167 bundles IIRC, from 1 bug and 2 not-incorrect but still avoidable change. It would have been prudent to have looked at jdt logs. 

So far I am with you. Where you lost me is what a compiler engineer is supposed to do when he has validated the comparator diffs in jdt core but still anticipates thousands of comparator errors across the SDK build. As it happened this time. I did inspect several dozens of class files before merging. 


> 
> 
> 
> @trancexpress yes I mean comparator errors in jdt.core, even if they are only a few, they should be analyzed and fixed as part of the PR otherwise it might result in "unexpected" changes that the put pressure on fixing them when they are already in.
> 
> 
> 
> Of course there is no guarantee for that we not discover anything later, but what could be analyzed and fixed locally should be done at JDT, at best one run the aggregator build and check a wider range of bundle like described here:


ISTM the raison d'etre for an integration build is to validate a build by bringing together components. And to discover problems in the process. 

Your suggestion seems to want to hoist the purpose of an integration build to outside the loop.

Ultimately it could be a problem of nomenclature. I challenge the labeling of builds with comparator errors as "unstable". That is poor choice perhaps. 

See that all tests across all SDK projects may be green and the build can still be labeled "unstable."





> 
> 
> 
> https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/?tab=readme-ov-file#build-with-custom-compiler
> 
> 
> 
> if you pass in `-Dcompare-version-with-baselines.skip=false` and `-Dtycho.debug.artifactcomparator` you can discover any changes (and even prepare PRs if you want) before running the full ibuild.

sravanlakkimsetti:
> I challenge the labeling of builds with comparator errors as "unstable". That is poor choice perhaps.

@srikanth-sankaran I am just listing the reasons for taking the approach of labelling "unstable". 

Earlier we used to publish the build and add its p2 repo to I-build composite. This has caused a problem when there are comparator errors specifically when we missed increment of swt versions(earlier this used to be a manual process). 
We used to have SWT jar having latest code and binaries jar having old code. This caused eclipse not to work. But comparator errors always caught this problem. Another side effect here is if anyone tried to upgrade using I build repository, their eclipse becomes unusable and recovery path is complete reinstallation.

To avoid this situation we came up with this approach Whenever there are comparator errors we label the build as unstable and not add to I-build repository. This way we have a clear warning to identify what is causing the trouble and at the same time protect upgrades from not using problematic build.

> See that all tests across all SDK projects may be green and the build can still be labeled "unstable."

When this labelling was introduced we have introduced a process to remove this label and add to I-build repository as well. [Mark Stable](https://ci.eclipse.org/releng/job/Builds/job/markStable/) job is specifically written to remove the label. This can be used if build is verified as stable build(tests are green and sanity tested).

We went with this approach to be better safe than sorry

PS: we also have [Mark Unstable](https://ci.eclipse.org/releng/job/Builds/job/markUnstable/) to mark a build unstable in case a build found to be problematic during the testing.

srikanth-sankaran:
Thanks for the explanation @sravanlakkimsetti

srikanth-sankaran:
> Opened issue for the comparator errors: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1923
> 
> 
> 
> @srikanth-sankaran , @sravanlakkimsetti , how do we continue here? Who checks if the changes are expected?
> 
> 
> 
> Ideally we would have had 2 builds between the changes from #2181 and #2171, unfortunately we have only 1 build for both changes. So we don't know which PR results in which comparator errors.

POA: I will come up with a script to mechanically check if there are any differences other than stack size going down. If there is no other differences and the tests are all green, we can move forward to updating the bundles.

laeubi:
> Let me understand. By a "clean" PR you mean a PR with no comparator errors against jdt core itself. Is that right ?

Correct.

> But a clean PR can produce comparator error in 200 other bundles legitimately, because a super basic super ubiquitous operation like string concatenation gets a bug fix ??

Yes that's something one currently can't avoid, but at least for the scope of the PR (this repository) everything is already fixed (so we save one additional commit afterwards) and probably an undesired change can be identified before the merge.

> If so, the steps you suggest above don't sound like a big win to me. Or have I misunderstood ??

Every little gain is worth it, you can't avoid the work anyways so why postpone it if it is guaranteed to fail?

> Where you lost me is what a compiler engineer is supposed to do when he has validated the comparator diffs in jdt core but still anticipates thousands of comparator errors across the SDK build. As it happened this time. I did inspect several dozens of class files before merging.

I'm not that deep inside the compiler topic itself but in the end it is the question who if not the PR writer is able to validate the side-effects?  So it is more if one want to make additional check before the submission, we can even enhance the tools if someone can give hints on how to do it, e.g can some things be automated (one example would be to map changed byte code back to the source code, give some comprehensive analysis about stack sizes and so on. But for that we need some support of the compiler experts.

**In this particular case the question would be:**

If I have two classes A and B and the only differ in the stack size, is this a relevant change that should lead to a comparator error? If not, a maybe mor sustainable fix would be to enhance the comparator to ignore stacksize changes.

srikanth-sankaran:
Good morning gentlemen,

The build looks good from tests p.o.v.  I inspected the tests that fail on all 5 platforms and we have only this one new:

- https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2193

(We also have https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2187 - don't know yet why this didn't fail in I-build)

srikanth-sankaran:
1. On a linux VM, downloaded and unzipped `artifactcomparisons.zip`  into current directory

Except for these errors - which I think are due to excessively long path names - the unzip went fine
```
error:  cannot create rt.equinox.p2/bundles/org.eclipse.equinox.p2.metadata.repository/target/artifactcomparison/no-classifier/ant_tasks/metadataRepository-ant.jar/org/eclipse/equinox/internal/p2/metadata/repository/ant/CreateCompositeMetadataRepositoryTask.class-baseline
        Input/output error
error:  cannot create rt.equinox.p2/bundles/org.eclipse.equinox.p2.metadata.repository/target/artifactcomparison/no-classifier/ant_tasks/metadataRepository-ant.jar/org/eclipse/equinox/internal/p2/metadata/repository/ant/CreateCompositeMetadataRepositoryTask.class-build
        Input/output error
error:  cannot create rt.equinox.p2/bundles/org.eclipse.equinox.p2.metadata.repository/target/artifactcomparison/no-classifier/ant_tasks/metadataRepository-ant.jar/org/eclipse/equinox/internal/p2/metadata/repository/ant/CreateCompositeMetadataRepositoryTask.class-baseline.class
        Input/output error
error:  cannot create rt.equinox.p2/bundles/org.eclipse.equinox.p2.metadata.repository/target/artifactcomparison/no-classifier/ant_tasks/metadataRepository-ant.jar/org/eclipse/equinox/internal/p2/metadata/repository/ant/CreateCompositeMetadataRepositoryTask.class-build.class
        Input/output error
error:  cannot create rt.equinox.p2/bundles/org.eclipse.equinox.p2.repository.tools/target/artifactcomparison/no-classifier/lib/repository-tools-ant.jar/org/eclipse/equinox/p2/internal/repository/tools/tasks/CreateCompositeArtifactRepositoryTask.class-baseline.class
        Input/output error
```
2. Issued these commands: 

`$ find 2>/tmp/err . -name '*.class-baseline'  > baseline-classes

$ cat /tmp/err   #### To verify no errors - yes no errors.

$ head baseline-classes 
./eclipse.jdt.core/org.eclipse.jdt.apt.tests/target/artifactcomparison/no-classifier/apt.jar/org/eclipse/jdt/apt/tests/annotations/readAnnotationType/AnnotationReader.class-baseline
./eclipse.jdt.core/org.eclipse.jdt.apt.tests/target/artifactcomparison/no-classifier/apt.jar/org/eclipse/jdt/apt/tests/annotations/pause/PauseAnnotationProcessor.class-baseline
./eclipse.jdt.core/org.eclipse.jdt.apt.tests/target/artifactcomparison/no-classifier/apt.jar/org/eclipse/jdt/apt/tests/annotations/nestedhelloworld/NestedHelloWorldAnnotationProcessor.class-baseline
./eclipse.jdt.core/org.eclipse.jdt.apt.tests/target/artifactcomparison/no-classifier/apt.jar/org/eclipse/jdt/apt/tests/annotations/mirrortest/MirrorUtilTestAnnotationProcessor.class-baseline
./eclipse.jdt.core/org.eclipse.jdt.apt.tests/target/artifactcomparison/no-classifier/apt.jar/org/eclipse/jdt/apt/tests/annotations/mirrortest/MirrorTestAnnotationProcessor.class-baseline
./eclipse.jdt.core/org.eclipse.jdt.apt.tests/target/artifactcomparison/no-classifier/apt.jar/org/eclipse/jdt/apt/tests/annotations/mirrortest/DefaultConstantProcessor.class-baseline
./eclipse.jdt.core/org.eclipse.jdt.apt.tests/target/artifactcomparison/no-classifier/apt.jar/org/eclipse/jdt/apt/tests/annotations/helloworld/HelloWorldAnnotationProcessor.class-baseline
./eclipse.jdt.core/org.eclipse.jdt.apt.tests/target/artifactcomparison/no-classifier/apt.jar/org/eclipse/jdt/apt/tests/annotations/generic/AbstractGenericProcessor.class-baseline
./eclipse.jdt.core/org.eclipse.jdt.apt.tests/target/artifactcomparison/no-classifier/apt.jar/org/eclipse/jdt/apt/tests/annotations/aptrounding/Round1GenAnnotationProcessor.class-baseline
./eclipse.jdt.core/org.eclipse.jdt.apt.tests/target/artifactcomparison/no-classifier/apt.jar/org/eclipse/jdt/apt/tests/annotations/apitest/APIAnnotationProcessorFactory$APIAnnotationProcessor.class-baseline

$ wc -l baseline-classes 
5236 baseline-classes

## So, OK we got `5236` classes to verify. Sanity check that we have 5236 classes from build

3. Using this script, I gathered all differences into a single text file:
```
#! /bin/bash
rm -f classfile-diff.out
for baseline_class in `cat baseline-classes`
do
    build_class=`dirname $baseline_class`/`basename -s .class-baseline $baseline_class`.class-build
    printf "diff $baseline_class $build_class\n" >> classfile-diff.out
    diff $baseline_class $build_class >> classfile-diff.out
done
``` 

### The accumulated text diffs look like this:
$head -30  classfile-diff.out
```
diff ./eclipse.jdt.core/org.eclipse.jdt.apt.tests/target/artifactcomparison/no-classifier/apt.jar/org/eclipse/jdt/apt/tests/annotations/readAnnotationType/AnnotationReader.class-baseline ./eclipse.jdt.core/org.eclipse.jdt.apt.tests/target/artifactcomparison/no-classifier/apt.jar/org/eclipse/jdt/apt/tests/annotations/readAnnotationType/AnnotationReader.class-build
127c127
<     MAXSTACK = 5
---
>     MAXSTACK = 2
diff ./eclipse.jdt.core/org.eclipse.jdt.apt.tests/target/artifactcomparison/no-classifier/apt.jar/org/eclipse/jdt/apt/tests/annotations/pause/PauseAnnotationProcessor.class-baseline ./eclipse.jdt.core/org.eclipse.jdt.apt.tests/target/artifactcomparison/no-classifier/apt.jar/org/eclipse/jdt/apt/tests/annotations/pause/PauseAnnotationProcessor.class-build
136c136
<     MAXSTACK = 5
---
>     MAXSTACK = 4
diff ./eclipse.jdt.core/org.eclipse.jdt.apt.tests/target/artifactcomparison/no-classifier/apt.jar/org/eclipse/jdt/apt/tests/annotations/nestedhelloworld/NestedHelloWorldAnnotationProcessor.class-baseline ./eclipse.jdt.core/org.eclipse.jdt.apt.tests/target/artifactcomparison/no-classifier/apt.jar/org/eclipse/jdt/apt/tests/annotations/nestedhelloworld/NestedHelloWorldAnnotationProcessor.class-build
34c34
<     MAXSTACK = 2
---
>     MAXSTACK = 1
diff ./eclipse.jdt.core/org.eclipse.jdt.apt.tests/target/artifactcomparison/no-classifier/apt.jar/org/eclipse/jdt/apt/tests/annotations/mirrortest/MirrorUtilTestAnnotationProcessor.class-baseline ./eclipse.jdt.core/org.eclipse.jdt.apt.tests/target/artifactcomparison/no-classifier/apt.jar/org/eclipse/jdt/apt/tests/annotations/mirrortest/MirrorUtilTestAnnotationProcessor.class-build
615c615
<     MAXSTACK = 8
---
>     MAXSTACK = 5
diff ./eclipse.jdt.core/org.eclipse.jdt.apt.tests/target/artifactcomparison/no-classifier/apt.jar/org/eclipse/jdt/apt/tests/annotations/mirrortest/MirrorTestAnnotationProcessor.class-baseline ./eclipse.jdt.core/org.eclipse.jdt.apt.tests/target/artifactcomparison/no-classifier/apt.jar/org/eclipse/jdt/apt/tests/annotations/mirrortest/MirrorTestAnnotationProcessor.class-build
136c136
<     MAXSTACK = 4
---
>     MAXSTACK = 3
diff ./eclipse.jdt.core/org.eclipse.jdt.apt.tests/target/artifactcomparison/no-classifier/apt.jar/org/eclipse/jdt/apt/tests/annotations/mirrortest/DefaultConstantProcessor.class-baseline ./eclipse.jdt.core/org.eclipse.jdt.apt.tests/target/artifactcomparison/no-classifier/apt.jar/org/eclipse/jdt/apt/tests/annotations/mirrortest/DefaultConstantProcessor.class-build
237c237
<     MAXSTACK = 6
---
>     MAXSTACK = 2
```

Now the job is to verify that only stack depth changes are encountered in this file. Anything else needs scruitiny

srikanth-sankaran:
I have mechanically verified that the textual diffs contains only stack size reductions. The one lone exception is called out below.

Now corelating the stacksize reductions to indy instructions count is not undertaken for the full set of `5236` classes. I had already verified it for several dozens of class files in JDT-Core before merge and given the tests are all green would assert we are good to go.

The exception is 

(1) this diff below that needs to be studied.  

```
diff ./eclipse.jdt.core/org.eclipse.jdt.core.tests.performance/target/artifactcomparison/no-classifier/org/eclipse/jdt/core/tests/performance/FullSourceWorkspaceBuildTests.class-baseline ./eclipse.jdt.core/org.eclipse.jdt.core.tests.performance/target/artifactcomparison/no-classifier/org/eclipse/jdt/core/tests/performance/FullSourceWorkspaceBuildTests.class-build
428c428
---
580c580
---
639c639
---
895c895
---
1191c1191
<       64: L6
---
>       38: L6
1201c1201
<       64: L10
---
>       38: L10
1248c1248
<       64: L19
---
>       38: L19
1258c1258
<       64: L23
---
>       38: L23
1326c1326
---
```
(2) the unzip errors: these class files would have to be analyzed somehow:

```
error:  cannot create rt.equinox.p2/bundles/org.eclipse.equinox.p2.metadata.repository/target/artifactcomparison/no-classifier/ant_tasks/metadataRepository-ant.jar/org/eclipse/equinox/internal/p2/metadata/repository/ant/CreateCompositeMetadataRepositoryTask.class-baseline
        Input/output error
error:  cannot create rt.equinox.p2/bundles/org.eclipse.equinox.p2.metadata.repository/target/artifactcomparison/no-classifier/ant_tasks/metadataRepository-ant.jar/org/eclipse/equinox/internal/p2/metadata/repository/ant/CreateCompositeMetadataRepositoryTask.class-build
        Input/output error
error:  cannot create rt.equinox.p2/bundles/org.eclipse.equinox.p2.metadata.repository/target/artifactcomparison/no-classifier/ant_tasks/metadataRepository-ant.jar/org/eclipse/equinox/internal/p2/metadata/repository/ant/CreateCompositeMetadataRepositoryTask.class-baseline.class
        Input/output error
error:  cannot create rt.equinox.p2/bundles/org.eclipse.equinox.p2.metadata.repository/target/artifactcomparison/no-classifier/ant_tasks/metadataRepository-ant.jar/org/eclipse/equinox/internal/p2/metadata/repository/ant/CreateCompositeMetadataRepositoryTask.class-build.class
        Input/output error
error:  cannot create rt.equinox.p2/bundles/org.eclipse.equinox.p2.repository.tools/target/artifactcomparison/no-classifier/lib/repository-tools-ant.jar/org/eclipse/equinox/p2/internal/repository/tools/tasks/CreateCompositeArtifactRepositoryTask.class-baseline.class
        Input/output error
```

I'll continue.

jarthana:
> > Its a bit unfortunate that this is not done, I would have expsected that a PR is never merged with comparator errors in JDT as otherwise build failures are guaranteed!
> 
> Sorry, JDT core is its own repository. You expect a PR in it to be blocked by the rest of the Eclipse platform having some checker that complains about the latest code in JDT core? That doesn't sound reasonable to me.
> 
> Maybe using the latest JDT for the platform build should be blocked by comparator errors, that sounds OK to me.

I agree. 
This is the reason why, imo, upgrading the compiler as a separate task periodically was better, as we will be expecting such troubles and even be better prepared.

srikanth-sankaran:
> I'm not that deep inside the compiler topic itself but in the end it is the question who if not the PR writer is able to validate the side-effects? So it is more if one want to make additional check before the submission, we can even enhance the tools if someone can give hints on how to do it,

I agree the PR author is in the best position to validate.

What would help me personally is:

1.  On a JDT/Core PR if a new "check" is added that would bubble up the presence of comparator errors. Checking logs is all OK, but humans being humans are going to fail/forget/neglect to do that with discipline. Tools calling attention would be welcome. If this check is implemented, let comparator errors be flagged with orange rather than with red X. Believe me sometimes, subtle changes in other phases could result in code generation change that the PR author would be completely surprised about. Happens.
2.  On the I-Build download/tests results page, I would like to view just failures. I don't want to search for a needle in a haystack
3. On the I-Build tests results page, I would like to see new failures color coded differently compared to failures in previous builds.
4. I would like flaky tests that repeatedly fail on a sporadic basis to be flagged as flaky - they should not "Fail" they should "Flaky Fail"
5. I would prefer the "Unstable" nomenclature to go away and become "Under-validation"  and red letters to become orange letters :)
6. Regarding Jay's comment above about upgrading the compiler as a periodic task: I may have had an hand in advocating the present practice - by having called out the "best practice" from javac/OpenJDK - every line of change you make is used to build the entire OpenJDK forest and it is the bootstrapped compiler that is used to run the test suites. But they don't have comparators :)

iloveeclipse:
> This is the reason why, imo, upgrading the compiler as a separate task periodically was better, as we will be expecting such troubles and even be better prepared.

The problem with that approach is, that it makes it harder to find out **which** change in compiler caused regression and to  provide and validate an appropriate fix in same release. We had bugs appearing in RC builds. Patching bugs in code generation at this time is risky, not patching is shipping buggy compiler. So I believe current immediate validation is the right thing. The extra work we have results in a better quality of the release.

laeubi:
I'll try to provide an extra validation step to JDT, given we have these comparator errors now seems a good way to test ist but I would maybe need some "extra power" from JDT commiters to allow workflow runs.

laeubi:
See PR here:
- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2194

PR #2175, opened 2024-03-18
Modify grammar to create switch nodes after parsing expression

### LINKED ISSUE #2155: ASTParser cannot recover this statement that SelectionParser can
For `ResolveTests1_5.testBrokenSwitch0` and `ResolveTests1_5.testBrokenSwitch1`, the SelectionParser is able to recover the statements in the method, but the AssistParser isn't. It would be nice to move the hacks from the SelectionParser to the AssistParser so that ASTs created through ASTParser can access the recovered statements.

Here is the class used in the first test case; the second test case is similar.

```java
public class Test {
    private static ILog test() {
        return new ILog() {
            @Override
            public void log(String status) {
                switch (status.length()) { // here
                case
                }
            }
        };
    }
}
```

datho7561:
## What it does
Modifies the parser so that switch statements can be recovered in more cases, such as when they're in the body of a method in an anonymous class that's declared in a method.

Fixes #2155

Signed-off-by: David Thompson <davthomp@redhat.com>

## How to test
View the attached test case.

I expect that this will mostly be helpful for clients consuming `ASTParser`, since many cases within jdt.core that need recovered nodes use the `SelectionParser` or the `CompletionParser`, which do a much better job of recovering broken AST nodes.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
@@ -8403,7 +8403,10 @@ protected void consumePostExpressionInIf() {
 }
 
 protected void consumePostExpressionInSwitch(boolean statSwitch) {
-	// for sub-types to hook into
+	SwitchStatement switchStatement = statSwitch ? new SwitchStatement() : new SwitchExpression();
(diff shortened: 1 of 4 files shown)


datho7561:
I abandoning this for now. It barely improves the fault tolerance of ASTParser, and will require a lot of changes in CompletionParser and SelectionParser in order to ensure they continue to work properly.

PR #2182, opened 2024-03-20
tests: Convert String to TextBlock

jukzi:
automated Cleanup


jukzi:
@jjohnstn is it intentional, that there are escaped newlines in the String block after clean-up? in org.eclipse.jdt.core.tests.performance.FullSourceWorkspaceModelTests.testPerfReconcileBigFileWithSyntaxError()
image
Honestly i don't understand that string before clean-up either. That test still passes.

jukzi:
superseeded by https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2282

PR #2015, opened 2024-02-13
fix zipSlip

jukzi:
https://github.com/eclipse-jdt/eclipse.jdt.core/security/code-scanning/34

PR #2185, opened 2024-03-20
fix zipSlip .. regression

jukzi:
https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2015#issuecomment-2009162226

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/util/Util.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/util/Util.java
@@ -3358,4 +3358,13 @@ public class Util {
 		}
 		return method;
 	}
+
+	public static String getEntryName(String zipfileName, ZipEntry entry) {
+		String entryName = entry.getName();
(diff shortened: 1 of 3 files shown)


jukzi:
@sratz

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/util/Util.java
```diff
@@ -3358,4 +3358,13 @@ public static IMethod findMethod(IType type, char[] selector, String[] paramType
 		}
 		return method;
 	}
+
+	public static String getEntryName(String zipfileName, ZipEntry entry) {
+		String entryName = entry.getName();
+		if (!java.nio.file.Path.of(zipfileName, entryName).normalize()
```
sratz, line 3364: I was confused at first how `Path` would be able to handle zip entries, but you're just using the `Path` API to handle normalization and `startsWith()` in a semantically correct way (not `String.startsWith()`).

Neat trick! Looks good

sratz:
Is this ready to be merged?

PR #2189, opened 2024-03-20
Provide an API to check if an IType is an implicitly declared class.

datho7561:
## What it does
Provide an API to check if an `IType` is an implicitly declared class.

## How to test
I've included a test case. The only way to manually test this would be to consume the API.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ImplicitTypeDeclaration.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ImplicitTypeDeclaration.java
@@ -12,7 +12,6 @@ package org.eclipse.jdt.internal.compiler.ast;
 
 import java.nio.file.Path;
 import java.nio.file.Paths;
-import java.text.MessageFormat;
 import java.util.stream.Stream;
 
 import org.eclipse.jdt.internal.compiler.CompilationResult;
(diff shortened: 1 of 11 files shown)


datho7561:
That's why there aren't any new test failures, then!

I saw that I was still removing the message format and assumed I was still making that change in this commit.

The test failure seems unrelated, and is present in `master`

datho7561:
@jarthana @jjohnstn @rgrunber if you have time, could you please take a look at this PR?

rgrunber:
Is there any reason to do a minor bump on `org.eclipse.jdt.core.tests.builder` & `org.eclipse.jdt.core.tests.compiler` ? Doesn't seem anything was changed there.

datho7561:
> Is there any reason to do a minor bump on `org.eclipse.jdt.core.tests.builder` & `org.eclipse.jdt.core.tests.compiler` ? Doesn't seem anything was changed there.

I'll try removing the bump

srikanth-sankaran:
Given the nomenclature change to implicitly declared classes, should we be using a mixture of "unnamed" and "implicitly declared" internally ? Does it serve a purpose ?

rgrunber:
It might make sense to just use "unnamed". The underlying implementation seems to be tied to unnamed classes. It wouldn't help someone determine if a class was generated by Lombok, but it might have consumers of the API wrongly thinking it might.

srikanth-sankaran:
Not sure I understand. "Unnamed classes" was old terminology. Subsequent JEP changed it to Implicitly declared classes - no ? So why would it be useful to continue to refer to withdrawn nomenclature ??

rgrunber:
Whoops, I overlooked the ["changed to implicitly declared classes"](https://openjdk.org/jeps/463) in your comment. Thought you were only implying the new API should change to match the existing "unnamed classes" one. It definitely makes sense to use the new terminology going forward.

datho7561:
I think we should use "implicitly declared class" consistently. I wasn't sure whether or not to make that change to the API AST node as a part of this PR, so I left it unchanged.

PR #2200, opened 2024-03-21
ASTParser doesn't parse task tags

### LINKED ISSUE #2179: ASTParser doesn't parse task tags
ASTParser never parses task tags. See https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java#L381

Can we relax this restriction?
The https://github.com/eclipse-jdtls/eclipse-jdt-core-incubator project uses ASTParser in https://github.com/eclipse-jdtls/eclipse-jdt-core-incubator/blob/dom-based-operations/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/CompilationUnit.java#L195 and need to parse task tags.

I have commented ASTParser.java, line 381 and all tests pass locally.

A related issue - https://github.com/eclipse-jdtls/eclipse-jdt-core-incubator/issues/84
See https://github.com/eclipse-jdtls/eclipse-jdt-core-incubator/pull/160/commits/b2e311daa3ebc2c4d9472078d3ca1b63de3fe450#diff-5155d744fdf221c04316cc2e0751b9ed207740927ccb1a36d82f85ca4e2bee5cR382

snjeza:
## What it does
Fixes #2179 

## How to test
Run the jdt.core tests

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
+++ b/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
@@ -378,7 +378,8 @@ public class ASTParser {
 			// copy client's options so as to not do any side effect on them
 			options = new HashMap<>(options);
 		}
-		options.remove(JavaCore.COMPILER_TASK_TAGS); // no need to parse task tags
+		// https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2179
+		// options.remove(JavaCore.COMPILER_TASK_TAGS); // no need to parse task tags
 		this.compilerOptions = options;
 	}
 


rgrunber:
Looking through the history, the only thing I could find on removing the task tags was a [bug](https://bugs.eclipse.org/bugs/show_bug.cgi?id=307295) where having an imbalance between task tags and their corresponding task priorities caused issues. But that is handled by `Util.fixTaskTags`, which is separate from this. So I think this should probably be ok.

mickaelistria:
IMO, this patch is actually a fix because the statement "no need to parse task tags" and explicitly removing a setting that the consumer might want to override seems really incorrect: it'd be fine if the task gets ignored by default, but if the options do explicitly require them, they must be obeyed.
Note that there is nothing in the API contract the specify this behavior, while the documentation tells that the options are used; so this statement partly breaks the API doc.

rgrunber:
I noticed a bunch of others instances.

https://github.com/eclipse-jdt/eclipse.jdt.core/blob/8fd859497d7031f90cf469758a4ba4058042f58d/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java#L287

https://github.com/eclipse-jdt/eclipse.jdt.core/blob/8fd859497d7031f90cf469758a4ba4058042f58d/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java#L659

https://github.com/eclipse-jdt/eclipse.jdt.core/blob/8fd859497d7031f90cf469758a4ba4058042f58d/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java#L699

https://github.com/eclipse-jdt/eclipse.jdt.core/blob/8fd859497d7031f90cf469758a4ba4058042f58d/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java#L839

It probably doesn't matter for the client, but might be good for the sake of consistency to apply the same change.

mickaelistria:
There is another more complex problem that makes that the output tasks are not accessible #2277 . I think fixing #2277 should almost be a pre-requisite before further changes here.

PR #2249, opened 2024-03-29
Fix IAE computing type signature for anonymous + var

mickaelistria:
Unlike what's specified by TypeBinding.signableType(), the return value is incorrect for anonymous type and that leads to an exception such as

  java.lang.IllegalArgumentException: new I(){}
  	at org.eclipse.jdt.core.Signature.createCharArrayTypeSignature(Signature.java:1109)
  	at org.eclipse.jdt.core.Signature.createTypeSignature(Signature.java:1305)
  	at org.eclipse.jdt.core.dom.VariableBinding.getUnresolvedJavaElement(VariableBinding.java:277)

Some specific code is added here to handle that case.

## What it does

Fixes an IllegalArgumentError

## How to test

Run provided test without patch to see the failure, then run with the fix to see the fix

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverter10Test.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverter10Test.java
@@ -13,20 +13,37 @@
  *******************************************************************************/
 package org.eclipse.jdt.core.tests.dom;
 
-import junit.framework.Test;
-
-import org.eclipse.jdt.core.dom.*;
-import org.eclipse.jdt.internal.compiler.classfmt.ClassFileConstants;
-import org.eclipse.jdt.internal.compiler.impl.CompilerOptions;
-
 import java.io.IOException;
 import java.util.Hashtable;
 
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.jdt.core.ICompilationUnit;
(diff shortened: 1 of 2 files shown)


org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverter10Test.java
```diff
@@ -13,20 +13,37 @@
  *******************************************************************************/
 package org.eclipse.jdt.core.tests.dom;
 
-import junit.framework.Test;
-
-import org.eclipse.jdt.core.dom.*;
```
gayanper, line 18: Any reason for not keeping the star import as it was ?

gayanper (approved):
LGTM

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverter10Test.java
mickaelistria: No, it was the auto-formatting which changed that. I can revert it if you prefer?

gayanper: If the format settings are same as what I configured in the project, we can keep the change.

akurtakov (approved):
Looks good to me too.

Issue #1728, opened 2023-12-11
UI Freezes since 2023-09


gigermocas:
I've been having constant UI freezes on my eclipse environment since updating to the 2023-09 release, triggering from simple copy/paste operations and function drill downs on java code. 2023-06 is working fine.

I did open first a more generic issue on https://github.com/eclipse-platform/.github/issues/157 first, but I think it was prematurely closed as not planed due to a miscommunication, and my last message was not seen (?)

Eclipse 2023-12 was released so I made a clean installation (with no external plugins), and imported all my projects and I'm experiencing the same behavior.
[trace.log](https://github.com/eclipse-jdt/eclipse.jdt.core/files/13633960/trace.log)

iloveeclipse:
What exactly is in your workspace, how many projects, targeting which Java version? 
Any chance for steps to reproduce ?

jukzi:
please take a serious sampling snapshot with for example https://visualvm.github.io/ and post stacktrace of hotspots or make the full result available for download

gigermocas:
I took a couple of snapshots here, hope it helps
[snapshots.zip](https://github.com/eclipse-jdt/eclipse.jdt.core/files/13660241/snapshots.zip)

jukzi:
Thanks, i can read them. however i don't see an obvious error despite that the compilation should not happen in the UI Thread
![image](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/51790620/c7e8494f-edac-4a39-a1b2-f47caf492714) but in the background. If you can contribute a solution that would be nice.

gigermocas:
Don't know if I'm following correctly, this should not trigger a compilation, I'm not changing any files here

gigermocas:
Ok, didn't knew it did that, wasn't expecting that through simply trying to copy text from the source file.

jukzi:
@gigermocas any contribution to make it better would be very welcome. I am closing this issue as "not planned". We could reopen if you could contribute a PR which makes it any better.

gigermocas:
Thanks, but the specific annotation that's causing issues for me is @XmlSeeAlso which is not TYPE_USE but USE, I guess that wouldn't help me much here.

gigermocas:
Content assist being slower in this instance may be unavoidable, but the thing that makes this extra processing more impactful is it being triggered during a simple select text + copy operation, which can cause an UI freeze, and is something that doesn't actually use any of that information (?).

[...]

PR #2270, opened 2024-04-02
UI Freezes since 2023-09

stephan-herrmann:
experimental fix for #1728

 disable annotation resolution / evaluation for some clients

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/LookupEnvironment.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/LookupEnvironment.java
@@ -166,6 +166,14 @@ public class LookupEnvironment implements ProblemReasons, TypeConstants {
 			return this; // no-change to signal "at end"
 		}
 
+		boolean isRequired(boolean buildFieldsAndMethods, boolean resolveAnnotations) {
+			return switch (this) {
+				case BUILD_FIELDS_AND_METHODS -> buildFieldsAndMethods;
+				case INTEGRATE_ANNOTATIONS_IN_HIERARCHY -> resolveAnnotations;
+				default -> true;
+			};
+		}
+
 		/** values without NONE */
 		static final CompleteTypeBindingsSteps[] realValues = Arrays.copyOfRange(values(), 1, values().length-1);
 
@@ -601,11 +609,14 @@ public void completeTypeBindings(CompilationUnitDeclaration parsedUnit) {
 * suitable replacement will be substituted (such as Object for a missing superclass)
 */
 public void completeTypeBindings(CompilationUnitDeclaration parsedUnit, boolean buildFieldsAndMethods) {
+	completeTypeBindings(parsedUnit, buildFieldsAndMethods, true);
+}
+public void completeTypeBindings(CompilationUnitDeclaration parsedUnit, boolean buildFieldsAndMethods, boolean resolveAnnotations) {
 	if (parsedUnit.scope == null) return; // parsing errors were too severe
 	LookupEnvironment rootEnv = this.root;
 	CompilationUnitDeclaration previousUnitBeingCompleted = rootEnv.unitBeingCompleted;
 	for (CompleteTypeBindingsSteps step : CompleteTypeBindingsSteps.realValues) {
-		if (step != CompleteTypeBindingsSteps.BUILD_FIELDS_AND_METHODS || buildFieldsAndMethods)
+		if (step.isRequired(buildFieldsAndMethods, resolveAnnotations))
 			step.perform((rootEnv.unitBeingCompleted = parsedUnit).scope);
 	}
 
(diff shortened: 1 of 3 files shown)


stephan-herrmann:
This draft change (for performance reasons) quite drastically pulls the plug on any treatment of annotations when invoked from **codeSelect** or **codeComplete** or when a UI action requests a **type hierarchy**. More specifically all of the "new" phase `INTEGRATE_ANNOTATIONS_IN_HIERARCHY` will be skipped for all types except for the focus type (this logic being applied only on the `askForType()` -> `accept()` path).

@srikanth-sankaran could you think of any reasons why such use cases could **require** resolved information about annotations on referenced types? Here 'referenced' includes those referenced as supertypes of the focus type.

stephan-herrmann:
@gigermocas would you be able to directly use this change for some testing? That would greatly help to find out:
* does the change improve performance?
* does it cause any regressions?

gigermocas:
Yes, this a bit more manageable. Actively editing a source file and using content assist is snappy.

Trying to copy some text while editing the file still freezes and locks the UI, there seems to be some contention there ([trace.log](https://github.com/eclipse-jdt/eclipse.jdt.core/files/14850124/trace.log)). I have to wait until the background compilation triggered by the file change completely stops, which takes several seconds. If there's no background compilation the copy operation is lighter there and by itself doesn't freeze the UI anymore.

I must admit I don't fully understand it, I would expect that copying some text should work regardless of background compilation, as triggering the content assist seems to be doing, does this make sense?

Otherwise I'm not seeing any other issues, I will daily drive this and report back if I see anything weird. Thanks!

stephan-herrmann:
> Yes, this a bit more manageable. Actively editing a source file and using content assist is snappy.

thanks for your feedback! I'm glad my guess at a culprit was not completely off the mark! :)

> Trying to copy some text while editing the file still freezes and locks the UI, there seems to be some contention there ([trace.log](https://github.com/eclipse-jdt/eclipse.jdt.core/files/14850124/trace.log)). I have to wait until the background compilation triggered by the file change completely stops, which takes several seconds. If there's no background compilation the copy operation is lighter there and by itself doesn't freeze the UI anymore.
> 
> I must admit I don't fully understand it, I would expect that copying some text should work regardless of background compilation, as triggering the content assist seems to be doing, does this make sense?

The connection here is, that `org.eclipse.jdt.internal.ui.javaeditor.ClipboardOperationAction` wants to add imports that are required to make the pasted code compilable. Computing these imports is what requires a resolved AST, which in turn requires access to workspace resources that could be locked by the build.

If you see room for improvement in this area please file a new issue for our brothers and sisters at eclipse.jdt.ui (with the latest log file attached)
 
> Otherwise I'm not seeing any other issues, I will daily drive this and report back if I see anything weird. Thanks!

Great, so I'll wait -- say -- one week for you to report back. If no new issues are observed, I'll go ahead an merge.

gigermocas:
I'm noticing that the type hierarchy is also requested through trying to open the  "Run As"/"Debug As" menu ([stack.txt](https://github.com/eclipse-jdt/eclipse.jdt.core/files/14874554/stack.txt)), resulting in a freeze as well. Can annotation processing be disabled from there as well?

> The connection here is, that org.eclipse.jdt.internal.ui.javaeditor.ClipboardOperationAction wants to add imports that are required to make the pasted code compilable. Computing these imports is what requires a resolved AST, which in turn requires access to workspace resources that could be locked by the build.

I see, and that behavior I can actually disable in the editor typing preferences, good to know!

stephan-herrmann:
> I'm noticing that the type hierarchy is also requested through trying to open the "Run As"/"Debug As" menu ([stack.txt](https://github.com/eclipse-jdt/eclipse.jdt.core/files/14874554/stack.txt)), resulting in a freeze as well. Can annotation processing be disabled from there as well?

The stack only shows the initial `completeTypeBindings(CompilationUnitDeclaration[], boolean[], int) ` for which work on annotations are not (yet) suppressed, but this shouldn't matter, the bulk of the work is done only via `accept()` (for additional types). One of the `accept()` methods already has the short-circuit, will add it to a second one just now.

If you still observe slowness, look out for `integrateAnnotationsInHierarchy()`, that's the guy we are trying to avoid in the current experiment.

gigermocas:
I got tripped up by that. The real culprit in this case is `org.eclipse.jdt.internal.junit.launcher.JUnit5TestFinder.internalIsTest`, which requires a resolved AST ([trace.log](https://github.com/eclipse-jdt/eclipse.jdt.core/files/14885933/trace.log)), this is something more like the copy with imports situation sadly.

stephan-herrmann:
> The real culprit in this case is `org.eclipse.jdt.internal.junit.launcher.JUnit5TestFinder.internalIsTest`, which requires a resolved AST

In this case the caller who knows what is needed (`JUnit5TestFinder`) is many stack levels away from `completeTypeBindings()` where I inserted the new flag. 

I suggest to postpone this for now. Once we are confident that the new flag does help (to some degree) and introduces no regressions, then we can think of ways to tell `ASTParser` or similar guys that annotations on super types are not of interest, so that clients like JUnit5TestFinder can simply opt-out.

stephan-herrmann:
@gigermocas with one more confirmation from your side I'd be ready to merge. Did you observe any ill effects of the change? 

Further use cases where the same strategy could be applied can be taken up separately.

gigermocas:
It's been working great here. I didn't encounter any related issues.

Regarding a more customizable AST parser, I hope you guys find that compelling enough to have a go at it. That would greatly improve the performance in the remaining areas.

PR #2275, opened 2024-04-03
Fix NPE in SourceRangeVerifier (when debugging)

mickaelistria:
Some nodes may return null for structuralFeatures if those nodes are usually not available at given AST/Preview level (eg StringTemplateExpression.propertyDescriptors(apiLevel, previewEnabled). So let's handle this possible null case more gracefully.

## What it does

Prevents a NPE

## How to test

Build an AST using StringTemplates via ASTParser, but without enabling preview, the AST is generated and reports some problem, but if SourceVerifier.DEBUG is enabled, then a NPE will be thrown.
The logic of the test would be relatively hard to maintain, as the example above will change, and thus test will either become irrelevant or fail for no good reason, when the StringTemplate become non-preview.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/SourceRangeVerifier.java
+++ b/org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/SourceRangeVerifier.java
@@ -57,6 +57,9 @@ public class SourceRangeVerifier extends ASTVisitor {
 		ASTNode previous = null;
 
 		List properties = node.structuralPropertiesForType();
+		if (properties == null) { // happens for some nodes that aren't usually available at AST level
+			return false;
+		}
 		for (Object p : properties) {
 			StructuralPropertyDescriptor property = (StructuralPropertyDescriptor) p;
 			if (property.isChildProperty()) {


gayanper:
@mickaelistria bit curious, returning preVisit -> false will stop visiting the rest of the tree, could there be a situation where this node is not valid because it's a preview-feature, but rest of the expression might be ?

mickaelistria:
>  could there be a situation where this node is not valid because it's a preview-feature, but rest of the expression might be ?

I can't think of any; but I'm not 100% confident here.

mickaelistria:
To reproduce this issue, you can add a test in EmbeddedExpressionCompletionTests
```java
	public void testASTParserCrashWhenPreviewNotSetAndDebug() throws JavaModelException {
		this.workingCopies = new ICompilationUnit[1];
		this.workingCopies[0] = getWorkingCopy(
				"/Completion/src/X.java",
				"""
				public class X {
					String v = "lol";
					String s = STR."Hello \\{v}";
				}
				""");
		SourceRangeVerifier.DEBUG = true;
		this.workingCopies[0].getJavaProject().setOption(JavaCore.COMPILER_PB_ENABLE_PREVIEW_FEATURES, JavaCore.DISABLED);
		ASTParser astParser = ASTParser.newParser(AST.JLS22);
		astParser.setSource(this.workingCopies[0]);
		astParser.setResolveBindings(true);
		astParser.setStatementsRecovery(true);
		astParser.setBindingsRecovery(true);
		ASTNode ast = astParser.createAST(null);
	}
```
But again, as it;s heavily relying on preview features which are moveable, adding it as a test case will cause trouble later.

gayanper:
Make sense, and given that currently it silently throws which stops the visitor and also cause a error, the fix here is much better. So LGTM unless someone raise any concerns.

gayanper (approved):

akurtakov:
Being null safe here is far better than throwing NPE.

Issue #2284, opened 2024-04-05
Bug561287Test fails when run on Java 22


akurtakov:
Details:
```
junit.framework.ComparisonFailure: Unexpected problem(s)!!!.
----------- Expected ------------

------------ but was ------------
Problem : The project was not built since its build path is incomplete. Cannot find the class file for java.lang.Object. Fix the build path then try building this project [ resource : </Bug561287Test> range : <-1,-1> category : <10> severity : <2>]
Problem : The type java.lang.Object cannot be resolved. It is indirectly referenced from required .class files [ resource : </Bug561287Test/src/bug561287/A0.java> range : <0,1> category : <10> severity : <2>]

---------------------- ----------
expected:<[]> but was:<[Problem : The project was not built since its build path is incomplete. Cannot find the class file for java.lang.Object. Fix the build path then try building this project [ resource : </Bug561287Test> range : <-1,-1> category : <10> severity : <2>]
Problem : The type java.lang.Object cannot be resolved. It is indirectly referenced from required .class files [ resource : </Bug561287Test/src/bug561287/A0.java> range : <0,1> category : <10> severity : <2>]
]>
at org.eclipse.jdt.core.tests.junit.extension.TestCase.assertStringEquals(TestCase.java:267)
at org.eclipse.jdt.core.tests.junit.extension.TestCase.assumeEquals(TestCase.java:287)
at org.eclipse.jdt.core.tests.builder.BuilderTests.expectingNoProblemsFor(BuilderTests.java:255)
at org.eclipse.jdt.core.tests.builder.BuilderTests.expectingNoProblemsFor(BuilderTests.java:241)
at org.eclipse.jdt.core.tests.builder.BuilderTests.expectingNoProblems(BuilderTests.java:235)
at org.eclipse.jdt.core.tests.builder.Bug561287Test.testBuilderRegression(Bug561287Test.java:66)
at java.base/java.lang.reflect.Method.invoke(Method.java:580) 
``` 
and visible at https://download.eclipse.org/eclipse/downloads/drops4/I20240404-1800/testresults/html/org.eclipse.jdt.core.tests.builder_ep432I-unit-cen64-gtk3-java22_linux.gtk.x86_64_22.html .

akurtakov:
If I change compliance level to 22 (TestingEnvironments modified to add support for it) test works so this makes me look in the ctsym direction. Any pointers are more than welcome.

akurtakov:
[...]

Issue #1875, opened 2024-01-16
[22][test] TypeHieararchyTests to be modified for JDK 22


mpalat:
The following tests fail when run on JDK 22 after applying pr#1841

- org.eclipse.jdt.core.tests.model.TypeHierarchyTests.testIndexQualificationLambdaReferences_AsReturnTypes
- org.eclipse.jdt.core.tests.model.TypeHierarchyTests.testIndexQualificationLambdaReferences_AsParameters

Fix/Modify these after pr #1841  is merged - these will be returning without running in the merged code for 22 JDK

Issue #1874, opened 2024-01-16
[22][test] ModuleBuilderTests to be modified for JDK22


mpalat:
These tests are failing once pr #1841  is used.

- org.eclipse.jdt.core.tests.model.ModuleBuilderTests.testConvertToModule
- org.eclipse.jdt.core.tests.model.ModuleBuilderTests.testConvertToModuleWithRelease9
- org.eclipse.jdt.core.tests.model.ModuleBuilderTests.testBug512053
- org.eclipse.jdt.core.tests.model.ModuleBuilderTests.testBug527569c
- org.eclipse.jdt.core.tests.model.ModuleBuilderTests.testReleaseOption1
- org.eclipse.jdt.core.tests.model.ModuleBuilderTests.testReleaseOption4
- org.eclipse.jdt.core.tests.model.ModuleBuilderTests.testReleaseOption5
- org.eclipse.jdt.core.tests.model.ModuleBuilderTests.testReleaseOption8
- org.eclipse.jdt.core.tests.model.ModuleBuilderTests.testReleaseOption9
- org.eclipse.jdt.core.tests.model.ModuleBuilderTests.testReleaseOption11
- org.eclipse.jdt.core.tests.model.ModuleBuilderTests.testReleaseOption12
- org.eclipse.jdt.core.tests.model.ModuleBuilderTests.testReleaseOption13

would return if jre22 in the pr - need to re-enable after fixing

jarthana:
I see something weird going on with the ct.sym of the JDK 22. I see that the classes not being loaded when --release is specified. I suspect they messed up the paths inside the ct.sym file.

@iloveeclipse Is this something you can look at?

iloveeclipse:
@jarthana : I can. Which exact JDK 22 build have you tried?

iloveeclipse:
I've tried to run ModuleBuilderTests locally on `openjdk-22-ea+32` build and commit 31ef12dd145f22210d90dc53596f0799d420482b (the latest one I see on BETA_JAVA22 branch) and everything is fine, I see no fails.

I've also briefly checked if the ct.sym structure / content looks different, but couldn't spot any major change (except new content for older releases). I'm using my very basic "JRT Browser" view which you can get importing project from 
https://github.com/iloveeclipse/jrtbrowser repo into workspace and run debugger IDE with that.

iloveeclipse:
[...]

PR #2286, opened 2024-04-05
Adopt release detection code to consider ct.sym changes in Java 22

iloveeclipse:
Adopt release detection code to consider ct.sym changes in Java 22

The ct.sym file contains with Java 22 no "system-modules" file in the "current" release directory anymore.

The structure of the root level of ct.sym file indicates that the release contents are in single letter directories and all diffs between releases seem to be in multiple letter directories.

To answer a question whether current JDK is 12+ or not it is enough to check that any single letter directory name is higher then 'B' (== Java 11) and we don't need to rely on existence of the "system-modules" file.

Additionally re-enabled few tests that were disabled in 31ef12dd145f22210d90dc53596f0799d420482b.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2284
Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1875
Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1874

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/CtSym.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/CtSym.java
@@ -399,9 +399,8 @@ public class CtSym {
 				}
 				try {
 					char releaseCode = rel.charAt(0);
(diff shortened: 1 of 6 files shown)


iloveeclipse:
Arrgh, I see some tests were muted via 31ef12dd145f22210d90dc53596f0799d420482b! I will fix that too now.

PR #2289, opened 2024-04-05
Fix ASTRewriteAnalyzer.ListRewriter to remove end keyword

### LINKED ISSUE #2288: Splitting try-with-resources at first resource statement creates invalid try statement
The initial try-with-resources is not correctly converted to a regular try statement (the trailing parentheses is not removed)

```
package tests;

import java.io.BufferedReader;
import java.io.FileNotFoundException;
import java.io.FileReader;
import java.io.IOException;
import java.io.Reader;

public class TestTryWithResourceExtract {

    public void foo() throws IOException {
        try (Reader s = new BufferedReader(new FileReader("c.d"));
                Reader r = new BufferedReader(new FileReader("a.b"));
                Reader t = new BufferedReader(new FileReader("e.f"))) {
            r.read();
            System.out.println("abc");
        } catch (FileNotFoundException e) {
            e.printStackTrace();
        } catch (IOException e) {
            e.printStackTrace();
        }
    }
}
```

changes to:

```
package tests;

import java.io.BufferedReader;
import java.io.FileNotFoundException;
import java.io.FileReader;
import java.io.IOException;
import java.io.Reader;

public class TestTryWithResourceExtract {

    public void foo() throws IOException {
        try) {
            try (Reader s = new BufferedReader(new FileReader("c.d"));
                    Reader r = new BufferedReader(new FileReader("a.b"));
                    Reader t = new BufferedReader(new FileReader("e.f"))) {
                r.read();
                System.out.println("abc");
            } 
        } catch (FileNotFoundException e) {
            e.printStackTrace();
        } catch (IOException e) {
            e.printStackTrace();
        }
    }
}

```

jjohnstn:
- when we remove all elements in list and we have an end keyword, we should remove the end keyword if it is specified
- fixes #2288

## What it does

Fixes ASTRewriteAnalyzer.ListRewriter to remove end keyword if specified and we are removing all items

## How to test

See issue.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/rewrite/ASTRewriteAnalyzer.java
+++ b/org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/rewrite/ASTRewriteAnalyzer.java
@@ -761,7 +761,7 @@ public final class ASTRewriteAnalyzer extends ASTVisitor {
 						currPos= end;
 						prevEnd= currEnd;
 						// if all removed and we are removing last item - we must remove end keyword
-						if (nextIndex == total && lastNonDelete == -1 && endKeyword.length() > 0) {
+						if (nextIndex == total && lastNonDelete == -1 && endKeyword != null && endKeyword.length() > 0) {
 							try {
 								TokenScanner scanner = getScanner();
 								int tempOffset = scanner.getNextEndOffset(currPos, true);
--- a/org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/rewrite/ASTRewriteAnalyzer.java
+++ b/org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/rewrite/ASTRewriteAnalyzer.java
@@ -760,6 +760,17 @@ public final class ASTRewriteAnalyzer extends ASTVisitor {
 						}
 						currPos= end;
 						prevEnd= currEnd;
+						// if all removed and we are removing last item - we must remove end keyword
+						if (nextIndex == total && lastNonDelete == -1 && endKeyword.length() > 0) {
+							try {
+								TokenScanner scanner = getScanner();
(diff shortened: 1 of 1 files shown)


org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/rewrite/ASTRewriteAnalyzer.java
```diff
@@ -603,6 +603,10 @@ protected boolean mustRemoveSeparator(int originalOffset, int nodeIndex) {
 			return true;
 		}
 
+		protected boolean mustRemoveEndKeywordWhenEmpty() {
```
gayanper, line 606: According the issue description, the last part we want to remove could be a token as well right ? Should we call this `LastToken` ? Or did you misunderstood the issue and naming used here :(

jjohnstn: Hi @gayanper I was basing on the passing of keyword/end-keyword to the rewriteList() method from try-with-resources where ")" was the end-keyword for the resources plus the fact that the remove all was commented as removing the keyword.  If you feel that should be mustRemoveLastToken() I am fine with renaming.  I could also change the arguments to pass in a token to remove so that it ensures that token is removed plus any preceding comments.  I assumed that an invalid token at the end of the resources list couldn't occur such that a removal of resources would be possible, but again I may be assuming to much.

gayanper (approved):
LGTM

org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/rewrite/ASTRewriteAnalyzer.java
gayanper: I think the current name looks good according to the above comment.

jjohnstn:
Thanks @gayanper, merged

PR #2296, opened 2024-04-07
Erroneous "Cannot infer type arguments" error

### LINKED ISSUE #1475: Erroneous "Cannot infer type arguments" error
With generics, the `Cannot infer type arguments` error is shown erroneously.

Take the following (simplified, so slightly bizarre) example:
```
public class CannotInferTypeArguments<V extends java.util.concurrent.Semaphore> {
    class Fish {
        public V getFlavour() {
            return null;
        }
    }

    class Shark<E extends Fish> {
    }

    <E extends Fish> Shark<E> fish() {
        // This compiles fine with javac, but will only work in Eclipse with new Shark<E>();
        return new Shark<>();
    }
}
```

This compiles fine with `javac`, but not in Eclipse.
When `new Shark<>()` is changed to `new Shark<E>()`, Eclipse is then happy as well.

This _may_ be related to [this bug](https://bugs.eclipse.org/bugs/show_bug.cgi?id=566989) which the same symptom but with a slightly different scenario.

I've reproduced this in Eclipse 2023-09 (4.29.0) on Windows (Java 19) & OSX (Java 17).

stephan-herrmann:
fixes #1475

Implementation of JLS 15.9.3. failed to substitute an enclosing type in
>  - The return type of mj is θj applied to D<F1, ..., Fp>.

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
@@ -5382,7 +5382,9 @@ public abstract class Scope {
 						}
 				}
 			}
-			staticFactory.returnType = environment.createParameterizedType(genericType, Scope.substitute(substitution, genericType.typeVariables()), originalEnclosingType);
+			staticFactory.returnType = environment.createParameterizedType(genericType,
+					Scope.substitute(substitution, genericType.typeVariables()),
+					(ReferenceBinding) Scope.substitute(substitution, originalEnclosingType));
 			staticFactory.parameters = Scope.substitute(substitution, method.parameters);
 			staticFactory.thrownExceptions = Scope.substitute(substitution, method.thrownExceptions);
 			if (staticFactory.thrownExceptions == null) {
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericsRegressionTest_1_8.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericsRegressionTest_1_8.java
@@ -10682,4 +10682,32 @@ public void testBug508834_comment0() {
 			"----------\n",
 			null, true, customOptions);
 	}
+	public void testGH1475() {
+		runConformTest(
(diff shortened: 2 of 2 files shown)


srikanth-sankaran (approved):
Looks good. Mmm what has taken so long for this to be caught ?

srikanth-sankaran:
A quick pass over some of the existing bugs: 

https://bugs.eclipse.org/bugs/show_bug.cgi?id=576846 - not reproducible with or without this fix - should be closed with a regression test.
https://bugs.eclipse.org/bugs/show_bug.cgi?id=569231 - reproducible with or without this fix
https://bugs.eclipse.org/bugs/show_bug.cgi?id=568586 - Lacks a full test case
https://bugs.eclipse.org/bugs/show_bug.cgi?id=566989 - not reproducible with or without this fix - should be closed with a regression test.
https://bugs.eclipse.org/bugs/show_bug.cgi?id=551929 - reproducible with or without this fix
https://bugs.eclipse.org/bugs/show_bug.cgi?id=509848 - not reproducible with or without this fix - should be closed with a regression test.

There could be value in coming up with regression tests from the working cases ((after comparing behavior with javac  of course) - even if they were not fixed by this PR - so we have our test suite running them on a regular basis

stephan-herrmann:
Some bisecting:

* https://bugs.eclipse.org/bugs/show_bug.cgi?id=569231
  * fixed by #1557 
* https://bugs.eclipse.org/bugs/show_bug.cgi?id=566989
  * also :smile:  fixed by #1557 
* https://bugs.eclipse.org/bugs/show_bug.cgi?id=509848
  * fixed since ecj-4.11 (v20190306-2240) - too old for my git-bisect environment :)

regression tests coming in a minute

stephan-herrmann:
> Mmm what has taken so long for this to be caught ?

Java provides too many features for all Java programmers on the planet to ever try all feature combinations? :smile: 

To trigger the bug you need:
* a generic member type
* hosted in a generic enclosing type
* a generic method of that enclosing type
* inside that method a diamond instantiation of the member type

jarthana:
We seem to be rejecting the following code (which Javac accepts) after this fix:
```
class Outer <T extends Number>{
    public void foo(){
        class ABC <U extends Number> {
            public ABC(int i, long l, String s, U t) {}
        }
        ABC<?> a = new ABC<>(5, 3L, "", 5) {};
    }
}
```

Issue #1591, opened 2023-11-12
ECJ fails to compile method invocation while javac successfully compiles


coehlrich:
```java
    public void test() {
    	Supplier<? extends List<? extends Serializable>> supplier = () -> null;
        error(supplier.get());
    }

    public <T, V extends Serializable> void error(List<V> v2) {}
```
ECJ fails to compile with the message `The method error(List<V>) in the type Outer is not applicable for the arguments (capture#1-of ? extends List<? extends Serializable>)` whereas javac successfully compiles the method.

jukzi:
probably related to https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1332

stephan-herrmann:
> probably related to #1332

I don't think so. Cast compatibility and type inference use different algorithms.

But I can report that the problem vanishes when you remove the unused type parameter `<T>`

OTOH, the problem is not cause by the fact that `<T>` is unused, as this variant demonstrates:
```java
import java.io.Serializable;
import java.util.List;
import java.util.function.Supplier;

public class Outer {
    public void test() {
    	Supplier<? extends List<? extends Serializable>> supplier = () -> null;
        error(supplier.get(), "");
    }

    public <T, V extends Serializable> void error(List<V> v2, T t) {}
    
}
```

Now `<T>` should have a suitable solution (String), but still we report
> The method error(List<V>, T) in the type Outer is not applicable for the arguments (capture#1-of ? extends List<? extends Serializable>, String)

stephan-herrmann:
Inference fails when it encounters this constraint: `⟨V#1 = ? extends java.io.Serializable⟩`

Legally, [JLS 18.2.4](https://docs.oracle.com/javase/specs/jls/se22/html/jls-18.html#jls-18.2.4) has no provision to accept this constraint (because for accepting, both sides must be of the same kind: both wildcards or none a wildcard).

So one of the following must be true:
* it is correct to reject the program, _or_
* our inference shouldn't produce that constraint in the first place, _or_
* somewhere implicitly the spec author assumed that an equality constraint involving an inference variable doesn't require reduction / can directly be translated into an equality type bound, _or_ can it be dropped?

stephan-herrmann:
> * somewhere implicitly the spec author assumed that an equality constraint involving an inference variable doesn't require reduction / can directly be translated into an equality type bound, _or_ can it be dropped?

[...]

PR #2305, opened 2024-04-09
ECJ fails to compile method invocation while javac successfully compiles

stephan-herrmann:
work on #1591

hapless attempt to mimic javac behavior, causing regressions in GenericsRegressionTest_1_8.testBug472851()


stephan-herrmann:
Here's what happened: analysis in #1591 did not give any hint about a bug in ecj (other than the fact that javac accepts, while ecj rejects).

For the time being there is nothing more we can do.

I have, however, added the test case to a new test suite, capturing the current outcome. Should some future code change effect this test, we will be informed by way of a failure (which then needs to be checked if that failure is indeed an improvement). See https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2305/files#diff-0e52fa5757c780704f8d75a913f25a639b18f55cebadf6957c383bd8fab3a85a

PR #2311, opened 2024-04-10
ASTParser uses CompilationUnit options

mickaelistria:
Instead of parent project one, as the CompilationUnit may override some options.

## What it does

Make that ASTParser honors ICompilationUnit compiler options, which may differ from the project options.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
+++ b/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
@@ -602,7 +602,9 @@ public class ASTParser {
 	 *
 	 * <p>This method automatically sets the project (and compiler
 	 * options) based on the given compilation unit, in a manner
-	 * equivalent to {@link #setProject(IJavaProject) setProject(source.getJavaProject())}.</p>
+	 * equivalent to {@link #setProject(IJavaProject) setProject(source.getJavaProject())}
+	 * and the custom compiler options supported by the compilation unit through
+	 * {@link ICompilationUnit#getCustomOptions() getCustomOptions()}.</p>
 	 *
 	 * <p>This source is not used when the AST is built using
 	 * {@link #createASTs(ICompilationUnit[], String[], ASTRequestor, IProgressMonitor)}.</p>
@@ -612,6 +614,9 @@ public class ASTParser {
 	 */
 	public void setSource(ICompilationUnit source) {
 		setSource((ITypeRoot)source);
+		if (source != null) {
+			setCompilerOptions(source.getOptions(true));
+		}
 	}
 
 	/**
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/CompilationUnit.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/CompilationUnit.java
@@ -1453,11 +1453,13 @@ public void setOptions(Map<String, String> newOptions) {
 
 @Override
 public Map<String, String> getCustomOptions() {
-	try {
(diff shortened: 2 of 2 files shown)


rgrunber (commented):
An `ICompilationUnit`'s default `getOptions(boolean)` is :

```
IJavaProject parentProject = getJavaProject();
Map<String, String> options = parentProject == null ? JavaCore.getOptions() : parentProject.getOptions(inheritJavaCoreOptions);
Map<String, String> customOptions = getCustomOptions();
if (customOptions != null) {
	options.putAll(customOptions);
}
return options;
```

If you look inside `setSource(ITypeRoot)` (immediately above the line you're adding), you'll see :

```
this.project = source.getJavaProject();
Map<String, String> options = this.project.getOptions(true);
options.remove(JavaCore.COMPILER_TASK_TAGS); // no need to parse task tags
this.compilerOptions = options;
```

So which compilation unit options are not making their way into the compiler options field ? Is it some custom options ?

mickaelistria:
> So which compilation unit options are not making their way into the compiler options field ? Is it some custom options ?

Yes, if one builds some CompilationUnit instance with some customOptions, those are fully ignored by ASTParser. In a future story. Note that custom options were recently introduced by @testforstephen in 21e1b31837b67234715496b5b73638bc3dae4268 , so I think their support in ASTParser was just forgotten.
In ASTParser, specific compilationUnit options can achieve very interesting stories, such as allowing somehow to target different Java version between src vs tests.

testforstephen:
When I introduce the custom options, the primary purpose is to allow storing the file-level formatting options within the Compilation Unit (CU) model. These options include the indentation settings such as `insertSpace` and `tabSize`. By doing so, we ensure that the ASTRewrite can accurately respect the formatting options specified at the file level. For instance, when rewriting a file, the ASTRewrite will apply the correct indentation based on the file’s formatting options.

pls note that the ASTParser, which handles the initial parsing of the file, does not require these formatting options.

mickaelistria:
> pls note that the ASTParser, which handles the initial parsing of the file, does not require these formatting options.

Yes, those particular options are not relevant to ASTParser, but some others which can now be customized could be. For example, some compilation units could override some options to ignore some particular warnings in test code or issues of that sort. There might not be user stories for it yet, but I think it'd still be consistent that the ASTParser uses the most contextual options (the ones from the CU) over more generic ones; this can serve many future stories.

rgrunber:
Yup, it seemed like this was capturing the custom options, but I knew those are only used for capturing per-file indentation options. With that said I agree that in the future (or for clients), other options might make sense to be passed in.

I would just update the javadocs. `setOptions(ICompilationUnit)` no longer just _"sets the project (and compiler options) based on the given compilation unit of class file, in a manner equivalent to `setProject(source.getJavaProject())`."_

There probably needs be something added to the end ".. in a manner equivalent to `setProject(source.getJavaProject())` with the additional custom compiler options supported by the compilation unit."

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
```diff
@@ -612,6 +614,7 @@ public void setSource(char[] source) {
 	 */
 	public void setSource(ICompilationUnit source) {
 		setSource((ITypeRoot)source);
+		setCompilerOptions(source.getOptions(true));
```
rgrunber, line 617: The test failures are because of this part. This needs a `null` check because :

```
@param source the Java model compilation unit whose source code
	 * is to be parsed, or <code>null</code> if none
```

The code formatter logic also has some code that attempts to parse code as various different kinds (cu, expression, body, statement, module, etc.) and when this new statement is now called, it fails because `source.getOptions(true)` triggers some hashcode logic that requires the CU owner, which is `null`.

PR #2327, opened 2024-04-11
null analysis should merge null contracts from equivalent super methods

### LINKED ISSUE #2325: null analysis should merge null contracts from equivalent super methods
In #2308 @howlger came up with this problematic snippet:
```java
import org.eclipse.jdt.annotation.Nullable;

interface InterfaceA {
    @Nullable Object get(); 
}

interface InterfaceB {
    @NonNull Object get(); 
}

interface InterfaceAB extends InterfaceA, InterfaceB {}

interface InterfaceBA extends InterfaceB, InterfaceA {}

class Sample {

    void ab(InterfaceAB ab) {
        @NonNull Object obj = ab.get();
                           // ^^^^^^^^
                           // ⚠ Null type mismatch (type annotations): required '@NonNull Object' but this expression has type '@Nullable Object'
                           // Expected: no "Null type mismatch" problem,
                           //           because the union of the two null constraints has to be @Nullable, the most restrictive null constraint
                           //           (@Nullable violates the null constraint given by InterfaceB; @NonNull fulfills both null constraints from InterfaceA and InterfaceB)
    }

    void ba(InterfaceBA ba) {
        @NonNull Object obj = ba.get(); // (no "Null type mismatch" as expected)
    }

}
```

Indeed null analysis should interpret the "phantom" methods in `InterfaceAB` and `InterfaceBA` as being constrained by the union of null contracts of both super methods, since any implementing class will be forced (via `The return type is incompatible with '@NonNull Object' returned from InterfaceA.get() (mismatching null constraints)`) to fulfill the stronger of both contracts.

stephan-herrmann:
fixes #2325

+ if method choice would be arbitrary prefer more specific nullness

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/NullAnnotationMatching.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/NullAnnotationMatching.java
@@ -816,6 +816,26 @@ public class NullAnnotationMatching {
 		return mainType;
 	}
 
+	/**
+	 * Help Scope.mostSpecificMethodBinding(MethodBinding[], int, TypeBinding[], InvocationSite, ReferenceBinding):
+	 * If choice between equivalent methods would otherwise be arbitrary, determine if m1 should be preferred due
+	 * to a more specific null contract.
+	 */
+	public static boolean hasMoreSpecificNullness(MethodBinding m1, MethodBinding m2) {
+		long nullness1 = m1.returnType.tagBits & TagBits.AnnotationNullMASK;
+		long nullness2 = m2.returnType.tagBits & TagBits.AnnotationNullMASK;
+		if (nullness1 == TagBits.AnnotationNonNull && nullness2 != TagBits.AnnotationNonNull)
+			return true;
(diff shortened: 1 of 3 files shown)


stephan-herrmann:
I would characterize this change as cheating with explicit permission by JLS :)

Here's why I claim the change to be legitimate:
* the code section handles the case in [JLS 15.12.2.5](https://docs.oracle.com/javase/specs/jls/se22/html/jls-15.html#jls-15.12.2.5)  where "no method is the most specific, because there are two or more methods that are maximally specific"
* we found that method `current` fulfills all criteria when checked against method `next`
* but then we find that `next` has the more specific null contract (which otherwise must be ignored during overload resolution)
* here I nudge the search to skip ahead to `next` as the new `current` (outer loop)
* this implies that the method with more specific null contract will be checked vis-a-vis all other candidates.

And voilà, while writing this, I see that I need a safety net: If that `next` method fails some test, I should be able to return to the original strategy, where `current` was a promising candidate already.

@srikanth-sankaran  Feel free to raise additional concerns.

stephan-herrmann:
> [23cabee](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2327/commits/23cabeecc4cd8224b800963a852e1340d1a6546b)

safety net added, phew.

PR #2341, opened 2024-04-13
Generate reference to classes holding constants used in the code

### LINKED ISSUE #1382: Generate reference to classes holding constants used in the code
Originally https://bugs.eclipse.org/bugs/show_bug.cgi?id=549455

We learned that https://bugs.openjdk.org/browse/JDK-7153958 deliberately made javac emit unnecessary CONSTANT_Class entries in class files, to record "dependencies" on classes of which constants have been inlined in the code.

This creates a hidden contract between javac and tools consuming that entry for their analysis, so it should have been driven by a spec change rather than an unannounced implementation change, but given that the change is close to trivial, ecj might as well follow suit.

stephan-herrmann:
fixes #1382

Emit a class constant into the constant pool for every class from which a constant is used.

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Clinit.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Clinit.java
@@ -357,6 +357,17 @@ public class Clinit extends AbstractMethodDeclaration {
 			codeStream.recordPositionsFrom(0, declaringType.sourceStart);
 			classFile.completeCodeAttributeForClinit(codeAttributeOffset, classScope);
 		}
+		// the following block must happen after constantPool.resetForClinit()
+		if (TypeDeclaration.kind(declaringType.modifiers) != TypeDeclaration.ENUM_DECL
+				&& fieldDeclarations != null) {
+			int constantFlags = ClassFileConstants.AccStatic | ClassFileConstants.AccFinal;
+			for (FieldDeclaration fieldDecl : fieldDeclarations) {
+				if ((fieldDecl.modifiers & constantFlags) == constantFlags
+						&& fieldDecl.initialization instanceof NameReference nr) {
+					nr.emitDeclaringClassOfConstant(codeStream);
+				}
+			}
+		}
 	}
 
 	@Override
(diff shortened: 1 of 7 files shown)


stephan-herrmann:
@iloveeclipse  results from API-Tools signal bytecode changes, indeed. This could cause quite a landslide of comparator errors in the SDK build. Please suggest how / when to proceed.

As mentioned in the issue #1382  this will eliminate one small difference wrt javac.

iloveeclipse:
@stephan-herrmann : 
1) we can merge this and build SDK now to get the list of (most likely all) affected bundles. 
2) After that one should create ticket for unstable SDK build at https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues and check if our expectation **why** the unstable build is unstable matches the class file diffs (there should be this change in constant pool only).
3) After that for every bundle that is reported to be "unstable" one has to "touch" it, set a PR and merge.
The work of touching bundles is boring, but can be automated via https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/wiki/Platform-Releng-Faq#how-to-fix-unstable-build
4) After that we can start next SDK build to validate everything is touched and stable.

So if you are confident in this PR, feel free to merge and we can continue with points above.

iloveeclipse:
Actually you can touch affected JDT bundles already that make this PR red as of now.

iloveeclipse:
@stephan-herrmann : if you plan to merge it today, please do it before 00:00 CET so the next build can pick it and we can start looking into "touching" bundles tomorrow without extra build.

stephan-herrmann:
> Actually you can touch affected JDT bundles already that make this PR red as of now.

done

> if you plan to merge it today, please do it before 00:00 CET so the next build can pick it and we can start looking into "touching" bundles tomorrow without extra build.

At that time I was already offline, and in fact I thought better not to mess up a Monday build (as those are the ones retained longer than other days, right?) - but then I got my timezone arithmetic wrong, so perhaps _now_ I should indeed wait until the Monday build it good.

iloveeclipse:
@stephan-herrmann : just merge if you are done, if you wish, now. I can help next day to fix the SDK build.

stephan-herrmann:
> @stephan-herrmann : just merge if you are done, if you wish, now. I can help next day to fix the SDK build.

Thanks!

iloveeclipse:
See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/1979.

Issue #2169, opened 2024-03-15
StackOverflowError on autocomplete imports on modular projects


AGT96:
Hello:

I found a bug when I autocomplete a import on JDK11 onwards based projects. When a JDK11+ is added to the module path in the project settings and I use the autocomplete feature on imports a java.lang.StackOverflowError is throwed.

I m using CTRL + Space to autocomplete imports on class imports section.

Can you fix this?

Anything tell me please.

[stacktrace.txt](https://github.com/eclipse-jdt/eclipse.jdt.core/files/14614708/stacktrace.txt)

![IMG1](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/42464066/20f02a65-7272-4c79-b77a-c7699a1d1265)
![IMG2](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/42464066/2f799381-95fd-4e19-b878-d1c7e40c3e33)

AGT96:
https://github.com/eclipse-jdt/eclipse.jdt.core/assets/42464066/3dae16a1-4fff-4ed2-80d9-a60b12ef6dd3

AGT96:
![stacktrace img autocomplete bug on jdk11+ eclipe ide](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/42464066/683e5f35-3ef7-4f58-b6dd-632676c632ac)

AGT96:
This bug appears when install the e(fx)clipse plugin. However if this plugin is not installed, the autocomplete doesn't show the suggestions to autocomplete the import and the exception is not thrown in this case.

iloveeclipse:
Relevant part of the stack
```
java.lang.StackOverflowError
	at org.eclipse.jdt.internal.core.JavaModelCache.getInfo(JavaModelCache.java:139)
	at org.eclipse.jdt.internal.core.JavaModelManager.getInfo(JavaModelManager.java:2289)
	at org.eclipse.jdt.internal.core.JavaElement.getElementInfo(JavaElement.java:306)
	at org.eclipse.jdt.internal.core.JavaElement.getElementInfo(JavaElement.java:294)
	at org.eclipse.jdt.internal.core.JavaProject.getModuleDescription(JavaProject.java:3763)
	at org.eclipse.jdt.internal.core.SearchableEnvironment.findPackagesFromRequires(SearchableEnvironment.java:356)
[...]

PR #2346, opened 2024-04-15
StackOverflowError on autocomplete imports on modular projects

iloveeclipse:
Converts recursive module dependencies search into stack based.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2169

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/ModuleDescriptionInfo.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/ModuleDescriptionInfo.java
@@ -13,8 +13,10 @@
  *******************************************************************************/
 package org.eclipse.jdt.internal.core;
 
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 
 import org.eclipse.jdt.core.IJavaElement;
 import org.eclipse.jdt.core.IModuleDescription;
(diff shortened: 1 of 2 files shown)


iloveeclipse:
@stephan-herrmann, @mpalat : this is the fix for the fix in c43bada8f824033f76fd2417f54d46b6a0eb569e.

It is a draft as I don't have a test case and not familiar enough with the modules in order to construct one.

From the https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2169 issue I assume either the javafx contributes lot of extra modules with extra dependencies that simply overflow the (limited) stack size, or they contain circular dependencies (however I've assumed that shouldn't be possible due the spec).

I believe the change itself is straight forward, with the hashcode/equals added to the `org.eclipse.jdt.internal.core.ModuleDescriptionInfo.ModuleReferenceInfo` that isn't considering "super" fields as they aren't used at all while constructing `ModuleReferenceInfo` instances.

Would be nice if you could review that .

iloveeclipse:
@stephan-herrmann, @mpalat : any objections merging this?

stephan-herrmann:
I'm looking into this, trying to come up with a test.

stephan-herrmann:
@iloveeclipse the fix is good essentially. Still I chose to create a new PR #2358 with a reduced version of your fix:
* keep the added `equals()` and `hashCode()` methods (although no other type in that hierarchy has them)
* keep collecting visited references and refuse to traverse any reference more than once
* OTOH I reverted to the recursive style, which IMHO is more direct and shouldn't hurt since the depth of any real life module graph should never come near creating a problem wrt stack size.

The good news: my test can demonstrate both problems:
* even in a "good" module graph, the same module may be reachable via many paths, causing (a lot of) redundant work
  * there is no assert for this, but just an added log statement easily demonstrates this
* adding a module cycle (though causing a compile error) would not stop completion to try going in cycles infinitely, i.e., the test succeeds in overflowing the stack when the fix is removed

iloveeclipse:
With https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2358 we can close this one. Thanks @stephan-herrmann !

PR #2370, opened 2024-04-19
Cache JRT-FileSystem created for JrtFileSystem instances

HannesWell:
## What it does

When obtaining the `JrtFileSystem` for different releases of the same JDK using `JRTUtil.getJrtSystem(File, String)` a new _JRT_ `FileSystem` is created for each release passed. But for the same JDK-installation the `FileSystem` is always backed by the same JRT file and since it is read-only I want to proposed to reuse the file-system for all releases.
This would reduce memory usage and avoid the performance penalty of creating the `FileSystem` more than once.

Fetching the `JrtFileSystem` for all releases since Java-9 for a JDK is a common pattern in PDE and its (future) use of the API proposed via https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/294. For example for a Java-21 VM we currently have 12 open JRT file-systems. 
 
## How to test

Using the API introduced with https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/294 one can use this _very_ simple benchmark test that must be run as JUnit Plug-in test.
```
public void testPerformance() throws CoreException {
	IVMInstall vm = JavaRuntime.getDefaultVMInstall();
	List<Set<String>> packages = new ArrayList<>();
	for (int r = 9; r <= 21; r++) {
		packages.add(JavaRuntime.getProvidedVMPackages(vm, "" + r));
	}
	System.out.println(packages);
}
```
On my computer (Ryzen 9 5900X, SSD drive) this test terminates in a little bit more than 1200ms without this change and in ~800ms with this change in repetitive runs.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
@@ -76,6 +76,7 @@ public class JRTUtil {
 	 * Map from JDK home path to ct.sym file (located in /lib in the JDK)
 	 */
 	private static final Map<Path, CtSym> ctSymFiles = new ConcurrentHashMap<>();
+	private static final Map<String, FileSystem> JRT_FILE_SYSTEMS = new ConcurrentHashMap<>();
 
 	static final SoftClassCache classCache = new SoftClassCache();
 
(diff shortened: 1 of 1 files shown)


iloveeclipse (changes requested):
I've noticed lot of code that tried to use jrtfs in a way that could be further improved, I will try to come up with a bigger patch.

iloveeclipse:
See https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2376

HannesWell:
> See #2376

Awesome. Thank you!
Closing this in favor of #2376.

PR #2371, opened 2024-04-19
Code simplifications in JRTUtil and ClasspathJRT classes

HannesWell:
## What it does
Apply multiple minor code simplifications where suitable to simplify reading the code.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/tool/JrtFileSystem.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/tool/JrtFileSystem.java
@@ -28,6 +28,7 @@ import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
@@ -42,7 +43,7 @@ import org.eclipse.jdt.internal.compiler.util.JRTUtil;
 
 public class JrtFileSystem extends Archive {
 
-	public HashMap<String, Path> modulePathMap;
+	Map<String, Path> modulePathMap;
 	Path modules;
 	private java.nio.file.FileSystem jrtfs;
 
@@ -55,7 +56,7 @@ public class JrtFileSystem extends Archive {
 		// initialize packages
 		this.modulePathMap = new HashMap<>();
 		if (this.file.exists()) {
-			this.jrtfs = JRTUtil.getJrtFileSystem(this.file.getAbsolutePath());
+			this.jrtfs = JRTUtil.getJrtFileSystem(this.file.toPath());
 			this.modules = this.jrtfs.getPath(JRTUtil.MODULES_SUBDIR);
 		} else {
 			return;
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
@@ -17,7 +17,6 @@ package org.eclipse.jdt.internal.compiler.util;
 import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.FileNotFoundException;
-import java.io.FileReader;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URI;
@@ -28,12 +27,11 @@ import java.nio.file.FileSystemAlreadyExistsException;
 import java.nio.file.FileSystemNotFoundException;
 import java.nio.file.FileSystems;
 import java.nio.file.FileVisitResult;
-import java.nio.file.FileVisitor;
 import java.nio.file.Files;
 import java.nio.file.NoSuchFileException;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.nio.file.ProviderNotFoundException;
+import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.util.ArrayList;
 import java.util.Collections;
@@ -54,7 +52,7 @@ public class JRTUtil {
 	public static final boolean DISABLE_CACHE = Boolean.getBoolean("org.eclipse.jdt.disable_JRT_cache"); //$NON-NLS-1$
 	public static final boolean PROPAGATE_IO_ERRORS = Boolean.getBoolean("org.eclipse.jdt.propagate_io_errors"); //$NON-NLS-1$
 
-	public static final String JAVA_BASE = "java.base".intern(); //$NON-NLS-1$
+	public static final String JAVA_BASE = "java.base"; //$NON-NLS-1$
 	public static final char[] JAVA_BASE_CHAR = JAVA_BASE.toCharArray();
 	public static final String MODULES_SUBDIR = "/modules"; //$NON-NLS-1$
 	static final String[] DEFAULT_MODULE = new String[]{JAVA_BASE};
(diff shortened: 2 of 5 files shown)


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
```diff
@@ -479,21 +450,19 @@ boolean sameRelease(String other) {
...
 		}
-		return home;
 	}
 
-	static String readJdkReleaseFile(String javaHome) throws IOException {
+	static String readJdkReleaseFile(Path javaHome) throws IOException {
 		Properties properties = new Properties();
-		try(FileReader reader = new FileReader(new File(javaHome, "release"))){ //$NON-NLS-1$
-			properties.load(reader);
+		try (InputStream in = Files.newInputStream(javaHome.resolve("release"))) { //$NON-NLS-1$
```
HannesWell, line 464: The `java.util.Properties` class states in its doc that it expect the content to be encoded using the `ISO 8859-1` encoding. Using a bare `InputStream` allows the implementation to handle it as it as expected instead of relying on a Reader that in this case uses the default encoding (which is UTF-8 in the latest JDKs).

That being said, I assume for the value read, the encoding should not really matter since it is simple ASCII.

HannesWell:
@iloveeclipse or @akurtakov, could you have a look at this?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
```diff
@@ -160,9 +133,8 @@ public static JrtFileSystem getJrtSystem(File image, String release) throws IOEx
 					throw new RuntimeIOException(errorMessage, e);
 				}
 			});
-			return system;
```
iloveeclipse, line 163: Please keep such code "as is" (here and below), it's otherwise hard to debug maps hits etc.

HannesWell: Done, restored it.
In such cases I just look at the `<methodName> returned` field in the Variables view of the Debug perspective, but having it as a variable makes it even simpler.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/ClasspathJrtWithReleaseOption.java
```diff
@@ -137,16 +134,16 @@ Map<String, SimpleSet> findPackagesInModules() {
 		if (this.modPathString == null) {
 			return Map.of();
 		}
-		Map<String, SimpleSet> cache = PackageCache.computeIfAbsent(this.modPathString, key -> {
+		return PackageCache.computeIfAbsent(this.modPathString, key -> {
```
iloveeclipse, line 137: Please keep such code "as is", it's otherwise hard to debug maps hits etc.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
```diff
@@ -144,11 +119,9 @@ public static JrtFileSystem getJrtSystem(File image) {
 	 */
 	public static JrtFileSystem getJrtSystem(File image, String release) throws IOException {
 		Jdk jdk = new Jdk(image);
-		String key = jdk.path;
-
-		if (release != null && !jdk.sameRelease(release)) {
-			key = key + "|" + release; //$NON-NLS-1$
-		}
+		String key = release != null && !jdk.sameRelease(release) //
```
iloveeclipse, line 122: For me the original version is easier to read, and the new ternary expression is harder. I would keep it "as is"

iloveeclipse (changes requested):
This will have merge conflicts with https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2376 I've just merged to mastzer, also see two comments below.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/ClasspathJrtWithReleaseOption.java
HannesWell: Revered to previous state.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
```diff
@@ -174,11 +148,11 @@ public static JrtFileSystem getJrtSystem(File image, String release) throws IOEx
 	 * @throws IOException
 	 *             on any error
 	 */
-	public static FileSystem getJrtFileSystem(String path) throws IOException {
+	public static FileSystem getJrtFileSystem(Path path) throws IOException {
 		try {
-			FileSystem fs = JRT_FILE_SYSTEMS.computeIfAbsent(path, p -> {
+			FileSystem fs = JRT_FILE_SYSTEMS.computeIfAbsent(path.toRealPath(), p -> {
```
HannesWell, line 153: This ensures that in all cases only one `FileSystem` exists for the same JRT-file, regardless of the difficulties the path contains (being relative or having `..` elements, symlinks, etc.).

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/tool/JrtFileSystem.java
```diff
@@ -55,7 +56,7 @@ public void initialize() throws IOException {
 		// initialize packages
 		this.modulePathMap = new HashMap<>();
 		if (this.file.exists()) {
-			this.jrtfs = JRTUtil.getJrtFileSystem(this.file.getAbsolutePath());
+			this.jrtfs = JRTUtil.getJrtFileSystem(this.file.toPath());
```
HannesWell, line 59: Btw. I noticed that this field is overwritten in the `JrtFileSystemWithOlderRelease` sub-class with the CtSystem, but `modRoot` isn't.
I assume that's all intentional? (I'm not very familiar with that code, I just wanted to make sure all is as it should be:) )

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
```diff
@@ -534,8 +507,8 @@ class JrtFileSystem {
 	private final Map<String, List<String>> packageToModules = new HashMap<>();
 
 	FileSystem fs;
```
HannesWell, line 509: Btw. I noticed that this field is overwritten in the `JrtFileSystemWithOlderRelease` sub-class with the CtSystem, but `modRoot` isn't.
I assume that's all intentional? (I'm not very familiar with that code, I just wanted to make sure all is as it should be:) )

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
HannesWell: Actually this can also be used in the two other caches using Path keys.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
iloveeclipse: Yes, it kind of intentional. The `JrtFileSystemWithOlderRelease` doesn't care about it because it overwrites the only method in superclass that uses this field. This isn't optimal, but for now we can leave it.

iloveeclipse (approved):

iloveeclipse:
Thanks.

PR #2375, opened 2024-04-22
Add Java 23 constants for AST

akurtakov:
## What it does
Backports AST Java 23 constants and helper methods of https://github.com/eclipse-jdt/eclipse.jdt.core/commit/3e78c6e69a239a51d3d4e7c8c070ec488ce9fa89 to master.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/AST.java
+++ b/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/AST.java
@@ -470,6 +470,22 @@ public final class AST {
 	 */
 	public static final int JLS22 = 22;
 	/**
+	 * Constant for indicating the AST API that handles JLS23.
+	 * <p>
+	 * This API is capable of handling all constructs in the
+	 * Java language as described in the Java Language
+	 * Specification, Java SE 23 Edition (JLS23).
+	 * JLS23 is a superset of all earlier versions of the
+	 * Java language, and the JLS23 API can be used to manipulate
+	 * programs written in all versions of the Java language
+	 * up to and including Java SE 23(aka JDK 23).
+	 * </p>
+	 *
+	 * @noreference This field is not intended to be referenced by clients. Java 23 support is WIP.
+	 * @since 3.38
+	 */
+	public static final int JLS23 = 23;
+	/**
 	 * Internal synonym for {@link #JLS15}. Use to alleviate
 	 * deprecation warnings once JLS15 is deprecated
 	 */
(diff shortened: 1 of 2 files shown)


akurtakov:
@jarthana @mpalat Is this one fine according to you? It makes it easier to experiment with Java 23 features.

jarthana:
As you can see from the doc, can be used to manipulate
	 * programs written in all versions of the Java language
	 * up to and including Java SE 22

jarthana:
As you can see from the doc, 
`JLS23 API can be used to manipulate programs written in all versions of the Java language up to and including Java SE 23`
this doesn't give the correct picture, does it? So, I am not for this change without the rest of the compiler changes.

akurtakov:
Enhancing DOM/AST layer is possible without compiler supporting it but it's up to you. I acted on request from https://github.com/eclipse-jdtls/eclipse-jdt-core-incubator/pull/276 .
Please close if you don't want to see this one in master.

mickaelistria:
> JLS23 API can be used to manipulate programs written in all versions of the Java language up to and including Java SE 23

Can we add a note to the doc to mention that this is provisional and incomplete; and we remove the note when releasing complete JLS23 support.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/AST.java
```diff
@@ -509,6 +524,11 @@ public final class AST {
 	 * deprecation warnings once JLS21 is deprecated
 	 */
 	static final int JLS22_INTERNAL = JLS22;
+	/**
+	 * Internal synonym for {@link #JLS22}. Use to alleviate
+	 * deprecation warnings once JLS21 is deprecated
```
mickaelistria, line 529: Should be `JLS23` and `JLS22` in the doc.

mickaelistria (commented):
Another approach to make progress here could be to contribute all the change excluding the `public static final int JLS23 = 23;`, so no API would be broken.

@jarthana what do you think it best: expose the public field with a "provisional" notice, or just not add it at all? In the end, I think just having the internal flag and the method is good enough to allow some progress on DOM side without waiting for compiler completion.

akurtakov:
If the change without the public API will be acceptable I'll do the necessary changes.

jarthana:
Historically, we have tagged such APIs with @noreference. Perhaps, we can consider that too here. I will wait to hear what others think.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/AST.java
akurtakov: Done

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/AST.java
```diff
@@ -509,6 +525,11 @@ public final class AST {
 	 * deprecation warnings once JLS21 is deprecated
 	 */
 	static final int JLS22_INTERNAL = JLS22;
+	/**
+	 * Internal synonym for {@link #JLS23}. Use to alleviate
+	 * deprecation warnings once JLS22 is deprecated
```
jarthana, line 530: This should also be JLS23.

akurtakov: Done. I fixed the JLS22_INTERNAL documentation too as it had the same issue.

jarthana (approved):

akurtakov:
Thanks, @jarthana !

PR #2326, opened 2024-04-11
Record pattern variable resolution inside switch, Fixes #2299

gayanper:
## What it does
Resolve record pattern variables which are declared as `var`

## How to test
Unit tests and sample code provided in #2299 

## Author checklist

- [x] I have thoroughly tested my changes
- [x] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [x] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

PR #2391, opened 2024-04-26
Record pattern variable resolution inside switch

### LINKED ISSUE #2299: RecordPattern variables are not resolved in Completions
Given the code with Java 21

```
public interface Named {}
public record Person(String name) implements Named {}

public foo(Named n) {
     switch(n) {
        case Person(var name) -> {
            n
        }
     }
} 
```

Completions after `n` inside the case statement show the local variable n as `Object` instead of string. But the compiler do resolve type which compiles fine.

stephan-herrmann:
Fixes #2299

+ approach by manual recovery of a switch statement
+ minimal fixes for elementStack handling of record patterns

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTestsForRecordPattern.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTestsForRecordPattern.java
@@ -19,10 +19,12 @@ import org.eclipse.jdt.core.JavaModelException;
 import junit.framework.Test;
 
 public class CompletionTestsForRecordPattern extends AbstractJavaModelCompletionTests {
+	private static int UNQUALIFIED_REL = R_DEFAULT + R_RESOLVED + R_CASE + R_INTERESTING + R_UNQUALIFIED
+			+ R_NON_RESTRICTED;
 
 
 	static {
-//		 TESTS_NAMES = new String[]{"test012"};
+		 //TESTS_NAMES = new String[]{"testGH2299_SwitchStatement"};
 	}
 
 	public CompletionTestsForRecordPattern(String name) {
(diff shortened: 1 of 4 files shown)


stephan-herrmann:
Some elements adopted from #2326:
* changes in CompletionNodeDetector
* use of `K_RECORD_PATTERN` to avoid interpreting "ID(" as the start of an invocation. I kept it for general correctness, although it may no longer be relevant for this issue

I removed from #2326:
* override of `requireExtendedRecovery()`
* recently introduced `K_ENHANCED_SWITCH`

New strategy:
* when the orphan assist node is passed to `buildMoreCompletionContext()`, detect if we are parsing a block nested in a switch statement or expression
* if so, manually collect the elements of the switch from ast and expression stacks and assemble the SwitchStatement  from them.
* then remove all `RecoveredElements` from `this.currentElement` which captured the parts of the switch, and replace those with a `RecoveredStatement` representing the switch.

This follows the example of an existing code block for `K_BETWEEN_CASE_AND_COLON`  and is intended to keep the complexity local to when we actually are in the specific situation. This block still needs more testing and likely some more checks. We should at least consider conditionally creating a `SwitchExpression` instead of `SwitchStatement`. Also the conditions of all those stack manipulations should be hardened.

If we decide that all this magic should only kick in within an enhanced switch, then re-introducing `K_ENHANCED_SWITCH` from #2326 should be our friend.

stephan-herrmann:
@gayanper would you like to prepare a few more tests so we can make sure the new code works in all relevant situations?

In a black box perspective we'd be looking for variations in these dimensions:
* where in the code is the switch located? (e.g., switch expression as a method argument)
* are there any case statements before the incomplete one?
* more complex structure of the case label
* what is the element being completed?
* is that element nested within some other structure?

Or from a white box perspective: different constellations of what we find on `astStack` and `expressionStack` plus the companion `astLengthStack` and `expressionLengthStack`. Also the structure in `Parser.currentElement` might have different shapes that need to be considered in the new code.

gayanper:
I can add some tests tomorrow, i think we can add some blackbox tests.

gayanper:
@stephan-herrmann I added some tests which some of them are failing

gayanper:
I just checked all these scenarios in both IntelliJ and NetBeans which both use javac parser based completions even though it might have difference, but both works without any glitches on all these tests scenarios :(

stephan-herrmann:
> @stephan-herrmann I added some tests which some of them are failing

Thanks, these were very helpful!

stephan-herrmann:
@srikanth-sankaran after some back and forth I think now is a good point in time for you to review this approach.

Just now I came across a code block in `buildInitialRecoveryState()` with this comment
```
		// Only relevant for Completion{Engine/ParSer}
		// Kludge, for the unfortunate case where recovery can't
		// construct a switch expression but simply creates a case statement (with type patterns)
		// and leaves it in the astStack
```

My current changes plough the same field, but they don't quite succeed to replace this previous approach. Still I have a feeling this should be possible if desired.

stephan-herrmann:
thanks @gayanper

Issue #2393, opened 2024-04-29
new jdt.internal.compiler.util.Jdk.<init> ()performance problem


jukzi:
After updating from ~0418 to I20240428-1800 (auto)rebuild of platform workspace took much longer then usual. I sampled the autobuild and see massive hotspots in JRTUtil .getJrtSystem() -> `Jdk<init>` that i have never seen before
![image](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/51790620/64f20181-3bbd-4aa3-b779-5bccaf7d0adb)

```
eclipse.buildId=4.32.0.I20240428-1800
java.version=17.0.10
java.vendor=Eclipse Adoptium
```

jukzi:
@HannesWell @iloveeclipse there where recent refactorings in that area:
f53ffcb035c4194af6a25ea11220f11cd2c56137
523d582e07d4ae1448f179fbbf5f42166da34343
from
https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2370

jukzi:
The use of  "path.toRealPath()" in JRTUtil is inaccaptable. It's an expensive IO call not suitable to compute a key on every call. Please fix or revert.

iloveeclipse:
> The use of "path.toRealPath()" in JRTUtil

@HannesWell : could you please remove this from `org.eclipse.jdt.internal.compiler.util.Jdk.Jdk(File)` constructor?

HannesWell:
> > The use of "path.toRealPath()" in JRTUtil
> 
> @HannesWell : could you please remove this from `org.eclipse.jdt.internal.compiler.util.Jdk.Jdk(File)` constructor?

Created https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2394 to use the much cheaper alternative `toAbsolutePath().normalize()`.

PR #2394, opened 2024-04-29
Avoid expensive calls to Path.toRealPath() in JRTUtil

HannesWell:
## What it does

Avoid calls to `Path.toRealPath()` in `JRTUtil` which are expensive, especially under Windows since each element of the path is checked for the real name of its element. Just calling `Path.toAbsolutePath().normalize()` already covers most cases.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2393

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
@@ -150,7 +150,7 @@ public class JRTUtil {
 	 */
 	public static FileSystem getJrtFileSystem(Path path) throws IOException {
 		try {
-			FileSystem fs = JRT_FILE_SYSTEMS.computeIfAbsent(path.toRealPath(), p -> {
+			FileSystem fs = JRT_FILE_SYSTEMS.computeIfAbsent(path.toAbsolutePath().normalize(), p -> {
 				try {
 					return FileSystems.newFileSystem(JRTUtil.JRT_URI, Map.of("java.home", p.toString())); //$NON-NLS-1$
 				} catch (IOException e) {
@@ -196,7 +196,7 @@ public class JRTUtil {
 	public static CtSym getCtSym(Path jdkHome) throws IOException {
 		CtSym ctSym;
 		try {
-			ctSym = ctSymFiles.compute(jdkHome.toRealPath(), (Path x, CtSym current) -> {
+			ctSym = ctSymFiles.compute(jdkHome.toAbsolutePath().normalize(), (Path x, CtSym current) -> {
 				if (current == null || !current.getFs().isOpen()) {
 					try {
 						return new CtSym(x);
@@ -440,7 +440,7 @@ class Jdk {
 	public Jdk(File jrt) throws IOException {
 		this.path = toJdkHome(jrt);
 		try {
-			this.release = pathToRelease.computeIfAbsent(this.path.toRealPath(), p -> {
+			this.release = pathToRelease.computeIfAbsent(this.path, p -> {
 				try {
 					return readJdkReleaseFile(p);
 				} catch (IOException e) {
@@ -477,7 +477,7 @@ class Jdk {
 
 	static Path toJdkHome(File jrt) {
 		Path home;
-		Path normalized = jrt.toPath().normalize();
+		Path normalized = jrt.toPath().toAbsolutePath().normalize();
 		if (jrt.getName().equals(JRTUtil.JRT_FS_JAR)) {
 			home = normalized.getParent().getParent();
 		} else {


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
```diff
@@ -477,7 +477,7 @@ boolean sameRelease(String other) {
 
 	static Path toJdkHome(File jrt) {
 		Path home;
-		Path normalized = jrt.toPath().normalize();
+		Path normalized = jrt.toPath().toAbsolutePath().normalize();
```
jukzi, line 480: When you need to introduce such Filesystem calls please provide measurement. Some of these Path method look harmless but system dependently do much more then they look like. May be its fast - may be not. It may even differ by the provided path and used filesystem. 
I am sure its faster then toRealPath but fast enough for this usecase? Does it really "operate without actually accessing the file-system."?
And why do you even introduce toAbsolutePath() here when it was not needed before?

HannesWell: > When you need to introduce such Filesystem calls please provide measurement. Some of these Path method look harmless but system dependently do much more then they look like. May be its fast - may be not. It may even differ by the provided path and used filesystem.

Sorry, but I can't provide such extensive testing. I only have a Windows computer available and not the time for a test-campaign.

> I am sure its faster then toRealPath but fast enough for this usecase? Does it really "operate without actually accessing the file-system."?

When looking at the `WindowsPath` implementation I can only see String operations. And when looking at the doc, while `Path.toRealPath()` requires the file to exist and may throw an `IOException` `toAbsolutePath()` does have this requirement and does not throw `IOExceptions`. At the same time the latter method is a NOP if the path is already absolute (which should be the case usually).
While I cannot guarantee that `toAbsolutePath()` does not access the file-system under all circumstances, I'm relatively confident that it does not at least in most cases.

For `Path.normalize()` the JavaDoc even mentions it explicitly:
`This method does not access the file system; the path may not locate a file that exists.`


> And why do you even introduce toAbsolutePath() here when it was not needed before?

In some places it was already used as you can see in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2371, some used `toAbsolutePath()` (respectively the File equivalent) and as you can see in this PR some used `normalized()` and because the created objects are retained for the entire lifetime of a JVM it actual duplicates should be avoided.

HannesWell:
@jukzi and @iloveeclipse what do you think about submitting this as it is and see how it behaves in the real world and if the assumptions are right or wrong? 
It should already improve things a lot. If it still isn't sufficient the resolution can still be reverted.

jukzi:
It is up to you to proof the effective benefit of all those changes. I don't see a reason why such measurement can not be done before submit. It's also a good practice for further contributions. If you can reproduce the issue, then measuring buildtime before/after all those changes should be enough.

iloveeclipse (approved):
Looks OK for me.

HannesWell:
Thank you @iloveeclipse.

@jukzi I have sampled my Eclipse SDK I use for platform+equinox+p2 development (which is my largest workspace) and on my Windows computer that's using a SSD the impact was not as sever as in your case:
grafik

The hard-drive is another factor that can impact the performance of IO-operations. Additionally the drives cache can also impact the performance, therefore an exhaustive benchmark would also have to distinguish between frequently and seldom touched files, which would make a definitive benchmark even more work.

But I'll sample my Eclipse SDK again after tonights I-build and post the difference here. It would also be interesting to see how the situation changed for you.

HannesWell:
> But I'll sample my Eclipse SDK again after tonights I-build and post the difference here. It would also be interesting to see how the situation changed for you.

With that I-Build that contains this change I didn't see the JDK constructor in the sampler logs at all and the calling `generateCode()` method also run much shorter (about 200ms now, but since there are multiple compiler-threads its hard to tell which threads exactly relate). So at least on my computer I would consider the performance regression fixed.

@jukzi I hope for you it's now also fast again?

jukzi:
It is faster now, but not as fast as it used to be. During building platform workspace it still takes a total of ~2 seconds in org.eclipse.jdt.internal.compiler.util.JRTUtil.getJrtSystem() on my computer. Almost all that time is spend in the Jdk constructor. The time is wasted across WindowsPath.hashCode(), .equals(), .normalize() and .getParent()
As far as i understand before it was just a plain String as key in Jdk.Path an now its a complex Path Object which is slow one such basic operations. 
It's not horrible slow anymore, but questionable that the changes where designed to be a performance improvement when they are in fact the opposite. The idea to cache something was good but changing Jdk.path from String to Path was bad. I don't see a reason for that. The normal case is that there are very few distinct Jdk.path while the Jdk constructor seems to be called  horrible often especially from org.eclipse.jdt.internal.core.builder.ClasspathJrt.isPackage(). Neither normalize() nor absolutepath() seem to be necessary., they are just contra productive when called that often.
So the key should either be changed back to String or it might also work to store the resulting JrtFileSystem in ClasspathJrt and hand that over to the JRTUtil methods, so that the constructor is just not called to so often.
image

HannesWell:
> So the key should either be changed back to String or it might also work to store the resulting JrtFileSystem in ClasspathJrt and hand that over to the JRTUtil methods, so that the constructor is just not called to so often.

I think referencing the JrtFileSystem in ClasspathJrt  would be the best way since it would save the repetitive map-look ups totally.
Unfortunately `JrtFileSystem` is package-private and can't be made public since it is another top-level class in `JRTUtil`. So it would need to become for example a nested class in `JRTUtil`.

> Neither normalize() nor absolutepath() seem to be necessary., they are just contra productive when called that often.

`toAbsolute()` is a no-op if the path is already absolute and normalize should also be quick if no `.` or `..` are contained. But since a big portion of the runtime is the computation of the hash, but because its value is cached within a (Windows)Path object, one way to recompute it would be to avoid re-creation of Path objects over and over again, i.e. to use more Path instead of String.

PR #2460, opened 2024-05-19
Capture test cases from hohwille

stephan-herrmann:
Tests extracted from https://github.com/m-m-m/util/issues/166

All of these tests pass with ecj and fail when invoked with
`-Drun.javac=enabled -Djdk.root=/home/java/jdk-22 -Dcompliance=22`
(insert your location of jdk-22, optionally select any other recent version).

I inserted tests into suite DubiousOutcomeTest, as the main purpose for now is to be informed when the outcome of any of these tests *change* during development, whereas the final call which outcome is correct/incorrect still needs to be made, since compilers do not agree.

### DIFF
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/AbstractRegressionTest.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/AbstractRegressionTest.java
@@ -837,6 +837,19 @@ protected static class JavacTestOptions {
 	Excuse excuseFor(JavacCompiler compiler) {
 		return null;
 	}
+	/** Difference were we're not sure which is correct, but want to be informed if either compiler changes. */
+	public static class DubiousOutcome extends Excuse {
+		DubiousOutcome(int mismatchType) {
+			super(mismatchType);
+		}
+		public static DubiousOutcome
+		EclipseErrorsJavacNone = RUN_JAVAC ?
+				new DubiousOutcome(MismatchType.EclipseErrorsJavacNone) : null,
+		JavacErrorsEclipseNone = RUN_JAVAC ?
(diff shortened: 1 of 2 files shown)


stephan-herrmann:
Initial findings for DubiousOutcomeTest.testJDK8319461

In https://bugs.openjdk.org/browse/JDK-8319461 Vicente reports:
> Now type inference creates this bound set:
V = V
V <: Object
P = P
P <: WritableProperty<V>
P <: ReadableProperty<V> 

and continues to conclude that rejecting is correct since P is not a subtype of WritableProperty.

Unfortunately he did not mark which types are type variables vs inference variables.

For the same example (possibly slightly different extract), ecj computes this bound set:
```
 	TypeBound  P#1 = P
	Dependency P#1 <: ReadableProperty<V#0>
	TypeBound  P#1 <: ReadableProperty<V>
	TypeBound  V#0 = V
	Dependency (? extends P#1)#2 <: P#1
	Dependency (? extends P#1)#2 <: ReadableProperty<V#0>
	TypeBound  (? extends P#1)#2 <: P
	TypeBound  (? extends P#1)#2 <: ReadableProperty<V>
	Dependency (? extends P#1)#2 <: WritableProperty<V#0>
	TypeBound  (? extends P#1)#2 <: WritableProperty<V>
```

Here the thing that should be a subtype of `WritableProperty` is an inference variable (substituting a wildcard), which we have no problem instantiating like this:
`(? extends P#1)#2	:	P & WritableProperty<V>`

Disclaimer: Debugging ecj I could observe some capture bounds, which we know are not handled during resolution as they should (because that part of the spec doesn't seem to "work"). This raises some initial doubt about ecj's outcome, but Vicente didn't base is argument on capture bounds in particular, so hopefully this is irrelevant here.

stephan-herrmann:
> In https://bugs.openjdk.org/browse/JDK-8319461 Vicente reports ...

That bug got re-opened linking back here :smile:

PR #2477, opened 2024-05-22
Avoid some catched ClassCastException

jukzi:
While hunting down an uncatched ClassCastException i found some ClassCastException that can be avoided by Pattern Matching

https://openjdk.org/jeps/394


stephan-herrmann (approved):
Thanks, this change is very welcome.
I kept wondering whether this strange pattern ever made sense back in 2001, but then I didn't bother enough.

jukzi:
ignoring instable build due to 6 API warnings like
`The API problem filter for: 'The constant value 21 of the field org.eclipse.jdt.core.dom.AST.JLS_Latest has been changed' is no longer used`

PR #2479, opened 2024-05-22
AbstractJavaModelTests performance: wakeup sleeping job

jukzi:
For example TypeResolveTests gets twice as fast: 8s -> 4s


jukzi:
image

PR #2480, opened 2024-05-22
performance: faster Java plugin shutdown

jukzi:
interrupts sleeping indexer

reproducable with ClasspathInitializerTests
14s->6s


jukzi:
image

Issue #2485, opened 2024-05-23
Empty stack error compiling project with broken classpath


iloveeclipse:
From one of many bugs reported I had a project in my workspace that was meant to be used to reproduce some specific error, however I wasn't able to properly setup the project (seem to need gradle etc), and left it "broken" & closed in my test workspace.

Now while trying to reproduce https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2484 I've opened this project again and got immediately this error:
```
eclipse.buildId=4.32.0.I20240522-1800
java.version=21
java.vendor=Oracle Corporation
BootLoader constants: OS=linux, ARCH=x86_64, WS=gtk, NL=en_US
Framework arguments:  -product org.eclipse.sdk.ide
Command-line arguments:  -product org.eclipse.sdk.ide -data /data/4x_platform_workspace/../runtime-Java_21 -dev file:/data/4x_platform_workspace/.metadata/.plugins/org.eclipse.pde.core/Java_21/dev.properties -os linux -ws gtk -arch x86_64 -consoleLog

org.eclipse.core.resources
Error
Thu May 23 09:35:42 CEST 2024
Problems occurred when invoking code from plug-in: "org.eclipse.core.resources".

java.util.EmptyStackException
	at java.base/java.util.Stack.peek(Stack.java:103)
	at java.base/java.util.Stack.pop(Stack.java:85)
	at org.eclipse.jdt.internal.compiler.codegen.OperandStack.pop(OperandStack.java:57)
	at org.eclipse.jdt.internal.compiler.codegen.CodeStream.areturn(CodeStream.java:418)
	at org.eclipse.jdt.internal.compiler.codegen.StackMapFrameCodeStream.areturn(StackMapFrameCodeStream.java:409)
	at org.eclipse.jdt.internal.compiler.codegen.CodeStream.generateReturnBytecode(CodeStream.java:2574)
	at org.eclipse.jdt.internal.compiler.ast.ReturnStatement.generateReturnBytecode(ReturnStatement.java:276)
	at org.eclipse.jdt.internal.compiler.ast.ReturnStatement.generateCode(ReturnStatement.java:260)
	at org.eclipse.jdt.internal.compiler.ast.AbstractMethodDeclaration.generateCode(AbstractMethodDeclaration.java:388)
	at org.eclipse.jdt.internal.compiler.ast.AbstractMethodDeclaration.generateCode(AbstractMethodDeclaration.java:324)
[...]

PR #2491, opened 2024-05-24
Empty stack error compiling project with broken classpath

srikanth-sankaran:
## What it does

Fix failure of operand stack management code in handling chained `goto`s

* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2485

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/codegen/BranchLabel.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/codegen/BranchLabel.java
@@ -28,7 +28,7 @@ public class BranchLabel extends Label {
 	protected int targetStackDepth = -1;
 	public final static int WIDE = 1;
 	public final static int USED = 2;
-	private OperandStack switchSaveTypeBindings;
+	private OperandStack operandStack;
 
 
 public BranchLabel() {
@@ -85,6 +85,29 @@ public void becomeDelegateFor(BranchLabel otherLabel) {
 	// other label is delegating to receiver from now on
 	otherLabel.delegate = this;
 
+	if (this.targetStackDepth == -1) {
+		if (this.codeStream.stackDepth < 0) {
+			this.codeStream.classFile.referenceBinding.scope.problemReporter()
+					.operandStackSizeInappropriate(this.codeStream.classFile.referenceBinding.scope.referenceContext);
+			this.codeStream.stackDepth = 0; // FWIW
+			this.codeStream.operandStack.clear();
+		}
+		this.targetStackDepth = this.codeStream.stackDepth;
+		this.operandStack = this.codeStream.operandStack.copy();
+		// TODO: check that operand stack contents slot count matches targetStackDepth
+	} else {
+		// Stack depth known at label having encountered a previous branch and/or having fallen through to label
+		if (this.targetStackDepth != this.codeStream.stackDepth) {
+			this.codeStream.classFile.referenceBinding.scope.problemReporter().operandStackSizeInappropriate(
+					this.codeStream.classFile.referenceBinding.scope.referenceContext);
+			if (this.targetStackDepth < this.codeStream.stackDepth) {
+				this.targetStackDepth = this.codeStream.stackDepth; // FWIW, pick the higher water mark.
+				this.operandStack = this.codeStream.operandStack.copy();
+			}
+		}
+		// TODO: check that operand stacks match.
+	}
+
 	// all existing forward refs to other label are inlined into current label
 	final int otherCount = otherLabel.forwardReferenceCount;
 	if (otherCount == 0) return;
@@ -136,7 +159,7 @@ protected void trackStackDepth(boolean branch) {
 				this.codeStream.operandStack.clear();
 			}
 			this.targetStackDepth = this.codeStream.stackDepth;
-			this.switchSaveTypeBindings = this.codeStream.operandStack.copy();
+			this.operandStack = this.codeStream.operandStack.copy();
 			// TODO: check that contents slot count matches targetStackDepth
 		} // else: previous instruction completes abruptly via goto/return/throw: Wait for a backward branch to be emitted.
 	} else {
@@ -145,13 +168,16 @@ protected void trackStackDepth(boolean branch) {
 			if (this.targetStackDepth != this.codeStream.stackDepth) {
 				this.codeStream.classFile.referenceBinding.scope.problemReporter().operandStackSizeInappropriate(
 						this.codeStream.classFile.referenceBinding.scope.referenceContext);
-				if (this.targetStackDepth < this.codeStream.stackDepth)
+				if (this.targetStackDepth < this.codeStream.stackDepth) {
 					this.targetStackDepth = this.codeStream.stackDepth; // FWIW, pick the higher water mark.
+					this.operandStack = this.codeStream.operandStack.copy();
+				}
 			}
-			// TODO: check that contents slot count matches targetStackDepth
+			// TODO: check that operand stacks match.
+		} else {
+			this.codeStream.stackDepth = this.targetStackDepth;
+			this.codeStream.operandStack = this.operandStack.copy();
 		}
-		this.codeStream.stackDepth = this.targetStackDepth;
-		this.codeStream.operandStack = this.switchSaveTypeBindings.copy();
 	}
 }
 /*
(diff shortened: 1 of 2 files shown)


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/codegen/BranchLabel.java
```diff
@@ -85,6 +85,29 @@ public void becomeDelegateFor(BranchLabel otherLabel) {
 	// other label is delegating to receiver from now on
 	otherLabel.delegate = this;
 
+	if (this.targetStackDepth == -1) {
```
mpalat, line 88: @srikanth-sankaran quick check - Should this be "< 0 " ?
[Essentially checking any scenario where this can be false]

srikanth-sankaran: Not really, it is checking for the initial value from line 28 above. If you check all other assignments to this field, they are all guaranteed to be 0 or +ve

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/codegen/BranchLabel.java
```diff
@@ -28,7 +28,7 @@ public class BranchLabel extends Label {
 	protected int targetStackDepth = -1;
 	public final static int WIDE = 1;
 	public final static int USED = 2;
-	private OperandStack switchSaveTypeBindings;
+	private OperandStack operandStack;
```
srikanth-sankaran, line 32: This is just a rename refactoring, to align the field name with what it is called elsewhere.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/codegen/BranchLabel.java
```diff
@@ -145,13 +168,16 @@ protected void trackStackDepth(boolean branch) {
 			if (this.targetStackDepth != this.codeStream.stackDepth) {
 				this.codeStream.classFile.referenceBinding.scope.problemReporter().operandStackSizeInappropriate(
 						this.codeStream.classFile.referenceBinding.scope.referenceContext);
-				if (this.targetStackDepth < this.codeStream.stackDepth)
+				if (this.targetStackDepth < this.codeStream.stackDepth) {
 					this.targetStackDepth = this.codeStream.stackDepth; // FWIW, pick the higher water mark.
+					this.operandStack = this.codeStream.operandStack.copy();
+				}
```
srikanth-sankaran, line 174: This corrects an oversight unrelated to this defect. There are two pieces of state associated with operand type stacks. A scalar the stackDepth - no of words on stack - and a vector - the stack - holding the type bindings themselves. Existing code merely checks the scalars match at branch targets and if not chooses the bigger - we are FUBAR anyways - but we try to handle it as best as we can. In this process, we should also update the vector in sync. That it was not is an oversight and addressed here

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/codegen/BranchLabel.java
```diff
@@ -145,13 +168,16 @@ protected void trackStackDepth(boolean branch) {
...
-			// TODO: check that contents slot count matches targetStackDepth
+			// TODO: check that operand stacks match.
+		} else {
+			this.codeStream.stackDepth = this.targetStackDepth;
+			this.codeStream.operandStack = this.operandStack.copy();
 		}
-		this.codeStream.stackDepth = this.targetStackDepth;
-		this.codeStream.operandStack = this.switchSaveTypeBindings.copy();
 	}
 }
```
srikanth-sankaran, line 182: The wrapping of the pre-existing statements 
```
                        this.codeStream.stackDepth = this.targetStackDepth;
			this.codeStream.operandStack = this.operandStack.copy();
```

is a fix unrelated to the defect. As the method [validOperandStack](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2442/files) shows, it is incorrect to expect that operand stack contents are identical along branch edges. They are only required to be compatible. (think of `cond ? new String() : new Date()` - as long as a LUB exists it is ok). So where `targetStackDepth` is known **AND** `sourceDepthKnown` is true, we should not copy over the operand stack. That could replace sharper values with blunter ones and vice versa.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/codegen/BranchLabel.java
```diff
@@ -85,6 +85,29 @@ public void becomeDelegateFor(BranchLabel otherLabel) {
...
+					this.codeStream.classFile.referenceBinding.scope.referenceContext);
+			if (this.targetStackDepth < this.codeStream.stackDepth) {
+				this.targetStackDepth = this.codeStream.stackDepth; // FWIW, pick the higher water mark.
+				this.operandStack = this.codeStream.operandStack.copy();
+			}
+		}
+		// TODO: check that operand stacks match.
+	}
+
 	// all existing forward refs to other label are inlined into current label
```
srikanth-sankaran, line 111: A `BranchLabel`s `targetStackDepth` get set under two circumstances. When a branch is issued on the label (`org.eclipse.jdt.internal.compiler.codegen.BranchLabel.branch()`) or when the `BranchLabel` is placed (`org.eclipse.jdt.internal.compiler.codegen.BranchLabel.place()`) such that the previous instruction fall through to the instruction with the label. 

In the defect scenario, `org.eclipse.jdt.internal.compiler.codegen.CodeStream.goto_(BranchLabel)` via `org.eclipse.jdt.internal.compiler.codegen.CodeStream.inlineForwardReferencesFromLabelsTargeting(BranchLabel, int)` finds out that it is a chained goto situation and arranges the target label to delegate to the one of the pre-existing labels and short circuits the emission of the `goto` - As the call to `org.eclipse.jdt.internal.compiler.codegen.BranchLabel.branch()` from `org.eclipse.jdt.internal.compiler.codegen.CodeStream.goto_(BranchLabel)` is skipped, we fail to the update the stack depth at the target of the goto. 

The fix is to update `org.eclipse.jdt.internal.compiler.codegen.BranchLabel.becomeDelegateFor(BranchLabel)` and propagate the target stack depth there

srikanth-sankaran:
@mpalat - this is ready for review. I have added detailed comments. I am happy to walk you through the changes.

mpalat (approved):
+1 Looks Good - Please merge
@srikanth-sankaran : Thanks for the details comments making it easier for review

srikanth-sankaran:
Consider this code below which is not the real code generated but for illustration purposes contains the extra `goto` at
bci 53:

```
  public static int convertOpcode(int, int);
    descriptor: (II)I
    flags: (0x0009) ACC_PUBLIC, ACC_STATIC
    Code:
      stack=2, locals=2, args_size=2
         0: iload_0
         1: tableswitch   { // 42 to 42

                      42: 20
                 default: 56
            }
        20: iload_1
        21: tableswitch   { // 42 to 42

                      42: 40
                 default: 45
            }
        40: bipush        42
        42: goto          53
        45: new           #16                 // class java/lang/UnsupportedOperationException
        48: dup
        49: invokespecial #18                 // Method java/lang/UnsupportedOperationException."<init>":()V
        52: athrow
        53: goto          64
        56: new           #16                 // class java/lang/UnsupportedOperationException
        59: dup
        60: invokespecial #18                 // Method java/lang/UnsupportedOperationException."<init>":()V
        63: athrow
        64: ireturn
```

I will compress this and show only the crucial instructions below:
```
// *** operand stack empty here ****
             40: bipush        42
             42: goto          L1
             ...
    L1:    53: goto          L2
            63: athrow
    L2:    64: ireturn
```
what does the operand stack contain at the start of bci 64 ? 

In this unoptimized sequence, it is easy to determine that stack depth at 64 is 1 word because there is an edge from bci 53 to bci 64 and at bci 53 the stack depth is 1 because there is an edge from bci 42 to bci 53

Now consider the `goto` chaining optimization which results in this code which elides the goto at bci 53 having marked that L1 delegates to L2.

```
// *** operand stack empty here ****
             40: bipush        42
             42: goto          L2      // **originally goto L1**
             ...
            60: athrow
    L2:    61: ireturn
```

In this scenario, we see only the label placement for L2 and never the branch to L2 since the instruction targetting it at bci 53 has been elided. (the `goto` at 42 mentions L2 but that is due to the delegation; at the time of goto emission, it was associated with the label L1.)

When we see the label placement without there ever being a direct branch, we don't know the real target depth there and bad things happen from there

Issue #2490, opened 2024-05-23
Support an alternative compiler (e.g., Javac) to build projects and generate class files


testforstephen:
From time to time, we observe discrepancies in compiler implementation between ECJ and Javac, particularly in areas such as generic type inference and annotation processing. Often, build tools like Maven or Gradle can compile the project successfully, but ECJ fails, causing compilation errors that block the run/debug behavior in the IDE.

To address this, we're experimenting with allowing the JDT to support an alternative compiler for building Java projects and generating class files. This new compiler would be used by JavaBuilder for project compilation only and would not impact basic language features like code completion and navigation, which would continue to use ECJ for the type system. We believe that offering an alternative compiler will benefit both the IDE and developers, allowing them to choose the compiler that best suits their needs.

To integrate a new compiler in JavaBuilder, we need to make minor refactorings to the AbstractImageBuilder, specifically to open up the API for initializing a different compiler. The implementation of the new compiler itself can be handled in a separate plugin.

We are opening this ticket to initiate the discussion. In the meantime, we will prepare a draft pull request to discuss the API design required to support a new compiler in JavaBuilder.

jukzi:
jdt is not only used to generate .class files but also for things like codecompletion, refactorings etc. it is an alternative to javac.
No one can stop you  from building a distinct builder, but you better don't expect any help from jdt team which has enough todo to keep ecj as java alternative alife.  
i think It would be even more frustrating for a user when the UI parses the java code different then the compiler. so it's better a replace both parsers or none.

see also
https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1984#issuecomment-2082236821

mickaelistria:
> No one can stop you from building a distinct builder

Indeed, and it's already done in a fork.

>  you better don't expect any help from jdt team which has enough todo to keep ecj as java alternative alife.

To be clear, the work is already done, and no-one is asking help nor anyone else to put extra manpower here as most of the work is already done in a fork.
We're more willing to evaluate whether JDT committers are open to the idea of 1. allowing to host some code to plug-in an alternative AbstractImageBuilder and 2. maybe host of javac-based implementation for it; so we can limit how our fork diverges from JDT and contribute in making JDT more powerful in the end.

I understand that current contributors can be busy, but we're not trying to make them any busier (beyond just saying "yes, we can welcome code that allows to use an alternative AbstractImageBuilder"); so this argument doesn't really stand here.
[...]

PR #2497, opened 2024-05-27
Add the option to use an alternative compiler for compiling classes in JavaBuilder

testforstephen:
This is a follow up to #2490

What does in this PR:
- Introduce a system property (e.g., `-DAbstractImageBuilder.compilerFactory=x.y.z.MyCompilerFactory`) to specify the compiler factory that produces the compiler for compiling classes in AbstractImageBuilder. Refer to MockCompiler in the test case org.eclipse.jdt.core.tests.builder.BasicBuildTests.MockCompiler for an example of a custom compiler implementation, and see [JavacCompiler](https://github.com/eclipse-jdtls/eclipse-jdt-core-incubator/blob/dom-with-javac/org.eclipse.jdt.core.javac/src/org/eclipse/jdt/internal/javac/JavacCompiler.java) for a real Javac implementation.
- Aggregate compiler settings (such as source path, classpath, output, annotation processors, and other compiler options) into a CompilerConfiguration model for reuse by the custom compiler.
  - Regarding the annotation processing workflow in Eclipse, project importers like m2e convert the annotation processing configuration from the build file into a .factorypath file, which the APT plugin uses for annotation processing. When switching to an alternative compiler like Javac, which can handle annotation processing directly, we don't need to invoke the APT plugin during compilation but still need it to resolve the factorypath configuration.


org.eclipse.jdt.core.tests.builder/META-INF/MANIFEST.MF
```diff
@@ -5,6 +5,7 @@ Bundle-SymbolicName: org.eclipse.jdt.core.tests.builder; singleton:=true
 Bundle-Version: 3.12.400.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
+Fragment-Host: org.eclipse.jdt.core
```
testforstephen, line 8: This ensures the tests shares the same classloader as jdt.core plugin, and the AbstractImageBuilder in jdt.core can create an instance of the MockCompiler in the tests.

org.eclipse.jdt.core/model/org/eclipse/jdt/core/compiler/CompilationParticipant.java
```diff
@@ -152,6 +152,31 @@ public void processAnnotations(BuildContext[] files) {
...
 }
 
+/**
+ * Returns where to find annotation processors. If this is not specified,
+ * then the class path is searched for processors.
+ *
+ * @param project the project to participate in
+ * @param isTest whether the annotation processor path is for test code
+ * @return the annotation processor paths
+ * @since 3.38
```
jukzi, line 162: Would be better if such API changes could be avoided, given that is more like an experiment

jukzi (changes requested):
In principle the change looks simple enough and purely technical. Please also add a Junit test for the case where there is a problem with a custom supplied builder. Currently that seems to be possibly inconsistent (fall back vs hardcoded change in annotation processing)

org.eclipse.jdt.core/model/org/eclipse/jdt/core/compiler/CompilationParticipant.java
testforstephen: Annotation processing configurations, such as the processor path and output directory, are part of the standard compiler arguments. It is better to explicitly provide these configurations within the host builder (for instance, AbstractImageBuilder) when invoking the custom compiler.

Currently, jdt.core utilizes the CompilationParticipant mechanism to interact with the APT plugin for annotation processing tasks. Extending the CompilationParticipant class appears to be a natural choice for obtaining APT configurations. Any other ideas we could consider for accessing APT information?

jukzi: > Any other ideas we could consider for accessing APT information?

Does your client code need to call or override that API or can jdt just typecast to its internal implementation?

testforstephen: > Does your client code need to call or override that API

My client code does not use this API. It's AbstractImageBuilder in jdt.core that calls the API to get the related information and pass the result to the client code.

> can jdt just typecast to its internal implementation?

jdt.core doesn't depend on the apt plugin directly. It uses extension point to discover AptCompilationParticipant instance. Not sure how to use typecast.

jukzi: Let a new intermediate jdt internal class 
AbstractAptCompilationParticipant extends CompilationParticipant, let 
AptCompilationParticipant extends AbstractAptCompilationParticipant (jdt apt should be a friend of jdt), 
downcast CompilationParticipant to AbstractAptCompilationParticipant if the participiant is an instance of that.  
That's a little detour but avoids otherwise unneeded public API.
 -- Or mark the new API as Provisional so that it could be removed again without further warning.

testforstephen: Good point. thanks for suggestion.

org.eclipse.jdt.core/model/org/eclipse/jdt/core/compiler/CompilationParticipant.java
```diff
@@ -152,6 +152,37 @@ public void processAnnotations(BuildContext[] files) {
 	// do nothing by default
 }
 
+/**
+ * Provisional API for experimental support for an alternative compiler
+ * and can be removed without notice.
```
stephan-herrmann, line 157: this looks like a case for `@noreference`

testforstephen: nice. @noreference tag added. thanks.

org.eclipse.jdt.core/model/org/eclipse/jdt/core/compiler/CompilerConfiguration.java
```diff
@@ -0,0 +1,100 @@
...
+import java.io.File;
+import java.util.List;
+import java.util.Map;
+
+import org.eclipse.jdt.internal.compiler.impl.CompilerOptions;
+
+/**
+ * @since 3.38
+ */
+public class CompilerConfiguration {
```
jarthana, line 25: What about this? Are we exposing this to the outside world too? If not, we need @noreference here as well. Otherwise, this could use some documentation.

testforstephen: This is not only used by internal AbstractImageBuilder. It will be also used by clients for retrieving the compiler options.

mickaelistria: But aren't all the expected clients basically implementing AbstractImageBuilder (so already consuming this internal package) ?
While I think all this will need to be API at some point, as it's early stage, I think we'd rather keep as much as possible in "internal" packages, until we feel comfortable enough with the design to turn it API.

testforstephen: The clients don't implement AbstractImageBuilder, but the ECJ internal Compiler class. Yes, we can put the new CompilerConfiguration within internal package since it remains accessible from other plugin.

Just curious what's guideline to split the public package and internal package in the JDT codebase?

mickaelistria: Ah yes, my bad!
The public package become API that we can then hardly modify and that we have to maintain as such. internal packages are basically not supported externally and can be changed more freely. So in general, unless there is a good reason for something to become API, we try to keep things internal as it makes further maintenance significantly easier.
But here, it can be a case where a new API class makes sense. I think it's actually good to have an actual object formalizing portable standard compiler configuration that could be used in multiple clients, rather than passing some options and (potentially unknown) participants.

testforstephen: Exactly. Currently, project compiler-related settings are scattered across multiple places. Introducing a centralized class to formalize them would make them easily portable for other clients to reuse.

mickaelistria:
@testforstephen can this one be rebased and updated and turned from draft to PR already?

testforstephen:
> @testforstephen can this one be rebased and updated and turned from draft to PR already?

Done.

Here’s the CI failure message. This CI error also occurs for other pull requests such as https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2519.  It seems to be a general environment issue when we start a new version (e.g., 4.33).

```
10:44:43  [ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.8:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.jdt.core.compiler.batch: Baseline and reactor have the same fully qualified version, but different content
10:44:43  [ERROR] different
10:44:43  [ERROR]    META-INF/ECLIPSE_.RSA: not present in baseline
10:44:43  [ERROR]    META-INF/ECLIPSE_.SF: not present in baseline
```

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java
```diff
@@ -573,12 +576,39 @@ protected Compiler newCompiler() {
...
+				IErrorHandlingPolicy.class,
+				CompilerConfiguration.class,
+				ICompilerRequestor.class,
+				IProblemFactory.class);
+			newCompiler = constructor.newInstance(this.nameEnvironment,
+				DefaultErrorHandlingPolicies.proceedWithAllProblems(),
+				prepareCompilerConfiguration(compilerOptions),
+				this,
+				ProblemFactory.getProblemFactory(Locale.getDefault()));
+			this.useDefaultCompiler = false;
```
jukzi, line 595: It's not good when the method newCompiler() changes the state of the Class, as successive calls could alter the behaviour. It would be better a flag Compiler.isDefaultCompiler. But ideally there would not be such flag at all but participant.isAnnotationProcessor() or participant.processAnnotations() would just behave different.

org.eclipse.jdt.core/model/org/eclipse/jdt/core/compiler/CompilerConfiguration.java
```diff
@@ -0,0 +1,169 @@
...
+ *
+ * @since 3.38
+ * @noextend This class is not intended to be subclassed by clients.
+ */
+public class CompilerConfiguration {
+	List<String> sourcepaths;
+	List<String> moduleSourcepaths;
+	List<String> classpaths;
+	List<String> modulepaths;
+	Map<File, File> sourceOutputMapping;
```
jukzi, line 40: I am worried, that this otherwise unneeded API 

- became that big,
- is not tested at all, 
- the implementing members are not private, 
- it is not threadsafe. 
- publishes internal Classes

Feels like it should be an immutable record.

The IDE even warns about it:
image

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java
```diff
@@ -573,12 +576,39 @@ protected Compiler newCompiler() {
...
-		DefaultErrorHandlingPolicies.proceedWithAllProblems(),
-		compilerOptions,
-		this,
-		ProblemFactory.getProblemFactory(Locale.getDefault()));
+	Compiler newCompiler = null;
+	String compilerClassName = System.getProperty(AbstractImageBuilder.class.getSimpleName() + ".compiler"); //$NON-NLS-1$
+	if (compilerClassName != null) {
+		try {
+			Class<? extends Compiler> compilerClass = (Class<? extends Compiler>) Class.forName(compilerClassName);
+			Constructor<? extends Compiler> constructor = compilerClass.getDeclaredConstructor(
```
jukzi, line 584: It's not obvious how the constructor of an alternative Compiler has to look like. Better use the Property to instantiate a parameterless compiler Factory that implements a well defined interface to produce a Compiler. At best JDT comes with its own default CompilerFactory and uses the same parameters instead of having a different constructor

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java
```diff
@@ -573,12 +576,39 @@ protected Compiler newCompiler() {
...
 	compilerOptions.performMethodsFullRecovery = true;
 	compilerOptions.performStatementsRecovery = true;
-	Compiler newCompiler = new Compiler(
-		this.nameEnvironment,
-		DefaultErrorHandlingPolicies.proceedWithAllProblems(),
-		compilerOptions,
-		this,
-		ProblemFactory.getProblemFactory(Locale.getDefault()));
+	Compiler newCompiler = null;
+	String compilerClassName = System.getProperty(AbstractImageBuilder.class.getSimpleName() + ".compiler"); //$NON-NLS-1$
```
jukzi, line 580: Please define a String constant for `AbstractImageBuilder.class.getSimpleName() + ".compiler"` and use it in the junit tests as well.

org.eclipse.jdt.core.tests.builder/src/org/eclipse/jdt/core/tests/builder/BasicBuildTests.java
```diff
@@ -687,4 +696,103 @@ public void testBug549942() throws JavaModelException {
...
+			expectingProblemsFor(
+					path,
+					"Problem : Compilation error from MockCompiler [ resource : </Project/src/p1/Hello.java> range : <0,1> category : <60> severity : <2>]"
+				);
+		} finally {
+			System.clearProperty(CUSTOM_COMPILER_KEY);
+		}
+	}
+
+	public void testFallbackForProblematicCompiler() throws JavaModelException {
```
jukzi, line 734: that test should utilize annotation processing

org.eclipse.jdt.core/model/org/eclipse/jdt/core/compiler/CompilerConfiguration.java
```diff
@@ -0,0 +1,169 @@
...
+ * arguments.
+ *
+ * Clients typically use this class when opting for an alternative compiler
+ * like Javac to compile Java files.
+ *
+ * @since 3.38
+ * @noextend This class is not intended to be subclassed by clients.
+ */
+public class CompilerConfiguration {
+	List<String> sourcepaths;
```
jukzi, line 36: It is not obvious that all those Strings refer to a file path. Better use a higher level Objects like org.eclipse.core.resources.IContainer for directory paths.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java
```diff
@@ -596,6 +626,72 @@ protected Compiler newCompiler() {
...
+		}
+	}
+	configuration.setAnnotationProcessorPaths(annotationProcessorPaths);
+	configuration.setGeneratedSourcePaths(generatedSourcePaths);
+	ClasspathLocation[] classpathLocations = this.nameEnvironment.binaryLocations;
+	List<String> classpaths = new ArrayList<>();
+	List<String> modulepaths = new ArrayList<>();
+	for (ClasspathLocation location : classpathLocations) {
+		if (location instanceof ClasspathDirectory cpDirectory) {
+			String filepath = cpDirectory.binaryFolder.getLocation().toFile().getAbsolutePath();
```
jukzi, line 656: possible NPE: .getLocation() is documented that it may return null. `if path is null, it means path variable refers to a non-local filesystem`

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java
mickaelistria: +1, introducing an explicitly "ICompilerFactory" with a `newCompiler(...)` method can make things much clearer here.

testforstephen: good suggestion.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java
testforstephen: useDefaultCompiler is removed.

org.eclipse.jdt.core/model/org/eclipse/jdt/core/compiler/CompilerConfiguration.java
testforstephen: Yes, it's cleaner to use record for this API. Also add unit test to cover the new API.

org.eclipse.jdt.core/model/org/eclipse/jdt/core/compiler/CompilerConfiguration.java
testforstephen: I don't see it as a strong reason to use IContainer. I will keep using the file path.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/compiler/CompilerConfiguration.java
```diff
@@ -0,0 +1,67 @@
...
+ * used to compile Java files. It provides methods to set and retrieve
+ * various compiler options, including source paths, class paths,
+ * output directories, annotation processing options, and other compiler
+ * arguments.
+ *
+ * Clients typically use this class when opting for an alternative compiler
+ * like Javac to compile Java files.
+ *
+ * @since 3.38
+ * @noextend This class is not intended to be subclassed by clients.
```
jukzi, line 31: `@noextend` can be enforced by declaring the record final instead. 
However it now became clear that this record is only used as argument to newCompiler(). Isn't it easier to understand the code if you just add all those members as separate argument to newCompiler() ?

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java
```diff
@@ -596,6 +620,93 @@ protected Compiler newCompiler() {
...
+			}
+		}
+
+		ClasspathLocation[] classpathLocations = this.nameEnvironment.binaryLocations;
+		Set<String> classpaths = new LinkedHashSet<>();
+		Set<String> modulepaths = new LinkedHashSet<>();
+		for (ClasspathLocation location : classpathLocations) {
+			if (location instanceof ClasspathDirectory cpDirectory) {
+				IPath binaryLocation = cpDirectory.binaryFolder.getLocation();
+				if (binaryLocation == null) {
```
jukzi, line 649: Such null check does not allow compilers to be compliant to handle external files. Better just collect the IFolders instead of string path such that external folders are not lost.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java
```diff
@@ -596,6 +620,93 @@ protected Compiler newCompiler() {
...
+			String absoluteSourcepath = sourceFolder.getAbsolutePath();
+			if (sourceLocation.isOnModulePath) {
+				moduleSourcepaths.add(absoluteSourcepath);
+			} else {
+				sourcepaths.add(absoluteSourcepath);
+			}
+		}
+
+		return new CompilerConfiguration(
+				new ArrayList<>(sourcepaths),
```
jukzi, line 695: nitpick: better hand over immutable copys: List.copyOf(sourcepaths)

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java
```diff
@@ -573,12 +575,34 @@ protected Compiler newCompiler() {
...
+	ICompilerFactory compilerFactory = null;
+	String compilerFactoryClassName = System.getProperty(COMPILER_FACTORY_KEY);
+	if (compilerFactoryClassName != null) {
+		try {
+			Class<? extends ICompilerFactory> compilerFactoryClass = (Class<? extends ICompilerFactory>) Class.forName(compilerFactoryClassName);
+			Constructor<? extends ICompilerFactory> constructor = compilerFactoryClass.getDeclaredConstructor();
+			compilerFactory = constructor.newInstance();
+		} catch (ClassNotFoundException e) {
+			ILog.get().error("Could not load class " + compilerFactoryClassName, e); //$NON-NLS-1$
+		} catch (NoSuchMethodException e) {
```
jukzi, line 588: Here and below the underlying Exception e is lost.  Better just merge all catch clauses to a single? The action is always the same: just log the exception

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java
```diff
@@ -596,6 +620,93 @@ protected Compiler newCompiler() {
...
+		return new CompilerConfiguration(
+				new ArrayList<>(sourcepaths),
+				new ArrayList<>(moduleSourcepaths),
+				new ArrayList<>(classpaths),
+				new ArrayList<>(modulepaths),
+				annotationProcessorPaths,
+				generatedSourcePaths,
+				sourceOutputMapping,
+				options.getMap());
+	} catch (Exception e) {
```
jukzi, line 703: the exception stacktrace is lost here

jukzi (commented):
overall it looks like its going in the right direction. Please fix the faiiing tests

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java
```diff
@@ -666,9 +777,10 @@ protected void processAnnotations(CompilationParticipantResult[] results) {
 		results[i].reset(foundAnnotations ? this.filesWithAnnotations.get(results[i].sourceFile) : null);
 	}
 
+	boolean isEcjUsed = Compiler.class.equals(this.compiler.getClass());
 	// even if no files have annotations, must still tell every annotation processor in case the file used to have them
 	for (CompilationParticipant participant : this.javaBuilder.participants)
-		if (participant.isAnnotationProcessor())
+		if (isEcjUsed && participant.isAnnotationProcessor())
```
fbricon, line 783: I believe the check for isEcjUsed  could be moved up prior to looping through the participants

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/compiler/CompilerConfiguration.java
```diff
@@ -0,0 +1,67 @@
...
+		 */
+		List<IContainer> sourcepaths,
+		/**
+		 * List of file paths where the compiler can find source files for modules.
+		 */
+		List<IContainer> moduleSourcepaths,
+		/**
+		 * List of file paths where the compiler can find user class files and annotation processors.
+		 */
+		List<Either<IContainer, String>> classpaths,
```
jukzi, line 45: I dislike the "Either" class. But i have not solution in hand :-(. I guess this is caused by the complicated definition of `org.eclipse.jdt.core.IClasspathEntry.getPath()` ? Can you instead pass a List of IClasspathEntry which is complicated but well defined?

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java
```diff
@@ -666,10 +756,13 @@ protected void processAnnotations(CompilationParticipantResult[] results) {
 		results[i].reset(foundAnnotations ? this.filesWithAnnotations.get(results[i].sourceFile) : null);
 	}
 
-	// even if no files have annotations, must still tell every annotation processor in case the file used to have them
-	for (CompilationParticipant participant : this.javaBuilder.participants)
-		if (participant.isAnnotationProcessor())
-			participant.processAnnotations(results);
+	boolean isEcjUsed = Compiler.class.equals(this.compiler.getClass());
```
jukzi, line 759: Since this codeblock depends on compiler you can move the code to Compiler (for example Compiler#processAnnotations()) and other compilers can overload it with a NOOP.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/compiler/CompilerConfiguration.java
testforstephen: They are inferred from NameEnvironment of JavaBuilder, where it defines multi kinds of o.e.j.i.c.builder.ClasspathLocation . The problem for ClasspathLocation is that its property modifier are package visible.

testforstephen: They are inferred from NameEnvironment of JavaBuilder, where it defines multi kinds of o.e.j.i.c.builder.ClasspathLocation . The problem for ClasspathLocation is that its property modifier are package visible.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java
testforstephen: The Compiler API is defined in a separate ECJ plugin org.eclipse.jdt.core.compiler.batch. I want to maintain a narrow focus for this pull request, and prefer not to modify too much bundles. Also ECJ has no concept of annotation processing, it's not good to move it to ECJ compiler API.

testforstephen:
Currently, I am reusing the `o.e.j.c.tests.builder` plugin to test the Custom Compiler option API. To do this, I need to make the tests.builder bundle a fragment of the host plugin `org.eclipse.jdt.core`. This allows the newCompiler API in AbstractImageBuilder class to access the MockCompilerFactory test class. However, this approach prevents the tests.builder bundle from exporting packages to other bundles (such as `o.e.j.c.tests.model`) for reuse, resulting in a CI failure with the error: `Unresolved requirement: Require-Bundle: org.eclipse.jdt.core.tests.builder`.

@mickaelistria @jukzi Do you know if it's possible to export packages from a fragment? If not, should I avoid reusing the existing tests.builder and instead create a separate tests bundle to test the new Custom Compiler option API?

mickaelistria:
> I need to make the tests.builder bundle a fragment of the host plugin

This can have undesired implications, some bundles may have a `Require-Bundle: org.eclipse.jdt.tests.builder` which would be broken.
Would it be possible to just create a new test fragment that defines the MockCompilerFactory and have the tests using an Import-Package to make sure this fragment is present when executing the tests?

> Do you know if it's possible to export packages from a fragment? 

Yes, it should be possible IIRC.

jukzi:
> @mickaelistria @jukzi Do you know if it's possible to export packages from a fragment?

I can't help with that. i only know two patterns we use in tests: either a custom extensions defined in the plugin.xml or provide a setter that is explicitly called from tests like .setCompilerFactory(TestFactory.class) should work

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java
jukzi: how about ICompilerFactory.processAnnotations(...)?

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/compiler/CompilerConfiguration.java
jukzi: How about using URI?
Resource.getLocationURI() / java.io.File.toURI()

testforstephen: URI is more general than path, and it's good for me to use URI instead.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java
testforstephen: It would move complex logic, such as CompilationParticipant, into the ICompilerFactory, which seems overly designed. Instead, I'm thinking if we can introduce a flag such as ICompilerFactory.canCompilerSupportAnnotationProcessing to control whether the AbstractImageBuilder should skip annotation processing.

mickaelistria: I think the current code is clearer than the proposals: in practice annotation processors as CompilationParticipants are only necessary for ECJ, and don't make sense for other known implementation. The current code makes it really clear and easy to understand. Adding some indirection will make things more complicated to grok.
So IMO, we should stick to `isECJUsed`, or maybe just

```java
if (this.compiler instanceof Compiler) { // annotation processors as compilation participants are only relevant to ECJ Compiler
	// even if no files have annotations, must still tell every annotation processor in case the file used to have them
	for (CompilationParticipant participant : this.javaBuilder.participants)
		if (participant.isAnnotationProcessor())
			participant.processAnnotations(results);
}
```

Let's keep in mind that the code here is internal and we'd like to keep things simple for the moment. Maybe a 3rd implementation will allow to sort things out better now, but with only 2 implementations, we won't be able to ensure we make the perfect final choice anyway.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/compiler/CompilerConfiguration.java
```diff
@@ -0,0 +1,69 @@
...
+		List<IContainer> generatedSourcePaths,
+		/**
+		 * The mapping of source files to output directories.
+		 */
+		Map<IContainer, IContainer> sourceOutputMapping,
+		/**
+		 * The compiler options used to control the compilation behavior.
+		 * See {@link org.eclipse.jdt.internal.compiler.impl.CompilerOptions} for a list of available options.
+		 */
+		CompilerOptions compilerOptions) {
```
testforstephen, line 68: Have to use CompilerOptions instead of Map, as converting to Map results in the loss of some options (such as CompilerOptions.OPTION_JdtDebugCompileMode). This is the reason why [some tests](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-2497/19/testReport/org.eclipse.jdt.core.tests.model/ModuleBuilderTests/test_no_conflicting_packages_for_debugger_global/) failed.

mickaelistria: That's IMO a good thing. CompilerOptions is a very good object.

testforstephen:
@jukzi all comments are addressed and CI has passed as well, do you think it's OK to merge now?

jukzi (approved):
Looks acceptable for me. Maybe someone else like @jarthana would like to take a second look?

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java
```diff
@@ -596,6 +618,79 @@ protected Compiler newCompiler() {
...
+				if (cpURI == null) {
+					continue;
+				}
+				if (cpDirectory.isOnModulePath) {
+					modulepaths.add(cpURI);
+				} else {
+					classpaths.add(cpURI);
+				}
+			} else if (location instanceof ClasspathJar cpJar) {
+				URI cpURI = new File(cpJar.zipFilename).toURI();
```
jukzi, line 656: just one more thing: please avoid File.toURI but use implementation like org.eclipse.jdt.internal.core.search.indexing.BinaryIndexer.toUri(String) to avoid FileSystem access  - it is known that this file is a file and not a directory

testforstephen: Ok, I will reuse the utility `org.eclipse.core.filesystem.URIUtil.toURI(filePath, true)`, which requires no IO operation.

jukzi: Cool, i did not know that method.

jukzi:
I would like to see this PR submitted by tomorrow. Objections anybody? @mpalat?
@testforstephen last formal thing todo: please squash commits to a single commit.

jarthana (approved):

testforstephen:
> last formal thing todo: please squash commits to a single commit.

We can leverage `Squash and merge` button of GitHub PR directly, which will merge all commits to a single one.

jukzi:
> We can leverage `Squash and merge` button of GitHub PR directly, which will merge all commits to a single one.

We could but the jdt.core committers prefer to squash before merge. Then the commits on github will be the same as in the repository. Please do us the favor.

jukzi:
> I'll try to get to it within a week's time.

@stephan-herrmann Oh, i have only read that by now. I will wait for your review. Thanks for the indication.

testforstephen:
> > We can leverage `Squash and merge` button of GitHub PR directly, which will merge all commits to a single one.
> 
> We could but the jdt.core committers prefer to squash before merge. Then the commits on github will be the same as in the repository. Please do us the favor.

Just my two cents: as a committer of jdt.core as well, I might have a different perspective. I prefer to keep the PR commit history as it is because the back-and-forth commits provide valuable learning material for other contributors. It helps them understand the types of comments a committer might make. Personally, I learned about version bumping things from the sub-commits of other PRs.

jukzi:
> committer of jdt.core as well

Hey cool, i did not know. Then you do not need me to press submit  :-)

jarthana:
> Just my two cents: as a committer of jdt.core as well, I might have a different perspective. I prefer to keep the PR commit history as it is because the back-and-forth commits provide valuable learning material for other contributors. It helps them understand the types of comments a committer might make. Personally, I learned about version bumping things from the sub-commits of other PRs.

Absolutely, especially, once a reviewer or more get involved. It's probably OK to squash before bringing something up for a review, but after that, it's best not to force push or squash. Of course, we should squash when merging.

stephan-herrmann:
Before I go into details,  one general question: will this option be advertised for a general audience, or is it only a hook for experimentation within the greater JDT umbrella?

Reason for asking: all this heavily depends on a great number of internal classes and interfaces, so at the moment there would not be a legal way for a third-party plug-in to leverage the option, right?

BTW, in Eclipse I see many "discouraged" warnings against org.eclipse.jdt.core.tests.builder.mockcompiler. A few things look inconsistent here:
* it seems maven ignores these due to reference to parent `tests-pom/pom.xml`
* the eclipse project does not declare its sources as "test"
* in the end the mockcompiler shouldn't even be considered as test source, because that's what eventually 3rd party would provide, right?
One way or other, those warnings should be avoided also in the IDE.

stephan-herrmann (changes requested):
See detailed comments, of which I consider the following as mandatory:
* A statement about whether/how this will be advertised
* A statement about supporting incremental builds
* A warning free workspace in Eclipse
* Error-free execution of `RunBuilderTests`.

Full test coverage of method `prepareCompilerConfiguration()` would be good to have / should be established before the next release.

Others comments are less urgent.

org.eclipse.jdt.core.tests.builder/src/org/eclipse/jdt/core/tests/builder/BasicBuildTests.java
```diff
@@ -687,4 +714,210 @@ public void testBug549942() throws JavaModelException {
...
+					package p1;
+					public class Hello {
+					   public static void main(String args[]) {
+					      System.out.println(\"Hello world\");
+					   }
+					}
+					"""
+					);
+
+			fullBuild(projectPath);
```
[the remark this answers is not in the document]
testforstephen: In the experiment repo, we have an issue to track the incremental build. https://github.com/eclipse-jdtls/eclipse-jdt-core-incubator/issues/443

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java
```diff
@@ -573,12 +577,31 @@ protected Compiler newCompiler() {
...
+	}
+
+	if (compilerFactory == null) {
+		compilerFactory = new DefaultCompilerFactory();
+	}
+
+	Compiler newCompiler = compilerFactory.newCompiler(
+			this.nameEnvironment,
+			DefaultErrorHandlingPolicies.proceedWithAllProblems(),
+			prepareCompilerConfiguration(compilerOptions),
```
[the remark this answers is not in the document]
testforstephen: Since the default ECJ implementation only relies on the `compilerOptions` property from the `CompilerConfiguration`, and `prepareCompilerConfiguration()` retains the original compilerOptions when creating the CompilerConfiguration. All other properties computation in it is for the experimental flag. I believe it will not impact the default implementation.

org.eclipse.jdt.apt.core/src/org/eclipse/jdt/apt/core/internal/AptCompilationParticipant.java
```diff
@@ -208,6 +215,45 @@ public void processAnnotations(BuildContext[] allfiles) {
 		}
 	}
 
+	@Override
+	public URI[] getAnnotationProcessorPaths(IJavaProject project, boolean isTest) {
```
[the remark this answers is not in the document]
testforstephen: The motivation is that the AbstractImageBuilder contains two compilation groups (MAIN and TEST), which invoke the CompilationParticipant separately. In the current experiment stage, it doesn't use this flag because I haven't encountered a use case that requires differentiating annotation processors for the test scope.

org.eclipse.jdt.core.tests.builder/src/org/eclipse/jdt/core/tests/builder/BasicBuildTests.java
```diff
@@ -687,4 +714,210 @@ public void testBug549942() throws JavaModelException {
...
+			if (compiler instanceof MockCompiler mockCompiler) {
+				configs.add(mockCompiler.compilerConfig);
+			}
+		};
+		File projectRoot = null;
+		IProject project = null;
+		try {
+			MockCompilerFactory.addListener(listener);
+			System.setProperty(AbstractImageBuilder.COMPILER_FACTORY_KEY, CUSTOM_COMPILER_VALUE);
+			projectRoot = copyFiles("autoValueSnippet", true);
```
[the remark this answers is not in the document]
testforstephen: nice catch. fixed.

testforstephen:
> See detailed comments, of which I consider the following as mandatory:
> 
> * A statement about whether/how this will be advertised

This is currently for experimental Javac support and not intended for a general audience. That's why we retain the internal class references and minimize intrusion into the existing mechanism.

> * A statement about supporting incremental builds

Already replied in the related comment.

> * A warning free workspace in Eclipse

The warning for require-bundle is fixed.
> in Eclipse I see many "discouraged" warnings against org.eclipse.jdt.core.tests.builder.mockcompiler.

No idea for this. I don't see the warnings for them in the Eclipse IDE.

> * Error-free execution of `RunBuilderTests`.

Fixed.

org.eclipse.jdt.apt.core/src/org/eclipse/jdt/apt/core/internal/AptCompilationParticipant.java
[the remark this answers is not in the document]
stephan-herrmann: I don't see any code in JDT that supports separate annotationProcessorPaths for MAIN vs TEST, only separate folders for generated sources. Am I missing anything?

Anyway this can be cleaned up later.

org.eclipse.jdt.core.tests.builder/src/org/eclipse/jdt/core/tests/builder/BasicBuildTests.java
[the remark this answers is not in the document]
stephan-herrmann: Thanks, I see you already found the relevant connections. You're right: while obviously class `Compiler` does much of the work, from the builder perspective `CompilationResult` is more relevant.

Another piece in the puzzle is `ClassFileReader.hasStructuralChanges()` which has its own dependencies on ecj internals, but that, too, needn't influence the current option.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java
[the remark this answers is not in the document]
stephan-herrmann: In the future perhaps this prepare logic could also be moved to the factory, to feed each compiler implementation those options it needs. Or you could perhaps demonstrate that `CompilerConfiguration` is indeed an abstraction suitable also for ecj, beyond that one field `compilerOptions`. But for the time being this design is acceptable. 

`prepareCompilerConfiguration()` just mustn't throw exceptions (looks fairly safe) nor cause performance issues (looks fairly fast).

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/compiler/ICompilerFactory.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * A factory used to produce a compiler to compile the Java files.
+ */
+public interface ICompilerFactory {
+
+	/**
+	 * Create a new compiler using the given name environment and compiler options.
+	 *
+	 * @param environment - the type system environment used for resolving types and packages
```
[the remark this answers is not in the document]
stephan-herrmann: Before going public this will undergo more review anyway ...

stephan-herrmann:
> > in Eclipse I see many "discouraged" warnings against org.eclipse.jdt.core.tests.builder.mockcompiler.
> 
> No idea for this. I don't see the warnings for them in the Eclipse IDE.

I could still see it, perhaps related to what target platform one had configured in Eclipse, but once I updated to [1627800](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2497/commits/1627800035b53f9c61b5cf81ba7a9d5473ca23bf) the warnings disappeared. Theory: with the redundant Require-Bundle perhaps we saw classes from o.e.jdt.core via the wrong dependency-path: when seen as a regular dependency, access is discouraged, when seen as owned by the fragment host, all is fine.

stephan-herrmann (approved):
Once we have a green build this is good to go - as the experimental option it is, perhaps to be revisited before the next release.

mickaelistria (approved):
I agree the current proposal is a (very) good enough iteration.
I think everything @stephan-herrmann suggested or remarked is relevant though, so if there is a chance to cover some of them easily right now, then I think it's better to do it before merging, so that's fewer reasons to come back to this topic later.
But if implemented the suggested change seems too much effort at the moment, I'm all for merging as is now.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/compiler/ICompilerFactory.java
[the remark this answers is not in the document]
testforstephen: this is just my understanding about the environment. I agree we will revisit this term before making it public.

testforstephen:
@jukzi @stephan-herrmann @jarthana @mickaelistria @fbricon thanks everyone for the review. I appreciate all your feedback.

org.eclipse.jdt.core.tests.builder/src/org/eclipse/jdt/core/tests/builder/BasicBuildTests.java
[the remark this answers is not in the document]
testforstephen: > Another piece in the puzzle is ClassFileReader.hasStructuralChanges() which has its own dependencies on ecj internals, but that, too, needn't influence the current option.

Thanks for pointing it out. Yes, ECJ has some special optimization for incremental build, I'm experimenting to see how to achieve similar capability with Javac.

mpalat:
Hey @testforstephen [Jinbo Wang] - After pulling this change, Am getting compilation errors 
[8 of them - putting the most relevant  below:
Neither an available bundle nor the associated JRE export package 'org.eclipse.jdt.core.tests.builder.mockcompiler'
shown in Manifest]. Is anyone else facing an issue?

jukzi:
me2
image

jarthana:
> Hey @testforstephen [Jinbo Wang] - After pulling this change, Am getting compilation errors [8 of them - putting the most relevant below: Neither an available bundle nor the associated JRE export package 'org.eclipse.jdt.core.tests.builder.mockcompiler' shown in Manifest]. Is anyone else facing an issue?

For the IDE, I guess you will have to import the new projects from the repo after pulling.

akurtakov:
Do you have [org.eclipse.jdt.core.tests.builder.mockcompiler](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2497/files#diff-3a426db964e90d3aa9493054ed05b93783781b5fa61c4cf86482439869b867b5) imported in your workspace?

mpalat:
Hey thanks all - yes, got it resolved. Didn't realize that a new project was added - a rare occurrence in jdt.core world - [and how stupid of me for not taking that possibility into consideration :( ] - thanks!

jukzi:
I am typical using OOmphs "Perform Setup Task" but in this case it does not work. Does the new project need to be registered somewhere for a OOmph setup? @merks

merks:
That depends.  The ones with * requirement will find all projects, the ones with requirements listed will transitively include all those requirements:

image

So if there is something new somewhere in the clones that do not use * and are not required in any feature already required, then it may need to be added.  Just to avoid reading the whole long thread, which specific project is of concern here?

merks:
Doing a pull and resolving the modular targlet resulted in:

> Importing project org.eclipse.jdt.core.tests.builder.mockcompiler

So that appears to be working okay.

jukzi:
> Just to avoid reading the whole long thread, which specific project is of concern here?
C:\Users\jkubitz\platform-2024-06-10\git\eclipse.jdt.core\org.eclipse.jdt.core.tests.builder.mockcompiler\.project

merks:
FYI, the generally running just these two Help -> Perform Setup Tasks will install the latest and update the target platform to the latest, importing new projects as a side effect:

image

jukzi:
my complaint is that this:
image
did not import the new project.
I just tried it again and it failed with

```
Collected 4 artifacts for https://download.eclipse.org/eclipse/updates/4.33-I-builds in 0.768s
Preparing to commit the provisioning operation.
Committing the provisioning operation.
ERROR: org.eclipse.oomph.targlets.core code=0 Resolution problems
  at org.eclipse.oomph.targlets.internal.core.TargletContainer.forceUpdate(TargletContainer.java:919)
  at org.eclipse.oomph.setup.targlets.impl.TargletTaskImpl$4.run(TargletTaskImpl.java:1232)
  at org.eclipse.oomph.util.pde.TargetPlatformUtil.runWithTargetPlatformService(TargetPlatformUtil.java:120)
  at org.eclipse.oomph.setup.targlets.impl.TargletTaskImpl.perform(TargletTaskImpl.java:1092)
  at org.eclipse.oomph.setup.internal.core.SetupTaskPerformer.doPerformNeededSetupTasks(SetupTaskPerformer.java:3864)
  at org.eclipse.oomph.setup.internal.core.SetupTaskPerformer$WorkspaceUtil$1.run(SetupTaskPerformer.java:5200)
  at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2452)
  at org.eclipse.core.internal.resources.Workspace.run(Workspace.java:2477)
  at org.eclipse.oomph.setup.internal.core.SetupTaskPerformer$WorkspaceUtil.performNeededSetupTasks(SetupTaskPerformer.java:5193)
  at org.eclipse.oomph.setup.internal.core.SetupTaskPerformer.performNeededSetupTasks(SetupTaskPerformer.java:3798)
  at org.eclipse.oomph.setup.internal.core.SetupTaskPerformer.performTriggeredSetupTasks(SetupTaskPerformer.java:3773)
  at org.eclipse.oomph.setup.internal.core.SetupTaskPerformer.perform(SetupTaskPerformer.java:3651)
  at org.eclipse.oomph.setup.ui.wizards.ProgressPage$9.run(ProgressPage.java:592)
  at org.eclipse.oomph.setup.ui.wizards.ProgressPage$11$1.run(ProgressPage.java:721)
  at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
  ERROR: org.eclipse.oomph.targlets.core code=0 Problems resolving composed target 'eclipse-sdk-prereqs'
    ERROR: org.eclipse.pde.core code=0 Problems occurred while resolving the target contents
      ERROR: org.eclipse.pde.core code=0 Problems loading repositories
        ERROR: org.eclipse.pde.core code=0 Unable to locate installable unit org.apache.lucene.core 9.11.0.v20240607-1000
        ERROR: org.eclipse.pde.core code=0 Unable to locate installable unit org.apache.lucene.analysis-smartcn 9.11.0.v20240607-1000
        ERROR: org.eclipse.pde.core code=0 Unable to locate installable unit org.apache.lucene.analysis-common 9.11.0.v20240607-1000
        ERROR: org.eclipse.pde.core code=0 Unable to locate installable unit org.apache.lucene.core.source 9.11.0.v20240607-1000
        ERROR: org.eclipse.pde.core code=0 Unable to locate installable unit org.apache.lucene.analysis-smartcn.source 9.11.0.v20240607-1000
        ERROR: org.eclipse.pde.core code=0 Unable to locate installable unit org.apache.lucene.analysis-common.source 9.11.0.v20240607-1000
      ERROR: org.eclipse.pde.core code=0 Problems loading repositories
        ERROR: org.eclipse.pde.core code=0 Unable to locate installable unit org.eclipse.emf.common.feature.group 2.32.0.v20240604-0832
        ERROR: org.eclipse.pde.core code=0 Unable to locate installable unit org.eclipse.emf.common.source.feature.group 2.32.0.v20240604-0832
        ERROR: org.eclipse.pde.core code=0 Unable to locate installable unit org.eclipse.emf.ecore.feature.group 2.38.0.v20240604-0832
        ERROR: org.eclipse.pde.core code=0 Unable to locate installable unit org.eclipse.emf.ecore.source.feature.group 2.38.0.v20240604-0832
        ERROR: org.eclipse.pde.core code=0 Unable to locate installable unit org.eclipse.emf.edit.feature.group 2.23.0.v20240604-0832
        ERROR: org.eclipse.pde.core code=0 Unable to locate installable unit org.eclipse.emf.edit.source.feature.group 2.23.0.v20240604-0832
        ERROR: org.eclipse.pde.core code=0 Unable to locate installable unit org.eclipse.emf.databinding.feature.group 1.12.0.v20240604-0832
        ERROR: org.eclipse.pde.core code=0 Unable to locate installable unit org.eclipse.emf.databinding.source.feature.group 1.12.0.v20240604-0832
        ERROR: org.eclipse.pde.core code=0 Unable to locate installable unit org.eclipse.emf.databinding.edit.feature.group 1.12.0.v20240604-0832
        ERROR: org.eclipse.pde.core code=0 Unable to locate installable unit org.eclipse.emf.databinding.edit.source.feature.group 1.12.0.v20240604-0832
Took 46 seconds.
There are failed tasks.
Press Back to choose different settings or Cancel to abort.
```
might be an unrelated error but prevented to import the new project?

merks:
Ever since we added resolving the actual *.target, this happens once in a while.   If it still fails after a restart, open that *.target in the PDE Editor:

image

and wait for it to resolve properly.

I think you can also do there here in the preferences:

image

It's an annoying problem.  😢

iloveeclipse:
Back to original code now.

While thinking about comments on https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2606 I've noticed that we use a system property to select compiler.

Sorry I haven't participated in the review here, is there any reason that the alternative compiler is supplied as a system property and not compiler option here?
https://github.com/eclipse-jdt/eclipse.jdt.core/blob/b9f1cdd638e42e3e7cc0795d4e0789ac19c87578/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java#L582

Wouldn't be natural to read which compiler is to be used few lines above from project options in https://github.com/eclipse-jdt/eclipse.jdt.core/blob/b9f1cdd638e42e3e7cc0795d4e0789ac19c87578/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java#L561

If I read the code right, this would be possible as the compiler is created per project builder and that also would allow have two compilers running in same IDE? Also this would allow any set of tests running in parallel with both compilers etc, avoid mocking of system properties.

Of course one could still have that property for projects without settings.

mickaelistria:
>  is there any reason that the alternative compiler is supplied as a system property and not compiler option here?

Currently, the only reason is convenience: a system property allows to dynamically switch compiler in the IDE. They allow to repeat the same operation with a simple flag that can be easily controlled in a debugger to compare results, they also allow easy "external" configuration (eg JDT-LS can just add a `-vmArg` to force usage of a compiler or the other).
That said, your proposal might be better; but the idea to use something else than a system property hadn't emerged so far, maybe because it's not useful yet.

> that also would allow have two compilers running in same IDE? 

Yes, that would be possible with your proposal to have distinct projects using distinct compilers in parallel. Currently, the system property only allows to switch the compiler globally for all projects (but dynamically).

iloveeclipse:
@mickaelistria : so would you or @testforstephen follow up on that proposal?

mickaelistria:
> @mickaelistria : so would you or @testforstephen follow up on that proposal?

This particular proposal is not so interesting for our current goals regarding Javac integration. However, if you are certain that this can be soon profitable for the JDT Core project, and since it's not too much code to change, I think we can invest the small effort to unleash extra value to the ecosystem. So the (genuine) question is now for you: do you really think if we develop it now, it's going to be useful soon?

stephan-herrmann:
> Sorry I haven't participated in the review here, is there any reason that the alternative compiler is supplied as a system property and not compiler option here?

I see the system property as an implementation of the statement that this is an experimental flag not to be exposed to any wider audience. Right now we wouldn't want the option to (automatically?) appear in `.settings/org.eclipse.jdt.core.prefs`, for people to look at and start searching for documentation :)

Creating a compiler option can IMHO wait until we discuss making this an official feature.

PR #2519, opened 2024-06-04
dom: fix ASTNode MT issues

jukzi:
Static code analysis complains that newChild.setParent had modified fields after initialize. fixes:

* 'Remove this dangerous instance of double-checked locking'
* 'Fully initialize ... before assigning it.'
* atomic increment
* wrong fieldname in GuardedPattern.getExpression()

found by: SonarLint

### DIFF
--- a/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTNode.java
+++ b/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTNode.java
@@ -2880,15 +2880,20 @@ public abstract class ASTNode {
 	/**
      * Begin lazy initialization of this node.
      * Here is the code pattern found in all AST
-     * node subclasses:
+     * node subclasses. For thread safety it uses a "double-checked locking idiom":
+     *
      * <pre>
-     * if (this.foo == null) {
-	 *    // lazy init must be thread-safe for readers
+     * private volatile ASTNode node; // has to be declared volatile and ASTNode has to be threadsafe!
+     * ...
+     * if (this.node == null) {
      *    synchronized (this) {
-     *       if (this.foo == null) {
+     *       if (this.node == null) { // double check
      *          preLazyInit();
-     *          this.foo = ...; // code to create new node
-     *          postLazyInit(this.foo, FOO_PROPERTY);
+     *          ASTNode node = ...; // code to create new node
+     *          node.xyz = ...; // initialize all fields
+     *          // Finally write the full initialized field:
+     *          this.node = postLazyInit(node, FOO_PROPERTY);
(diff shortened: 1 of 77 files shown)


org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTNode.java
```diff
@@ -2879,20 +2879,6 @@ final void checkModifiable() {
...
-	 *    // lazy init must be thread-safe for readers
-     *    synchronized (this) {
-     *       if (this.foo == null) {
-     *          preLazyInit();
-     *          this.foo = ...; // code to create new node
-     *          postLazyInit(this.foo, FOO_PROPERTY);
-     *       }
-     *    }
-     * }
-     * </pre>
```
stephan-herrmann, line 2895: I didn't quite understand what's wrong with this code pattern, but if you change it, wouldn't it be nice to update this documentation, rather than removing?

Just recently I referred to this documentation in a review because I found it helpful.

jukzi: ` this.foo = ...` writes to memory before executing postLazyInit, so any concurrent thread will see `this.foo != null` in the first check `if (this.foo == null) {` before entering synchronized. The other thread willl therefor not wait for postLazyInit to happen and may use a object that has is parent not initialized - if by chance postLazyInit() has not been executed yet.
I removed the documentation because it was a wrong pattern. All callers should now use same safe pattern where it can be copied from. But if you want i can also update the javadoc.

stephan-herrmann: > ` this.foo = ...` writes to memory before executing postLazyInit, so any concurrent thread will see `this.foo != null` in the first check `if (this.foo == null) {` before entering synchronized. The other thread willl therefor not wait for postLazyInit to happen and may use a object that has its parent not initialized - if by chance postLazyInit() has not been executed yet.

understood, thanks for explaining.

> I removed the documentation because it was a wrong pattern. All callers should now use same safe pattern where it can be copied from. But if you want i can also update the javadoc.

If we want all future additions to adhere to the new pattern, then documentation looks like a helpful device, no?

mpalat: good catch @jukzi ! 

Also, as @stephan-herrmann please update javadoc since DOM AST is always evolving and it would help in future commits. thanks!

jukzi: updated javadoc

PR #2555, opened 2024-06-12
[performance] Use new IFile API: IFile.readAllBytes() / readAllChars()

jukzi:
Instead of streaming IFile.getContents(true)

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/classfmt/ClassFileReader.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/classfmt/ClassFileReader.java
@@ -118,12 +118,12 @@ public static ClassFileReader read(File file, boolean fullyInitialize) throws Cl
 	return classFileReader;
 }
 
-public static ClassFileReader read(InputStream stream, String fileName) throws ClassFormatException, IOException {
-	return read(stream, fileName, false);
+public static ClassFileReader read(byte[] classFileBytes, String fileName) throws ClassFormatException, IOException {
+	return read(classFileBytes, fileName, false);
 }
 
-public static ClassFileReader read(InputStream stream, String fileName, boolean fullyInitialize) throws ClassFormatException, IOException {
-	byte classFileBytes[] = Util.getInputStreamAsByteArray(stream);
+public static ClassFileReader read(byte[] classFileBytes, String fileName, boolean fullyInitialize)
+		throws ClassFormatException {
 	ClassFileReader classFileReader = new ClassFileReader(classFileBytes, fileName.toCharArray());
 	if (fullyInitialize) {
 		classFileReader.initialize();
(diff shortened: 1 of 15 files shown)


iloveeclipse:
This breaks Xtext, see https://github.com/eclipse/xtext/issues/3089.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/classfmt/ClassFileReader.java
```diff
@@ -118,12 +118,12 @@ public static ClassFileReader read(File file, boolean fullyInitialize) throws Cl
 	return classFileReader;
 }
 
-public static ClassFileReader read(InputStream stream, String fileName) throws ClassFormatException, IOException {
-	return read(stream, fileName, false);
+public static ClassFileReader read(byte[] classFileBytes, String fileName) throws ClassFormatException, IOException {
```
szarnekow, line 121: This change breaks existing installations that use Xtext. Why wasn’t it possible to do a backwards compatible modification instead?

szarnekow: Also this change here doesn't seem to be related to IFile.readAllBytes

org.eclipse.jdt.core/META-INF/MANIFEST.MF
```diff
@@ -42,9 +42,9 @@ Export-Package: org.eclipse.jdt.core,
...
  org.eclipse.jdt.internal.formatter.linewrap;x-friends:="org.eclipse.jdt.core.tests.model, org.eclipse.jdt.core.tests.compiler, org.eclipse.jdt.core.tests.builder, org.eclipse.jdt.core.tests.performance, org.eclipse.jdt.ui.tests",
  org.eclipse.jdt.internal.formatter.old;x-friends:="org.eclipse.jdt.core.tests.model, org.eclipse.jdt.core.tests.compiler, org.eclipse.jdt.core.tests.builder, org.eclipse.jdt.core.tests.performance, org.eclipse.jdt.ui.tests"
-Require-Bundle: org.eclipse.core.resources;bundle-version="[3.18.0,4.0.0)",
+Require-Bundle: org.eclipse.core.resources;bundle-version="[3.21.0,4.0.0)",
  org.eclipse.core.runtime;bundle-version="[3.29.0,4.0.0)",
- org.eclipse.core.filesystem;bundle-version="[1.7.0,2.0.0)",
+ org.eclipse.core.filesystem;bundle-version="[1.11.0,2.0.0)",
  org.eclipse.text;bundle-version="[3.6.0,4.0.0)",
  org.eclipse.team.core;bundle-version="[3.1.0,4.0.0)";resolution:=optional,
  org.eclipse.jdt.core.compiler.batch;bundle-version="3.38.0";visibility:=reexport
```
szarnekow, line 50: This is re-exported from 3.38.0 but the dependencies within this repo now use API that's only available in 3.38.100

iloveeclipse: I assume during 4.33 the ecj version will be bumped to 3.39 anyway. Is that OK, or do you see problems for M1/M2 time frame and need a fix?

PR #2560, opened 2024-06-12
An extension point allowing third parties to override source to dom tree implementations

robstryker:
## What it does

This extension point would allow third party products to override the conversion between java source files and the JDT dom tree with a custom implementation. In order to prevent rogue extenders from simply breaking people's installations, a system property selecting the preferred ICompilationUnitResolver must be set, as well. 

As consumers of JDT become more diverse and also have been operating in environments not previously expected, being able to properly define the borders and interactions between and among the various parts of JDT becomes important. As resources to maintain various parts of the codebase become scarce, allowing the community to experiment with potentially more efficient solutions also becomes important. 

Ensuring that the community can contribute to a common codebase for the vast majority of the shared work is critical to providing that community  the cohesion, common goals, and shared visions that maintains a project. And allowing different elements of the community the freedom, via extension points or otherwise, to investigate alternatives to parts of the current codebase will serve to keep the project growing long term. 

## How to test

Other than exposing an extension point, this change does not significantly change any behavior. To test, one could attempt to make a plugin using this extension point and verifying that the various interface methods are called in a runtime workbench. I am not familiar with how the jdt.core project prefers to automatically test extension points, especially ones that require a sysprop to also be set. 

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/CompilationUnitResolverDiscoveryTest.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/CompilationUnitResolverDiscoveryTest.java
@@ -1,0 +1,184 @@
+/*******************************************************************************
+ * Copyright (c) 2024 Red Hat Inc and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.jdt.core.tests.dom;
+
+import java.util.List;
+import java.util.Map;
+
+import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.jdt.core.ICompilationUnit;
+import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.jdt.core.JavaCore;
+import org.eclipse.jdt.core.JavaModelException;
+import org.eclipse.jdt.core.WorkingCopyOwner;
+import org.eclipse.jdt.core.dom.AST;
+import org.eclipse.jdt.core.dom.ASTNode;
+import org.eclipse.jdt.core.dom.ASTRequestor;
+import org.eclipse.jdt.core.dom.CompilationUnit;
+import org.eclipse.jdt.core.dom.FileASTRequestor;
+import org.eclipse.jdt.core.dom.IMethodBinding;
+import org.eclipse.jdt.core.dom.MethodDeclaration;
+import org.eclipse.jdt.core.dom.PackageDeclaration;
+import org.eclipse.jdt.internal.compiler.batch.FileSystem.Classpath;
+import org.eclipse.jdt.internal.compiler.classfmt.ClassFileConstants;
+import org.eclipse.jdt.internal.compiler.impl.CompilerOptions;
+import org.eclipse.jdt.internal.core.dom.ICompilationUnitResolver;
+
+import junit.framework.Test;
+
+public class CompilationUnitResolverDiscoveryTest extends ConverterTestSetup {
+
+	ICompilationUnit workingCopy;
+
+	@Override
+	public void setUpSuite() throws Exception {
+		super.setUpSuite();
+		this.ast = AST.newAST(AST.getJLSLatest(), false);
+	}
+
+	public CompilationUnitResolverDiscoveryTest(String name) {
+		super(name);
+	}
+
+	public static Test suite() {
+		String javaVersion = System.getProperty("java.version");
+		int index = -1;
+		if ( (index = javaVersion.indexOf('-')) != -1) {
+			javaVersion = javaVersion.substring(0, index);
+		} else {
+			if (javaVersion.length() > 3) {
+				javaVersion = javaVersion.substring(0, 3);
+			}
+		}
+		long jdkLevel = CompilerOptions.versionToJdkLevel(javaVersion);
+		if (jdkLevel >= ClassFileConstants.JDK9) {
+			isJRE9 = true;
+		}
+		return buildModelTestSuite(CompilationUnitResolverDiscoveryTest.class);
+	}
+
+	@Override
+	protected void tearDown() throws Exception {
+		super.tearDown();
+		if (this.workingCopy != null) {
+			this.workingCopy.discardWorkingCopy();
+			this.workingCopy = null;
+		}
+	}
+
+	public void testCompilationUnitResolverNoSysprop() throws JavaModelException {
+		String SELECTED_SYSPROP = "ICompilationUnitResolver";
+		String original = System.getProperty(SELECTED_SYSPROP);
+		try {
+			System.clearProperty(SELECTED_SYSPROP);
+			ICompilationUnit sourceUnit = getCompilationUnit("Converter9" , "src", "testBug497719_001", "X.java"); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$
+			ASTNode result = runConversion(this.ast.apiLevel(), sourceUnit, true, true);
+			assertTrue("Not a compilation unit", result.getNodeType() == ASTNode.COMPILATION_UNIT);
+			CompilationUnit compilationUnit = (CompilationUnit) result;
+			ASTNode node = getASTNode(compilationUnit, 0, 0);
+			assertEquals("Not a compilation unit", ASTNode.METHOD_DECLARATION, node.getNodeType());
+			MethodDeclaration methodDeclaration = (MethodDeclaration) node;
+			IMethodBinding mb = methodDeclaration.resolveBinding();
+			assertEquals(mb.getClass().getName(), "org.eclipse.jdt.core.dom.MethodBinding");
+		} finally {
+			System.setProperty(SELECTED_SYSPROP, original);
+		}
+	}
+
+	public void testCompilationUnitResolverInvalidSysprop() throws JavaModelException {
+		String SELECTED_SYSPROP = "ICompilationUnitResolver";
+		String original = System.getProperty(SELECTED_SYSPROP);
+		try {
+			System.setProperty(SELECTED_SYSPROP, "doesNotExist");
+			ICompilationUnit sourceUnit = getCompilationUnit("Converter9" , "src", "testBug497719_001", "X.java"); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$
+			ASTNode result = runConversion(this.ast.apiLevel(), sourceUnit, true, true);
+			assertTrue("Not a compilation unit", result.getNodeType() == ASTNode.COMPILATION_UNIT);
+			CompilationUnit compilationUnit = (CompilationUnit) result;
+			ASTNode node = getASTNode(compilationUnit, 0, 0);
+			assertEquals("Not a compilation unit", ASTNode.METHOD_DECLARATION, node.getNodeType());
+			MethodDeclaration methodDeclaration = (MethodDeclaration) node;
+			IMethodBinding mb = methodDeclaration.resolveBinding();
+			assertEquals(mb.getClass().getName(), "org.eclipse.jdt.core.dom.MethodBinding");
+		} finally {
+			System.setProperty(SELECTED_SYSPROP, original);
+		}
+	}
+
+	public void testCompilationUnitResolverValidSysprop() throws JavaModelException {
+		String SELECTED_SYSPROP = "ICompilationUnitResolver";
+		String original = System.getProperty(SELECTED_SYSPROP);
+		try {
+			System.setProperty(SELECTED_SYSPROP, "org.eclipse.jdt.core.tests.model.resolver1");
+			ICompilationUnit sourceUnit = getCompilationUnit("Converter9" , "src", "testBug497719_001", "X.java"); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$
+			ASTNode result = runConversion(this.ast.apiLevel(), sourceUnit, true, true);
+			assertTrue("Not a compilation unit", result.getNodeType() == ASTNode.COMPILATION_UNIT);
+			CompilationUnit compilationUnit = (CompilationUnit) result;
+			PackageDeclaration pd = compilationUnit.getPackage();
+			assertNotNull(pd);
+			String pdName = pd.getName().toString();
+			assertEquals(pdName, "compilationUnitResolverDiscoveryTest");
+		} finally {
+			System.setProperty(SELECTED_SYSPROP, original);
+		}
+	}
+
+
+	/*
+	 * Custom made interface that implements exactly to the test
+	 */
+	public static final class TEST_RESOLVER implements ICompilationUnitResolver {
+		@Override
+		public CompilationUnit toCompilationUnit(org.eclipse.jdt.internal.compiler.env.ICompilationUnit sourceUnit,
+				boolean initialNeedsToResolveBinding, IJavaProject project, List<Classpath> classpaths,
+				int focalPosition, int apiLevel, Map<String, String> compilerOptions,
+				WorkingCopyOwner parsedUnitWorkingCopyOwner, WorkingCopyOwner typeRootWorkingCopyOwner, int flags,
+				IProgressMonitor monitor) {
+			// Return a mostly-invalid hard-coded dom tree for the purpose of this test
+			AST ast = AST.newAST(apiLevel, JavaCore.ENABLED.equals(compilerOptions.get(JavaCore.COMPILER_PB_ENABLE_PREVIEW_FEATURES)));
+			CompilationUnit res = ast.newCompilationUnit();
+			PackageDeclaration pack = ast.newPackageDeclaration();
+			pack.setName(ast.newSimpleName("compilationUnitResolverDiscoveryTest"));
+			res.setPackage(pack);
+			return res;
+		}
+
+		@Override
+		public void resolve(String[] sourceFilePaths, String[] encodings, String[] bindingKeys,
+				FileASTRequestor requestor, int apiLevel, Map<String, String> compilerOptions,
+				List<Classpath> classpathList, int flags, IProgressMonitor monitor) {
+		}
+
+		@Override
+		public void parse(ICompilationUnit[] compilationUnits, ASTRequestor requestor, int apiLevel,
+				Map<String, String> compilerOptions, int flags, IProgressMonitor monitor) {
+			// TODO Auto-generated method stub
+
+		}
+
+		@Override
+		public void parse(String[] sourceFilePaths, String[] encodings, FileASTRequestor requestor, int apiLevel,
+				Map<String, String> compilerOptions, int flags, IProgressMonitor monitor) {
+			// TODO Auto-generated method stub
+
+		}
+
+		@Override
+		public void resolve(ICompilationUnit[] compilationUnits, String[] bindingKeys, ASTRequestor requestor,
+				int apiLevel, Map<String, String> compilerOptions, IJavaProject project,
+				WorkingCopyOwner workingCopyOwner, int flags, IProgressMonitor monitor) {
+			// TODO Auto-generated method stub
+
+		}
+	}
+}
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/CompilationUnitResolverDiscoveryTest.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/CompilationUnitResolverDiscoveryTest.java
@@ -9,7 +9,7 @@
  * SPDX-License-Identifier: EPL-2.0
  *
  * Contributors:
- *     IBM Corporation - initial API and implementation
+ *     Red Hat Inc - initial API and implementation
  *******************************************************************************/
 package org.eclipse.jdt.core.tests.dom;
 
@@ -162,23 +162,20 @@ public class CompilationUnitResolverDiscoveryTest extends ConverterTestSetup {
 		@Override
 		public void parse(ICompilationUnit[] compilationUnits, ASTRequestor requestor, int apiLevel,
 				Map<String, String> compilerOptions, int flags, IProgressMonitor monitor) {
-			// TODO Auto-generated method stub
-
+			// irrelevant for test
 		}
 
 		@Override
 		public void parse(String[] sourceFilePaths, String[] encodings, FileASTRequestor requestor, int apiLevel,
 				Map<String, String> compilerOptions, int flags, IProgressMonitor monitor) {
-			// TODO Auto-generated method stub
-
+			// irrelevant for test
 		}
 
 		@Override
 		public void resolve(ICompilationUnit[] compilationUnits, String[] bindingKeys, ASTRequestor requestor,
 				int apiLevel, Map<String, String> compilerOptions, IJavaProject project,
 				WorkingCopyOwner workingCopyOwner, int flags, IProgressMonitor monitor) {
-			// TODO Auto-generated method stub
-
+			// irrelevant for test
 		}
 	}
 }
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/CompilationUnitResolverDiscoveryTest.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/CompilationUnitResolverDiscoveryTest.java
@@ -92,7 +92,11 @@ public class CompilationUnitResolverDiscoveryTest extends ConverterTestSetup {
 			IMethodBinding mb = methodDeclaration.resolveBinding();
 			assertEquals(mb.getClass().getName(), "org.eclipse.jdt.core.dom.MethodBinding");
 		} finally {
-			System.setProperty(SELECTED_SYSPROP, original);
+			if (original == null) {
+				System.clearProperty(SELECTED_SYSPROP);
+			} else {
+				System.setProperty(SELECTED_SYSPROP, original);
+			}
 		}
 	}
 
@@ -111,7 +115,11 @@ public class CompilationUnitResolverDiscoveryTest extends ConverterTestSetup {
 			IMethodBinding mb = methodDeclaration.resolveBinding();
 			assertEquals(mb.getClass().getName(), "org.eclipse.jdt.core.dom.MethodBinding");
 		} finally {
-			System.setProperty(SELECTED_SYSPROP, original);
+			if (original == null) {
+				System.clearProperty(SELECTED_SYSPROP);
+			} else {
+				System.setProperty(SELECTED_SYSPROP, original);
+			}
 		}
 	}
 
@@ -129,7 +137,11 @@ public class CompilationUnitResolverDiscoveryTest extends ConverterTestSetup {
 			String pdName = pd.getName().toString();
 			assertEquals(pdName, "compilationUnitResolverDiscoveryTest");
 		} finally {
-			System.setProperty(SELECTED_SYSPROP, original);
+			if (original == null) {
+				System.clearProperty(SELECTED_SYSPROP);
+			} else {
+				System.setProperty(SELECTED_SYSPROP, original);
+			}
 		}
 	}
 
--- a/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
+++ b/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
@@ -48,6 +48,7 @@ import org.eclipse.jdt.internal.core.ClassFileWorkingCopy;
 import org.eclipse.jdt.internal.core.DefaultWorkingCopyOwner;
 import org.eclipse.jdt.internal.core.JavaModelManager;
 import org.eclipse.jdt.internal.core.PackageFragment;
+import org.eclipse.jdt.internal.core.dom.ICompilationUnitResolver;
 import org.eclipse.jdt.internal.core.dom.util.DOMASTUtil;
 import org.eclipse.jdt.internal.core.util.CodeSnippetParsingUtil;
 import org.eclipse.jdt.internal.core.util.RecordedParsingInformation;
--- a/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
+++ b/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
@@ -33,7 +33,6 @@ import org.eclipse.jdt.core.JavaModelException;
 import org.eclipse.jdt.core.WorkingCopyOwner;
 import org.eclipse.jdt.core.compiler.CategorizedProblem;
 import org.eclipse.jdt.core.compiler.CharOperation;
-import org.eclipse.jdt.internal.compiler.ast.CompilationUnitDeclaration;
 import org.eclipse.jdt.internal.compiler.ast.ConstructorDeclaration;
 import org.eclipse.jdt.internal.compiler.ast.ExplicitConstructorCall;
 import org.eclipse.jdt.internal.compiler.batch.FileSystem.Classpath;
@@ -221,6 +220,8 @@ public class ASTParser {
 	 */
 	private int bits;
 
+	private final ICompilationUnitResolver unitResolver;
+
 	/**
 	 * Creates a new AST parser for the given API level.
 	 * <p>
@@ -233,6 +234,7 @@ public class ASTParser {
 	ASTParser(int level) {
 		DOMASTUtil.checkASTLevel(level);
 		this.apiLevel = level;
+		this.unitResolver = CompilationUnitResolverDiscovery.getInstance();
 		initializeDefaults();
 	}
 
@@ -245,10 +247,13 @@ public class ASTParser {
 			}
 			if (this.sourcepaths != null) {
 				for (int i = 0, max = this.sourcepaths.length; i < max; i++) {
-					String encoding = this.sourcepathsEncodings == null ? null : this.sourcepathsEncodings[i];
-					main.processPathEntries(
-							Main.DEFAULT_SIZE_CLASSPATH,
-							allClasspaths, this.sourcepaths[i], encoding, true, false);
+					String sourcePath = this.sourcepaths[i];
+					if (sourcePath != null) {
+						String encoding = this.sourcepathsEncodings == null ? null : this.sourcepathsEncodings[i];
+						main.processPathEntries(
+								Main.DEFAULT_SIZE_CLASSPATH,
+								allClasspaths, sourcePath, encoding, true, false);
+					}
 				}
 			}
 			if (this.classpaths != null) {
@@ -957,9 +962,9 @@ public class ASTParser {
 				if ((this.bits & CompilationUnitResolver.BINDING_RECOVERY) != 0) {
 					flags |= ICompilationUnit.ENABLE_BINDINGS_RECOVERY;
 				}
-				CompilationUnitResolver.resolve(compilationUnits, bindingKeys, requestor, this.apiLevel, this.compilerOptions, this.project, this.workingCopyOwner, flags, monitor);
+				this.unitResolver.resolve(compilationUnits, bindingKeys, requestor, this.apiLevel, this.compilerOptions, this.project, this.workingCopyOwner, flags, monitor);
 			} else {
-				CompilationUnitResolver.parse(compilationUnits, requestor, this.apiLevel, this.compilerOptions, flags, monitor);
+				this.unitResolver.parse(compilationUnits, requestor, this.apiLevel, this.compilerOptions, flags, monitor);
 			}
 		} finally {
 			// reset to defaults to allow reuse (and avoid leaking)
@@ -1052,9 +1057,9 @@ public class ASTParser {
 				if ((this.bits & CompilationUnitResolver.BINDING_RECOVERY) != 0) {
 					flags |= ICompilationUnit.ENABLE_BINDINGS_RECOVERY;
 				}
-				CompilationUnitResolver.resolve(sourceFilePaths, encodings, bindingKeys, requestor, this.apiLevel, this.compilerOptions, getClasspath(), flags, monitor);
+				this.unitResolver.resolve(sourceFilePaths, encodings, bindingKeys, requestor, this.apiLevel, this.compilerOptions, getClasspath(), flags, monitor);
 			} else {
-				CompilationUnitResolver.parse(sourceFilePaths, encodings, requestor, this.apiLevel, this.compilerOptions, flags, monitor);
+				this.unitResolver.parse(sourceFilePaths, encodings, requestor, this.apiLevel, this.compilerOptions, flags, monitor);
 			}
 		} finally {
 			// reset to defaults to allow reuse (and avoid leaking)
@@ -1159,9 +1164,8 @@ public class ASTParser {
 				}
 				break;
 			case K_COMPILATION_UNIT :
-				CompilationUnitDeclaration compilationUnitDeclaration = null;
 				try {
-					NodeSearcher searcher = null;
+					boolean useSearcher = false;
 					org.eclipse.jdt.internal.compiler.env.ICompilationUnit sourceUnit = null;
 					WorkingCopyOwner wcOwner = this.workingCopyOwner;
 					if (this.typeRoot instanceof ClassFileWorkingCopy) {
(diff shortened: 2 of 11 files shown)


org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
```diff
@@ -245,10 +247,13 @@ private List<Classpath> getClasspath() throws IllegalStateException {
 			}
 			if (this.sourcepaths != null) {
 				for (int i = 0, max = this.sourcepaths.length; i < max; i++) {
-					String encoding = this.sourcepathsEncodings == null ? null : this.sourcepathsEncodings[i];
```
robstryker, line 248: This is a mostly-unrelated NPE check that we found was required in some instances. If you wish this to be removed from the PR I will do so.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
```diff
@@ -1229,68 +1233,28 @@ private ASTNode internalCreateASTCached(IProgressMonitor monitor) {
 						throw new IllegalStateException();
 					}
 					if ((this.bits & CompilationUnitResolver.PARTIAL) != 0) {
-						searcher = new NodeSearcher(this.focalPointPosition);
```
robstryker, line 1232: Small note on this code: rather than pass around the NodeSearcher, I have opted to pass the focalPointPosition (or -1) around, as NodeSearcher is not API and should not be in any interfaces.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ICompilationUnitResolver.java
```diff
@@ -0,0 +1,108 @@
...
+	 * @param classpaths A list of classpaths to use during this operation
+	 * @param focalPosition a position to focus on, or -1 if N/A
+	 * @param apiLevel Level of AST API desired.
+	 * @param compilerOptions Compiler options. Defaults to JavaCore.getOptions().
+	 * @param parsedUnitWorkingCopyOwner The working copy owner of the unit
+	 * @param typeRootWorkingCopyOwner The working copy owner of the type
+	 * @param flags Flags to to be used during this operation
+	 * @param monitor A progress monitor
+	 * @return A CompilationUnit
+	 */
```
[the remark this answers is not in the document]
robstryker: I think package-private is fine for this type of situation and serves as a great halfway point for separations and refactors but before an API stage is warranted.

robstryker:
@testforstephen Would you mind giving a review on this PR?  

Thanks.

testforstephen:
> Would you mind giving a review on this PR?

sure, I'll take a look next week.

rgrunber:
@testforstephen , also something to consider. I noticed your compiler extension PR extends via system property containing the FQN + `Class.forName(..)` on it. Rob's current approach is to use the extension point framework. It seems like it would be better for all extensions to standardize on a single approach.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
```diff
@@ -1229,68 +1233,28 @@ private ASTNode internalCreateASTCached(IProgressMonitor monitor) {
...
-						wcOwner,
-						needToResolveBindings ? new DefaultBindingResolver.BindingTables() : null,
-						flags,
-						monitor,
-						this.project != null,
-						this.project);
-					result.setTypeRoot(this.typeRoot);
-					return result;
+
+					CompilationUnit result2 = this.unitResolver.toCompilationUnit(sourceUnit, needToResolveBindings, this.project, getClasspath(), useSearcher ? this.focalPointPosition : -1, this.apiLevel, this.compilerOptions, this.workingCopyOwner, wcOwner, flags, monitor);
```
testforstephen, line 1252: What is the reason for extracting this part of the logic into a new API `toCompilationUnit`?

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolverDiscovery.java
```diff
@@ -0,0 +1,58 @@
...
+import org.eclipse.core.runtime.Platform;
+import org.eclipse.jdt.core.JavaCore;
+
+class CompilationUnitResolverDiscovery {
+	private static final String COMPILATION_UNIT_RESOLVER_EXTPOINT_ID = "compilationUnitResolver" ; //$NON-NLS-1$
+	private static boolean ERROR_LOGGED = false;
+
+
+	static ICompilationUnitResolver getInstance() {
+		String compilationUnitResolverId = System.getProperty(ICompilationUnitResolver.class.getSimpleName());
```
testforstephen, line 27: it's more straightforward to define the property key as a string constant.

testforstephen (commented):
Overall, the extension point looks clean to me. Could you add some unit tests for it?

testforstephen:
> @testforstephen , also something to consider. I noticed your compiler extension PR extends via system property containing the FQN + `Class.forName(..)` on it. Rob's current approach is to use the extension point framework. It seems like it would be better for all extensions to standardize on a single approach.

I'd keep the compiler extension PR as it is for now. If we encounter any limitations with this approach, we can consider changing it to an extension point later.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
mickaelistria: I guess it's more convenience. What do you think would be better here? Making the whole `internalCreateASTForKind` abstract or part of the interface?

testforstephen: The CompilationUnitResolver provides three basic operations: parse, resolve, and convert. I'm wondering why we don't simply extract these operations into the new API. Then, we would only need to replace the resolver instance in the ASTParser.

mickaelistria: The issue with convert is that it is parser specific. So we cannot easily `convert` as it in a parser-independent interface. I think that's why we extract `toCompilationUnit` in 1st place, to hide the internal parser objects and the parser-specific conversion work.

robstryker: The `convert` method has `DefaultBindingResolver.BindingTables bindingTables` as one of its parameters. Not only is this internal, but we also intend on using a binding resolver class, and so forcing this parameter into the interface is inappropriate. 

I believe the `toCompilationUnit` method is appropriate as designed here. It has a signature that allows extenders the most flexibility with no extraneous parameters that wouldn't make sense for a custom implementation.

testforstephen: Understood, the new API `toCompilationUnit` makes sense. A side note is the new public API uses two internal ECJ classes such as `org.eclipse.jdt.internal.compiler.env.ICompilationUnit` and `org.eclipse.jdt.internal.compiler.batch.FileSystem.Classpath`. How about passing the raw data (contents, ... sourcepaths, ...) or wrapping them in records?

For example, 
`public record SourceUnit(contents, packageName, fileName, javaElement)`
`public record Classpath(sourcepaths, classpaths, includeVMBootpath)`

robstryker:
Working on test cases. Will update PR asap.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
mickaelistria: I'm not sure it would be immediately profitable. Those internal compiler classes are likely to stay for a while in practice and it could be a bit too early to refactor this. Maybe later we'll have some concrete case as to why *not* requiring them, but we're not there yet IMO, so it's simpler and faster to stick to those (useful) classes for an internal/experimental extension point which is constrained on many aspects already.

As to using a record, then one issue would be that we would have to actually resolve all values at creation. I'm not so familiar with implementation of eg ICompilationUnit, but I'm wondering what if those are dynamic; more dynamic than what a record can capture? This is also 1 reason why I think refactoring that immediately may not be best.

IMO, it would be better to keep the current state of this proposal, make annotations here and there wherever it makes sense to clearify that all that is experimental/provisional; and then let more experience lead to improving the design later.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ICompilationUnitResolver.java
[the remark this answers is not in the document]
robstryker: We ended up switching to a public interface in an internal package.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
robstryker: Unit tests have been added

mickaelistria (approved):
I agree the current proposal is a (very) good enough iteration.
I think everything @stephan-herrmann suggested or remarked is relevant though, so if there is a chance to cover some of them easily right now, then I think it's better to do it before merging, so that's fewer reasons to come back to this topic later.
But if implemented the suggested change seems too much effort at the moment, I'm all for merging as is now.

org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/ICompilationUnitResolver.java
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * This interface is used to resolve a jdt dom tree from source files.
+ * It is contributed to via the compilationUnitResolver extension point.
+ * This interface is currently internal only, and is not considered API.
+ * This interface may be modified, changed, or removed at any time.
+ *
+ * @since 3.38
+ */
+public interface ICompilationUnitResolver {
```
jukzi, line 33: This is in internal package, meant as provisional API? doesn't that contradict plugin mechanism?

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/CompilationUnitResolverDiscoveryTest.java
```diff
@@ -0,0 +1,184 @@
...
+		@Override
+		public void resolve(String[] sourceFilePaths, String[] encodings, String[] bindingKeys,
+				FileASTRequestor requestor, int apiLevel, Map<String, String> compilerOptions,
+				List<Classpath> classpathList, int flags, IProgressMonitor monitor) {
+		}
+
+		@Override
+		public void parse(ICompilationUnit[] compilationUnits, ASTRequestor requestor, int apiLevel,
+				Map<String, String> compilerOptions, int flags, IProgressMonitor monitor) {
+			// TODO Auto-generated method stub
```
jukzi, line 165: open TODOs -> mark as draft?

org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/ICompilationUnitResolver.java
mickaelistria: The extension point itself is provisional: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2560/files#diff-593bbeeaa5bd91d942bb0a52abe6bc4829af2b970e665caa842759bf0ff7691fR9 says "This extension point is not considered API.". We hope to make it API later if there is interest, when we have consolidated with the Javac experience, but while it is already valuable enough to have an extension point, it's still too early to guarantee API stability.

org.eclipse.jdt.core/schema/compilationUnitResolver.exsd
```diff
@@ -0,0 +1,121 @@
...
+      <annotation>
+         <documentation>
+            Definition of a compilation unit resolver.
+         </documentation>
+      </annotation>
+      <complexType>
+         <attribute name="class" type="string" use="required">
+            <annotation>
+               <documentation>
+                  The class that implements this compilation unit resolver. This class must implement the  &lt;code&gt;org.eclipse.jdt.core.dom.ICompilationUnitResolver&lt;/code&gt; interface with a public 0-arg constructor.
```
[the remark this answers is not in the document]
mickaelistria: See the next lines of this documentation that mention that this extension point is currently internal/non-API too.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/CompilationUnitResolverDiscoveryTest.java
mickaelistria: As it's just testing the extension point mechanism, the implementation doesn't matter. This TODO should be replaced with some `// does not matter for test`

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolverDiscovery.java
```diff
@@ -0,0 +1,60 @@
...
+import org.eclipse.core.runtime.IExtensionPoint;
+import org.eclipse.core.runtime.ILog;
+import org.eclipse.core.runtime.Platform;
+import org.eclipse.jdt.core.JavaCore;
+import org.eclipse.jdt.internal.core.dom.ICompilationUnitResolver;
+
+class CompilationUnitResolverDiscovery {
+	private static final String SELECTED_SYSPROP = "ICompilationUnitResolver"; //$NON-NLS-1$
+	private static final String COMPILATION_UNIT_RESOLVER_EXTPOINT_ID = "compilationUnitResolver" ; //$NON-NLS-1$
+	private static boolean ERROR_LOGGED = false;
```
jukzi, line 25: can ERROR_LOGGED) accessed from different threads?

org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/ICompilationUnitResolver.java
jukzi: then the javadoc has to clearly state so

org.eclipse.jdt.core/schema/compilationUnitResolver.exsd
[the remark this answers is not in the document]
jukzi: the package name is wrong

org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/ICompilationUnitResolver.java
mickaelistria: Javadoc already says " This interface is currently internal only, and is not considered API. This interface may be modified, changed, or removed at any time.". What exactly do you think is missing?

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolverDiscovery.java
mickaelistria: Possibly yes.

jukzi: then the access need some synchronization. for example volatile modifier

org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/ICompilationUnitResolver.java
jukzi: `<strong>EXPERIMENTAL</strong>.`
https://github.com/eclipse-platform/eclipse.platform/blob/master/docs/Provisional_API_Guidelines.md

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolver.java
```diff
@@ -76,12 +76,52 @@
...
+
+		@Override
+		public CompilationUnit toCompilationUnit(org.eclipse.jdt.internal.compiler.env.ICompilationUnit sourceUnit, final boolean initialNeedsToResolveBinding, IJavaProject project, List<Classpath> classpaths, int focalPosition,
+			int apiLevel, Map<String, String> compilerOptions, WorkingCopyOwner parsedUnitWorkingCopyOwner, WorkingCopyOwner typeRootWorkingCopyOwner, int flags, IProgressMonitor monitor) {
+			return CompilationUnitResolver.toCompilationUnit(sourceUnit, initialNeedsToResolveBinding, project,
+					classpaths, focalPosition == -1 ? null : new NodeSearcher(focalPosition), apiLevel, compilerOptions, parsedUnitWorkingCopyOwner, typeRootWorkingCopyOwner, flags, monitor);
+		}
+	}
+
+	public static final ECJCompilationUnitResolver FACADE = new ECJCompilationUnitResolver();
```
jukzi, line 123: better an ECJCompilationUnitResolver.INSTANCE that is only initialized on demand

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/CompilationUnitResolverDiscoveryTest.java
```diff
@@ -0,0 +1,184 @@
...
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
```
jukzi, line 12: Red Hat

org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/ICompilationUnitResolver.java
```diff
@@ -0,0 +1,113 @@
...
+	 * @param sourceFilePaths the compilation units to create ASTs for
+	 * @param encodings the given encoding for the source units
+	 * @param requestor the AST requester that collects abstract syntax trees and bindings
+	 * @param apiLevel Level of AST API desired.
+	 * @param compilerOptions Compiler options. Defaults to JavaCore.getOptions().
+	 * @param flags Flags to to be used during this operation
+	 * @param monitor A progress monitor
+	 */
+	void parse(String[] sourceFilePaths, String[] encodings, FileASTRequestor requestor, int apiLevel,
+			Map<String, String> compilerOptions, int flags, IProgressMonitor monitor);
```
jukzi, line 75: in other consumers it was stated that Map<String, String> i snot

robstryker: I'm sorry, this comment is unclear.

jukzi: sry for the broken comment. 
In https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2497/files#diff-e70e3008348d77cb89e082387f8a534ee23302c5b4ad4243f114b917e8aee21dR68 it was stated that map<string,string> is not enough and you would need CompilerOptions

robstryker: Thanks for the update.  

That's a different PR with a different part of the code, specifically the compiler, and not the converter from source to dom. 

We prefer to pass around the raw settings, almost exclusively fetched from `JavaCore.getOptions()`.   However, the default implementation will then use that Map to create a CompilerOptions, and so should anyone making use of this non-API functionality. 

The CompilerOptions class is internal, and we prefer to not use it in this interface. If we have hopes that this interface may one day become API, we prefer to avoid (wherever possible) using classes that would then also need to be made public. 

If an implementation wishes to take our raw settings / Map, and use it to create a CompilerOptions class or their own implementation thereof, then we are ok with that.  But making CompilerOptions part of the interface doesn't seem right to us. 

Part of the issue here is that (from the comments) it appears they were downgrading FROM CompilerOptions TO Map, whereas we are loading the map from the raw settings and allowing any clients to convert to CompilerOptions at their leisure. 

I'm open to alternatives, but I'm also about to go on vacation, and considering the workflow of this particular part of the code, I genuinely believe using the raw map from settings in the interface is appropriate, and clients can expand that to the more fully-featured CompilerOptions as they find convenient to do so.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
```diff
@@ -245,10 +248,13 @@ private List<Classpath> getClasspath() throws IllegalStateException {
 			}
 			if (this.sourcepaths != null) {
 				for (int i = 0, max = this.sourcepaths.length; i < max; i++) {
-					String encoding = this.sourcepathsEncodings == null ? null : this.sourcepathsEncodings[i];
-					main.processPathEntries(
-							Main.DEFAULT_SIZE_CLASSPATH,
-							allClasspaths, this.sourcepaths[i], encoding, true, false);
+					String sourcePath = this.sourcepaths[i];
+					if (sourcePath != null) {
+						String encoding = this.sourcepathsEncodings == null ? null : this.sourcepathsEncodings[i];
```
jukzi, line 253: Making the encoding part of the API seems wrong as currently the encoding is not used consistently: Only Batch Compiler uses it in org.eclipse.jdt.internal.compiler.batch.CompilationUnit.getContents() while the IDE uses org.eclipse.jdt.internal.core.util.Util.getResourceContentsAsCharArray(IFile) which disrespect the given classpath encoding but uses the encoding configured in workspace and the utf bom. You may want to harmonize that first.

jukzi (dismissed):
Please make sure that only copies or immutable data structures are passed to third party to prevent that changes by third party can make errors in jdt. IProgressMonitor is an exceptional case.
int is OK,
String is OK, 
Arrays are mutable -> copy or better use a immutable List. 
Map -> make a immutable copy
FileASTRequestor  is mutable -> i don't have solution for that at hand

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
mickaelistria: Fixing existing issues is going to make things much harder and longer on this review. We cannot afford fixing all pre-existing issues right now. Can we just open an issue about this `encoding` inconsistency in existing code and look at it later (including changing the new ICompilationUnitResolver interface if necessary)?

mickaelistria:
The FileASTRequestor is intentionally mutable. It's even expected that one of its field gets modified as part of the process. It kind of captures some of the state. I believe we won't be able to change that.
For other structure, I think we should go to copies over immutable. The issue with immutable is that the immutability doesn't surface in the API, so one can try to change them and get a crash at runtime, which is a worse behavior than changing something which doesn't have an effect outside of the change scope.

org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/ICompilationUnitResolver.java
mickaelistria: The compiler options are indeed the right API at compiler level, but from the Java model (ICompilationUnit and so on), it looks like the Map is more commonly used, so it's consistent to stick to it.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
jukzi: > Can we just open an issue about this `encoding` inconsistency

sure, just please extend the javadoc of the proposed provisional API with a reference to it so that it gets fixed before publishing such API.

mickaelistria: I've opened https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2641

robstryker:
> Please make sure that only copies or immutable data structures are passed to third party to prevent that changes by third party can make errors in jdt. IProgressMonitor is an exceptional case. int is OK, String is OK, Arrays are mutable -> copy or better use a immutable List. Map -> make a immutable copy FileASTRequestor is mutable -> i don't have solution for that at hand

I've updated most arrays to be cloned before sending them to the unitResolver.

mickaelistria:
Is anything else needed before this can be merged?

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolverDiscovery.java
```diff
@@ -0,0 +1,70 @@
...
+				for (final IConfigurationElement configElement : configElements) {
+					String elementId = configElement.getAttribute("id"); //$NON-NLS-1$
+					if( compilationUnitResolverId.equals(elementId)) {
+						String elementName =configElement.getName();
+						if (!("resolver".equals(elementName))) { //$NON-NLS-1$
+							continue;
+						}
+						try {
+							Object executableExtension = configElement.createExecutableExtension("class"); //$NON-NLS-1$
+							if( executableExtension instanceof ICompilationUnitResolver icur) {
```
testforstephen, line 44: inconsistent formatting.
 `if( executableExtension ...` -> `if (executableExtension ...`

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolverDiscovery.java
```diff
@@ -0,0 +1,70 @@
...
+								return icur;
+							}
+						} catch (CoreException e) {
+							if( !setErrorLogged()) {
```
testforstephen, line 48: formatting.

`if( !setErrorLogged())` -> `if (!setErrorLogged())`

org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/ICompilationUnitResolver.java
```diff
@@ -0,0 +1,123 @@
...
+	 * @param requestor the AST requestor that collects abstract syntax trees and bindings
+	 * @param apiLevel Level of AST API desired.
+	 * @param compilerOptions Compiler options. Defaults to JavaCore.getOptions().
+	 * @param flags Flags to to be used during this operation
+	 * @param monitor A progress monitor
+	 */
+	void parse(ICompilationUnit[] compilationUnits, ASTRequestor requestor, int apiLevel,
+			Map<String, String> compilerOptions, int flags, IProgressMonitor monitor);
+
+	/**
```
testforstephen, line 74: Only parameter doc, no doc for the method itself.

org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/ICompilationUnitResolver.java
```diff
@@ -0,0 +1,123 @@
...
+	 * @param apiLevel Level of AST API desired.
+	 * @param compilerOptions Compiler options. Defaults to JavaCore.getOptions().
+	 * @param flags Flags to to be used during this operation
+	 * @param monitor A progress monitor
+	 */
+	void parse(String[] sourceFilePaths, String[] encodings, FileASTRequestor requestor, int apiLevel,
+			Map<String, String> compilerOptions, int flags, IProgressMonitor monitor);
+
+	/**
+	 *
```
testforstephen, line 88: No doc for the method itself.

org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/ICompilationUnitResolver.java
```diff
@@ -0,0 +1,123 @@
...
+	 * @param classpathList A list of classpaths to use during this operation
+	 * @param flags Flags to to be used during this operation
+	 * @param monitor A progress monitor
+	 */
+	void resolve(String[] sourceFilePaths, String[] encodings, String[] bindingKeys, FileASTRequestor requestor,
+			int apiLevel, Map<String, String> compilerOptions, List<Classpath> classpathList, int flags,
+			IProgressMonitor monitor);
+
+	/**
+	 * Parse the given source paths with the following options.
```
testforstephen, line 62: -> Parse the ASTs for the given source units using the following options.

org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/ICompilationUnitResolver.java
```diff
@@ -0,0 +1,123 @@
...
+* </p>
+*
+* See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2641 for discussion on possible
+* changes to this interface
+*
+* @since 3.38
+ */
+public interface ICompilationUnitResolver {
+	/**
+	 * Resolve the given source paths with the following options.
```
testforstephen, line 45: As I understand it, 'Resolve' includes two steps: parse AST and resolve binding. Can you make the doc to be more specific about what it does?

-> Parse the ASTs and resolve the bindings for the given source files using the following options.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolver.java
```diff
@@ -76,12 +76,59 @@
...
+		public CompilationUnit toCompilationUnit(org.eclipse.jdt.internal.compiler.env.ICompilationUnit sourceUnit, final boolean initialNeedsToResolveBinding, IJavaProject project, List<Classpath> classpaths, int focalPosition,
+			int apiLevel, Map<String, String> compilerOptions, WorkingCopyOwner parsedUnitWorkingCopyOwner, WorkingCopyOwner typeRootWorkingCopyOwner, int flags, IProgressMonitor monitor) {
+			return CompilationUnitResolver.toCompilationUnit(sourceUnit, initialNeedsToResolveBinding, project,
+					classpaths, focalPosition == -1 ? null : new NodeSearcher(focalPosition), apiLevel, compilerOptions, parsedUnitWorkingCopyOwner, typeRootWorkingCopyOwner, flags, monitor);
+		}
+	}
+
+	private static ECJCompilationUnitResolver FACADE;
+	public static synchronized ICompilationUnitResolver getInstance() {
+		if( FACADE == null ) {
```
testforstephen, line 125: formatting issue.

`if( FACADE == null )` -> `if (FACADE == null)`

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolverDiscovery.java
```diff
@@ -0,0 +1,70 @@
...
+		return CompilationUnitResolver.getInstance();
+	}
+
+	/*
+	 * Set the ERROR_LOGGED field to true.
+	 * Return the previous value of ERROR_LOGGED.
+	 */
+	private static synchronized boolean setErrorLogged() {
+		boolean prev = ERROR_LOGGED;
+		if( !ERROR_LOGGED) {
```
testforstephen, line 65: formatting issue.

testforstephen (commented):
It has some formatting issues, others look good to me.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
```diff
@@ -957,15 +965,23 @@ public void createASTs(ICompilationUnit[] compilationUnits, String[] bindingKeys
...
 			} else {
-				CompilationUnitResolver.parse(compilationUnits, requestor, this.apiLevel, this.compilerOptions, flags, monitor);
+				this.unitResolver.parse(safeCopyOf(compilationUnits), requestor, this.apiLevel, safeUnmodifiableMap(this.compilerOptions), flags, monitor);
 			}
 		} finally {
 			// reset to defaults to allow reuse (and avoid leaking)
 			initializeDefaults();
 		}
 	}
+    @SuppressWarnings("unchecked")
```
jukzi, line 977: fix all warnings

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolverDiscovery.java
```diff
@@ -0,0 +1,70 @@
...
+	static ICompilationUnitResolver getInstance() {
+		String compilationUnitResolverId = System.getProperty(SELECTED_SYSPROP);
+		IExtensionPoint extension = Platform.getExtensionRegistry().getExtensionPoint(JavaCore.PLUGIN_ID, COMPILATION_UNIT_RESOLVER_EXTPOINT_ID);
+		if (extension != null && compilationUnitResolverId != null && !compilationUnitResolverId.isEmpty()) {
+			IExtension[] extensions = extension.getExtensions();
+			for (IExtension ext : extensions) {
+				IConfigurationElement[] configElements = ext.getConfigurationElements();
+				for (final IConfigurationElement configElement : configElements) {
+					String elementId = configElement.getAttribute("id"); //$NON-NLS-1$
+					if( compilationUnitResolverId.equals(elementId)) {
```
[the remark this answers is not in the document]
mickaelistria: should be fixed now.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
```diff
@@ -1229,68 +1244,28 @@ private ASTNode internalCreateASTCached(IProgressMonitor monitor) {
...
-						wcOwner,
-						needToResolveBindings ? new DefaultBindingResolver.BindingTables() : null,
-						flags,
-						monitor,
-						this.project != null,
-						this.project);
-					result.setTypeRoot(this.typeRoot);
-					return result;
+
+					CompilationUnit result2 = this.unitResolver.toCompilationUnit(sourceUnit, needToResolveBindings, this.project, getClasspath(), useSearcher ? this.focalPointPosition : -1, this.apiLevel, safeUnmodifiableMap(this.compilerOptions), this.workingCopyOwner, wcOwner, flags, monitor);
```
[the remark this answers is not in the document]
mickaelistria: should be fixed now.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolverDiscovery.java
mickaelistria: should be fixed now.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolverDiscovery.java
mickaelistria: should be fixed now.

org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/ICompilationUnitResolver.java
mickaelistria: should be fixed now.

org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/ICompilationUnitResolver.java
mickaelistria: should be fixed now (by the way, don't hesitate to use the "suggestion" block for such proposal, so we can just Apply it directly from GitHub UI)

org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/ICompilationUnitResolver.java
mickaelistria: should be fixed now.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolver.java
mickaelistria: should be fixed now.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
mickaelistria: should be fixed now.

mickaelistria:
Unless I missed one, all former review comments are now covered.

testforstephen (approved):
LGTM

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/CompilationUnitResolverDiscoveryTest.java
```diff
@@ -0,0 +1,181 @@
...
+import org.eclipse.jdt.core.dom.MethodDeclaration;
+import org.eclipse.jdt.core.dom.PackageDeclaration;
+import org.eclipse.jdt.internal.compiler.batch.FileSystem.Classpath;
+import org.eclipse.jdt.internal.compiler.classfmt.ClassFileConstants;
+import org.eclipse.jdt.internal.compiler.impl.CompilerOptions;
+import org.eclipse.jdt.internal.core.dom.ICompilationUnitResolver;
+
+import junit.framework.Test;
+
+public class CompilationUnitResolverDiscoveryTest extends ConverterTestSetup {
```
iloveeclipse, line 40: The test should be added to a test suite, otherwise it is not executed

iloveeclipse (changes requested):

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolverDiscovery.java
```diff
@@ -0,0 +1,68 @@
...
+	private static final String COMPILATION_UNIT_RESOLVER_EXTPOINT_ID = "compilationUnitResolver" ; //$NON-NLS-1$
+	private static boolean ERROR_LOGGED = false;
+
+
+	static ICompilationUnitResolver getInstance() {
+		String compilationUnitResolverId = System.getProperty(SELECTED_SYSPROP);
+		IExtensionPoint extension = Platform.getExtensionRegistry().getExtensionPoint(JavaCore.PLUGIN_ID, COMPILATION_UNIT_RESOLVER_EXTPOINT_ID);
+		if (extension != null && compilationUnitResolverId != null && !compilationUnitResolverId.isEmpty()) {
+			IExtension[] extensions = extension.getExtensions();
+			for (IExtension ext : extensions) {
```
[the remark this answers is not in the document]
mickaelistria: The system property can change during the lifecycle of the IDE, so we may need to consider a different extension when this happens. And we do really prefer instantiating 1 new object one every query instead of using a singleton as this can allow to keep state information in the class.
I'm going to make a comment about it.

Note that the usual pattern is 1 reason why we have to restart Eclipse after installing/uninstalling plugins. If all consumers were re-reading the extension points on demand (or better if they were reacting to changes in extension registry), then we wouldn't have to restart the IDE for changes to take effect.
I'll try to implement a cache so that we keep the "current" extension until the system property change.

mickaelistria:
@iloveeclipse I appended a commit that should cover all your comments.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolverDiscovery.java
```diff
@@ -0,0 +1,91 @@
...
+	private static final String SELECTED_SYSPROP = "ICompilationUnitResolver"; //$NON-NLS-1$
+	private static final String COMPILATION_UNIT_RESOLVER_EXTPOINT_ID = "compilationUnitResolver" ; //$NON-NLS-1$
+	private static boolean ERROR_LOGGED = false;
+
+	private static String lastId;
+	private static IConfigurationElement lastExtension;
+
+	static ICompilationUnitResolver getInstance() {
+		String id = System.getProperty(SELECTED_SYSPROP);
+		IConfigurationElement configElement = getConfigurationElement(id);
```
iloveeclipse, line 34: As before, can we please avoid the code looking into the OSGI registry if system property is not set, to avoid that on every call we enter sync blocks and do extra work?

What I want is to make the "default" SDK case where property is **not** set as fast as it was before this change, and that is easily done by checking system property first.

iloveeclipse (changes requested):

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolverDiscovery.java
mickaelistria: Please see the 1st lines of `getConfigurationElement(id)` method, they immediately return null when there is no setting, so extension registry is not even read.

mickaelistria:
@iloveeclipse I think all your review comments were addressed, can you please have another look?

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolverDiscovery.java
iloveeclipse: OK, this is a bit surprising, but I see it now.

iloveeclipse (approved):
All concerns cleared.

akurtakov:
Are we waiting for someone else to review?

akurtakov (approved):
Unless there is some concern I plan to merge this one on Monday.

akurtakov:
3 committers approved the patch - merging.

mickaelistria:
Thanks to everyone involved!

PR #2570, opened 2024-06-14
Use new IFile.write() to create or replace files

jukzi:


jukzi:
I plan to submit this tomorrow after fixing the build problem. If there are any concerns (xtext?) please state them timely @szarnekow @LorenzoBettini @cdietrich @merks

org.eclipse.jdt.apt.core/src/org/eclipse/jdt/apt/core/internal/env/BinaryFileOutputStream.java
```diff
@@ -46,40 +43,20 @@ public BinaryFileOutputStream(IFile file, BuildEnv env) {
...
-				finally {
-					closeInputStream(in);
-					closeInputStream(oldData);
-				}
-				if (contentsChanged) {
-					contents.reset();
-					saveToDisk(contents, false);
-				}
+			// Only write the contents if the data is different
+			byte[] oldContent = _file.readAllBytes();
```
merks, line 50: I suppose it's a trade-off to avoid an exception by checking `_file.exists()` versus the extra cost of the test itself for the very common case that the target already exists, right?

org.eclipse.jdt.apt.pluggable.core/src/org/eclipse/jdt/internal/apt/pluggable/core/filer/IdeNonSourceOutputStream.java
```diff
@@ -48,40 +45,20 @@ public IdeNonSourceOutputStream(IdeProcessingEnvImpl env, IFile file, Collection
...
-				}
-				finally {
-					closeInputStream(in);
-					closeInputStream(oldData);
-				}
-				if (contentsChanged) {
-					contents.reset();
-					saveToDisk(contents, false);
-				}
+			// Only write the contents if the data is different
```
merks, line 51: The fact this is a copy of the previous code is an existing issue.

merks (commented):
I'm not so qualified to review this.

I guess somewhere else there are MANIFEST.MF changes to indicate that there is a dependency on the latest resources API now.

jukzi:
> I guess somewhere else there are MANIFEST.MF changes to indicate that there is a dependency on the latest resources API now.

org.eclipse.jdt.apt.core/src/org/eclipse/jdt/apt/core/internal/env/BinaryFileOutputStream.java
jukzi: org.eclipse.jdt.internal.apt.pluggable.core.filer.IdeNonSourceOutputStream,
org.eclipse.jdt.internal.apt.pluggable.core.filer.IdeClassOutputStream,
org.eclipse.jdt.apt.core.internal.env.BinaryFileOutputStream,

are now refactored to use common FileSystemUtil.saveToDisk(IFile, byte[])

jukzi: > I suppose it's a trade-off to avoid an exception by checking _file.exists() versus the extra cost of the test itself for the very common case that the target already exists, right?

yes, readAllBytes(); will throw CoreException if the file does not exists, which will result in the common case of saveToDisk because contentsChanged. so there is not need for an extra fie.exists() check.

org.eclipse.jdt.apt.pluggable.core/src/org/eclipse/jdt/internal/apt/pluggable/core/filer/IdeNonSourceOutputStream.java
jukzi: org.eclipse.jdt.internal.apt.pluggable.core.filer.IdeNonSourceOutputStream,
org.eclipse.jdt.internal.apt.pluggable.core.filer.IdeClassOutputStream,
org.eclipse.jdt.apt.core.internal.env.BinaryFileOutputStream,

are now refactored to use common FileSystemUtil.saveToDisk(IFile, byte[])

Issue #2571, opened 2024-06-14
Show declaration (F3) & Mouseover Popups not working


DanskerDave:
One line of code causes F3 & Mouseover popups in the Java Editor to stop working.

Here's an example.

```
package org.eclipse;

import java.util.function.Predicate;
import java.util.function.Supplier;

public class PopupKiller {

	public final Supplier<Predicate<String>> predicateSupplier = () -> new Predicate<>() {

		@SuppressWarnings("unused")
		public static boolean THIS_VARIABLE_KILLS_JAVADOC_POPUPS_AND_F3 = true;

		@Override
		public boolean test(final String record) {

			System.out.println("<-- try to Mouseover System.out.: nothing!");

			return true;
		}
	};
}

```

Version.: newly downloaded Eclipse 2024-06 RCP/RAP
OS.: Windows 11

stephan-herrmann:
Thanks, I could observe the following exception, which is silently swallowed unless `SelectionEngine.DEBUG` is set:
```
java.lang.ArrayIndexOutOfBoundsException: Index -1 out of bounds for length 30
	at org.eclipse.jdt.internal.codeassist.impl.AssistParser.commit(AssistParser.java:2334)
	at org.eclipse.jdt.internal.codeassist.impl.AssistParser.triggerRecoveryUponLambdaClosure(AssistParser.java:596)
	at org.eclipse.jdt.internal.codeassist.impl.AssistParser.consumeFieldDeclaration(AssistParser.java:736)
	at org.eclipse.jdt.internal.compiler.parser.Parser.consumeRule(Parser.java:7026)
	at org.eclipse.jdt.internal.compiler.parser.Parser.parse(Parser.java:12707)
	at org.eclipse.jdt.internal.compiler.parser.Parser.parse(Parser.java:12965)
	at org.eclipse.jdt.internal.codeassist.select.SelectionParser.parse(SelectionParser.java:1910)
```
(I happened to be in a workspace with sources as of some time in April this year).

Scanner state at that point is
```
----------------Scanner--------------
package org.eclipse;

import java.util.function.Predicate;
import java.util.function.Supplier;

public class PopupKiller {

	public final Supplier<Predicate<String>> predicateSupplier = () -> new Predicate<>() {

		@SuppressWarnings("unused")
		public static final boolean THIS_VARIABLE_KILLS_JAVADOC_POPUPS_AND_F3 = true;

		
===============================
Starts here -->@<-- Ends here
[...]

PR #2573, opened 2024-06-14
Show declaration (F3) & Mouseover Popups not working

stephan-herrmann:
Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2571

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ResolveTests18.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ResolveTests18.java
@@ -14,8 +14,6 @@
 
 package org.eclipse.jdt.core.tests.model;
 
-import junit.framework.Test;
-
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.jdt.core.BindingKey;
 import org.eclipse.jdt.core.ICodeAssist;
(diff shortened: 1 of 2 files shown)


stephan-herrmann:
The culprit was quickly found, but finding a good remedy was not so clear, because first attempts (which looked straight-forward) created regressions elsewhere.

In the current issue, we "messed" with an outer lambda while parsing an anonymous class within the lambda, so I figured it's clearly not the time to even consider the outer lambda.

A simple `break` caused a regression in `CompletionTests.testBug421469a` which I could fix by changing `break` to `return false;` as to avoid the (premature?) `popElement()` at the bottom of the method.

At this point `CompletionParserTest18.test446765` was unwell, in that we dropped some intermediate structure:
Given this:
```
new Stepper<Void>(r -> {}) {
            private void step1() {
                Y y = new Y();
                y.request(response -> {
                    if (response.)
                });
            }
        }.run(); 
```
we expect a parent completion node of `if (<CompleteOnName:response.>)` but the if statement does not show up anywhere. Apparently, there is some interaction even with the anonymous class containing lambdas in two independent locations.

I could experimentally repair this by saying:
```diff
		if (expression.sourceStart >= statementStart && expression.sourceEnd <= statementEnd) {
			this.elementPtr = i - 1;
			lambdaClosed = true;
+			break;
		} else {
```

but then `CompletionTests18.testBug577883_expectCompletions_onOuterLambdaVars_inNestedLambdas` no longer gives any completions.

This takes us back to https://git.eclipse.org/r/c/jdt/eclipse.jdt.core/+/188986 where I mentioned to @gayanper 
> I'm not able to 100% explain why it works. I did observe that the effect is by pushing the automaton into more parsing after detecting an error, and only this parsing will eventually trigger attachOrphanCompletionNode() which is necessary to build the RecoveredBlock containing all the lambda magic.
>
> I'll mention this to Srikanth when I speak to him next time, maybe he can enlighten me more about the concept of committing a snapshot 😊

@srikanth-sankaran  can you give a few words of what is the concept of "committing a snapshot"?

As my third attempt, I'll avoid any `break;` or `return false;` but only skip the final `popSnapShot()` when inside a type declaration. Wish me luck :)

stephan-herrmann:
@srikanth-sankaran  I merged this because it is clear that seeing K_TYPE_DELIMITER on the element stack needs to be handled specifically (otherwise stack imbalance occurred). What exact steps should and should not happen in that case was determined heuristically (based on test regressions). If you have any thoughts to put this on a more solid foundation I'm all ears :)

PR #2596, opened 2024-06-18
org.eclipse.jdt.internal.core.Buffer: threadsafe

jukzi:
Some field accesses had not been synchronized which may have caused NPEs in access from different thread.

https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/736


jukzi:
@stephan-herrmann i have found some potential concurrency issues in Buffer.

Issue #2536, opened 2024-06-07
Drop support of older Java releases in ecj


iloveeclipse:
### Discussed in https://github.com/eclipse-jdt/eclipse.jdt.core/discussions/820

<div type='discussions-op-text'>

<sup>Originally posted by **iloveeclipse** March  9, 2023</sup>
Stephan [asked good question](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/809#issuecomment-1461692502), I'm not sure we've discussed that before.

The question is: 

> * Did we discuss dropping support for some ancient Java versions? What's the status there?

and my take on this is: if we can make our life easier, we should follow what javac does, and it does what was proposed in [JEP-182](https://openjdk.org/jeps/182): with each Java LTS release they seem to drop support for one of previous releases.

See 
- https://openjdk.org/jeps/182
- https://bugs.openjdk.org/browse/JDK-8046172
- https://bugs.openjdk.org/browse/JDK-8173605

Current state in Java 17 (still same in 19):
```
/usr/lib/jvm/java-17/bin/javac --help | grep release
        Enable preview language features. To be used in conjunction with either -source or --release.
  --release <release>
        Compile for the specified Java SE release. Supported releases: 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17
  --source <release>, -source <release>
        Provide source compatibility with the specified Java SE release. Supported releases: 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17
  --target <release>, -target <release>
        Generate class files suitable for the specified Java SE release. Supported releases: 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17
```

for Java 21 we have 

```
  --release <release>
        Compile for the specified Java SE release.
        Supported releases: 
            8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21
  -s <directory>               Specify where to place generated source files
  --source <release>, -source <release>
        Provide source compatibility with the specified Java SE release.
        Supported releases: 
            8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21
  --source-path <path>, -sourcepath <path>
        Specify where to find input source files
  --system <jdk>|none          Override location of system modules
  --target <release>, -target <release>
        Generate class files suitable for the specified Java SE release.
        Supported releases: 
[...]

PR #2606, opened 2024-06-19
Add new API for supported/unsupported Java versions

iloveeclipse:
- `List<String> getAllJavaProjectVersions()` - all Java versions that could be used for Java projects inside Eclipse. The difference to existing `getAllVersions()` API is that later one knows almost all Java versions ever released and might be used not only in JDT core but also in debugger/PDE area.
- `boolean isSupportedJavaProjectVersion(String version)` - differs from existing `isSupportedJavaVersion()` in the same way as explained above
- `String getFirstSupportedJavaVersion()` - similar to existing `latestSupportedJavaVersion()` and should return minimal "default" version supported by JDT.

API above will be used in JDT UI, Debug (PR's are following), and (most likely) PDE.

**Internal** API added in batch compiler `CompilerOptions`:
- `getFirstSupportedJavaVersion()`
- `getFirstSupportedJdkLevel()`

Bumped minor version segments on both core and compiler, even if compiler "only" provides internal API, it is more convenient to have them in-sync.

Related UI/Debugger changes that use new API are:

- https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/452
- https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1469

See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2536

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
@@ -3322,9 +3322,27 @@ public final class JavaCore extends Plugin {
 	 * @category OptionValue
 	 */
 	public static final String VERSION_CLDC_1_1 = "cldc1.1"; //$NON-NLS-1$
-	private static List<String> allVersions = Collections.unmodifiableList(Arrays.asList(VERSION_CLDC_1_1, VERSION_1_1, VERSION_1_2, VERSION_1_3, VERSION_1_4, VERSION_1_5,
+	private static final List<String> allVersions = Collections.unmodifiableList(Arrays.asList(VERSION_CLDC_1_1, VERSION_1_1, VERSION_1_2, VERSION_1_3, VERSION_1_4, VERSION_1_5,
 			VERSION_1_6, VERSION_1_7, VERSION_1_8, VERSION_9, VERSION_10, VERSION_11, VERSION_12, VERSION_13, VERSION_14, VERSION_15, VERSION_16, VERSION_17, VERSION_18,
 			VERSION_19, VERSION_20, VERSION_21, VERSION_22));
+
+	/**
+	 * Unordered set of all Java source versions <b>not supported</b> by compiler anymore.
+	 * The values are from {@link JavaCore}{@code #VERSION_*}.
+	 */
+	private static final Set<String> UNSUPPORTED_VERSIONS = CompilerOptions.UNSUPPORTED_VERSIONS;
+
+	/**
+	 * Ordered set (from oldest to latest) of all Java source versions <b>supported</b> by compiler.
+	 * The values are from {@link JavaCore}{@code #VERSION_*}.
+	 */
+	private static final List<String> SUPPORTED_VERSIONS;
+	static {
+		ArrayList<String> temp = new ArrayList<>();
+		temp.addAll(allVersions);
+		temp.removeAll(UNSUPPORTED_VERSIONS);
+		SUPPORTED_VERSIONS = Collections.unmodifiableList(temp);
+	}
 
 	/**
 	 * Returns all {@link JavaCore}{@code #VERSION_*} levels in the order of their
@@ -3335,6 +3353,22 @@ public final class JavaCore extends Plugin {
 	 */
 	public static List<String> getAllVersions() {
 		return allVersions;
+	}
+
+	/**
+	 * Returns all Java source versions fully supported by Eclipse compiler in the order of their introduction. For
+	 * example, {@link JavaCore#VERSION_1_8} appears before {@link JavaCore#VERSION_10}.
+	 * <p>
+	 * Note, some not included older or newer Java versions might be known by Eclipse compiler internally but not
+	 * exposed via this API because compiler does not support these anymore (or yet).
+	 *
+	 * @return all Java source versions fully supported by Eclipse compiler
+	 * @see #isJavaSourceVersionSupportedByCompiler(String)
+	 * @see #getFirstJavaSourceVersionSupportedByCompiler()
+	 * @since 3.39
+	 */
+	public static List<String> getAllJavaSourceVersionsSupportedByCompiler() {
+		return SUPPORTED_VERSIONS;
 	}
 
 	/**
@@ -3350,6 +3384,22 @@ public final class JavaCore extends Plugin {
 	 */
 	public static boolean isSupportedJavaVersion(String version) {
 		return CompilerOptions.versionToJdkLevel(version, false) > 0;
+	}
+
+	/**
+	 * Not all known Java versions are supported by Eclipse compiler. This method answers if the given Java source
+	 * version is fully supported.
+	 *
+	 * @return {@code true} if the given string represents Java language version is fully supported by Eclipse compiler
+	 * @see #getAllJavaSourceVersionsSupportedByCompiler()
+	 * @see #getFirstJavaSourceVersionSupportedByCompiler()
+	 * @since 3.39
+	 */
+	public static boolean isJavaSourceVersionSupportedByCompiler(String version) {
+		if(version == null || version.isBlank()) {
+			return false;
+		}
+		return SUPPORTED_VERSIONS.contains(version);
 	}
 
 	/**
@@ -6453,6 +6503,20 @@ public final class JavaCore extends Plugin {
 	public static String latestSupportedJavaVersion() {
 		return allVersions.get(allVersions.size() - 1);
 	}
+
+	/**
+	 * First (oldest) Java source version supported by the Eclipse compiler.
+	 * This is the first entry from {@link JavaCore#getAllJavaSourceVersionsSupportedByCompiler()}.
+	 *
+	 * @return first (oldest) Java source version supported by the Eclipse compiler
+	 * @see #getAllJavaSourceVersionsSupportedByCompiler()
+	 * @see #isJavaSourceVersionSupportedByCompiler(String)
+	 * @since 3.39
+	 */
+	public static String getFirstJavaSourceVersionSupportedByCompiler() {
+		return SUPPORTED_VERSIONS.get(0);
+	}
+
 	/**
 	 * Compares two given versions of the Java platform. The versions being compared must both be
 	 * one of the supported values mentioned in
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/impl/CompilerOptions.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/impl/CompilerOptions.java
@@ -35,6 +35,7 @@ import java.io.InputStreamReader;
 import java.io.UnsupportedEncodingException;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Set;
 
 import org.eclipse.jdt.core.compiler.CharOperation;
 import org.eclipse.jdt.internal.compiler.Compiler;
@@ -258,6 +259,25 @@ public class CompilerOptions {
 	 * Note: Whenever a new version is added, make sure getLatestVersion()
 	 * is updated with it.
 	 */
+
+	/**
+	 * Unsupported JLS versions
+	 */
+	/*
+	 * Note: Whenever a new version is obsoleted, make sure getFirstVersion() is updated.
+	 */
+	public static Set<String> UNSUPPORTED_VERSIONS = Set.of(
+			VERSION_1_1,
+			VERSION_1_2,
+			VERSION_1_3,
+			VERSION_1_4,
+			VERSION_JSR14,
+			VERSION_CLDC1_1,
+			VERSION_1_5,
+			VERSION_1_6,
+			VERSION_1_7
+			);
+
 	public static final String ERROR = "error"; //$NON-NLS-1$
 	public static final String WARNING = "warning"; //$NON-NLS-1$
 	public static final String INFO = "info"; //$NON-NLS-1$
@@ -647,6 +667,18 @@ public class CompilerOptions {
 		this.parseLiteralExpressionsAsConstants = parseLiteralExpressionsAsConstants;
 	}
 
+	/**
+	 * Return the first (oldest) Java language version supported by the Eclipse compiler
+	 */
+	public static String getFirstSupportedJavaVersion() {
+		return VERSION_1_8;
+	}
+	/**
+	 * Return the first (oldest) Java language level supported by the Eclipse compiler
+	 */
+	public static long getFirstSupportedJdkLevel() {
+		return ClassFileConstants.JDK1_8;
+	}
 	/**
 	 * Return the latest Java language version supported by the Eclipse compiler
 	 */


iloveeclipse:
@jarthana, @stephan-herrmann, @srikanth-sankaran : could you please check new API? Does it make sense for you?

Related UI/Debugger changes that use new API are:

- https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/452
- https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1469

iloveeclipse:
The doubts I have is why to have two "supported" sets of Java versions in the API, which seem to be awkward.

The initial idea that with this PDE (and whoever else) could still use "old" API with all possible Java / EE versions, while JDT UI only sees "new" Java versions.

The negative part of it is extra (kind of duplicated) API and that "whoever else" would need code change if the code wanted to see "fully" supported Java versions.

I'm currently thinking about **not** adding `getAllJavaProjectVersions()`  and `isSupportedJavaProjectVersion(String version)` but **changing behavior** / javadoc of `getAllVersions()`  and `isSupportedJavaVersion(String version)` to point out that it doesn't supply **ALL** of `{@link JavaCore}{@code #VERSION_*} levels` as stated today. This would be breaking change and that is my dilemma.

iloveeclipse:
Another interesting use case for the new `getAllJavaProjectVersions()` API would be this issue: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1448.

The problem there is that in 4.32 we've introduced `JavaCore.VERSION_23` which is **not supported** by 4.32 at all! So it shouldn't appear in `getAllVersions()` and should report `false` in `isSupportedJavaVersion(String version)` in 4.32.

With the API proposed here we would say: yep, `JavaCore.VERSION_23` is known by JDT Core but is not supported for anything visible in UI, debugger etc, all places that involve projects etc.

So assuming JDT Core would continue to add Java versions (for JDT LS) that aren't fully supported yet by JDT UI, the new APi would make sense.

@akurtakov, @mickaelistria : WDYT?

akurtakov:
> Another interesting use case for the new `getAllJavaProjectVersions()` API would be this issue: [eclipse-jdt/eclipse.jdt.ui#1448](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1448).
> 
> The problem there is that in 4.32 we've introduced `JavaCore.VERSION_23` which is **not supported** by 4.32 at all! So it shouldn't appear in `getAllVersions()` and should report `false` in `isSupportedJavaVersion(String version)` in 4.32.

+1

> 
> With the API proposed here we would say: yep, `JavaCore.VERSION_23` is known by JDT Core but is not supported for anything visible in UI, debugger etc, all places that involve projects etc.

The reason to have (https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/426 https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/429) is so projects can declare they require newer Java EE for the sake of new APIs being added and not for being able to target this bytecode level. So some visibility in projects is required - a prototype how to handle the case of EE being newer than target level compiler supports is at https://github.com/akurtakov/eclipse.jdt.ui/commit/7a2ac1b9b2010b2a76052c42dec676749a5d5716.

mickaelistria:
This is all a corollary of the deep question "what does JDT support Java N mean ?": projects targetting this version can load and be developed, ECJ can compile to this version, all features of this versions are supported in JDT Core (ie DOM), all features of this version are supported in UI...

I think your proposal still assumes that JavaCore is only capable of dealing with what ECJ supports, while in practice things such as rewrite could allow to manipulate some different versions (older or newer) than what ECJ supports.
If we're to allow other parser/compiler than ECJ, then we cannot statically know what versions are supported. In theory, I think JDT should simply state the minimal version it's capable to handle in its AST, and everything else such as "this version is not supported" should be produced by the compiler and only surface as errors on the file. That would dis-couple further from ECJ, allowing AST to evolve ahead of ECJ, reduce maintenance when moving forward as less things would have to change. As for Java versions in the future, I'm even wondering whether it makes sense to have a list of them.

I know my answer is not helping on short-term, it's more food for thought, sorry for highjacking the PR.
So more precisely about the PR, I'm more in a "whatever" opinion at the moment, since the deeper thought presented make that I cannot even define what "supported" mean so I cannot build an opinion about such change.

iloveeclipse:
> In theory, I think JDT should simply state the minimal version it's capable to handle in its AST

AST is not the only part of JDT that needs Java version and actually AST itself has its own set of versions it supports.

> and everything else such as "this version is not supported" should be produced by the compiler and only surface as errors on the file.

This is too simplistic view and against IDE concept of guiding / helping user. We have wizards and preference pages that **do not** need to run compiler to detect that project options are incompatible with supported Java.

> If we're to allow other parser/compiler than ECJ, then we cannot statically know what versions are supported.

Not quite correct, we now that as IDE we can't support more as we have coded in various wizards/preference pages etc. 

This "other parser/compiler" story is of very disruptive to the static nature of `JavaCore` and `CompilerOptions` constants / methods, but so far this part doesn't surface in the "classic" IDE at all. The `JavaCore` constants/methods were developed with a monolithic Java IDE in mind, now the "Java IDE" can internally be decomposed into ("ecj" vs "alternative compiler") + ("JDT Core AST" vs "alternative AST") + ("JDT UI " vs "Java LS  / whatever").

Ideally "what is supported" could be also queried per project and derived from the intersection of particular IDE capabilities and project options like "use alternative compiler".

In ideal world, IDE as a whole should be able to know which Java (JLS, runtime) is supported, in which use case, scope and to which degree. This is required to avoid obviously surprising issues like https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1448 where left hand doesn't know what the right hand does. So even if we know that there will be Java 24, 25 etc, it doesn't make sense to "allow" these versions to appear anywhere in the UI or preferences because we can't really support them with our current version.

So far we only had one set of options / versions we could support, and they didn't really distinguish between compiler, core, ast, UI, debugger etc. Everyone referenced `JavaCore` versions and in few cases we had special handling for "special" versions. Therefore my proposal here: let continue assume `JavaCore` would still serve the "monolithic" IDE purpose as before, and look at this PR from this point of view.

jarthana:
Note: I am bumping the versions for jdt.core and compiler.batch (in advance) before the merge onto BETA_JAVA23. This is being done via https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2615

iloveeclipse:
> Note: I am bumping the versions for jdt.core and compiler.batch (in advance) before the merge onto BETA_JAVA23. This is being done via #2615

OK. @jarthana : could you please check comments here and give your opinion? [I'm in doubt if that is right way to do or not](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2606#issuecomment-2178466102).

org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
```diff
@@ -3322,10 +3322,28 @@ public final class JavaCore extends Plugin {
 	 * @category OptionValue
 	 */
 	public static final String VERSION_CLDC_1_1 = "cldc1.1"; //$NON-NLS-1$
-	private static List<String> allVersions = Collections.unmodifiableList(Arrays.asList(VERSION_CLDC_1_1, VERSION_1_1, VERSION_1_2, VERSION_1_3, VERSION_1_4, VERSION_1_5,
+	private static final List<String> allVersions = Collections.unmodifiableList(Arrays.asList(VERSION_CLDC_1_1, VERSION_1_1, VERSION_1_2, VERSION_1_3, VERSION_1_4, VERSION_1_5,
```
jukzi, line 3325: The modern version is  `List.of(...)`
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/List.html#of(E...)

szarnekow: True. But it shall be noted that the returned List implements an extremely rigor null-contract. It'll throw an NPE when asked whether it `contains(null)` instead of just returning false. I learned the hard way that this *can* be quite  painful.

srikanth-sankaran:
> > Note: I am bumping the versions for jdt.core and compiler.batch (in advance) before the merge onto BETA_JAVA23. This is being done via #2615
> 
> OK. @jarthana : could you please check comments here and give your opinion? [I'm in doubt if that is right way to do or not](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2606#issuecomment-2178466102).

@jarthana and I reviewed this together and discussed this in detail and I am capturing our comments.

> The doubts I have is why to have two "supported" sets of Java versions in the API, which seem to be awkward.

The present effort with https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2536 is **only** to drop support for **compiling** Java source files at levels 1.7 and below. It is obviously not the case that **all** support is dropped for those Java levels 1.7 and below. We must still be able to consume binary projects, class files, jars and such artifacts produced at levels not supported by compiler anymore. Various non-compilation operations (search, hierarchy, model construction etc) should span the scope of these binary artifacts produced at levels not supported for compilation.

As a result, it is best not to modify the semantics and operations of existing API points, but introduce additional new ones that may satisfy the needs of any and all interested parties.

However, the present choice of names can be a source of confusion. Here are suggested alternatives that may be considered:

Rather than ` JavaCore.getAllJavaProjectVersions()`  how about `JavaCore.getAllJavaVersionsSupportedAtSource()` or alternatively `JavaCore.getAllJavaSourceVersionsSupportedByCompiler()`

Likewise, instead of `JavaCore.isSupportedJavaProjectVersion(String version)` how about `JavaCore.isJavaVersionSupportedAtSource(String version)` or alternatively `JavaCore.isJavaSourceVersionSupportedByCompiler(String version)`


Likewise instead of `JavaCore.getFirstSupportedJavaVersion()` how about `JavaCore.getFirstJavaVersionSupportedAtSource()` or `JavaCore.getFirstJavaSourceVersionSupportedByCompiler()`

org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
iloveeclipse: Thanks Sebastian, will leave "as is".

iloveeclipse:
Thanks Srikanth & Jay. I've checked the usage of these methods and I agree that the "compiler" related method names fit better:
```
JavaCore.getAllJavaSourceVersionsSupportedByCompiler()
JavaCore.isJavaSourceVersionSupportedByCompiler(String version)
JavaCore.getFirstJavaSourceVersionSupportedByCompiler()
```
Here are callers and it makes sense looking who consumes all that:
image

I will proceed with that for now.

org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
```diff
@@ -3337,6 +3355,22 @@ public static List<String> getAllVersions() {
...
+	 * <p>
+	 * Note, some not included older or newer Java versions might be known by Eclipse compiler internally but not
+	 * exposed via this API because compiler does not support these anymore (or yet).
+	 *
+	 * @return all Java source versions fully supported by Eclipse compiler
+	 * @see #isJavaSourceVersionSupportedByCompiler(String)
+	 * @see #getFirstJavaSourceVersionSupportedByCompiler()
+	 * @since 3.39
+	 */
+	public static List<String> getAllJavaSourceVersionsSupportedByCompiler() {
```
HannesWell, line 3370: Have you considered to use a `SequencedSet<String>` here?
As far as I can see the uses of this method either fetch the first or last element or test if a String is contained in the list. The latter would be faster if a `LinkedHashSet` were used, which implements `SequencedSet`. That interface also provide `getFirst()` and `getLast()` and therefore seems to be an ideal fit for this use-case.

The only 'problem' is that `SequencedSet` was only added in Java-21. But since SimRel permits Java-21 now this might be a good opportunity to move JDT (at least parts of it) to Java-21 now instead of waiting for a few months more and having a List-based API forever?

iloveeclipse: Yes, I considered, but I didn't wanted to change JRE requirement because of that.

HannesWell: OK, thanks.
That's relatable, but IMHO one more reason pro moving the SDK to Java-21 in the SDK.

HannesWell: By chance I was remembered that we already have `SortedSet` (which inherits from `SequencedSet` in Java-21) and I think that would be a good fit as well:
Please have a look at https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2770

laeubi:
May I ask If something was changed in PDE as part of this? Or should change? As we again seem to face issues with an "unuspported" Java Version in PDE again.

iloveeclipse:
Tests are green again, it is just https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/746 that was missing on master.

laeubi:
> Tests are green again, it is just [eclipse-jdt/eclipse.jdt.debug#746](https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/746) that was missing on master.

Looking at the code, it seems to check each java version and then if nothing is found return Java 1.3 (!). As it seems an recurring problem here with new java versions, would it not be better to reverse that check to always return the _HIGHEST_ as default if nothing is found instead of the _LOWEST_? Or even throw an exception?

In Tycho we even simply use `JavaSE-(\d+)` as a pattern to extract the java version from the EE String as it does not change often, as it here even only checks for the number being part of it maybe one can even use `(\d+)` as a pattern.

iloveeclipse:
> Looking at the code, it seems to check each java version and then if nothing is found return Java 1.3 (!). As it seems an recurring problem here with new java versions, would it not be better to reverse that check to always return the _HIGHEST_ as default if nothing is found instead of the _LOWEST_? Or even throw an exception?

You can open a ticket to refactor `EnvironmentsManager`, the code is largely undocumented, I have no idea why 1.3 was selected as default at the moment the code in question was added. Probably something that was considered "safe choice" at that time.

laeubi:
I have created
- https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/750

so lets see if anything breaks ...

stephan-herrmann:
for an overview of all version related changes I suggest to have a look at the start of any recent branch like BETA_JAVA25 in each of jdt.{core,debug.ui}.

For jdt.core you may also refer to https://github.com/eclipse-jdt/eclipse.jdt.core/wiki/Adding-support-for-new-Java-version-and-Features-in-JDT

And mind you, if an unrecognized Java version is classified as "highest" this will raise expectations that may not be satisfied.

mickaelistria:
(thinking aloud here, I hope that this can be interesting to anyone else...)

I think such questions shows a possible confusion with current JDT design (that we may be able to improve at some point) and that is making adoption of newer Java versions, or adoption of Javac for instance relatively difficult in my understanding:
Currently JDT has a centralized concept of "supported versions" and "latest" in `JavaCore`, but the meaning of "supported" is variable. Concretely, supporting java N in JDT can mean:
1. Class files of version N can be read and used to feed compiler/DOM/model...
2. Java N is usable as Execution Environment
3. ECJ can parse/compile to Java N
4. The AST/DOM API has the Java N features
5. The Java model has Java N features (less frequent)
6. Launching using Java N works (probably comes almost directly from item 2)
7. Debugging Java N works (usually not requiring much changes AFAIK as the debug protocol seems very stable?)

Some of these are relatively independent. There are actually often different underlying APIs for each component: ECJ has `ClassFileConstants.getLatestJDKLevel()`, DOM has `AST.getJLSLatest()`, EE vould have some `IExecutionEnvironmentManager.getLatest()` or similar... The separation is already relatively well in place. So the main trouble seems to come from `JavaCore.get*Version*` methods not clearly telling what flavor of "supported" they mean, and downstream consumers potentially misusing those instead of more specific APIs.

Also, it's not clear which ones of those require particular legal caution, and particularly, require to be implemented in a branch and released separately.
I can imagine supporting officially a new Java version in compiler is constrained. But for exampe, il supporting an EE actually legally constrained? Or even being able to read class files of higher versions?...

stephan-herrmann:
Another way to disentangle "Java version" into components is telling apart the different parts of "Java":
* JLS: this is where we have most tight coupling, since rules are not only added but some rules actually change over time
  * grammar - AST: some language changes do not involve any AST changes, e.g., flexible constructor bodies does not introduce any new AST structure, so code using that feature can be mapped to AST of previous versions.
* JVMS - bytecode: whenever the classfile format is changed, we need an update not only of our own class reading implementation, but also of ASM (for advanced source lookup)
* JVM: launching old programs on a new JVM should mostly work
  * some new features may involve new command line options that are rejected by older VMs
  * some old programs need new command line options to run on a newer VM (think of `--add-opens` and friends)
  * some options are generated behind the scenes, in particular for running tests of a modular application.
 * JDK library: limited compatibility 
   * methods / classes even modules may come and go
   * reading API from an older version may break when using `--release` on a unknown new JDK if the format of `ct.sym` changes again.

Sure, running today's programs on a future JVM is the most promising mix of versions. In that vein, **now is a good time to enable launching programs on JDK 26 ea**, no kidding.

But the above list should also indicate that a smooth and consistent developer experience can never be guaranteed for future versions, even when your own code does not use any new features.

mickaelistria:
> now is a good time to enable launching programs on JDK 26 ea, no kidding.

That's good to read it! Can this be done directly on master?

laeubi:
> > now is a good time to enable launching programs on JDK 26 ea, no kidding.
> 
> That's good to read it! Can this be done directly on master?

See also here:
- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4260#issuecomment-3134983658

datho7561:
@stephan-herrmann I'm interested in setting up a PR for enabling running apps with Java 26. Does anything need to be done outside of adding the constants Christoph linked to?

stephan-herrmann:
> @stephan-herrmann I'm interested in setting up a PR for enabling running apps with Java 26. Does anything need to be done outside of adding the constants Christoph linked to?

Feel free to take inspiration also from https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/675 - not all of that may be relevant, though.

Issue #2582, opened 2024-06-17
Switch exhaustiveness error with enum and 'case null' 


rovarga:
For this bit of code:
```
    public enum Foo {
        FOO, BAR;
    }

    public void dispatch(final Foo foo) {
        switch (foo)  {
            // forces JEP-441 exhaustiveness
            case null -> throw new NullPointerException("null foo");
            case FOO -> foo();
            case BAR -> bar();
        }
    }

    private void foo() {
    }

    private void bar() {
    }
```

javac-21.0.3 happily accepts it and does the right thing.
JDT elicits an `An enhanced switch statement should be exhaustive; a default label expected` error.

Seems to be present it current development (I tested with JDT 3.19.600.v20240616-1800).

srikanth-sankaran:
> Seems to be present it current development (I tested with JDT 3.19.600.v20240616-1800).

Are you sure about the versions ? This seems to be a duplicate of https://github.com/eclipse-jdt/eclipse.jdt.core/issues/277 - the test case is almost identical and as of the comment https://github.com/eclipse-jdt/eclipse.jdt.core/issues/277#issuecomment-1851795520 dated Dec 12, 2023 itself the problem is not reproducible. I will release a regression test proving that it works and close this.

I am using this test case: 

```
enum Foo {
        FOO, BAR;
    }

public class X {
    public void dispatch(final Foo foo) {
        switch (foo)  {
            // forces JEP-441 exhaustiveness
            case null -> throw new NullPointerException("null foo");
            case FOO -> foo();
            case BAR -> bar();
        }
    }

    private void foo() {
    	System.out.println("Foo");
    }

    private void bar() {
    	System.out.println("Bar");
    }
    public static void main(String[] args) {
    	new X().dispatch(Foo.FOO);
    	new X().dispatch(Foo.BAR);
    	try {
    	new X().dispatch(null);
    	} catch (NullPointerException npe) {
    		System.out.println("Null");
    	}
	}
}
```

In future always provide full test cases - the methods `dispatch` `foo` and `bar` are not enclosed in any class - Or are you saying you are using the preview feature implicitly named classes ?? I doubt that since the code is lacking `main` method and won't compile without it and you claim it compiles.

srikanth-sankaran:
I have added a regression test to prove this works ok - and that test passes. If you still see the problem, please open with a full test case and details needed to reproduce against next I build

stephan-herrmann:
[...]

PR #2611, opened 2024-06-19
Switch exhaustiveness error with enum and 'case null'

stephan-herrmann:
failing test as witness for
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2582


stephan-herrmann:
Sorry, @srikanth-sankaran I broke it! :)

We have an unfortunate interaction between our optional *warning* (configurable) vs. the hard error per JLS.

Of course we could decide to let our optional analysis shut-up for enhanced switches that are subject to exhaustiveness checks. Perhaps evaluating two related but different sets of rules on the same construct is confusing anyway, right?

srikanth-sankaran:
I have included the test case into the PR: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2613

Thanks so much!

Issue #2552, opened 2024-06-12
mac: Random failing CopyMoveElementsTests.testCopyFieldForce


jukzi:
failed in two MacOS builds:
```
junit.framework.AssertionFailedError: No delta found for class Y [in Y.java [in <default> [in src [in P]]]]
boolean foo
int foo#2, all deltas: [Java Model[*]: {CHILDREN}
P[*]: {CHILDREN}
src[*]: {CHILDREN}
<default>[*]: {CHILDREN}
Y.java[*]: {CONTENT | PRIMARY RESOURCE}]
at junit.framework.Assert.fail(Assert.java:57)
at junit.framework.TestCase.fail(TestCase.java:223)
at org.eclipse.jdt.core.tests.model.CopyMoveTests.copyPositive(CopyMoveTests.java:122)
at org.eclipse.jdt.core.tests.model.CopyMoveElementsTests.testCopyFieldForce(CopyMoveElementsTests.java:215)

```
both:
https://download.eclipse.org/eclipse/downloads/drops4/I20240611-0150/testresults/html/org.eclipse.jdt.core.tests.model_ep433I-unit-macM1-java17_macosx.cocoa.aarch64_17.html
https://download.eclipse.org/eclipse/downloads/drops4/I20240611-0150/testresults/html/org.eclipse.jdt.core.tests.model_ep433I-unit-mac64-java17_macosx.cocoa.x86_64_17.html

May relate to another random fail of CopyMoveElementsTests.testCopyConstructorInDifferentProject:
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1094

srikanth-sankaran:
`testCopyFieldInDifferentProject` occasionally fails too

srikanth-sankaran:
`org.eclipse.jdt.core.tests.model.CopyMoveElementsTests.testCopyField` occasionally fails too

stephan-herrmann:
Look, what I found in the console when `testCopyField` failed:
```
[2024-06-22T20:10:28.199Z] java.lang.Throwable: Refreshing workspace [Worker-0: Refreshing workspace] Caller of IElementChangedListener#elementChanged with delta Java Model[*]: {CHILDREN}
[2024-06-22T20:10:28.199Z] 	P[*]: {CHILDREN}
[2024-06-22T20:10:28.199Z] 		src[*]: {CHILDREN}
[...]

PR #2629, opened 2024-06-23
mac: Random failing CopyMoveElementsTests.testCopyFieldForce

stephan-herrmann:
hopefully fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2552


stephan-herrmann:
@jukzi do you think this is a valid fix? Or cheating? Or maybe not even effective? (see my recipe for reproduction in the issue (#2552).

jukzi (approved):
IF it helps - OK - i can not reproduce on my computer,
The same code is in org.eclipse.jdt.core.tests.model.CopyMoveElementsTests.setUpSuite()

stephan-herrmann:
> IF it helps - OK 

thanks

> i can not reproduce on my computer,

The recipe I gave *should* work deterministically on any computer :)

> The same code is in org.eclipse.jdt.core.tests.model.CopyMoveElementsTests.setUpSuite()

The same call to `waitForAutoBuild()`, yes. But I don't think we have a problem during `setUpSuite()`. What matters is that the change performed during `copy()` has been built before all the assertions below.

PR #2637, opened 2024-06-25
Refactor `ASTRequestor` to reduce the access it has to the compiler

datho7561:
## What it does
This reduces the access that `ASTRequestor` has to the `CompilationUnitResolver` to just the method it needs, `createBinding`.

This also makes it possible to use an alternate additional binding resolver instead of hardcoding the one implemented by `CompilationUnitResolver`.

## How to test
Run the AST Converter test suite (I'm sure this feature is used elsewhere, but this is the only place that I know uses it for sure).

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTRequestor.java
+++ b/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTRequestor.java
@@ -13,6 +13,8 @@
  *******************************************************************************/
 package org.eclipse.jdt.core.dom;
 
+import java.util.function.Function;
+
 import org.eclipse.jdt.core.ICompilationUnit;
 
 /**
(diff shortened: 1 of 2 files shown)


rgrunber:
If the only reason to pass in the compilation unit resolver is for creating bindings, and if clients may want to swap that out with something else, then this seems like a nice decoupling. Is it worth being more explicit on `additionalBindingResolver` with the fact that `String` must be a binding key ?

datho7561:
Good point, I'll make the Javadoc more specific

mickaelistria:
@rgrunber Can you please re-review, and merge if everything is fine?

mickaelistria:
As @rgrunber is on PTO, can @jjohnstn or @testforstephen you please have a look and merge if appropriate?

testforstephen (approved):
this just made minor changes to the internal implementation without affecting the API. It's safe to me.

PR #2647, opened 2024-06-29
Prevent excessive creation of intermediate SplitPackageBindings #2646

### LINKED ISSUE #2646: Prevent excessive creation of intermediate SplitPackageBinding instances
I found that in a code base with many Java modules (explicit modules containing a module-info file) and a shared package prefix, the `SplitPackageBinding.combine` produces an unnecessary amount of binding instances.
To make things worse, even if the calling code discards the previous SplitPackageBinding instance which was combined into a new one the old instance will still remain in memory because the old instance is still referenced as a `PackageBinding.wrappingSplitPackageBindings` in all `PlainPackageBinding`s it wraps in addition to the newly combined instance and this is done again and again and again.

In a large code base I encountered `PlainPackageBinding`s which hold references to more than 600,000 SplitPackageBinding instances. You can imagine that my heap memory does not like that and it slows down the compiler code which later iterates over this huge list.

It might be sensible to remove the old SplitPackageBinding from the wrappingSplitPackageBindings when creating a new one but that might require to replace the `ArrayList` with a `(Linked)HashSet` and I have neither tried it nor investigated into the possible consequences and problems of such a change.

Because I found a simpler and from my understanding safe solution for my problem.

I think my main problem is rooted in how `SplitPackageBinding` decides to reuse or create a new instance. I found two code paths which trigger bad behaviour with the current implementation.

First `org.eclipse.jdt.internal.compiler.lookup.ModuleBinding.combineWithPackagesFromOtherRelevantModules(PackageBinding, char[][], char[][])`. This one produces a SplitPackageBinding for a given package with all required modules. Imagine the Eclipse project would be modularized and you got a 'root' module which depends on all others. Now you got a long list of modules which all contain the package `org` (because all Eclipse code is placed in `org.eclipse...` and you also get all external dependencies having `org` packages) and will all be combined into a SplitPackageBinding.

The `SplitPackageBinding.combine` method first tries to reuse one of the bindings however, in this case you can expect that the rank is the same for all (at least for the cases I debugged). Also since we iterate the modules you will never `subsumes` one binding with the other. Therefore each loop iteration will produce a new `SplitPackageBinding` instance.
Each new `SplitPackageBinding` will take over the `PlainPackageBinding` incarnations from its predecessor and will register itself as wrapping binding on the `PlainPackageBinding`.

The seconds one, probably _a lot_ worse, way is through `SplitPackageBinding.getPackage0`, `SplitPackageBinding.getPackage0Any` and `SplitPackageBinding.combineWithSiblings`. If I understand it correctly this might be called to find a sub-package and if not found directly in the `knownPackages` it will check its plain package incarnations. If more than one incarnation contains  the searched sub-package it will produce a new `SplitPackageBinding`, again step by step, and with each step will register the intermediate bindings as wrapping on the plain binding. This grows quite large, quite quickly.

My fix is to extend the logic when to reuse an existing `SplitPackageBinding` and when to create a new one when combing bindings. From my understanding of the code there is no need to create a brand new `SplitPackageBinding` if the `primaryModule` has not changed. In the problematic method `combineWithPackagesFromOtherRelevantModules` I described above the `primaryModule` is always the same and it feels way more natural to have _one_ `SplitPackageBinding` here and add all incarnations and declaring modules one by one to it.

Also for the even more problematic uses of `combine` in `SplitPackageBinding` it would prevent creation of a lot of intermediate `SplitPackageBinding`s.

What might still be problematic is, that `SplitPackageBinding.getPackage0` and `SplitPackageBinding.getPackage0Any` still can produce ephemeral(?) `SplitPackageBinding` which are still referenced as wrapping bindings on the plain bindings for way longer as the split binding might be used by the caller.

PPazderski:
should fix #2646

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/SplitPackageBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/SplitPackageBinding.java
@@ -47,7 +47,13 @@ public class SplitPackageBinding extends PackageBinding {
 			return previous;
 		if (binding.subsumes(previous))
 			return binding;
-		SplitPackageBinding split = new SplitPackageBinding(previous, primaryModule);
+
+		final SplitPackageBinding split;
+		if (previous instanceof SplitPackageBinding previousSplit && previousSplit.enclosingModule == primaryModule) {
+			split = previousSplit;
+		} else {
+			split = new SplitPackageBinding(previous, primaryModule);
+		}
 		split.add(binding);
 		return split;
 	}


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/SplitPackageBinding.java
```diff
@@ -47,7 +47,13 @@ public static PackageBinding combine(PackageBinding binding, PackageBinding prev
...
 			return binding;
-		SplitPackageBinding split = new SplitPackageBinding(previous, primaryModule);
+
+		final SplitPackageBinding split;
+		if (previous instanceof SplitPackageBinding previousSplit && previousSplit.enclosingModule == primaryModule) {
+			split = previousSplit;
+		} else {
+			split = new SplitPackageBinding(previous, primaryModule);
+		}
 		split.add(binding);
```
jukzi, line 57: Attention: this modifies the previous PackageBinding. A caller would not expect that from the signature.

PPazderski: Of cause, and still I somehow missed it :/
Thanks for pointing it out. This can be a problem. I will replace it with a different change soon which should fix my problem.

PPazderski: Well, seem to be a bit more tricky than I thought to find get a different approach. Wouldn't mind if someone finds a different solution to the problem I described in #2646.

stephan-herrmann:
I'll look into this soonish. Please don't merge yet. Thanks.

PPazderski:
Change is now obsolete with #2730

Issue #1937, opened 2022-11-03
Caching Compiler Threads?


laeubi:
While debugging an Eclipse Instance I noticed that when many projects are rebuild I see a lot of "Compiler Processing Thread" come and go in the Debugger. Maybe it would be good to have a (cached) Threadpool instead of creating and destroying Threads instead?

jukzi:
i worked on such pool in https://git.eclipse.org/r/c/jdt/eclipse.jdt.core/+/177563

PR #2665, opened 2024-07-03
ProcessTaskManager: reuse Compiler Thread #1937

jukzi:
For easier monitoring.
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1937


jukzi:
Example sampling clean build platform workspace, showing a single thread instead of thousands before:
image

PR #2666, opened 2024-07-03
ReadManager: reuse reader threads #1937

jukzi:
For easier monitoring.
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1937

Also refactored the read-ahead-queue with java.util.concurrent, making the source smaller.


jukzi:
Example sampling building platform workspace showing the pooled threads (instead of thousands before)):
image

jukzi:
Almost no time spend in ReadManager without the former `CUSHION`. Files are all read by Reader Threads and Builder has almost never to wait for them:
image

Issue #2679, opened 2024-07-05
Bug 533327 - [9] Implement JEP 211


stephan-herrmann:
See https://bugs.eclipse.org/533327 :

> Brought to you via http://openjdk.java.net/jeps/211 is this change in JLS §9.6.4.6:
> A Java compiler must produce a deprecation  warning [...] unless:
> [...]
> •  The use is within an import declaration that imports the ordinarily deprecated
type or member."

stephan-herrmann:
Wondering about "ordinarily" in the above I see the JLS has the same import exception regarding "ordinarily deprecated" and "terminally deprecated"  (causing otherwise a "removal warning").

So, nothing to worry about, in this regard all deprecation is treated the same.

jukzi:
Thanks for implementing this. Should render @SuppressWarnings at class level for the imports unneeded, allowing more fine grained @SuppressWarnings

PR #2680, opened 2024-07-05
Bug 533327 - [9] Implement JEP 211

stephan-herrmann:
fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2679


stephan-herrmann:
@jarthana @iloveeclipse  I'd like to apply the same also to warnings / errors regarding restricted access (according to access rules):

Currently, just like with deprecation warnings, those warnings ("Discouraged access ...") can only be suppressed by adding `@SuppressWarnings("restriction")` on the entire class, which is typically undesirable.

So perhaps we should also stop raising such warnings on imports? Warnings on actual usage will of course remain.

Any objections?

Strictly speaking the rule in JLS was changed in version 9, but I don't think anybody will complain if the same also applies at 1.8, right?

stephan-herrmann:
Test failures show that I "accidentally" already did away with warnings for restricted access reported on import declarations.

I'll wait for agreement / objections before adjusting those tests _or_ limiting the implementation change to pure deprecation warnings.

iloveeclipse:
> Any objections?

Not from me, makes sense.

> Strictly speaking the rule in JLS was changed in version 9, but I don't think anybody will complain if the same also applies at 1.8, right?

Is that a problem / big overhead to perform it on 9+ level only? I would follow the spec, simply to be consistent with it even if the change should be fine with 1.8.

stephan-herrmann:
> > Any objections?
> 
> Not from me, makes sense.

thanks

> Is that a problem / big overhead to perform it on 9+ level only?

Not a problem, only means instead of deleting code I'll have to add code :)

iloveeclipse:
> Not a problem, only means instead of deleting code I'll have to add code :)

I would prefer the opposite. 
@jarthana, @srikanth-sankaran : any opinions?

stephan-herrmann:
> > Not a problem, only means instead of deleting code I'll have to add code :)
> 
> I would prefer the opposite. @jarthana, @srikanth-sankaran : any opinions?

If "opposite" refers to removing code rather than adding, this is at conflict with your previous suggestion.

Anyway, not a big deal, see [5024190](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2680/commits/5024190031e796e636c6e167831f8f1d765f6015) for warning re-introduced at 1.8-

iloveeclipse:
@stephan-herrmann : this PR was merged "too fast", as it introduced new warnings that failed the SDK build (see https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2178).

I don't have a solution yet for how we can avoid that in the future (except to avoid treating warnings as errors in project settings), because it is of course a lot of effort / time needed to compile everything locally with a new compiler version.

stephan-herrmann:
@iloveeclipse I'm all ears if you have any feasible suggestions. For the recently added "unnecessary casts" warnings, I was happy to fix them ahead of time, but that certainly was a one-time favor :)

Issue #2691, opened 2024-07-08
Clean build: SourceFile(s) read twice


jukzi:
During Project/"Clean all projects" SourceFile(s) are read multiple times.
Once from ReadManager - exepected.
And later another time from Parser.getMethodBodies() - without further digging i guess this is wrong.  It is prepared to use a ReadManager there, but readManager=null.
```
Thread [Compiler Processing Task] (Suspended (breakpoint at line 11694 in org.eclipse.jdt.internal.compiler.parser.Parser))	
	org.eclipse.jdt.internal.compiler.parser.Parser.getMethodBodies(org.eclipse.jdt.internal.compiler.ast.CompilationUnitDeclaration) line: 11694	
	org.eclipse.jdt.internal.compiler.Compiler.process(org.eclipse.jdt.internal.compiler.ast.CompilationUnitDeclaration, int) line: 890	
	org.eclipse.jdt.internal.compiler.ProcessTaskManager.compile() line: 163	
```
![image](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/51790620/26390531-caca-4567-b1ec-c65143c2cea0)

jukzi:
Both reads happen during` org.eclipse.jdt.internal.compiler.Compiler.compile(ICompilationUnit[], boolean)`
The first read happens during `org.eclipse.jdt.internal.compiler.Compiler.beginToCompile(org.eclipse.jdt.internal.compiler.env.ICompilationUnit[]) line: 393	`
and the second read is triggered from
`org.eclipse.jdt.internal.compiler.Compiler.processCompiledUnits(int, boolean) line: 606	`

jukzi:
In Compiler.internalBeginToCompile() a decision is made if a full parse is done or only a dietParse. By default it is only a dietParse because parseThreshold<=0. That diet parse skips parsing the method bodies. To get them later a second parse has to occur. That 2nd parse is done in another Thread ("Compiler Processing Task"). Nevertheless the main compilation has to wait on that result anyway. 
I measured an experiment with diet parse vs full parse.
![image](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/51790620/fe19a133-9095-453f-aab3-d2fa40414e93)
![image](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/51790620/6d3f9bd4-9155-4e94-a658-74a4e0f7ca46)
[...]

PR #2694, opened 2024-07-09
Avoid read source twice during compile #2691

jukzi:
Use parse() instead of dietParse() to avoid parse during Parser.getMethodBodies()

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -28,7 +28,6 @@ pipeline {
 					# via configuration/argLine property in pom.xml
 					# export MAVEN_OPTS="-Xmx2G"
 					
-					mvn clean install -f org.eclipse.jdt.core.compiler.batch -DlocalEcjVersion=99.99 -Dmaven.repo.local=$WORKSPACE/.m2/repository -DcompilerBaselineMode=disable -DcompilerBaselineReplace=none
 					
 					mvn -U clean verify --batch-mode --fail-at-end -Dmaven.repo.local=$WORKSPACE/.m2/repository \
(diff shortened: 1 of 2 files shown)


iloveeclipse:
Should this be a draft too (like https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2695/)? I assume you are experimenting with the right solution?

jukzi:
i do not understand why the build fails before executing tests.because jdt compiles itself? how to get test results ?

iloveeclipse:
> i do not understand why the build fails before executing tests.because jdt compiles itself? how to get test results ?

If JDT can't compile itself it is already a test failure :-)

iloveeclipse:
The background is: we first compile batch compiler from sources using old compiler and then continue with compilation of the code with the freshly built compiler. This is another way to test real code not just some test snippets.

iloveeclipse:
So basically it says that compiler is unable to build JDT anymore with your changes.

jukzi:
yea, can i disable that to get a little bit more detail why it fails to compile?

iloveeclipse:
You could try to temporarily disable that in Jenkins file, but may be you can try to compile same code in the IDE in debugger first either from Main or IDE from IDE (not sure if that would be specific to batch compiler invocation)?

jukzi:
i compiled the whole platform workspace first locally in the ide- successully  :-/

iloveeclipse:
So you can try to run org.eclipse.jdt.internal.compiler.batch.Main directly from debugger with the source in question.

jukzi:
@stephan-herrmann in principle this change looks harmeless simple but it causes Problem markers when imports are only used in javadoc:
image
Can you help to solve that riddle? I guess its a problem that could and should be solved independently of this PR.

PR #2700, opened 2024-07-10
A flag to resolve codeSelect using ASTParser

mickaelistria:
## What it does

Introduce a CompilationUnit.DOM_BASED_OPERATIONS flag which instead of using the ECJ-derived parser uses ASTParser to create a DOM and then uses the DOM to implement codeSelect

This has the benefit of:
* Simplifying the code as an crawling DOM is usually simpler than dealing with lower-level parser
* allowing other parsers as backend if ASTParser is configured for it

## How to test

Set system property `-DCompilationUnit.DOM_BASED_OPERATIONS=true` and see codeSelect (hover, ctrl+click) working with ASTParser.

## Author checklist


jjohnstn (approved):
Seeing this as being part of the javac effort and is completely distinct from regular JDT compiler without special flag, LGTM.

PR #2701, opened 2024-07-10
A flag to resolve populating index using ASTParser

mickaelistria:
## What it does

Introduce a SourceIndexer.DOM_BASED_INDEXER flag which instead of using the ECJ-derived parser uses ASTParser to create a DOM and then crawls this DOM to populate index.

This has the benefit of:
* Simplifying the code as an crawling DOM is usually simpler than dealing with lower-level parser
* allowing other parsers as backend if ASTParser is configured for it

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/DOMToIndexVisitor.java
+++ b/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/DOMToIndexVisitor.java
@@ -1,0 +1,345 @@
+/*******************************************************************************
+ * Copyright (c) 2023 Red Hat, Inc. and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.jdt.internal.core.search.indexing;
+
+import java.nio.file.Path;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Objects;
+import org.eclipse.jdt.core.Signature;
+import org.eclipse.jdt.core.dom.*;
+
+class DOMToIndexVisitor extends ASTVisitor {
+
+	private SourceIndexer sourceIndexer;
+
+	private char[] packageName;
+	private List<AbstractTypeDeclaration> enclosingTypes = new LinkedList<>();
+
+	public DOMToIndexVisitor(SourceIndexer sourceIndexer) {
+		super(true);
+		this.sourceIndexer = sourceIndexer;
+	}
+
+	private AbstractTypeDeclaration currentType() {
+		return this.enclosingTypes.get(this.enclosingTypes.size() - 1);
+	}
+
+	@Override
+	public boolean visit(PackageDeclaration packageDeclaration) {
+		this.packageName = packageDeclaration.getName().toString().toCharArray();
+		return false;
+	}
+
+	@Override
+	public boolean visit(TypeDeclaration type) {
+		char[][] enclosing = type.isLocalTypeDeclaration() ? IIndexConstants.ONE_ZERO_CHAR :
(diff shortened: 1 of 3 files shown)


org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/SourceIndexer.java
```diff
@@ -88,6 +98,10 @@ public SourceIndexer(SearchDocument document) {
 	}
 	@Override
 	public void indexDocument() {
+		if (Boolean.getBoolean(getClass().getSimpleName() + ".DOM_BASED_INDEXER")) { //$NON-NLS-1$
```
jjohnstn, line 101: Why is this flag per class vs a project or workspace preference?  For a large project, this could result in a lot of system properties.  As well, why are you are using the simple name as there could be collision with similarly named classes in different packages.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/SourceIndexer.java
```diff
@@ -213,6 +227,15 @@ private void purgeMethodStatements(TypeDeclaration type) {
 
 	@Override
 	public void indexResolvedDocument() {
+		// TODO We need to rebuild the DOM with binding resolution enabled
```
jjohnstn, line 230: What are the ramifications of having this code commented out?  Can the PR be tested without it?  You should have an issue cited here which is to fix the problem you are noting.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/DOMToIndexVisitor.java
```diff
@@ -0,0 +1,371 @@
+/*******************************************************************************
```
jjohnstn, line 1: The title of this PR is a bit misleading.  It implies you are just adding a new flag but you are adding this new code and this should be spelled out in the PR title and commit comment (e.g. Convert DOM to Index based on new flag setting)

jjohnstn (changes requested):

mickaelistria:
I'm converting as a draft as one of the remaining TODO is critical.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/DOMToIndexVisitor.java
mickaelistria: I fixed it.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/SourceIndexer.java
mickaelistria: Using a simple name is for consistency with other system properties such as `ICompilationUnitResolver` which are already used in cases. I can switch it to fully qualified name if this is a must-have.
Note that we do not want to offer it as a preference or something more usable. This switch is deep in the code so that it's mostly known of people who maintain the code, and will not be part of any API supported by JDT. Sometimes in the future, the switch will be removed together with legacy code.

mickaelistria:
With latest version of everything, I couldn't reproduce the issue that triggered to commenting out the indexDocumentFromDOM in indexResolveDocument. So I think this is now ready for review.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/SourceIndexer.java
jjohnstn: Ok

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/SourceIndexer.java
jjohnstn: This has been resolved.

jjohnstn (approved):
With the commented out code now being called and the code being completely optional and distinct from regular JDT usage, I am merging.

Issue #2642, opened 2024-06-26
False positive warning "Potential resource leak" inside "try with resources"


iloveeclipse:
Given this example and enabled `org.eclipse.jdt.core.compiler.problem.potentiallyUnclosedCloseable=warning` option in `.settings/org.eclipse.jdt.core.prefs`

![image](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/964108/2d47d904-9fca-4392-9c58-f42513ca6f95)

```
import java.io.FileWriter;
public class DemoNonCloseableWarning {
    public static void main(String[] args) throws Exception {
        try (FileWriter writer = new FileWriter("/dev/null")) {
            writer.append("\n");
        }
    }
}
```

we report false positive warning `Potential resource leak: '<unassigned Closeable value>' may not be closed`.

Of course this is option is not enabled by default, and of course it is **potential**, but here we have undoubtedly closed writer, so not idea where the "potential" warning is coming from.

@stephan-herrmann : FYI.

iloveeclipse:
Fun fact: it worked at least in 3.8.2 Eclipse on Java 1.7 :-)

stephan-herrmann:
This looks spooky on first glance:
```
----------
1. WARNING in /tmp/DemoNonCloseableWarning.java (at line 5)
        writer.append("\n");
        ^^^^^^^^^^^^^^^^^^^
Potential resource leak: '<unassigned Closeable value>' may not be closed
----------
1 problem (1 warning)
```

Not allocating the writer bad calling `append()` triggers the warning. But wait: append *returns* a Writer (which could in theory be a fresh new resource).

We do have provision for this situation, see `org.eclipse.jdt.internal.compiler.lookup.TypeConstants.FLUENT_RESOURCE_CLASSES`.  Something confuses the detection of this "fluent" method.

iloveeclipse:
> But wait: append _returns_ a Writer

OK, I see it now where AI hallucinates :-). Compiler assumes the return value has to be closed...

stephan-herrmann:
> OK, I see it now where AI hallucinates :-). Compiler assumes the return value has to be closed...

Don't worry: with AI in the picture any causalities between input program and warnings would be much more obscure.. ;-P

PR #2718, opened 2024-07-16
False positive warning "Potential resource leak" inside "try with resources"

stephan-herrmann:
Traverse supers to detect fluent methods

Test:
+ fine tune test230_sourcepath_vs_classpath to avoid superclass loading

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2642

### DIFF
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/ResourceLeakTests.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/ResourceLeakTests.java
@@ -7269,4 +7269,26 @@ public void testGH2129() {
 		"",
 		options);
 }
+public void testGH2642() {
+	Map options = getCompilerOptions();
+	options.put(CompilerOptions.OPTION_ReportPotentiallyUnclosedCloseable, CompilerOptions.ERROR);
+	options.put(CompilerOptions.OPTION_ReportUnclosedCloseable, CompilerOptions.ERROR);
+	options.put(CompilerOptions.OPTION_ReportExplicitlyClosedAutoCloseable, CompilerOptions.ERROR);
+	runLeakTest(
+		new String[] {
+			"DemoNonCloseableWarning.java",
+			"""
+			import java.io.FileWriter;
+			public class DemoNonCloseableWarning {
+			    public static void main(String[] args) throws Exception {
+			        try (FileWriter writer = new FileWriter("/dev/null")) {
+			            writer.append("\\n");
+			        }
+			    }
+			}
+			"""
+		},
+		"",
+		options);
+}
 }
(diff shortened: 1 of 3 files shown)


org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/ResourceLeakTests.java
```diff
@@ -7269,4 +7269,26 @@ public void testGH2129() {
...
+	options.put(CompilerOptions.OPTION_ReportUnclosedCloseable, CompilerOptions.ERROR);
+	options.put(CompilerOptions.OPTION_ReportExplicitlyClosedAutoCloseable, CompilerOptions.ERROR);
+	runLeakTest(
+		new String[] {
+			"DemoNonCloseableWarning.java",
+			"""
+			import java.io.FileWriter;
+			public class DemoNonCloseableWarning {
+			    public static void main(String[] args) throws Exception {
+			        try (FileWriter writer = new FileWriter("/dev/null")) {
```
iloveeclipse, line 7284: @stephan-herrmann : this doesn't work on Windows, sorry for providing this example.

We have new test fails: https://download.eclipse.org/eclipse/downloads/drops4/I20240716-1800/testresults/html/org.eclipse.jdt.core.tests.compiler_ep433I-unit-win32-java17_win32.win32.x86_64_17.html

stephan-herrmann: I don't have a windows box to test what works on windows.
Do you have a better example, or must I add a dummy error to make this a negative test not trying to execute the program?

iloveeclipse: Why do we execute the snippet at all? Easiest is not to run it.

iloveeclipse: Another possibility is to use some other "writer" like StringWriter (haven't checked if that would show same issue, but I assume it should).

stephan-herrmann: > Why do we execute the snippet at all? Easiest is not to run it.

We have two strategies:
* runNegativeTest() look for expected compile errors
* runConformTest() expect no compiler errors but then run the program and expect correct output from it

Adding a new parameter to runConformTest() would need to trickle down a myriad of delegating overloads, not a nice option.

So that standard trick to avoid executing is to use runNegativeTest() - with a dummy error like a field with type Zork :)

Issue #2748, opened 2024-07-23
SplitPackageBinding might be added as child of PlainPackageBinding


PPazderski:
There is some additional context for this [at the end](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2646#issuecomment-2243606326) of #2646.

In that issue [I described](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2646#issuecomment-2242353968) a false "package not found" error I got from the compiler which is, at least for my project setup, avoided with #2730.

Stephan mentioned and added an assert to ensure a `SplitPackageBinding` is never a child (`knownPackage`) of a `PlainPackageBinding` but I still see this happening. This time I was able to create a reproducing example.

[multi-module-project.zip](https://github.com/user-attachments/files/16354075/multi-module-project.zip)

The zip contains four projects. If I compile the "app" project (or all) I will run into the new assertion.
https://github.com/eclipse-jdt/eclipse.jdt.core/blob/7ea3685728d498c68f7f132a034049a7ba94da06/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/PlainPackageBinding.java#L50

I also transformed this project setup into a unit test. Will link it in the next comment. (they are not exactly equal but should behave the same)

PPazderski:
JUnit test which should trigger the assertion can be found at

https://github.com/PPazderski/eclipse.jdt.core/commit/a4ca0b9d29eec9915fce5cdd05ed35a9d7bb7291

PR #2749, opened 2024-07-23
JUnit test for setup where split package becomes child of plain package

PPazderski:
unit test for #2748 (fails only if run with assertions enabled)


stephan-herrmann:
The fix looks trivial: just one more location to treat auto modules like the unnamed module.

I'm not quite sure, though, about some details in existing code, relating to unnamed and automatic modules. One part has been outsourced to #2764.

Still I'll try to move the fix in this PR closer to the specification as I read it today.

stephan-herrmann:
> Still I'll try to move the fix in this PR closer to the specification as I read it today.

That failed, as now we are inconsistent regarding what an auto module reads. I reverted to the previous fix and leave the remaining issue to #2764 entirely.

I also slightly improved readability of the otherwise well-crafted test: use text blocks where possible :)

stephan-herrmann:
Thanks @PPazderski !

PR #2770, opened 2024-07-28
Return a SortedSet in getAllJavaSourceVersionsSupportedByCompiler()

HannesWell:
The list current returned by `JavaCore.getAllJavaSourceVersionsSupportedByCompiler()` is mainly/only used to either fetch the first or last element or test if a version-String is contained.

Especially for the last operation a Set is a better choice regarding runtime. A `SortedSet` additionally also offers the methods `first()`/`last()`, which are semantically more expressive than calling List.get(0)/List.get(List.size()-1).

Furthermore using a `SortedSet` would also better reflect the general semantic of the returned collection and element-access is usually not of interest. It would also reduce the number of methods provided by JavaCore and therefore make it's API smaller.

@iloveeclipse what do you think?

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
@@ -121,6 +121,7 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Hashtable;
@@ -128,6 +129,8 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.SortedSet;
+import java.util.TreeSet;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.zip.ZipFile;
@@ -3336,12 +3339,19 @@ public final class JavaCore extends Plugin {
 	 * Ordered set (from oldest to latest) of all Java source versions <b>supported</b> by compiler.
 	 * The values are from {@link JavaCore}{@code #VERSION_*}.
 	 */
-	private static final List<String> SUPPORTED_VERSIONS;
+	private static final SortedSet<String> SUPPORTED_VERSIONS;
 	static {
-		ArrayList<String> temp = new ArrayList<>();
+		Comparator<String> byVersion = Comparator.comparingDouble((String v) -> {
+			try {
+				return Double.parseDouble(v);
+			} catch (RuntimeException e) {
+				return 0;
+			}
+		}).thenComparing(Comparator.naturalOrder());
+		SortedSet<String> temp = new TreeSet<>(byVersion);
 		temp.addAll(allVersions);
 		temp.removeAll(UNSUPPORTED_VERSIONS);
-		SUPPORTED_VERSIONS = Collections.unmodifiableList(temp);
+		SUPPORTED_VERSIONS = Collections.unmodifiableSortedSet(temp);
 	}
 
 	/**
@@ -3367,7 +3377,7 @@ public final class JavaCore extends Plugin {
 	 * @see #getFirstJavaSourceVersionSupportedByCompiler()
 	 * @since 3.39
 	 */
-	public static List<String> getAllJavaSourceVersionsSupportedByCompiler() {
+	public static SortedSet<String> getAllJavaSourceVersionsSupportedByCompiler() {
 		return SUPPORTED_VERSIONS;
 	}
 
@@ -6472,7 +6482,7 @@ public final class JavaCore extends Plugin {
 	 * @since 3.39
 	 */
 	public static String getFirstJavaSourceVersionSupportedByCompiler() {
-		return SUPPORTED_VERSIONS.get(0);
+		return SUPPORTED_VERSIONS.first();
 	}
 
 	/**


iloveeclipse:
Maybe later this week

org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
```diff
@@ -3396,6 +3406,7 @@ public static boolean isSupportedJavaVersion(String version) {
 	 * @since 3.39
 	 */
 	public static boolean isJavaSourceVersionSupportedByCompiler(String version) {
+		//TODO: drop this again and just mention the use-case in the doc of getAllJavaSourceVersionsSupportedByCompiler()?
```
iloveeclipse, line 3409: I would not drop it, I believe it is easier to read that compared to `getAllJavaVersionsSupportedByCompiler().contains(someVersion)` and also less error prone, as `contains(Object)` allows to pass everything inside.
So please remove TODO.

org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
```diff
@@ -6472,7 +6483,8 @@ public static String latestSupportedJavaVersion() {
 	 * @since 3.39
 	 */
 	public static String getFirstJavaSourceVersionSupportedByCompiler() {
-		return SUPPORTED_VERSIONS.get(0);
+		return SUPPORTED_VERSIONS.first();
+		//TODO: drop this again and just mention the use-case in the doc of getAllJavaSourceVersionsSupportedByCompiler()?
```
iloveeclipse, line 6487: This could be done in a second PR, because currently JDT UI uses this method. Once this PR is merged, we could use `getAllJavaVersionsSupportedByCompiler().first()`.

org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
```diff
@@ -3367,7 +3377,7 @@ public static List<String> getAllVersions() {
 	 * @see #getFirstJavaSourceVersionSupportedByCompiler()
 	 * @since 3.39
 	 */
-	public static List<String> getAllJavaSourceVersionsSupportedByCompiler() {
+	public static SortedSet<String> getAllJavaSourceVersionsSupportedByCompiler() {
```
iloveeclipse, line 3380: THis, if merged, would break JDT Debug build.
Please provide PR to workaround this, like
```
diff --git a/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/jres/JREsPreferencePage.java b/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/jres/JREsPreferencePage.java
index fbeced8..898e5b5 100644
--- a/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/jres/JREsPreferencePage.java
+++ b/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/jres/JREsPreferencePage.java
@@ -17,3 +17,4 @@
 import java.io.File;
-import java.util.List;
+import java.util.SortedSet;
+import java.util.TreeSet;
 
@@ -200,4 +201,4 @@
 		else {
-			List<String> allVersions = JavaCore.getAllJavaSourceVersionsSupportedByCompiler();
-			String latest = allVersions.get(allVersions.size() - 1);
+			SortedSet<String> allVersions = new TreeSet<>(JavaCore.getAllJavaSourceVersionsSupportedByCompiler());
+			String latest = allVersions.last();
 			String vmver = null;
diff --git a/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/jres/StandardVMPage.java b/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/jres/StandardVMPage.java
index 14b82a6..acbe8f4 100644
--- a/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/jres/StandardVMPage.java
+++ b/org.eclipse.jdt.debug.ui/ui/org/eclipse/jdt/internal/debug/ui/jres/StandardVMPage.java
@@ -18,3 +18,4 @@
 import java.net.URL;
-import java.util.List;
+import java.util.SortedSet;
+import java.util.TreeSet;
 
@@ -249,4 +250,4 @@
 		if (s.isOK()) {
-			List<String> allVersions = JavaCore.getAllJavaSourceVersionsSupportedByCompiler();
-			String latest = allVersions.get(allVersions.size() - 1);
+			SortedSet<String> allVersions = new TreeSet<>(JavaCore.getAllJavaSourceVersionsSupportedByCompiler());
+			String latest = allVersions.last();
 			IVMInstallType vmType = fVM.getVMInstallType();
```
After that we could merge this PR.

iloveeclipse (changes requested):

org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
HannesWell: Understand.
And maybe if we remove the `get` prefix?
I would find
`JavaCore.allJavaVersionsSupportedByCompiler().contains(someVersion)` readable. And usually either Linters or even the compiler warn if the type is unrelated (IIRC).

But it's just a suggestion, if you still prefer to keep it, I'm fine.

HannesWell:
Thank you for your review I'm fine with all of your remarks. I'm away till the beginning of next week but will then update this PR and provide the fixes for jdt.ui.

org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
HannesWell: Done in https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/468.

HannesWell:
@iloveeclipse I applied all requested changes and created the PR to adapt JDT-Debug to the changed API
- https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/468

When this is submitted+ I-build I'll create follow-ups to adjust to replace `JavaCore.getFirstJavaSourceVersionSupportedByCompiler()`.

iloveeclipse:
The change in jdt.debug will not compile yet, it depends on this one, right? Have no IDE right now to check. So we have to merge both at same time.

HannesWell:
> The change in jdt.debug will not compile yet, it depends on this one, right? Have no IDE right now to check. So we have to merge both at same time.

Yes that's right. As far as I can tell I have all JDT projects in my workspace and with both changes they are error free.

> So we have to merge both at same time.

Yes. Since there is an I-build soon I think now or short before it would be a good time to avoid an extra I-build without disturbing many others.

PR #2776, opened 2024-07-30
WIP: Fix DefaultBindingResolver.isResolvedTypeInferredFromExpectedType()

### LINKED ISSUE #2775: "Inline local variable" produces broken code with JLS >= 1.8
Follow up from https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2536.

Found while analyzing test failure of `org.eclipse.jdt.ui.tests.refactoring.InlineTempTests.test31()` in https://download.eclipse.org/eclipse/downloads/drops4/I20240725-1800/testresults/html/org.eclipse.jdt.ui.tests.refactoring_ep433I-unit-cen64-gtk3-java17_linux.gtk.x86_64_17.html

The test worked with the project compliance set to 1.4 or 1.5 (not sure which was used in the hierarchy of the test).
With the project compliance 1.8 inline refactoring of `emptyMap` variable produces compilation error at line `Integer first= Collections.emptyMap(/*nada*/).values().iterator().next();`

**In**
```
import java.util.Collections;
import java.util.Map;

class A extends Sup<String> {
	public A(Map<String, Integer> map) {
		Map<String, Integer> emptyMap= Collections.emptyMap(/*nada*/);
		method(emptyMap);
		method2(emptyMap);
		method3(emptyMap);
		new A(emptyMap);
		super.sup(emptyMap);
		
		Map<String, Integer> emptyMap2= emptyMap;
		emptyMap2= emptyMap;
		Object o= emptyMap;
		Map<? extends String, ?> emptyMap3= emptyMap;
		
		Integer first= emptyMap.values().iterator().next();
	}

	void method(Map<String, Integer> map) { }
	void method2(Map<? extends String, ?> map) {	}
	void method3(Object map) {	}
}

class Sup<S> {
	<A extends S, B> void sup(Map<A, B> map) {}
}
```

**Out after "inline local variable"**
```
import java.util.Collections;
import java.util.Map;

class A extends Sup<String> {
	public A(Map<String, Integer> map) {
		method(Collections.emptyMap(/*nada*/));
		method2(Collections.emptyMap(/*nada*/));
		method3(Collections.emptyMap(/*nada*/));
		new A(Collections.emptyMap(/*nada*/));
		super.sup(Collections.emptyMap(/*nada*/));
		
		Map<String, Integer> emptyMap2= Collections.emptyMap(/*nada*/);
		emptyMap2= Collections.emptyMap(/*nada*/);
		Object o= Collections.emptyMap(/*nada*/);
		Map<? extends String, ?> emptyMap3= Collections.emptyMap(/*nada*/);
		
		Integer first= Collections.emptyMap(/*nada*/).values().iterator().next(); // <----- ERROR here
	}

	void method(Map<String, Integer> map) { }
	void method2(Map<? extends String, ?> map) {	}
	void method3(Object map) {	}
}

class Sup<S> {
	<A extends S, B> void sup(Map<A, B> map) {}
}
```

jjohnstn:
- fixes #2775

## What it does

Fixes UI refactoring test failures after moving to Java 1.8.

## How to test

See issue.

## Author checklist


jjohnstn:
Needs rework

stephan-herrmann (commented):
I plan to revert the second commit and amend the change in DefaultBindingResolver, OK?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ParameterizedGenericMethodBinding.java
```diff
@@ -344,6 +344,7 @@ else if (invocationSite instanceof ReferenceExpression)
 								((ReferenceExpression) invocationSite).registerInferenceContext(methodSubstitute, infCtx18); // keep context so we can finish later
 						}
 					}
+					methodSubstitute.inferredReturnType= invocationTypeInferred;
```
stephan-herrmann, line 347: these two variables actually mean slightly different things:
* `inferredReturnType`: was the return type determined by type inference?
* `invocationTypeInferred`: has invocation type inference completed?

jjohnstn:
> I plan to revert the second commit and amend the change in DefaultBindingResolver, OK?

Please do.

stephan-herrmann:
@jjohnstn with my pending fix I see 3 failures in InlineTempTests: test32, test37, test38.

It looks like answering true to isResolvedTypeInferredFromExpectedType() is correct here, but that still doesn't mean that explicit type arguments are necessary, because after inlining the same target type will be present, to let inference succeed still. Does refactoring have any additional checks for this situation, or did it expect a false from   isResolvedTypeInferredFromExpectedType() also for these cases?

jjohnstn:
Hi @stephan-herrmann, I figured there was additional logic required on the UI side as I ran into those tests myself.

The following code is where the indirect check for the flag is made:

```
		if (Invocations.isResolvedTypeInferredFromExpectedType(initializer)) {
			if (!(referenceContext instanceof VariableDeclarationFragment)
					&& !(referenceContext instanceof SingleVariableDeclaration)
					&& !(referenceContext instanceof Assignment)) {
				ITypeBinding[] typeArguments= Invocations.getInferredTypeArguments(initializer);
				if (typeArguments != null) {
					String newSource= createParameterizedInvocation(initializer, typeArguments, rewrite);
					return (Expression) rewrite.getASTRewrite().createStringPlaceholder(newSource, initializer.getNodeType());
				}
			}
		}

```

Is there some existing additional calls the UI code can make to verify if the same target type will result?  If not, one way to solve this is to use the current refactoring logic which already verifies the refactoring by compiling the changed code and checks if any new problems are introduced.  We could mark the method invocation as possibly needing type parameters and then if the regular compile check reveals new problems at those locations, do the modifications and compile again to do a recheck for problems.  Not ideal, but it should work.  Judging by the few tests that fail, the recompile shouldn't be needed often.

stephan-herrmann:
> Hi @stephan-herrmann, I figured there was additional logic required on the UI side as I ran into those tests myself.
> 
> The following code is where the indirect check for the flag is made: [...]

This indeed looks incomplete vis-a-vis the mentioned tests.

> Is there some existing additional calls the UI code can make to verify if the same target type will result?

I'm afraid not. Even if the expression is inlined as an argument of an invocation where the corresponding parameter seems to supply the desired target type, this target type could have been computed during type inference based on the actual argument:

```
A<T> a = someExpr;
B b = m(a);
```
When inlining `a` and if `m` involves type inference, then `B b= m(someExpr);` could produce a different inference result or even fail, although inspection of `m(a)` shows a parameter of type `A<T>` as the target type.

> If not, one way to solve this is to use the current refactoring logic which already verifies the refactoring by compiling the changed code and checks if any new problems are introduced. We could mark the method invocation as possibly needing type parameters and then if the regular compile check reveals new problems at those locations, do the modifications and compile again to do a recheck for problems. Not ideal, but it should work. Judging by the few tests that fail, the recompile shouldn't be needed often.

This sounds good and appropriate.

Issue #2782, opened 2024-08-01
[23] ECJ Crashes with CCE


jarthana:
Trying to compile this code with the BETA_JAVA23 branch, the compiler crashes:
```
package test;	
public class StringVar2 {
  public void test () {
    "foo".
    if (true);
  }
}
```
This is happening after [the merger of this PR](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2685)

For the record, this is a testcase from one of the jdt.ui tests - `org.eclipse.jdt.text.tests.contentassist.PostFixCompletionTest.testStringVar2()`

@stephan-herrmann , can you please take a look or provide some pointers so I can take a look.

stephan-herrmann:
sorry, what exactly should I look at?
This is what I see:
```
$ ecj-j23.sh  -23 test/StringVar2.java 
----------
1. ERROR in /tmp/test/StringVar2.java (at line 5)
        if (true);
        ^^
Syntax error on token "if", super expected
----------
1 problem (1 error)
```

stephan-herrmann:
Weird, the parser produces:
```
public void test() {
  "foo".super(true);
}
```

During batch compilation the syntax error causes the method body to remain empty.

stephan-herrmann:
github didn't understand the "Fixes" stance in the PR? Closing manually.

PR #2785, opened 2024-08-01
[23] ECJ Crashes with CCE #2782

stephan-herrmann:
+ do expect ExplicitConstructorCall even in regular method

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2782


jarthana (approved):

jarthana:
I don't know what's causing the order change. The test pass in my local set-up too.

stephan-herrmann:
> I don't know what's causing the order change. The test pass in my local set-up too. Can we try `assertProblemMarkers()` instead of `assertMarkers()`?

I didn't see your suggestion, but in effect, it's the same `assertProblemMarkers()` == `sortMarkers() + assertMarkers()` :)

Issue #2758, opened 2024-07-25
Investigate tests disabled during transition to minimal 1.8 target


iloveeclipse:
Commit https://github.com/eclipse-jdt/eclipse.jdt.core/commit/1bcb15a140b9989d0f33b2552989af431376ede7 disabled few failing tests that started to fail after changing minimal compilation level to 1.8 in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2551

Some of disabled tests have already own tickets, the rest must be analyzed & either tests updated or removed or code fixed.

To find tests in the workspace, just search for `_2551_`.
Current list has 21 (2024-12-31) matches left, here the overview:

- [x] ASTConverter15JLS4Test.java - org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom (2 matches)                                             
- [x] ASTConverter15JLS8Test.java - org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom (2 matches)                                             
- [x] ASTConverter15Test.java - org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom (2 matches)                                                 
- [ ] ASTConverter17Test.java - org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom (6 matches)                                                 
- [x] ASTConverter18Test.java - org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom (2 matches)                                                 
- [x] ASTConverterTest2.java - org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom                                                              
- [x] ASTConverterTestAST4_2.java - org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom (7 matches)                                             
- [x] AttachSourceTests.java - https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2754                                                         
- [ ] BatchASTCreationTests.java - org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom                                                          
- [x] ClasspathTests.java - org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model                                                               
- [ ] CompletionTests.java - org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model                                                              
[...]

Issue #2716, opened 2024-07-16
ClasspathTests / JavaModelTests seem to be unstable


iloveeclipse:
See https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2551/checks?check_run_id=27502347557

Two tests failed (fully unrelated to the changes on the branch https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2551), common is that something basic what the test setup did in the project/file system wasn't "seen" by the test assertions moments later. 

I have not yet an explanation for that, except there are some jobs running in the background and we run into some asynchronous update issues...

**JavaModelTests.testGetJavaProjects1**
```
junit.framework.AssertionFailedError: Project P should be present
	at junit.framework.Assert.fail(Assert.java:57)
	at junit.framework.Assert.assertTrue(Assert.java:22)
	at junit.framework.TestCase.assertTrue(TestCase.java:192)
	at org.eclipse.jdt.core.tests.model.JavaModelTests.testGetJavaProjects1(JavaModelTests.java:402)
```

**ClasspathTests.testUnknownElements1**
```
junit.framework.ComparisonFailure: 
Unexpected content.
----------- Expected ------------
<?xml version="1.0" encoding="UTF-8"?>\n
<classpath>\n
	<classpathentry kind="src" path="src2"/>\n
	<classpathentry kind="src" path="src1">\n
		<unknown>\n
			<test kind=""/>\n
		</unknown>\n
	</classpathentry>\n
	<classpathentry kind="output" path="bin"/>\n
</classpath>\n

------------ but was ------------
<?xml version="1.0" encoding="UTF-8"?>\n
<classpath>\n
	<classpathentry kind="var" path="JCL18_LIB"/>\n
	<classpathentry kind="src" path=""/>\n
	<classpathentry kind="output" path=""/>\n
</classpath>\n

---------------------- ----------
 expected:<...lasspathentry kind="[src" path="src2"/>\n
	<classpathentry kind="src" path="src1">\n
		<unknown>\n
			<test kind=""/>\n
		</unknown>\n
	</classpathentry>\n
	<classpathentry kind="output" path="bin]"/>\n
</classpath>\n...> but was:<...lasspathentry kind="[var" path="JCL18_LIB"/>\n
	<classpathentry kind="src" path=""/>\n
	<classpathentry kind="output" path="]"/>\n
</classpath>\n...>
[...]

PR #2789, opened 2024-08-03
[#2758] Check tests disabled during transition to 1.8 minimal target

zelenyhleb:
Hi folks.
For `org.eclipse.jdt.core.tests.compiler.regression.Compliance_1_4` two test cases were enabled and aligned with the current state-of-thigs:

### test080 
Removed expected error since it was produced because of a [bug](https://bugs.java.com/bugdatabase/view_bug?bug_id=5080917) that was fixed before 1.8 release, and error is not shown in future compiler versions. For 1.4 the problem still exists, we can still notify user about with custom validation or something, but it seems to be not needed if 1.4 is no longer supported.

### test088
Added raw types warnings to expected output, generics were added in Java 1.5 and using raw types produce warnings. Same for test080.
<hr>
IMO, warnings from 1.8 compiler make no sense for 1.4 code and you won't be even able to fix them. So if we say that older java versions are no longer supported, wouldn't it be better to just remove the whole regression test class (and also other old compliance tests)? 

How do we benefit from testing not supported feature? WDYT, @iloveeclipse?

Thank you.

### DIFF
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/Compliance_1_4.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/Compliance_1_4.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2019 IBM Corporation and others.
+ * Copyright (c) 2000, 2019, 2024 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -11,6 +11,7 @@
  * Contributors:
  *     IBM Corporation - initial API and implementation
  *     Stephan Herrmann <stephan@cs.tu-berlin.de> - Contribution for bug 185682 - Increment/decrement operators mark local variables as read
+ *     Nikifor Fedorov <nikifor.fedorov@arsysop.ru> - Bug #2758 - expected compiler outputs extended with 1.8 warnings
  *******************************************************************************/
 package org.eclipse.jdt.core.tests.compiler.regression;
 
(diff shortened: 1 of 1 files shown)


iloveeclipse:
> So if we say that older java versions are no longer supported, wouldn't it be better to just remove the whole regression test class (and also other old compliance tests)?

The reason I've tried to keep all tests is that I haven't analyzed whether we will get a coverage problem - the remaining tests might not cover functionality that were tested by "old compliance" tests. Also 1.4 compliance test doesn't automatically mean the test is irrelevant now, simply because most of 1.x Java spec is still in 20+ Java spec.

Of course if you see same functionality covered by other tests, then "older" test versions could be removed.

zelenyhleb:
Test failure is unrelated (see #2716)

zelenyhleb:
It seems I am lucky to have unstable tests always failing in my MRs, same as when I tried to contribute to platform a year ago 😕
Can this be related to some negligible thing like branch name (it starts with digits)?

iloveeclipse:
> Can this be related to some negligible thing like branch name (it starts with digits)?

No idea, but I guess it is something else.

zelenyhleb:
> it is something else

Definitely

zelenyhleb:
Hi @iloveeclipse! 
Is there still something what you expect from me to do in scope of this MR?

iloveeclipse:
Sorry, I was on vacation, I will check this PR now. Please rebase on latest master state.

zelenyhleb:
Done @iloveeclipse

iloveeclipse:
OK, lgtm. The only problem now is that master branch is closed as we are in 4.33 RC2 phase and we prepare for 4.34 development. Once master will be open, we can merge.

iloveeclipse:
@zelenyhleb : sorry for the long wait time and thanks for contribution.

PR #2549, opened 2024-06-11
Boxing conversion

mpalat:
## What it does
Boxing part of issue #2298 
## Author checklist

- [ ] I have thoroughly tested my changes
- [ ] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [ ] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

Issue #2298, opened 2024-04-08
[23] JEP 455: Primitive Types in Patterns, instanceof, and switch (Preview)


mpalat:
ref: https://openjdk.org/jeps/455

Summary
Enhance pattern matching by allowing primitive type patterns in all pattern contexts, and extend instanceof and switch to work with all primitive types.

spec delta: https://cr.openjdk.org/~abimpoudis/instanceof/latest/

srikanth-sankaran:
Please include me in all design and implementation discussions and reviews.

As a first step, I suggest we figure out if parser/scanner changes are required by capturing all snippets from the JEP document into test cases - do we have any parse failures if we do that ?

jarthana:
> As a first step, I suggest we figure out if parser/scanner changes are required by capturing all snippets from the JEP document into test cases - do we have any parse failures if we do that ?

Is there an easy/standard way to do that or are you saying we just look for syntax errors?

srikanth-sankaran:
> > As a first step, I suggest we figure out if parser/scanner changes are required by capturing all snippets from the JEP document into test cases - do we have any parse failures if we do that ?
> 
> Is there an easy/standard way to do that or are you saying we just look for syntax errors?

We can see what errors we get, or better yet whether diagnose parse gets entered while compiling the snippets.

stephan-herrmann:
> Please include me in all design and implementation discussions and reviews.

 ... and include me on all spec issues :)

srikanth-sankaran:
When this is scheduled for standardization I will assume ownership - for now, please feel free to engage me for any design or implementation consultation and code review as needed. Thanks

stephan-herrmann:
> When this is scheduled for standardization I will assume ownership

by "standardization" you mean having a [stable spec](https://cr.openjdk.org/~iris/se/23/spec/pr/java-se-23-pr-spec-01/specs/instanceof-jls.html), or maturing from preview to standard?

srikanth-sankaran:
> > When this is scheduled for standardization I will assume ownership
> 
> 
> 
> by "standardization" you mean having a [stable spec](https://cr.openjdk.org/~iris/se/23/spec/pr/java-se-23-pr-spec-01/specs/instanceof-jls.html), or maturing from preview to standard?

I meant the latter.

mpalat:
[...]

PR #2790, opened 2024-08-04
JEP 455: instanceof involving boxing & test reorganization

stephan-herrmann:
Implementation:
* boxing plus widening reference conversions
* boxing in patterns
 
Testing:
* generic complete tests for primitive widening & narrowing
* generic tests for tests involving boxing
* enable run.javac with 23-ea

based on #2549 (squashed and rebased).

addresses https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2298

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/InstanceOfExpression.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/InstanceOfExpression.java
@@ -39,6 +39,7 @@ import org.eclipse.jdt.internal.compiler.codegen.*;
 import org.eclipse.jdt.internal.compiler.flow.*;
 import org.eclipse.jdt.internal.compiler.impl.CompilerOptions;
 import org.eclipse.jdt.internal.compiler.impl.Constant;
+import org.eclipse.jdt.internal.compiler.impl.JavaFeature;
 import org.eclipse.jdt.internal.compiler.lookup.*;
 
 public class InstanceOfExpression extends OperatorExpression {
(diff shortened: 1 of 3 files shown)


stephan-herrmann:
@mpalat here is a proposal that might replace #2549, containing these changes:
* slight implementation clarification, see https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2549/commits/432b4118f9dc81950c0876f52e582f506de095c0
* test re-organization: fold lots of plain tests into a few generic tests
  * IMHO this makes it easier to vouch for complete coverage
  * tests run significantly faster (because launching each individual test is costly).

@srikanth-sankaran do you want to comment on the strategy for tests? I'm aware that typically tests should be straight forward to challenge exactly one thing at a time with little risk of messing up by complexity in the test code. In this case I simply lost track among 100+ tests, that all looked very similar. That's why I figured that something like matrix testing (widen all primitives to all applicable other primitives etc. multiplied with different styles of instanceof expressions) might rather be programmed than enumerated explicitly. Are you fine with this style?

mpalat:
@stephan-herrmann my personal preference would be to have the test in the original form such that it tests only one part. However, since @srikanth-sankaran wants to own this feature later, will wait for his take.

stephan-herrmann:
> [...] such that it tests only one part.

I won't object to that aim, but we seem to disagree on what is "one part". Where I see one test case (e.g.: test primitive identity conversions) you seem to see 24(!) relevantly different tests. One way or other, I think we do need more structure to have at least some chance to detect duplicate tests and omissions.

> However, since @srikanth-sankaran wants to own this feature later, will wait for his take.

:+1:

stephan-herrmann:
Some code changes have been merged via other PRs, comments on the code are (a) moved to https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2880, (b) no longer relevant.

Re-organized tests have been merged additionally in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2878

We're done.

PR #2816, opened 2024-08-12
remove recursion from VMInputStream

hduelme:
## What it does
Currently `VMInputStream` uses recursion to retry reads in the `int read()` method. I replaced the recursive call with a while loop. This removes unnecessary overhead while running tests.

## How to test
This doesn't change any behavior.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/runtime/VMInputStream.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/runtime/VMInputStream.java
@@ -57,14 +57,16 @@ public boolean markSupported() {
 }
 @Override
 public int read() throws IOException {
-	try {
-		return this.input.read();
-	} catch (IOException e) {
-		if (isRunning()) {
-			return read();
-		}
-		throw e;
+	while (true) {
+		try {
+			return this.input.read();
+		} catch (IOException e) {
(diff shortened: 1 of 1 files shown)


hduelme:
I did some testing. If the build is successful no recursive calls happen. Only if something is written to the `errorStream` the `IOException` is thrown. In this case a found between 10 to 500 recursive calls.

stephan-herrmann:
> In this case a found between 10 to 500 recursive calls.

What's wrong with that?

When you said "overhead" I was expecting some time measurements... a recursive call is not overhead per se.

hduelme:
I agree. I was also expecting different results. I got here in the first place due to a exception with a large stacktrace

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/runtime/VMInputStream.java
```diff
@@ -57,14 +57,16 @@ public boolean markSupported() {
...
 @Override
 public int read() throws IOException {
-	try {
-		return this.input.read();
-	} catch (IOException e) {
-		if (isRunning()) {
-			return read();
-		}
-		throw e;
+	while (true) {
```
jukzi, line 60: So this would be an endless loop instead of Stackoverflow? Doesn't sound better, but yes the recursion seems wrong too.

Issue #2808, opened 2024-08-11
[23] markdown javadoc should detect code regions


stephan-herrmann:
Follow-up after #2738:

In markdown two ways exist to mark a block of code:
* indentation
* "fenced" within a pair of ` ``` `

Parsers should recognize these regions and in particular avoid parsing contained `@` prefixed words as tags.

See the second example in https://openjdk.org/jeps/467#JavaDoc-tags

stephan-herrmann:
In order to correctly detect indented blocks, we need to keep track of relative amount of whitespace between `///` and text.

According to https://openjdk.org/jeps/467#Syntactical-details individual lines of a markdown block are normalized to start at the start column of the least indented line. 

According to those rules the following comment should start with a code region:
```
	///     void m() {}
	///
	/// Plain text
```
The javadoc tool, however, seems to have problems with code blocks at the beginning or end of a comment, the above is rendered as plain text (in contrast to github rendering). To achieve compatibility it might suffice if we detect _increase_ of indentation without worrying about subsequent lines with less leading whitespace.

stephan-herrmann:
> The javadoc tool, however, seems to have problems with code blocks at the beginning or end of a comment, the above is rendered as plain text (in contrast to github rendering).

This was not quite correct, that example is rendered correctly by javadoc. Apparent bugs are:
* javadoc doesn't accept a Java annotation (like `@Override`) as the first thing within a markdown comment, despite correct indentation
* javadoc silently drops a trailing indented region

stephan-herrmann:
I have local changes that ensure that significant whitespace is actually part of the DOM TextElements.  With this and with specific parsing of ` ``` ` both kinds of code blocks will be parsed without interpreting `@tags` and with proper rendering on the markdown side.

Also locally fixed: detecting more than 3 `/`, to include any excess slashes in TextElements, too.

To somewhat decouple these changes from standard AbstractCommentParser, I created a little companion class that keeps track of markdown related state.

@jarthana fyi. With these local changes all new tests in https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1587 pass :)

stephan-herrmann:
@jarthana  another fyi: while preparing a PR for this issue, I saw my changes causing a few regressions in (disabled) ASTConverterMarkdownTest. While trying to avoid such regressions I'm making a few obvious changes in those tests. So for your next round on those tests you may want to wait just a little while, OK?

[...]

PR #2821, opened 2024-08-14
[23] markdown javadoc should detect code regions

stephan-herrmann:
+ finetune parsing start of line by new companion IMarkdownCommentHelper
  + detect excess slashes after '///'
  + determine minimum amount of whitespace per markdown comment
  + preserve additional whitespace at beginning of line

Also:
+ avoid showing closing ']' of links
+ resolve one warning (redundant null check)
+ fix and extend compliance settings in RunCompletionParserTests
+ fix & extend test source in ../Converter_23/src/markdown/testBug228648

fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2808

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/AbstractCommentParser.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/AbstractCommentParser.java
@@ -421,7 +421,7 @@ public abstract class AbstractCommentParser implements JavadocTagConstants {
 							}
 						} else if (this.lineStarted && isDomParser) {
 							textEndPosition = this.index;
-						} else if (this.markdownHelper.recordSignificantLeadingSpace()) {
+						} else if (!this.lineStarted && this.markdownHelper.recordSignificantLeadingSpace()) {
 							if (this.textStart == -1)
 								this.textStart = this.index; // first relevant whitespace is start of text
 						}
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/AbstractCommentParser.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/AbstractCommentParser.java
@@ -105,6 +105,8 @@ public abstract class AbstractCommentParser implements JavadocTagConstants {
 	private int nonRegionTagCount, inlineTagCount;
 	final static String SINGLE_LINE_COMMENT = "//"; //$NON-NLS-1$
 
+	protected IMarkdownCommentHelper markdownHelper;
+
 	// Line pointers
 	private int linePtr, lastLinePtr;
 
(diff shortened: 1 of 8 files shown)


stephan-herrmann:
@jarthana once this PR is merged, it would be cool to have another Y-build, also incorporating the latest merge master->BETA in jdt.debug (thanks!). Based on the fresh Y-build also https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1472 will have good chances to build successfully :)

Apparently, Y-builds aren't run regularly (yet)? Who owns the big button to trigger a Y-build?

jarthana:
> @jarthana once this PR is merged, it would be cool to have another Y-build, also incorporating the latest merge master->BETA in jdt.debug (thanks!). Based on the fresh Y-build also [eclipse-jdt/eclipse.jdt.ui#1472](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1472) will have good chances to build successfully :)
> 
> Apparently, Y-builds aren't run regularly (yet)? Who owns the big button to trigger a Y-build?

I used to have the permission, but not anymore. Now the rights are with @MohananRahul. Anyway, I see that we have a new successful Y build yesterday. I will grab and that play around with markdown on editor.

jarthana:
Stephan, something I noticed with the Y build - if we have two sections of markdown comments (separated by an empty line), we seem to be consider the first one also as part of the following element. Is this intentional? I recollect the JEP saying the opposite - i.e. only the most recent section should be attached to the element. Javadoc tool agrees with that too. For e.g:

```
///
/// 
/// @param an array of String

///
```
produces the following in Javadoc view:
```
Parameters:
array of String
args
```

I am not sure I added a test for this. So, I am wondering if this was happening already after my work going in.

jarthana:
> Stephan, something I noticed with the Y build - if we have two sections of markdown comments (separated by an empty line), we seem to be consider the first one also as part of the following element. Is this intentional? I recollect the JEP saying the opposite - i.e. only the most recent section should be attached to the element. Javadoc tool agrees with that too. For e.g:

OK, this happens only when we have an empty line with a markdown beginning. I will investigate in another issue. Might have been an issue all along since my first patch.

stephan-herrmann:
_(your comment came in while I was writing this, submitting any way :) )._

Thanks, @jarthana  for playing with this. Apparently, handling the beginning and end of a comment is particularly difficult, see also https://mail.openjdk.org/pipermail/javadoc-dev/2024-August/007884.html

Two observations regarding your example:

(1) I see an error "Javadoc: Parameter an is not declared" which explains that "args" appears after the description, since the description cannot be associated to the parameter.

(2) If you only add one character (may be a blank) to the empty comment line, then the preceding comment is correctly dropped. 
  
So the problem boils down to the corner case that a syntactically valid comment block is follow by (a) a blank line and then (b) an empty markdown comment, empty as in "zero chars content".

The problem can alternatively be avoided by removing the space between `///` and `@param`. This tells me: we determine that markdown content starts after `///` plus one blank. In that sense the empty comment line actually has `-1` chars of content :)

Compared to the obvious bugs in javadoc I'd consider this a very exotic example.

stephan-herrmann:
I could narrow it down to `org.eclipse.jdt.internal.core.Member.getJavadocRange()` answering the range for the first block, rather then the empty second. This again can be traced back to this break: https://github.com/eclipse-jdt/eclipse.jdt.core/blob/9b33f168119500e61bc71845b62a3cde77721d5e/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java#L1953

Not having seen any content we simply don't produce `TokenNameCOMMENT_MARKDOWN`. The following insertion (partial copy from below) before the break seems to fix this:
```
						int token = TokenNameCOMMENT_MARKDOWN;
						recordComment(token);
						if (this.tokenizeComments) {
							return token;
						}
```

I have no idea if this will break anything.

Issue #2824, opened 2024-08-14
[23] Parsing details for markdown comments


stephan-herrmann:
Part of #2429:
* ensure code blocks are detected only (a) at the beginning of a comment, or (b) after a blank line
* proper escaping of array syntax `[]` within links (`see [#meth(String[])]`)
* check handling of different styles of code fences
* ...

PR #2831, opened 2024-08-17
[23] Parsing details for markdown comments #2824

stephan-herrmann:
+ codeblock cannot interrupt a paragraph
+ separate handling of code blocks indented vs fenced
  + fence does not terminate indented code block
  + handle various forms of fences:
    + \`  vs. ~
    + different fence lengths
    + no mixed fences
    + fences inside fenced region
+ handle (and require) escaping of `[` `]` inside references (method arguments)

fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2824


stephan-herrmann:
Note: since in DOM tests we cannot see how code block is handled in the end, I will add corresponding tests also to org.eclipse.jdt.ui.tests.hover.MarkdownCommentTests.

Issue #2834, opened 2024-08-17
[23] DOM support for JEP 476: Module Import Declarations (Preview)


stephan-herrmann:
The tiny feature https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1589 has a tiny follow-up https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1589 where all the sudden it seems we need to extend the DOM AST:

In a regular compilation unit `module` is not a keyword and hence it would be suitable to apply semantic highlighting (category restricted identifiers). But then we'd need an ASTNode that informs us about the position of the modifier.

It's a real pity that introduction of `import static` was answered just with a boolean flag in `ImportDeclaration`. At this point we'd be better off with a list of `Modifier`.

I have a quick-n-dirty draft of such addition, but I'd like to discuss this approach first, because now it looks a bit odd to have `isStatic()` _and_ `modifiers()`. Should the latter answer a `Modifier` also for static?? Should we migrate the internal storage to list of Modifier and only map `isStatic()` and `setStatic()` to corresponding acrobatics? This looks similar to how `Modifier` lists were introduced for JLS3 in the first place.

@jarthana @mpalat  wdyt?

stephan-herrmann:
If module imports were a strongly debated feature I'd say let's wait until it leaves preview state, but specifically with its connection to implicitly declared classes I guess it's safe to assume that this feature will stay.

jarthana:
> I have a quick-n-dirty draft of such addition, but I'd like to discuss this approach first, because now it looks a bit odd to have `isStatic()` _and_ `modifiers()`. Should the latter answer a `Modifier` also for static?? Should we migrate the internal storage to list of Modifier and only map `isStatic()` and `setStatic()` to corresponding acrobatics? This looks similar to how `Modifier` lists were introduced for JLS3 in the first place.

Yes, we can may be have a general modifiers() and perhaps a retirement plan for the isStatic() and family. It will be redundant for a while, but definitely future-proof.

PR #2836, opened 2024-08-19
[23] DOM support for JEP 476: Module Import Declarations (Preview)

stephan-herrmann:
+ parser to record modifier start position of imports
+ avoid code duplication in SourceElementParser
+ new list ImportDeclaration.modifiers()
  + at 23 even 'static' is represented in that list
+ existing accessors may scan that list if present

fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2834

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ImportReference.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ImportReference.java
@@ -33,6 +33,7 @@ public class ImportReference extends ASTNode {
 	public int declarationSourceStart;
 	public int declarationSourceEnd;
 	public int modifiers; // 1.5 addition for static imports
+	public int modifiersSourceStart;
 	public Annotation[] annotations;
 	// star end position
 	public int trailingStarPosition;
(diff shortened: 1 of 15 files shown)


stephan-herrmann:
@jarthana while this already produces desired results (and passes even JDT/UI tests), I'm a short in additional tests. From the experience of previous AST additions, can you recommend which test would be good to extend here?

stephan-herrmann:
Documenting one oddity:

To cleanly separate strategies before vs since JLS23 I had to let JLS23-mode represent *all* modifiers using list `modifiers()`.
* `ASTConverter` can use valid source positions (thanks to a small addition in Parser) to create the `Modifier` instance
* existing clients like `CreateImportOperation` still want to use `setStatic(boolean)`
  * in JLS23 mode this now synthesizes a `Modifier` *without* source positions

For code manipulation those synthesized `Modifier`s work fine, we just need to be sure, that `setStatic(boolean)` is never called in a situation where valid source positions are needed (I made a corresponding note in the method's javadoc).

stephan-herrmann:
> @jarthana while this already produces desired results (and passes even JDT/UI tests), I'm a short in additional tests. From the experience of previous AST additions, can you recommend which test would be good to extend here?

I'm still open to suggestions :)

Issue #2815, opened 2024-08-12
ephemeral memory allocation in JRTUtil.getJrtSystem()


jukzi:
Currently JDK Mission Control reports JRTUtil.getJrtSystem() as top most memory allocator during full build

caused by https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2370
see also https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2394#issuecomment-2109776599 ff

![image](https://github.com/user-attachments/assets/a7c2c126-9982-4aab-829b-52ca28f32995)

HannesWell:
> caused by #2370

How can a PR that was not submitted cause a regression? :P
I guess you mean https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2376, don't you?

> see also [#2394 (comment)](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2394#issuecomment-2109776599) ff

As written in the ff, if possible it would be best to save a reference to the file-system at an appropriated place. This would save also the look-ups or at least use Path all the time.

In general it would be good to mention the build you monitored, this allows other to reproduce the results. One of the Eclipse top-level projects would of course be handy, maybe even the entire SDK.

But from the screenshot you provided I read that `org.eclipse.jdt.internal.compiler.util.lJdk.toJdkHome()` is the main cause isn't it?
That method calls `Path.getParent()` twice since its inception in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/217.
So it doesn't look like a regression.

jukzi:
There was a chain of multiple PRs in that area....
My testcase was simply setting up a workspace as descripted in https://github.com/eclipse-platform/.github/blob/main/CONTRIBUTING.md and measuring Project/clean.../"Clean all projects".

@HannesWell do you plan to work on it? Then i would assign it to you. Otherwise i would just leave this open as a reminder that there is room for improvement. It's not urgent but JDT is much optimized for avoiding memory allocations and it's absurd that now this code allocates more memory then all the sourcecodes compiled. 

> So it doesn't look like a regression.

I trust in measurements more then in speculations.

HannesWell:
> @HannesWell do you plan to work on it? Then i would assign it to you. Otherwise i would just leave this open as a reminder that there is room for improvement. It's not urgent but JDT is much optimized for avoiding memory allocations and it's absurd that now this code allocates more memory then all the sourcecodes compiled.

[...]

PR #2837, opened 2024-08-20
cache JrtFileSystem in ClasspathJrt.jrtFileSystem #2815

jukzi:
to avoid ephemeral memory allocations

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2815

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
@@ -14,7 +14,6 @@
  *******************************************************************************/
 package org.eclipse.jdt.internal.compiler.util;
 
-import java.io.ByteArrayInputStream;
 import java.io.File;
 import java.io.FileNotFoundException;
 import java.io.IOException;
@@ -22,7 +21,6 @@ import java.io.InputStream;
 import java.io.UncheckedIOException;
 import java.net.URI;
 import java.nio.channels.ClosedByInterruptException;
-import java.nio.file.DirectoryStream;
 import java.nio.file.FileSystem;
 import java.nio.file.FileSystemAlreadyExistsException;
 import java.nio.file.FileSystemNotFoundException;
@@ -34,15 +32,11 @@ import java.nio.file.Path;
 import java.nio.file.ProviderNotFoundException;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Predicate;
-import java.util.stream.Stream;
 
 import org.eclipse.jdt.internal.compiler.classfmt.ClassFileReader;
 import org.eclipse.jdt.internal.compiler.classfmt.ClassFormatException;
@@ -236,16 +230,11 @@ public class JRTUtil {
 	 * @param visitor an instance of JrtFileVisitor to be notified of the entries in the JRT image.
 	 * @param notify flag indicating the notifications the client is interested in.
 	 */
-	public static void walkModuleImage(File image, final JRTUtil.JrtFileVisitor<Path> visitor, int notify) throws IOException {
-		JrtFileSystem system = getJrtSystem(image, null);
-		if (system == null) {
-			return;
-		}
-		system.walkModuleImage(visitor, notify);
+	public static void walkModuleImage(File image, JRTUtil.JrtFileVisitor<Path> visitor, int notify) throws IOException {
+		walkModuleImage(getJrtSystem(image, null), visitor, notify);
 	}
 
-	public static void walkModuleImage(File image, String release, final JRTUtil.JrtFileVisitor<Path> visitor, int notify) throws IOException {
-		JrtFileSystem system = getJrtSystem(image, release);
+	public static void walkModuleImage(JrtFileSystem system, JRTUtil.JrtFileVisitor<Path> visitor, int notify) throws IOException {
 		if (system == null) {
 			return;
 		}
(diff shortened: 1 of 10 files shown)


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JrtFileSystem.java
```diff
@@ -0,0 +1,322 @@
+package org.eclipse.jdt.internal.compiler.util;
```
stephan-herrmann, line 1: Is this class moved verbatim or did you change any details? 

If the latter, could you create the change in two steps, where the first is just the verbatim move, so we can see the modifications in the history?

stephan-herrmann:
Since the change is not small, I'd very much appreciate a description of your strategy. What is being improved and how?

jukzi:
> Since the change is not small, I'd very much appreciate a description of your strategy. What is being improved and how?

Don't know what it not clear: instead of calling JRTUtil.method(File, ... ) i outlined and used JRTUtil.method(JrtFileSystem, ... ). The JrtFileSystem used is eager created in advance in the constructor ClasspathJrt(String zipFilename) and stored in ClasspathJrt.jrtFileSystem instead of just the Filename "jrtFile. Then it can be reused in subsequent calls. For example in the heavy used ClasspathJrt.isPackage().
I also added modifiers "private final" as much as possible for all members, which required to inline ClasspathJrtWithReleaseOption.initialize()

stephan-herrmann:
> > Since the change is not small, I'd very much appreciate a description of your strategy. What is being improved and how?
> 
> Don't know what it not clear: 

initially, when looking at the PR "nothing" is clear, since we can't read the intentions inside your head :)

> instead of calling JRTUtil.method(File, ... ) i outlined and used JRTUtil.method(JrtFileSystem, ... ).

I don't see "instead", but "additionally / optionally", meaning that only one client of `JRTUtil` is affected, others will still use the old strategy, right?

>  The JrtFileSystem used is eager created in advance in the constructor ClasspathJrt(String zipFilename) and stored in ClasspathJrt.jrtFileSystem instead of just the Filename "jrtFile. Then it can be reused in subsequent calls. For example in the heavy used ClasspathJrt.isPackage().

ok

> I also added modifiers "private final" as much as possible for all members, which required to inline ClasspathJrtWithReleaseOption.initialize()

thanks, that explains the other "big" change :)

FYI, I'm asking these questions mainly so we better know what's going on when later we have to revisit this code in another bug report.

jukzi:
Such questions are very welcome as they help me to challenge my code. Now about the benefit of this change: I took a measurement of building the platform workspace within the IDE. Sampling did yield this:
image
With this change the same usecase results in:
image
One can see ClasspathJrt.isPackage() is reduced by ~1 second and so does the the overall build time in org.eclipse.core.internal.events.BuildManager.basicBuildLoop(). So i am sure the time did not just move to the constructor :-)
Note that ClasspathJrt.findClass() is also reported to be reduced, but it includes a call to isPackage():
image

The changes assumes the JRT on the disk does does not change during the runtime. Due to the eager init there might be behavioral changes if the JRT is not found on disk. That should not happen anyway.

I did not improve the Batch compiler callers as i am not interested in that. Probably same optimization could be done for org.eclipse.jdt.internal.compiler.batch.ClasspathJrt.file

If there are no concerns i plan to submit next M1.

jukzi:
Regarding the batch compiler it seems straight forward to add a batch.ClasspathJrt.jrtFileSystem but i don't understand how/if it should be used as argument for
org.eclipse.jdt.internal.compiler.tool.EclipseFileManager.getJrtFileSystem(File)
and why the batch compiler never calls getJrtSystem(, release) with any non-null release as ClasspathJrtWithReleaseOption does.

stephan-herrmann:
> why the batch compiler never calls getJrtSystem(, release) with any non-null release

have you seen `FileSystem.getOlderSystemRelease()`?

Also, looking at `initialize()` methods being called in constructors of  `..batch.ClasspathJrt` and subclasses I don't see any problem with memory / initialization. Do you?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JrtFileSystem.java
HannesWell, line 1: Please add a Copyright header for this file. It should consider when and by whom `JrtFileSystem` was added.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
```diff
@@ -235,16 +229,11 @@ public static void reset() {
...
-			return;
-		}
-		system.walkModuleImage(visitor, notify);
+	public static void walkModuleImage(File image, JRTUtil.JrtFileVisitor<Path> visitor, int notify) throws IOException {
+		walkModuleImage( getJrtSystem(image, null), visitor, notify);
 	}
 
-	public static void walkModuleImage(File image, String release, final JRTUtil.JrtFileVisitor<Path> visitor, int notify) throws IOException {
-		JrtFileSystem system = getJrtSystem(image, release);
+	public static void walkModuleImage(JrtFileSystem system, JRTUtil.JrtFileVisitor<Path> visitor, int notify) throws IOException {
```
HannesWell, line 236: Is it really necessary to keep these static methods that just call a similarly named method on the first argument here and below?
Some handle the case where the `JRTFileSystem` is null, but I wonder if that would be better handled by the caller?
And some don't even do null-checks and just call the corresponding method on `JRTFileSystem`.

HannesWell (commented):
Thank you Jörg for this one.
I think this is going into the right direction.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
jukzi: I suggest you answer that question yourself by looking into the call hierarchy.

HannesWell: If I do that I see that one just has to change the visibility of the called method to public and then the same question arises, plus should that method not be public (together with the corresponding other methods)?

But in general, since I patiently answer your question and address your remarks on my PRs, I was hoping that you would return the favor and answer my questions/remarks on your PRs.

jukzi: There are 4 callers. i don't see a benefit to add the null check to all of them.

jukzi:
> > why the batch compiler never calls getJrtSystem(, release) with any non-null release
> 
> have you seen `FileSystem.getOlderSystemRelease()`?

they store release in ClasspathJep247.compliance, but never call getJrtSystem. i don't get why batch compiler have two different "ClasspathJrt" implementations.

> Also, looking at `initialize()` methods being called in constructors of `..batch.ClasspathJrt` and subclasses I don't see any problem with memory / initialization. Do you?

I don't see a problem in the source code but i have never used or even measured the batch compiler in any way. I will add a commit for changes that can be similar done in batch compiler  - without knowing how big the benefit will be there.

Issue #2838, opened 2024-08-20
 [23] Selection for module imports


stephan-herrmann:
It is not yet possible to F3-navigate from a module import to the module declaration.

PR #2839, opened 2024-08-20
[23] Selection for module imports

stephan-herrmann:
+ implement selection for module imports
+ includes new ResolveTests23
+ also add forgotten ResolveTests21 to the suite
  - but remove tests for withdrawn string template feature

fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2838


stephan-herrmann:
For posterity: this fix contains a tweak to ensure module names are treated as one name, regardless of any contained `.`. Otherwise selecting on `java` of `java.base` would pull up *all modules starting with "java"*. 

For references within `module-info.java` this same concept is already implemented in `SelectionParser.createAssistModuleReference(int)`

PR #2841, opened 2024-08-21
classfmt: fix "Implicit narrowing conversion" warnings

jukzi:
ClassFileStruct.u4At(int) is supposed to return an unsigned integer. The return value is currently always used as primitive integer (signed) to compute readOffset inside the class file (which will then passed as argument to the method again). So any negative value would trigger an ArrayIndexOutOfBoundsException when used as index into the byte array.

The intermediate usage of primitive long did not help and only triggered warnings at the call sites.

https://github.com/eclipse-jdt/eclipse.jdt.core/security/code-scanning/18 ff

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/classfmt/ClassFileStruct.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/classfmt/ClassFileStruct.java
@@ -54,9 +54,17 @@ public int u2At(int relativeOffset) {
 	int position = relativeOffset + this.structOffset;
 	return ((this.reference[position++] & 0xFF) << 8) | (this.reference[position] & 0xFF);
 }
-public long u4At(int relativeOffset) {
+
+/**
+ * Returns an 32 bit integer created of 4 bytes. The return value should be treated as an <strong>unsigned</strong> integer! i.e. if the caller does math with
+ * the result he should use the unsigned methods from Integer like {@link java.lang.Integer#toUnsignedLong(int)} or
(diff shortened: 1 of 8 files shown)


stephan-herrmann:
I understand that carelessly using `int` will cause trouble for class files beyond 2 GiB. Apparently ecj hasn't yet produced or read a file of that size :)

But to really fix this, wouldn't we need to use `long x` for every `this.reference[x]` (of which there are *plenty*)? Am I missing anything?

IOW, you might be resolving some warnings, but I doubt that you are fixing a bug with your approach.

jukzi:
Java arrays take only int as parameter. On top of that the java vm restricts array sizes to max int minus a hand full of bytes overhead. So jdt will never be able to read such a large class into an array. And i don't think there is any real world usecase for such. So this commit does not improve on that or fix any usecase but avoids needless cast to long and back to int. As a effect it will avoid the false positive warning.

stephan-herrmann:
> Java arrays take only int as parameter. On top of that the java vm restricts array sizes to max int minus a hand full of bytes overhead. So jdt will never be able to read such a large class into an array. And i don't think there is any real world usecase for such. So this commit does not improve on that or fix any usecase but avoids needless cast to long and back to int. As a effect it will avoid the false positive warning.

Thanks for explaining.

So in "_any negative value would trigger an ArrayIndexOutOfBoundsException_" you are discussing a purely hypothetical situation, right? I assume the situation would have blown up earlier as `NegativeArraySizeException` when trying to create that array.

jukzi:
The exception may also depend on if the classfile is just too big (OOME) or broken / handcrafted with just illegal values inside. Such a thing might even cause infinite loops. This PR doesn't fix that.

Issue #2851, opened 2024-08-22
[23] Formatter support for markdown doc comments


stephan-herrmann:
To complete #2845  we'd need to upgrade the formatter, too.

Currently, trying to format a document with markdown comments will throw exceptions like:
```
java.lang.AssertionError
	at org.eclipse.jdt.internal.formatter.TokenManager.firstIndexIn(TokenManager.java:194)
	at org.eclipse.jdt.internal.formatter.CommentsPreparator.visit(CommentsPreparator.java:590)
	at org.eclipse.jdt.core.dom.Javadoc.accept0(Javadoc.java:191)
	at org.eclipse.jdt.core.dom.ASTNode.accept(ASTNode.java:3313)
	at org.eclipse.jdt.internal.formatter.DefaultCodeFormatter.prepareComments(DefaultCodeFormatter.java:431)
	at org.eclipse.jdt.internal.formatter.DefaultCodeFormatter.prepareFormattedCode(DefaultCodeFormatter.java:233)
```

or
```
java.lang.IndexOutOfBoundsException: Index 29 out of bounds for length 29
	at java.base/jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:100)
	at java.base/jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:106)
	at java.base/jdk.internal.util.Preconditions.checkIndex(Preconditions.java:302)
	at java.base/java.util.Objects.checkIndex(Objects.java:365)
	at java.base/java.util.ArrayList.get(ArrayList.java:428)
	at org.eclipse.jdt.internal.formatter.TokenManager.get(TokenManager.java:79)
	at org.eclipse.jdt.internal.formatter.TokenManager.findIndex(TokenManager.java:173)
	at org.eclipse.jdt.internal.formatter.TokenManager.firstIndexIn(TokenManager.java:193)
	at org.eclipse.jdt.internal.formatter.CommentsPreparator.visit(CommentsPreparator.java:590)
	at org.eclipse.jdt.core.dom.Javadoc.accept0(Javadoc.java:191)
	at org.eclipse.jdt.core.dom.ASTNode.accept(ASTNode.java:3313)
	at org.eclipse.jdt.internal.formatter.DefaultCodeFormatter.prepareComments(DefaultCodeFormatter.java:431)
[...]

PR #2858, opened 2024-08-24
 [23] Formatter support for markdown doc comments

stephan-herrmann:
Minimal fix for https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2851

Restore ability to format Java in file with markdown comments:
+ avoid AssertionError in TokenManager.firstIndexIn
  - visit(Javadoc) needs to look for TokenNameCOMMENT_MARKDOWN specifically
+ fix infinite loop in tokenizeMultilineComment()
  - detect markdown comment
  - search for '/' not '*'
  - token should not include the final '\n'

Leave markdown comments unchanged by formatting for now

Includes a draft addition to support `Javadoc.isMarkdown()` in DOM AST.

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Javadoc.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Javadoc.java
@@ -42,6 +42,7 @@ public class Javadoc extends ASTNode {
 	// bug https://bugs.eclipse.org/bugs/show_bug.cgi?id=153399
 	// Store value tag positions
 	public long valuePositions = -1;
+	public boolean isMarkdown;
 
 	public Javadoc(int sourceStart, int sourceEnd) {
 		this.sourceStart = sourceStart;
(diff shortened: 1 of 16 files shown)


stephan-herrmann:
@jarthana  this PR includes one commit for exposing Javadoc.isMarkdown() in DOM: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2858/commits/a16d943bb3dcf7bf8801606e4f8f561f4e2af0e3

WDYT?

I included both commits in a single PR to see how far I could get regarding the Formatter. We can still discuss how to release this:
* rebase and merge to keep both commits separate?
* create a new PR just for the DOM changes?

stephan-herrmann:
As for the actual formatter changes:

Some changes were needed just to avoid AssertionError and an infinite loop, which **prevented any formatting in a text file with just one markdown comment present.**

Beyond that, I believe we should not hastily let the formatter alter the format of markdown comments, because in this case some **whitespace is actually semantically relevant**. IOW, to provide useful reformatting we would need to understand major parts of the markdown specification, which I believe is out of scope for JDT.

The best I could think of would be to select regions which can / cannot be reformatted:
* never format text with tricky indentation, itemization tokens, code fences or such
* otherwise:
  * try to format contiguous text regions  
  * try to format javadoc tags with description

But I believe it's OK when we postpone this until the feature leaves preview.

@mateusz-matela  please just let us know if you already have some tricks and ideas up the sleeve. It would also be good to know if my draft changes violate your design, or look dangerous in any way. For the rest let's make haste slowly.

PR #2866, opened 2024-08-27
[23] JEP 455: Primitive Types in Patterns, instanceof, and switch (Preview)

### LINKED ISSUE #2298: [23] JEP 455: Primitive Types in Patterns, instanceof, and switch (Preview)

stephan-herrmann:
- removed unused arg typeId from invokeDynamic() (several)

+ new switchBit Primitive identifies switch over primitive
+ switching over all int-ish primitives
  - not yet: long, float, double
+ route primitive constants through indy typeSwitch with 2 args:
  - the constant (int or boolean)
  - invocation of ConstantBootstraps.primitiveClass
+ for boolean constants specifically route the constant through invocation of ConstantBootstraps.getStaticFinal ( TRUE | FALSE )
+ uniquely generate new bootstraps controlled by ClassFile.PRIMITIVE_CLASS__TOKEN and ..GET_STATIC_FINAL__TOKEN
+ no requireNonNull() for primitives

+ generic test for most primitives with identity conversion
  + using a constant case & a pattern case

fixes #2298

@mpalat fyi, this is quite a mouthful, I hope the explanations above and in the code help. I did re-introduce my reorganized test class (under a new name and not yet wired to the suite).

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ClassFile.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ClassFile.java
@@ -100,6 +100,7 @@ import org.eclipse.jdt.internal.compiler.codegen.StackMapFrameCodeStream;
 import org.eclipse.jdt.internal.compiler.codegen.StackMapFrameCodeStream.ExceptionMarker;
 import org.eclipse.jdt.internal.compiler.codegen.TypeAnnotationCodeStream;
 import org.eclipse.jdt.internal.compiler.codegen.VerificationTypeInfo;
+import org.eclipse.jdt.internal.compiler.impl.BooleanConstant;
 import org.eclipse.jdt.internal.compiler.impl.CompilerOptions;
 import org.eclipse.jdt.internal.compiler.impl.Constant;
 import org.eclipse.jdt.internal.compiler.impl.StringConstant;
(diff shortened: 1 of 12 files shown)


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ClassFile.java
```diff
@@ -3959,6 +3970,55 @@ private int addClassDescBootstrap(int localContentsOffset, TypeBinding type, Map
 
 		return localContentsOffset;
 	}
+	private int addPrimitiveClassBootstrap(int localContentsOffset, Map<String, Integer> fPtr) {
+		final int contentsEntries = 4;
+		int idx = fPtr.get(CONSTANT_BOOTSTRAP__PRIMITIVE_CLASS);
```
stephan-herrmann, line 3975: The tokens  `PRIMITIVE_CLASS__TOKEN` and `GET_STATIC_FINAL__TOKEN`  can be refactored to descriptors to the end that this method and `addGetBooleanConstantsBootstrap` can be folded into one.

stephan-herrmann: done in [a62bba7](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2866/commits/a62bba785f264a1f3f884aefa2e1bf416868fc4c)

stephan-herrmann:
This PR focuses on positive cases to ensure code gen works and produces the expected behavior.

Test class `PrimitiveInPatternsTestSH` demonstrates the following (some of which was working already before this PR):
* **instanceof** (traditional and pattern) supports
  * identity testing for all primitives
     * this is independent of the shape of the lhs (variable or message send)
  * widening:
    * from all primitive types other than boolean to all larger numerical types
    * widening *to char* should be possible, too, but needs special treatment for proper printout in tests
  * narrowing:
    * from all primitive types other than boolean to all smaller primitive types except boolean (byte doesn't have such a smaller type)
    * narrowing from & to char is tested separately
  * boxing and unboxing for each primitive type
  * boxing plus widening to `Comparable` for all primitives
  * boxing plus widening to `Number` for all numerical primitives
  * Some groups of tests explicit demonstrate that the shape of the lhs has no impact on JEP 455
* **switch** on primitive or boxing types supports type identity and narrowing for
   * constants of all primitive types
   * patterns of all primitive types

  
Additionally, new errors are reported:
* in a switch on `long`, `float`, `double`, `boolean` (or their boxes) when a case constant does not have that exact primitive type
* duplicate `boolean` constant cases (was masked by incomplete logic around `null` cases, which share the constant `IntConstant(-1)` and thus require special treatment).

stephan-herrmann:
Here's an outline of relevant implementation changes regarding code generation:
* bootstrap method `SwitchBootstraps.typeSwitch` is used with new kinds of `Dynamic` arguments:
  * for type tests (patterns):
     * `ConstantBootstraps.primitiveClass()` to determine the primitive class (like `byte.class`) from the signature (like `B`)
  * for boolean constants:
     * `ConstantBootstraps.getStaticFinal()` to retrieve `Boolean.TRUE` / `Boolean.FALSE` fields
  * this implies that `SwitchBootstraps.typeSwitch` now appears with different signatures
* these "nested" bootstrap methods are singletons, i.e., they are declared only once and specific arguments are passed as part of the outer bootstrap method
   * all other constant values are passed directly into `typeSwitch`
* lifecycle of each "nested" bootstrap:
   * it is registered during `SwitchStatement.generateCodeSwitchPatternPrologue()` (see `recordSingletonBootstrapMethod()`)
     * its idx is stored in `ResolvedCase.primitivesBootstrapIdx`
  * descriptor generated during `ClassFile.generateBootstrapMethods()` (see `ClassFile.addSingletonBootstrap()`)
  * the above `primitivesBootstrapIdx` is then added to the `typeSwitch` bootstrap in `ClassFile.addBootStrapTypeSwitchEntry()` (both for boolean constants and primitive patterns)


We perform one optimization not found in javac:
* a pattern-less switch over `Boolean` or `boolean` avoids the use of `typeSwitch` by switching on the corresponding int value
  * this arose by the addition of long, float, double and boolean to the possible switch types, where boolean is the only type in this group that can be easily represented as int (and on the bytecode level already is).

mpalat:
> 
> @mpalat fyi, this is quite a mouthful, I hope the explanations above and in the code help. I did re-introduce my reorganized test class (under a new name and not yet wired to the suite).

thanks @stephan-herrmann  - going through them - yep, its a mouthful. And the explanations help much.  Very much welcome for the new test class - at this point shall we keep the PrimiInPatternsTest also alongwith the new one PIPTSH until post release atleast. Meanwhile let me try to see if there are any leftovers and will raise issues accordingly for each

Issue #2579, opened 2024-06-17
Enable "organize imports on save"


jukzi:
I would like to enable organize imports on save and would like to get feedback from jdt.core commiters.

Known benefit: i don't have to manually remove no longer used imports. 
Known downslde:  The first commit on each file will have a longer  - unrelated - patch size.

optional workaround: reorganize all imports once for all files.
history: at some point of time the sort order for imports was changed.

---
If you start a discussion and mention me, I will add more details there. Hint: if the change of strategy is done well, I'm not strictly against it.

_Originally posted by @stephan-herrmann in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1700#discussion_r1641778187_

iloveeclipse:
Yes, it makes sense to organize imports on save.
Not sure how many changes would happen on master, but if that action would be done for all sources would be good too.

jjohnstn:
I also thinks this makes sense.  I would also be in favor of reorganizing the imports in a mass change.

stephan-herrmann:
There is one small issue, why I'm not 100% happy with automatic organize import: it doesn't respect existing wildcard imports. I know that many people are actively opposed to wildcard imports, so here is my reason for liking them: I actually look at imports once in a while to get an idea of what a class depends on. When organize imports expands a nice `import org.eclipse.jdt.internal.compiler.ast.*;` into 147 lines of imports, the import list is simply overwhelming / useless for the human eye.

As a workaround I hope we can agree on a threshold for automatic wildcard imports. I'd propose a value of 10.

stephan-herrmann:
> Yes, it makes sense to organize imports on save. Not sure how many changes would happen on master, but if that action would be done for all sources would be good too.

I agree. Doing this incrementally would mean a lot of trouble with experimental changes, which trigger organize imports, to the effect that reverting the experimental change will not put the file back into a clean state.

A mass change on imports is acceptable, because import lists are not the typical subject of research into the git history :)

The typical caveat remains: we need to be 100% sure that the mass change does not introduce any unrelated changes, and it must happen in the same PR that enables automatic organize imports on save, to avoid any drift.

And a second caveat: it would create one more reason why changes must be prepared in the original Eclipse project context, to avoid any drift later on. For comparison: remove-trailing whitespace was globally enabled some years ago, and still once in a blue moon I see unrelated changes on save, because some commit succeeded to sneak in unwanted trailing white space :)

[...]

PR #2871, opened 2024-08-29
[tests] organize imports #2579

jukzi:
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2579


jukzi:
@stephan-herrmann when is a good timepoint to submit this? Its purely the automatic cleanup (takes ~1min). As a second step we could then change the options. If beta branch is different it is possibly simpler to do the same refactoring there instead of solving merge conflicts if any.

stephan-herrmann:
> @stephan-herrmann when is a good timepoint to submit this? Its purely the automatic cleanup (takes ~1min). As a second step we could then change the options.

We should not leave much time between automatic cleanup and changing the options to avoid fresh drift between both actions. My preference: two commits (not-squashed) within one PR.

> If beta branch is different it is possibly simpler to do the same refactoring there instead of solving merge conflicts if any.

Generally I'd do such things only when there's only one writable branch. This will be guaranteed once the beta branch has been merged to master (around 17 September)

Alternatively, doing this in the beta branch is an interesting idea, but that would conflict with any other changes between Simrel GA and JDK 23 GA. We don't want to close master until the latter.

Also, the beta branch is under some stress until release, so probably better not to stir up unnecessary dust there right now.

So if you could hold your horses just 3 more weeks that would be safest :)

stephan-herrmann:
On the positive side #2679 should help avoid that import changes result in warning changes :)

jukzi:
looks like beta branch has been merged - ready for an update?

stephan-herrmann:
> looks like beta branch has been merged - ready for an update?

indeed so - you're unblocked ...

jukzi:
superseeded by https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3032

PR #2874, opened 2024-08-29
Fix deprecated check routines in ASTNode to also check since value

### LINKED ISSUE #2873: deprecation markers with since > BREE are not shown in Problems view but in Editor
example: org.eclipse.e4.ui.css.swt.helpers.CSSSWTImageHelper shows a warning
`The constructor URL(String) is deprecated since version 20` in the editor but not in the Problems View.
Even if i configure the warning to be an error it is not shown in the Problems view while compile warings/errors of other type in the same file are shown.
image
The project is configured to have 17 compliance. When i change the the compliance level to 21 then both editor and problem view are consistently showning the marker.
image

I expect the warning in the editor to match the warnings in the problems view, no matter which compliance level i selected - shouldn't i?

jjohnstn:
- fixes #2873

## What it does

Fixes deprecated warnings and errors being shown when the release setting is below the since threshold of a
deprecated method, type, field.

## How to test

See issue.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ASTNode.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ASTNode.java
@@ -58,6 +58,7 @@ import org.eclipse.jdt.internal.compiler.ASTVisitor;
 import org.eclipse.jdt.internal.compiler.ast.TypeReference.AnnotationPosition;
 import org.eclipse.jdt.internal.compiler.classfmt.ClassFileConstants;
 import org.eclipse.jdt.internal.compiler.env.AccessRestriction;
+import org.eclipse.jdt.internal.compiler.impl.StringConstant;
 import org.eclipse.jdt.internal.compiler.lookup.*;
 
 @SuppressWarnings({"rawtypes", "unchecked"})
@@ -498,6 +499,8 @@ public abstract class ASTNode implements TypeConstants, TypeIds {
 		// inside same unit - no report
 		if (scope.isDefinedInSameUnit(field.declaringClass)) return false;
 
+		if (sinceValueUnreached(field, scope)) return false;
+
 		// if context is deprecated, may avoid reporting
 		if (!scope.compilerOptions().reportDeprecationInsideDeprecatedCode && scope.isInsideDeprecatedCode()) return false;
 		return true;
@@ -548,6 +551,8 @@ public abstract class ASTNode implements TypeConstants, TypeIds {
 		// inside same unit - no report
 		if (scope.isDefinedInSameUnit(method.declaringClass)) return false;
 
+		if (sinceValueUnreached(method, scope)) return false;
+
 		// non explicit use and non explicitly deprecated - no report
 		if (!isExplicitUse &&
 				(method.modifiers & ClassFileConstants.AccDeprecated) == 0) {
@@ -557,6 +562,35 @@ public abstract class ASTNode implements TypeConstants, TypeIds {
 		// if context is deprecated, may avoid reporting
 		if (!scope.compilerOptions().reportDeprecationInsideDeprecatedCode && scope.isInsideDeprecatedCode()) return false;
 		return true;
+	}
+
+	private boolean sinceValueUnreached(Binding binding, Scope scope) {
+		AnnotationBinding[] annotations= binding.getAnnotations();
+		for (AnnotationBinding annotation : annotations) {
+			if (String.valueOf(annotation.getAnnotationType().readableName()).equals("java.lang.Deprecated")) { //$NON-NLS-1$
+				ElementValuePair[] pairs= annotation.getElementValuePairs();
+				for (ElementValuePair pair : pairs) {
+					if (String.valueOf(pair.getName()).equals("since")) { //$NON-NLS-1$
+						if (pair.getValue() instanceof StringConstant strConstant) {
+							try {
+								String value= strConstant.stringValue();
+								int sinceValue= Integer.parseInt(value);
+								// As long as the AST levels and ClassFileConstants.MAJOR_VERSION grow simultaneously,
+								// we can use the offset of +44 to compute the Major version from the given AST Level
+								long sinceLevel= ClassFileConstants.getComplianceLevelForJavaVersion(sinceValue + 44);
+								long sourceLevel= scope.compilerOptions().sourceLevel;
+								if (sourceLevel < sinceLevel) {
+									return true;
+								}
+							} catch (NumberFormatException e) {
+								// do nothing and fall through
+							}
+						}
+					}
+				}
+			}
+		}
+		return false;
 	}
 
 	public boolean isSuper() {
@@ -618,6 +652,8 @@ public abstract class ASTNode implements TypeConstants, TypeIds {
 
 		// inside same unit - no report
 		if (scope.isDefinedInSameUnit(refType)) return false;
+
+		if (sinceValueUnreached(refType, scope)) return false;
 
 		// if context is deprecated, may avoid reporting
 		if (!scope.compilerOptions().reportDeprecationInsideDeprecatedCode && scope.isInsideDeprecatedCode()) return false;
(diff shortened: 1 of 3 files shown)


jjohnstn:
Hi @stephan-herrmann Can this be reviewed for M2?

stephan-herrmann:
> Hi @stephan-herrmann Can this be reviewed for M2?

yes, M2 sounds realistic.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ASTNode.java
```diff
@@ -589,6 +595,40 @@ public final boolean isMethodUseDeprecated(MethodBinding method, Scope scope,
...
 
+	private boolean sinceValueUnreached(Binding binding, Scope scope) {
+		if (binding instanceof TypeBinding typeBinding) {
+			if (!typeBinding.isReadyForAnnotations()) {
+				return false;
+			}
+		}
+		AnnotationBinding[] annotations = binding.getAnnotations();
+		for (AnnotationBinding annotation : annotations) {
+			if (annotation != null && String.valueOf(annotation.getAnnotationType().readableName()).equals("java.lang.Deprecated")) { //$NON-NLS-1$
```
stephan-herrmann, line 606: Simpler check without a magic string literal:
`annotation.getAnnotationType().id == TypeIds.T_JavaLangDeprecated`

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ASTNode.java
```diff
@@ -589,6 +595,40 @@ public final boolean isMethodUseDeprecated(MethodBinding method, Scope scope,
...
+				ElementValuePair[] pairs = annotation.getElementValuePairs();
+				for (ElementValuePair pair : pairs) {
+					if (String.valueOf(pair.getName()).equals("since")) { //$NON-NLS-1$
```
stephan-herrmann, line 609: `CharOperation.equals(pair.getName(), TypeConstants.SINCE)`

stephan-herrmann (changes requested):
I dropped some suggestions for technical code simplification.

But I'm not sure we are on 100% firm ground conceptually:
* do we know why editor and problems view show different results? A corresponding test in `ReconcilerTests9` could be helpful to observe the difference.
* I'd expect `since` values to be used mostly by JDK classes (because the version is a JDK version, not the version of 3rd party code) - and in the ticket the affected class was `java.net.URL` indeed.
  * in that case we shouldn't even see the class version with the deprecation annotations, IFF `--release` is used. This, too, is worth being double-checked.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/Deprecated9Test.java
```diff
@@ -915,7 +915,15 @@ public void testBug533063_2() throws Exception {
 			"----------\n";
 		runner.runWarningTest();
 	}
-	public void testBug534304() throws Exception {
+	public void testBug534304_1() throws Exception {
+		Map<String, String> options= getCompilerOptions();
+		String compliance= options.get(CompilerOptions.OPTION_Compliance);
+		if (compliance.equals(CompilerOptions.VERSION_9) ||
```
stephan-herrmann, line 921: In compiler tests we typically check compliance like this:
`this.complianceLevel < ClassFileConstants.JDK13`

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ASTNode.java
jjohnstn: Changed after rebasing.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ASTNode.java
jjohnstn: Changed after rebasing.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ASTNode.java
```diff
@@ -589,6 +595,40 @@ public final boolean isMethodUseDeprecated(MethodBinding method, Scope scope,
...
+						if (pair.getValue() instanceof StringConstant strConstant) {
+							try {
+								String value = strConstant.stringValue();
+								int sinceValue = Integer.parseInt(value);
+								// As long as the AST levels and ClassFileConstants.MAJOR_VERSION grow simultaneously,
+								// we can use the offset of +44 to compute the Major version from the given AST Level
+								long sinceLevel = ClassFileConstants.getComplianceLevelForJavaVersion(sinceValue + 44);
```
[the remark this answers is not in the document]
jjohnstn: Changed after rebasing.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/Deprecated9Test.java
jjohnstn: Changed after rebasing.

jjohnstn:
> I dropped some suggestions for technical code simplification.
> 
> But I'm not sure we are on 100% firm ground conceptually:
> 
>     * do we know why editor and problems view show different results? A corresponding test in `ReconcilerTests9` could be helpful to observe the difference.
> 
>     * I'd expect `since` values to be used mostly by JDK classes (because the version is a JDK version, not the version of 3rd party code) - and in the ticket the affected class was `java.net.URL` indeed.
>       
>       * in that case we shouldn't even see the class version with the deprecation annotations, IFF `--release` is used. This, too, is worth being double-checked.

Thanks for the code suggestions.  I agree that the since value will likely only be used with JDK classes.   I don't know how the release option works under the covers.  Should the release option change the URL class in this case since we are talking deprecation and not an API addition, change, or removal?  When I use JDK22 and specify compliance 19 with --release option checked, the Javadoc I get from the hover for the URL constructor in question is the JDK22 javadoc specifying the deprecation with since value.

stephan-herrmann:
> I don't know how the release option works under the covers.

`--release` selects class stubs from withing `ct.sym` that represent the requested version of all JDK classes. I.e., we basically see the byte code minus the code attribute in exactly the requested version.

> Should the release option change the URL class in this case since we are talking deprecation and not an API addition, change, or removal? 

The signature entry for URL at versions 20+ contains:
```
    Deprecated: true
    RuntimeVisibleAnnotations:
      0: #42(#43=s#44)
        java.lang.Deprecated(
          since="20"
        )
```

When specifying `--release 19` we get a version of URL without the deprecation info. Hence the compiler will not issue a warning when one of those contructors is used. This works as specified.

> When I use JDK22 and specify compliance 19 with --release option checked, the Javadoc I get from the hover for the URL constructor in question is the JDK22 javadoc specifying the deprecation with since value.

That's a problem: we do have the correct information in byte code format, but we don't have version specific source attachment and javadoc! There is only the JDK22 version in your setup.

stephan-herrmann:
> do we know why editor and problems view show different results? A corresponding test in ReconcilerTests9 could be helpful to observe the difference.

@jjohnstn did you succeed to write a test that demonstrates this difference between compiler and reconciler?

jjohnstn:
> > do we know why editor and problems view show different results? A corresponding test in ReconcilerTests9 could be helpful to observe the difference.
> 
> @jjohnstn did you succeed to write a test that demonstrates this difference between compiler and reconciler?

No, unfortunately I had to go to a wedding in Texas and have been doing some catch up in getting some UI changes into M2.

jjohnstn:
Hi @stephan-herrmann I added a test to ReconcilerTests9.  Without the patch, there is a warning for both methods (one is deprecated since 9 and the other is deprecated since 10).  With the patch, it only shows a warning for the method that is deprecated since 9.  Is that sufficient?

jjohnstn:
@stephan-herrmann The reason for the warning in the editor is that the IMethodBinding returns true for isDeprecated() due to the compiler MethodBinding which sets the flag based on the ASTNode.isMethodUseDeprecated() method which I changed.

stephan-herrmann:
For the actual difference between editor and builder (which can only be demonstrated with deprecation in an original JDK class) I filed #3168, which comes with its own set of open questions, and hence may not be resolved in the near future. For that reason the current fix is a valid workaround for a situation that ideally shouldn't even occur, but does occur currently.

stephan-herrmann (approved):

stephan-herrmann:
thanks @jjohnstn

PR #2888, opened 2024-09-03
Merge master into BETA_JAVA23

jarthana:
## Author checklist

- [ ] I have thoroughly tested my changes
- [ ] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [ ] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

PR #2892, opened 2024-09-03
Restore changes lost by force push

stephan-herrmann:
Cherry-pick from a local branch those changes that got lost by force-push of #2888


jarthana:
Different test failed from the same test suite this time!

stephan-herrmann:
> Different test failed from the same test suite this time!

Stars are ill-aligned for us, today.

jarthana:
> @jarthana : I've missed the point: **why** do we force push at all on **any** "official" project branch, beta or not? This is not OK, because in worst case there is no one to "repair" the rewritten history.
> 
> Can't we simply push **without** force? Git would say that your and remote branches are diverged in that case.

Yep, my bad. I should have been more careful since I directly push to the origin rather than my fork.

Issue #2884, opened 2024-09-02
Changing JRE of Execution Environment is needlessly slow


jukzi:
For example in the platform workspace changing the JRE freezes the UI for ~ 17sec.
![image](https://github.com/user-attachments/assets/a0602c89-b09b-421c-bf62-b6a3a65a8f0c)
![image](https://github.com/user-attachments/assets/911c6d5a-a4a1-4a3b-8055-f8071b493e3a)
![image](https://github.com/user-attachments/assets/81f5f25a-7685-4edb-9a08-e477612d18f5)

There seem to be multiple reasons. Biggest:

1. 10 seconds to touch all projects:

![image](https://github.com/user-attachments/assets/2e5d40d4-7ca8-48ff-8584-99fab4e66acc)

2. 5 seconds in UI thread: to repeated scan the whole JRT  for **every affected java project** in workspace:

![image](https://github.com/user-attachments/assets/6680e08a-d9dd-428a-ad24-7a54f866c948)

The result (list of files in the JDK) could be shared between projects. 
If that happens also during eclipse restart it might be worth for an optimization.

PR #2893, opened 2024-09-03
[performance] share JRT entries across projects #2884

jukzi:
JarPackageFragmentRootInfo.rawPackageInfo hold a list of all files in the JRT but each projects gets its own JarPackageFragmentRootInfo. Calculate that files independent of the instance and reuse if for all projects. To safely share the result across threads an unmodifiable datastructure is used.

* use record instead of String[2] for classes vs resources
* reduce rawtypes, unchecked conversions
* avoid char[] to String conversions

tested by JavaProjectTests, AttachedJavadocTests

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JarPackageFragment.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JarPackageFragment.java
@@ -14,9 +14,11 @@
 package org.eclipse.jdt.internal.core;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
-import java.util.Iterator;
+import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 
 import org.eclipse.core.resources.IResource;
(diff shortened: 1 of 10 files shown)


org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JarPackageFragment.java
```diff
@@ -132,11 +134,11 @@ private Object[] computeNonJavaResources(ArrayList entryNames) {
 			}
 		}
 	}
-	Iterator entries = childrenMap.entrySet().iterator();
+	Iterator<Entry<IPath, ArrayList<IPath>>> entries = childrenMap.entrySet().iterator();
 	while (entries.hasNext()) {
```
gzsombor, line 138: You can also simplify it with:
```
  for (var entry : childrenMap.entrySet()) {
	IPath entryPath = entry.getKey();
	ArrayList entryValue =  entry.getValue();
```

jukzi: ok, done. thanks.

Issue #2564, opened 2024-06-13
Incorrect startPosition of a SingleVariableDeclaration in the catch block


tkravchuk-parasoft:
![AST view](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/154594176/14730d7e-9201-4a71-8efc-1e46bb93f4a0)

Such a problem we have observed only when javadoc comment is placed before a catch keyword.

subyssurendran666:
@mpalat could you please assign it to me ?

jarthana:
Closing this issue.

PR #2896, opened 2024-09-03
Incorrect startPosition of a SingleVariableDeclaration in the catch block

subyssurendran666:
## What it does
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2564

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverter18Test.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverter18Test.java
@@ -5483,4 +5483,36 @@ public void testLambdaParameterSourcePosition() throws JavaModelException {
 	assertEquals(1, variableElement.getSourceRange().getLength());
 }
 
+public void testSVDStartPositionIssue() throws JavaModelException {
+	String contents = """
+				public class X {
+					public static void example() {
+						try {
+							System.out.println("try");
+						}
+						/** */
+						catch (RuntimeException e) {
+							System.out.println("catch");
+						}
+					}
+				}
+			""";
+	this.workingCopy = getWorkingCopy("/Converter22/src/xyz/X.java", true/*resolve*/);
+	CompilationUnit cu = (CompilationUnit) buildAST(contents, this.workingCopy);
+	TypeDeclaration typedeclaration = (TypeDeclaration) getASTNode(cu, 0);
+	MethodDeclaration methodDeclaration = (MethodDeclaration) typedeclaration.bodyDeclarations().get(0);
+	Block block = methodDeclaration.getBody();
+	List<ASTNode> statements = block.statements();
+	TryStatement tryStatement = (TryStatement) statements.get(0);
+	List<ASTNode> catchClauseList = tryStatement.catchClauses();
+	CatchClause catchClause = (CatchClause) catchClauseList.get(0);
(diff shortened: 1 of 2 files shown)


org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/util/CommentRecorderParser.java
```diff
@@ -203,6 +203,11 @@ public int flushCommentsDefinedPriorTo(int position) {
 		return position;
 	}
 
+	@Override
+	protected void consumeExitTryBlock() {
+		flushCommentsDefinedPriorTo(this.scanner.currentPosition);
```
jarthana, line 200: @subyssurendran666 See some of the other overridden consume* methods in this class that call the equivalent methods on super after doing what they are doing. I think we should do the same here as well.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverter18Test.java
```diff
@@ -5483,4 +5483,36 @@ interface I {
...
+	List<ASTNode> statements = block.statements();
+	TryStatement tryStatement = (TryStatement) statements.get(0);
+	List<ASTNode> catchClauseList = tryStatement.catchClauses();
+	CatchClause catchClause = (CatchClause) catchClauseList.get(0);
+	SingleVariableDeclaration svd = catchClause.getException();
+
+	assertEquals("Not a Single Variable Declaration", ASTNode.SINGLE_VARIABLE_DECLARATION, svd.getNodeType());
+	assertEquals("Not a Simple Type", ASTNode.SIMPLE_TYPE, svd.getType().getNodeType());
+	assertEquals("Not a Simple Name", ASTNode.SIMPLE_NAME, svd.getName().getNodeType());
+	assertEquals("Single Variable Declaration length is not correct", svd.getLength(), 18);
```
jarthana, line 5514: For anyone verifying this change or even those touching this testcase in future, it's easier if we use actual content to get the expected source range. For e.g "RuntimeException e".length(), contents.indexOf() etc.

jarthana (changes requested):

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/util/CommentRecorderParser.java
subyssurendran666: Done

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverter18Test.java
subyssurendran666: Done

jarthana:
@subyssurendran666 Sorry for input in a piecemeal manner. Can you also add a testcase with a comment on the catch parameter? Like this:
`catch(/** abc*/ Exception e) {}`
We should test that the source positions includes the comment.

subyssurendran666:
> @subyssurendran666 Sorry for input in a piecemeal manner. Can you also add a testcase with a comment on the catch parameter? Like this: `catch(/** abc*/ Exception e) {}` We should test that the source positions includes the comment.

@jarthana I have added a new test case. Could you please review my changes ?

jarthana (approved):

PR #2910, opened 2024-09-04
[performance] Avoid reading SourceFile twice

jukzi:
During compile parsing happens in two stages:
1. diet parse (any blocks like method bodies are skipped)
2. parse bodies Both phases did read the source .java file from file system. With this change the file contents is kept until no longer needed. It is cached in a SoftReference to avoid OutOfMemoryError.

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/Compiler.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/Compiler.java
@@ -422,7 +422,15 @@ public class Compiler implements ITypeRequestor, ProblemSeverities {
 	}
 
 	public void compile(ICompilationUnit[] sourceUnits) {
-		compile(sourceUnits, false);
+		try {
+			compile(sourceUnits, false);
+		} finally {
+			for (ICompilationUnit cu:sourceUnits) {
+				if (cu != null) {
+					cu.releaseContent();
+				}
+			}
+		}
 	}
 	/**
 	 * General API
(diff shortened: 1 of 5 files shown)


jukzi:
@stephan-herrmann could you review, please?
On my computer building the whole platform workspace it reduces reading sources from 7 to 3.5 seconds - which is expected (halfling that time)

stephan-herrmann:
> @stephan-herrmann could you review, please? On my computer building the whole platform workspace it reduces reading sources from 7 to 3.5 seconds - which is expected (halfling that time)

My focus is still on shipping complete support for Java 23. Any other reviews will have to wait until after 17 Sept.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/Compiler.java
```diff
@@ -851,30 +859,31 @@ protected void internalBeginToCompile(ICompilationUnit[] sourceUnits, int maxUni
...
-						parsedUnit = this.parser.dietParse(sourceUnits[i], unitResult);
+						parsedUnit = this.parser.dietParse(sourceUnit, unitResult);
 					}
 					long resolveStart = System.currentTimeMillis();
 					this.stats.parseTime += resolveStart - parseStart;
 					// initial type binding creation
 					this.lookupEnvironment.buildTypeBindings(parsedUnit, null /*no access restriction*/);
 					this.stats.resolveTime += System.currentTimeMillis() - resolveStart;
-					addCompilationUnit(sourceUnits[i], parsedUnit);
+					addCompilationUnit(sourceUnit, parsedUnit);
```
stephan-herrmann, line 886: why is this method changed?

stephan-herrmann (commented):
This solution looks fine, but I still wonder if SourceFile is the best location for this cache.

Given the large number of clients of this class, and many of them calling `getContents()`, it is difficult to see what will be the effect for clients other than `Compiler`. 
* Do any of those read the same source more than once?
* It would be difficult to define clean-up protocol for each of them.
   * hence we need to ask: is holding unused `SoftReference`s for free, or does it put some kind of pressure on the garbage collector?

Since compiler always accesses the `SourceFile` via `CompilationResult`, perhaps the latter could be made responsible for caching (encapsulated as `CompilationResult.getContents()`). And then cleanup could be hooked into the existing `CompilationUnitDeclaration.cleanUp()`, to the advantage that we don't add a new protocol to be observed.

jukzi:
> encapsulated as `CompilationResult.getContents()`)

please see https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3030/ for an alternate version based on CompilationResult

Issue #2580, opened 2024-06-17
jclMin22.jar contains classes compiled against Java 21


iloveeclipse:
Found while fixing tests in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2551.

[jclMin22.jar](https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/org.eclipse.jdt.core.tests.model/JCL/jclMin22.jar) contains classes compiled against Java 21. The expectation is that every such jar contains classes built for exact that Java API version it supposed to represent, so compiler tests can see / assert **matching** class file version if looking on classes from jclMin.jar's. Beside this, `jclMin22.jar` contains `.classpath` file that refers to `JavaSE-14` which also looks wrong.

I see that there was an attempt made to provide some systematic approach for jclMin* & converterJclMin* jars (see https://bugs.eclipse.org/bugs/show_bug.cgi?id=560658 and https://github.com/eclipse-jdt/eclipse.jdt.core/commit/d925ea3982e460755a90587fe85b43218963473b), however I do not see any trace of actual automation for **jclMin** jars, nor I see any explanation what is expected content of those jars and what is the difference between **jclMin** and **converterJclMin** jars.

I would be very grateful if @stephan-herrmann or @jarthana would add something to the https://github.com/eclipse-jdt/eclipse.jdt.core/wiki/Adding-support-for-new-Java-version-and-Features-in-JDT page (or any other Wiki page that fits to the "jcl" jars).

Once I would know the expected procedure I could update `jclMin22.jar` with proper content (and may be fix other jars as I can see that  `jclMin17.jar` contains `.classpath` file that refers to `JavaSE-14`).

PS.
In the changed code, it's `org.eclipse.jdt.core.tests.model.ClasspathTests.test232816f()` that found the problem

stephan-herrmann:
> I see that there was an attempt made to provide some systematic approach for jclMin* & converterJclMin* jars (see https://bugs.eclipse.org/bugs/show_bug.cgi?id=560658 and [d925ea3](https://github.com/eclipse-jdt/eclipse.jdt.core/commit/d925ea3982e460755a90587fe85b43218963473b)), however I do not see any trace of actual automation for **jclMin** jars ...

Right. Seems we ran out of steam before completing the task :(

iloveeclipse:
> Right. Seems we ran out of steam before completing the task :(

[...]

PR #2912, opened 2024-09-04
jclMin22.jar contains classes compiled against Java 21

iloveeclipse:
Re-built jclMin22.jar from jclMin22src.jar with proper project compilation target and enabled test232816f().

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2580


iloveeclipse:
@laeubi : any idea where this is coming from:

```
ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.8:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.jdt.core.compiler.batch: Baseline and reactor have the same fully qualified version, but different content
18:40:45  [ERROR] different
18:40:45  [ERROR]    META-INF/ECLIPSE_.RSA: not present in baseline
18:40:45  [ERROR]    META-INF/ECLIPSE_.SF: not present in baseline
```

Does it mean, ecj is not signed in the 4.33?

laeubi:
According to the log the jar was replaced with the baseline so there should be no baseline diff. Maybe this needs a rebase on master now some settings have changed?

iloveeclipse:
> According to the log the jar was replaced with the baseline so there should be no baseline diff. Maybe this needs a rebase on master now some settings have changed?

Isn't the error about different content between baseline and reactor? I'm sure this commit is on top of master.

laeubi:
> Isn't the error about different content between baseline and reactor? 

Yes but the logfile says that the main artifact (of the reactor) was replaced, what makes sense as you do not changed anything for ECJ (as far as I can see), so the error is really unexpected.

And thats also what the error complains, it says the jars have the _exact same version_ but different contents...

While on this one https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-2911/2/console

it complains that only qualifier changed (what usually indicates a missing version bump).

Issue #2915, opened 2024-09-05
[23] exhaustiveness signalling difference between javac and ecj


mpalat:
Given,
 
```
	public class X  {
	    public static int foo(Integer myInt) {
	        return switch (myInt) {
	            case int i  -> i;
	            default -> 0;
	        };
	    }
	
	    public static void main(String argv[]) {
	    	Integer i = 100;
	    	System.out.println(X.foo(i) == i);
	    }
	}  
```

ecj bails out with the error:
"Switch case cannot have both unconditional pattern and default label"
while javac compiles.

Further, spec says - 
Section 5.7.2 says - "Unconditional exactness is used at compile time to check the dominance of one pattern over another in a switch block ([14.11.1](https://cr.openjdk.org/~abimpoudis/instanceof/jep455-20240424/specs/instanceof-jls.html#jls-14.11.1)), and whether a switch block as a whole is exhaustive ([14.11.1.1](https://docs.oracle.com/javase/specs/jls/se22/html/jls-14.html#jls-14.11.1.1))."

and also, includes "a boxing conversion" for unconditionally exact. However, this is an unboxing condition [from Integer to int] and not a boxing conversion -and hence not unconditionally exact. and hence it looks like ecj should allow this - ie, should not consider "case int i" as covering Integer. 
But, what could be a value  of myInt such that default is taken instead of case int i? @stephan-herrmann any thoughts on this?

stephan-herrmann:
> But, what could be a value of myInt such that default is taken instead of case int i? @stephan-herrmann any thoughts on this?

`null` ?

mpalat:
> > But, what could be a value of myInt such that default is taken instead of case int i? @stephan-herrmann any thoughts on this?
> 
> `null` ?

@stephan-herrmann, I expected this question :) 
null anyway is not covered by default. Still, even if I add null
	            case int i  -> i;
                     case null -> -1;
	            default -> 0;
now also similar behaviour difference between javac and ecj - javac accepts, ecj doesn't. Between int i and null, now all the cases are covered, I believe? can you think of a case where default  case will be exercised?

stephan-herrmann:
Firstly this seems to apply (14.11.1.1):

> A set of case elements, CE, covers a type T if one of the following cases applies:
> * ...
[...]

PR #2925, opened 2024-09-05
[23] exhaustiveness signalling difference between javac and ecj

stephan-herrmann:
+ precise implementation of TypePattern.isUnconditional()
  - remove implementations in parent / sibling classes
+ defer setting SwitchStatement.totalPattern until we know if a default case is present (otherwise we would generated inconsistent code for default).

+ clarify that flagDuplicateDefault() flags only conditionally,
  renamed to checkDuplicateDefault()
  - inside this method clarify that only one error is reported per loc.
+ remove all conflict reporting from CaseStatement.resolveCasePattern
  - will be done within SwitchStatement.resolve() anyway
+ remove IProblem.DuplicateTotalPattern and related code
  - all errors reported here coincided with a dominance error
+ adjust tests: no longer expect secondary errors

fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2915

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/.settings/.api_filters
+++ b/org.eclipse.jdt.core.compiler.batch/.settings/.api_filters
@@ -23,6 +23,12 @@
                 <message_argument value="DisallowedStatementInPrologue"/>
             </message_arguments>
         </filter>
+        <filter comment="this preview constant (marked @noreference) is unnecesary" id="405864542">
+            <message_arguments>
+                <message_argument value="org.eclipse.jdt.core.compiler.IProblem"/>
+                <message_argument value="DuplicateTotalPattern"/>
+            </message_arguments>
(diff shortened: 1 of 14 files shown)


stephan-herrmann:
@mpalat @srikanth-sankaran  commit https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2925/commits/f508a96cd748c7809ddb40aea278e7937f250657 fully removes `IProblem.DuplicateTotalPattern` and related code. This is probably worth another pair of eyeballs looking at that change. 

You'll find an explanation in https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2915#issuecomment-2335143598

stephan-herrmann:
Merged in the interest of time.

Comments are still wellcome.

Issue #2928, opened 2024-09-09
[23] Unexpected operand at stack top error with double as switch type


jarthana:
Test case: 
```
public class X {
    public boolean foo(double d) {
    	boolean b;
        switch (d) {
            case 1d -> b = true;
            default -> b = false;
        }
        return b;
    }
} 
```

The ECJ crashes with the following exception:
Caused by: java.lang.AssertionError: Unexpected operand at stack top
	at org.eclipse.jdt.internal.compiler.codegen.OperandStack.pop(OperandStack.java:111)
	at org.eclipse.jdt.internal.compiler.codegen.CodeStream.tableswitch(CodeStream.java:7569)
	at org.eclipse.jdt.internal.compiler.ast.SwitchStatement.generateCode(SwitchStatement.java:836)
	at org.eclipse.jdt.internal.compiler.ast.AbstractMethodDeclaration.generateCode(AbstractMethodDeclaration.java:387)
	at org.eclipse.jdt.internal.compiler.ast.AbstractMethodDeclaration.generateCode(AbstractMethodDeclaration.java:323)
	at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.generateCode(TypeDeclaration.java:759)
	at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.generateCode(TypeDeclaration.java:829)
	at org.eclipse.jdt.internal.compiler.ast.CompilationUnitDeclaration.generateCode(CompilationUnitDeclaration.java:416)
	at org.eclipse.jdt.internal.compiler.Compiler.process(Compiler.java:935)

PR #2930, opened 2024-09-09
Issue #2928 unexpected operand error with case double

mpalat:
fixes issue #2928 

## Author checklist


stephan-herrmann:
@mpalat for posterity it would be good to have a little explanation while your memory is still fresh.

When I saw the stacktrace in #2928 my first thought was: someone tried to pop an int from the stack when in fact a double was at the top. Your fix seems to indicate that code gen actually had a different bug. What wrong did code gen do prior to your change?

The code you added in `needPatternDispatchCopy()` tests the flag `isPrimitiveSwitch`. Do you assume that we can enter this code (part of code gen) with `isPrimitiveSwitch` being `false`? If yes, then what should code gen do as an `else` to the new `if`? If no, then why do we need this `if`? Looking from a distance I think, the combination (switch over double and `!isPrimitiveSwitch`) will have raised an error during resolve, implying that code gen will never be performed for that method. If that is correct, then promoting `isPrimitiveSwitch` from local to field would be unnecessary.

PR #2944, opened 2024-09-12
Fixes #2933 - Byte to float case

### LINKED ISSUE #2933: [23] ECJ rejects a default case when it shouldn't
Here's the test:
```
public class X {
	public static float byteToFloat(Byte a) {
		return switch (a) {
			case float y -> y;
			default -> 2;
		};
	}
}
```
ECJ report on the default that Switch case cannot have both unconditional pattern and default label". Looks like somewhere we are omitting the fact that the float is not matching Byte.

mpalat:
fixes #2933 

## Author checklist


mpalat:
with the latest code the error had gone but the run-time results were wrong - it was branching to default-  this gets corrected via the pr above.

PR #2948, opened 2024-09-12
[performance] BatchImageBuilder: write .class files in batches

jukzi:
ProcessTaskManager
* use java.util.concurrent for queue
* signal Cancel/Exception/Stop via the queue
* implements AutoCloseable for try-with-resource
* drain as much Elements as possible from the queue

Improves the performance of "Clean all projects"

For example building platform workspace on Windows AbstractImageBuilder.compile(): 120 sec -> 91 sec

With this change the Compiler is actually waiting for parsing most time and not for the write to FileSystem anymore.

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ProcessTaskManager.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ProcessTaskManager.java
@@ -14,30 +14,34 @@
 
 package org.eclipse.jdt.internal.compiler;
 
-import java.util.concurrent.ExecutionException;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.ArrayBlockingQueue;
+import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
-import java.util.concurrent.TimeUnit;
-import java.util.concurrent.TimeoutException;
 
 import org.eclipse.jdt.internal.compiler.ast.CompilationUnitDeclaration;
 import org.eclipse.jdt.internal.compiler.problem.AbortCompilation;
 import org.eclipse.jdt.internal.compiler.util.Messages;
 
-public class ProcessTaskManager {
+public class ProcessTaskManager implements AutoCloseable {
 
 	private final Compiler compiler;
 	private final int startingIndex;
-	private volatile Future<?> processingTask; // synchronized write, volatile read
-	CompilationUnitDeclaration unitToProcess;
-	private Throwable caughtException;
+	private volatile boolean processing;
+	private final Future<?> processingTask; // synchronized write, volatile read
+	/** synchronized access **/
+	private CompilationUnitDeclaration unitWithError;
 
-	// queue
-	volatile int currentIndex, availableIndex, size, sleepCount;
-	private final CompilationUnitDeclaration[] units;
+	/** contains CompilationUnitDeclaration or something else as stop signal **/
+	private final BlockingQueue<Object> units;
 
-	public static final int PROCESSED_QUEUE_SIZE = 100;
+	private static final int PROCESSED_QUEUE_SIZE = 100;
+	private static final Object STOP_SIGNAL = new Object();
 
 	/** Normally a single thread is created an reused on subsequent builds **/
 	private static final ExecutorService executor = Executors.newCachedThreadPool(r -> {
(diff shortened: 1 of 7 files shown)


jukzi:
needs https://github.com/eclipse-platform/eclipse.platform/pull/1549

jukzi:
Example sampling. Before:
image
With this PR:
image

jukzi:
If anyone wants to review please let me know. Otherwise i plan to submit tomorrow.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/BatchImageBuilder.java
```diff
@@ -24,15 +24,30 @@
...
+	private final CompilationGroup compilationGroup;
+
+	/*  leave 2 threads for compiler + reader.*/
+	private static final ExecutorService WRITER_EXECUTOR = Executors.newFixedThreadPool(Math.max(1, Runtime.getRuntime().availableProcessors() - 2), r -> {
+				Thread t = new Thread(r, "Compiler Class File Writer"); //$NON-NLS-1$
+				t.setDaemon(true);
+				return t;
+			});
+	private final Map<IFile, byte[]> classContents = new HashMap<>();
+	private int classContentsBytesQueued = 0;
```
iloveeclipse, line 48: Should it be long? In an int one can only have ~2 GB of content. Surely not usual, but there is no need to limit the compiler. E.g. we run our application with 31 GB heap on 256 GB workstations and workspaces of hundreds thousands of files, where 2 GB is nothing.

Small style: don't init to default zero please.

jukzi: i can change that. however i have hardcoded a 100MB limit.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/BatchImageBuilder.java
```diff
@@ -348,4 +363,50 @@ protected void storeProblemsFor(SourceFile sourceFile, CategorizedProblem[] prob
...
+
+@Override
+public void startBatch() {
+	this.batchMode = true;
+}
+
+@Override
+protected void writeClassFileContents(ClassFile classFile, IFile file, String qualifiedFileName, boolean isTopLevelType,
+		SourceFile compilationUnit) throws CoreException {
+	if (JavaBuilder.DEBUG) {
```
iloveeclipse, line 375: This is not ambiguous and should be split into two traces: **writing** file and **caching** file contents for later write

jukzi:
Sampling from latest revision. writing the class files took only 11sec :-) :
image
double checked:
image

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/BatchImageBuilder.java
```diff
@@ -348,4 +363,50 @@ protected void storeProblemsFor(SourceFile sourceFile, CategorizedProblem[] prob
...
+		trace("Writing changed class file " + file.getName());//$NON-NLS-1$
+	}
+	byte[] content = classFile.getBytes();
+	if (this.batchMode) {
+		// flush before 100MB to avoid OOME:
+		if (!this.classContents.isEmpty() && this.classContentsBytesQueued + content.length >= 100_000_000) {
```
iloveeclipse, line 381: Could this please be a constant? CLASS_CONTENT_CACHE_LIMIT or something?

jukzi: done - MAX_CLASS_CONTENTS_BYTES_QUEUED

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ProcessTaskManager.java
```diff
@@ -46,159 +51,131 @@ public class ProcessTaskManager {
...
-		this.sleepCount = 0;
 	}
 
-	this.units[this.availableIndex++] = newElement;
-	if (this.availableIndex >= this.size)
-		this.availableIndex = 0;
-	if (this.sleepCount <= -1)
-		notify(); // wake up writing thread to accept next unit - could be the last one - must avoid deadlock
-}
+	private Object lastSignal = null;
```
iloveeclipse, line 72: please don't init to defaults, this makes debugger always jump to every useless init line...

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ProcessTaskManager.java
```diff
@@ -46,159 +51,131 @@ public class ProcessTaskManager {
...
-		//if (this.sleepCount < 0) throw new IllegalStateException(Integer.valueOf(this.sleepCount).toString());
-		this.sleepCount = 1;
+	private void addNextUnit(Object newElement) {
 		try {
-			wait(250);
-		} catch (InterruptedException ignore) {
-			// ignore
+			this.units.put(newElement);
+		} catch (InterruptedException interrupt) {
+			throw new RuntimeException(interrupt);
```
iloveeclipse, line 68: 1) Why is this wrapped? Usually if interrupted exception is handled, it should be flagged by Thread.interrupted() call, but neither this code nor caller does that.

2) Shouldn't this be AbortCompilation exception if we encounter problems during queue writing?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ProcessTaskManager.java
```diff
@@ -46,159 +51,131 @@ public class ProcessTaskManager {
...
-		}
+				if (!elements.isEmpty()) {
+					if (next !=null) {
+						// defer any stop signal until all CompilationUnitDeclaration read
+						this.lastSignal= next;
+					}
+					return elements;
+				}
+			} catch (InterruptedException interrupt) {
+				throw new RuntimeException(interrupt);
```
iloveeclipse, line 98: Same here regarding InterruptedException

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ProcessTaskManager.java
```diff
@@ -46,159 +51,131 @@ public class ProcessTaskManager {
...
-					return null;
+	/** returns null when no more elements can be expected **/
+	public Collection<CompilationUnitDeclaration> removeNextUnits() throws Error {
+		List<CompilationUnitDeclaration> elements = new ArrayList<>();
+		do {
+			Object next = this.lastSignal;
+			this.lastSignal = null;
+			try {
+				// wait until at least 1 element is available:
+				if (next ==null) {
```
iloveeclipse, line 82: `==null`  space missing

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ProcessTaskManager.java
```diff
@@ -46,159 +51,131 @@ public class ProcessTaskManager {
...
-
-		this.units[this.currentIndex++] = null;
-		if (this.currentIndex >= this.size)
-			this.currentIndex = 0;
-		if (this.sleepCount >= 1 && ++this.sleepCount > 4) {
-			notify(); // wake up processing thread to add next unit but only after removing some elements first
-			yield = this.sleepCount > 8;
-		}
+				if (!elements.isEmpty()) {
+					if (next !=null) {
```
iloveeclipse, line 91: ` !=null` space missing

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ProcessTaskManager.java
```diff
@@ -46,159 +51,131 @@ public class ProcessTaskManager {
...
-			do {
-				if (this.processingTask == null) {
-					if (this.caughtException != null) {
-						// rethrow the caught exception from the processingThread in the main compiler thread
-						if (this.caughtException instanceof Error)
-							throw (Error) this.caughtException;
-						throw (RuntimeException) this.caughtException;
-					}
-					return null;
+	/** returns null when no more elements can be expected **/
```
iloveeclipse, line 74: Can we return empty collection instead? It is always more convenient on the caller side to check for empty and not for null/not null

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ProcessTaskManager.java
```diff
@@ -46,159 +51,131 @@ public class ProcessTaskManager {
...
+			if (next instanceof Error error) {
+				throw error;
+			}
+			if (next instanceof RuntimeException runtimeException) {
+				throw runtimeException;
+			}
+			if (next == STOP_SIGNAL) {
+				return null;
+			}
+			throw new IllegalStateException(String.valueOf(next));
```
iloveeclipse, line 109: Could we have a speaking exception message like `"Received unexpected element to process: " + next`

iloveeclipse:
@szarnekow : would be good if you could take a look here too.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ProcessTaskManager.java
```diff
@@ -46,159 +51,131 @@ public class ProcessTaskManager {
...
+						this.compiler.reportProgress(Messages.bind(Messages.compilation_processing,
+								new String(unitToProcess.getFileName())));
+						if (this.compiler.options.verbose)
+							this.compiler.out.println(Messages.bind(Messages.compilation_process,
+									new String[] { String.valueOf(index + 1), String.valueOf(this.compiler.totalUnits),
+											new String(unitToProcess.getFileName()) }));
+						try {
+							this.compiler.process(unitToProcess, index);
+						} catch (AbortCompilation abortCompilation) {
+							unitToProcess.cleanUp();
```
iloveeclipse, line 144: `unitToProcess.cleanUp();` seem to be added here - why? It is done in finally below.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ProcessTaskManager.java
```diff
@@ -46,159 +51,131 @@ public class ProcessTaskManager {
...
-			}
-		}
-		if (t != null) {
-			t.get(250, TimeUnit.MILLISECONDS); // do not wait forever
+	@Override
+	public void close() {
+		try {
+			this.processingTask.get(250, TimeUnit.MILLISECONDS); // do not wait forever
+		} catch (InterruptedException | ExecutionException | TimeoutException ignored) {
+			// ignore
```
iloveeclipse, line 178: Not changed in this PR, but shouldn't Thread.interrupted() flag been cleared here?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ProcessTaskManager.java
```diff
@@ -46,159 +51,131 @@ public class ProcessTaskManager {
...
-				this.processingTask = null;
-				notifyAll();
-			}
-		}
-		if (t != null) {
-			t.get(250, TimeUnit.MILLISECONDS); // do not wait forever
+	@Override
+	public void close() {
+		try {
+			this.processingTask.get(250, TimeUnit.MILLISECONDS); // do not wait forever
```
iloveeclipse, line 176: should this be `this.processingTask.cancel(true)`?

iloveeclipse (commented):
I will be available next week for further review comments.

It would be nice if someone else also would review this PR if you plan to submit this before next week.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ProcessTaskManager.java
jukzi: ok

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ProcessTaskManager.java
jukzi: InterruptedException is a checked exception. that is not supported by executor.submit().
By now nothing should try to interrupt the executor, so it would be an error to receive one!
All other places in JDT just ignore InterruptedException. we could do that here too but that would need an endless loop until not interrupted. AbortCompilation is not supported here either as nothing reads the future. Only close() does - but that happens anyway only at the end.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ProcessTaskManager.java
jukzi: in this thread we can use throw `new AbortCompilation(true/* silent */, new RuntimeException(interrupt));`

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ProcessTaskManager.java
jukzi: ok

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ProcessTaskManager.java
jukzi: ok

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ProcessTaskManager.java
jukzi: Currently the call blocks until some elements are returned or end reached. But the idea was that if the queue is empty the caller could meanwhile flush the result. I forgot why it is not needed that anymore. I may think about it, should not be a relevant difference.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ProcessTaskManager.java
jukzi: we should never see it, but ok.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ProcessTaskManager.java
jukzi: good question, i forgot why. i will remove it

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ProcessTaskManager.java
jukzi: If you want some consistent InterruptedException behaviour you would need to change it everywhere. by now just ignoring is the pattern used all over in jdt.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ProcessTaskManager.java
jukzi: I tested: both units.clear() and cancel(true) let addNextUnit() progress. However cancel would only work if not anything else in the progress catches the interrupt. So i added both.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ProcessTaskManager.java
jukzi: I am happy you asked that. Since the current implementation returns as many CUs as available we know the next call would block until new units available=> the CPU has time to flush. I added flush after every call. This way it looks like the flush would take longer (more often only few files flushed), but the overall build time gets even faster, as the latency reduces:
image

iloveeclipse:
In debugger I see 30 idle threads staying after compilation (I have 32 core workstation). I wonder if they could be shut down after some time?

image

jukzi:
> I wonder if they could be shut down after some time?

https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/ThreadPoolExecutor.html#%3Cinit%3E(int,int,long,java.util.concurrent.TimeUnit,java.util.concurrent.BlockingQueue)
has a keepAliveTime. But you did not complain about the "Source File Reader" threads either so i wonder what is the benefit of shutting the threads down?

jukzi:
@iloveeclipse take your time next week, this can wait

iloveeclipse:
We often have to read thread dumps and having 30+ more threads there would make my life harder as it should be. Beside this, I don't like to waste resources (memory/system) if not needed. At the end, the number of (real) threads is limited per process.

jukzi:
there was a request from @laeubi  https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1937 to reuse threads. What keepAliveTime would you suggest?

iloveeclipse:
> there was a request from @laeubi #1937 to reuse threads. What keepAliveTime would you suggest?

Let say 5 minutes.

jukzi:
> Let say 5 minutes.

done

jukzi:
@iloveeclipse do plan to continue review?

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/BatchImageBuilder.java
```diff
@@ -348,4 +368,57 @@ protected void storeProblemsFor(SourceFile sourceFile, CategorizedProblem[] prob
...
+public void flushBatch() {
+	try {
+		if (JavaBuilder.DEBUG) {
+			this.classContents.keySet().forEach(file -> trace("Writing changed class file " + file.getName()));//$NON-NLS-1$
+		}
+		ResourcesPlugin.getWorkspace().write(this.classContents, true, true, false, null, WRITER_SERVICE);
+	} catch (CoreException e) {
+		// Already existing class files should not happen:
+		// Duplicate classes get marked earlier with a "The type {} is already defined"
+		Util.log(e, "JavaBuilder handling CoreException"); //$NON-NLS-1$
```
iloveeclipse, line 418: `"JavaBuilder handling CoreException"` - better something like `"Failed to write some of the class files: " + this.classContents.keySet()`.

The main question I have here is what happens in the IDE if the files weren't written - the user & JDT will assume that everything is OK after build, however the actual class files would be out-of-date / not existing, causing various unexpected issues. 

Ideally this failure here should be visible to user in some way, just logging is not enough as many users don't have Error log opened at all.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ProcessTaskManager.java
```diff
@@ -46,159 +50,128 @@ public class ProcessTaskManager {
...
-							String.valueOf(this.compiler.totalUnits),
-							new String(this.unitToProcess.getFileName())
-						}));
+	private void processing() {
+		try {
+			int unitIndex = this.startingIndex;
+			boolean noAnnotations = this.compiler.annotationProcessorManager == null;
+			while (this.processing) {
+				int index = unitIndex++;
+				boolean cleanup = noAnnotations || this.compiler.shouldCleanup(index);
```
iloveeclipse, line 118: The `shouldCleanup(index)` is running here with `startingIndex + 1`, while in the original code it was `startingIndex`. The `shouldCleanup()` compares `index < this.annotationProcessorStartIndex`  and it seem that this would be a bug here.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ProcessTaskManager.java
```diff
@@ -46,159 +50,128 @@ public class ProcessTaskManager {
...
+				CompilationUnitDeclaration unitToProcess = this.compiler.getUnitToProcess(index);
```
iloveeclipse, line 119: Here too, we start now with `startingIndex + 1` not with `startingIndex`

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/BatchImageBuilder.java
iloveeclipse: I believe `flushBatch()` should declare `throws CoreException` and don't catch it here.

iloveeclipse (changes requested):
I wonder we don't have tests showing `startingIndex + 1`.

Also would be cool to have a test that validates that one not written class file from the batch will be reported as compilation error.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/BatchImageBuilder.java
jukzi: i will add a problem marker for the project.

iloveeclipse: I haven't tried, but I'm curious what the builder infrastructure would do if we simply throw exception here?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ProcessTaskManager.java
jukzi: the ++ is a postfix, i.e. "index" has the unincremented unitIndex. "unitIndex" is not reused - only "index" is.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/BatchImageBuilder.java
jukzi: The errorhandling was taken from the existing 
https://github.com/eclipse-jdt/eclipse.jdt.core/blame/ff84b0ea54a020defe593d1e3956701879d0b0f9/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/AbstractImageBuilder.java#L222

i do not plan to change it.

jukzi: image

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ReadManager.java
```diff
@@ -32,19 +35,20 @@ public class ReadManager {
...
 	 */
 	private static int THREAD_COUNT = Math.max(0, Math.min(CACHE_SIZE, Runtime.getRuntime().availableProcessors() - 2));
-	private static final ExecutorService executor = THREAD_COUNT <= 0 ? null
-			: Executors.newFixedThreadPool(THREAD_COUNT, r -> {
-				Thread t = new Thread(r, "Compiler Source File Reader"); //$NON-NLS-1$
-				t.setDaemon(true);
-				return t;
-			});
+	private static final ExecutorService READER_SERVICE = THREAD_COUNT <= 0 ? null
+			: new ThreadPoolExecutor(THREAD_COUNT, THREAD_COUNT, /* keepAliveTime */ 5, TimeUnit.MINUTES,
```
iloveeclipse, line 39: I believe in order to stop idling threads, it should be something like `new ThreadPoolExecutor(1, THREAD_COUNT)`. At least I don't see threads disappearing with the current patch state.

Same for BatchImageBuilder

jukzi: Unfortunately that API seems to be not intuitive - corePoolSize=1 would never create more then 1 thread with an unbounded queue.
I'll take this workaround:
https://stackoverflow.com/a/19528622

iloveeclipse (approved):
OK, thanks for the updated thread pool fix.

Issue #2937, opened 2024-09-10
[23] VerifyError during conversion between T extends Short and short


jarthana:
Here's the affected testcase, which is a slightly modified version from #2936 :
```
record Record<T extends Short>(T t) {}
public class X {
	public static <T extends Short> short foo(Record<T> s) {
		return switch (s) {
			case Record(short s1) -> s1;
			default -> 0;
		};
	}
	public static void main(String[] args) {}
}
```

Running this with the latest from BETA_JAVA23 crashes with the following:
```
Error: Unable to initialize main class X
Caused by: java.lang.VerifyError: Inconsistent stackmap frames at branch target 62
Exception Details:
  Location:
    X.foo(LRecord;)S @52: goto
  Reason:
    Current frame's stack size doesn't match stackmap.
  Current Frame:
    bci: @52
    flags: { }
    locals: { 'Record', 'Record', integer }
    stack: { }
  Stackmap Frame:
    bci: @62
    flags: { }
    locals: { 'Record', 'Record', integer }
    stack: { 'java/lang/Short' }
```

stephan-herrmann:
Resolved by #2958

PR #2958, opened 2024-09-15
[23] VerifyError during conversion between T extends Short and short

stephan-herrmann:
when checking primitive conversions involving a type variable, always use the superclass to find a boxing type, knowing that all boxing types are classes, so other bounds are irrelevant.

fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2937


stephan-herrmann:
We have been using `erasure()` already in a few placed to find a boxing type hidden in a type variable, but we missed (at least) one relevant location.

OTOH, I don't think we have a card blanche to use `erasure()` for other situations, so I tried to establish the following principles:
* consistently use `superclass()` for the substitution rather than `erasure()` with the following justification:
   * we only look for boxing types among supers
   * since boxing types are classes, they must be found in `superclass`
* only perform such substitution if both are true:
   * the original type is a type variable (including capture)
   * the superclass is a boxing type indeed

The first item should not be a semantic change, just document that and why looking in this one location is sufficient.

I left one questionable use of `erasure()` in the context of `NARROWING_AND_UNBOXING_CONVERSION` which has a TODO anyway.

PR #2978, opened 2024-09-18
fix javadoc warning: Tag @author cannot be used in method documentation.

jukzi:
It can only be used in the following types of documentation: overview, module, package, class/interface.


akurtakov:
IMO such comments can be deleted entirely as they don't help in anyway .

Issue #2982, opened 2024-09-19
SwitchPatternTests failing in I-builds


akurtakov:
See https://download.eclipse.org/eclipse/downloads/drops4/I20240918-0950/testresults/html/org.eclipse.jdt.core.tests.compiler_ep434I-unit-cen64-gtk3-java21_linux.gtk.x86_64_21.html and

jarthana:
I did a quick comparison between R4_33_maintenance and master (post merge) and I find some difference in .class output of the typeSwitch signature:

Here's at 22:
```
         8: invokedynamic #16,  0             // InvokeDynamic #0:typeSwitch:(Ljava/lang/Object;I)I
...
...
  #16 = InvokeDynamic      #0:#17         // #0:typeSwitch:(Ljava/lang/Object;I)I
  #17 = NameAndType        #18:#19        // typeSwitch:(Ljava/lang/Object;I)I
  #18 = Utf8               typeSwitch
  #19 = Utf8               (Ljava/lang/Object;I)I
```

And what's produced at 23:
 ```
        8: invokedynamic #16,  0             // InvokeDynamic #0:typeSwitch:(Ljava/lang/Integer;I)I
...
...
  #16 = InvokeDynamic      #0:#17         // #0:typeSwitch:(Ljava/lang/Integer;I)I
  #17 = NameAndType        #18:#19        // typeSwitch:(Ljava/lang/Integer;I)I
  #18 = Utf8               typeSwitch
  #19 = Utf8               (Ljava/lang/Integer;I)I
```

I don't yet know what causing this and if this is expected or not.

jarthana:
Here's another interesting fact: The same code compiled with the post-merge compiler at compliance 22, runs OK with JRE 23. This explains why they were passing in Y builds and our setup.

jarthana:
I believe this is related to the recent work in the BETA branch wrt to the primitive pattern, even though the code itself doesn't contain any primitive patterns. Re-assigning to @mpalat

mpalat:
will take a look @jarthana

stephan-herrmann:
This seems to be a corner case where a switch was legal already in Java 22, but in 23 ecj "re-classifies" it as applying JEP 455.

In my work concerning JEP 455 I assumed that it suffices to reject at 22- any switches that **need** JEP 455. Looks like we need additional checks for switches that swing both ways, in order to generate code that is compatible with the target JRE.

[...]

PR #2986, opened 2024-09-19
SwitchPatternTests failing in I-builds

mpalat:
fixes issue #2982

## What it does
fixes issue #2982 

The return value of the CallSite [used by invokedynamic of typeSwitch] has changed in Java 23 - we need to check that if its a primitive enabled switch pattern (meaning that its 23 + preview then use the new signature else old]

## How to test

run 
org.eclipse.jdt.core.tests.compiler.regression.SwitchPatternTest testBug574525_03 - at level 21 and 22 with a 21 and 22 jre respectively.
-- | --

<br class="Apple-interchange-newline">

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/SwitchStatement.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/SwitchStatement.java
@@ -1023,7 +1023,9 @@ public class SwitchStatement extends Expression {
 		char[] arg1 = switch (exprType.id) {
 			case TypeIds.T_JavaLangLong, TypeIds.T_JavaLangFloat, TypeIds.T_JavaLangDouble, TypeIds.T_JavaLangBoolean,
 				TypeIds.T_JavaLangByte, TypeIds.T_JavaLangShort, TypeIds.T_JavaLangInteger, TypeIds.T_JavaLangCharacter->
-				exprType.signature();
+				this.isPrimitiveSwitch
+				? exprType.signature()
+				: "Ljava/lang/Object;".toCharArray(); //$NON-NLS-1$
 			default -> {
 				if (exprType.id > TypeIds.T_LastWellKnownTypeId && exprType.erasure().isBoxedPrimitiveType())
 					yield exprType.erasure().signature(); // <T extends Integer> / <? extends Short> ...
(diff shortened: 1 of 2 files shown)


akurtakov:
Have you looked at the failing test in the verification build? Hopefully, it's not caused by the change here.

akurtakov:
Btw, if we(committers) find it too much of an effort to fill the PR template maybe we should remove it? @eclipse-jdt/eclipse-jdt-committers  wdyt?

mpalat:
> Have you looked at the failing test in the verification build? Hopefully, it's not caused by the change here.

That failures is unrelated - so is the failure that has happened with the repeat run. I have ran those tests locally and are passing.

mpalat:
> Btw, if we(committers) find it too much of an effort to fill the PR template maybe we should remove it? @eclipse-jdt/eclipse-jdt-committers wdyt?

I have filled the template now - thanks for pointing out.

jukzi:
> PR template

PR template is too much overhead to fill out on daily bases, but as long as i can delete that part i don't care. For non-committers it's a good explanation what we expect though.

stephan-herrmann:
> Btw, if we(committers) find it too much of an effort to fill the PR template maybe we should remove it? @eclipse-jdt/eclipse-jdt-committers wdyt?

I had the same thought many times.

For first time contributors the template may make sense as a guideline of what information we need. Do non-committer contributors use the template as intended? I don't have any figures.

When committers leave the template untouched, this sets a poor example. I personally figure, I don't need to specify how to test, e.g., because I will already include unit tests that give that exact information. Being a committer also adhering to coding conventions and ECA can be assumed even if I don't specify this every time. But even we should keep in mind that github issues and PRs are an essential source of information for reviewers (now) and maintainers (later). 

As a consequence, I personally have the habit of replacing the template with information that is to the point for committer-to-committer communication. Does anyone object to that?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/SwitchStatement.java
```diff
@@ -1023,7 +1023,9 @@ char[] typeSwitchSignature(TypeBinding exprType) {
 		char[] arg1 = switch (exprType.id) {
 			case TypeIds.T_JavaLangLong, TypeIds.T_JavaLangFloat, TypeIds.T_JavaLangDouble, TypeIds.T_JavaLangBoolean,
 				TypeIds.T_JavaLangByte, TypeIds.T_JavaLangShort, TypeIds.T_JavaLangInteger, TypeIds.T_JavaLangCharacter->
-				exprType.signature();
+				this.isPrimitiveSwitch
+				? exprType.signature()
+				: "Ljava/lang/Object;".toCharArray(); //$NON-NLS-1$
```
stephan-herrmann, line 1028: there are two more locations below this block which still use `exprType.signature()` unconditionally. Shouldn't those have the same guard, too?

stephan-herrmann: ... or, simpler still, make an early return if `!this.isPrimitiveSwitch`.

mpalat: yep, will take a look - will wait for an I build first to confirm the fix

PR #2994, opened 2024-09-21
[Build] Remove duplicated configuration of tycho-p2-extras-plugin

HannesWell:
## What it does
Remove duplicated configuration of tycho-p2-extras-plugin. Most parts of the configuration are already inherited from the `eclipse-platform-parent`:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/73b7934ea89cbf0a8b0235572c8e80e84aed0104/eclipse-platform-parent/pom.xml#L372-L388

This would also help in the future in case JDT wants to use the workflow for automated version increments in pull-requests:
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2352

Also leverage the new overwritable property to permanently disable the baseline check, which is introduced in
- https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2381

@akurtakov, @iloveeclipse or @jukzi could you please have a look at this?

## Author checklist


jukzi:
> Remove duplicated configuration of tycho-p2-extras-plugin. Most parts of the configuration are already inherited from the `eclipse-platform-parent`:

I am not an expert on this, so i have a basic question: Where is it configured that JDT uses eclipse-platform-parent?

akurtakov:
> > Remove duplicated configuration of tycho-p2-extras-plugin. Most parts of the configuration are already inherited from the `eclipse-platform-parent`:
> 
> I am not an expert on this, so i have a basic question: Where is it configured that JDT uses eclipse-platform-parent?

https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/pom.xml#L15
https://github.com/eclipse-jdt/eclipse.jdt.debug/blob/master/pom.xml#L15
https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/master/pom.xml#L15

HannesWell:
Now it is ready, but requires https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/2381 to succeed.
The advantage is that we can now go full pomless, even with the annotations-v1.

HannesWell:
@akurtakov thanks for the update. The test-failure looks unrelated to me.

Issue #2835, opened 2024-08-19
ASTRewrite.rewriteAST fails with AssertionFailedException for records when it contains inner class


helospark:
Given the following `record`:

```java
record Test(String name, String address) {

    public static Builder builder() {
    }

    public static final class Builder {
    }
}
```

trying to remove the first inner node using `rewriteAST` results in the following exception:

```
org.eclipse.core.runtime.AssertionFailedException: assertion failed: 
	at org.eclipse.core.runtime.Assert.isTrue(Assert.java:113)
	at org.eclipse.core.runtime.Assert.isTrue(Assert.java:99)
	at org.eclipse.text.edits.TextEdit.<init>(TextEdit.java:153)
	at org.eclipse.text.edits.DeleteEdit.<init>(DeleteEdit.java:36)
	at org.eclipse.jdt.internal.core.dom.rewrite.ASTRewriteAnalyzer.doTextRemove(ASTRewriteAnalyzer.java:409)
	at org.eclipse.jdt.internal.core.dom.rewrite.ASTRewriteAnalyzer$ListRewriter.rewriteList(ASTRewriteAnalyzer.java:752)
...
```

The exception happens because `len` in DeleteEdit is negative.

However the same code works properly if the Test is a class instead of a record:
```java
class Test {
```

also works if the inner class and the method is in the opposite order.

Minimum reproducible example:
```java

    @Test
    public void minimumRepro() throws IOException, URISyntaxException, JavaModelException {
        String input = ""
                + "record Test(String name, String address) {\n"
                + "\n"
                + "    public static Builder builder() {\n"
                + "    }\n"
                + "\n"
                + "    public static final class Builder {\n"
                + "    }\n"
                + "\n"
                + "}";
        CompilationUnit result = getCompilationUnit(input);
        AbstractTypeDeclaration mainType = (AbstractTypeDeclaration) result.types().get(0);

        AST ast = result.getAST();
        ASTRewrite rewriter = ASTRewrite.create(ast);
        ListRewrite listRewriter = rewriter.getListRewrite(mainType, mainType.getBodyDeclarationsProperty());

        listRewriter.remove((ASTNode) mainType.bodyDeclarations().get(0), null); // remove builder() type

        Document document = new Document(input);
        TextEdit edits = rewriter.rewriteAST(document, null); // commit code change
    }

    private CompilationUnit getCompilationUnit(String input) throws IOException, URISyntaxException {
[...]

PR #3002, opened 2024-09-24
ASTRewrite.rewriteAST fails with AssertionFailedException for records when it contains inner class 

subyssurendran666:
## What it does
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2835

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/rewrite/describing/ASTRewritingRecordDeclarationTest.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/rewrite/describing/ASTRewritingRecordDeclarationTest.java
@@ -1402,5 +1402,43 @@ public class ASTRewritingRecordDeclarationTest extends ASTRewritingTest {
 
 	}
 
+	public void testRecord_029() throws Exception {
+		if (checkAPILevel()) {
+			return;
+		}
+		IPackageFragment pack1= this.sourceFolder.createPackageFragment("test1", false, null);
+
+		StringBuilder buf= new StringBuilder();
+		buf.append("package test1;\n");
+		buf.append("record Test(String name) {\n");
+		buf.append("	public static Builder builder() {}\n");
+		buf.append("	public static final class Builder {}\n");
+		buf.append("}\n");
+
+		ICompilationUnit cu= pack1.createCompilationUnit("Test.java", buf.toString(), false, null);
+
+		CompilationUnit astRoot= createAST(cu);
+		ASTRewrite rewrite= ASTRewrite.create(astRoot.getAST());
+		AST ast= astRoot.getAST();
+
+		assertTrue("Parse errors", (astRoot.getFlags() & ASTNode.MALFORMED) == 0);
+		List <RecordDeclaration> methods= astRoot.types();
+		MethodDeclaration methodDecl= findMethodDeclaration(methods.get(0), "builder");
+		{
+			rewrite.remove(methodDecl, null);
+		}
+
+		String preview= evaluateRewrite(cu, rewrite);
+
+		buf = new StringBuilder();
+		buf.append("package test1;\n");
+		buf.append("record Test(String name) {\n");
+		buf.append("	\n");
+		buf.append("	public static final class Builder {}\n");
+		buf.append("}\n");
+
+		assertEqualString(preview, buf.toString());
+	}
+
 }
 
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/rewrite/describing/ASTRewritingRecordDeclarationTest.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/rewrite/describing/ASTRewritingRecordDeclarationTest.java
@@ -1402,7 +1402,7 @@ public class ASTRewritingRecordDeclarationTest extends ASTRewritingTest {
 
 	}
 
-	public void testRecord_029() throws Exception {
+	public void testRecord_029_a() throws Exception {
 		if (checkAPILevel()) {
 			return;
 		}
@@ -1410,7 +1410,7 @@ public class ASTRewritingRecordDeclarationTest extends ASTRewritingTest {
 
 		String code = """
 					package test1;
-					public record Test(String name) {
+					record Test(String name) {
 					    public static Builder builder() {}
 					    public static final class Builder {}
 					}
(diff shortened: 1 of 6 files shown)


org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
```diff
@@ -3323,7 +3323,7 @@ public static List<String> getAllVersions() {
 	 *
 	 * @return all Java source versions fully supported by Eclipse compiler
 	 * @see #isJavaSourceVersionSupportedByCompiler(String)
-	 * @since 3.39
+	 * @since 3.40
```
jarthana, line 3326: Why is this required? This was added last release and we don't change the @since versions once they are set.

jarthana:
@subyssurendran666 Can you say a word or two about the root cause of the problem and what the fix does? Thanks!

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/rewrite/describing/ASTRewritingRecordDeclarationTest.java
```diff
@@ -1402,5 +1402,43 @@ public void testRecord_028() throws Exception {
...
+		buf.append("record Test(String name) {\n");
+		buf.append("	public static Builder builder() {}\n");
+		buf.append("	public static final class Builder {}\n");
+		buf.append("}\n");
+
+		ICompilationUnit cu= pack1.createCompilationUnit("Test.java", buf.toString(), false, null);
+
+		CompilationUnit astRoot= createAST(cu);
+		ASTRewrite rewrite= ASTRewrite.create(astRoot.getAST());
+		AST ast= astRoot.getAST();
```
jarthana, line 1422: There's a warning here.

jarthana (changes requested):
As I said on the PR, a word or two about why the fix is correct would help speed up the review. For e.g., it's not clear why this is not a problem for regular classes but only a problem for records?

subyssurendran666:
I have a code snippet
```
record Test(String name, String address) {

    public static Builder builder() {
    }

    public static final class Builder {
    }
}
```
and trying to rewrite the inner node using the ASTRewrite, then it throws an exception

```
org.eclipse.core.runtime.AssertionFailedException: assertion failed: 
	at org.eclipse.core.runtime.Assert.isTrue(Assert.java:113)
	at org.eclipse.core.runtime.Assert.isTrue(Assert.java:99)
	at org.eclipse.text.edits.TextEdit.<init>(TextEdit.java:153)
	at org.eclipse.text.edits.DeleteEdit.<init>(DeleteEdit.java:36)
	at org.eclipse.jdt.internal.core.dom.rewrite.ASTRewriteAnalyzer.doTextRemove(ASTRewriteAnalyzer.java:409)
	at org.eclipse.jdt.internal.core.dom.rewrite.ASTRewriteAnalyzer$ListRewriter.rewriteList(ASTRewriteAnalyzer.java:752)
```

The exception occurs when the length of DeleteEdit becomes negative. However, when I change the record in the test to a class, the issue resolves.

In the method `ASTRewriteAnalyzer.doTextRemove()`, it already checks if the `len == 0` and returns null accordingly. I applied a similar check to handle cases where the value becomes negative.

subyssurendran666:
> I have a code snippet
> 
> ```
> record Test(String name, String address) {
> 
>     public static Builder builder() {
>     }
> 
>     public static final class Builder {
>     }
> }
> ```
> 
> and trying to rewrite the inner node using the ASTRewrite, then it throws an exception
> 
> ```
> org.eclipse.core.runtime.AssertionFailedException: assertion failed: 
> 	at org.eclipse.core.runtime.Assert.isTrue(Assert.java:113)
> 	at org.eclipse.core.runtime.Assert.isTrue(Assert.java:99)
> 	at org.eclipse.text.edits.TextEdit.<init>(TextEdit.java:153)
> 	at org.eclipse.text.edits.DeleteEdit.<init>(DeleteEdit.java:36)
> 	at org.eclipse.jdt.internal.core.dom.rewrite.ASTRewriteAnalyzer.doTextRemove(ASTRewriteAnalyzer.java:409)
> 	at org.eclipse.jdt.internal.core.dom.rewrite.ASTRewriteAnalyzer$ListRewriter.rewriteList(ASTRewriteAnalyzer.java:752)
> ```
> 
> The exception occurs when the length of DeleteEdit becomes negative. However, when I change the record in the test to a class, the issue resolves.
> 
> In the method `ASTRewriteAnalyzer.doTextRemove()`, it already checks if the `len == 0` and returns null accordingly. I applied a similar check to handle cases where the value becomes negative.

@jarthana I've noticed another issue related to the AST: the starting position of the MethodDeclaration is invalid. I'm currently debugging this issue.

Screenshot 2024-10-11 at 3 08 29 PM

subyssurendran666:
The method A`STConvert.convert(TypeDeclaration)` is responsible for converting the Compiler AST into a DOM AST, utilising `ASTConverter.convertToRecordDeclaration()`. During this conversion, it invokes ASTConverter.buildBodyDeclarations() to handle the body declarations. In this process, the typeDeclaration may include a default constructor, even if the test code doesn't explicitly define one. This is due to the automatic constructor generation by the Eclipse Compiler for Java (ECJ).

As the ASTConverter attempts to generate a RecordDeclaration from the Compiler AST, it processes the body in a specific order: fields first, then methods, and finally member types. In the current scenario, there are no fields, so the converter immediately moves on to the methods. The first method encountered is the default constructor generated by the ECJ, which is correctly omitted. The process then moves to the member types.

The issue arises in the subsequent iteration of the loop, which only processes members. This behavior leads to the BAD AST issue, as the ASTConverter incorrectly handles the generated RecordDeclaration. This misprocessing is also the root cause of the reported issue with ASTRewrite. The ASTRewrite process ends up rewriting the wrong inner node inside the record, which results in an invalid AST structure.

mpalat (approved):
@subyssurendran666 since this is an issue primarily due to conversion, can you please add a junit test case for DOM AST conversion. Once review comments are incorporated please go ahead and merge. Thanks!

subyssurendran666:
> As I said on the PR, a word or two about why the fix is correct would help speed up the review. For e.g., it's not clear why this is not a problem for regular classes but only a problem for records?

Done.

mpalat (approved):

jarthana (approved):

mpalat:
@jarthana took the liberty of merging - didn't want to risk another rebase :)

PR #3006, opened 2024-09-25
Issue with code fence > 3 in markdown comment

### LINKED ISSUE #3004: Issue with code fence > 3 in markdown comment
Inspecting the code change in #2989 it seems to break handling of code fences of four or more characters. Consider the following contrived example:

```java
	/// ````
	/// ```
	/// @param is not a tag here because this is nested literal *markdown*
	/// ```
	/// ````
	void m() {}
```

Here we should recognize a code block delimited by four backticks, where three backticks don't suffice to terminate the block. As a result `@param` should *not* be interpreted as a javadoc tag, but as literal text.

The problem relates to field `org.eclipse.jdt.internal.compiler.parser.MarkdownCommentHelper.fenceLength` no longer being increased to a value greater than 3.

jarthana:
## What it does
Fixes #3004 

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/AbstractCommentParser.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/AbstractCommentParser.java
@@ -186,6 +186,8 @@ public abstract class AbstractCommentParser implements JavadocTagConstants {
 			if (!this.markdown) {
 				realStart = this.linePtr==1 ? this.javadocStart : this.scanner.getLineEnd(this.linePtr-1)+1;
 				if (realStart < this.javadocStart) realStart = this.javadocStart;
+			} else {
+				this.linePtr = getLineNumber(realStart);
 			}
 			this.scanner.resetTo(realStart, this.javadocEnd);
 			this.index = realStart;
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/AbstractCommentParser.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/AbstractCommentParser.java
@@ -180,9 +180,13 @@ public abstract class AbstractCommentParser implements JavadocTagConstants {
 			int lastStarPosition = -1;
 
 			// Init scanner position
+			this.markdown = this.source[this.javadocStart + 1] == '/';
 			this.linePtr = getLineNumber(this.firstTagPosition);
-			int realStart = this.linePtr==1 ? this.javadocStart : this.scanner.getLineEnd(this.linePtr-1)+1;
-			if (realStart < this.javadocStart) realStart = this.javadocStart;
+			int realStart = this.javadocStart;
+			if (!this.markdown) {
+				realStart = this.linePtr==1 ? this.javadocStart : this.scanner.getLineEnd(this.linePtr-1)+1;
+				if (realStart < this.javadocStart) realStart = this.javadocStart;
+			}
 			this.scanner.resetTo(realStart, this.javadocEnd);
 			this.index = realStart;
 			if (realStart == this.javadocStart) {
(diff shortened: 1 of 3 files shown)


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/IMarkdownCommentHelper.java
```diff
@@ -138,8 +138,9 @@ public void recordFenceChar(char previous, char next, boolean lineStarted) {
 		int required = this.insideFencedCodeBlock ? this.fenceLength : 3;
 		if (++this.fenceCharCount == required) {
 			this.insideFencedCodeBlock^=true;
-			this.fenceLength = this.fenceCharCount;
 		}
+		if (this.insideFencedCodeBlock && this.fenceCharCount >= this.fenceLength)
+			this.fenceLength = this.fenceCharCount;
```
stephan-herrmann, line 143: This part looks OK. 
I admit that the logic in this method is not very clear.
Do you want me to try an alternate design as an explicit state machine?
I'm thinking of states NONE, INCOMPLETE_OPEN, COMPLETE_OPEN, INSIDE, INCOMPLETE_CLOSE, where in particular COMPLETE_OPEN would be the state where a fence of 3 can still grow until a non-fence character is seen.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/AbstractCommentParser.java
```diff
@@ -180,9 +180,13 @@ protected boolean commentParse() {
...
 
 			// Init scanner position
+			this.markdown = this.source[this.javadocStart + 1] == '/';
 			this.linePtr = getLineNumber(this.firstTagPosition);
-			int realStart = this.linePtr==1 ? this.javadocStart : this.scanner.getLineEnd(this.linePtr-1)+1;
-			if (realStart < this.javadocStart) realStart = this.javadocStart;
+			int realStart = this.javadocStart;
+			if (!this.markdown) {
+				realStart = this.linePtr==1 ? this.javadocStart : this.scanner.getLineEnd(this.linePtr-1)+1;
+				if (realStart < this.javadocStart) realStart = this.javadocStart;
```
stephan-herrmann, line 188: I wonder if the difference is really between markdown and traditional, or perhaps `firstTagPosition` is the culprit, which may point to a `@` that actually does not start a tag because it's inside a code block. 
But avoiding a bogus firstTagPosition would require even the scanner (where `commentTagStarts` is maintained) to detect code blocks, which doesn't sound like a task for the scanner.

Do you know how this is handled in traditional javadoc, i.e., does the scanner signal a tag when we have `{@code @param is not a tag here}`?

stephan-herrmann (commented):
I see no fault, but some questions for improving our understanding of the strategy.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/IMarkdownCommentHelper.java
jarthana: If we don't come up with lot of issues in this regard, we can live with this, I guess.

jarthana:
I have no idea why the build is failing. Only theory I can think of is the compilation pref has changed somehow (tycho?).

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/AbstractCommentParser.java
stephan-herrmann: > I see that even with a regular Javadoc, the firstTagPosition does point to the @code. So, there's no difference there. Just that the parser handles it after that.

That **is** a difference: in traditional javadoc you *need* a tag even to enter a code section, in markdown looking for the first `@` is not sufficient to understand the relevant structure.

IOW, starting at `{@code` is fine, everything before that is uninteresting for semantic analysis. But in markdown starting at `@param` lets you miss that fact that we are inside a code block.

stephan-herrmann:
> I have no idea why the build is failing. Only theory I can think of is the compilation pref has changed somehow (tycho?).

This:
```
```
is used only inside javadoc, so if javadoc analysis doesn't find the reference, then it will be marked as unused. Did anyone just change javadoc analysis? ;-P

jarthana:
OK, I see that setting the linePtr indiscriminately is wrong. Will take a look.

jarthana:
Something's messed up in my last commit that undid the changes to one file. This needs another attempt.

PR #3015, opened 2024-09-26
[24] Create Java 24 constants and test framework changes #2905

### LINKED ISSUE #2905: [24] Create Java 24 constants and test framework changes
Ref issue #2268 for a similar issue for Java 23

also, to raise once the issue #2212 gets closed - https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/4497

jarthana:
## What it does
Fixes #2905 

## Author checklist


org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/util/DOMASTUtil.java
```diff
@@ -214,7 +214,7 @@ public static void checkASTLevel(int level) {
 	private static final String[] AST_COMPLIANCE_MAP = {"-1","-1",JavaCore.VERSION_1_2, JavaCore.VERSION_1_3, JavaCore.VERSION_1_7, //$NON-NLS-1$ //$NON-NLS-2$
 			JavaCore.VERSION_1_7, JavaCore.VERSION_1_7, JavaCore.VERSION_1_7, JavaCore.VERSION_1_8, JavaCore.VERSION_9, JavaCore.VERSION_10,
 			JavaCore.VERSION_11, JavaCore.VERSION_12, JavaCore.VERSION_13, JavaCore.VERSION_14, JavaCore.VERSION_15, JavaCore.VERSION_16, JavaCore.VERSION_17, JavaCore.VERSION_18,
-			JavaCore.VERSION_19, JavaCore.VERSION_20, JavaCore.VERSION_21, JavaCore.VERSION_22, JavaCore.VERSION_23};
+			JavaCore.VERSION_19, JavaCore.VERSION_20, JavaCore.VERSION_21, JavaCore.VERSION_22, JavaCore.VERSION_23, JavaCore.VERSION_23};
```
akurtakov, line 217: Shouldn't this be VERSION_24?

jarthana: Yes, thanks. Will fix it.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/eval/DebugEvaluationSetup.java
```diff
@@ -55,7 +60,10 @@ protected void setUp() {
 				LocalVMLauncher launcher;
 				try {
 					launcher = LocalVMLauncher.getLauncher();
-					launcher.setVMArguments(new String[]{"-verify"});
+					long vmVersion = Util.getMajorMinorVMVersion();
+					if (vmVersion < ClassFileConstants.JDK24) {
+						launcher.setVMArguments(new String[]{"-verify"});
```
stephan-herrmann, line 65: is option `-verify` being removed, or what kind of trouble does it cause?

jarthana: Not yet but will cause tons of tests to fail with this warning. We need a strategy if we continue to require this by our tests.

jarthana: > > Not yet but will cause tons of tests to fail with this warning. We need a strategy if we continue to require this by our tests.
> 
> Are you saying this from experience with previous versions, or is anything becoming stricter specifically in 24?
> 
> I'm just curious what could make classfiles generated by ecj fail the VM's verification?

This part, I don't know. I haven't spent time on this yet. I was hoping someone would already know the answer :) 
All I know now is the test are failing with unexpected warning about the -verify argument.

stephan-herrmann: >  All I know now is the test are failing with unexpected warning about the -verify argument.

Where could I such such failures?

jarthana: Unfortunately, the jenkins that failed with that error is gone. But here's what I did to reproduce locally - just set the "passed" to false after the call to `verifyClassFiles()` in line number 3574 of `AbstractRegressionTest`.

stephan-herrmann: > "Warning: -verify option is deprecated and may be removed in a future release."

Well ...

Actually, I can't find `-verify` in any documentation of the `java` command, recent or old.

[At 1.8 ](https://docs.oracle.com/javase/8/docs/technotes/tools/unix/java.html) I see "-Xverify:mode".

And even `-Xverify` is [deprecated since 13](https://www.oracle.com/java/technologies/javase/13all-relnotes.html)

It looks like verification will happen always, and hence the option is plain useless.

How to test: revert one of the recent fixes in JAVA_BETA23 that fixed a verify error, and run the test without -verify. It should still fail :)

jarthana: So what do you reckon? Just removing it altogether instead of doing it conditionally?

stephan-herrmann:
@jarthana looking at the changes how jclMin  are built, I'm a bit at loss what is the new strategy.

In a comment in ConverterTestSetup I learned that you don't want to have separate jclMin and converterJclMin any more. Fair enough. But the build.xml  now seems to create same-named output files for targets with different inputs (below 24).

Also (after latest merge) build.xml has a comment close without a comment start. So It's quite unclear what this file should actually look like.

jarthana:
> @jarthana looking at the changes how jclMin are built, I'm a bit at loss what is the new strategy.
> 
> In a comment in ConverterTestSetup I learned that you don't want to have separate jclMin and converterJclMin any more. Fair enough. But the build.xml now seems to create same-named output files for targets with different inputs (below 24).
> 
> Also (after latest merge) build.xml has a comment close without a comment start. So It's quite unclear what this file should actually look like.

Sorry, I didn't pay attention to your changes. So, I suppose we can remove the targets `allConverterJcls `and `converterJclMin`. What do you say?

PR #3063, opened 2024-10-09
ASTParser should detect and report when no suitable system library has been configured #3047

### LINKED ISSUE #3047: ASTParser should detect and report when no suitable system library has been configured
I have a case when I got NullPointerException, it seems that it is thrown when we are parsing more than one file and java lib is not set (when calling method parser.setEnvironment() includeRunningVMBootclasspath flag is set to false, changing to true solves the problem). 

java.lang.NullPointerException: Cannot read field "declaringClass" because "mb" is null
        at org.eclipse.jdt.internal.compiler.ast.ReferenceExpression.generateCode(ReferenceExpression.java:375)
        at org.eclipse.jdt.internal.compiler.ast.LocalDeclaration.generateCode(LocalDeclaration.java:174)
        at org.eclipse.jdt.internal.compiler.ast.AbstractMethodDeclaration.generateCode(AbstractMethodDeclaration.java:387)
        at org.eclipse.jdt.internal.compiler.ast.AbstractMethodDeclaration.generateCode(AbstractMethodDeclaration.java:323)
        at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.generateCode(TypeDeclaration.java:759)
        at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.generateCode(TypeDeclaration.java:829)
        at org.eclipse.jdt.internal.compiler.ast.CompilationUnitDeclaration.generateCode(CompilationUnitDeclaration.java:412)
        at org.eclipse.jdt.internal.compiler.Compiler.process(Compiler.java:935)
        at org.eclipse.jdt.core.dom.CompilationUnitResolver.resolve(CompilationUnitResolver.java:1173)
        at org.eclipse.jdt.core.dom.CompilationUnitResolver.resolve(CompilationUnitResolver.java:786)
        at org.eclipse.jdt.core.dom.CompilationUnitResolver$ECJCompilationUnitResolver.resolve(CompilationUnitResolver.java:92)
        at org.eclipse.jdt.core.dom.ASTParser.createASTs(ASTParser.java:1071)
        at npe.JDTParsingExample.main(JDTParsingExample.java:52)

Steps to reproduce:
download and unzip the attached project
make sure you are under Java 17
then run: 

gradlew run

As a result on the console, you should get the listed exception. The project is using jdt.core version 3.39.0

[npe_while_parsing.zip](https://github.com/user-attachments/files/17260145/npe_while_parsing.zip)

jarthana:
## What it does
Fixes #3047

## Author checklist


stephan-herrmann:
[Commit 19be8b3](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3063/commits/19be8b31607295fa3050e47dde7dac37b03c5355) got a bit more complex than expected for two reasons:
* Several tests were indeed ill-configured using minimal system libraries that didn't match the project compliance etc. Also one such library was found to be quite incomplete.
  * a bunch of tests configured for "1.8" indeed do not supply any system library at all:
`createJavaProject("P", new String[] { "" }, new String[0], "");`
* based on the decision to look for module `java.base`, it get's a bit blurry if that module is not found:
  * I added a fallback looking for class `java.lang.Object` if no module was found at all.
  * the code needs to differentiate if explicit classpaths are provided or if that information should come from an IJavaProject
  * still at 1.8 the code keeps silent (for pragmatic reasons, partly because of the number of lib-less tests)

The current state is a compromise trying to raise the exception in most relevant situations.

We could probably simplify the code and also check at compliance 1.8 if we only look for class `java.lang.Object`. I didn't make this the one and only criterion because I feared that searching deep into details of all classpath entries might degrade performance, more so than just looking for `module-info` at the root of each entry.

If, however, we assume when `RESOLVE_BINDING` is requested, the compiler will need to lookup `java.lang.Object` anyway, then we could reduce the checks to that class lookup (but would need to "fix" more tests which don't have system libraries configured).

@jarthana  wdyt?

stephan-herrmann:
> * Several tests were indeed ill-configured using minimal system libraries that didn't match the project compliance

Some locations seemed to suffer from the confusion between 1.8 and 18. E.g., project `Converter_15_1` uses compliance "15" but as library "CONVERTER_JCL18_LIB" was used, which is for "1.8"!

jarthana:
Looks good, Stephan. I just looked at what we do in the IDE. There, we seem to somehow turn that into a build path problem. Probably lot of work goes behind that.

Issue #3066, opened 2024-10-10
[Switch][Record Patterns] Unchecked conversion error missing from ECJ


srikanth-sankaran:
Forked off from https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1735#issuecomment-2373079869

This program compiles with ECJ but it shouldn't. Object cannot be safely cast to X<String> and so the program should be rejected and javac does correctly reject it.

```
public record X<T>(int x) {
	public static void main(String[] args) {
		Object o = new Object();
		switch (o) {
		case X<String>(int x):
		default:
		}
	}
}
```

This problem arises from this line in `org.eclipse.jdt.internal.compiler.ast.CaseStatement.resolveCasePattern(BlockScope, TypeBinding, TypeBinding, SwitchStatement, Pattern, boolean)` that reads:

```
		if (!type.isReifiable()) {
			if (expressionType != TypeBinding.NULL && !(e instanceof RecordPattern)) {
				boolean isLegal = e.checkCastTypesCompatibility(scope, type, expressionType, e, false);
				if (!isLegal || (e.bits & ASTNode.UnsafeCast) != 0) {
					scope.problemReporter().unsafeCastInInstanceof(e, type, expressionType);
				}
			}
		}
```

The additional gating for  `!(e instanceof RecordPattern)` looks suspect - If I remove it the program above is rejected correctly but
there are 23 additional failures now:

```
 org.eclipse.jdt.core.tests.compiler.regression.SwitchPatternTest.testGH1955 
 org.eclipse.jdt.core.tests.compiler.regression.SwitchPatternTest.testIssueExhaustiveness_005 
 org.eclipse.jdt.core.tests.compiler.regression.SwitchPatternTest.testGH1955 
 org.eclipse.jdt.core.tests.compiler.regression.SwitchPatternTest.testIssueExhaustiveness_005 
 org.eclipse.jdt.core.tests.compiler.regression.RecordPatternTest.testIssue1224_7 
 org.eclipse.jdt.core.tests.compiler.regression.RecordPatternTest.testIssue1336_2 
 org.eclipse.jdt.core.tests.compiler.regression.RecordPatternTest.testIssue1804 
 org.eclipse.jdt.core.tests.compiler.regression.RecordPatternTest.testIssue1804_4 
 org.eclipse.jdt.core.tests.compiler.regression.RecordPatternTest.testIssue1804_5 
[...]

Issue #3038, opened 2024-10-01
[Sealed types] ParameterizedTypeBinding.permittedTypes() needs more rigor


srikanth-sankaran:
In order to fix https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3009, I have made some major changes to `org.eclipse.jdt.internal.compiler.lookup.ParameterizedTypeBinding.permittedTypes()` that are in the right direction - the prior implementation is highly suspect.

However, we may have to iteratively evolve a more rigorous implementation that would handle all complex cases. 

In private conversations with @stephan-herrmann, three things were discussed that would/could need follow up:

1. Do we need the counterpart of `org.eclipse.jdt.internal.compiler.lookup.TypeBinding.findSuperTypeOriginatingFrom(TypeBinding)` say `findSubtypeResultingIn` that would answer given a supertype possibly parameterized, what subtype possibly parameterized would have the given supertype as its supertype.
2.  Do we even need to compute the subtype parameterization at all to determine switch exhaustiveness ?
3. To quote Stephan: "As this is a statement about all possible instantiations of a generic type, I really feel that type inference is the suitable means to find the unknown type, or prove its absence. Inference also has all the means for expressing in its solution type arguments that are not or only partially determined. I don't see how else you can handle arbitrary type parameters introduced in the subclass, with arbitrary dependencies etc."

Quite plausibly. 

This ticket is raised to follow up on these

srikanth-sankaran:
> In order to fix #3009, I have made some major changes to `org.eclipse.jdt.internal.compiler.lookup.ParameterizedTypeBinding.permittedTypes()` that are in the right direction - the prior implementation is highly suspect.
> 
> However, we may have to iteratively evolve a more rigorous implementation that would handle all complex cases.
> 
> In private conversations with @stephan-herrmann, three things were discussed that would/could need follow up:
> 
> 1. Do we need the counterpart of `org.eclipse.jdt.internal.compiler.lookup.TypeBinding.findSuperTypeOriginatingFrom(TypeBinding)` say `findSubtypeResultingIn` that would answer given a supertype possibly parameterized, what subtype possibly parameterized would have the given supertype as its supertype.

We need the functionality - yes - but may be not such an internal API - this function is now buried into `org.eclipse.jdt.internal.compiler.lookup.ParameterizedTypeBinding.permittedTypes()`

> 2. Do we even need to compute the subtype parameterization at all to determine switch exhaustiveness ?

[...]

Issue #2720, opened 2024-07-17
[Sealed Types + Enhanced Switch] Incorrect diagnostic about switch not being exhaustive


srikanth-sankaran:
Reported here by @nlisker : https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2719#issuecomment-2232303720

I've just ran into what seems to be the same problem:

```
sealed interface I {

	enum E implements I {
		A, B, C;
	}
} 

class Test {

	void d(I i) {
		switch (i) { // error: An enhanced switch statement should be exhaustive; a default label expected
			case I.E.A -> {}
			case I.E.B -> {}
			case I.E.C -> {}
		}
	}
}
```
But there are no other valid cases (I must be an E, which must be A, B or C).

Didn't test with javac.

Version: 2024-06 (4.32)
Build id: I20240601-0610

nlisker:
Incorrect behavior with 4.33 as well. This time I tested with javac 22 and compilation passes (correctly).

PR #3069, opened 2024-10-10
[Switch][Record Patterns] Unchecked conversion error missing from ECJ

stephan-herrmann:
+ try to balance responsibility about type checking between
  - Pattern.isApplicable() - don't yet report in nested position
  - RecordPattern.resolveType() - failed to check isApplicable()
  - InstanceOfExpression.resolveType() - avoid pattern-agnostic checks

+ added the 1st(!) test for inapplicable record pattern in switch

fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3066
fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3074

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/InstanceOfExpression.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/InstanceOfExpression.java
@@ -289,7 +289,7 @@ public TypeBinding resolveType(BlockScope scope) {
 			if (expressionType != TypeBinding.NULL) {
 				boolean isLegal = checkCastTypesCompatibility(scope, checkedType, expressionType, this.expression, true);
 				if (!isLegal || (this.bits & ASTNode.UnsafeCast) != 0) {
-					scope.problemReporter().unsafeCastInInstanceof(this.expression, checkedType, expressionType);
+					scope.problemReporter().unsafeCastInTestingContext(this.expression, checkedType, expressionType);
 				} else  {
 					checkRefForPrimitivesAndAddSecretVariable(scope, checkedType, expressionType);
 				}
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/InstanceOfExpression.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/InstanceOfExpression.java
@@ -277,8 +277,12 @@ public TypeBinding resolveType(BlockScope scope) {
 	if (expressionType == null || checkedType == null)
 		return null;
 
-	if (this.pattern instanceof RecordPattern) // subsequent tests are not relevant for RecordPatterns
+	if (this.pattern != null) {
+		if (this.pattern.isApplicable(expressionType, scope, this)) {
+			checkForPrimitives(scope, checkedType, expressionType);
+		}
 		return this.resolvedType = TypeBinding.BOOLEAN;
+	}
 
 	if (!checkedType.isReifiable()) {
 		CompilerOptions options = scope.compilerOptions();
@@ -297,17 +301,18 @@ public TypeBinding resolveType(BlockScope scope) {
 		}
 	} else if (checkedType.isValidBinding()) {
 		// if not a valid binding, an error has already been reported for unresolved type
-		if ((expressionType != TypeBinding.NULL && expressionType.isBaseType()) // disallow autoboxing
-				|| checkedType.isBaseType()
-				|| !checkCastTypesCompatibility(scope, checkedType, expressionType, null, true)) {
-			checkForPrimitives(scope, checkedType, expressionType);
-		}
+		checkForPrimitives(scope, checkedType, expressionType);
 	}
 
 	return this.resolvedType = TypeBinding.BOOLEAN;
 }
 
 private void checkForPrimitives(BlockScope scope, TypeBinding checkedType, TypeBinding expressionType) {
+	boolean needToCheck = (expressionType != TypeBinding.NULL && expressionType.isBaseType()) // disallow autoboxing
+				|| checkedType.isBaseType()
+				|| !checkCastTypesCompatibility(scope, checkedType, expressionType, null, true);
+	if (!needToCheck)
+		return;
 	PrimitiveConversionRoute route = Pattern.findPrimitiveConversionRoute(checkedType, expressionType, scope);
 	this.testContextRecord = new TestContextRecord(checkedType, expressionType, route);
 
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/InstanceOfExpression.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/InstanceOfExpression.java
@@ -277,6 +277,9 @@ public TypeBinding resolveType(BlockScope scope) {
 	if (expressionType == null || checkedType == null)
 		return null;
 
+	if (this.pattern instanceof RecordPattern) // subsequent tests are not relevant for RecordPatterns
+		return this.resolvedType = TypeBinding.BOOLEAN;
+
 	if (!checkedType.isReifiable()) {
 		CompilerOptions options = scope.compilerOptions();
 		// Report same as before for older compliances
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Pattern.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Pattern.java
@@ -165,7 +165,7 @@ public abstract class Pattern extends Expression {
 				applicable = false;
 			}
 			if ((this.bits & ASTNode.UnsafeCast) != 0) {
-				scope.problemReporter().unsafeCastInInstanceof(this, this.outerExpressionType, patternType);
+				scope.problemReporter().unsafeCastInTestingContext(this, patternType, this.outerExpressionType);
 			}
 		}
 		return applicable;
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Pattern.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Pattern.java
@@ -139,36 +139,39 @@ public abstract class Pattern extends Expression {
 	}
 
 	// 14.30.3 Properties of Patterns: A pattern p is said to be applicable at a type T if ...
-	protected boolean isApplicable(TypeBinding other, BlockScope scope) {
-		TypeBinding patternType = this.resolvedType;
-		if (patternType == null) // ill resolved pattern
-			return false;
+	protected boolean isApplicable(TypeBinding expressionType, BlockScope scope, ASTNode location) {
+		if (expressionType == TypeBinding.NULL)
+			return true;
+		TypeReference typeRef = getType();
+		if (typeRef == null)
+			return true; // nothing to be checked for wildcard '_'
+		TypeBinding patternType = typeRef.resolvedType;
+		if (patternType == null || !patternType.isValidBinding() || !expressionType.isValidBinding())
+			return false; // problem already reported
+
 		// 14.30.3 Properties of Patterns doesn't allow boxing nor unboxing, primitive widening/narrowing (< JLS23)
-		if (patternType.isBaseType() != other.isBaseType() && !JavaFeature.PRIMITIVES_IN_PATTERNS.isSupported(scope.compilerOptions())) {
-			scope.problemReporter().incompatiblePatternType(this, other, patternType);
+		if (patternType.isBaseType() != expressionType.isBaseType() && !JavaFeature.PRIMITIVES_IN_PATTERNS.isSupported(scope.compilerOptions())) {
+			scope.problemReporter().notCompatibleTypesError(location, expressionType, patternType);
 			return false;
 		}
-		boolean applicable = true;
 		if (patternType.isBaseType()) {
 			PrimitiveConversionRoute route = Pattern.findPrimitiveConversionRoute(this.resolvedType, this.outerExpressionType, scope);
-			if (!TypeBinding.equalsEquals(other, patternType)
+			if (!TypeBinding.equalsEquals(expressionType, patternType)
 					&& route == PrimitiveConversionRoute.NO_CONVERSION_ROUTE) {
-				scope.problemReporter().incompatiblePatternType(this, other, patternType);
+				scope.problemReporter().notCompatibleTypesError(location, expressionType, patternType);
 				return false;
 			}
 		} else {
-			if (!checkCastTypesCompatibility(scope, patternType, other, null, true)) {
-				if (this.enclosingPattern != null) {
-					scope.problemReporter().incompatiblePatternType(this, other, patternType);
-					return false;
-				}
-				applicable = false;
+			if (!checkCastTypesCompatibility(scope, patternType, expressionType, null, true)) {
+				scope.problemReporter().notCompatibleTypesError(location, expressionType, patternType);
+				return false;
 			}
 			if ((this.bits & ASTNode.UnsafeCast) != 0) {
-				scope.problemReporter().unsafeCastInTestingContext(this, patternType, this.outerExpressionType);
+				scope.problemReporter().unsafeCastInTestingContext(location, patternType, this.outerExpressionType);
+				return false;
 			}
 		}
-		return applicable;
+		return true;
 	}
 
 	public abstract boolean dominates(Pattern p);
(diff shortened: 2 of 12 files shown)


stephan-herrmann:
@srikanth-sankaran  please have a look at the different parties involved in compatibility checking for (record) patterns. Quite likely the code can be simplified by being more explicit about which part is responsible for which aspect of checking.

srikanth-sankaran:
Thanks for the quick follow up Stephan. I'll review this Monday first thing - tomorrow is a holiday in India. Not just a regular holiday but one where you are not supposed to use any tools of your trade!  😊

(https://en.m.wikipedia.org/wiki/Ayudha_Puja) 

And after I wrap up https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2720 early next week I’ll join you in the grammar work in full earnest.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Pattern.java
```diff
@@ -148,18 +148,27 @@ protected boolean isApplicable(TypeBinding other, BlockScope scope) {
...
-			scope.problemReporter().incompatiblePatternType(this, other, patternType);
-			return false;
+		} else {
+			if (!checkCastTypesCompatibility(scope, patternType, other, null, true)) {
+				if (this.enclosingPattern != null) {
+					scope.problemReporter().incompatiblePatternType(this, other, patternType);
+					return false;
+				}
+				applicable = false;
+			}
```
srikanth-sankaran, line 166: The way this is coded it appears there could be a scenario where the cast is illegal but also unsafe. I doubt that to be valid ever.

If I change `applicable = false;` to be `return applicable = false;` I don't see any test failing. 

If this is a valid observation, it could allow for some simplification of the code by we could get rid of the new variable

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/RecordPatternTest.java
```diff
@@ -4686,4 +4686,54 @@ static void foo(Object o) {
...
+				}
+				"""
+			},
+			"""
+			----------
+			1. ERROR in X.java (at line 5)
+				case X<String>(int x):
+				     ^^^^^^^^^^^^^^^^
+			Type X<String> cannot be safely cast to Object
+			----------
```
srikanth-sankaran, line 4711: This looks backwards - no ? The error is that Object cannot be safely cast to X<String> and not the other way around ??

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/RecordPatternTest.java
```diff
@@ -2008,8 +2008,8 @@ public void testRecordPatternTypeInference_011() {
 				"----------\n" +
 				"1. ERROR in X.java (at line 10)\n" +
 				"	if (p instanceof R<>(String a)) {\n" +
-				"	    ^^^^^^^^^^^^^^^^^^^^^^^^^^\n" +
-				"Incompatible conditional operand types R<capture#1-of ? extends I> and R\n" +
+				"	                 ^^^^^^^^^^^^^\n" +
+				"Type mismatch: cannot convert from R<capture#1-of ? extends I> to R\n" +
```
srikanth-sankaran, line 2012: This test doesn't compile with javac so unable to verify behavior - bug in javac 

I get:

```
X.java:10: error: illegal start of type
        if (p instanceof R<>(String a)) {
                           ^
1 error

```

srikanth-sankaran (changes requested):
I am happy to look over once the current set of feedback is addressed.

Still feel it is a bit of a smell that we have in CaseStatement this:

`if (expressionType != TypeBinding.NULL && !(e instanceof RecordPattern)) {` 

There is something in the flow that could be adjusted so we don't have to have this check for type test patterns alone - what obviates it for record patterns, why doesn't the same thing obviate it for non-record patterns too ? Don't know yet

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/RecordPatternTest.java
stephan-herrmann: ups, right, we keep confusing ourselves by naming parameters as "right" and "left", which was ok-ish, while only one syntax was involved, but in pattern matching, where is right, where is left??
(ironically, I just fixed the same kind of bug a few lines up, and then introduced a new instanced of the bug here).
Next commit will include a bit of renaming.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/InstanceOfExpression.java
```diff
@@ -277,6 +277,9 @@ public TypeBinding resolveType(BlockScope scope) {
 	if (expressionType == null || checkedType == null)
 		return null;
 
+	if (this.pattern instanceof RecordPattern) // subsequent tests are not relevant for RecordPatterns
+		return this.resolvedType = TypeBinding.BOOLEAN;
+
```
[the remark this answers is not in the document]
stephan-herrmann: reworked in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3069/commits/df50953411e7818a54e426a88611a7642e8c7d1b - do you like it better now?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Pattern.java
stephan-herrmann: reworked in https://github.com/eclipse-jdt/eclipse.jdt.core/commit/df50953411e7818a54e426a88611a7642e8c7d1b - do you like it better now?

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/RecordPatternTest.java
stephan-herrmann: Fixed in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3069/commits/bc53322ad7d9e8d9e2ceabcac50db530e08b6e45

stephan-herrmann:
@srikanth-sankaran after my first commit focused on (local) correctness, in https://github.com/eclipse-jdt/eclipse.jdt.core/commit/df50953411e7818a54e426a88611a7642e8c7d1b I made an attempt to define a basic rule for assigning responsibilities:
* `Pattern.isApplicable()` is now responsible for most type checking of patterns
* "clients" `InstanceOfExpression` and `CaseStatement` are now responsible for calling `isApplicable()` and integrating this call with there overall task of semantic analysis. Similarly for `RecordPattern` regarding nested patterns.

I could imagine that this can be taken even one step further. There are still differences between  `InstanceOfExpression` and `CaseStatement` for which I don't readily see the reason, e.g., regarding integration with `isReifiable()` and handling of primitives.

I also slightly unified problem reporting: let the problem reporter select the problemID based on the error location rather than defining separate reporting methods. As one effect this allowed me to remove a reporting method, which was a bit sloppy with short vs long readable names.

I'm not yet 100% convinced of the idea to share `IncompatibleTypesInEqualityOperator` and `IncompatibleTypesInConditionalOperator`. Trying to clean up this area (overlap with checking for `ConditionalExpression`) also led to filing #3089 

Let me know if you like better now, or if you see potential for further clarification.

srikanth-sankaran:
Yes, this is on my radar. I will review this  before tomorrow. Thanks for your patience.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/SwitchPatternTest.java
```diff
@@ -2651,11 +2651,6 @@ public void testBug573921_7() {
 				"	case List<String> s: \n" +
 				"	     ^^^^^^^^^^^^^^\n" +
 				"Type Object cannot be safely cast to List<String>\n" +
-				"----------\n" +
-				"3. ERROR in X.java (at line 9)\n" +
-				"	case List<String> s: \n" +
-				"	     ^^^^^^^^^^^^^^\n" +
-				"This case label is dominated by one of the preceding case labels\n" +
 				"----------\n");
```
srikanth-sankaran, line 2654: I gather this diagnostic going away is because of the early exit in CaseStatement ? If so, OK

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/PrimitiveInPatternsTestSH.java
```diff
@@ -2302,6 +2302,28 @@ int foo4(int myInt) {
...
+				},
+				"""
+				----------
+				1. ERROR in X.java (at line 3)
+					if (this instanceof int i)
+					    ^^^^^^^^^^^^^^^^^^^^^
+				Incompatible conditional operand types X and int
+				----------
+				""");
+	}
```
srikanth-sankaran, line 2325: I so much like javac's message : `incompatible types: X cannot be converted to int` - I don't know where this conditional operand types terminology originated but I am OK with that staying and perhaps being taken up separately.

srikanth-sankaran (approved):
Looks good - except for the minor points called out.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/problem/ProblemReporter.java
```diff
@@ -7377,28 +7377,18 @@ public void notCompatibleTypesError(EqualExpression expression, TypeBinding left
...
-public void notCompatibleTypesError(Expression expression, TypeBinding leftType, TypeBinding rightType) {
-	String leftName = new String(leftType.readableName());
-	String rightName = new String(rightType.readableName());
-	String leftShortName = new String(leftType.shortReadableName());
-	String rightShortName = new String(rightType.shortReadableName());
-	if (leftShortName.equals(rightShortName)){
-		leftShortName = leftName;
-		rightShortName = rightName;
+	int problemId = IProblem.IncompatibleTypesInEqualityOperator; // name is misleading, also used for instanceof patterns
+	if (location instanceof Pattern p && p.getEnclosingPattern() instanceof RecordPattern) {
```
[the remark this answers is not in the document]
stephan-herrmann: right, the comment looks a bit stale, but the general question stays: is it appropriate to use problem ids designed for conditional expression and equality expression when reporting problems of pattern matching?

I'll remove the comment and raise a new issue for that discussion.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/SwitchPatternTest.java
stephan-herrmann: Right, when `isApplicable()` returns false, we've seen enough, `resolveCasePattern()` returns `NotAConstant`, hence no `ResolvedCase` is created and `SwitchStatement.resolve()` has nothing to check dominance for.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/PrimitiveInPatternsTestSH.java
stephan-herrmann: 'conditional operand types' has been in use for instanceof expressions for at least very very long time (like: for ever?), see, e.g., org.eclipse.jdt.core.tests.compiler.regression.CastTest.test027()

I filed #3119 for follow-up.

stephan-herrmann:
@srikanth-sankaran  the failure in SwitchPatternTest22.testIssue3109 reminds me of our discussion around #3024:
```java
			import java.util.function.Predicate;

			public sealed interface Maybe<T>  {
...
			  default Maybe<T> filter(Predicate<T> predicate){
			    return switch (this) {
			      case Some(T value) when predicate.test(value) -> this;
			      case Some(_), None<T> _ -> empty();
			    };
			  }
			  record Some<T>(T value) implements Maybe<T> {}
			  record None<T>() implements Maybe<T>{}
			}
```
With the patch here we report 
```
 case Some(T value) when predicate.test(value) -> this;
           ^^^^^^^
Type Object cannot be safely cast to T
```

But from `this : MayBe<T>` we _could_ "deduce" that a matching `Some` would automatically be a `Some<T>`.

Do your think you implementation from #3024 could be extracted / generalized so it can be used for `isApplicable()` to "improve" the provided `expressionType`? Currently we get Object to be matched against `T` but from the switchExpressionType `MayBe<T>` I would hope we would come up with a better type??

srikanth-sankaran:
> But from `this : MayBe<T>` we _could_ "deduce" that a matching `Some` would automatically be a `Some<T>`.
> 
> Do your think you implementation from #3024 could be extracted / generalized so it can be used for `isApplicable()` to "improve" the provided `expressionType`? Currently we get Object to be matched against `T` but from the switchExpressionType `MayBe<T>` I would hope we would come up with a better type??

Quite possibly. I am planning to work on https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3038 which should solve https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3121 later this week - I will have the concern from here in the reckoning too.

(In my mind an _ideal_ fix for the present problem would get rid of the instanceof check `if (expressionType != TypeBinding.NULL && !(e instanceof RecordPattern)) {` from `org.eclipse.jdt.internal.compiler.ast.CaseStatement.resolveCasePattern` and that may become possible with the approach you hint at)

srikanth-sankaran:
I'll try and propose a patch for  https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3038 that will also address https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3121 and the present ticket https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3069

Bonus points if I manage to get https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3119 and https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3074

I’ll be in touch as I need consulting help.

stephan-herrmann:
> (In my mind an _ideal_ fix for the present problem would get rid of the instanceof check `if (expressionType != TypeBinding.NULL && !(e instanceof RecordPattern)) {` from `org.eclipse.jdt.internal.compiler.ast.CaseStatement.resolveCasePattern` and that may become possible with the approach you hint at)

Sorry, I simply forgot to remove that part, done in [7d2f381](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3069/commits/7d2f3816e3019b7612b1795282a19f6022a10e02) :)

stephan-herrmann:
> I'll try and propose a patch for #3038 that will also address #3121 and the present ticket #3069

Actually I do prefer to let this PR resolve itself, rather than mixing it with unrelated issues. Well, there is an indirect connection: in [e329739](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3069/commits/e329739e14d06e0c296cd8a77a4e89fa124fde1f) I simply extracted part of your work on sealed types (#3024) and applied it for type checking of patterns, which indeed fixes the single regression from the previous version of this PR

If jenkins agrees then this PR should be ready for another round of review. Thanks.

stephan-herrmann:
> > I'll try and propose a patch for #3038 that will also address #3121 and the present ticket #3069
> 
> Actually I do prefer to let this PR resolve itself, rather than mixing it with unrelated issues. Well, there is an indirect connection: in [e329739](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3069/commits/e329739e14d06e0c296cd8a77a4e89fa124fde1f) I simply extracted part of your work on sealed types (#3024) and applied it for type checking of patterns, which indeed fixes the single regression from the previous version of this PR

Actually the entire connection to sealed types (via #3024) was a red herring. The real problem was a missing Pattern#outerExpressionType. This is being fixed by letting GuardedPattern and EitherOrPattern propagate to the contained patterns (filling in an omission from #2827). Then with a valid outerExpressionType **type inference** kicks in :smile: and finds the more specific pattern type, TADA :tada: 

If jenkins agrees then this PR should be ready for another round of review. Thanks.

srikanth-sankaran:
I'll look through the changes later today.

srikanth-sankaran:
> In that light I guess I can proceed based on your previous approval, or do you want to do a final round of reviewing, @srikanth-sankaran ?

The changes look good. Seeing that it also addresses the wrong diagnostic that is the subject matter of https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3074, I have assigned that ticket to you and linked that bug to this PR.

Looks like existing test `org.eclipse.jdt.core.tests.compiler.regression.RecordPatternTest.test47()` - so if you prefer to release as is without adding a regression test for https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3074 it is OK. 

The comment `// this one should report an unsafe cast error\n"` in RecordPatternTest.java:1561 is bogus, but ...

Please proceed to integrate. Thanks.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/RecordPatternTest.java
stephan-herrmann: Good catch. To be continued in #3157.

PR #3114, opened 2024-10-18
Stop building jdt.annotations v1

akurtakov:
Ensure that o.e.jdt.annotations 1.x is not picked by accident when there is unversioned require (like
https://github.com/eclipse-jdt/eclipse.jdt/pull/118 ) as it will no longer be in the reactor.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -58,7 +58,7 @@
 
   <modules>
     <module>org.eclipse.jdt.annotation</module>
-    <module>org.eclipse.jdt.annotation_v1</module>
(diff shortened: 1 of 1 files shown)


stephan-herrmann (approved):
I almost knew I was missing one part, but couldn't quite think of which.
Thanks!

akurtakov:
I'll start new I-build now in case there is smth else missed.

stephan-herrmann:
It's [gone](https://download.eclipse.org/eclipse/updates/4.34-I-builds/I20241018-0730/plugins/):
image

:+1:

PR #3117, opened 2024-10-18
May fix issue #2716

mpalat:
## What it does
Possible fix for issue https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2716
Scenario: JMT.testGetNonJavaResources().

After delete_SP1, model.getNonJavaResources() is used to check whether its deleted or not.
This uses ElementTree tree; which possibly may be out of sync.
The javadoc of Workspace says:
"The tree caches
* the structure and state of files and directories on disk (their existence
* and last modified times). When external parties make changes to
* the files on disk, this representation becomes out of sync. A local refresh
* reconciles the state of the files on disk with this tree (https://github.com/link {https://github.com/link IResource#refreshLocal(int, IProgressMonitor)})."

So, a refreshLocal() may solve this out-of-sync issue.
Point to ponder: performance ?
## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelInfo.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelInfo.java
@@ -14,7 +14,10 @@
 package org.eclipse.jdt.internal.core;
 
 import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IResource;
+import org.eclipse.core.resources.IWorkspaceRoot;
 import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.core.runtime.CoreException;
 
 /**
  * Implementation of IJavaModel. A Java Model is specific to a
@@ -28,7 +31,14 @@ public class JavaModelInfo extends OpenableElementInfo {
  * Compute the non-java resources contained in this java project.
  */
 private Object[] computeNonJavaResources() {
-	IProject[] projects = ResourcesPlugin.getWorkspace().getRoot().getProjects();
+	IWorkspaceRoot workspaceRoot = ResourcesPlugin.getWorkspace().getRoot();
+	try {
+		workspaceRoot.refreshLocal(IResource.DEPTH_INFINITE, null);
+	} catch (CoreException e) {
+		e.printStackTrace();
+		return null;
+	}
+	IProject[] projects = workspaceRoot.getProjects();
 	int length = projects.length;
 	Object[] resources = null;
 	int index = 0;


stephan-herrmann:
@mpalat the **commit message** is really not great. Please **always** include at least a readable headline (like from the issue, **not** just the number) and a description of the solution approach. We want to be able to browse the history in eclipse and make sense of what each commit may contain. Please.

As for the chosen approach: I would be surprised if this helps, because refresh is intended to update workspace resources from the file system if files have been modified outside eclipse (or using java.io.File API). In the case of the test resource deletion happens properly using `IResource` API which to the best of my knowledge guarantees the workspace to be "fresh".

mpalat:
> @mpalat the **commit message** is really not great. Please **always** include at least a readable headline (like from the issue, **not** just the number) and a description of the solution approach. We want to be able to browse the history in eclipse and make sense of what each commit may contain. Please.

Sure @stephan-herrmann  - will take care of this in future.

> As for the chosen approach: I would be surprised if this helps, because refresh is intended to update workspace resources from the file system if files have been modified outside eclipse (or using java.io.File API). In the case of the test resource deletion happens properly using `IResource` API which to the best of my knowledge guarantees the workspace to be "fresh".

possible - investigating..
interestingly I see 1400+ test failures in jdt ui - https://download.eclipse.org/eclipse/downloads/drops4/I20241019-0120/testresults/html/org.eclipse.jdt.ui.tests_ep434I-unit-cen64-gtk3-java17_linux.gtk.x86_64_17.html - resource issues.. need to check whether this is caused by this change.

mpalat:
> possible - investigating.. interestingly I see 1400+ test failures in jdt ui - https://download.eclipse.org/eclipse/downloads/drops4/I20241019-0120/testresults/html/org.eclipse.jdt.ui.tests_ep434I-unit-cen64-gtk3-java17_linux.gtk.x86_64_17.html - resource issues.. need to check whether this is caused by this change.

update: No failures in the next batch of jdt.ui tests :
https://download.eclipse.org/eclipse/downloads/drops4/I20241019-1800/testresults/html/org.eclipse.jdt.ui.tests_ep434I-unit-cen64-gtk3-java17_linux.gtk.x86_64_17.html

Also, no failures in jdt.core tests

iloveeclipse:
There are few issues with the code change itself, one of them is https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3125.

Others: 
- in JDT core, errors should never be logged to system error/out, they should be logged. So I have no clue which actual error happened because I don't start IDE from a shell and never see system output.
- refresh operation may take lot of time, requires workspace lock and may cause further resource events processed - non of that is expected to happen on `IJavaModel.getNonJavaResources()` call that is supposed to be a plain getter.

mpalat:
> There are few issues with the code change itself, one of them is #3125.
> 
> Others:
> 
> * in JDT core, errors should never be logged to system error/out, they should be logged to the Eclipse log. So I have no clue which actual error happened because I don't start IDE from a shell and never see system output.
> * refresh operation may take lot of time, requires workspace lock and may cause further resource events processed - non of that is expected to happen on `IJavaModel.getNonJavaResources()` call that is supposed to be a plain getter.

Thanks @iloveeclipse - I was also suspecting this PR to be the root cause of the jdt.ui failures. 

The original issue of "SP1" still being present after deletion is found again in the latest I build..

PR #3140, opened 2024-10-22
[23] static methods from java.io.IO not imported in implicitly declared classes #3137

### LINKED ISSUE #3137: [23] static methods from java.io.IO not imported in implicitly declared classes
As per JEP 477, the following code should compile, but ECJ reports error on println():

```
	println("Hello");
}
```

jarthana:
## What it does
Fixes #3137

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/parser/ImplicitlyDeclaredClassesTest.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/parser/ImplicitlyDeclaredClassesTest.java
@@ -341,4 +341,27 @@ public class ImplicitlyDeclaredClassesTest extends AbstractRegressionTest9 {
 				"Zork cannot be resolved to a type\n" +
 				"----------\n");
 	}
+	public void testGH3137a() {
+		runConformTest(new String[] {
+				"X.java",
+				"""
+				public static void main(String[] args) {
+					println("Hello1");
+					println("Hello2");
+				}"""
+		},
+		"Hello1\n" +
+		"Hello2");
+	}
+	public void testGH3137b() {
+		runConformTest(new String[] {
+				"X.java",
+				"""
+				public static void main(String[] args) {
+					String str = readln("Enter:");
+					println(str);
+				}"""
+		},
+		"Enter:");
+	}
 }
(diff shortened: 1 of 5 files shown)


org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/parser/ImplicitlyDeclaredClassesTest.java
```diff
@@ -341,4 +341,27 @@ void main() {
...
+		},
+		"Hello1\n" +
+		"Hello2");
+	}
+	public void testGH3137b() {
+		runConformTest(new String[] {
+				"X.java",
+				"""
+				public static void main(String[] args) {
+					String str = readln("Enter:");
```
stephan-herrmann, line 361: @jarthana running this test locally, it just sits there and waits for input from a terminal that does not exist.

Do you have an explanation how that test should work?

stephan-herrmann: Ah, there's more to it:
* in normal test mode, we just wait twice 10 seconds, then the test silently aborts
* with run.javac we use different code for executing which does not have such timeout.

The typical tricks don't help:
* avoiding a main method is not allowed in implicitly declared class
* raising an unrelated error defeats the purpose when we try to compare ecj<->javac.

For now I will just add `JavacTestOptions.SKIP` (for which I need to move the test to the regression package. Hope that's OK.

jarthana: > @jarthana running this test locally, it just sits there and waits for input from a terminal that does not exist.
> 
> Do you have an explanation how that test should work?

All the test cares about is, `readln` is compiled and invoked. It doesn't really care what happens after that. Our tests are only equipped to read from the console up until that point. So, I am OK with that. We can try and send something to the console (I don't know if possible), but I don't think it's worth the effort.

jarthana: Stephan, if you are happy with my explanation, we can skip skipping the tests ;)
Otherwise, I can convert this into a simple compilation test without execution. It's sufficient, IMO.

stephan-herrmann: Sorry to be dense, but I see how your answers help.

Not caring about the execution means: tests will invoke the test program, and wait for timeout, while the test program runs until the physical machine is shut down. I actually suspect this very test to be the reason for lots of exceptions thrown in the sequel.

Without `JavacTestOptions.SKIP` running the test in run.javac mode simply hangs infinitely, which is not a sufficient solution.

So as clean solutions I only see two options:
* send input to the program as you suggest, but I don't have the implementation up my sleeve either :)
* pass a flag to the generic test() method that lets us skip execution even after successful compilation (the opposite of `forceExecution`) - _yet another parameter, oh my_ :smile:

Issue #2522, opened 2024-06-04
[Switch Expression][Null] Bogus null contract violation diagnostic at yield statements


agentgt:
Using the code from #2521 as it is related and null analysis turned on Eclipse frequently thinks the result of a pattern matching expression is nullable. @stephan-herrmann 

```java
public sealed interface PatternMatching {
    
    record Stuff() implements PatternMatching {}
    
    static Stuff match(PatternMatching pm) {
        Stuff r = return switch(pm) {
            case Stuff s -> s;
        }; 
        return r; // Eclipse will say "r" is Nullable here.
    }
}
```

Notice even if we pattern match on `null` making it impossible for r to be `null`:

```java
public sealed interface PatternMatching {
    
    record Stuff() implements PatternMatching {}
    
    static Stuff match(PatternMatching pm) {
        Stuff r = return switch(pm) {
            case Stuff s -> s;
            case null -> throw new NullPointerException();
        }; 
        return r; // Eclipse will say "r" is Nullable here.
    }
}
```

It still does not work.

Version I'm using is JDT RC (3.19.500.v20240601-0610)

Happy to help in anyway I can.

srikanth-sankaran:
Much water has flown under the bridge - I don't see a problem with the snippet above (after mending it for the misplaced return)

But on master HEAD - this is the annotated behavior I see and a couple of surprised I have called out:

```
import org.eclipse.jdt.annotation.NonNull;

public sealed interface PatternMatching {
    
    record Stuff() implements PatternMatching {}
    
    @NonNull
    static Stuff match(PatternMatching pm, int v) {
    	if (v == 0) {
			Stuff r = switch (pm) {
			case Stuff s -> s;
			case null -> throw new NullPointerException();
			};
			return r; // no error here - good
    	} else if (v == 1) {
			Stuff r = switch (pm) {
			case Stuff s -> s;
			case null -> throw new NullPointerException();
			};
			return null; // get error here	- good
    	} else if (v == 2) {
			Stuff r = switch (pm) {
			case Stuff s -> s;
			};
			return r; // no error here -- good
    	} else if (v == 3) {
			Stuff r = switch (pm) {
			case Stuff s -> null; // <<<<<---------------------------- Line 28 - error Why ???
			};
			return r; // get error here - good
    	} else if (v == 4) {
			Stuff r = switch (pm) {
			case Stuff s -> s;
[...]

PR #3154, opened 2024-10-24
Bogus null contract violation diagnostic at yield statements

srikanth-sankaran:
## What it does
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2522

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/YieldStatement.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/YieldStatement.java
@@ -55,8 +55,6 @@ public FlowInfo analyseCode(BlockScope currentScope, FlowContext flowContext, Fl
 
 	flowInfo = this.expression.analyseCode(currentScope, flowContext, flowInfo);
 	this.expression.checkNPEbyUnboxing(currentScope, flowContext, flowInfo);
-	if (flowInfo.reachMode() == FlowInfo.REACHABLE && currentScope.compilerOptions().isAnnotationBasedNullAnalysisEnabled)
-		checkAgainstNullAnnotation(currentScope, flowContext, flowInfo, this.expression);
 
 	targetContext.recordAbruptExit();
 	targetContext.expireNullCheckedFieldInfo();
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/NullAnnotationTests21.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/NullAnnotationTests21.java
@@ -1198,4 +1198,50 @@ public class NullAnnotationTests21 extends AbstractNullAnnotationTest {
 				""";
 		runner.runNegativeTest();
 	}
+
+	// https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2522
+	// Pattern matching on sealed classes cannot infer NonNull (JDK 21)
+	public void testIssue2522_2() {
+		Runner runner = getDefaultRunner();
+		runner.testFiles = new String[] {
+			"PatternMatching.java",
+			"""
+			import org.eclipse.jdt.annotation.*;
+
+			public sealed interface PatternMatching {
+
+			    record Stuff() implements PatternMatching {}
+
+			    @NonNull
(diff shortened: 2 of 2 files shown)


srikanth-sankaran:
@stephan-herrmann my first contribution to null analysis after 10 years hiatus. Be gentle and kind in review comments 😆

stephan-herrmann:
> `but the provided value is null` reads better as `but the provided value may be null` but that is nitpicking

Is that a linguistic hint? AFAICS there is no doubt about null at the two error locations in the new test, so 'is' is more precise than 'may be'. Do you disagree about this analysis?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/YieldStatement.java
```diff
@@ -55,8 +55,6 @@ public FlowInfo analyseCode(BlockScope currentScope, FlowContext flowContext, Fl
 
 	flowInfo = this.expression.analyseCode(currentScope, flowContext, flowInfo);
 	this.expression.checkNPEbyUnboxing(currentScope, flowContext, flowInfo);
-	if (flowInfo.reachMode() == FlowInfo.REACHABLE && currentScope.compilerOptions().isAnnotationBasedNullAnalysisEnabled)
-		checkAgainstNullAnnotation(currentScope, flowContext, flowInfo, this.expression);
```
stephan-herrmann, line 59: So you successfully removed a bogus analysis for the price that NPE caused by `yield` will not be detected at all, right?

I admit that the current code is only useful in the narrow situation of `return switch(v) { ... yield x; }`

I believe the better course of action would be to identify the expected annotated type to which the yield value will be assigned, without taking the shortcut to the method return type.

stephan-herrmann (changes requested):
Welcome back to null analysis, @srikanth-sankaran :)

You found the location that triggered incorrect behaviour. I hope you find my detailed comment helpful to further improve the change.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/YieldStatement.java
srikanth-sankaran: > So you successfully removed a bogus analysis for the price that NPE caused by `yield` will not be detected at all, right?
> 
> I admit that the current code is only useful in the narrow situation of `return switch(v) { ... yield x; }`
>

Hold on, please take a look again! 

I added a new test via a follow up commit that proves that a null injected by `yield` of a `switch` expression that is the directly `return`ed *will be* detected: See `org.eclipse.jdt.core.tests.compiler.regression.NullAnnotationTests21.testIssue2522_2()`

Why would yield require special treatment under return anymore than ternary expression would ? 
 
> I believe the better course of action would be to identify the expected annotated type to which the yield value will be assigned, without taking the shortcut to the method return type.

This happens automatically! Just the same way this is complained against:

```
@NonNull Stuff l = switch (pm) {
					case Stuff s -> s;
					case null -> null; // error: Null type mismatch: required 'PatternMatching.@NonNull Stuff' but the provided value is null
					};
```

So do you have a snippet that shows a case where we would not detect a null injection due to the change here ?? I

srikanth-sankaran:
> > `but the provided value is null` reads better as `but the provided value may be null` but that is nitpicking
> 
> Is that a linguistic hint? AFAICS there is no doubt about null at the two error locations in the new test, so 'is' is more precise than 'may be'. Do you disagree about this analysis?

No, please ignore that comment - I didn't realize we *also* have a variant `but the provided value is inferred as @Nullable` where the expression may be but also may not be null. So we are good.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/YieldStatement.java
stephan-herrmann: Apologies, I forgot about how a switch expression collects result expressions and how various null-checking methods will drill into those result expressions, without the yield expression needing to do anything. The fact that I wrote my previous review comment on a train is only a weak excuse.

stephan-herrmann (approved):
Thanks, @srikanth-sankaran,  for insisting. After more scrutiny I see no problem with this PR. Please go ahead.

Issue #3195, opened 2024-10-29
[24] ModuleImportASTConverterTest.test002 fails after recent merge from master


stephan-herrmann:
org.eclipse.jdt.core.tests.dom.ModuleImportASTConverterTest.test002:
```
Unexpected errors.
----------- Expected ------------

------------ but was ------------
1. ERROR in /Converter_23/src/X.java (at line 1)\n
 /** */\n
 ^\n
The preview feature Implicitly Declared Classes and Instance Main Methods is only available with source level 24 and above\n

---------------------- ----------
 expected:<[]> but was:<[1. ERROR in /Converter_23/src/X.java (at line 1)\n
 /** */\n
 ^\n
The preview feature Implicitly Declared Classes and Instance Main Methods is only available with source level 24 and above\n
]>
```

PR #3196, opened 2024-10-29
[24] ModuleImportASTConverterTest.test002 fails after recent merge from master

stephan-herrmann:
module import is now a preview for **24**, switch to new test project

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3195


stephan-herrmann:
@jarthana the test was introduced on master via 26e33c791. Now in BETA_JAVA24 it needed this adjustment. Are any jenkins builds run for those merge-from-master commits?

Issue #520, opened 2022-11-08
Errors running builder 'Java Builder' on project 'org.eclipse.tycho.core.shared'.


laeubi:
While refactoring i go the following error message:

```
java.lang.NullPointerException: Cannot invoke "org.eclipse.jdt.internal.compiler.lookup.TypeBinding.kind()" because "receiverType" is null
	at org.eclipse.jdt.internal.compiler.lookup.Scope.getMethod(Scope.java:3042)
	at org.eclipse.jdt.internal.compiler.ast.MessageSend.findMethodBinding(MessageSend.java:1020)
	at org.eclipse.jdt.internal.compiler.ast.MessageSend.resolveType(MessageSend.java:835)
	at org.eclipse.jdt.internal.compiler.ast.Expression.resolveTypeExpecting(Expression.java:1147)
	at org.eclipse.jdt.internal.compiler.ast.SwitchExpression.resolveTypeInternal(SwitchExpression.java:522)
	at org.eclipse.jdt.internal.compiler.ast.SwitchExpression.resolveType(SwitchExpression.java:459)
	at org.eclipse.jdt.internal.compiler.ast.ASTNode.resolvePolyExpressionArguments(ASTNode.java:763)
	at org.eclipse.jdt.internal.compiler.ast.MessageSend.findMethodBinding(MessageSend.java:1027)
	at org.eclipse.jdt.internal.compiler.ast.MessageSend.resolveType(MessageSend.java:835)
	at org.eclipse.jdt.internal.compiler.ast.Expression.resolve(Expression.java:1113)
	at org.eclipse.jdt.internal.compiler.ast.AbstractMethodDeclaration.resolveStatements(AbstractMethodDeclaration.java:662)
	at org.eclipse.jdt.internal.compiler.ast.MethodDeclaration.resolveStatements(MethodDeclaration.java:388)
	at org.eclipse.jdt.internal.compiler.ast.AbstractMethodDeclaration.resolve(AbstractMethodDeclaration.java:571)
	at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.resolve(TypeDeclaration.java:1503)
	at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.resolve(TypeDeclaration.java:1628)
	at org.eclipse.jdt.internal.compiler.ast.CompilationUnitDeclaration.resolve(CompilationUnitDeclaration.java:667)
[...]

PR #3208, opened 2024-10-31
Alternate fixes for https://bugs.eclipse.org/bugs/show_bug.cgi?id=576861  & https://github.com/eclipse-jdt/eclipse.jdt.core/issues/520

srikanth-sankaran:
Alternate fixes for https://bugs.eclipse.org/bugs/show_bug.cgi?id=576861 & https://github.com/eclipse-jdt/eclipse.jdt.core/issues/520

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/SwitchExpression.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/SwitchExpression.java
@@ -31,7 +31,15 @@ import org.eclipse.jdt.internal.compiler.flow.FlowContext;
 import org.eclipse.jdt.internal.compiler.flow.FlowInfo;
 import org.eclipse.jdt.internal.compiler.impl.CompilerOptions;
 import org.eclipse.jdt.internal.compiler.impl.Constant;
-import org.eclipse.jdt.internal.compiler.lookup.*;
+import org.eclipse.jdt.internal.compiler.lookup.BlockScope;
+import org.eclipse.jdt.internal.compiler.lookup.FieldBinding;
+import org.eclipse.jdt.internal.compiler.lookup.LocalVariableBinding;
+import org.eclipse.jdt.internal.compiler.lookup.LookupEnvironment;
+import org.eclipse.jdt.internal.compiler.lookup.MethodBinding;
+import org.eclipse.jdt.internal.compiler.lookup.PolyTypeBinding;
+import org.eclipse.jdt.internal.compiler.lookup.Scope;
+import org.eclipse.jdt.internal.compiler.lookup.TypeBinding;
+import org.eclipse.jdt.internal.compiler.lookup.TypeIds;
 
 public class SwitchExpression extends SwitchStatement implements IPolyExpression {
 
@@ -40,8 +48,6 @@ public class SwitchExpression extends SwitchStatement implements IPolyExpression
 	private boolean isPolyExpression = false;
 	private TypeBinding[] originalValueResultExpressionTypes;
 	private TypeBinding[] finalValueResultExpressionTypes;
-	/* package */ Map<Expression, TypeBinding> originalTypeMap = new HashMap<>();
-
 
 	private int nullStatus = FlowInfo.UNKNOWN;
 	public List<Expression> resultExpressions = new ArrayList<>(0);
@@ -330,9 +336,6 @@ public class SwitchExpression extends SwitchStatement implements IPolyExpression
 			if (this.constant != Constant.NotAConstant) {
 				this.constant = Constant.NotAConstant;
 
-				if (this.originalTypeMap == null)
-					this.originalTypeMap = new HashMap<>();
-
 				super.resolve(upperScope);
 
 				resultExpressionsCount = this.resultExpressions.size();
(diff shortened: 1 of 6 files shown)


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/SwitchExpression.java
```diff
@@ -341,31 +344,26 @@ public TypeBinding resolveType(BlockScope upperScope) {
...
+			} else { // re-resolving of poly expression:
 				resultExpressionsCount = this.resultExpressions.size();
 				if (resultExpressionsCount == 0)
 					return this.resolvedType = null; // error flagging would have been done during the earlier phase.
 				boolean yieldErrors = false;
 				for (int i = 0; i < resultExpressionsCount; i++) {
 					Expression resultExpr = this.resultExpressions.get(i);
-					TypeBinding origType = this.originalTypeMap.get(resultExpr);
-					// NB: if origType == null we assume that initial resolving failed hard, rendering re-resolving impossible
-					if (origType != null &&  origType.kind() == Binding.POLY_TYPE) {
```
stephan-herrmann, line 368: I'd need a good explanation why the previous fix was not good. It was quite expensive to achieve at.

srikanth-sankaran: > I'd need a good explanation why the previous fix was not good. It was quite expensive to achieve at.

Hi Stephan, Just so we are clear, I did not say the fix for https://github.com/eclipse-jdt/eclipse.jdt.core/issues/520 was "not good."

But it is not meaningful, if the code whose problem it is fixing is going away.

Let us summarize the picture:

- A switch expression is a poly expression when it features in an assignment or invocation context.
- During the initial resolve of a poly switch expression, target type may or may not be available (invocation - not available, assignment - available)
- When target type is not available in the initial resolve in invocation context, we return a PolyTypeBinding
- When target type becomes available after overload resolution, the switch needs be re-resolved against that target type.
- In this last step, it is only needed to re-resolve poly result expressions. Any standalone result expressions of the switch will not be influenced by the target type imposition and so there is no point in re-resolving them.
- So we need to ask which result expressions need to be re-resolved: For this the fix for `https://bugs.eclipse.org/bugs/show_bug.cgi?id=576861` invented a new mechanism using the newly introduced state `originalTypeMap` which needs to be maintained involving some new code.

I am saying no such new mechanism needs to be invented to map a result expression to be a binding and then querying the kind thusly: `kind() == Binding.POLY_TYPE`

You can simply ask the result expression (or any expression for that matter that has undergone an initial resolve) using the existing API: `org.eclipse.jdt.internal.compiler.ast.Expression.isPolyExpression()`

So this code on master:

```
                                        Expression resultExpr = this.resultExpressions.get(i);
					TypeBinding origType = this.originalTypeMap.get(resultExpr);
					// NB: if origType == null we assume that initial resolving failed hard, rendering re-resolving impossible
					if (origType != null &&  origType.kind() == Binding.POLY_TYPE) {
```
 
becomes 

```
                                        Expression resultExpr = this.resultExpressions.get(i);
                                        if (resultExpr.isPolyExpression()) {
```

How does that sound ?

srikanth-sankaran:
The so called `originalTypeMap` is polluted during re-resolution of the result expressions. That such pollution doesn't have any ill-effect is side issue. It takes some analysis and reasoning to satisfy oneself that the pollution is harmless. Code is much cleaner if new mechanisms are not invented when existing ones will do well enough is the motivation

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/SwitchExpression.java
```diff
@@ -341,31 +344,26 @@ public TypeBinding resolveType(BlockScope upperScope) {
...
+					if (resultExpr.isPolyExpression()) {
```
srikanth-sankaran, line 366: Including a null check here: `if (resultExpr.resolvedType != null && resultExpr.isPolyExpression()) {` would make it fully equivalent to what is on master/HEAD I think

stephan-herrmann:
I was missing the following bit: many parts of a complex switch expression may actually be resolvable even if the switch occurs in invocation context without a target type yet (which makes it different from a lambda with elided types). It is strictly *only poly result expressions* that need to wait for the target type, and those do not have any influence on parts that may have been perfectly resolved during the first iteration.

OTOH, I was a bit confused by your reference in https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3209#issuecomment-2449901822

Once I find the time to fully absorb what you wrote here and in #3209 the fog should subside.

srikanth-sankaran:
> I was missing the following bit: many parts of a complex switch expression may actually be resolvable even if the switch occurs in invocation context without a target type yet (which makes it different from a lambda with elided types). 

And they are already!

See the very early call to `super.resolve(upperScope);` in `org.eclipse.jdt.internal.compiler.ast.SwitchExpression.resolveType(BlockScope)`

The entire switch expression is resolved visiting all the switch rules or the switch block group statements, 

>It is strictly _only poly result expressions_ that need to wait for the target type, and those do not have any influence on parts that may have been perfectly resolved during the first iteration.

Correct. Which is why we want to see if a a result expression is a poly expression in the second entry and only resolve them against the now available target type.

> OTOH, I was a bit confused by your reference in [#3209 (comment)](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3209#issuecomment-2449901822)

That was a comment written up while I myself am/was still in an investigation mode. After having progressed much more in the investigation, I can say this: The control flow in SwitchExpression.resolveType() is not very clear or obviously structured.

The else arm of

```
if (this.constant != Constant.NotAConstant) {
				this.constant = Constant.NotAConstant;

				if (this.originalTypeMap == null)
					this.originalTypeMap = new HashMap<>();

				super.resolve(upperScope);
                                // ...
				}
			} else {
                               // ...
                       }
```

should not actually fall through. The entire block of code after that till the end of the method is ONLY valid for stand alone switch expressions. 

The only "connection" between ConditionalExpression and SwitchExpression is in this lack of precise structuring of the concerns of poly vs standalone and poly initial phase vs poly final phase  
> 
> Once I find the time to fully absorb what you wrote here and in #3209 the fog should subside.

srikanth-sankaran:
I have a better understanding of how things are structured in `SwitchExpression.resolveType` - It is very much modelled after `ConditionalExpression.resolveType` including adopting variable names. By acknowledging key differences between them (in CE, the "result expressions" are not buried deep as is the case for SE), we can actually come up with a cleaner implementation than taking CE's implementation and trying to retrofit that for SE

srikanth-sankaran:
> Once I find the time to fully absorb what you wrote here and in #3209 the fog should subside.

Let me know if it is ok to integrate into master, TIA

stephan-herrmann:
> > Once I find the time to fully absorb what you wrote here and in #3209 the fog should subside.
> 
> Let me know if it is ok to integrate into master, TIA

Please don't wait for me to catch up...

srikanth-sankaran:
> I have a better understanding of how things are structured in `SwitchExpression.resolveType` - It is very much modelled after `ConditionalExpression.resolveType` including adopting/adapting variable names. By acknowledging key differences between them (in CE, the "result expressions" are not buried deep as is the case for SE), we can actually come up with a cleaner implementation than taking CE's implementation and trying to retrofit that for SE

Not only are the result expressions buried deep in SE, we also don't know how many are there upfront, unlike a CE which has exactly two and structurally very visible result expressions `valueIfTrue` and `valueIfFalse` - so a straightforward adaptation of CE into SE may look inorganic. I'll work on these separately.

PR #3214, opened 2024-11-01
Fixes #3213 - NPE and stack overflow protection

### LINKED ISSUE #3213: Stackoverflow protection during recursive call
SourceTypeBinding.methods() and SourceTypeBinding.resolveTypesFor(MethodBinding etc) both make use of 'scope'. If the binding has already been cleaned up, scope is null, and these methods are not null safe. 

Furthermore, getMethods(char[] selector) has a recursive call with no check to avoid a recursive stack overflow. 

Details: 

We're doing some unique things with JDT, and we've run into an NPE, and, after fixing that, a stack overflow, in some unique situations. 

I believe that, upon review, you'll see the patch is safe and unlikely to cause any issues. As for the exact cause, I haven't completely figured it out. An ASTParser parses a file, creates the bindings, and then cleans up. For some reason, a later call to get a specific binding ends up looking for the scope object, which has been cleaned up already. 

While I doubt the exact nature of these details is relevant, and the root cause itself may or may not be worth tracking down (since we're extending and providing alternate implementations for individual parts of JDT, that may be out of scope), but this patch at least attempts to allow us to deal with the results in a safer manner.

robstryker:
## What it does

Please see https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3213 for details. 

## How to test

Automated tests should provide sufficient coverage for this issue. 

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/SourceTypeBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/SourceTypeBinding.java
@@ -1711,6 +1711,9 @@ public RecordComponentBinding getComponent(char[] componentName, boolean needRes
 // NOTE: the return type, arg & exception types of each method of a source type are resolved when needed
 @Override
 public MethodBinding[] getMethods(char[] selector) {
+	return getMethods(selector, true);
+}
+public MethodBinding[] getMethods(char[] selector, boolean shouldTryAgain) {
 	if (!isPrototype())
 		return this.prototype.getMethods(selector);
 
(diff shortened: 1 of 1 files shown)


stephan-herrmann:
This:
```
19:00:37  [ERROR] /home/jenkins/agent/workspace/eclipse.jdt.core-Github_PR-3214/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/CharDelegateMap.java: 
19:00:37  [ERROR] 	/*******************************************************************************
19:00:37  [ERROR] 	^
19:00:37  [ERROR] Internal compiler error: java.lang.RuntimeException: Internal Error compiling /home/jenkins/agent/workspace/eclipse.jdt.core-Github_PR-3214/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/CharDelegateMap.java at org.eclipse.jdt.internal.compiler.ProcessTaskManager.processing(ProcessTaskManager.java:139)
19:00:37  [ERROR] Internal Error compiling /home/jenkins/agent/workspace/eclipse.jdt.core-Github_PR-3214/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/CharDelegateMap.java
```
could possibly indicate that the change actually breaks something. Note that we first build the compiler using changes from the PR and then run the actual jdt.core build using that compiler. So the exception is thrown from the compiler _with_ your changes.

I'd be willing to review this (once tests are green) _if_ you can provide a test case that let's me see the problem :)

robstryker:
Thanks for the feedback and the history. I'll see what I can do.

robstryker:
Based on the history, and the fact that I discovered a workaround (or rather, am likely using the API as intended now), I will not be pursuing this PR any further. 

Workaround is to perform my visitors and fetching bindings within the ASTRequestor's acceptAST method, before cleanup occurs.

PR #3216, opened 2024-11-01
Fixes #3215 - ClassFileWorkingCopy.getFileName() does not follow API doc

### LINKED ISSUE #3215: ClassFileWorkingCopy.getFileName() does not follow spec of IDependent
ClassFileWorkingCopy extends CompilationUnit, which implements org.eclipse.jdt.internal.compiler.env.ICompilationUnit which implements IDependent. 

IDependent has following method: `char[] getFileName();`

The javadoc on this method indicates getFileName() should return a string like `/lib/some.zip|/com/q/Y.class`

CompilationUnit (and it's subclass, ClassFileWorkingCopy) returns the following for this method call:  getPath().toString().toCharArray();

`getPath()` is from IJavaElement, and says it must return basically the path to the archive itself, without the pipe character and package and class details within the archive. 

With this in mind, if a ClassFileWorkingCopy is from inside an archive, it should not be using `getPath().toString().toCharArray()` as its `getFileName()` implementation, because this will only return the path to the archive, and leave out the remaining requirements from `IDependent.getFileName()`. 

	// BIG PROBLEM HERE!!!
	// IJavaElement#getPath() has different specs than
	// org.eclipse.jdt.internal.compiler.env.IDependent#getFileName()

robstryker:
See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3215 

## Author checklist


org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/ClassFileWorkingCopy.java
```diff
@@ -79,6 +83,26 @@ public IPath getPath() {
...
+	PackageFragmentRoot root = getPackageFragmentRoot();
+	if (root == null)
+		// working copy not in workspace
+		ret = new Path(getElementName()).toString().toCharArray();
+	else if (root.isArchive()) {
+		String clazzFileName = this.classFile.getElementName();
+		String parentPath = this.classFile.getParent().getPath().toString();
+		IPackageFragment enclosingPackage = (IPackageFragment)getAncestor(IJavaElement.PACKAGE_FRAGMENT);
+		String pack = enclosingPackage == null ? "" : enclosingPackage.getElementName();
+		String packReplaced = pack.length() > 0 ? pack.replaceAll("\\.", "/") + "/" : "";
```
stephan-herrmann, line 98: where do you see a requirement to use a specific separator?

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/util/Util.java
```diff
@@ -1001,6 +1003,9 @@ public static IPackageFragment getPackageFragment(char[] fileName, int pkgEnd, i
 		if (jarSeparator != -1) {
 			String jarMemento = new String(fileName, 0, jarSeparator);
 			PackageFragmentRoot root = (PackageFragmentRoot) JavaCore.create(jarMemento);
+			if( root == null ) {
+				root = getJarPkgFragmentRoot(new String(fileName), jarSeparator, jarMemento);
+			}
```
stephan-herrmann, line 1008: what motivates this change in a PR addressing a problem in ClassFileWorkingCopy?

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/ClassFileWorkingCopy.java
robstryker: The documentation for IDependent.getFileName(); 

In summary, the two examples provided are:
```
 *  "c:\\lib\\some.jar|/com/p/X.class" or
 *  "/lib/some.zip|/com/q/Y.class".
```

It uses a system path for the resource, followed by the jar separator, followed by the package with forward slashes only for both major operating systems.  

```
/**
 * Answer the file name which defines the type.
 *
 * The path part (optional) must be separated from the actual
 * file proper name by a separator suitable for the type (java.io.File.separator for example),
 * e.g.
 *  "c:\\source\\com\\p\\X.java" or
 *  "/com/p/Y.java".
 *
 * The path to the zip or jar file (optional) must be separated
 * from the actual path part by JAR_FILE_ENTRY_SEPARATOR,
 * e.g.
 *  "c:\\lib\\some.jar|/com/p/X.class" or
 *  "/lib/some.zip|/com/q/Y.class".
 *
 * The proper file name includes the suffix extension (e.g.&nbsp;".java")
 * e.g.&nbsp;"c:/org/eclipse/jdt/internal/compileri/env/IDependent.java"
 *
 * Return null if no file defines the type.
 */

char[] getFileName();
```

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/util/Util.java
robstryker: The tests continued to fail will the following stacktrace. I realize this will be impossible to actually follow, as you don't have our underlying implementations. I could try to break this part out and come up with a new specific test.  But anyway, here goes: 

```
Util.getPackageFragment(char[], int, int) line: 1011	
Util.getClassFile(char[]) line: 810	
Util.getUnresolvedJavaElement(TypeBinding, WorkingCopyOwner, Util$BindingsToNodesMap) line: 1611	
TypeBinding.getUnresolvedJavaElement(TypeBinding) line: 566	
TypeBinding.getUnresolvedJavaElement() line: 557	
TypeBinding.getJavaElement() line: 539	
```

`getPackageFragment(char[] fileName, int pkgEnd, int jarSeparator)` appears to be set up to handle inside-jar paths, but, for some reason, it wasn't working for me. In this case, the values passed in were:

```
fileName=/home/rob/apps/eclipse/workspaces/jclMin1.8.jar|java/lang/RuntimeException.class
pkgEnd=57
jarSeparator=47
```

The call to `PackageFragmentRoot root = (PackageFragmentRoot) JavaCore.create(jarMemento);` was returning null  when `jarMemento` has a value of `/home/rob/apps/eclipse/workspaces/jclMin1.8.jar`

I'm not familiar enough with this code to know why it's doing things this way. It appears to me that the currently implementation simply did not function for this specific case. But I did know i had browsed similar code that did properly turn a path of this structure into a PackageFragmentRoot without error, and so I copied it over, and it worked. 

I still maintain that this change to locating the PackageFragmentRoot is at the very least closely related to changing the `getFileName()` implementation from returning the path of the jar to the full path of the resource inside the jar. If you prefer, I can try to break this specific issue out and try to develop a test case for it.

robstryker:
The most recent build of this PR still has 2 failures, but I am unable to replicate those failures locally :|

robstryker:
The two failing tests are not replicatable locally. 

Breakpoints set at the following locations: 
ClassFileWorkingCopy [line: 88] - getFileName()	
CompilationUnit [line: 243] - getFileName()	
Util [line: 1007] - getPackageFragment(char[], int, int)	

Only the breakpoint at CompilationUnit.getFileName() appears to be hit in these two tests, however, the new implementation appears to return the exact same results as the superclass / previous results. So I'm really not sure what could be happening here. 

Any advice is appreciated.

robstryker:
I really believe this issue is worth looking at, though, because the current code is out of spec with its own documentation. With only two failures (that I can't replicate locally), I think it's unlikely these changes have significantly altered the code or results, especially since the two failing tests don't even seem to enter any of the changes I've made except 1.

robstryker:
Is anyone else able to reproduce the 3 failures? I can't reproduce them on my local machine.

stephan-herrmann:
> Is anyone else able to reproduce the 3 failures? I can't reproduce them on my local machine.

Sorry, I can't reproduce running the tests on Linux. Which OS are you on?

As two failures are constantly reported on jenkins, we can't just ignore those, unfortunately. Also I can't find any reports about those tests being unstable.

The failures mention jclMin1.8.jar as the location of expected results, so I also checked if perhaps an outdated jar file is being used in tests, but the logs consistently show `jclMin1.8.jar', file size: 14941` which corresponds to what I see in my workspace.

robstryker:
Thanks for the update. 

I've modified the PR to see if  using the separator character may let the tests pass. Might as well try it and see, right?

robstryker:
The same two tests are still failing: 

[org.eclipse.jdt.core.tests.model.JavaSearchBugsTests.testBug98378](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-3216/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugsTests/testBug98378/)
    [org.eclipse.jdt.core.tests.model.JavaSearchBugsTests.testBug98378b](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-3216/9/testReport/junit/org.eclipse.jdt.core.tests.model/JavaSearchBugsTests/testBug98378b/)

It seems to me that these tests are likely failing due to the actual following of the spec by adding the jar path and the jar separator, and are unrelated to the particular separator character used.

robstryker:
I originally tested these on linux, and could not replicate the issues. 

I've now tested both versions of this patch on a windows machine, and still cannot replicate the failures.

Issue #3192, opened 2024-10-29
Error in JDT Core during AST creation


ax-lothas:
Exception Stack Trace
```
java.lang.ClassCastException: class org.eclipse.jdt.internal.compiler.lookup.LocalVariableBinding cannot be cast to class org.eclipse.jdt.internal.compiler.lookup.FieldBinding (org.eclipse.jdt.internal.compiler.lookup.LocalVariableBinding and org.eclipse.jdt.internal.compiler.lookup.FieldBinding are in unnamed module of loader org.eclipse.osgi.internal.loader.EquinoxClassLoader @13aa8247)
	at org.eclipse.jdt.internal.compiler.ast.NameReference.fieldBinding(NameReference.java:90)
	at org.eclipse.jdt.internal.compiler.flow.FlowContext.nullCheckedFields(FlowContext.java:195)
	at org.eclipse.jdt.internal.compiler.ast.IfStatement.analyseCode(IfStatement.java:71)
	at org.eclipse.jdt.internal.compiler.ast.MethodDeclaration.analyseCode(MethodDeclaration.java:172)
	at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.internalAnalyseCode(TypeDeclaration.java:961)
	at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.analyseCode(TypeDeclaration.java:305)
	at org.eclipse.jdt.internal.compiler.ast.CompilationUnitDeclaration.analyseCode(CompilationUnitDeclaration.java:137)
	at org.eclipse.jdt.core.dom.CompilationUnitResolver.resolve(CompilationUnitResolver.java:1375)
	at org.eclipse.jdt.core.dom.CompilationUnitResolver.resolve(CompilationUnitResolver.java:837)
	at org.eclipse.jdt.core.dom.CompilationUnitResolver.toCompilationUnit(CompilationUnitResolver.java:1467)
	at org.eclipse.jdt.core.dom.CompilationUnitResolver$ECJCompilationUnitResolver.toCompilationUnit(CompilationUnitResolver.java:117)
	at org.eclipse.jdt.core.dom.ASTParser.internalCreateASTCached(ASTParser.java:1263)
	at org.eclipse.jdt.core.dom.ASTParser.lambda$0(ASTParser.java:1142)
	at org.eclipse.jdt.internal.core.JavaModelManager.cacheZipFiles(JavaModelManager.java:5770)
[...]

PR #3217, opened 2024-11-01
Error in JDT Core during AST creation

stephan-herrmann:
+ QualifiedNameReference: fieldBinding() != lastFieldBinding()

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3192


stephan-herrmann:
Method `NameReference.fieldBinding()` even has this warning:
```
/**
 * Use this method only when sure that the current reference is <strong>not</strong>
 * a chain of several fields (QualifiedNameReference with more than one field).
 * Otherwise use {@link #lastFieldBinding()}.
 */
```
I just should've read the warning when preparing #1716...

Issue #2727, opened 2024-07-18
Systematic maintenance of jclMinX.jar and .src.zip


stephan-herrmann:
In https://bugs.eclipse.org/560658 we started to organize maintenance of jcl stubs more systematically, but covered only convertJclMin, not regular jclMin.

Most recently, follow-ups of https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2551 would like to work on a cleaned-up foundation :)

To consider:
* do we actually need two flavors, or can one set of jclMin cover all use cases?
* are all source projects configured for IDE and build using `org.eclipse.jdt.core.tests.model/JCL/build.xml`?
* [2712](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2712) brought up the issue of `JCL_SRCROOT`
  * this is used to configure classpath entries with a src path inside the src.zip archive
  * it will be used by all tests accessing sources of binary types from jclMin (example: `BufferTests.testAppendReadOnly()`)
  * as per `AbstractJavaModelTests.getExternalJCLRootSourcePath()` paths inside src.zip should start with "src"
  * if this is misconfigured, tests may still pass, but require analysing jar and zip to tediously figure out the source root path

stephan-herrmann:
Found some confusion between jclMin & jclFull: Tests CompletionTests18.testBug443091_expectLambdaCompletions_forFunctionalInterfaceArgumentAssignment() and friends from [bug 443091](https://bugs.eclipse.org/bugs/show_bug.cgi?id=443091) use `jclFull1.8.jar` but expect to find a parameter name via the source attachment `jclMin1.8src.zip`.  This is despite the fact that we do have a `jclFull1.8src.zip`, too.

I'd call this a bug in `AbstractJavaModelTests.setUpJCLClasspathVariables(String, boolean)` where `getExternalJCLPathString()` has a parameter `useFullJcl` but `getExternalJCLSourcePath()` constantly uses the min version, hence `jclFull1.8src.zip` some 100% unused atm.

PR #3223, opened 2024-11-02
Systematic maintenance of jclMinX.jar and .src.zip

stephan-herrmann:
+ remove jclMin < 1.8 and adjust ClasspathTests accordingly
+ created jclMin projects from existing jclMin*src.zip
+ added module-info to all modular jcl
+ declared --patch-module in .classpath & build.xml
+ minor clean-up
+ regenerate all jclMin
+ did NOT re-generate converterJclMin - need to check diffs first
+ fix source attachment for jclFull1.8.jar

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2727


stephan-herrmann:
No rocket science, just a bunch of eclipse projects and additions to the existing build.xml.

As we still keep the resulting .jar and .zip in git, rebuilding has to be triggered on every change.

It will be interesting to see, if regenerating on different machines will result in changes of artifacts where the source where not touched.

Currently the ant build will actually pick up .class files created by ecj in the IDE, so perhaps a new version of ecj may create differences in class files.

One issue to highlight: for levels 9+ we would normally get a conflict with module java.base, since we still need to link a JRE for building. I solved this by `--patch-module java.base=src` both in .classpath and build.xml. I was actually surprised to see, that I was still allowed to add module-info.java declaring that vary module `java.base`. _Interesting_.

Issue #3224, opened 2024-11-03
[Internal][Switch Expressions]  Eliminate double book keeping in SwitchExpression handling


srikanth-sankaran:
PR https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3219 has changed the collection used to gather result expressions from being a List to a Set. This is to ensure that we don't add the same result expression multiple times as we resolve poly result expressions more than once.

But certain operations on these expressions assume their representation to be a List. PR https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3219 has already cleaned most of the places where any Iterable<Expression> would suffice.

The left over ones will be addressed via this ticket. The goal is to get rid of the declaration `org.eclipse.jdt.internal.compiler.ast.SwitchExpression.resultExpressions` but have everything else work as it should.

srikanth-sankaran:
@stephan-herrmann - Along the lines of `org.eclipse.jdt.core.tests.RunVariousPatternsTests` and `org.eclipse.jdt.core.tests.RunVariousSealedTypeTests` is there a single wrapper that would run ALL null annotation tests (compiler, model ...) ? Doesn't look like it ? It could be very valuable as a smoke test that we can run in just a few minutes

srikanth-sankaran:
Another query: `org.eclipse.jdt.internal.compiler.lookup.Scope.lowerUpperBound(TypeBinding[])` seems to lose nullness information.  The sole purpose of the code in `org.eclipse.jdt.internal.compiler.ast.SwitchExpression.analyseCode(BlockScope, FlowContext, FlowInfo)` seems to be to compensate for this ??

Issue #3225, opened 2024-11-03
[4.35+] Refactor ConditionalExpression and SwitchExpression by making them implement INumericChoiceContext


srikanth-sankaran:
JLS 5.6 classifies `SwitchExpression` and `ConditionalExpression` as constructs in numeric choice context (when all result expressions are numeric) The computation of result type, their polyness or not all are all similarly defined. Essentially a switch is multi-way-conditional.

As such there is enormous room for code sharing by hoisting contract and default behavior to a new interface type say `INumericChoiceContext` that both switch expressions and ternary operators implement.

Comparing `org.eclipse.jdt.internal.compiler.ast.SwitchExpression.resolveType(BlockScope)` and  `org.eclipse.jdt.internal.compiler.ast.ConditionalExpression.resolveType(BlockScope)` it is possible for the latter to be as minimal as the former.

PR #3235, opened 2024-11-04
[Internal][Switch Expressions] Eliminate double book keeping in SwitchExpression handling

srikanth-sankaran:
* Eliminate double book keeping by getting rid of `org.eclipse.jdt.internal.compiler.ast.SwitchExpression.resultExpressions`
* Eliminate the convenience copy state held in  `org.eclipse.jdt.internal.compiler.ast.SwitchExpression.resultExpressions`resultExpressionNullStatus` as it makes assumptions about the collection (that it is a List/supports `get(int)` method which doesn't gel with the latest changes.
* Rather than use the convenience copy of the null status of result expressions, reach for them from the result expressions when required.
* Inline  `org.eclipse.jdt.internal.compiler.ast.SwitchExpression.computeNullStatus()` into `analyzeCode()` and simplify a bit
* Null annotation tests have carried a measure of bit rot. A bunch of tests were not running - fixed.
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3224


srikanth-sankaran:
@stephan-herrmann, I have merged this PR as I felt the changes are straightforward, but I invite you to look through the changes. 
I am happy to follow up on any problems reported or suggestions.

Detailed description of the changes available in the commit message as well in https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3224

If deemed fine, I can do similar changes for ConditionalExpression via https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3225

PR #3254, opened 2024-11-05
Alternate code generation approach for YieldStatement

srikanth-sankaran:
(Work in progress - still failing a handful of tests in `RunVariousSwitchTests`)


srikanth-sankaran:
The class file comparison is is due a nop being emitted in the new compiler in some scenarios. A follow up ticket will cover it. It is harmless. I have pushed a qualifier update trigger

srikanth-sankaran:
@jarthana - this is ready for your perusal.

srikanth-sankaran:
> The class file comparison is is due a nop being emitted in the new compiler in some scenarios. A follow up ticket will cover it. It is harmless. I have pushed a qualifier update trigger

I have fixed the code not to emit the nop in most scenarios - so we don't need the qualifier update.

PR #3279, opened 2024-11-11
4.35: Accumulated defect fixes - Do not integrate into 4.34

srikanth-sankaran:


srikanth-sankaran:
@jarthana/@iloveeclipse  - who creates milestones labels settable here ? I would like to request labels for 4.35 M1 - I have started accumulating patches here for 4.35 that need not make it to 4.34 on account of these not being critical weighed against the late stage 4.34 is in.

PR #3281, opened 2024-11-11
Withdraw support for the undocumented capability to share finally blocks.

srikanth-sankaran:
* Apropos https://github.com/eclipse-jdt/eclipse.jdt.core/discussions/3246

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TryStatement.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TryStatement.java
@@ -52,6 +52,7 @@ import org.eclipse.jdt.internal.compiler.impl.CompilerOptions;
 import org.eclipse.jdt.internal.compiler.impl.Constant;
 import org.eclipse.jdt.internal.compiler.impl.JavaFeature;
 import org.eclipse.jdt.internal.compiler.lookup.*;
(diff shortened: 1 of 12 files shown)


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TryStatement.java
```diff
@@ -52,6 +52,7 @@
 import org.eclipse.jdt.internal.compiler.impl.Constant;
 import org.eclipse.jdt.internal.compiler.impl.JavaFeature;
 import org.eclipse.jdt.internal.compiler.lookup.*;
+import org.eclipse.jdt.internal.compiler.tool.EclipseCompiler;
```
stephan-herrmann, line 55: @srikanth-sankaran  I saw this flying by, while adopting recent changes to ObjectTeams: just from looking at this import this is a very unexpected relationship.

o.e.j.compiler.tool used to be a separate fragment, invisible to the compiler. Why should the ast package outsource anything into this package?

srikanth-sankaran: Could you suggest a better place ?? I seem to need a C style global variable. I would like to be able to use such for asserting/documenting control flow. Rather than construct an exception object on the spot each time.

I don't know if the compliance we are in converting to enums and have the compiler insert a default is an option. But enum switches may not be the only spots to use this global.

Issue #3283, opened 2024-11-11
[Switch Expressions] ECJ fails to recognize "yield" as a contextual keyword when there is a case fall-through


jarthana:
This was a valid code few days ago, but compilation fails now on HEAD:

```
public class X {
	public static int var = 0;
	public static void main(String argv[]) {
        int j = switch(var){
        case 0: yield (0); // If this is removed, there's a different error
        case 1: // If I remove this line, the error goes away
        case 2: yield 1; // multiple errors
        default:
          throw new IllegalArgumentException("Unexpected value: " + var);
    };
    }
}
```

srikanth-sankaran:
Thanks Jay! I am on it

PR #3284, opened 2024-11-11
ECJ fails to recognize "yield" as a contextual keyword when there is a case fall-through

srikanth-sankaran:
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3283


srikanth-sankaran:
@jarthana - this PR passes RunVariousSwitchTests locally. While we wait for the Jenkins run to finish, you are welcome to play around with it. After it is green, we can schedule a review. Fix is simple and I can walk you through it.

PR #3286, opened 2024-11-11
ECJ fails to recognize "yield" as a contextual keyword when there is a case fall-through

srikanth-sankaran:
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3283


jarthana (approved):

jarthana:
Patch looks good to me. +1 for RC1.

srikanth-sankaran:
Thanks Jay!

PR #3299, opened 2024-11-12
Downgrade ReadOnly Exception to warning

jukzi:
Throwing an Exception here currently causes more pain then just continuing.
Ideally one would rework the getter JavaProject.getResolvedClasspath() to never modify the javamodel, but it's not clear how todo that in a compatible way.

https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1742

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/DeltaProcessor.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/DeltaProcessor.java
@@ -30,6 +30,7 @@ import java.util.Map;
 import java.util.Set;
 import org.eclipse.core.resources.*;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.ILog;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.ISafeRunnable;
@@ -2022,7 +2023,9 @@ public class DeltaProcessor {
 	 */
 	public void registerJavaModelDelta(IJavaElementDelta delta) {
 		if (JavaModelManager.isReadOnly()) {
-			throw new IllegalStateException("Its not allow to modify JavaModel during ReadOnly action. delta=" + delta); //$NON-NLS-1$
+			ILog.get().warn("JavaModel change during read only operation", new IllegalStateException( //$NON-NLS-1$
+					"JavaModel modified during 'read only' operation. Consider to report this warning to https://github.com/eclipse-jdt/eclipse.jdt.core/issues. delta=" //$NON-NLS-1$
+							+ delta));
 		}
 		this.javaModelDeltas.add(delta);
 	}


org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/DeltaProcessor.java
```diff
@@ -2022,7 +2023,8 @@ public void resetProjectCaches() {
 	 */
 	public void registerJavaModelDelta(IJavaElementDelta delta) {
 		if (JavaModelManager.isReadOnly()) {
-			throw new IllegalStateException("Its not allow to modify JavaModel during ReadOnly action. delta=" + delta); //$NON-NLS-1$
+			ILog.get().warn("JavaModel change during readOnly Operation", new IllegalStateException( //$NON-NLS-1$
+					"Its not allow to modify JavaModel during ReadOnly action. delta=" + delta)); //$NON-NLS-1$
```
iloveeclipse, line 2027: Spelling / wording: 
`Its not allow to modify JavaModel during ReadOnly action`
 -> 
`JavaModel modified during "read only" operation. Consider to report this warning to https://github.com/eclipse-jdt/eclipse.jdt.core/issues`.

jukzi: done

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/DeltaProcessor.java
```diff
@@ -2022,7 +2023,9 @@ public void resetProjectCaches() {
 	 */
 	public void registerJavaModelDelta(IJavaElementDelta delta) {
 		if (JavaModelManager.isReadOnly()) {
-			throw new IllegalStateException("Its not allow to modify JavaModel during ReadOnly action. delta=" + delta); //$NON-NLS-1$
+			ILog.get().warn("JavaModel change during readOnly Operation", new IllegalStateException( //$NON-NLS-1$
```
iloveeclipse, line 2026: Sorry, that must be consistent: `readOnly Operation` -> `'read only' operation`

iloveeclipse:
@jukzi : I assume we want this in 4.34 RC2? If so, @trancexpress please also review this PR.

Background: the feature https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/933 that added `org.eclipse.jdt.internal.core.JavaModelManager.isReadOnly()` and a hard check here caused a lot of regressions in related JDT/JDT UI code, last one is https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1742.

So instead of hard failing with exception (and breaking caller code) this PR now allows operation but logs a warning.
From this point of view this shouldn't harm and should fix potential regressions introduced with the original change, therefore my assumption that it would be beneficial for 4.34 to have this PR merged.

jukzi:
> I assume we want this in 4.34 RC2

yes

iloveeclipse (approved):

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/DeltaProcessor.java
```diff
@@ -2022,7 +2023,9 @@ public void resetProjectCaches() {
 	 */
 	public void registerJavaModelDelta(IJavaElementDelta delta) {
 		if (JavaModelManager.isReadOnly()) {
-			throw new IllegalStateException("Its not allow to modify JavaModel during ReadOnly action. delta=" + delta); //$NON-NLS-1$
+			ILog.get().warn("JavaModel change during read only operation", new IllegalStateException( //$NON-NLS-1$
+					"JavaModel modified during 'read only' operation. Consider to report this warning to https://github.com/eclipse-jdt/eclipse.jdt.core/issues. delta=" //$NON-NLS-1$
+							+ delta));
```
trancexpress, line 2028: Do we tell the user to report errors to `github` anywhere else in logged messages? if not I would omit this. Especially on a warning.

Maybe this should be an error though, considering we are expecting trouble to follow after this log (I assume caching somehow will go wrong?).

trancexpress (approved):

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/DeltaProcessor.java
iloveeclipse: > Do we tell the user to report errors to github anywhere else in logged messages? if not I would omit this. Especially on a warning.

In other places we've suggested to report to bugzilla, but that time is over :-)
And yes, it would make sense if users would report these warnings, so we can fix inappropriate code.

Issue #3181, opened 2024-10-29
 [23][Primitive Patterns][Preview] VerifyError during conversion between T extends Long and double


jarthana:
This is similar to #2937 but happens with Long and double. Here's the testcase:

```
record Record<T extends Long>(T t) {}
public class X  {
	public static <T extends Long> double foo(Record<T> s) {
		return switch (s) {
			case Record(double s1) -> s1;
			default -> 0;
		};
	}
    public static void main(String[] args) {
    }
}
```
And the verify error:
```
Caused by: java.lang.VerifyError: Inconsistent stackmap frames at branch target 63
Exception Details:
  Location:
    X.foo(LRecord;)D @52: goto
  Reason:
    Current frame's stack size doesn't match stackmap.
  Current Frame:
    bci: @52
    flags: { }
    locals: { 'Record', 'Record', integer }
    stack: { }
  Stackmap Frame:
    bci: @63
    flags: { }
    locals: { 'Record', 'Record', integer }
    stack: { 'java/lang/Long' }
```

srikanth-sankaran:
This one triggers a `Internal inconsistency: Inappropriate operand stack size encountered during translation` error too

srikanth-sankaran:
@stephan-herrmann - Thanks for taking a look.

Issue #3129, opened 2024-10-21
VerifyError with instanceof record patterns with conversion from double to Long


jarthana:
Testcase:
```
record Record(Long i) {}
public class X  {
    public static double convert(Record r) {
    	return r instanceof Record(double d) ? d : 0;
    }
    public static void main(String[] args) {}
}
```

Exception thrown when run:
```
Caused by: java.lang.VerifyError: Inconsistent stackmap frames at branch target 35
Exception Details:
  Location:
    X.convert(LRecord;)D @15: ifnull
  Reason:
    Current frame's stack size doesn't match stackmap.
  Current Frame:
    bci: @15
    flags: { }
    locals: { 'Record' }
    stack: { 'java/lang/Long', 'java/lang/Long' }
  Stackmap Frame:
    bci: @35
    flags: { }
    locals: { 'Record' }
    stack: { }
```

Note: It's also cool to see the compiler error about "Internal inconsistency: Inappropriate operand stack size encountered during translation". After some more testing, may be we can make that an error instead of warning - @srikanth-sankaran ?

srikanth-sankaran:
Byte code generated:

```
  public static double convert(Record);
    descriptor: (LRecord;)D
    flags: (0x0009) ACC_PUBLIC, ACC_STATIC
    Code:
      stack=4, locals=3, args_size=1
         0: aload_0                                
         1: instanceof    #16                 // class Record
         4: ifeq          35                        // <---------------  reach branch target with stackDepth == 0 
         7: aload_0
         8: checkcast     #16                 // class Record
        11: invokevirtual #18                 // Method Record.i:()Ljava/lang/Long;
        14: dup
        15: ifnull        35                     //  <---------------  reach branch target with stackDepth == 1
        18: iconst_1
        19: ifne          26
        22: pop
        23: goto          35
        26: invokevirtual #22                 // Method java/lang/Long.longValue:()J
        29: l2d
        30: dstore_1
        31: dload_1
        32: goto          36
        35: dconst_0
        36: dreturn
        37: new           #28                 // class java/lang/MatchException
        40: dup_x1
        41: swap
        42: dup
        43: invokevirtual #30                 // Method java/lang/Throwable.toString:()Ljava/lang/String;
        46: swap
        47: invokespecial #36                 // Method java/lang/MatchException."<init>":(Ljava/lang/String;Ljava/lang/Throwable;)V
        50: athrow
```

[...]

PR #3301, opened 2024-11-12
[23][Primitive Patterns][Preview] VerifyError during conversion between T extends Long and double #3181

stephan-herrmann:
+ attempt to compensate for jumps with different start stack depth

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3181

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/RecordPattern.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/RecordPattern.java
@@ -248,15 +248,19 @@ public class RecordPattern extends Pattern {
 					componentType.erasure()))
 				codeStream.checkcast(componentType); // lastComponent ? [C] : [R, C]
 			if (p instanceof RecordPattern || !p.isTotalTypeNode) {
-				if (!p.isUnnamed())
+				BranchLabel innerFailLabel = matchFailLabel;
+				if (!p.isUnnamed()) {
 					codeStream.dup(componentType); // lastComponent ? named ? ([C, C] : [R, C, C]) : ([C] : [R, C])
+					innerFailLabel = new BranchLabel(codeStream);
+				}
 				if (p instanceof TypePattern) {
-					((TypePattern) p).generateTypeCheck(currentScope, codeStream, matchFailLabel);
+					((TypePattern) p).generateTypeCheck(currentScope, codeStream, innerFailLabel);
 				} else {
 					codeStream.instance_of(p.resolvedType); // lastComponent ? named ? ([C, boolean] : [R, C, boolean]) : ([boolean] : [R, boolean])
 				}
 				BranchLabel innerTruthLabel = new BranchLabel(codeStream);
 				codeStream.ifne(innerTruthLabel); // lastComponent ? named ? ([C] : [R, C]) : ([] : [R])
+				innerFailLabel.place();
 				int pops = p.isUnnamed() ? 0 : TypeIds.getCategory(componentType.id); // Not going to store into the component pattern binding, so need to pop, the duped value.
 				Pattern current = p;
 				RecordPattern outer = this;
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/PrimitiveInPatternsTest.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/PrimitiveInPatternsTest.java
@@ -7090,6 +7090,27 @@ public class PrimitiveInPatternsTest extends AbstractRegressionTest9 {
 			},
 			"2");
    }
+
+	public void testGH3181() {
+		runConformTest(new String[] {
+			"X.java",
+			"""
+			record Record<T extends Long>(T t) {}
+			public class X  {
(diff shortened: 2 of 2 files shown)


stephan-herrmann:
@srikanth-sankaran this is the smaller of two fixes I could think of, but I'm neither sure if it's complete, nor if it is good design:

I'm a bit uneasy about the asymmetry between these existing two branches in `RecordPattern.generateCode()`:
```java
	if (p instanceof TypePattern) {
		((TypePattern) p).generateTypeCheck(currentScope, codeStream, innerFailLabel);
	} else {
		codeStream.instance_of(p.resolvedType); // lastComponent ? named ? ([C, boolean] : [R, C, boolean]) : ([boolean] : [R, boolean])
	}
```

While `instance_of` only converts a object value into 0 or 1, as the input to subsequent `ifne` (the positive branch), inside `generateTypeCheck()` different things can happen, one of them being:
```java
				codeStream.ifnull(falseLabel);
				codeStream.iconst_1();
```
i.e., in one case we already *perform* the jump *for the negative case*. This caused different jumps to the same label from locations with different stackDepth!

Would it be cleaner if `generateTypeCheck()` would refrain from any jumping, and only signal some "code" indicating which instruction should be used for type checking (if any)? Unfortunately, the JVM has no bytecode for comparison against `null` *without* jumping (i.e., some `a != null ? 1 : 0` idiom).

Alternatively, we could just flip the jumps to target innerTruthLabel, rather than matchFailLabel/innerFailLabel.

Things are a bit more involved since `InstanceOfExpression.generateOptimizedBoolean()` uses `generateTypeCheck()` *always* and then has several downstream branches with different jump-strategies.

srikanth-sankaran:
> Unfortunately, the JVM has no bytecode for comparison against `null` _without_ jumping (i.e., some `a != null ? 1 : 0` idiom).

Sure, it does! 😆 

This patch fixes the present problem but I need to see if it gels well with `org.eclipse.jdt.internal.compiler.ast.InstanceOfExpression.generateOptimizedBoolean(BlockScope, CodeStream, BranchLabel, BranchLabel, boolean)`'s usage of `generateTypeCheck`

```
diff --git a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/IGenerateTypeCheck.java b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/IGenerateTypeCheck.java
index 5565d8a..b9819b5 100644
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/IGenerateTypeCheck.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/IGenerateTypeCheck.java
@@ -46,8 +46,7 @@
 			}
 			case WIDENING_REFERENCE_AND_UNBOXING_COVERSION,
 			WIDENING_REFERENCE_AND_UNBOXING_COVERSION_AND_WIDENING_PRIMITIVE_CONVERSION -> {
-				codeStream.ifnull(falseLabel);
-				codeStream.iconst_1();
+				codeStream.instance_of(scope.getJavaLangObject());
 				setPatternIsTotalType();
 			}
 			case NARROWING_AND_UNBOXING_CONVERSION -> {
@@ -56,8 +55,7 @@
 			}
 			case UNBOXING_CONVERSION,
 			UNBOXING_AND_WIDENING_PRIMITIVE_CONVERSION -> {
-				codeStream.ifnull(falseLabel);
-				codeStream.iconst_1();
+				codeStream.instance_of(scope.getJavaLangObject());
 				setPatternIsTotalType();
 			}
 			case NO_CONVERSION_ROUTE -> {

```

srikanth-sankaran:
> Would it be cleaner if `generateTypeCheck()` would refrain from any jumping

I much prefer that as it gives the caller flexibility over what to do.

srikanth-sankaran:
The patch passes RVPT - ergo it doesn't irritate InstanceOfExpression.generateOptimizedBoolean - I didn't white box validate that. 

So if this is the final solution, you can start a tally. Seeing that you are getting beverages from others 😆

srikanth-sankaran:
You can use the comments of the form `// lastComponent ? named ? ([C, boolean] : [R, C, boolean]) : ([boolean] : [R, boolean])` in `org.eclipse.jdt.internal.compiler.ast.RecordPattern.generateCode(BlockScope, CodeStream, BranchLabel, BranchLabel)` to keep track of operand stack and ensure that every path to a bci matches up every other path. So statically we can be sure there is no shape violation

stephan-herrmann:
> > Unfortunately, the JVM has no bytecode for comparison against `null` _without_ jumping (i.e., some `a != null ? 1 : 0` idiom).
> 
> Sure, it does! 😆

I said "without jumping", so show me the byte code!

stephan-herrmann:
@srikanth-sankaran  feel free to grab this one.

srikanth-sankaran:
> @srikanth-sankaran feel free to grab this one.

Ok, I will take this forward and help wrap up both https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3129 and https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3181

srikanth-sankaran:
I tried pushing to here, but due to some glitch in my secure store setup I wasn't. So I have created a fresh PR: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3307

Given where we are, it may require two reviews. I'll raise a request there

PR #1698, opened 2023-12-06
Add test for reproducing #1654

haubi:
## What it does
Add test for #1654, based on existing test_github844() method.
The breakage is caused by this combination:
- import some class generated by some annotation processor
- to some other annotation, pass a value that is defined right within the currently compiled project

Beyond the breakage this test also shows small code changes which cause the compiler not to break.

## How to test
Run `AnnotationProcessingCompilerToolTest.test_github1654()` as JUnit Plug-in Test

## Author checklist

- [x] I have thoroughly tested my changes
- [x] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [x] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

PR #3324, opened 2024-11-18
Fix classes being 'lost' when earlier rounds of annotation processing succeed before a later round fails

facboy:
## What it does

Annotation processing can fail with a `SourceTypeCollisionException` when a to-be generated type is referenced by an 'existing' source file. When this happens the compiler 'recompiles' the original files together with the newly generated files. The current implementation is buggy because annotation processing can go through multiple rounds, with each round generating new files. It is possible for earlier rounds to 'succeed', only for a later round to fail with a `SourceTypeCollisionException` requiring a restart. The current implementation 'loses' those files generated in earlier successful rounds.

This PR fixes this issue by ensuring that files from earlier rounds are also included in the retry.

## How to test
Unfortunately I only have a local project I am working on (which is how I discovered the bug). It's a WIP but I could make it available.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/Compiler.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/Compiler.java
@@ -17,9 +17,11 @@
 package org.eclipse.jdt.internal.compiler;
 
 import java.io.PrintWriter;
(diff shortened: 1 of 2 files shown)


jukzi:
Please add a junit test as a separate commit that shows the error and gets solved by this commit. You can use this as sample: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1698

PR #3345, opened 2024-11-25
[GH-2620] consider the cursor location in completions

### LINKED ISSUE #2620: Faulty/Undesired CodeCompletion behaviour
Hello. This situation has been driving me ([and others](https://bugs.eclipse.org/bugs/show_bug.cgi?id=575025)) crazy every minute(!) I am coding with eclipse since forever, so I finally took some time out to at least pinpoint the offending plugin and commit. Hope this helps!

This behaviour started back with eclipe 2021-06 while 2021-03 still behaved "well".

Code completion is behaving in an undesired manner in the following situation:
```Java
public class Demo {
		System.out.println("hello");
		if (true) {
			System.out.print|cursor|ln("world");
		}
	}
}
```
Triggering code completion (ctrl+space) while the cursor is located as indicated gives a list of proposals, so far so good.
Now, after selecting the first proposal (in my case the "print" method), the code gets transformed as follows:
```Java
public class Demo {
		System.out.println("hello");
		if (true) {
			System.out.print(false);
		}
	}
}
```
The expected behaviour would be that the existing parameter "world" remains in place and only the "ln" part of "println" gets removed, i.e. the method call is updated from "println" to "print", like so:
```Java
public class Demo {

		System.out.println("hello");
		if (true) {
			System.out.print("world");
		}
	}
}
```
Note that I have enclosed the second println statement in an if block (the true doesnt matter). An empty block does not show the bad behaviour. Here is an (incomplete) overview:

```Java
public class Test {

  public static void main (String[] args) {
    // good behaviour
    System.out.println("hello");

    { // good behaviour
      System.out.println("world");
    }

    if (true) { // bad behaviour
      System.out.println("world");
    }

    while (true) { // bad behaviour
      System.out.println("world");
      break;
    }

    for (int i = 0; i < 10; i++) { // bad behaviour
      System.out.println("world");
    }

    Runnable a = () -> { // bad behaviour
      System.out.println("world");
    };

    Nested n = new Nested();
    // good behaviour for deeper1, bad behaviour for deeper2
    n.deeper1(1, 2).deeper2(3, 4);
  }

  public static class Nested {
    public Nested deeper;

    public Nested deeper1 (int a, int b) {
      return deeper;
    }

    public Nested deeper2 (int a, int b) {
      return deeper;
    }
  }

}
```

My content assist settings:
settings

I was able to pinpoint the issue to the jdt.core, specifically the following commit: e3517ddc41f3c9536a29ef9be4e7dd3104993ab2

To arrive at this commit I used a git bisect:
- endpoints: 18d780b and 1e898ec
- last good: 1e8d2c330d6c9f355e1893333382d51809c457c3
- first bad: e3517ddc41f3c9536a29ef9be4e7dd3104993ab2

Due to the complexity of that specific commit, me looking into the eclipse source code for the first time, and my lack of time, I was hoping someone more experienced could pick it up from here.

There are some other completion issues I have been encountering which are possibly related to this, so looking into it is hopefully time well spent.

Additionally possibly related:

#1770 
#1016 

Thank you!

angelickite:
## What it does
Add cursor awareness to various auto-completion situations.

Fixes #2620 specifically and probably also [bug 575025](https://bugs.eclipse.org/bugs/show_bug.cgi?id=575025), eclipse-jdt/eclipse.jdt.ui#1804, and probably some others.
## How to test
I've included some tests. A thorough description can be found in #2620.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2023 IBM Corporation and others.
+ * Copyright (c) 2000, 2024 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests.java
@@ -606,7 +606,7 @@ public void testCamelCaseField1() throws JavaModelException {
 }
 // https://bugs.eclipse.org/bugs/show_bug.cgi?id=102572
 public void testCamelCaseLocalVariable1() throws JavaModelException {
-	String old = getSetCodeAssistProperty(JavaCore.CODEASSIST_VISIBILITY_CHECK, JavaCore.ENABLED);
+	String old = getSetCodeAssistProperty(JavaCore.CODEASSIST_CAMEL_CASE_MATCH, JavaCore.ENABLED);
 	try {
 		this.workingCopies = new ICompilationUnit[1];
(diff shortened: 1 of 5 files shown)


jukzi:
A merge is possible but we typically use ammend or even create a new PR to also get the PR title updated.

Jörg Kubitz

PR #3350, opened 2024-11-26
Faulty/Undesired CodeCompletion behaviour #2620

angelickite:
## What it does
Add cursor awareness to various auto-completion situations.

Completions for CompletionOnMessageSendName as well as for QualifiedNameReference did not take the cursor position into consideration, leading to various downstream issues like failing to complete at all, not properly overwriting a source range or overwriting too eagerly.

fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2620

## How to test
I've included some tests. A thorough description can be found in https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2620.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
+++ b/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
@@ -3244,9 +3244,14 @@ public final class CompletionEngine
 		if (!this.requestor.isIgnored(CompletionProposal.METHOD_REF)) {
 			CompletionOnMessageSendName messageSend = (CompletionOnMessageSendName) astNode;
 
-			setTokenRange(messageSend.sourceStart, messageSend.sourceEnd);
-			if (messageSend.statementEnd > messageSend.sourceStart)
-				setSourceRange(messageSend.sourceStart, messageSend.statementEnd);
+			// we take the cursor location into consideration
+			if (messageSend.cursorIsToTheLeftOfTheLParen) {
+				setSourceAndTokenRange(messageSend.sourceStart, messageSend.sourceEnd);
+			} else {
+				setTokenRange(messageSend.sourceStart, messageSend.sourceEnd);
+				if (messageSend.statementEnd > messageSend.sourceStart)
+					setSourceRange(messageSend.sourceStart, messageSend.statementEnd);
+			}
 
 			this.completionToken = messageSend.selector;
 
@@ -3562,7 +3567,41 @@ public final class CompletionEngine
 	private void internalCompletionOnQualifiedReference(ASTNode ref, ASTNode enclosingNode, Binding qualifiedBinding, Scope scope,
 			boolean insideTypeAnnotation, boolean isInsideAnnotationAttribute, InvocationSite site, char[][] tokens, long[] sourcePositions)
 	{
-		long completionPosition = sourcePositions[sourcePositions.length - 1];
+		// we take the cursor location into consideration.
+		//
+		// void x(TreeNode node) {
+		//   TreeNode other = node.ch|.child.child.parent;
+		//   // ...
+		// }
+		//
+		// in the above example, bestPosition will match against the cursor after 'ch', i.e. sourcePositions[1].
+		// (sourcePositions[0] points to 'node', sourcePositions[4] to 'parent')
+		long bestPosition;
+		{
+			bestPosition = -1L;
+			for (int i = 0; i < sourcePositions.length; i++) {
+				long p = sourcePositions[i];
+				int start = (int) (p >>> 32);
+				int end = (int) (p);
+				{ // in specific cases like "node.|.child" the start might be larger than the end (see GH-2620)
+					if (start > end) {
+						int tmp = start;
+						start = end;
+						end = tmp;
+					}
+				}
+				boolean cursorWithinBounds = (start <= this.actualCompletionPosition && this.actualCompletionPosition <= end);
+				if (cursorWithinBounds) {
+					bestPosition = p;
+				}
+			}
+		}
+		if (bestPosition == -1L) {
+			// fallback to the previous behaviour, just in case.
+			bestPosition = sourcePositions[sourcePositions.length - 1];
+		}
+		long completionPosition = bestPosition;
+
 		if (qualifiedBinding.problemId() == ProblemReasons.NotFound) {
 			setSourceAndTokenRange((int) (completionPosition >>> 32), (int) completionPosition);
 			// complete field members with missing fields type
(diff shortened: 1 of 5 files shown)


org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
```diff
@@ -3562,7 +3567,41 @@ private void completionOnQualifiedNameReference(ASTNode astNode, ASTNode enclosi
...
+		//
+		// void x(TreeNode node) {
+		//   TreeNode other = node.ch|.child.child.parent;
+		//   // ...
+		// }
+		//
+		// in the above example, bestPosition will match against the cursor after 'ch', i.e. sourcePositions[1].
+		// (sourcePositions[0] points to 'node', sourcePositions[4] to 'parent')
+		long bestPosition;
+		{
```
jukzi, line 3580: why start a codeblock here?

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
```diff
@@ -3562,7 +3567,41 @@ private void completionOnQualifiedNameReference(ASTNode astNode, ASTNode enclosi
...
+			bestPosition = -1L;
+			for (int i = 0; i < sourcePositions.length; i++) {
```
jukzi, line 3582: You can use a for-each loop here? would be easier to read.

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
```diff
@@ -3562,7 +3567,41 @@ private void completionOnQualifiedNameReference(ASTNode astNode, ASTNode enclosi
...
+				long p = sourcePositions[i];
+				int start = (int) (p >>> 32);
+				int end = (int) (p);
+				{ // in specific cases like "node.|.child" the start might be larger than the end (see GH-2620)
+					if (start > end) {
+						int tmp = start;
```
jukzi, line 3588: if this code should swap start and end, you could add it as comment.

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/complete/CompletionOnMessageSendName.java
```diff
@@ -27,14 +28,19 @@
...
 
 	public boolean nextIsCast;
+	public boolean cursorIsToTheLeftOfTheLParen;
 
-	public CompletionOnMessageSendName(char[] selector, int start, int end, boolean nextIsCast) {
+	public CompletionOnMessageSendName(char[] selector, int start, int end) {
+		this(selector, start, end, setup -> {/* use defaults */});
+	}
+
+	public CompletionOnMessageSendName(char[] selector, int start, int end, Consumer<CompletionOnMessageSendName> setup) {
```
jukzi, line 37: instead of using a Consumer for setup it would be easier to understand and maintain if you directly pass the initial values to the constructor. probably "cursorIsToTheLeftOfTheLParen" misses a "t" at the end and could be final.

stephan-herrmann: In order to unblock the change I'm applying some of these suggestions, but here's one correction: LParen is correct compiler lingo to refer to '(' (short for left parenthesis), nothing to do with parents :)

stephan-herrmann:
Trying to unblock this issue I ...
* rebased the existing commit on HEAD (without any changes)
* added a commit of code simplifications (also  considering @jukzi's comments) and additional tests.

One more observations, and one remaining task:

(1) We should call out that the change in `CompletionTest.testCamelCaseLocalVariable1()` fixed a test bug, which would affect proposals of `Object.clone()` and `Object.finalize()` in unrelated tests, if run after this one.

(2) I'm seeing a regression when trying the following in the IDE:
```
public class GH2620_1 {
	public static void main (String[] args) {
		int j = 1;
		Test local;
		if (true) { // the if-statement is crucial for this test.
			/*x*/local.ref.test|();
		}
	}
	public static class Test {
		public Test ref;
		public static void test(int i) {}
	}
}
```

Cursor location is indicated by `|`.
* Old behavior: argument `j` is proposed.
* New behavior: argument `j` is proposed only when completion operates in 'insert' mode. When selecting 'overwrite' no argument is proposed.

@angelickite  can you find out, if this is a bug in JDT/UI, or can/should be fixed here?

angelickite:
@stephan-herrmann Are we on the same page that this regression is foremost a side effect of the new approach? What I mean by that:

- The new approach is generally well behaved
- Specifically when the completion is operating on any part names (i.e.. `local`, `ref` and `test` in the above example), then the completion should only affect these parts names.
- If the method name is the target, only the method name should be operated on and the argument list not be touched, i.e existing arguments should remain as they are. (e.g. `System.out.print|("Hello")` -> `System.out.println("Hello")`)

So the regression then manifests itself in the case where we have an empty argument list and can potentially safely modify the argument list too as a bonus feature as there is "nothing to break"?

If this is the situation you refer to as the regression, looking back I did see this case appear but never gave it any attention or thought. Reason being that the completion behaviour around arguments always felt janky anyways as well as it not negatively impacting my workflow.

If this feature is truly highly important, I will take a look where it originates and whether there is a sensible solution.

Lastly, I personally would prefer this sidegrade every time over having the old behaviour back.

angelickite:
The regression is here: https://github.com/angelickite/eclipse.jdt.core/blob/aa0c66d368586cdc30c18511f260519c80c59ff7/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java#L3263-L3270

So it stems directly from the new behaviour where "cursor to left of lparen" is given special consideration.

Something like this just might do the trick (limited testing):

```
// we take the cursor location into consideration
if (messageSend.cursorIsToTheLeftOfTheLParen && messageSend.arguments != null) {
  setSourceAndTokenRange(messageSend.sourceStart, messageSend.sourceEnd);
else {
  setTokenRange(messageSend.sourceStart, messageSend.sourceEnd);
  if (messageSend.statementEnd > messageSend.sourceStart)
    setSourceRange(messageSend.sourceStart, messageSend.statementEnd);
}
```

stephan-herrmann:
@angelickite are you interested in taking this to completion?

angelickite:
@stephan-herrmann Should be good to go. The various holes in codecompletion go quite deep... so I've tried to find a balance between "keeping sanity", "recovering lost value", "maintaining the status quo" and "taking a step towards a better future".

angelickite:
@stephan-herrmann Another ping.

Issue #3278, opened 2024-11-10
Annotation-based analysis breaks when annotations are not on the build path


basilevs:
# Test steps
Create a class in an Eclipse plugin:

```
import java.io.Closeable;
import java.io.IOException;
import java.util.ArrayList;
import java.util.List;

import org.eclipse.jdt.annotation.NotOwning;
import org.eclipse.jdt.annotation.Owning;

public class TestNotOwning implements Closeable {
    private final List<Closeable> toClose = new ArrayList<>();

    @NotOwning
    public <T extends Closeable> T register(@Owning
    T closeable) throws IOException {
        closeable.close();
        return closeable;
    }

    @Override
    public void close() throws IOException {
        for (Closeable closeable : toClose.reversed()) {
            closeable.close(); // Ignore error handling for this demonstration
        }
    }

}
```
Enable annotation analysis and ensure that:
```
try (TestNotOwning t = new TestNotOwning()) {
    var a = new Closeable() {} ; // produces warning
    var b = t.register(new Closeable() {} ); // produces no warning
}
```

Create a fragment of the same plugin. Enable annotation analysis. Try same sample:
```
try (TestNotOwning t = new TestNotOwning()) {
    var a = new Closeable() {} ; // produces warning
    var b = t.register(new Closeable() {} ); // produces warning
}
```
## Expected result
no warnings for line `var b = ...` or a warning requesting a missing dependency

Workaround: add an explicit dependency on `org.eclipse.jdt.annotation` in the fragment.

For some reason annotation analysis ignores implicit dependency on `org.eclipse.jdt.annotation` from host plugin.

stephan-herrmann:
I'll try to have a look during the next release cycle.

If there is a bug indeed, it will likely be an issue of PDE, because JDT knows nothing of bundles, fragments and all that. All that JDT works on is a readily resolved classpath.

basilevs:
JDT annotations are absent from classpath (as they are not used in the fragment).

stephan-herrmann:
> JDT annotations are absent from classpath (as they are not used in the fragment).

We'll have to check: shouldn't PDE implicitly put them on the classpath? Do you see them when you expand "Plug-in Dependencies" in the Package Explorer?

basilevs:
Attached is a reproducer for pure JDT. A "client" project has no direct dependency on annotations package, but still produces the potential resource leak warning.
[...]

PR #3355, opened 2024-11-26
Annotation-based analysis breaks when annotations are not on the build path

stephan-herrmann:
+ recognize missing annotation types if relevant for analysis
+ treat Closeable lambda as fresh resource

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3278


stephan-herrmann:
I don't see a way to bring this approach to a complete conclusion.

Instead a different direction is taken by #3374

Issue #2096, opened 2024-03-05
Textual problem indicator goes wild with lamda


laeubi:
I just noticed that in the build logs ([full log can be seen here](https://github.com/laeubi/eclipse.jdt.core/actions/runs/8156485127/job/22294148290)):

```
[INFO] Compiling 81 source files to /home/runner/work/eclipse.jdt.core/eclipse.jdt.core/platform/equinox/bundles/org.eclipse.equinox.http.servlet/target/classes using Eclipse Compiler for Java(TM) 3.38.0.v20240305-1247
Warning:  /home/runner/work/eclipse.jdt.core/eclipse.jdt.core/platform/equinox/bundles/org.eclipse.equinox.http.servlet/src/org/eclipse/equinox/http/servlet/internal/HttpServiceImpl.java:[72] 
	AccessController.doPrivileged((PrivilegedExceptionAction<Void>) () -> {
	^^^^^^^^^^^^^^^^
The type AccessController has been deprecated since version 17 and marked for removal
Warning:  /home/runner/work/eclipse.jdt.core/eclipse.jdt.core/platform/equinox/bundles/org.eclipse.equinox.http.servlet/src/org/eclipse/equinox/http/servlet/internal/HttpServiceImpl.java:[72] 
	AccessController.doPrivileged((PrivilegedExceptionAction<Void>) () -> {
				httpServiceRuntime.registerHttpServiceFilter(bundle, alias, filter, initparams, httpContextHolder);
				return null;
			});
	                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 The method doPrivileged(PrivilegedExceptionAction<Void>) from the type AccessController is deprecated
```

while this might be not so nice already, it even can go further:

```
Warning:  /home/runner/work/eclipse.jdt.core/eclipse.jdt.core/platform/equinox/bundles/org.eclipse.equinox.http.servlet/src/org/eclipse/equinox/http/servlet/internal/servlet/ResourceServlet.java:[74] 
[...]

PR #3358, opened 2024-11-27
Textual problem indicator goes wild with lamda

srikanth-sankaran:
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2096


srikanth-sankaran:
I didn't go searching for otter such needles in the haystack. This being a non correctness issue, let us cross the bridge when we get to it.

Issue #3273, opened 2024-11-09
java.lang.StackOverflowError during "Requesting Java AST from selection"


nlisker:
Eclipse soft crashes during hover over some fields with a SOE and asks to restart. This also breaks the compilation, as after this error, many red lines appear saying that various methods can't be found suddenly.
![image](https://github.com/user-attachments/assets/264f5dbb-98a5-466f-be5d-089429ad053c)
![image](https://github.com/user-attachments/assets/c82af755-eae9-463c-8c62-4beb8b42087b)

The first lines of the stack traces are different every time, but the recursion is always the same. Here is the log with various cases of these errors: [crashes.log](https://github.com/user-attachments/files/17687075/crashes.log)

Here is the infinite recursion (I don't know what the entry point was):
```
	at org.eclipse.jdt.internal.compiler.Compiler.accept(Compiler.java:365)
	at org.eclipse.jdt.core.dom.CompilationUnitResolver.accept(CompilationUnitResolver.java:226)
	at org.eclipse.jdt.core.dom.CompilationUnitResolver.accept(CompilationUnitResolver.java:221)
	at org.eclipse.jdt.internal.compiler.lookup.LookupEnvironment.askForType(LookupEnvironment.java:391)
	at org.eclipse.jdt.internal.compiler.lookup.PackageBinding.getTypeOrPackage(PackageBinding.java:275)
	at org.eclipse.jdt.internal.compiler.lookup.CompilationUnitScope.findImport(CompilationUnitScope.java:643)
	at org.eclipse.jdt.internal.compiler.lookup.CompilationUnitScope.resolveImports(CompilationUnitScope.java:325)
	at org.eclipse.jdt.internal.compiler.lookup.CompilationUnitScope.checkAndSetImports(CompilationUnitScope.java:284)
	at org.eclipse.jdt.internal.compiler.lookup.LookupEnvironment$CompleteTypeBindingsSteps.perform(LookupEnvironment.java:191)
	at org.eclipse.jdt.internal.compiler.lookup.LookupEnvironment.completeTypeBindings(LookupEnvironment.java:602)
	at org.eclipse.jdt.internal.compiler.Compiler.accept(Compiler.java:365)
	at org.eclipse.jdt.core.dom.CompilationUnitResolver.accept(CompilationUnitResolver.java:226)
	at org.eclipse.jdt.core.dom.CompilationUnitResolver.accept(CompilationUnitResolver.java:221)
```

[...]

PR #3373, opened 2024-11-30
java.lang.StackOverflowError during "Requesting Java AST from selection"

stephan-herrmann:
reproduces https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3273

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/Compiler.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/Compiler.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corporation and others.
+ * Copyright (c) 2000, 2024 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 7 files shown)


stephan-herrmann:
In https://bugs.eclipse.org/bugs/show_bug.cgi?id=544306 @brychcy implemented a check that ensures that `NameLookup.getModuleDescription()` answers `null` for a "foreign" package fragment root (i.e., from dependency), if the dependency is on the classpath rather than modulepath.

In the current change I opportunistically restrict that check to modular projects.

This may not be the last word on this matter. Perhaps we should rather hide any `module-info.java` that is brought in via the classpath.

stephan-herrmann:
> In https://bugs.eclipse.org/bugs/show_bug.cgi?id=544306 @brychcy implemented a check that ensures that `NameLookup.getModuleDescription()` answers `null` for a "foreign" package fragment root (i.e., from dependency), if the dependency is on the classpath rather than modulepath.
> 
> In the current change I opportunistically restrict that check to modular projects.
> 
> This may not be the last word on this matter. 

This indeed looked inconsistent.

In [3cadc59](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3373/commits/3cadc596bac64577baa3ddd8cd1057594eac7200) I reversed the strategy: instead of surfacing modules in more situations, I added a callback in SearchableNameEnvironment where the compiler can check if a CU is on the module path or class path. If the latter, than the CU will not appear as associated to any module.

PR #3374, opened 2024-12-02
Annotation-based analysis breaks when annotations are not on the build path

stephan-herrmann:
+ tag elements annotated with a missing type relevant for analysis
+ report the missing annotation at client sites when possibly relevant
+ also: treat Closeable lambda as fresh resource

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3278


jarthana:
@stephan-herrmann  we usually keep the versions of compiler.batch and jdt.core in sync. I see that the former is at 3.41 and the latter is still at 3.40.100. At the very least, the jdt.core should reference the correct version of compiler.batch. Do you mind I update the version of compiler.batch or see a reason to keep it as is?

iloveeclipse:
If batch compiler updated minor version segment, jdt core must follow, because it reexports API from batch compiler.

PR #3375, opened 2024-12-02
Use Import-Package instead of Require-Bundle for o.e.c.runtime to prevent importing unnecessary classes

testforstephen:
## What it does
Based on the discussion in https://github.com/eclipse-equinox/equinox/discussions/697#discussioncomment-11262689, using `Require-Bundle` for `org.eclipse.core.runtime` will pull in all of `org.eclipse.osgi` which in-turn exposes you to all the JDK packages because this horrible re-export:

https://github.com/eclipse-platform/eclipse.platform/blob/6dd67323474b1b49d541bcd1e4ea8007ab2a36a4/runtime/bundles/org.eclipse.core.runtime/META-INF/MANIFEST.MF#L12C17-L12C87.

In our downstream project, we try to override some part of JRE packages within our fragment of `org.eclipse.jdt.core`. However, the default JDK packages always take precedence because of this `Require-Bundle` thing. The equinox team's recommendation is to use `Import-Package` instead of `Require-Bundle` for finer control over the imported classes.

## How to test
This PR just minimizes the imported packages from the bundle `o.e.c.runtime`, without affecting the overall functionality of the feature.

### DIFF
--- a/org.eclipse.jdt.core/META-INF/MANIFEST.MF
+++ b/org.eclipse.jdt.core/META-INF/MANIFEST.MF
@@ -6,6 +6,16 @@ Bundle-Version: 3.40.100.qualifier
 Bundle-Activator: org.eclipse.jdt.core.JavaCore
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
+Import-Package: org.eclipse.core.internal.runtime,
+ org.eclipse.core.runtime,
+ org.eclipse.core.runtime.content,
+ org.eclipse.core.runtime.jobs,
+ org.eclipse.core.runtime.preferences,
+ org.eclipse.equinox.app,
+ org.eclipse.osgi.service.debug,
+ org.eclipse.osgi.util,
+ org.osgi.framework,
+ org.osgi.service.prefs
 Export-Package: org.eclipse.jdt.core,
  org.eclipse.jdt.core.compiler,
  org.eclipse.jdt.core.dom,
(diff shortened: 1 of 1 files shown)


org.eclipse.jdt.core/META-INF/MANIFEST.MF
```diff
@@ -6,6 +6,16 @@ Bundle-Version: 3.40.100.qualifier
 Bundle-Activator: org.eclipse.jdt.core.JavaCore
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
+Import-Package: org.eclipse.core.internal.runtime,
```
jukzi, line 9: why importing an internal package?

szarnekow: Also why are these imported without a version contraint?

testforstephen: > why importing an internal package?

The use of internal packages was not introduced by this PR; it originates from existing code in jdt.core, which already references these internal packages.

testforstephen:
> By arguing this PR affect your project you imply that the require-bundle silently is some public API. APIs should not be changed. I do not understand what exactly is going on there, but i have doubts that this is the right way to fix it.

Both `Import-Package` and `Require-Bundle` are standard mechanisms for referencing packages from other bundles. And following https://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#i3198951, `Import-Package` is even the preferred way for wiring bundles. Please clarify what issue do you see with this PR change.

org.eclipse.jdt.core/META-INF/MANIFEST.MF
mickaelistria, line 9: Based on discussion at https://github.com/eclipse-equinox/equinox/discussions/697 , it looks like we can add `;bundle-symbolic-name="org.eclipse.core.runtime";bundle-version="[3.29.0, 4.0)"` to each Import-Package to keep some versioning.

mickaelistria: > Also why are these imported without a version constraint?

A version constraint was added in last revision of this PR.

Issue #3381, opened 2024-12-03
[Enhanced Switch][Null] Missing Null pointer access warning with total/unconditional patterns


srikanth-sankaran:
Found by code inspection and white box testing:

While ECJ rejects this program with an error (`Null pointer access: This expression of type Integer is null but requires auto-unboxing`)

```
import org.eclipse.jdt.annotation.Nullable;

public class X {
	public static void main(String[] args) {
		@Nullable Integer i = null;
		switch (i) {
		// case Integer ii -> System.out.println();
		default -> System.out.println();
		}
	}
} 
``` 

this one below is compiled without any diagnostic, but crashes at runtime with an NPE:

```
import org.eclipse.jdt.annotation.Nullable;

public class X {
	public static void main(String[] args) {
		@Nullable Integer i = null;
		switch (i) {
			case Integer ii -> System.out.println();
		}
	}
} 
```

srikanth-sankaran:
Trouble stems from `org.eclipse.jdt.internal.compiler.ast.SwitchStatement.isNullHostile()`'s last line that reads: `return this.totalPattern == null;` 

Why should we check for whether any case arm contains an unconditional pattern is a mystery.

See that while this line stands in this form due to a recent refactoring, problem goes back to 4.30 at least.

Issue #3319, opened 2024-11-16
[Enhanced Switch][Null] Inconsistent null propagation


srikanth-sankaran:
Found by code inspection:

With null analysis enabled, we report two warnings in foo(Object)  `Null comparison always yields false: The variable o cannot be null at this location` but none in foo(X) (with redundant null check set to warning)

```
import org.eclipse.jdt.annotation.NonNull;

public class X {
	static @NonNull Object foo(Object o) {
		switch (o) {
			case String s -> {
				if (o == null) {
					System.out.println("o cannot be null at all!");
				}
				System.out.println();
			}
			default -> {
				if (o == null) {
					System.out.println("o cannot be null at all!");
				}
				System.out.println();	
			}
		}
		return new Object();
	}
	
	static @NonNull Object foo(X o) {
		switch (o) {
			case X s  -> {
				if (o == null) {
					System.out.println("o cannot be null at all!");
				}
				System.out.println(s);
			}
		}
		return new Object();
	}
}
```

This is because of the incorrect asymmetry in `org.eclipse.jdt.internal.compiler.ast.SwitchStatement.analyseCode(BlockScope, FlowContext, FlowInfo)` between a regular case and a default.

srikanth-sankaran:
> This is because of the incorrect asymmetry in org.eclipse.jdt.internal.compiler.ast.SwitchStatement.analyseCode(BlockScope, FlowContext, FlowInfo) between a regular case and a default.

Not entirely sure of this. Need to study in detail. 

At the outset it does look like there is some inconsistent behavior. Cause may be different than hypothesized.

srikanth-sankaran:
The first commit from https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3382 also this issue - basically the isNullHostile method's wrong result in the presence of total patterns was inhibiting a diagnostic from being reported.

PR #3382, opened 2024-12-03
[Enhanced Switch][Null] Missing Null pointer access warning with total/unconditional patterns in case labels

srikanth-sankaran:
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3381

### DIFF
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/NullAnnotationTests21.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/NullAnnotationTests21.java
@@ -107,7 +107,7 @@ public class NullAnnotationTests21 extends AbstractNullAnnotationTest {
 				  "				case Number n0 -> consumeNumber(n0);\n" +
 				  "			}\n" +
 				  "		} catch (NullPointerException npe) {\n" +
-				  "			// ignoring the unchecked warning, and expecting the NPE:\n" +
+				  "			// Expecting an NPE because selector is null\n" +
 				  "			System.out.print(npe.getMessage());\n" +
 				  "		}\n" +
 				  "	}\n" +
@@ -115,6 +115,7 @@ public class NullAnnotationTests21 extends AbstractNullAnnotationTest {
 				  "		System.out.print(i);\n" +
 				  "	}\n" +
 				  "	void consumeNumber(@NonNull Number n) {\n" +
+				  "     System.out.println(\"consumeNumber \");\n" +
 				  "		System.out.print(n.toString());\n" +
 				  "	}\n" +
 				  "	public static void main(String... args) {\n" +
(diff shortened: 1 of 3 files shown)


srikanth-sankaran:
Expecting some tests to fail

org.eclipse.jdt.core.tests.compiler.regression.NullAnnotationTests21.test_totalTypePatternDoesNotAdmitNull()
org.eclipse.jdt.core.tests.compiler.regression.NullAnnotationTests21.test_totalTypePatternNullableExpression()

but these need to be studied to see if expectation is valid

srikanth-sankaran:
Looking at `org.eclipse.jdt.core.tests.compiler.regression.NullAnnotationTests21.test_totalTypePatternNullableExpression()` it seems the understanding is that an unconditional pattern would match null - but that is not so.

So when `case Number n0` is matched. n0 can never be null, so the warning `Null type mismatch: required \'@NonNull Number\' but the provided value is inferred as @Nullable` expected by the test looks wrong.

@stephan-herrmann - The proposed patch is still WIP, comments are welcome

srikanth-sankaran:
I think I have remastered the "failing" tests with right expectations. @stephan-herrmann thanks for your review.

stephan-herrmann:
> So when `case Number n0` is matched. n0 can never be null

looks like this changed between Java 18 and 19.

> I think I have remastered the "failing" tests with right expectations. @stephan-herrmann thanks for your review.

looking into it now.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/NullAnnotationTests21.java
```diff
@@ -1244,4 +1238,46 @@ static Stuff match(PatternMatching pm, int v) {
...
+				1. ERROR in X.java (at line 6)
+					switch (i) {
+					        ^
+				Null pointer access: The variable i can only be null at this location
+				----------
+				2. ERROR in X.java (at line 12)
+					switch (i) {
+					        ^
+				Null pointer access: This expression of type Integer is null but requires auto-unboxing
+				----------
```
stephan-herrmann, line 1279: @srikanth-sankaran  do you have an idea why these are different error messages?

stephan-herrmann:
@srikanth-sankaran  your turn again :)

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/NullAnnotationTests21.java
srikanth-sankaran: > @srikanth-sankaran do you have an idea why these are different error messages?

Yes, I had looked into this. In the former case its is a type switch which doesn't call for unboxing and in the latter switch on the unboxed integer. This results in a special message in the latter case.

srikanth-sankaran:
Your additional changes look good, Thanks @stephan-herrmann!

srikanth-sankaran:
The first commit from here also fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3319 - basically the isNullHostile method's wrong result in the presence of total patterns was inhibiting a diagnostic from being reported.

Issue #3327, opened 2024-11-19
[Performance] Full build takes more time since 2024-09 (type inference)


iloveeclipse:
### Discussed in https://github.com/eclipse-jdt/eclipse.jdt.core/discussions/3305

<div type='discussions-op-text'>

<sup>Originally posted by **fedejeanne** November 13, 2024</sup>
Hi,
I've been looking into some performance issues that started with **2024-09**, specifically the time it takes to perform a **full build**. Interestingly enough, the time it takes to do a _clean all_ also increased in 2024-09 but the biggest increase happened in the build phase.

- In **2024-06** it takes ~12 minutes to do a full clean and build.
- In **2024-09** it takes ~30 minutes to do a full build (sampled **after** the _clean all_)

[samples.zip](https://github.com/user-attachments/files/17728799/samples.zip)

The part that caught my eye was the increase in amount of calls (_Hits_) and therefore the _Self Time (CPU)_ for the method(s) `org.eclipse.jdt.internal.compiler.lookup.TypeBinding.findSuperTypeOriginatingFrom(...)` (there are 2 flavors of this method).

**2024-06**
- 431 hits
- ~20s spent in these methods
![image](https://github.com/user-attachments/assets/2b876849-c0d4-4546-82b4-86625cacaf11)

**2024-09**
- 6972 hits
- ~9m spent in these methods
![image](https://github.com/user-attachments/assets/b809a235-e0da-44d0-bf7e-26a816497bfd)

## Question 1
- Why are there suddenly so many more calls to these methods ?

## Question 2
- Would it be OK to simply build a cache to speed up some subsequent repeated calls or would this introduce any undesired effects (_e.g._ inconsistencies)?

I ask because I noticed that many of the calls are "repeated" _i.e._ the input parameters are the same so a cache would be a trivial optimization. </div>

jukzi:
my bet it's caused by changes in org.eclipse.jdt.internal.compiler.lookup.BoundSet.allSuperPairsWithCommonGenericType ():
![image](https://github.com/user-attachments/assets/e520199d-ec4a-49c2-8e54-5398d3a3cd22)
@fedejeanne can you test if a revert of https://github.com/eclipse-jdt/eclipse.jdt.core/commit/cc0e66ea57dbcef4054c360d84d2a0ecf8e9b5d4 helps?

jukzi:
@fedejeanne it may help if you share a script that generates an artificial project with such a complex type hierarchy

[...]

PR #3333, opened 2024-11-21
Try with a simpler search in BoundSet::incorporate(...) before falling back to a more thorough (and expensive) search

fedejeanne:
## What it does

This PR brings back the simpler (faster) approach that was used in the method `org.eclipse.jdt.internal.compiler.lookup.BoundSet.incorporate(InferenceContext18, TypeBound[], TypeBound[])` and only falls back to the newer (slower) approach if the simple one failed.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3327

## Implementation details
This PR brings back the old approach by re-adding these methods:
- `deriveTypeArgumentConstraints` (renamed to `deriveTypeArgumentConstraintsSimple`)
- `superTypesWithCommonGenericType`

### Workspace and performance details
This has a drastic performance improvement when doing a full **clean and build** for our tool.

Some details about our workspace:
- Projects: **> 2000** (plugins, features and fragments) .
- Full build time up until version _2024-06_: **~10 minutes**.
- Full build time starting at _2024-09_: **~30 minutes**.
- Full build time with this PR: **~12 minutes** ([Sample.zip](https://github.com/user-attachments/files/17847208/Sample.zip))

### Option 1: regression tests
- Run these regression tests in `org.eclipse.jdt.core.tests.compiler.regression.GenericsRegressionTest_1_8`:
   - `testGH2413`
   - `testGH2413_mini`
   - `testGH2413_direct`

### Option 2: Snippet
The following snippet should compile

```java
public class EjcBug2413 {

	public interface EventSource<L> {
	}

	public interface ObservableEventListener<V> {
	}

	public interface WritableProperty<V> extends EventSource<ObservableEventListener<V>> {
		static <P extends WritableProperty<?>> P getReadOnly(P property) {
			return null;
		}
	}

	public static class Property<V> implements EventSource<ObservableEventListener<V>>, WritableProperty<V> {
	}

	public static abstract class Bug2413<V, P extends Property<V>> {
		public void foo(P property) {
			P readOnly = WritableProperty.getReadOnly(property);
		}
	}
}
```

## Author checklist

- [x] I have thoroughly tested my changes
- [x] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [x] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

PR #3384, opened 2024-12-03
[Performance] Full build takes more time since 2024-09 (type inference)

stephan-herrmann:
optimize 1:
+ don't record pairs without any type arguments

optimize 2:
+ stop traversal when both types are the same

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3327

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/BoundSet.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/BoundSet.java
@@ -1267,11 +1267,14 @@ class BoundSet {
 		if (s == null || s.id == TypeIds.T_JavaLangObject || t == null || t.id == TypeIds.T_JavaLangObject)
 			return Collections.emptyList();
 		List<Pair<TypeBinding>> result = new ArrayList<>();
-		if (TypeBinding.equalsEquals(s.original(), t.original())) {
+		if ((s.isParameterizedType() || t.isParameterizedType()) // optimization #1: clients of this method only want to inspect type arguments
+				&& TypeBinding.equalsEquals(s.original(), t.original())) {
 			result.add(new Pair<>(s, t));
 		}
+		if (TypeBinding.equalsEquals(s,  t))
+			return result; // optimization #2: nothing interesting above equal types
 		TypeBinding tSuper = t.findSuperTypeOriginatingFrom(s);
-		if (tSuper != null) {
+		if (tSuper != null && s.isParameterizedType() && tSuper.isParameterizedType()) { // optimization #1 again
 			result.add(new Pair<>(s, tSuper));
 		}
 		result.addAll(allSuperPairsWithCommonGenericType(s.superclass(), t));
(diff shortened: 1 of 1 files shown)


stephan-herrmann:
I repeated the measurements  from https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3333#issuecomment-2514729478 after applying both optimizations:

ecj

| | min | max |
|-------|---------|------|
|real   |  0m4,210s  | 0m6,360s |
| user  |  0m8,628s | 0m9,131s |
| sys |    0m0,316s | 0m0,393s |

----

for reference the original results

ecj

| | min | max |
|-------|---------|------|
|real   |  0m8,335s  | 0m8,799s|
| user  |  0m13,848s | 0m14,265s|
| sys |     0m0,427s |  0m0,489s|



javac

| | min | max |
|-------|---------|------|
| real  |  0m14,618s | 0m16,737s|
| user  |  0m19,790s | 0m20,369s|
| sys  |    0m0,293s  | 0m0,436s|

I'm curious how this micro benchmark will translate into practical experience ...

jukzi:
It's an improvement, but does not fix the O(n^2) behavior.
For example try https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3327#issuecomment-2515120465 with I40. You better save your workspace before-

stephan-herrmann:
> It's an improvement, but does not fix the O(n^2) behavior.

JLS requires to inspect all pairs. Can you do that in linear time? I can't.

stephan-herrmann:
> It's an improvement, but does not fix the O(n^2) behavior. For example try [#3327 (comment)](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3327#issuecomment-2515120465) with I40. You better save your workspace before-

See #3387

fedejeanne:
> @fedejeanne the first optimization is effective if many non-generic types are involved. Does your original code contain lots of generics? If so, this particular optimization may not help a lot.

As a rule of thumb, our code contains a lot of everything (it's just a huge code base 😅 ). I will try this optimization and see how much performance was gained.

fedejeanne:
@stephan-herrmann I just compiled the whole workspace with yesterday's I-BUILD (**v20241203-1917**) and it went down from ~30 min to **~17 min**! 👏 

Here's a sample (sampled every **20ms**): 
[v20241203-1907-Full build (clean workspace).zip](https://github.com/user-attachments/files/18004671/v20241203-1907-Full.build.clean.workspace.zip)


And here one can see that even though `allSuperPairsWithCommonGenericType` is called as often as before, it is faster and that `findSuperTypesOriginatingFrom` is called more often (???) but it takes less time. I would take these assumptions with a pinch of salt though, since I am comparing with the original screenshots in https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3327#issue-2671954211 but I don't remember if back then I sampled every 100ms or every 20ms.

image

jukzi:
@fedejeanne The "Hits" column does not relate to call counting. it is only the number of samples taken that includes the method.
You may now want to measure with https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3387 on top, which should delete the method almost entirely from the samples.

fedejeanne:
> You may now want to measure with #3387 on top, which should delete the method almost entirely from the samples.

Just for future reference, I tried it and you were right: the methods went all the way down in the _hotspots_ --> https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3387#issuecomment-2516609355

PR #3386, opened 2024-12-03
Improve performance in CharOperation

robstryker:
When no matching characters / replacements are found in the string, the current replace method seems to be performing at an n^2 rate. The performance is actually really really bad. 

Specifically, if the indexOf(etc) call returns -1, it clearly means there's no more work left to do here, but instead the existing logic continues to loop, and the indexOf(etc) call itself has a loop. 

It seems not only needlessly complicated, but also inefficient. 

This code seems to have originated in org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/pdom/util/CharArrayUtils.java, a file that was later renamed to org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/nd/util/CharArrayUtils.java.  

The replace(etc) method in CharArrayUtils differs from that in CharOperation.  The performance of the version in CharArrayUtils appears normal in comparison to that in CharOperation. 

However, I can't tell how long ago this code was written, but, it would appear to me that simply using String functions performs comparably, so there's no reason whatsoever for all this complicated code? 

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/core/compiler/CharOperation.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/core/compiler/CharOperation.java
@@ -3750,7 +3750,7 @@ public static final char[] replace(
 		next : for (int i = 0; i < max;) {
 			int index = indexOf(toBeReplaced, array, true, i);
 			if (index == -1) {
-				i++;
+				i = max; // end
 				continue next;
 			}
 			if (occurrenceCount == starts.length) {
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/CharOperationTest.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/CharOperationTest.java
@@ -170,4 +170,26 @@ public void test012() {
 			4,
 			true));
 }
+
+public void testReplacePerformance001() {
+	// This is the test that takes an excessively long time
+	// The improvement here is drastic, 99% reduction in time
+	testReplaceImpl("This is one line with no matches\n", 9, 0.01);
+}
+
+private void testReplaceImpl(String oneLine, int power, double multiplier) {
+	String total = oneLine;
+	for( int i = 0; i < power; i++ ) {
+		total = total + total; // Double the length
+	}
+	total = oneLine + total;
+
+	// Now replace
+	long start = System.currentTimeMillis();
+	char[] found = CharOperation.replace(total.toCharArray(), new char[]{'/'}, new char[] {'z'});
+	assertNotNull(found);
+	long end = System.currentTimeMillis();
+	long spent = end - start;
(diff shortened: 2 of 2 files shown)


org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/CharOperationTest.java
```diff
@@ -170,4 +170,25 @@ public void test012() {
...
+	String total = oneLine;
+	for( int i = 0; i < power; i++ ) {
+		total = total + total; // Double the length
+	}
+	total = oneLine + total;
+
+	// Now replace
+	long start = System.currentTimeMillis();
+	char[] found = CharOperation.replace(total.toCharArray(), new char[]{'/'}, new char[] {'z'});
+	long end = System.currentTimeMillis();
```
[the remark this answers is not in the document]
jukzi: Infact the hotspot is allowed to not even execute code that is not needed, so the result should be stored in a volatile sink   to be guaranteed to be executed.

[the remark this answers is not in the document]
robstryker: I've added the assert, on the assumption that that should be sufficient to get the code to execute. Let me know if I'm wrong.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/core/compiler/CharOperation.java
```diff
@@ -3734,59 +3734,15 @@ public static final void replace(char[] array, char[] toBeReplaced, char replace
...
-
-	int[] starts = new int[5];
-	int occurrenceCount = 0;
-
-	if (!equals(toBeReplaced, replacementChars)) {
-
-		next : for (int i = 0; i < max;) {
-			int index = indexOf(toBeReplaced, array, true, i);
-			if (index == -1) {
-				i++;
```
szarnekow, line 3753: The actual bug is here: If index == -1, there is not hit. No need to scan again from the next offset.

What I like about this impl here: it's garbage-free whereas the proposed solution needs to encode the char arrays as ISO or UTF-8.

Do you have time to measure the pressure on the GC if your version of replace is used in tight loops?

jukzi: agree, that sounds like a better fix. One should compare the measurement for both solutions.

robstryker: If you guys are happier with a smaller change, then I'm fine with it.  I'll just abort the loop at that point and be done with it.

jukzi: using "break" instead of "continue" and removing the "next" label would be prettier, but i am OK with this fix

PR #3387, opened 2024-12-03
[Performance] Full build takes more time since 2024-09 (type inference)

stephan-herrmann:
optimize 3:
+ never visit the same super type more than once

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3327

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/BoundSet.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/BoundSet.java
@@ -1264,7 +1264,13 @@ class BoundSet {
 	}
 
 	protected List<Pair<TypeBinding>> allSuperPairsWithCommonGenericType(TypeBinding s, TypeBinding t) {
+		return allSuperPairsWithCommonGenericTypeRecursive(s, t, new HashSet<>());
+	}
+
+	private List<Pair<TypeBinding>> allSuperPairsWithCommonGenericTypeRecursive(TypeBinding s, TypeBinding t, HashSet<TypeBinding> visited) {
 		if (s == null || s.id == TypeIds.T_JavaLangObject || t == null || t.id == TypeIds.T_JavaLangObject)
+			return Collections.emptyList();
+		if (!visited.add(s.prototype()))
 			return Collections.emptyList();
 		List<Pair<TypeBinding>> result = new ArrayList<>();
 		if (s.isParameterizedType() && t.isParameterizedType() // optimization #1: clients of this method only want to compare type arguments
@@ -1277,11 +1283,11 @@ class BoundSet {
 		if (tSuper != null && s.isParameterizedType() && tSuper.isParameterizedType()) { // optimization #1 again
 			result.add(new Pair<>(s, tSuper));
 		}
-		result.addAll(allSuperPairsWithCommonGenericType(s.superclass(), t));
+		result.addAll(allSuperPairsWithCommonGenericTypeRecursive(s.superclass(), t, visited));
 		ReferenceBinding[] superInterfaces = s.superInterfaces();
 		if (superInterfaces != null) {
 			for (ReferenceBinding superInterface : superInterfaces) {
-				result.addAll(allSuperPairsWithCommonGenericType(superInterface, t));
+				result.addAll(allSuperPairsWithCommonGenericTypeRecursive(superInterface, t, visited));
 			}
 		}
 		return result;


stephan-herrmann:
For a moment I was tricked by the [discussion](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3333#issuecomment-2515142140) about `O(n^2)` vs `O(n)` without saying what is `n` :)

The requirement put forward by JLS (and thus not subject to our choosing) is similar to `O(n^2)`, by saying that all pairs of supertypes of types `s` and `t` must be inspected. The implementation actually cuts this down to two linear-ish searches. See that `s` never changes in all recursive invocations. This part of the equation is always handled using `findSuperTypeOriginatingFrom()` which performs a _limited_ search in that hierarchy (limited in that it stops when it find a "matching" type).

What still looks like some combinatorial explosion is in fact a difference between the number of paths vs nodes in the inheritance hierarchy. Viz, the original implementation traversed all paths from `t` up to `j.l.Object` whereas visiting each node in that graph is fully sufficient. This is less an issue of the depth of a hierarchy but more of its "volume", it must be deep and wide with lots of joining edges in the graph for effects to show.

Limiting that traversal is what the cache achieves. In contrast to https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3333#issuecomment-2514836251 no cache of pairs is needed, since `s` is constant during a traversal :)

Additionally, using `prototype()` I ensure that differently annotated variants of the same type are still recognized as same.

I'm still a bit uneasy about using `TypeBinding` in a hashed collection, but looking at the comment in `ReferenceBinding.hashCode()` this could still be safe as long as no `UnresolvedReferenceBinding` are in the lot (because those intentionally break uniqueness of hashCode()).

stephan-herrmann:
FWIW, using the [reproducer](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3333#issuecomment-2514555217) my poor man's measurements yield:

| | min | max |
|-------|---------|------|
|real   |  0m4,050s | 0m4,656s|
|user | 0m8,669s | 0m10,040s |
|sys | 0m0,355s | 0m0,423s |

which is not such a significant gain over results from #3384:

| | min | max |
|-------|---------|------|
|real   |  0m4,210s  | 0m6,360s |
| user  |  0m8,628s | 0m9,131s |
| sys |    0m0,316s | 0m0,393s |

stephan-herrmann:
@fedejeanne I marked this PR as draft, because I'd like to get feedback on results from #3384 in real life, before stacking more speculative optimization on top of the previous one. I mean we've seen the differences in counts, but not so much in compile time.

stephan-herrmann:
> which is not such a significant gain over results from #3384:

The reason is: while we still visit a significant number of types, we don't perform any actual work. The list of pairs already has length zero thanks to #3384 :)

fedejeanne:
@stephan-herrmann understood, I was about to check the performance gain anyway when I read your comments :-)

I am setting up my IDE right now and I will measure performance with #3384 right away.

jukzi:
> I'm still a bit uneasy about using `TypeBinding` in a hashed collection, but looking at the comment in `ReferenceBinding.hashCode()` this could still be safe as long as no `UnresolvedReferenceBinding` are in the lot (because those intentionally break uniqueness of hashCode()).

1. This PR uses hashCode(), equals() only during the search. The cache is not kept on heap. The search itself should not (but may? - that would be bad) modify the hashcode during the search.
2. The hashcode implementation looks horribly bad: equals() and hashCode() should always be in sync. I wonder where this hashcode is already used and if it would not be better fixed.
3. As workaround this PR could use ReferenceBindingSetWrapper that already tried to workaround that problem.
4. The Implementation of org.eclipse.jdt.internal.compiler.lookup.ReferenceBinding.identityHashCode() is bad, as it relies on the fact that super does not override hashCode. ReferenceBindingSetWrapper should use System.identityHashCode instead.

fedejeanne:
FTR, even in its current state and with the bad implementations of `hashCode()` / `equals()`, I rebased this PR on `master` (so #3384 is also applied) and run a full build: it took **~12min** 👏 

I just happened to have VisualVM still open so I am adding also a screenshot of the _Monitor_ view. FYI the build took time between ~9:23 to ~9:35 (sorry that I can not be more precise, I wasn't looking at the watch).

image

This is the sample (sampled every **20ms**): [v20241203-1907 and 3387-Full build (clean workspace).zip](https://github.com/user-attachments/files/18005233/v20241203-1907.and.3387-Full.build.clean.workspace.zip)

And one can see that the hotspots changed:

image

`allSuperPairsWithCommonGenericType` and `allSuperPairsWithCommonGenericTypeRecursive` are way down in the hotspots view:
image

image

jukzi:
Good, that's what i expected.

BTW: you should not trust the Heap/Memory view while you are sampling. Sampling does too much memory overhead.
You may also want to investigate in new Issues why you have hotspots in indexOfChildren() and findSuperTypeOriginatingFrom() [the later may be similar problem as this]. Never ever should you have any CPU hotspot during compile expect File IO, as compiling is not really a computational intensive problem but should be mostly lookups.

fedejeanne:
> BTW: you should not trust the Heap/Memory view while you are sampling. Sampling does too much memory overhead. 

I wasn't aware of that, thank you for the tip!

> You may also want to investigate in new Issues why you have hotspots in `indexOfChildren()` and `findSuperTypeOriginatingFrom()` [the later may be similar problem as this]. Never ever should you have any CPU hotspot during compile expect File IO, as compiling is not really a computational intensive problem but should be mostly lookups.

Hm, I didn't know that either. Do you have any hints about what should I look for when debugging these methods? I honestly assumed that these hotspots were "normal" since they were already there in the sampling I did with _the good old fast **2024-06**_ in the original issue (https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3327#issue-2671954211)

385661685-2b876849-c0d4-4546-82b4-86625cacaf11

Also, this PR brings down the number of invocations on those 2 methods almost to those of **2024-06** so I don't know how much more there is to improve in there. I can certainly take a look at them after merging this PR though (since this PR improves both of them, I mean). I am always happy to help track and improve performance bottlenecks :-)

jukzi:
> Do you have any hints about what should I look for when debugging these methods?

Same procedure: first step is to gather data why it is slow. Typically a breakpoint where it takes longest time (i.e. most elements in the loop) or statistic of the arguments used. second: create a reproducer. At best you could even write a junit test that triggers the  ill behavior. At that point others can easily jump in to find a solution.
The potential gain is a reduction of up to another 1.5 minutes (total time cpu):
image

stephan-herrmann:
> > I'm still a bit uneasy about using `TypeBinding` in a hashed collection, but looking at the comment in `ReferenceBinding.hashCode()` this could still be safe as long as no `UnresolvedReferenceBinding` are in the lot (because those intentionally break uniqueness of hashCode()).
> 
>  1. This PR uses hashCode(), equals() only during the search. The cache is not kept on heap. The search itself should not (but may? - that would be bad) modify the hashcode during the search.

no modification in sight
 
> 2. The hashcode implementation looks horribly bad: equals() and hashCode() should always be in sync. I wonder where this hashcode is already used and if it would not be better fixed.

Please let's be extremely careful about "fixing". The current implementation is necessary for semantic correctness and has served this purpose without any reported issues for 20 years.

In fact, scanning the compiler code, there are several locations using TypeBinding as keys in a HashMap. Again with no reported anomalies.

Given those precedents, I'll stop worrying at this particular location. If desired, the general topic can be picked up in a follow-up ticket.

stephan-herrmann:
> You may also want to investigate in new Issues why you have hotspots in indexOfChildren() and findSuperTypeOriginatingFrom() [the later may be similar problem as this].

`findSuperTypeOriginatingFrom()` is part of the algorithm being optimized here, so no surprise if you ask me.

> Never ever should you have any CPU hotspot during compile expect File IO, as compiling is not really a computational intensive problem but should be mostly lookups.

I'm not sure I'm following the reasoning here. Perhaps there's an unintended "not" involved? Or should it read except instead of expect? As you see I'm clueless :)

stephan-herrmann:
> FTR, even in its current state and with the bad implementations of `hashCode()` / `equals()`, I rebased this PR on `master` (so #3384 is also applied) and run a full build: it took **~12min** 👏

So, in the real world case there seems to be something that is not captured in the reproducer. With this gain I agree that there is value in the 3rd level optimization and will merge it, so it will get thorough field testing before release.

fedejeanne:
> > FTR, even in its current state and with the bad implementations of `hashCode()` / `equals()`, I rebased this PR on `master` (so #3384 is also applied) and run a full build: it took **~12min** 👏
> 
> So, in the real world case there seems to be something that is not captured in the reproducer. With this gain I agree that there is value in the 3rd level optimization and will merge it, so it will get thorough field testing before release.

Great, thank you! 

I'm looking forward to tomorrow's I-BUILD and I will use it ASAP.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/BoundSet.java
```diff
@@ -1277,11 +1283,11 @@ protected List<Pair<TypeBinding>> allSuperPairsWithCommonGenericType(TypeBinding
 		if (tSuper != null && s.isParameterizedType() && tSuper.isParameterizedType()) { // optimization #1 again
```
szarnekow, line 1283: Assuming `s.isParameterizedType` is a very cheap operation and `findSuperTypeOriginatingFrom` is not:
The lookup of tSuper can be guarded by s.isParameterizedType():
```java
TypeBinding tSuper = s.isParameterizedType() ? t.findSuperTypeOriginatingFrom(s) : null;
if (tSuper != null && tSuper.isParameterizedType()) {
  result.add(new Pair<>(s, tSuper));
}
```

fedejeanne: I can try this tomorrow and report back. Thank you for the hint!

fedejeanne: Incredible... I had to sample twice because I couldn't believe my eyes 😄 

Thank you, @szarnekow ! --> https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3411

srikanth-sankaran:
> > 3\. The hashcode implementation looks horribly bad: equals() and hashCode() should always be in sync. I wonder where this hashcode is already used and if it would not be better fixed.
> 
> Please let's be extremely careful about "fixing". The current implementation is necessary for semantic correctness and has served this purpose without any reported issues for 20 years.
> 
> In fact, scanning the compiler code, there are several locations using TypeBinding as keys in a HashMap. Again with no reported anomalies.
> 
> Given those precedents, I'll stop worrying at this particular location. If desired, the general topic can be picked up in a follow-up ticket.

"equals() and hashCode() should always be in sync" - this really means when equals() is overridden, it is extremely likely a defect to not override hashCode() but not necessarily the other way around - no ? Particularly when equals() is coming from j.l.O ??

jukzi:
> "equals() and hashCode() should always be in sync" - this really means when equals() is overridden, it is extremely likely a defect to not override hashCode() but not necessarily the other way around - no ? Particularly when equals() is coming from j.l.O ??

If equals() is overridden but hashCode() is not then it most likely violates the contract.
If hashCode() is overridden but equals() is not then its unlikely to violate the contract (even `return 0` works) but it's needless and most likely introduces hash collisions which will introduce poor performance.  Any custom hashCode can even be exploited for denial of service attack. See https://openjdk.org/jeps/180 and for example https://www.javaspecialists.eu/archive/Issue262-How-Java-Maps-Protect-Themselves-from-DOS-Attacks.html for actual collisions
Lets think of types 
```
class ARbyguv {}
class ARbygvW {}
class ARbyhVv {}
...
// or whatever names make your custom hashcode collide, will force performance to explode
```
Which is impossible when using runtime depended hashcode from j.i.O.

fedejeanne:
FWIW this is how it looks like in our big WS (Sample from #3411)

image

---

image

---

JDT is in green.

For what I see in these pictures (I haven't looked into the code in depth) `hashCode()` seems to be "good enough" since `equals()` is not called that often (_i.e._ there are not that many collisions).

If you guys want me to test something in our big WS, just let me know and I'll fire up VisualVM with any configuration you want :-)

jukzi:
@fedejeanne i'd say you waste 17s  in org.eclipse.jdt.internal.compiler.lookup.InferenceVariable.hashCode()
image
can you post the stacktrace for that?

srikanth-sankaran:
> i'd say you waste 11s in org.eclipse.jdt.internal.compiler.lookup.InferenceVariable.hashCode()

You meant 11 ms - right ?

jukzi:
> You meant 11 ms - right ?

yes, sorry.

srikanth-sankaran:
> Lets think of types
> 
> ```
> class ARbyguv {}
> class ARbygvW {}
> class ARbyhVv {}
> ...
> // or whatever names make your custom hashcode collide, will force performance to explode
> ```
> 
> Which is impossible when using runtime depended hashcode from j.i.O.

I would urge that we base any decisions on real data as opposed to intuition such as "will force performance to explode" - Even super talented engineers can be completely off when it comes to speculating about performance bottlenecks! :) This particular hashCode implementation dates back to 2004 and we haven't had complaints

fedejeanne:
> > You meant 11 ms - right ?
> 
> yes, sorry.

No no, that's actually **11 s**. It's a "digit grouping symbol" and not a "Decimal symbol"

image

... I told you, it's a big WS 😅 

> can you post the stacktrace for that?

With this debug output...

image

... I get always this stack trace for the initial 10% of the compilation:

```java
java.lang.RuntimeException: InferenceVariable.InferenceVarKey.hashCode() = <some_hashcode>
        at org.eclipse.jdt.internal.compiler.lookup.InferenceVariable$InferenceVarKey.hashCode(InferenceVariable.java:42)
        at java.base/java.util.HashMap.hash(HashMap.java:338)
        at java.base/java.util.HashMap.put(HashMap.java:618)
        at org.eclipse.jdt.internal.compiler.lookup.InferenceVariable.get(InferenceVariable.java:80)
        at org.eclipse.jdt.internal.compiler.lookup.InferenceContext18.addInitialTypeVariableSubstitutions(InferenceContext18.java:307)
        at org.eclipse.jdt.internal.compiler.lookup.InferenceContext18.createInitialBoundSet(InferenceContext18.java:224)
        at org.eclipse.jdt.internal.compiler.lookup.ConstraintExpressionFormula.inferInvocationApplicability(ConstraintExpressionFormula.java:387)
        at org.eclipse.jdt.internal.compiler.lookup.InferenceContext18.inferInvocationApplicability(InferenceContext18.java:369)
        at org.eclipse.jdt.internal.compiler.lookup.ParameterizedGenericMethodBinding.computeCompatibleMethod18(ParameterizedGenericMethodBinding.java:252)
        at org.eclipse.jdt.internal.compiler.lookup.ParameterizedGenericMethodBinding.computeCompatibleMethod(ParameterizedGenericMethodBinding.java:92)
        at org.eclipse.jdt.internal.compiler.lookup.Scope.computeCompatibleMethod(Scope.java:849)
        at org.eclipse.jdt.internal.compiler.lookup.Scope.computeCompatibleMethod(Scope.java:806)
        at org.eclipse.jdt.internal.compiler.lookup.Scope.findMethod0(Scope.java:1775)
        at org.eclipse.jdt.internal.compiler.lookup.Scope.findMethod(Scope.java:1677)
        at org.eclipse.jdt.internal.compiler.lookup.Scope.getImplicitMethod(Scope.java:2669)
        at org.eclipse.jdt.internal.compiler.ast.MessageSend.findMethodBinding(MessageSend.java:1136)
        at org.eclipse.jdt.internal.compiler.ast.MessageSend.resolveType(MessageSend.java:872)
        at org.eclipse.jdt.internal.compiler.ast.LocalDeclaration.resolve(LocalDeclaration.java:401)
        at org.eclipse.jdt.internal.compiler.ast.LocalDeclaration.resolve(LocalDeclaration.java:257)
        at org.eclipse.jdt.internal.compiler.ast.Statement.resolveWithBindings(Statement.java:507)
        at org.eclipse.jdt.internal.compiler.ast.ASTNode.resolveStatements(ASTNode.java:735)
        at org.eclipse.jdt.internal.compiler.ast.AbstractMethodDeclaration.resolveStatements(AbstractMethodDeclaration.java:721)
        at org.eclipse.jdt.internal.compiler.ast.MethodDeclaration.resolveStatements(MethodDeclaration.java:397)
        at org.eclipse.jdt.internal.compiler.ast.AbstractMethodDeclaration.resolve(AbstractMethodDeclaration.java:619)
        at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.resolve(TypeDeclaration.java:1556)
        at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.resolve(TypeDeclaration.java:1685)
        at org.eclipse.jdt.internal.compiler.ast.CompilationUnitDeclaration.resolve(CompilationUnitDeclaration.java:661)
        at org.eclipse.jdt.internal.compiler.Compiler.process(Compiler.java:957)
        at org.eclipse.jdt.internal.compiler.ProcessTaskManager.processing(ProcessTaskManager.java:135)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
        at java.base/java.lang.Thread.run(Thread.java:1583)
...
```

Raw output: [first10percent.txt](https://github.com/user-attachments/files/18487270/first10percent.txt)

fedejeanne:
Here's the output for the last 10%: [last10percent.txt](https://github.com/user-attachments/files/18487488/last10percent.txt)

There is a bit more variation in there, _e.g._ this stack trace:

```java
java.lang.RuntimeException: InferenceVariable.InferenceVarKey.hashCode() = 15763705
        at org.eclipse.jdt.internal.compiler.lookup.InferenceVariable$InferenceVarKey.hashCode(InferenceVariable.java:42)
        at java.base/java.util.HashMap.hash(HashMap.java:338)
        at java.base/java.util.HashMap.getNode(HashMap.java:576)
        at java.base/java.util.HashMap.get(HashMap.java:564)
        at org.eclipse.jdt.internal.compiler.lookup.InferenceVariable.get(InferenceVariable.java:74)
        at org.eclipse.jdt.internal.compiler.lookup.InferenceContext18.addInitialTypeVariableSubstitutions(InferenceContext18.java:307)
        at org.eclipse.jdt.internal.compiler.lookup.InferenceContext18.createInitialBoundSet(InferenceContext18.java:224)
        at org.eclipse.jdt.internal.compiler.lookup.ConstraintExpressionFormula.inferInvocationApplicability(ConstraintExpressionFormula.java:387)
        at org.eclipse.jdt.internal.compiler.lookup.InferenceContext18.inferInvocationApplicability(InferenceContext18.java:369)
        at org.eclipse.jdt.internal.compiler.lookup.ParameterizedGenericMethodBinding.computeCompatibleMethod18(ParameterizedGenericMethodBinding.java:252)
        at org.eclipse.jdt.internal.compiler.lookup.ParameterizedGenericMethodBinding.computeCompatibleMethod(ParameterizedGenericMethodBinding.java:92)
        at org.eclipse.jdt.internal.compiler.lookup.Scope.computeCompatibleMethod(Scope.java:849)
        at org.eclipse.jdt.internal.compiler.lookup.Scope.computeCompatibleMethod(Scope.java:806)
        at org.eclipse.jdt.internal.compiler.lookup.Scope.getStaticFactory(Scope.java:5489)
        at org.eclipse.jdt.internal.compiler.ast.AllocationExpression.inferDiamondConstructor(AllocationExpression.java:656)
        at org.eclipse.jdt.internal.compiler.ast.AllocationExpression.inferConstructorOfElidedParameterizedType(AllocationExpression.java:632)
        at org.eclipse.jdt.internal.compiler.ast.AllocationExpression.isCompatibleWith(AllocationExpression.java:613)
        at org.eclipse.jdt.internal.compiler.lookup.PolyTypeBinding.isCompatibleWith(PolyTypeBinding.java:42)
        at org.eclipse.jdt.internal.compiler.lookup.Scope.parameterCompatibilityLevel(Scope.java:5261)
        at org.eclipse.jdt.internal.compiler.lookup.Scope.parameterCompatibilityLevel(Scope.java:5220)
        at org.eclipse.jdt.internal.compiler.lookup.Scope.computeCompatibleMethod(Scope.java:881)
        at org.eclipse.jdt.internal.compiler.lookup.Scope.computeCompatibleMethod(Scope.java:806)
        at org.eclipse.jdt.internal.compiler.lookup.Scope.findMethod0(Scope.java:1775)
        at org.eclipse.jdt.internal.compiler.lookup.Scope.findMethod(Scope.java:1677)
        at org.eclipse.jdt.internal.compiler.lookup.Scope.getMethod(Scope.java:3161)
        at org.eclipse.jdt.internal.compiler.ast.MessageSend.findMethodBinding(MessageSend.java:1137)
        at org.eclipse.jdt.internal.compiler.ast.MessageSend.resolveType(MessageSend.java:872)
        at org.eclipse.jdt.internal.compiler.ast.Expression.resolve(Expression.java:1119)
        at org.eclipse.jdt.internal.compiler.ast.Statement.resolveWithBindings(Statement.java:507)
        at org.eclipse.jdt.internal.compiler.ast.ASTNode.resolveStatements(ASTNode.java:735)
        at org.eclipse.jdt.internal.compiler.ast.Block.resolve(Block.java:143)
        at org.eclipse.jdt.internal.compiler.ast.Statement.resolveWithBindings(Statement.java:507)
        at org.eclipse.jdt.internal.compiler.ast.IfStatement.resolve(IfStatement.java:316)
        at org.eclipse.jdt.internal.compiler.ast.Statement.resolveWithBindings(Statement.java:507)
        at org.eclipse.jdt.internal.compiler.ast.IfStatement.resolve(IfStatement.java:319)
        at org.eclipse.jdt.internal.compiler.ast.Statement.resolveWithBindings(Statement.java:507)
        at org.eclipse.jdt.internal.compiler.ast.ASTNode.resolveStatements(ASTNode.java:735)
        at org.eclipse.jdt.internal.compiler.ast.AbstractMethodDeclaration.resolveStatements(AbstractMethodDeclaration.java:721)
        at org.eclipse.jdt.internal.compiler.ast.MethodDeclaration.resolveStatements(MethodDeclaration.java:397)
        at org.eclipse.jdt.internal.compiler.ast.AbstractMethodDeclaration.resolve(AbstractMethodDeclaration.java:619)
        at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.resolve(TypeDeclaration.java:1556)
        at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.resolve(TypeDeclaration.java:1685)
        at org.eclipse.jdt.internal.compiler.ast.CompilationUnitDeclaration.resolve(CompilationUnitDeclaration.java:661)
        at org.eclipse.jdt.internal.compiler.Compiler.process(Compiler.java:957)
        at org.eclipse.jdt.internal.compiler.ProcessTaskManager.processing(ProcessTaskManager.java:135)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
        at java.base/java.lang.Thread.run(Thread.java:1583)
```

jukzi:
> I would urge that we base any decisions on real data

@fedejeanne can you measure with `removing InferenceVariable.hashCode() / equals()` https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3585 if it gets any second faster?

jukzi:
> ```java
> java.lang.RuntimeException: InferenceVariable.InferenceVarKey.hashCode() = <some_hashcode>
> ```

@fedejeanne that output is not what i meant. The File InferenceVariable contains two `hashCode() ` you gave the stacktrace for the wrong one "InferenceVarKey" - the Key class did not show up in your measurement and is supposed to have a custom equals/hash - while InferenceVariable (without "Key" postfix) is not.

srikanth-sankaran:
Are we barking up the wrong tree ? If hashCode() distribution is poor and results in collision shouldn't we looking at whether equals() dominates runtime ??

jukzi:
May depend on what the caller does. Lets wait for the stacktrace.

jukzi:
that InferenceVariable.hashCode() may relate to https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3590#issuecomment-2606523240

fedejeanne:
> May depend on what the caller does. Lets wait for the stacktrace.

Added:

image

And got:

[Stacktrace-13percent.txt](https://github.com/user-attachments/files/18502977/Ddevdev-SimRel-staging-Deveclipsedevplatformc.exe-13percent.txt)

Adding the line made the compilation extremely slow (as expected) so I aborted the compilation after 10 minutes (at 14%).

Moving on to measuring the compilation time as requested in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3387#issuecomment-2604159387. I will report back.

fedejeanne:
@jukzi I sampled again after removing `hashCode` and `equals` (#3585) and I see no improvement, it still takes ~10.5 minutes (**636 s**).

Here are the hotspots:
image

Here's the raw data: [20250120-1800 without hashCode and equals in InferenceVariable.zip](https://github.com/user-attachments/files/18503569/20250120-1800.without.hashCode.and.equals.in.InferenceVariable.zip)

jukzi:
I created a separate issue for  InferenceVariable.hashCode(): https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3593

jukzi:
> Here's the raw data: [20250120-1800 without hashCode and equals in InferenceVariable.zip](https://github.com/user-attachments/files/18503569/20250120-1800.without.hashCode.and.equals.in.InferenceVariable.zip)

that file reads like you still have a problem in org.eclipse.jdt.internal.compiler.lookup.TypeBinding.findSuperTypeOriginatingFrom(TypeBinding), probably due to your redundant superinterfaces:
image

Issue #3194, opened 2024-10-29
[24] don't request access to enclosing class if it's never used


stephan-herrmann:
Given this example:
```java
	import java.util.function.Supplier;
	@SuppressWarnings("unused")
	class Outer {
		void m() { }
		class Inner {
			Inner() {
				class Foo {
					void g() {
//						m();
					}
				}
				super();
				class Bar {
					static void r() {
						Supplier<Foo> sfoo = Foo::new;
					}
				};
			}
		}
	}
```

ecj complains against `Foo::new`:
```
No enclosing instance of the type Outer is accessible in scope
```
The message is true, but the ctor of Foo does not **need** that enclosing instance. This would only be the case when, e.g., the invocation `m()` is uncommented.

This shows that the mechanics introduced for flexible constructors creates synthetic arguments where none are needed (which then causes an error, when no value can be supplied to that synthetic argument).

stephan-herrmann:
Fixed by https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3198

stephan-herrmann:
As of 24-ea+26 javac now answers:
```
Outer.java:15: error: local class Foo cannot be instantiated from a static context
                                                Supplier<Foo> sfoo = Foo::new;
                                                                     ^
```
With this shift of focus we are actually challenging a different set of rules, viz. (spec version 20241101):

> § 15.13.1
> It is a compile-time error if the method reference expression is of the form ClassType :: [TypeArguments] new and a compile-time error would occur when determining an enclosing instance for ClassType as specified in [15.9.2](https://cr.openjdk.org/~gbierman/jep492/jep492-20241101/specs/flexible-constructor-bodies-jls.html#jls-15.9.2) (treating the method reference expression as if it were an unqualified class instance creation expression).

So we proceed to 15.9.2 as if `Foo::new` were `new Foo()`:

> §15.9.2: 
> * If C is an inner local class, then:
>   * If C occurs in a static context, ...
>   * Otherwise, if the class instance creation expression occurs in a static context, then a compile-time error occurs.

With our example:
* Foo is an inner local class
  * Foo is not declared in a static context, so proceed to the next:
  * The instantiation occurs in the static context of method `r()`

Ergo: we should raise an error

(_reminder to self: conversely, also the error on `m()` when uncommented should be re-checked. It is not raised by javac_).

stephan-herrmann:
[...]

PR #3390, opened 2024-12-03
[24] don't request access to enclosing class if it's never used

stephan-herrmann:
Fixes after re-opening:
+ essentially revert previous change
+ replace with specific detection wrt local type alloc in static context

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3194

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/core/compiler/IProblem.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/core/compiler/IProblem.java
@@ -8,6 +8,10 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  *
+ * This is an implementation of an early-draft specification developed under the Java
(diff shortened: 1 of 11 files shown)


Jenkinsfile
```diff
@@ -33,7 +33,7 @@ pipeline {
 					mvn -U clean verify --batch-mode --fail-at-end -Dmaven.repo.local=$WORKSPACE/.m2/repository \
 						-Ptest-on-javase-24 -Pbree-libs -Papi-check -Pjavadoc -Pp2-repo \
 						-Dmaven.test.failure.ignore=true \
-						-Dcompare-version-with-baselines.skip=false \
+						-Dcompare-version-with-baselines.skip=true \
```
laeubi, line 36: Please make sure BETA_24 is in sync with master instead of skipping baseline checks as otherwise you probably see other issues later on.

stephan-herrmann: we're going in circles.

Issue #2413, opened 2024-05-05
ECJ fails: Problem detected during type inference: Unknown error at invocation of getReadOnly(P)


hohwille:
# Problem
I have a case with generics that is not compiled by EJC even though it seems perfectly valid and is accepted by javac as well.
It occurs in my OSS project exactly here:
https://github.com/m-m-m/entity/blob/f64a597454f32c7544826a2c4a171459f1937eb5/bean/src/test/java/io/github/mmm/entity/property/PropertyTest.java#L78
```java
P readOnly = WritableProperty.getReadOnly(property);
```

The inital generic `P` here is defined like this:
https://github.com/m-m-m/entity/blob/f64a597454f32c7544826a2c4a171459f1937eb5/bean/src/test/java/io/github/mmm/entity/property/PropertyTest.java#L23
```java
PropertyTest<V, P extends Property<V>>
```

The signature of the static method is as following:
https://github.com/m-m-m/property/blob/b71d7b3c8a320beaa841e24ee29bd5039b647bda/core/src/main/java/io/github/mmm/property/WritableProperty.java#L51
```java
static <P extends WritableProperty<?>> P getReadOnly(P property) { ... }
```

And of course `Property` implements `WritableProperty`:
https://github.com/m-m-m/property/blob/b71d7b3c8a320beaa841e24ee29bd5039b647bda/core/src/main/java/io/github/mmm/property/Property.java#L24
```java
public abstract class Property<V> extends AbstractWritableObservableValue<V> implements WritableProperty<V>, Cloneable
```

# Expected result
Code compiles in Eclipse without errors.

# Actual result
Compiler error:
```
Problem detected during type inference: Unknown error at invocation of getReadOnly(P)
```

![image](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/1923119/3c980bb3-0612-461c-94e6-d70ad5f1a0a7)

If you need further details, I should test anything or I should isolate a minimum code-snippet to reproduce the error, do not hesitate to ask.

# Side note
I seem to be one of the few Java developers on this planet that uses generics extensively and have also found many bugs in javac in this regard that Oracle seems to ignore. If you want to waste some time feel invited to have a look into my collection of such bugs:
https://github.com/m-m-m/util/issues/166
[...]

PR #3411, opened 2024-12-06
Guard lookup in BoundSet::allSuperPairsWithCommonGenericTypeRecursive

fedejeanne:
See proposal from @szarnekow in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3387#discussion_r1871077718

The compilation time was further reduced from ~12 min to **~10.5 min** 👏 

[Sample.zip](https://github.com/user-attachments/files/18036065/Sample.zip)

Contributes to https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3327

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/BoundSet.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/BoundSet.java
@@ -1279,8 +1279,9 @@ class BoundSet {
 		}
 		if (TypeBinding.equalsEquals(s,  t))
 			return result; // optimization #2: nothing interesting above equal types
-		TypeBinding tSuper = t.findSuperTypeOriginatingFrom(s);
-		if (tSuper != null && s.isParameterizedType() && tSuper.isParameterizedType()) { // optimization #1 again
+
+		TypeBinding tSuper = s.isParameterizedType() ? t.findSuperTypeOriginatingFrom(s) : null;
+		if (tSuper != null && tSuper.isParameterizedType()) { // optimization #1 again
 			result.add(new Pair<>(s, tSuper));
 		}
 		result.addAll(allSuperPairsWithCommonGenericTypeRecursive(s.superclass(), t, visited));
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/BoundSet.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/BoundSet.java
@@ -1264,34 +1264,38 @@ class BoundSet {
 	}
 
 	protected List<Pair<TypeBinding>> allSuperPairsWithCommonGenericType(TypeBinding s, TypeBinding t) {
-		return allSuperPairsWithCommonGenericTypeRecursive(s, t, new HashSet<>());
+		ArrayList<Pair<TypeBinding>> result = new ArrayList<>();
+		allSuperPairsWithCommonGenericTypeRecursive(s, t, result, new HashSet<>());
+		return result;
 	}
 
-	private List<Pair<TypeBinding>> allSuperPairsWithCommonGenericTypeRecursive(TypeBinding s, TypeBinding t, HashSet<TypeBinding> visited) {
+	private void allSuperPairsWithCommonGenericTypeRecursive(TypeBinding s, TypeBinding t, List<Pair<TypeBinding>> result, HashSet<Integer> visited) {
 		if (s == null || s.id == TypeIds.T_JavaLangObject || t == null || t.id == TypeIds.T_JavaLangObject)
-			return Collections.emptyList();
-		if (!visited.add(s.prototype()))
-			return Collections.emptyList();
-		List<Pair<TypeBinding>> result = new ArrayList<>();
-		if (s.isParameterizedType() && t.isParameterizedType() // optimization #1: clients of this method only want to compare type arguments
-				&& TypeBinding.equalsEquals(s.original(), t.original())) {
-			result.add(new Pair<>(s, t));
-		}
-		if (TypeBinding.equalsEquals(s,  t))
-			return result; // optimization #2: nothing interesting above equal types
+			return;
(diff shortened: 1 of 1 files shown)


fedejeanne:
cc @stephan-herrmann , @jukzi

jukzi:
While the change itself looks good, the description looks wrong. in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3387#issuecomment-2516609355 you stated your build took only 9 min so the impact of this change alone should not be that drastic. Please create a dedicated Issue for this problem (its not a regression as in the other PRs), link it in the PR and provide before/after measurements that only account to this PR.

fedejeanne:
> While the change itself looks good, the description looks wrong. in [#3387 (comment)](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3387#issuecomment-2516609355) you stated your build took only 9 min so the impact of this change alone should not be that drastic. Please create a dedicated Issue for this problem (its not a regression as in the other PRs), link it in the PR and provide before/after measurements that only account to this PR.

I think some wires got crossed: in the previous comment I stated that the build took **~12 min** and it took time (or better said: _took **place**_) between ~9:23 and ~9:35

image

In any case, I can confirm that the build time after merging #3384 and #3387 took **721 s** ...

image

... and the build with this PR took **626 s**

image

So the total gain of this PR is **721s - 626s = 95s** which is a reduction of **95 / 721 = 0.13 --> 13%**.

jukzi:
or 40 sec below org.eclipse.pde.ds.internal.annotations.DSAnnotationCompilationParticipant.canSkipFile () really?

fedejeanne:
But the 7 min stated in [#3411 (comment)](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3411#issue-2722559894) still do not match 626s (10,4 min)

😓  I changed the description, thank you!

> Please also note that the provided snapshot still contains hotspots that should not be there. for example spending 65 sec below org.eclipse.jdt.internal.core.builder.ClasspathDirectory.isPackage() does not make sense, so there should be much more room for optimization. or still 36 sec below org.eclipse.jdt.internal.compiler.lookup.TypeBinding.findSuperTypeOriginatingFrom () 12 sec below org.eclipse.osgi.internal.resolver.StateHelperImpl.getPackages () WTH
> 
> I hope you can contribute more performance improvements :-)

I'll look into them and I'll try to provide PRs for them too 👍 

But that doesn't affect this PR. A performance gain of 95s is also worth merging, don't you think?

jukzi:
> But that doesn't affect this PR. A performance gain of 95s is also worth merging, don't you think?

sure.

fedejeanne:
@jukzi does this mean that this PR can be merged today? . I would like to roll-out this change in our IDE next week so some of our developers can start battle-testing it :-)

stephan-herrmann:
Thanks @szarnekow  @fedejeanne  @jukzi , your continued interest in this issue led me to one more debugging session, where I made a few "interesting" observations:
* my use of `TypeBinding.prototype()` was well-intentioned but wrong, since `ParameterizedTypeBinding` constantly answers `null` , ups!
   * on closer look, we have something much better `TypeBinding.id` (the same id that is also used in `TypeBinding.equalsEquals()`. So no more worries about `hashCode()` in this issue :)
* to check effectiveness of the algorithm I tried to make the original tests from #2413 fail, but I couldn't 
  * this could simply depend on some order that changed in the meanwhile
  * I'll definitely follow up on that
  * it's an unfortunate property of these details deep in type inference, that creating a test case to challenge one particular code region can be arbitrarily difficult.

With these observations, I left no stone unturned trying to let theory demonstrate correctness, where testing fails. As part of this polish I also avoid creating fresh result lists for every recursive invocation.

As for the performance impact my poor-man's measurements still show a tiny improvement over #3387

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/BoundSet.java
```diff
@@ -1264,33 +1264,38 @@ private boolean superOnlyRaw(TypeBinding g, TypeBinding s, LookupEnvironment env
...
+		if (!visited.add(s.id))
+			return;
+
+		// optimization: nothing interesting above equal types
 		if (TypeBinding.equalsEquals(s,  t))
-			return result; // optimization #2: nothing interesting above equal types
-		TypeBinding tSuper = t.findSuperTypeOriginatingFrom(s);
-		if (tSuper != null && s.isParameterizedType() && tSuper.isParameterizedType()) { // optimization #1 again
-			result.add(new Pair<>(s, tSuper));
+			return;
```
szarnekow, line 1280: Previously we added `(s, t)` before checking equalsEquals. It's no longer the same algorithm, right?

Old:
```
s -> List<String>
t ->  Collection<String>
s.superInterfaces recurses into visiting (Collection<String>, Collection<String>) and adds that pair
```
Now:
```
s -> List<String>
t ->  Collection<String>
s.superInterfaces recurses into visiting (Collection<String>, Collection<String>) and exists immediately
```

Is that intentional?

szarnekow: That being said, would it be valid to exit after adding the pair in line 1278 of the old version without relying on equalsEquals(s, t)?

stephan-herrmann: > Previously we added `(s, t)` before checking equalsEquals. It's no longer the same algorithm, right?
> ...
> Is that intentional?

Yes, adding a pair of equal types is unnecessary. This change avoids that.

stephan-herrmann: > That being said, would it be valid to exit after adding the pair in line 1278 of the old version without relying on equalsEquals(s, t)?

That might be difficult to show. At that point we only know that generic types (`original()`) are equal. How the differences in type arguments percolate into super types is a non-trivial question. As ecj now is more than twice as fast than javac (in my poor-man's measurements), I'd rather not invest more in this particular code section. There are probably areas more in need of optimization.

Issue #3412, opened 2024-12-06
TypeBinding hashCode may be bad


jukzi:
Especially UnresolvedReferenceBinding.hashCode()
See discussion around https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3387#issuecomment-2518767219
It is not obvious if the method violates equals contract or is a good hashCode(), given that ReferenceBindingSetWrapper tries to work around it.
I will investigate.

jukzi:
- ArrayBinding overrides hashCode() but not equals() without explanation.
- LocalTypeBinding overrides hashCode() but not equals() without explanation.
- WildcardBinding overrides hashCode() but not equals() without explanation.
- ReferenceBinding overrides hashCode() but not equals() - comment that it should hash same as UnresolvedReferenceBinding
- TypeSystem.HashedParameterizedTypes.PTBKey overrides hashCode() using System.identityHashCode for some cases only (looks inconsistent)

There where 2 commits that introduced similar workarounds: PTBKey (https://bugs.eclipse.org/bugs/show_bug.cgi?id=521438, ReferenceBindingSetWrapper (https://bugs.eclipse.org/bugs/show_bug.cgi?id=544921) for noticed performance problems.

A clean solution would be to no require UnresolvedReferenceBindings have same hash as ReferenceBinding or to copy the hashCode from the unresolved to the resolved one.
The question is where that property is used. @stephan-herrmann

jukzi:
property is used in org.eclipse.jdt.internal.compiler.lookup.TypeSystem.updateCaches(UnresolvedReferenceBinding, ReferenceBinding) to update keys in a org.eclipse.jdt.internal.compiler.util.SimpleLookupTable TypeSystem.annotationTypes

carstenartur:
Maybe while at it some hashcode implementions could be updated in case they are not updated already (did not check)

https://git.eclipse.org/r/c/jdt/eclipse.jdt.core/+/170961

stephan-herrmann:
> property is used in org.eclipse.jdt.internal.compiler.lookup.TypeSystem.updateCaches(UnresolvedReferenceBinding, ReferenceBinding) to update keys in a org.eclipse.jdt.internal.compiler.util.SimpleLookupTable TypeSystem.annotationTypes

[...]

PR #3415, opened 2024-12-06
TypeBinding: What happens without custom hash()?

jukzi:

DO NOT SUBMIT


stephan-herrmann:
The failure might be suitable for analyzing the intended semantics:
```
org.eclipse.jdt.core.tests.builder.IncrementalTests18.testBug442452

Unexpected problem(s)!!!.

Problem : The interface Comparable cannot be implemented more than once with different arguments: Comparable<IBasicItem> and Comparable<IBasicItem> [ resource : </Project/Entity.java> range : <13,19> category : <40> severity : <2>]
```

Apparently, this test run succeeded to create two bindings for one and the same "IBasicItem", which is incorrect.

PR #3416, opened 2024-12-07
Fix source type binding key on windows

BoykoAlex:
On Windows since path separator is `\` the binding key ends up being `Lp/<path to the java file>~X`, i.e. includes file path. On Linux and Mac due to separator being `/` the key is `Lp/X;` as expected.
PR should fix the binding key on Windows

### DIFF
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/dom/StandAloneASTParserTest.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/dom/StandAloneASTParserTest.java
@@ -1943,7 +1943,73 @@ public class StandAloneASTParserTest extends AbstractRegressionTest {
         assertNotNull(ctorBinding);
         IAnnotationBinding[] annotations = ctorBinding.getAnnotations();
         assertNotNull(annotations);
-}
+	}
+
+	public void testBindingKey() throws IOException {
+		File rootDir = new File(System.getProperty("java.io.tmpdir"));
+		ASTParser parser = ASTParser.newParser(AST_JLS_LATEST);
+		parser.setEnvironment(null, null, null, true);
+		parser.setResolveBindings(true);
+		parser.setStatementsRecovery(true);
+		parser.setBindingsRecovery(true);
+		parser.setCompilerOptions(getCompilerOptions());
+
+		final IBinding[] bindings = new IBinding[1];
+
+		String contents =
+			"package p;\n" +
+			"public class X {\n" +
+			"}";
+
+		File packageDir = new File(rootDir, "p");
+		packageDir.mkdir();
+		File file = new File(packageDir, "X.java");
+		Writer writer = null;
+		try {
+			writer = new BufferedWriter(new FileWriter(file));
+			writer.write(contents);
+		} finally {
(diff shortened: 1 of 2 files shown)


BoykoAlex:
Could someone please look into merging this PR? It's rather simple fix with a unit test... (cc: @martinlippert @jukzi)

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/dom/StandAloneASTParserTest.java
```diff
@@ -1961,7 +1961,73 @@ class com {
...
+		String contents =
+			"package p;\n" +
+			"public class X {\n" +
+			"}";
+
+		File packageDir = new File(rootDir, "p");
+		packageDir.mkdir();
+		File file = new File(packageDir, "X.java");
+		Writer writer = null;
+		try {
```
jukzi, line 1986: please use try-with-resource

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/dom/StandAloneASTParserTest.java
```diff
@@ -1961,7 +1961,73 @@ class com {
...
+				try {
+					writer.close();
+				} catch(IOException e) {
+					// ignore
+				}
+			}
+		}
+
+		try {
+			final String canonicalPath = file.getCanonicalPath();
```
jukzi, line 2000: why do you need super slow canonicalPath here? is absolute path enough?

martinlippert:
Thanks a lot @jukzi for jumping on this and for providing feedback, very much appreciated. Look like @BoykoAlex incorporated your suggestions already. Do you think this is ready for merging into master?

iloveeclipse:
Merge commits are not allowed. Please rebase your changes on top of master branch instead.

BoykoAlex:
Squashed and rebased. Should be good to merge now.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/dom/StandAloneASTParserTest.java
jukzi: this question was not answered

jukzi:
> Look like @BoykoAlex incorporated your suggestions already. Do you think this is ready for merging into master?

does not look like. Or did it get lost during merge?

martinlippert:
> > Look like @BoykoAlex incorporated your suggestions already. Do you think this is ready for merging into master?
> 
> does not look like. Or did it get lost during merge?

Weird. I think I saw those questions addressed in the code, but maybe I am wrong here. I am sure @BoykoAlex can shed some light on this... :-)

BoykoAlex:
I don't know how it lost the additional "corrections" commit. I squashed 2 commits into one. I see that indeed those changes are gone. I've just added the changes again myself and pushed additional commit to my branch but the PR isn't showing this additional commit :-\

jukzi:
you need to force push

jukzi:
Or create a new PR, if you having trouble.

PR #3423, opened 2024-12-09
Fix "Redundant superinterface" warning depends on Interface name

jukzi:
SuperTypeTest.test016(), SuperTypeTest.test017() failed when implementation of ReferenceBinding.hashCode() changed.

The message was either
"already defined by IChangeRulerColumn" or
"already defined by IRevisionRulerColumn"
depending on which Interface had the lower hashCode.

Now using a LinkedHashMap to rely on the insertion order rather then the hashcode.

relates to


srikanth-sankaran:
Thanks @jukzi I'll review this tomorrow

PR #3425, opened 2024-12-09
batch compiler.util: writeToDisk use Files.write()

jukzi:
instead of streaming, because less source and it is a bit faster. For example MethodVerifyTest on windows: 65s->64s


jukzi:
image
->
image

PR #3426, opened 2024-12-09
ExceptionHandlingFlowContext: use java.util.IdentityHashMap

jukzi:
* avoids raw types
* deletes custom ObjectCache

ObjectCache implemented a HashMap using hashCode() but not equals(). Instead it used == identity. JDK offers a special Map for identity since 1.4, consistently using System.identityHashCode().

relates to https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3412


srikanth-sankaran:
Thanks @juzki I will review these tomorrow

Issue #3393, opened 2024-12-04
6 new compiler test errors on Java 17 builds since I20241203-1800


iloveeclipse:
https://download.eclipse.org/eclipse/downloads/drops4/I20241203-1800/testresults/html/org.eclipse.jdt.core.tests.compiler_ep435I-unit-linux-x86_64-java17_linux.gtk.x86_64_17.html

```
testGH3278_missingAnnotations - 1.8	Failure	Invalid problem log . ----------- Expected ------------ ------------ but was ------------ ----------\n 1. ERROR in annotated\TestNotOwning.java (at line 23)\n for (Closeable closeable : toClose.reversed()) {\n ^^^^^^^^\n The method reversed() is undefined for the type List<Closeable>\n ----------\n ---------------------- ---------- expected:<[]> but was:<[----------\n 1. ERROR in annotated\TestNotOwning.java (at line 23)\n for (Closeable closeable : toClose.reversed()) {\n ^^^^^^^^\n The method reversed() is undefined for the type List<Closeable>\n ----------\n ]>

junit.framework.ComparisonFailure: Invalid problem log .
----------- Expected ------------

------------ but was ------------
----------\n
1. ERROR in annotated\TestNotOwning.java (at line 23)\n
for (Closeable closeable : toClose.reversed()) {\n
^^^^^^^^\n
The method reversed() is undefined for the type List<Closeable>\n
----------\n

---------------------- ----------
expected:<[]> but was:<[----------\n
1. ERROR in annotated\TestNotOwning.java (at line 23)\n
for (Closeable closeable : toClose.reversed()) {\n
^^^^^^^^\n
The method reversed() is undefined for the type List<Closeable>\n
----------\n
]>
at org.eclipse.jdt.core.tests.junit.extension.TestCase.assertStringEquals(TestCase.java:265)
at org.eclipse.jdt.core.tests.junit.extension.TestCase.assertEquals(TestCase.java:240)
at org.eclipse.jdt.core.tests.compiler.regression.AbstractRegressionTest.checkCompilerLog(AbstractRegressionTest.java:1599)
at org.eclipse.jdt.core.tests.compiler.regression.AbstractRegressionTest.runTest(AbstractRegressionTest.java:3587)
at org.eclipse.jdt.core.tests.compiler.regression.AbstractRegressionTest.runTest(AbstractRegressionTest.java:3290)
[...]

PR #3427, opened 2024-12-09
Resolve 3 remaining test failures

stephan-herrmann:
It seems `File.separator` is an escape character on windows ;-P

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3393


jukzi:
Worked, thanks.

PR #3432, opened 2024-12-10
[performance] avoid File.getCanonicalPath()

jukzi:
File.getCanonicalPath() is a IO Operation that is costly especially on windows OS.
Instead use a NON-IO .toPath().normalize().toAbsolutePath().toString()

https://bugs.eclipse.org/bugs/show_bug.cgi?id=571614


jukzi:
Found it as hotspot in org.eclipse.jdt.internal.compiler.batch.ClasspathDirectory.<init> () while runnning NullTypeAnnotationTest
This PR reduces NullTypeAnnotationTest from 102 sec to 92 sec
image

image

PR #3433, opened 2024-12-10
[performance] optimized ClasspathDirectory

jukzi:
* java.io.File.list() returns null if the file is not a directory, so File.isDirectory() is not needed.
* use ConcurrentHashMap instead of Hashtable
* avoid rawtypes

ClasspathDirectory.directoryList(String) is a hotspot for example during NullTypeAnnotationTest

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathDirectory.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathDirectory.java
@@ -359,6 +359,7 @@ public char[][] listPackages() {
 public void reset() {
 	super.reset();
 	this.directoryCache.clear();
+	this.packageSecondaryTypes.clear();
 }
 @Override
 public String toString() {
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathDirectory.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathDirectory.java
@@ -26,10 +26,10 @@ import java.nio.file.Path;
 import java.nio.file.SimpleFileVisitor;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.util.HashSet;
-import java.util.Hashtable;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


jukzi:
This PR reduces NullTypeAnnotationTest from 102 sec to 99 sec
image
image

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathDirectory.java
```diff
@@ -360,7 +358,7 @@ public FileVisitResult visitFile(Path file, BasicFileAttributes attrs) {
 @Override
 public void reset() {
 	super.reset();
-	this.directoryCache = new Hashtable(11);
+	this.directoryCache.clear();
```
szarnekow, line 361: Do you know why teh secondaryTypes are not cleared?

jukzi: I also wondered about that. i decided to only make a performance change here keeping logic as is.

iloveeclipse: @mpalat : any clue? `packageSecondaryTypes` field was added in c39467a5717299e0743e249af795ac84f7541935

mpalat: > @mpalat : any clue? `packageSecondaryTypes` field was added in [c39467a](https://github.com/eclipse-jdt/eclipse.jdt.core/commit/c39467a5717299e0743e249af795ac84f7541935)

I believe its a bug - @jukzi  could you please add code to clear this also in reset alongwith this PR.

jukzi: will do

mpalat:
@jukzi There is an issue in this code:
	String[] cached = this.directoryCache.computeIfAbsent(qualifiedPackageName, this::computeDirectoryList);
Here this could result in recursive computeIfAbsent() which may end up in IllegalStateException due to recursive update since the doesFileExist also may call this on the same ConcurrentHashMap - This needs to be corrected

raised issue #3445 for the same. Thanks

PR #3436, opened 2024-12-11
fix PTBKey updateWhenSettingTypeAnnotations

jukzi:
Changing PTBKey.arguments[idx] changes PTBKey.equals() and PTBKey.hashCode().
So the old PTBKey has to be removed and a new has to be inserted. It was wrong to keep the PTBKey in hashedParameterizedTypes. Found by codereview. It is not clear if that could produce an error other then having bad performance by having garbage entries in the HashMap that could furthermore not found because the hash/equals contract was violated.

Code is executed for example during
compiler.apt.tests.Java8ElementsTests.testTypeAnnotations()

relates to

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeSystem.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeSystem.java
@@ -94,9 +94,10 @@ public class TypeSystem {
 								TypeVariableBinding typeVariableBinding = (TypeVariableBinding) argument;
 								Consumer<TypeVariableBinding> previousConsumer = typeVariableBinding.updateWhenSettingTypeAnnotations;
 								typeVariableBinding.updateWhenSettingTypeAnnotations = (newTvb) -> {
-									// update the TVB argument and simulate a re-hash:
-									ParameterizedTypeBinding[] value = HashedParameterizedTypes.this.hashedParameterizedTypes.get(this);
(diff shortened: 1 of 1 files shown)


jukzi:
failed 

```
1. ERROR in abc\d\DMessageHandlerRegistryImpl.java (at line 4)\r\n
	public class DMessageHandlerRegistryImpl<@NonNull C extends DConnection>\r\n
	             ^^^^^^^^^^^^^^^^^^^^^^^^^^^\n
The interface MessageHandlerRegistry cannot be implemented more than once with different arguments: 
MessageHandlerRegistry<C,CharSequence,DIncomingMessageData> and 
MessageHandlerRegistry<C,CharSequence,DIncomingMessageData>\n
]>
	at org.eclipse.jdt.core.tests.junit.extension.TestCase.assertStringEquals(TestCase.java:265)
	at org.eclipse.jdt.core.tests.junit.extension.TestCase.assertEquals(TestCase.java:240)
	at org.eclipse.jdt.core.tests.compiler.regression.AbstractRegressionTest.runTest(AbstractRegressionTest.java:3600)
	at org.eclipse.jdt.core.tests.compiler.regression.AbstractRegressionTest.runTest(AbstractRegressionTest.java:3405)
	at org.eclipse.jdt.core.tests.compiler.regression.AbstractRegressionTest$Runner.runConformTest(AbstractRegressionTest.java:126)
	at org.eclipse.jdt.core.tests.compiler.regression.NullTypeAnnotationTest.testGH2158(NullTypeAnnotationTest.java:19313)

```

PR #3437, opened 2024-12-11
remove HashedParameterizedTypes

jukzi:
The PTBKey looks complicated and errorprone (updating hashCode on an Element in a HashMap). Let's find out what was really slow and then see if we find a better solution for
"linear effort during lookup"
(https://bugs.eclipse.org/bugs/show_bug.cgi?id=434326#c2)

partial revert commit 510e79670ef0450a78c7882799bbb7cd266ab2e6 (Bug 434326 - Slow compilation with a significant amount of generics)

keeps binary search in
TypeSystem.cacheDerivedType(TypeBinding keyType, TypeBinding derivedType)


jukzi:
EXPERIMENT - DO NOT MERGE

PR #3444, opened 2024-12-12
[23] Add implementation for Types#stripAnnotations() #3439

### LINKED ISSUE #3439: [23] Add implementation for Types#stripAnnotations()
This was added in Java 23. ECJ should provide implementation in the `TypesImpl`.

jarthana:
## What it does
Fixes #3439

## Author checklist


stephan-herrmann:
@jarthana  I was wondering if version bumps like in this MR could contribute to the build failures in BETA_JAVA24 (until those changes are merged into the branch). 

Which lets me ask if such interference could possibly be avoided by smart use of different increments? I recall precise rules when to use +100, +50 or even just +1 in the service segment, but I can't remember where I should look for such rules. Since BETA_JAVA24 will be released after the current master stream, would it make sense to ensure that all versions in the beta branch are and remain greater than those in master?

jarthana:
> @jarthana I was wondering if version bumps like in this MR could contribute to the build failures in BETA_JAVA24 (until those changes are merged into the branch).
> 
> Which lets me ask if such interference could possibly be avoided by smart use of different increments? I recall precise rules when to use +100, +50 or even just +1 in the service segment, but I can't remember where I should look for such rules. Since BETA_JAVA24 will be released after the current master stream, would it make sense to ensure that all versions in the beta branch are and remain greater than those in master?

You know, before I fixed the version wondering about the same. I first added 50 and then thought it would be better if we kept n(100) on master and so we can have n(100) + 50 on the BETA branch if it changes. But I don't understand why you said all versions should be greater than master? We have always only done that for the bundles that changed.

stephan-herrmann:
> But I don't understand why you said all versions should be greater than master? We have always only done that for the bundles that changed.

My idea was, if bundles in beta have +100 then master can remain at +0 or increment +50 at any time without impacting beta builds. And since beta spans two regular release cycles, master could even do another +10 to still remain below beta. Perhaps this strategy makes most sense to test bundles, which hardly ever increase the minor segment, and micro-bumps are spread more over the release cycle. Also most test bundles have no clients to which these version bumps would ripple etc., i.e. a preemptive bump for test bundles shouldn't harm, should it?

Anyway, if more frequent merges  - at least after each version change in master - is not too much of a burden then this is fine. I was just trying to minimize the need for this.

PR #3388, opened 2024-12-03
update for run.javac testing wrt JDK-8341408

stephan-herrmann:
https://bugs.openjdk.org/browse/JDK-8341408 has been resolved 

Since preview features can be tested only against latest there is no use in keeping Excuse JavacBug8341408 => remove

PR #3446, opened 2024-12-12
[BUG] [JSR-199] ECJ cannot resolve JPMS modules if using a user-provided file manager

stephan-herrmann:
ClasspathJsr199:
+ various drafty additions for handling modules

ModuleFinder:
+ find automatic modules, too
+ allow invocation without a parser for binary-only scans

EclipseFileManager:
+ allow scanning all packages (by specifying an empty package name)
+ establish symmetry between setLocation() and setLocationFromPaths()
+ catch disruptive IAE from ModuleFinder.scanForModule()

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/958


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathJsr199.java
```diff
@@ -156,12 +160,29 @@ public char[][][] findTypeNames(String aQualifiedPackageName, String moduleName)
...
+		} else if (this.location.isModuleOrientedLocation()) {
+			Iterable<Set<Location>> locationsForModules = this.fileManager.listLocationsForModules(this.location);
+			for (Set<Location> locs: locationsForModules) {
+				for (Location loc : locs) {
+					if (loc instanceof LocationWrapper wrapper) {
+						for (Path locPath : wrapper.getPaths()) {
+							File file = locPath.toFile();
+							IModule mod = ModuleFinder.scanForModule(this, file, null, true, null);
+							if (mod != null) {
+								return;
```
stephan-herrmann, line 172: Several locations now assume that each ClasspathJsr199 is either JRT or a single module classpath location. See in particular new uses of the inherited field `module`. At a second look this is probably wrong, and each ClasspathJsr199 must be able to handle multiple modules.

Is this initialization the place that should actually create a `Map<Location,IModule>`? I'm a bit confused by the multi-level multiplicities: an Iterable of Sets of Locations where each Location has multiple Paths. At which of these levels can we assume a unique module? Is a LocationWrapper the right level? So if a module is in a jar, can we assume that it is represented by one LocationWrapper containing exactly one path?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathJsr199.java
```diff
@@ -156,12 +160,29 @@ public char[][][] findTypeNames(String aQualifiedPackageName, String moduleName)
...
+			}
 		}
 	}
 
 	@Override
 	public void acceptModule(IModule mod) {
-		// do nothing
+		if (this.jrt != null)
+			return; // do nothing
+		this.module = mod;
```
stephan-herrmann, line 185: If we need to handle multiple modules, then we probably need to let `initialize()` remember the module, since only there we know the corresponding LocationWrapper.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathJsr199.java
```diff
@@ -195,6 +216,29 @@ public char[][] getModulesDeclaringPackage(String aQualifiedPackageName, String
 		return singletonModuleNameIf(result);
 	}
 
+	@Override
+	public char[][] listPackages() {
+		Set<String> packageNames = new HashSet<>();
+		try {
+			for (JavaFileObject fileObject : this.fileManager.list(this.location, "", fileTypes, true)) { //$NON-NLS-1$
```
stephan-herrmann, line 223: Listing all class files in all packages is surely a performance issue. Also passing "" to signal no filtering is a bit crude (needed an adjustment inside ModuleFinder, too). Unfortunately I found no better we to list all packages, which, however, the compiler needs particularly for auto modules.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathJsr199.java
```diff
@@ -195,6 +216,29 @@ public char[][] getModulesDeclaringPackage(String aQualifiedPackageName, String
...
+				}
+			}
+			char[][] result = new char[packageNames.size()][];
+			int i = 0;
+			for (String s : packageNames) {
+				result[i++] = s.toCharArray();
+			}
+			return result;
+		} catch (IOException e) {
+			// ??
```
stephan-herrmann, line 237: I don't know what's the best strategy to handle exceptions in this code area.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathJsr199.java
```diff
@@ -250,6 +294,22 @@ public boolean hasAnnotationFileFor(String qualifiedTypeName) {
...
+			Set<String> moduleNames = new HashSet<>();
+			try {
+				for (Set<Location> locs : this.fileManager.listLocationsForModules(this.location)) {
+					for (Location loc : locs) {
+						String moduleName = this.fileManager.inferModuleName(loc);
+						if (moduleName != null)
+							moduleNames.add(moduleName);
+					}
+				}
+				return moduleNames;
```
stephan-herrmann, line 307: If the implementation is good, it could perhaps benefit from caching.

But even the sequence of invocations looks funny: we list module locations, so those locations are already known to contain a module. Is it really necessary to then "infer" the module name? Wait, it seems we could use `if (loc instanceof ModuleLocationWrapper wrapper) .. wrapper.getModuleName()`.

To do so we'd need to
* make that class public
* provide a `getModuleName()` accessor

Next, I'm not even sure if it is safe to use our classes like LocationWrapper. Wouldn't that break if a custom file manager is used??

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathJsr199.java
```diff
@@ -269,6 +329,11 @@ public IModule getModule(char[] name) {
 		return super.getModule(name);
 	}
 
+	@Override
+	public IModule getModule() {
+		return this.module;
```
stephan-herrmann, line 334: Called by the compiler. If we handle multiple modules, how should the compiler ask for a specific module?

At this point I wonder if the ClasspathJsr199 with Location MODULE_PATH should actually be resolved to a **set** of classpaths, just like ClasspathContainers in the IDE?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/tool/EclipseFileManager.java
```diff
@@ -1421,15 +1430,19 @@ public void setLocationFromPaths(Location location, Collection<? extends Path> p
...
+							new Parser(problemReporter, true), null, true, this.releaseVersion);
+					for (Classpath cp : mp) {
+						Collection<String> moduleNames = cp.getModuleNames(null);
+						for (String string : moduleNames) {
+							Path p = Paths.get(cp.getPath());
+							setLocationForModule(location, string,  Collections.singletonList(p));
+						}
 					}
+				} catch (IllegalArgumentException iae) { // e.g., from ModuleFinder.scanForModule(Classpath, File, Parser, boolean, String)
+					// FIXME ignore for now
```
stephan-herrmann, line 1445: CompilerToolJava9Tests.testBug566749() expects a regular compile error, which now additionally surfaces as the IllegalArgumentException caught here.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathJsr199.java
stephan-herrmann: > Wait, it seems we could use ... ModuleLocationWrapper ...

Never mind, `inferModuleName()` is implemented in that exact way I was considering for the client side, so this part looks OK :)

Still caching the set of module names might be a good idea.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathJsr199.java
```diff
@@ -156,12 +160,29 @@ public char[][][] findTypeNames(String aQualifiedPackageName, String moduleName)
 	public void initialize() throws IOException {
 		if (this.jrt != null) {
 			this.jrt.initialize();
+		} else if (this.location.isModuleOrientedLocation()) {
+			Iterable<Set<Location>> locationsForModules = this.fileManager.listLocationsForModules(this.location);
+			for (Set<Location> locs: locationsForModules) {
+				for (Location loc : locs) {
+					if (loc instanceof LocationWrapper wrapper) {
```
stephan-herrmann, line 167: This use of LocationWrapper leaks an abstraction, it will not work with other file managers.

But then, how to find the contained paths??

jarthana: If it were a StandardJavaFileManager we could use `getLocationAsPaths(Location)`

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathJsr199.java
jarthana: Yep, looks like it's not adequate for such scenarios. ClasspathJsr199 is more of a version of ClasspathLocation that knows a bit about Location but not doing a good job of morphing into a Location.

stephan-herrmann: Should I try to move the loops from `ClasspathJsr199.initialize()` into `EclipseCompilerImpl.handleLocations()` so that we can create individual `ClasspathJsr199` per module?

nettozahler:
During this analysis and overhaul maybe you could also have a look at PR #3057 since it also touches ClasspathJsr199. Eventually #3057 becomes obsolet with the changes proposed here.

stephan-herrmann:
In 477cab7f43975cf6cf7620b76adefbcd8affd350 I changed the strategy towards using one ClasspathJsr199 **per module**.

@jarthana from my p.o.v. this already looks much better. So perhaps you could take another look?

I have some local changes attempting to support also MODULE_SOURCE_PATH, but it seems, ClasspathJsr199 is not suitable for handling sources.

stephan-herrmann:
Test failed with the same bogus error as has been blocking #3388 for some time now.

laeubi:
> Test failed with the same bogus error as has been blocking #3388 for some time now.

I would try to rebase the change on top of master as a new release-stream has started in the meanwhile.

stephan-herrmann:
> > Test failed with the same bogus error as has been blocking #3388 for some time now.
> 
> I would try to rebase the change on top of master as a new release-stream has started in the meanwhile.

The change is based on I20241211-1800. Are you saying a new stream has started just within the last three days?? It looks like the stream change happened on 2024-11-23.

laeubi:
At the time of my comment there where three commits and at least one of them looks quite old, the other was a merge commit and then a "recent" one.

Now you fixed that up, the build seem working fine (all checks are green).

stephan-herrmann:
> At the time of my comment there where three commits and at least one of them looks quite old, the other was a merge commit and then a "recent" one,

I have no clue where you saw this, but of course after you made me rebase the branch, I'm no longer able to prove that the branch indeed started from HEAD of master only 4 days ago.

> beside that github showed merge conflicts. All these can probably lead to the build being confused about the actual baseline in use.
> 
> Now you fixed that up, the build seem working fine (all checks are green).

I wouldn't call this strategy "fixing" but working around same intermittent bug. And of course none of this is of any help for #3388

laeubi:
> I wouldn't call this strategy "fixing" but working around same intermittent bug

Contributions are always welcome to fix such "bugs", but as JDT is not a standalone project several things influence the outcome (e.g. changes to aggregator parent pom, I-Build repositories and so on) of a PR that depends on such setups.

stephan-herrmann:
> > I wouldn't call this strategy "fixing" but working around same intermittent bug
> 
> Contributions are always welcome to fix such "bugs", but as JDT is not a standalone project several things influence the outcome (e.g. changes to aggregator parent pom, I-Build repositories and so on) of a PR that depends on such setups.

This is not helpful.

laeubi:
> This is not helpful.

I showed you how to resolve the issue and it works, I'm not sure how to be more helpful... instead the solution is claimed a "workaround" for a bug.

stephan-herrmann:
> > This is not helpful.
> 
> I showed you how to resolve the issue and it works, I'm not sure how to be more helpful... instead the solution is claimed a "workaround" for a bug.

I'm hoping for a solution that is applicable also for BETA_JAVA24 without requiring that branches need to be tightly in sync, which would be an unacceptable requirement. Also I'm hoping for a solution where I'm able to understand the connection between the error and the remedy.

laeubi:
> I'm hoping for a solution that is applicable also for BETA_JAVA24 without requiring that branches need to be tightly in sync, which would be an unacceptable requirement

The build infrastructure for (feature) branches is currently to ensure they can be merged into master at some point in time what requires them to be at least close to the target branch. Why this is _unacceptable requirement_ is beyond my knowledge it more feels like a personal preference than a technical requirement, at least most contributors have adapted to the process to keep branches up-to date.

Anyways there are some *minimal requirements* such a branch must fulfill e.g. it requires to include changes in version of bundles compared to the current integration build and it should not have any merge conflicts.

> Also I'm hoping for a solution where I'm able to understand the connection between the error and the remedy.

I hoped to explain it [here](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2660#issuecomment-2543776898) its a bit hard to keep track of what is said where, but to summarize for this particular issue.

### Where it was failing before

Lets look at the last failing [build logs here](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-3446/4/consoleText) and we see it fails with:

```
[2024-12-15T09:58:04.486Z] [ERROR] Failed to execute goal org.eclipse.tycho.extras:tycho-p2-extras-plugin:4.0.10:compare-version-with-baselines (compare-attached-artifacts-with-release) on project org.eclipse.jdt.compiler.apt.tests: Baseline and reactor have the same fully qualified version, but different content
[2024-12-15T09:58:04.486Z] [ERROR] different
[2024-12-15T09:58:04.486Z] [ERROR]    META-INF/ECLIPSE_.RSA: present in baseline only
[2024-12-15T09:58:04.486Z] [ERROR]    META-INF/ECLIPSE_.SF: present in baseline only
[2024-12-15T09:58:04.486Z] [ERROR] -> [Help 1]
```

If we search up for the offending project we will see this (**baseline-replace with latest ibuild**):

```
[2024-12-15T09:56:33.139Z] [INFO] --- tycho-p2-plugin:4.0.10:p2-metadata-default (default-p2-metadata-default) @ org.eclipse.jdt.compiler.apt.tests --- 
[2024-12-15T09:56:33.139Z] [INFO] No baseline version org.eclipse.jdt:org.eclipse.jdt.compiler.apt.tests:eclipse-test-plugin:1.3.600-SNAPSHOT
```

We see that actually there is no baseline version found, because its [version was changed on master here](https://github.com/eclipse-jdt/eclipse.jdt.core/commit/b25d09198ef4f6a89f749bd067ca7af5780e2e17) 5 days ago (as of today) to version `1.3.700` while your branch is still `1.3.600`

Now the next check kicks in (**baseline-compare** with latest release):

```
[2024-12-15T09:56:59.730Z] [INFO] --- tycho-p2-extras:4.0.10:compare-version-with-baselines (compare-attached-artifacts-with-release) @ org.eclipse.jdt.compiler.apt.tests ---
[2024-12-15T09:56:59.730Z] [ERROR] Baseline and reactor have the same fully qualified version, but different content
[2024-12-15T09:56:59.730Z] different
[2024-12-15T09:56:59.730Z]    META-INF/ECLIPSE_.RSA: present in baseline only
[2024-12-15T09:56:59.730Z]    META-INF/ECLIPSE_.SF: present in baseline only
```

it complains that you have build _the exact same version as in the previous release_ but the content slightly differs what is classified as an error for platform builds as it usually means you have a missing version increment or something else is badly wrong.

**Please note that at this point there are no ignores!** And of course this is not a bug...

### How it was fixed

You have _rebased_ your branch on top of master and thus now it includes the version change as well, you could have *merged* master but we don't like merge commits here... another option would have been a squash merge (and later rebase) but this would pollute your PR with unrelated diffs / code changes maybe.

### Why it is working afterwards

Lets look at the last successful [build logs here](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-3446/5/consoleText) and we see:

```
[2024-12-15T15:13:27.543Z] [INFO] --- tycho-p2-plugin:4.0.10:p2-metadata-default (default-p2-metadata-default) @ org.eclipse.jdt.compiler.apt.tests ---
[2024-12-15T15:13:27.543Z] [INFO] MavenProject: org.eclipse.jdt:org.eclipse.jdt.compiler.apt.tests:1.3.700-SNAPSHOT @ /home/jenkins/agent/workspace/eclipse.jdt.core-Github_PR-3446/org.eclipse.jdt.compiler.apt.tests/pom.xml
[2024-12-15T15:13:27.543Z]     The main artifact has been replaced with the baseline version.
[2024-12-15T15:13:27.543Z]     The following attached artifacts have been replaced with the baseline version: [sources]
```

So we see the baseline artifact from the last ibuild was found and your currently build jar was compared to the baseline  and it was found to be _semantically equal_ (**and this is where we taking ignores into account**) and therefore replaced by the jar from the last ibuild.

Later on then we see

```
[2024-12-15T15:13:56.002Z] [INFO] --- tycho-p2-extras:4.0.10:compare-version-with-baselines (compare-attached-artifacts-with-release) @ org.eclipse.jdt.compiler.apt.tests ---
[2024-12-15T15:13:56.002Z] [INFO] 
```

so no error at all here, because both artifacts have _different content_ (as the bundle was changed) but also _different version_ that is higher than in the last release.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathJsr199.java
stephan-herrmann: Obsoleted by creating one ClasspathJsr199 for each module / location.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathJsr199.java
stephan-herrmann: Done

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathJsr199.java
stephan-herrmann: Obsoleted by creating one `ClasspathJsr199` for each module.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathJsr199.java
stephan-herrmann: I now have alternative implementations using `this.fileManager.list()`, i.e., relying neither on `LocationWrapper` nor on `StandardJavaFileManager`.

jarthana (approved):
I am fine with the patch as is. But my point about encoding might be worth considering. 
I see the exception handling was also raised in https://bugs.eclipse.org/bugs/show_bug.cgi?id=188796
but no clear answer was given by anyone. Olivier has suggested it to be handled case by case and
when we must abort the compilation, we (re)throw the exception. So, in that sense, throwing AbortCompilation is
fine too as long as we embed the cause.

stephan-herrmann:
@ascopes  I think this PR is already a substantial improvement over the current state. If you agree, we could merge this PR and continue investigating further details in follow-up issues, OK?

_(I still plan to slightly re-arrange the PR to separate squashed implementation changes from version bumps)_

ascopes:
Sounds good to me!

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/apt/dispatch/BatchFilerImpl.java
```diff
@@ -43,13 +48,32 @@ public class BatchFilerImpl implements Filer {
...
 		this._dispatchManager = dispatchManager;
 		this._fileManager = env._fileManager;
 		this._env = env;
 		this._createdFiles = new HashSet<>();
+		this._encoding = main.getDefaultEncoding();
+		if (this._fileManager.hasLocation(StandardLocation.SOURCE_PATH)) {
+			try {
+				for (JavaFileObject javaFileObject : this._fileManager.list(StandardLocation.SOURCE_PATH, "", //$NON-NLS-1$
+						Collections.singleton(Kind.SOURCE), false)) {
+					if (javaFileObject.getName().equals(IModule.MODULE_INFO_JAVA)) {
```
ascopes, line 65: small question about this... I was just checking the JLS for this for Java 9... and they seem to suggest `module-info.jav` is also a potentially valid name?

> If and only if packages are stored in a file system ([§7.2](https://docs.oracle.com/javase/specs/jls/se9/html/jls-7.html#jls-7.2)), the host system may choose to enforce the restriction that it is a compile-time error if a module declaration is not found in a file under a name composed of module-info plus an extension (such as .java or .jav).

This seems a bit odd to me, do we know if we're supposed to respect that as well (both on your side and my side)?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/tool/EclipseFileManager.java
```diff
@@ -151,13 +152,14 @@ private void collectAllMatchingFiles(Location location, File file, String normal
...
 				String key = normalizedPackageName;
+				boolean findAll = key.length() == 0;
 				if (!normalizedPackageName.endsWith("/")) {//$NON-NLS-1$
 					key += '/';
 				}
 				// we have an archive file
 				if (recurse) {
 					for (String packageName : archive.allPackages()) {
-						if (packageName.startsWith(key)) {
+						if (findAll || packageName.startsWith(key)) {
```
ascopes, line 162: if `findAll` is true, the `key` will have zero length... looks like `.startsWith` will still return true in that case

```
jshell> "foo".startsWith("")
$1 ==> true
```
is the `findAll` check needed?

ascopes:
Thanks for spending the time working on this, it is greatly appreciated 🎉

ascopes:
Might have found an edge case that is missed here. Will document it in #958.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/tool/EclipseFileManager.java
stephan-herrmann: Thanks for checking, but key will never have length 0, see `key += '/'` just above.

stephan-herrmann:
> Might have found an edge case that is missed here. Will document it in #958.

As per https://github.com/eclipse-jdt/eclipse.jdt.core/issues/958#issuecomment-2564696990 I don't see a direct connection to JPMS. So that case should not block this current PR. @ascopes  could you please file new issues for any remaining problems?

ascopes:
Does that include the existing one I flagged?

stephan-herrmann:
> Does that include the existing one I flagged?

Yes please. And when doing so feel free to @mention me. Any issues with relation to JPMS I will take a look.

Issue #3445, opened 2024-12-12
Possible IllegalStateException due to recursive call on ConcurrentHashMap in directory list.


mpalat:
ref: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3433 - The performance optimization due to introduction of ConcurrentHashMap :

In ClassPathDirectory.directoryList() method
String[] cached = this.directoryCache.computeIfAbsent(qualifiedPackageName, this::computeDirectoryList);
Here this could result in recursive computeIfAbsent() which may end up in IllegalStateException due to recursive update since the doesFileExist also may call this on the same ConcurrentHashMap - This needs to be corrected

@jukzi can you please take a look?

jukzi:
will do, thanks for analysis.

jukzi:
![image](https://github.com/user-attachments/assets/40396e3a-817c-460a-b952-aecdf3641266)

jukzi:
@mpalat is there any workflow / junitTest where it does actually happen?

jukzi:
Can you give me a hint what that special handling of 'uppercase character' is about? I could not find any special treatmeant for lower/uppercase in the JLS nor in Character.isJavaIdentifierStart() or Character.isJavaIdentifierPart()

mpalat:
Let me try to get an example

mpalat:
@jukzi Was able to create a test case - please take a look at PR #3449.
 I would request you to revert that commit of ConcurrentHashMap addition for now. re-opening this. Thanks.

PR #3450, opened 2024-12-13
ClasspathDirectory.directoryList(): check parent package first #3445

jukzi:
Avoid nested directoryCache.computeIfAbsent to fix possible IllegalStateException with long upper case packages.

fixes
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3445


jukzi:
cc @szarnekow

mpalat (approved):
@jukzi  Thanks for reverting - Looks fine.

+1

PR #3452, opened 2024-12-13
Make PTBKey not extend ReferenceBinding

jukzi:
* Introduces interface TypeBindingWrapper.
* Removes little code from PTBKey that is irrelevant for the wrapper

relates to https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3412

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeBinding.java
@@ -1685,11 +1685,6 @@ public char[] signature() {
 
 public abstract char[] sourceName();
 
-public void swapUnresolved(UnresolvedReferenceBinding unresolvedType,
-		ReferenceBinding resolvedType, LookupEnvironment environment) {
-	// subclasses must override if they wrap another type binding
-}
-
 TypeBinding [] typeArguments () {
 	return null;
 }
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeSystem.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeSystem.java
@@ -69,7 +69,7 @@ public class TypeSystem {
 
 	public final class HashedParameterizedTypes {
 
-		private final class PTBKey extends ReferenceBinding { // extends ReferenceBinding so it can be used as wrapper
+		private final class PTBKey implements HotSwappable {
 			protected ReferenceBinding type; // must ensure the type is resolved
 			public TypeBinding[] arguments;
 			private ReferenceBinding enclosingType;
@@ -87,8 +87,6 @@ public class TypeSystem {
 							TypeBinding argument = arguments[i];
 							if (argument instanceof UnresolvedReferenceBinding)
 								((UnresolvedReferenceBinding) argument).addWrapper(this, environment);
-							if (argument.hasNullTypeAnnotations())
-								this.tagBits |= TagBits.HasNullTypeAnnotation;
 							if (argument.getClass() == TypeVariableBinding.class) {
 								final int idx = i;
 								TypeVariableBinding typeVariableBinding = (TypeVariableBinding) argument;
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/UnresolvedReferenceBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/UnresolvedReferenceBinding.java
@@ -20,10 +20,10 @@ package org.eclipse.jdt.internal.compiler.lookup;
 
 import org.eclipse.jdt.core.compiler.CharOperation;
 
-public class UnresolvedReferenceBinding extends ReferenceBinding {
+public class UnresolvedReferenceBinding extends ReferenceBinding implements HotSwappable {
 
 ReferenceBinding resolvedType;
-TypeBinding[] wrappers;
+HotSwappable[] wrappers;
 UnresolvedReferenceBinding prototype;
 ReferenceBinding requestingType;
 
@@ -53,7 +53,7 @@ public TypeBinding clone(TypeBinding outerType) {
 	return copy;
 }
 
-void addWrapper(TypeBinding wrapper, LookupEnvironment environment) {
+void addWrapper(HotSwappable wrapper, LookupEnvironment environment) {
 	if (this.resolvedType != null) {
 		// the type reference B<B<T>.M> means a signature of <T:Ljava/lang/Object;>LB<LB<TT;>.M;>;
 		// when the ParameterizedType for Unresolved B is created with args B<T>.M, the Unresolved B is resolved before the wrapper is added
@@ -61,10 +61,10 @@ void addWrapper(TypeBinding wrapper, LookupEnvironment environment) {
 		return;
 	}
 	if (this.wrappers == null) {
-		this.wrappers = new TypeBinding[] {wrapper};
+		this.wrappers = new HotSwappable[] {wrapper};
 	} else {
 		int length = this.wrappers.length;
-		System.arraycopy(this.wrappers, 0, this.wrappers = new TypeBinding[length + 1], 0, length);
+		System.arraycopy(this.wrappers, 0, this.wrappers = new HotSwappable[length + 1], 0, length);
 		this.wrappers[length] = wrapper;
 	}
 }
@@ -148,7 +148,7 @@ void setResolvedType(ReferenceBinding targetType, LookupEnvironment environment)
 	// must ensure to update any other type bindings that can contain the resolved type
 	// otherwise we could create 2 : 1 for this unresolved type & 1 for the resolved type
 	if (this.wrappers != null)
-		for (TypeBinding wrapper : this.wrappers)
+		for (HotSwappable wrapper : this.wrappers)
 			wrapper.swapUnresolved(this, targetType, environment);
 }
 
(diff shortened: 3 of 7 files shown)


jukzi:
@srikanth-sankaran may i ask for review? I would like to improve ReferenceBinding ini small steps.

srikanth-sankaran:
> @srikanth-sankaran may i ask for review? I would like to improve ReferenceBinding in small steps.

I'll try - but can't promise, I am trying to wrap up some projects I am working on before this week since post that I am away till 2nd Jan.

jukzi:
@srikanth-sankaran ping, please review.

srikanth-sankaran:
Sure, I will look into these this week. Happy new year!

srikanth-sankaran:
> Sure, I will look into these this week. Happy new year!

Sorry, it looks like my full return to work will be delayed until a few days more. I will be at work 3 out of 5 days next week and I am hopeful the review can be completed well ahead of M2.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeBindingWrapper.java
```diff
@@ -0,0 +1,5 @@
+package org.eclipse.jdt.internal.compiler.lookup;
+
+interface TypeBindingWrapper {
+	void swapUnresolved(UnresolvedReferenceBinding unresolvedType, ReferenceBinding resolvedType, LookupEnvironment env);
+}
```
srikanth-sankaran, line 5: This file is missing suitable copyrights

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeBinding.java
```diff
@@ -55,7 +55,7 @@
  *
  * null is NOT a valid value for a non-public field... it just means the field is not initialized.
  */
-abstract public class TypeBinding extends Binding {
+abstract public class TypeBinding extends Binding implements TypeBindingWrapper {
```
srikanth-sankaran, line 59: I prefer `implements Resolvable`

(edited) Not sure actually. I am still trying to understand what this additional new name buys us. Any suggested name would depend on that

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeSystem.java
```diff
@@ -87,8 +87,6 @@ public PTBKey(ReferenceBinding type, TypeBinding[] arguments, ReferenceBinding e
 							TypeBinding argument = arguments[i];
 							if (argument instanceof UnresolvedReferenceBinding)
 								((UnresolvedReferenceBinding) argument).addWrapper(this, environment);
-							if (argument.hasNullTypeAnnotations())
-								this.tagBits |= TagBits.HasNullTypeAnnotation;
 							if (argument.getClass() == TypeVariableBinding.class) {
```
srikanth-sankaran, line 90: What explains the deletion of these lines ? I don't see a ready rationale.

(Edit) OK, that seems fine after all,

srikanth-sankaran (commented):
See comments, need some clarifications on what exactly we are trying to accomplish here.

I am yet to come to grips with what the new additional interface buys us.

Also the title of the PR `PTBKey is not a ReferenceBinding` - need clarification. What exactly due to mean by that.

I don't see anything seriously wrong/faulty - but simply don't get yet where we are going with this and why

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/UnresolvedReferenceBinding.java
```diff
@@ -53,18 +53,18 @@ public TypeBinding clone(TypeBinding outerType) {
 	return copy;
 }
 
-void addWrapper(TypeBinding wrapper, LookupEnvironment environment) {
+void addWrapper(TypeBindingWrapper wrapper, LookupEnvironment environment) {
```
jukzi, line 56: This is the only usage of the proposed interface. Here the instance is called "wrapper" and the method name add"Wrapper" also suggest to have "Wapper" in the interface name. Thats why i chose "TypeBindingWrapper". but i would be fine with any other Name.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeSystem.java
jukzi: These lines just do not make sense for a Resolvable and have been probably only introduced by copy&paste from an actual ReferenceBinding.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeBinding.java
jukzi: renamed to Resolvable as requested

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeBindingWrapper.java
jukzi: added copyright

srikanth-sankaran:
Not sure if you are planning to explain the PR title change - the new one still doesn't clear the mystery for me :)

Do you have a scenario where the PTBKey is actually not a reference binding ?? 

Without the additional name TypeBindingWrapper being used, things were working so far - so what does extracting the swapUnresolved behavior into its own type buy us ?

jukzi:
> Without the additional name TypeBindingWrapper being used, things were working so far - so what does extracting the swapUnresolved behavior into its own type buy us ?

I am trying to clean up the ReferenceBinding equals/hashCode contract as it hard to understand and errorprone to accidentialy use when using a ReferenceBinding as key (for example https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3561 ). As this PTBKey does not need to implement ReferenceBinding one could just ignore the PTBKey in such scenarios. At the end it will be fine if the PTBKey will have a custom equals/hashcode, but any actual ReferenceBinding probably won't need to override equals/hashcode => no more worries in the future.

srikanth-sankaran:
> > Without the additional name TypeBindingWrapper being used, things were working so far - so what does extracting the swapUnresolved behavior into its own type buy us ?
> 
> I am trying to clean up the ReferenceBinding equals/hashCode contract as it hard to understand and errorprone to accidentialy use when using a ReferenceBinding as key (for example #3561 ). As this PTBKey does not need to implement ReferenceBinding one could just ignore the PTBKey in such scenarios. At the end it will be fine if the PTBKey will have a custom equals/hashcode, but any actual ReferenceBinding probably won't need to override equals/hashcode => no more worries in the future.

PTBKey is not "implementing" `ReferenceBinding` but ISA `ReferenceBinding` - do you know of cases where it is NOT a ReferenceBinding ? If not this whole PR will become moot and will only serve to obfuscate by introducing additional names...

jukzi:
> PTBKey is not "implementing" `ReferenceBinding` but ISA `ReferenceBinding` - do you know of cases where it is NOT a ReferenceBinding ? If not this whole PR will become moot and will only serve to obfuscate by introducing additional names...

It was a ReferenceBinding but did not need to be ReferenceBinding. It is only a wrapper used as key for ParameterizedTypeBinding. It is not used as ReferenceBinding itself anywhere. Especially PTBKey.hash(TypeBinding) will never call PTBKey.hashCode() which was not clear from the type hierarchy.

srikanth-sankaran:
> > PTBKey is not "implementing" `ReferenceBinding` but ISA `ReferenceBinding` - do you know of cases where it is NOT a ReferenceBinding ? If not this whole PR will become moot and will only serve to obfuscate by introducing additional names...
> 
> It was a ReferenceBinding but did not need to be ReferenceBinding. It is only a wrapper used as key for ParameterizedTypeBinding. It is not used as ReferenceBinding itself anywhere. Especially PTBKey.hash(TypeBinding) will never call PTBKey.hashCode() which was not clear from the type hierarchy.

OK, I finally have an understanding of what you are trying to do. With that in mind, taking a fresh relook ...

srikanth-sankaran:
I have reviewed the fix and I don't have major objections to it. 

Yes, the PTBKey behavior does not call for it to be a `ReferenceBinding` per se and to factor out the relevant behavior into an interface is OK.

(1) I am still not very happy with the name - 'Resolvable` is way better than `TypeBindingWrapper` which was plain WRONG. In the original patch, `TypeBinding` was implementing `TypeBindingWrapper` - That is wrong. TypeBinding is actually a Wrap**pable** and not a Wra**pper** (but see below)

But Resolvable is not precise enough. I wonder if a name like `HotSwappable` is more to the point. Sorry for overriding my own earlier suggestion and causing rework. 

(2) A `TypeBinding` itself is not `HotSwappable` - ONLY those abstractions in the hierarchy that implement the method `org.eclipse.jdt.internal.compiler.lookup.TypeBinding.swapUnresolved(UnresolvedReferenceBinding, ReferenceBinding, LookupEnvironment)` are. That presently `TypeBinding` has a dummy/nop/placeholder implementation does not count.


I think this PR may look a lot better 

    - if `Resolvable` is renamed to be `HotSwappable` and
    - if  `HotSwappable` is implemented ONLY by `ArrayBinding`, `ParameterizedTypeBinding`, `WildCardBinding`, `PTBKey` and `UnresolvedReferenceBinding` 
    - if `TypeBinding` itself DOES NOT implement it

What do you think ? If you agree, you may proceed to make the changes and integrate once tests are green. Thanks

srikanth-sankaran (approved):
Changes look good. Please proceed to integrate assuming no surprises in testing.

jukzi:
the failing test is known to randomly fail: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2716#issuecomment-2510794711

PR #3466, opened 2024-12-17
Jenkinsfile: longer history, less artifacts

jukzi:
Keep more builds to get better statistics for master and BETA branches. But less builds for others (especially PRs).

https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/2653


jukzi:
like https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/576

Issue #3470, opened 2024-12-17
UI freeze with hotspot in JavaModelManager.determineIfOnClasspath()


jukzi:
I just had a ~ 2min UI freeze after changing branches while Package Explorer was open
![image](https://github.com/user-attachments/assets/9c1ebacb-29a4-432b-9647-0acbf435942f)

eclipse.buildId=4.35.0.I20241216-1800

Example stack trace:

```
"main" #1 [29680] prio=6 os_prio=0 cpu=537109.38ms elapsed=9777.28s tid=0x0000023f2ceb4fa0 nid=29680 runnable  [0x000000cc5592c000]
   java.lang.Thread.State: RUNNABLE
        at org.eclipse.core.internal.resources.ProjectPreferences.internalGet(ProjectPreferences.java:525)
        at org.eclipse.core.internal.preferences.EclipsePreferences.get(EclipsePreferences.java:447)
        at org.eclipse.jdt.internal.core.JavaModelManager.getOption(JavaModelManager.java:2349)
        at org.eclipse.jdt.internal.core.JavaProject.getOption(JavaProject.java:2061)
        at org.eclipse.jdt.internal.core.JavaModelManager.determineIfOnClasspath(JavaModelManager.java:1121)
        at org.eclipse.jdt.internal.core.JavaModelManager.create(JavaModelManager.java:1011)
        at org.eclipse.jdt.core.JavaCore.create(JavaCore.java:3799)
        at org.eclipse.jdt.internal.ui.packageview.PackageExplorerContentProvider.getHierarchicalPackagesInFolder(PackageExplorerContentProvider.java:490)
        at org.eclipse.jdt.internal.ui.packageview.PackageExplorerContentProvider.getFolderContent(PackageExplorerContentProvider.java:284)
        at org.eclipse.jdt.ui.StandardJavaElementContentProvider.getChildren(StandardJavaElementContentProvider.java:193)
        at org.eclipse.jdt.internal.ui.packageview.PackageExplorerContentProvider.getChildren(PackageExplorerContentProvider.java:309)
        at org.eclipse.jdt.internal.ui.navigator.JavaNavigatorContentProvider.getChildren(JavaNavigatorContentProvider.java:177)
        at org.eclipse.jdt.ui.StandardJavaElementContentProvider.hasChildren(StandardJavaElementContentProvider.java:240)
[...]

PR #3475, opened 2024-12-18
[performance] determineIfOnClasspath with last project used #3470

jukzi:
Instead of randomly looping over all projects try the last successful first.

Tested by JavaModelTests.testCreatePkgHandleInDifferentProject()

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3470

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java
@@ -980,6 +980,8 @@ public class JavaModelManager implements ISaveParticipant, IContentTypeChangeLis
 		return null;
 	}
 
+	private static volatile String lastProjectNameUsed;
+
 	/**
 	 * Returns the package fragment or package fragment root corresponding to the given folder,
 	 * its parent or great parent being the given project.
@@ -999,6 +1001,15 @@ public class JavaModelManager implements ISaveParticipant, IContentTypeChangeLis
 			project = JavaCore.create(folder.getProject());
 			element = determineIfOnClasspath(folder, project);
 			if (element == null) {
+				IJavaProject lastProject = lastProjectNameUsed == null ? null
+						: JavaModelManager.getJavaModelManager().getJavaModel().getJavaProject(lastProjectNameUsed);
+				if (lastProject != null) {
+					// try to avoid searching through all projects
+					element = determineIfOnClasspath(folder, lastProject);
+					if (element != null) {
+						return element;
+					}
+				}
 				// walk all projects and find one that have the given folder on its classpath
 				IJavaProject[] projects;
 				try {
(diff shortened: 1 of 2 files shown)


org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java
```diff
@@ -999,6 +999,14 @@ public static IJavaElement create(IFolder folder, IJavaProject project) {
 			project = JavaCore.create(folder.getProject());
 			element = determineIfOnClasspath(folder, project);
 			if (element == null) {
+				IJavaProject lastProject = lastProjectUsed;
+				if (lastProject != null) {
+					// try to avoid searching through all projects
+					element = determineIfOnClasspath(folder, lastProject);
```
iloveeclipse, line 1005: Not sure this is valid to use previously used project that could correspond to already closed/deleted resource project and be not present in JavaModel anymore

iloveeclipse: So what should happen is something like below, I haven't checked it for performance:

`lastProject = JavaModel.getJavaProject(lastProjectUsed.getElementName())`

iloveeclipse (changes requested):

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java
jukzi: I have extended the Junit test to show that it works without such workaround (as the IJavaProject is a Handle only)

iloveeclipse: This is a negative test only. What I'm more concerned about is the positive test - project exists, folder exists, we remember P1. Now P1 is closed (deleted), we still "see" P1 in lastProject and answer something weird.

iloveeclipse: Beside this, I don't like an idea to keep a static reference to a Java Project, as a potential for a memory leak because GC will not release this project (and resource project and all the references that might be connected to them recursive) anymore.

jukzi: > Now P1 is closed

The test does it and even opens the project afterwards again. I don't get what you miss. I don't see a problem with a kept project, as it is only a handle, but i could use some weaker reference.

iloveeclipse: No. The test checks if a non existing folder in project P2 does not exist in a project P2. 
Fine, it does not.

What I miss is that we have first a positive check  (folder F1 exists in project P1), close the project and try to create again.

Regarding some weaker reference: I assume just remembering `getElementName()` for the last project hit would do what you want.

jukzi: ok, using the String  name now

iloveeclipse (approved):
OK, thanks.

Do you have performance numbers for that?

jukzi:
> Do you have performance numbers for that?

I did not test it beside junit yet, as i could not reproduce it yet. It's just that the stacktrace clearly indicated that 1000 unrelated projects might have been searched.

jukzi:
Since also other reporters like https://bugs.eclipse.org/bugs/show_bug.cgi?id=541040#c7 observed that the perfromance problem can not be deterministically reproduced i think its best to submit even without real world benchmark and keep monitoring.

PR #3479, opened 2024-12-18
Mark `RequiresDirective`'s name as `MANDATORY`

datho7561:
Elsewhere in the code, it's treated as a mandatory property (for instance in AST rewrites), so it makes sense to update the property descriptor and Javadoc to reflect this.

## What it does
- Mark `RequiresDirective`'s name field as `MANDATORY`
- Switch the value of the lazily initialized name to be a `SimpleName`
- Document that `RequiredDirective.setName(Name name)` throws an `IllegalArgumentException` when attempting to set the name to `null`

## How to test
- [x] Write an integration test

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/RequiresDirective.java
+++ b/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/RequiresDirective.java
@@ -39,7 +39,7 @@ public class RequiresDirective extends ModuleDirective {
 	 * The module structural property of this node type (child type: {@link Name}).
 	 */
 	public static final ChildPropertyDescriptor NAME_PROPERTY =
-		new ChildPropertyDescriptor(RequiresDirective.class, "name", Name.class, OPTIONAL, NO_CYCLE_RISK); //$NON-NLS-1$
+		new ChildPropertyDescriptor(RequiresDirective.class, "name", Name.class, MANDATORY, NO_CYCLE_RISK); //$NON-NLS-1$
 
 	/**
 	 * A list of property descriptors (element type:
@@ -204,8 +204,7 @@ public class RequiresDirective extends ModuleDirective {
 			synchronized (this) {
 				if (this.name == null) {
 					preLazyInit();
-					this.name = postLazyInit(
-							this.ast.newQualifiedName(new SimpleName(this.ast), new SimpleName(this.ast)),
+					this.name = postLazyInit(new SimpleName(this.ast),
 							NAME_PROPERTY);
 				}
 			}
@@ -221,6 +220,7 @@ public class RequiresDirective extends ModuleDirective {
 	 * <ul>
 	 * <li>the node belongs to a different AST</li>
 	 * <li>the node already has a parent</li>
+	 * <li>the node is null</li>
 	 * </ul>
 	 */
 	public void setName(Name name) {
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverter9Test.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverter9Test.java
@@ -1596,6 +1596,39 @@ public class ASTConverter9Test extends ConverterTestSetup {
 			deleteProject("P2");
 		}
 	}
+	public void testRequiresDirectiveNameIsMandatory() throws JavaModelException, CoreException {
+		try {
+			assertTrue(RequiresDirective.NAME_PROPERTY.isMandatory());
+
+			IJavaProject project1 = createJavaProject("ASTParserModelTests", new String[] { "src" },
+					new String[] { "CONVERTER_JCL9_LIB" }, "bin", "9");
+			project1.open(null);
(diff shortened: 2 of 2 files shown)


datho7561:
@mpalat could you please review this code once you have a chance? From my understanding, you implemented `RequiresDirective`, so you probably have the best understanding of it. I don't understand why the name was being lazily initialized, and I don't understand why a qualified name was used in the lazy initalization.

mpalat:
> @mpalat could you please review this code once you have a chance? From my understanding, you implemented `RequiresDirective`, so you probably have the best understanding of it. I don't understand why the name was being lazily initialized, and I don't understand why a qualified name was used in the lazy initalization.

hmm.. well, let me scratch my old memory :)

Looking at the specs, I believe the name should have been MANDATORY rather than optional while it was modifiers that had to be optional - referring section - referring Section 7.7 of JLS :
requires {[RequiresModifier](https://docs.oracle.com/javase/specs/jls/se23/html/jls-7.html#jls-RequiresModifier)} [ModuleName](https://docs.oracle.com/javase/specs/jls/se23/html/jls-6.html#jls-ModuleName). So it's best to change the name to mandatory and leave lazy initialization in tact.

And to answer your second question on why a qualified name was used - I don't see why it should not be used. Guess that it so coded since this mimics import.  We can use either a Simple name or a QualifiedName since the ModuleName can be one of those - ref: [Section 6.5](https://docs.oracle.com/javase/specs/jls/se23/html/jls-6.html#jls-ModuleName). Am open for a change to SimpleName if it makes life easier in any case - but modules names tend to be qualified names mostly [no, I don;t have any statistics to prove that].

mpalat (changes requested):
>@mpalat could you please review this code once you have a chance? From my understanding, you implemented RequiresDirective, so you probably have the best understanding of it. I don't understand why the name was being lazily initialized, and I don't understand why a qualified name was used in the lazy initalization.

hmm.. well, let me scratch my old memory :)

Looking at the specs, I believe the name should have been MANDATORY rather than optional while it was modifiers that had to be optional - referring section - referring Section 7.7 of JLS :
requires {[RequiresModifier](https://docs.oracle.com/javase/specs/jls/se23/html/jls-7.html#jls-RequiresModifier)} [ModuleName](https://docs.oracle.com/javase/specs/jls/se23/html/jls-6.html#jls-ModuleName). So it's best to change the name to mandatory and leave lazy initialization in tact.

And to answer your second question on why a qualified name was used - I don't see why it should not be used. Guess that it so coded since this mimics import. We can use either a Simple name or a QualifiedName since the ModuleName can be one of those - ref: [Section 6.5](https://docs.oracle.com/javase/specs/jls/se23/html/jls-6.html#jls-ModuleName). Am open for a change to SimpleName if it makes life easier in any case - but modules names tend to be qualified names mostly [no, I don;t have any statistics to prove that].

So I believe the name should be made mandatory and Javadoc updated rather than allowing null.

datho7561:
Okay sounds good. I'll update this PR to mark the name as MANDATORY and update the Javadoc

datho7561:
I've added a new commit that marks the name as MANDATORY and switches the default name over to a SimpleName. I'm going to wait for the CI results before removing the first commit

datho7561:
Alright, it should be good to review: I've cleaned up the commit and added a test.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/RequiresDirective.java
```diff
@@ -39,7 +39,7 @@ public class RequiresDirective extends ModuleDirective {
 	 * The module structural property of this node type (child type: {@link Name}).
 	 */
 	public static final ChildPropertyDescriptor NAME_PROPERTY =
-		new ChildPropertyDescriptor(RequiresDirective.class, "name", Name.class, OPTIONAL, NO_CYCLE_RISK); //$NON-NLS-1$
+		new ChildPropertyDescriptor(RequiresDirective.class, "name", Name.class, MANDATORY, NO_CYCLE_RISK); //$NON-NLS-1$
```
mpalat, line 42: @datho7561 Since this implies an API change, we should increment `@since`. It is indeed a grey area when we look at the API revision doc, but from a conservative point of view, I believe we should increment the version number. 
@jarthana - checking with you on your opinion.

mpalat (changes requested):
Thanks @datho7561 for the change. Only one more comment which I have tagged @jarthana for his take as well.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/RequiresDirective.java
jarthana: I am not sure about the @since version option. But two things worth noting:
1. No where in the Javadoc of the property do we mention about the option/mandatory aspect.
2. One of the usual clients, JDT/UI doesn't seem to be referencing this property at all, surprisingly.

So, we might get away with just "fixing" this bug, IMO.

mpalat: > So, we might get away with just "fixing" this bug, IMO.

@jarthana - thanks for chiming in.  you are right that this is not mentioned in the JavaDOC.
@datho7561 Let's take a call to go ahead and merge this change. +1

mpalat (approved):

PR #3491, opened 2024-12-28
Remove "negative" javadoc tests for pre 1.5 mode

akurtakov:
JavadocTest_1_5.java contains same tests with same input but with "proper" expectations for modern java. As pre 1.5 is no longer supported deleting the tests directly.


akurtakov:
Covers JavadocTest* from https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2758

PR #3492, opened 2024-12-28
Cleanup disabled tests in ASTConverterTestAST4_2

akurtakov:
Some of the tests had asserts for pre 1.8 which is no longer accepted with a corresponding "uptodate" test in ASTConverterTestAST8_2 - these are deleted.
Other are enabled as they work now, most likely after the fixes to jclmin.
Cleaned ASTConverterTestAST8_2 tests from pre 1.8 expectations which are useless nowadays.


akurtakov:
Covers ASTConverterTestAST4_2 in https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2758

PR #3494, opened 2024-12-28
Remove tests checking lambda is not accepted at 1.7 level

akurtakov:
ECJ mandates 1.8 thus always accepts it and that's why the tests have been commented out.


akurtakov:
Part of https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2758

Issue #2941, opened 2024-09-10
Incorrect overload resolution ends in type error.


rluble:
The following code:
```
import static java.util.stream.Collectors.collectingAndThen;

import java.util.function.Function;
import java.util.stream.Collector;
import java.util.stream.Stream;

public abstract class BiStream<K, V> {
  public static <E, K, V> Collector<E, ?, BiStream<K, V>> toBiStream(
      Function<? super E, ? extends K> toKey, Function<? super E, ? extends V> toValue) {
    return collectingAndThen(toStream(), stream -> from(stream));
  }

  private static <T> Collector<T, ?, Stream<T>> toStream() {
    return null;
  }

  public static <T, K, V> BiStream<K, V> from(Iterable<T> elements) {
    return null;
  }

  public static <T, K, V> BiStream<K, V> from(Stream<T> stream) {
    return null;
  }
}

```

Fails to compile in ecj-4.32 with the following error:

```
1. ERROR in /Users/rluble/Downloads/BiStream.java (at line 10)
	return collectingAndThen(toStream(), stream -> from(stream));
	       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Type mismatch: cannot convert from Collector<Object,Object,Object> to Collector<E,?,BiStream<K,V>>
```

This is the result of selecting the wrong overload and incorrectly depending on declaration order.

If the overloads are declared in a different order as in:
```
import static java.util.stream.Collectors.collectingAndThen;

import java.util.function.Function;
import java.util.stream.Collector;
import java.util.stream.Stream;

public abstract class BiStream<K, V> {
  public static <E, K, V> Collector<E, ?, BiStream<K, V>> toBiStream(
      Function<? super E, ? extends K> toKey, Function<? super E, ? extends V> toValue) {
    return collectingAndThen(toStream(), stream -> from(stream));
  }

  private static <T> Collector<T, ?, Stream<T>> toStream() {
    return null;
  }

  public static <T, K, V> BiStream<K, V> from(Stream<T> stream) {
    return null;
  }

  public static <T, K, V> BiStream<K, V> from(Iterable<T> elements) {
    return null;
  }
}

```

the code compiles.

stephan-herrmann:
ECJ 3.35 and earlier (Eclipse 4.29) successfully compile the original example.

I'll check what commit caused this change.

PR #3515, opened 2025-01-04
Incorrect overload resolution ends in type error.

stephan-herrmann:
+ record when Poly*Method is part of a set of overloads
  + in that case don't propagate premature bounds inner -> outer

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2941

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
@@ -691,6 +691,8 @@ public class InferenceContext18 {
 			MethodBinding innerMethod = invocation.binding();
 			if (innerMethod == null)
 				return true; 		  // -> proceed with no new C set elements.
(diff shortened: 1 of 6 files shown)


stephan-herrmann:
Initially during inference, resolving `from(stream)` could not select an overload due to insufficient type information. At this point three things are created: two `PolyParameterizedMethodBinding`, one for each of the candidates, and one `ProblemMethodBinding(Ambiguous)`  holding one of the candidates as its `closestMatch`. This opportunistic selection is where declaration order influenced the outcome. Somehow that closest match surfaced in our implementation of [18.5.2.2.](https://docs.oracle.com/javase/specs/jls/se23/html/jls-18.html#jls-18.5.2.2) (specifically: `addConstraintsToC_OneExpr()`), where the type information of the inner `from(stream)` influenced the outer inference. When the wrong overload is used, this led to inference not finding the desired solution.

To fix this we'll now refuse to add further constraints based on type information from unresolved overload disambiguation. To detect the situation, at the point where we create the `ProblemMethodBinding(Ambiguous)` we tag the `closestMatch`  `PolyParameterizedMethodBinding`  as having overloads. That flag then prevents spilling premature information into outer inference.

Issue #3424, opened 2024-12-09
`StringLiteral.getLiteralValue` is not thread-safe


martinlippert:
It looks like `StringLiteral.getLiteralValue` is throwing `IllegalStateExceptions` in some cases when used from multiple threads, most likely due to the internal scanner object being used in a stateful way.

jukzi:
Please paste the Exception Stacktrace and why it is used in multiple threads

martinlippert:
Example stack traces can be found here: https://github.com/spring-projects/sts4/issues/1434 - but this comes from the Spring Tools using JDT Core to parse and analyze source code, so the stack traces have no origins in JDT or Eclipse itself.

We use the AST over there to implement content-assist, do additional validations, and extract index information from a Spring perspective.

jukzi:
That reference only shows a IllegalArgumentException, not a IllegalStateException , also the line number does not match master.
The IllegalArgumentException complains tokenType!=TerminalTokens.TokenNameStringLiteral

I don't see a connection to concurrent use. You should be able to set a breakpoint to find the sourcecode in question that cannot be parsed.

martinlippert:
> That reference only shows a IllegalArgumentException, not a IllegalStateException , also the line number does not match master. The IllegalArgumentException complains tokenType!=TerminalTokens.TokenNameStringLiteral

You are right, my bad, it is an `IllegalArgumentException`, not an `IllegalStateException`. The line numbers are from using JDT Core 3.39.

> I don't see a connection to concurrent use. You should be able to set a breakpoint to find the sourcecode in question that cannot be parsed.

I am not yet convinced about this not being related to concurrent use, but let me try to dig deeper here. Always hard to reproduce those things, but let me see what I can find out with our hints here. Thanks so much for looking into this so quickly and for providing the feedback here.

jukzi:
This issue still has not enough information to process.

martinlippert:
@martinlippert I am on it and will provide more details / information as soon as possible... :-)

martinlippert:
I created a code snippet to reproduce this issue with concurrent calls to `getLiteralValue`:

```
ASTParser parser = ASTParser.newParser(AST.JLS21);
parser.setSource("\"hello\\nworld\"".toCharArray());
parser.setKind(ASTParser.K_EXPRESSION);
ASTNode ast = parser.createAST(null);

final StringLiteral literal = (StringLiteral) ast;

[...]

PR #3509, opened 2025-01-03
StringLiteral.getLiteralValue is not thread-safe #3424

jukzi:
org.eclipse.jdt.core.dom.StringLiteral.getLiteralValue() unsynchronized use of shared org.eclipse.jdt.core.dom.AST.scanner

reproducer for
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3424

PR #3521, opened 2025-01-06
Make StringLiteral.getLiteralValue thread safe

martinlippert:
Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3424 + slightly improved tests from https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3509

The exact same issue happens for CharacterLiteral.charValue, for which I added a test case and a fix as well.

### DIFF
--- a/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CharacterLiteral.java
+++ b/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CharacterLiteral.java
@@ -205,7 +205,18 @@ public class CharacterLiteral extends Expression {
 	 * @exception IllegalArgumentException if the literal value cannot be converted
 	 */
 	public char charValue() {
-		Scanner scanner = this.ast.scanner;
+		// create a new local scanner to allow concurrent use
+		Scanner scanner = new Scanner(
+				this.ast.scanner.tokenizeComments,
+				this.ast.scanner.tokenizeWhiteSpace,
+				this.ast.scanner.checkNonExternalizedStringLiterals,
+				this.ast.scanner.sourceLevel,
+				this.ast.scanner.complianceLevel,
+				this.ast.scanner.taskTags,
+				this.ast.scanner.taskPriorities,
+				this.ast.scanner.isTaskCaseSensitive,
+				this.ast.scanner.previewEnabled);
+
 		char[] source = this.escapedValue.toCharArray();
 		scanner.setSource(source);
 		scanner.resetTo(0, source.length);
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTVisitorTest.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTVisitorTest.java
@@ -15,6 +15,11 @@
 package org.eclipse.jdt.core.tests.dom;
 
 import java.lang.reflect.Method;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import junit.framework.Test;
 import org.eclipse.jdt.core.dom.*;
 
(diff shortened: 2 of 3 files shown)


jukzi (approved):

mpalat:
@jukzi - @SarikaSinha  had moved out of Eclipse "officially" sometime back although she contributes at her "free" time occasionally. So wouldn't bother her unless its really required and hence removed her name from the review.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/AST.java
```diff
@@ -4003,4 +4003,28 @@ public boolean isPreviewEnabled() {
...
+
+	/**
+	 * Returns a new instance of the AST scanner with the same configuration. Since the scanner is
+	 * stateful, using a cloned copy in read methods on AST nodes (e.g. StringLiteral.getLiteralValue)
+	 * allows those read methods to be used concurrently
+	 *
+	 * @return new Scanner instance with the same config as AST.scanner
+	 * @since 3.41
+	 */
+	Scanner clonedScanner() {
```
mpalat, line 4015: @martinlippert : Would recommend not to have this addition since(1) this is more about a convenience method and not related to AST as such and (2) this class is an API and wouldn't want to change the same unless absolutely necessary.

mpalat (changes requested):
@martinlippert - As mentioned in the review comment in the AST, would not prefer this change. Since its just a couple of places - StringLiteral and CharacterLiteral, creating a new scanner from this.ast.scanner values would be better than having a convenience method at an API level. In case we need such similar changes later in other places, we can think of a convenience method in Scanner class (or a nested sub-class similar to VanguardScanner) - that would not constitute an API change since they are internal classes. So in short, would request to remove the API changes and create new Scanners at the charValue() and StringValue() methods.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/AST.java
martinlippert: @mpalat @jukzi Sure, no problem, I pushed an additional commit that moves the creation of the new scanner into `StringLiteral.getLiteralValue` and `CharacterLiteral.charValue`.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CharacterLiteral.java
```diff
@@ -205,7 +205,18 @@ void internalSetEscapedValue(String value) {
 	 * @exception IllegalArgumentException if the literal value cannot be converted
 	 */
 	public char charValue() {
-		Scanner scanner = this.ast.clonedScanner();
+		// create a new local scanner to allow concurrent use
```
jukzi, line 208: better make it reusable with a org.eclipse.jdt.internal.compiler.parser.Scanner.clone()?

jukzi (approved):
LGTM and no new API

martinlippert:
And thanks a lot @jukzi and @mpalat for your quick feedback and the reviews here. Really awesome. And very much appreciated!!!

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
```diff
@@ -5906,4 +5906,18 @@ protected void growCommentInfoArrays(int length, int newLength) {
 	System.arraycopy(this.commentTagStarts, 0, this.commentTagStarts = new int[newLength], 0, length);
 }
 
+@Override
+public Scanner clone() {
+	return new Scanner(
```
szarnekow, line 5911: I think this is misleading. Scanner has subclasses that now also offer `clone()` but don't return a cloned instance of their subclass. I still prefer Manoj's proposal and inline this expression into CharLiteral and StringLiteral.

szarnekow: One alternative would be a copy constructor instead of `clone()` but that would also add new API to Scanner. Not sure it's worth it.

martinlippert: No worries, let me revert the previous commit, than we are back at inline creation of separate scanner instances.

iloveeclipse:
@martinlippert : could you please squash all 5 commits to a **single one** with **appropriate** commit message?
See https://github.com/eclipse-jdt/.github/blob/main/CONTRIBUTING.md#commit-message-recommendations

iloveeclipse:
@mpalat : could you pleas update your review? Thanks.

mpalat:
> Yepp, no problem. Last commit reverted, back at inline scanner creation now.
Thanks @martinlippert !

mpalat (approved):

Issue #3418, opened 2024-12-09
org.eclipse.jdt.internal.core.search.matching.MatchLocator.bindings may be wrong


jukzi:
The type unsafe SimpleLookupTable may hide an error: char[] are added as keys, but it is tried to removed keys of type SearchPattern.

i will investigate.

PR #3522, opened 2025-01-06
fix search.matching.MatchLocator.bindings #3418

jukzi:
A map with keys char[] did only find keys by identity (but not by equal characters as needed).

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3418

tested by RunJavaSearchTests


iloveeclipse:
Wouldn't this cause more memory pressure and more GC work by creating new Strings, how often is this code used and in which context? Is this index related? I haven't checked this in the IDE.

jukzi:
This PR can even reduce memory pressure (and computational effort) as equal char[] are not stored multiple times in the Map. Not only are equal keys avoided but also their TypeBinding values.
Take for example org.eclipse.jdt.core.tests.model.JavaSearchGenericTypeTests.testParameterizedArrayTypeMultipleArguments02()
where "Ljava.lang.String" was stored multiple times in MatchLocator.bindings:
image
The "cache" just did not cache anything.

Issue #3516, opened 2025-01-05
CCE while performing Java Search


ssttaarr:
Getting ar error while searching for occurrences of a type `com.google.gson.Gson` (Ctrl-Shift-G)  
```
!ENTRY org.eclipse.core.jobs 4 2 2025-01-05 18:27:21.571
!MESSAGE An internal error occurred during: "Java Search".
!STACK 0
java.lang.ClassCastException: class org.eclipse.jdt.internal.compiler.lookup.ProblemReferenceBinding cannot be cast to class org.eclipse.jdt.internal.compiler.lookup.BinaryTypeBinding (org.eclipse.jdt.internal.compiler.lookup.ProblemReferenceBinding and org.eclipse.jdt.internal.compiler.lookup.BinaryTypeBinding are in unnamed module of loader org.eclipse.osgi.internal.loader.EquinoxClassLoader @27342d64)
        at org.eclipse.jdt.internal.compiler.lookup.PackageBinding.initDefaultNullness(PackageBinding.java:330)
        at org.eclipse.jdt.internal.compiler.lookup.PackageBinding.getDefaultNullness(PackageBinding.java:338)
        at org.eclipse.jdt.internal.compiler.lookup.BinaryTypeBinding.scanTypeForNullDefaultAnnotation(BinaryTypeBinding.java:2227)
        at org.eclipse.jdt.internal.compiler.lookup.BinaryTypeBinding.cachePartsFrom(BinaryTypeBinding.java:467)
        at org.eclipse.jdt.internal.compiler.lookup.LookupEnvironment.createBinaryTypeFrom(LookupEnvironment.java:1093)
        at org.eclipse.jdt.internal.compiler.lookup.LookupEnvironment.cacheBinaryType(LookupEnvironment.java:550)
        at org.eclipse.jdt.internal.compiler.lookup.LookupEnvironment.cacheBinaryType(LookupEnvironment.java:537)
        at org.eclipse.jdt.internal.core.search.matching.MatchLocator.cacheBinaryType(MatchLocator.java:451)
        at org.eclipse.jdt.internal.core.search.matching.ClassFileMatchLocator.locateMatches(ClassFileMatchLocator.java:269)
        at org.eclipse.jdt.internal.core.search.matching.MatchLocator.process(MatchLocator.java:1982)
        at org.eclipse.jdt.internal.core.search.matching.MatchLocator.locateMatches(MatchLocator.java:1351)
[...]

PR #3528, opened 2025-01-07
BinaryTypeBinding: enrich RuntimeException with filename

jukzi:
During internal errors it is not clear which file did contain the problematic binary. For example:

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3516

Example output will be like: "java.lang.RuntimeException: RuntimeException caching =P/libGh375.jar|TestGh375.class"


stephan-herrmann:
If this is just for investigating #3516 and will be removed afterwards, I have no objections.

If, however, this is intended to stay, wouldn't it be better to wrap the exception in or near BTB.cachePartsFrom()? That would benefit more scenarios, no?

As for the exception message: it's not really the caching that creates problems, but loading the binary type does :)

jukzi:
> If, however, this is intended to stay, wouldn't it be better to wrap the exception in or near BTB.cachePartsFrom()? That would benefit more scenarios, no?

I would like to let it stay as similar error may appear in the future. I moved it to BTB as requested and changed the text to "loading". ok?

PR #3533, opened 2025-01-08
Make AST map pre 1.8 versions to 1.8

akurtakov:
As the compiler mandates 1.8 already it makes sense for the AST to do the same.

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverterTest2.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverterTest2.java
@@ -35,11 +35,9 @@ import org.eclipse.jdt.internal.compiler.impl.CompilerOptions;
 @SuppressWarnings({"rawtypes", "unchecked"})
 public class ASTConverterTest2 extends ConverterTestSetup {
 
-	/** @deprecated using deprecated code */
-	@Override
 	public void setUpSuite() throws Exception {
 		super.setUpSuite();
-		this.ast = AST.newAST(AST.JLS2, false);
+		this.ast = AST.newAST(AST.getAllSupportedVersions().get(0), false);
 	}
 
 	public ASTConverterTest2(String name) {
@@ -54,23 +52,7 @@ public class ASTConverterTest2 extends ConverterTestSetup {
 		return buildModelTestSuite(ASTConverterTest2.class);
 	}
 	/**
-	 * Internal access method to MethodDeclaration#thrownExceptions() for avoiding deprecated warnings.
-	 * @deprecated
-	 */
-	private static List internalThrownExceptions(MethodDeclaration methodDeclaration) {
-		return methodDeclaration.thrownExceptions();
-	}
-
-	/**
-	 * @deprecated
-	 */
-	private Type componentType(ArrayType array) {
-		return array.getComponentType();
-	}
-
-	/**
 	 * http://dev.eclipse.org/bugs/show_bug.cgi?id=22560
-	 * @deprecated using deprecated code
 	 */
 	public void test0401() throws JavaModelException {
 		ICompilationUnit sourceUnit = getCompilationUnit("Converter" , "src", "test0401", "A.java"); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$
@@ -88,7 +70,7 @@ public class ASTConverterTest2 extends ConverterTestSetup {
 		Expression expression = returnStatement.getExpression();
 		assertNotNull("there is no expression", expression); //$NON-NLS-1$
 		// call the default initialization
-		methodDeclaration.getReturnType();
+		methodDeclaration.getReturnType2();
 		ITypeBinding typeBinding = expression.resolveTypeBinding();
 		assertNotNull("No typebinding", typeBinding); //$NON-NLS-1$
 		assertEquals("wrong name", "int", typeBinding.getName()); //$NON-NLS-1$ //$NON-NLS-2$
@@ -280,7 +262,7 @@ public class ASTConverterTest2 extends ConverterTestSetup {
 		assertNotNull(node);
 		assertTrue("Not a method declaration", node.getNodeType() == ASTNode.METHOD_DECLARATION); //$NON-NLS-1$
 		MethodDeclaration methodDeclaration = (MethodDeclaration) node;
-		Type type = methodDeclaration.getReturnType();
+		Type type = methodDeclaration.getReturnType2();
 		assertTrue("Not a simple type", type.isSimpleType()); //$NON-NLS-1$
 		SimpleType simpleType = (SimpleType) type;
 		Name name = simpleType.getName();
@@ -316,7 +298,7 @@ public class ASTConverterTest2 extends ConverterTestSetup {
 			compilationUnit.accept(bindingsCollectorVisitor);
 			assertEquals("wrong number", 3, bindingsCollectorVisitor.getUnresolvedNodesSet().size()); //$NON-NLS-1$
 			Map bindingsMap = bindingsCollectorVisitor.getBindingsMap();
-			assertEquals("wrong number", 205, bindingsMap.size()); //$NON-NLS-1$
+			assertEquals("wrong number", 212, bindingsMap.size()); //$NON-NLS-1$
 			ASTNodesCollectorVisitor nodesCollector = new ASTNodesCollectorVisitor();
 			compilationUnit.accept(nodesCollector);
 			Set detachedNodes = nodesCollector.getDetachedAstNodes();
@@ -466,9 +448,9 @@ public class ASTConverterTest2 extends ConverterTestSetup {
 		assertNotNull(node);
 		assertTrue("Not a method declaration", node.getNodeType() == ASTNode.METHOD_DECLARATION); //$NON-NLS-1$
 		MethodDeclaration methodDeclaration = (MethodDeclaration) node;
-		List throwsException = internalThrownExceptions(methodDeclaration);
+		List throwsException = methodDeclaration.thrownExceptionTypes();
 		assertEquals("wrong size", 2, throwsException.size()); //$NON-NLS-1$
-		Name name = (Name) throwsException.get(0);
+		Name name = ((SimpleType)throwsException.get(0)).getName();
 		IBinding binding = name.resolveBinding();
 		assertNotNull("No binding", binding); //$NON-NLS-1$
 		assertEquals("LIOException;", binding.getKey());
@@ -476,7 +458,6 @@ public class ASTConverterTest2 extends ConverterTestSetup {
 
 	/**
 	 * http://bugs.eclipse.org/bugs/show_bug.cgi?id=23734
-	 * @deprecated using deprecated code
 	 */
 	public void test0414() throws JavaModelException {
 		ICompilationUnit sourceUnit = getCompilationUnit("Converter" , "src", "test0414", "A.java"); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$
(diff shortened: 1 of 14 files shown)


akurtakov:
@iloveeclipse What do you think?

iloveeclipse:
> @iloveeclipse What do you think?

Jenkins is not accessible, how many test fails are there? I'm not AST expert, I can only remember I've tried to use latest AST in JDT debug and that failed with tons of errors...

iloveeclipse:
> My question is more about the goal not that much about the work involved ( I know it will require quite some).

I assume working with "old" level AST for Java version we don't support anymore should be more of academic interest? I don't know if consumers like Xtext use AST at some minimal JLS level internally and expect us to keep it going forever. @szarnekow ?

> Btw, can you point me what you have tried in jdt.debug so I can use/test it with this one?

I see I've finally managed to fix it: https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/109, so that should be fine.

akurtakov:
@iloveeclipse FYI, there are no test failures, presumably because the maps changes make tests run with 1.8 level which all the tests are supposedly working.

iloveeclipse:
Cool.

akurtakov:
May I assume that if @szarnekow (or someone else ) doesn't raise serious concern this approach if fine to proceed with?

iloveeclipse:
@mpalat, @srikanth-sankaran, @jarthana , @noopur2507 : any objections?

iloveeclipse:
@akurtakov : looking at the code, you only touched the mapping. But the AST constructor still uses `JLS4_INTERNAL` , `JLS2_INTERNAL` etc and only has the check below: 

 ```
if (level < JLS2_INTERNAL && level > JLS_Latest) {
	throw new IllegalArgumentException("Unsupported JLS level : " + level); //$NON-NLS-1$
}
```
Do you plan to update this as well?

**Comment by akurtakov** [comment]
Yes, I do. Also cleaning the tests to not run old compliance e.g  jls2 compliance (https://github.com/eclipse-jdt/eclipse.jdt.core/blob/0c3986bbe9b8f48b42efe9a35aa39dc0e5f5d79f/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ConverterTestSetup.java#L621 leading to https://github.com/eclipse-jdt/eclipse.jdt.core/blob/0c3986bbe9b8f48b42efe9a35aa39dc0e5f5d79f/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ConverterTestSetup.java#L628 ) which significantly complicates, slows down and prevent the suite to be useful to verify "supported" JDT features are implemented proper (a good number of ast failures in https://ci.eclipse.org/ls/job/jdt-core-incubator/job/dom-with-javac/lastSuccessfulBuild/testReport/ ).

**Comment by jarthana** [comment]
> @mpalat, @srikanth-sankaran, @jarthana , @noopur2507 : any objections?

I have none from jdt.core pov. Still would be good to hear from @noopur2507 on how this affect the UI.

**Comment by noopur2507** [comment]
I don't see any concerns with this approach from the UI side. However, updates or cleaning up the usage of old AST levels will be needed. Also, the AST view can be updated to support only 1.8 and above.

It will be good to run the UI and Debug tests as well with this PR before the merge.

File: org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTTest.java [hunk 1]
```diff
@@ -1323,12 +1323,11 @@ private static int getApiLevel(String s) {
...
-        case JavaCore.VERSION_1_3: return AST.JLS3;
-        case JavaCore.VERSION_1_4: return AST.JLS4;
-        case JavaCore.VERSION_1_5: return AST.JLS4;
-        case JavaCore.VERSION_1_6: return AST.JLS4;
-        case JavaCore.VERSION_1_7: return AST.JLS4;
+		case JavaCore.VERSION_1_2 : return AST.JLS8;
+        case JavaCore.VERSION_1_3: return AST.JLS8;
+        case JavaCore.VERSION_1_4: return AST.JLS8;
+        case JavaCore.VERSION_1_5: return AST.JLS8;
+        case JavaCore.VERSION_1_6: return AST.JLS8;
```
srikanth-sankaran, line 1330: This may not be a correctness issue - but I see the case for VERSION_1_7 is dropped - why

akurtakov: My bad. I'll actually start an enhancement to AST (in separate PR) adding getAllJLSVersionsSupported/isSupportedJLSVersions/etc. mimicking JavaCore methods so we can have a more consistent API rather than this wild refering of constants like in https://github.com/eclipse-jdt/eclipse.jdt.ui/blob/41eae491f2720d3febbc0063ee644b01f0e5bb40/org.eclipse.jdt.astview/src/org/eclipse/jdt/astview/views/ASTView.java#L507 . And this part will be rewritten based on it.

akurtakov:
> I don't see any concerns with this approach from the UI side. However, updates or cleaning up the usage of old AST levels will be needed. Also, the AST view can be updated to support only 1.8 and above.

Here it is https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1965

akurtakov:
I consider this one ready. It would be best to land this in ASAP so there is enough time to fix tests in jdt.debug/ui if there are any problems (running them all without I-build is quite heavy task).
If there are no problems/objections my intention is to merge it next Monday(Feb 3rd).

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverterTest2.java
```diff
@@ -4070,7 +4010,8 @@ public void test0534() throws JavaModelException {
 		assertNotNull("No node", node);
 		assertTrue("not a field declaration", node.getNodeType() == ASTNode.FIELD_DECLARATION);
 		FieldDeclaration declaration = (FieldDeclaration) node;
-		ASTParser parser = ASTParser.newParser(AST.JLS2);
+		@SuppressWarnings("deprecation")
+		ASTParser parser = ASTParser.newParser(AST.JLS8);
```
iloveeclipse, line 4014: I wonder if we should use `AST.getAllSupportedVersions().get(0)` instead of `AST.JLS8`?

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTTest.java
```diff
@@ -1312,46 +1309,31 @@ void genericPropertyListTest(ASTNode node, List children, Property prop) {
...
-        case JavaCore.VERSION_20: return AST.JLS20;
-        default:  return AST.JLS2;
-		}
+        try {
+            int apiLevel = Integer.valueOf(s);
+            if (AST.isSupportedVersion(apiLevel)) {
+                return apiLevel;
+            }
+        } catch (NumberFormatException e) {
+            if (s.equals(JavaCore.VERSION_1_8)) {
```
iloveeclipse, line 1318: Wouldn't be better to have a dedicated `equals(JavaCore.VERSION_1_8)` check before and hard fail on anything that throws NumberFormatException? Or are there still too many 1.x users in the tests?

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/AST.java
```diff
@@ -1234,8 +1208,6 @@ private AST(int level, boolean previewEnabled) {
 	 *    indicates the api level and source compatibility mode (as per <code>JavaCore</code>) - defaults to 1.3
```
iloveeclipse, line 1208: `(as per <code>JavaCore</code>) - defaults to 1.3` should be updated now

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTTest.java
akurtakov: Done.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverterTest2.java
akurtakov: Done. In a whole lot of other places. There is a lot to be desired in how jls is set in tests which I intend to do as follow up commits to not make this one huge.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverterTest2.java
```diff
@@ -4045,7 +3983,8 @@ public void test0533() throws JavaModelException {
 		assertNotNull("No node", node);
 		assertTrue("not a method declaration", node.getNodeType() == ASTNode.METHOD_DECLARATION);
 		MethodDeclaration declaration = (MethodDeclaration) node;
-		ASTParser parser = ASTParser.newParser(AST.JLS2);
+		@SuppressWarnings("deprecation")
+		ASTParser parser = ASTParser.newParser(AST.JLS8);
```
iloveeclipse, line 3987: `AST.JLS8` left over

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTTest.java
```diff
@@ -1312,46 +1309,32 @@ void genericPropertyListTest(ASTNode node, List children, Property prop) {
...
-        case JavaCore.VERSION_17: return AST.JLS17;
-        case JavaCore.VERSION_18: return AST.JLS18;
-        case JavaCore.VERSION_19: return AST.JLS19;
-        case JavaCore.VERSION_20: return AST.JLS20;
-        default:  return AST.JLS2;
-		}
+		if (s.equals(JavaCore.VERSION_1_8)) {
+            return AST.JLS8;
+        }
+		try {
```
iloveeclipse, line 1315: indentation

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTTest.java
```diff
@@ -1312,46 +1309,32 @@ void genericPropertyListTest(ASTNode node, List children, Property prop) {
...
-        case JavaCore.VERSION_14: return AST.JLS14;
-        case JavaCore.VERSION_15: return AST.JLS15;
-        case JavaCore.VERSION_16: return AST.JLS16;
-        case JavaCore.VERSION_17: return AST.JLS17;
-        case JavaCore.VERSION_18: return AST.JLS18;
-        case JavaCore.VERSION_19: return AST.JLS19;
-        case JavaCore.VERSION_20: return AST.JLS20;
-        default:  return AST.JLS2;
-		}
+		if (s.equals(JavaCore.VERSION_1_8)) {
```
iloveeclipse, line 1312: indentation

iloveeclipse:
I have no objections to merge.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTTest.java
akurtakov: fixed

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTTest.java
akurtakov: fixed

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverterTest2.java
akurtakov: Done.

akurtakov:
Merging this one and starting new I-build to verify jdt.ui/debug and pde test results.

PR #3538, opened 2025-01-09
Limit scope of JavaModelManager#rootPathToRawEntries field

laeubi:
Currently rootPathToRawEntries is a public field and as those can possibly written/modified outside of control of the containing class.

This now makes the field private and adds an accessory method for its consumers removing the need for some boilerplate checks as well.

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java
@@ -1305,7 +1305,7 @@ public class JavaModelManager implements ISaveParticipant, IContentTypeChangeLis
 		public volatile boolean writtingRawClasspath;
 		public volatile IClasspathEntry[] resolvedClasspath;
 		public volatile IJavaModelStatus unresolvedEntryStatus;
-		public volatile Map<IPath, IClasspathEntry> rootPathToRawEntries; // reverse map from a package fragment root's path to the raw entry
+		private volatile Map<IPath, IClasspathEntry> rootPathToRawEntries; // reverse map from a package fragment root's path to the raw entry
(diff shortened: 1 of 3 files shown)


jukzi (commented):
I have bad experience with using Optional in business code when null could be used instead. Making something more private sounds good. (but unimportant, so that such refactoring has to outweigh the changes in git history)

laeubi:
> I have bad experience with using Optional in business code when null could be used instead

I don't really understand what this should mean, I use `Optional` quite often and never seen any bad effects, and don't see any problem here beside it is more expressive and flexible use.

jarthana:
Considering what happened despite review from several people including myself, I will feel lot comfortable if we can run all the IDE tests once with the patch. Is there a way we can do it before merging this change?

laeubi:
Given the complex infrastructure and resource demands for i-builds I don't see any obvious option for this.

jukzi:
> I don't really understand what this should mean, I use `Optional` quite often and never seen any bad effects,

Most time it adds boiler plate code to wrap and unwrap null and additional memory allocation. In other contexts colleagues also stumbled upon that it is not well handled by libraries that serialize or persist values. Google "java Optional gone wrong" gives a lot of hits.

laeubi:
> Most time it adds boiler plate code to wrap and unwrap null and additional memory allocation. 

Optional is a [value-based](https://docs.oracle.com/javase/8/docs/api/java/lang/doc-files/ValueBased.html) class so it does not require any additional memory allocation. 

> In other contexts colleagues also stumbled upon that it is not well handled by libraries that serialize or persist values

I don't see we need to persist anything here.

> Google "java Optional gone wrong" gives a lot of hits.

Google is not really a reference you will get a "reference" for anything you want to "proof" these days in either direction ...

iloveeclipse (changes requested):
The general idea (hide filed, provide getters) is OK, but from the code style it simply doesn't look consistent to the rest of the surrounding code. 

So I would prefer to skip the "Optional" part of the PR and keep the rest.

Issue #1834, opened 2024-01-10
silent NumberFormatException in SourceMapper


jukzi:
for example java.lang.NumberFormatException `For input string: "1.class"` during quick type hierarchy for java.lang.Object
```
Stack Trace	Count	Percentage
NumberFormatException java.lang.NumberFormatException.forInputString(String, int):67	747	100 %
int java.lang.Integer.parseInt(String, int):661	747	100 %
int java.lang.Integer.parseInt(String):777	747	100 %
ISourceRange org.eclipse.jdt.internal.core.SourceMapper.mapSource(NamedMember, char[], IBinaryType, IJavaElement):1597	747	100 %
void org.eclipse.jdt.internal.core.SourceMapper.mapSource(NamedMember, char[], IBinaryType):1537	747	100 %
IBuffer org.eclipse.jdt.internal.core.ClassFile.mapSource(SourceMapper, IBinaryType, IClassFile):502	747	100 %
IBuffer org.eclipse.jdt.internal.core.ClassFile.openBuffer(IProgressMonitor, Object):478	747	100 %
IBuffer org.eclipse.jdt.internal.core.Openable.getBuffer():300	747	100 %
IBuffer org.eclipse.jdt.internal.core.AbstractClassFile.getBuffer():245	747	100 %
ISourceRange org.eclipse.jdt.internal.core.BinaryMember.getSourceRange():146	747	100 %
void org.eclipse.jdt.internal.core.manipulation.JavaElementLabelComposerCore.appendTypeLabel(IType, long):991	747	100 %
void org.eclipse.jdt.internal.core.manipulation.JavaElementLabelComposerCore.appendElementLabel(IJavaElement, long):189	747	100 %
void org.eclipse.jdt.ui.JavaElementLabels.getElementLabel(IJavaElement, long, StyledString):545	747	100 %
StyledString org.eclipse.jdt.ui.JavaElementLabels.getStyledElementLabel(IJavaElement, long):520	747	100 %
StyledString org.eclipse.jdt.ui.JavaElementLabels.getStyledTextLabel(Object, long):448	747	100 %
StyledString org.eclipse.jdt.internal.ui.viewsupport.JavaUILabelProvider.getStyledText(Object):176	747	100 %
StyledString org.eclipse.jface.viewers.DelegatingStyledCellLabelProvider.getStyledText(Object):209	747	100 %
[...]

PR #3543, opened 2025-01-09
fix silent NumberFormatException in SourceMapper #1834

jukzi:
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1834

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/SourceMapper.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/SourceMapper.java
@@ -1568,12 +1568,16 @@ public class SourceMapper
 
 				char[] fullName = info.getName();
 				if (isAnonymousClass) {
-					String eltName = this.binaryTypeOrModule.getParent().getElementName();
-					eltName = eltName.substring(eltName.lastIndexOf('$') + 1, eltName.length());
+					String eName = this.binaryTypeOrModule.getParent().getElementName();
+					String eltName = eName.substring(eName.lastIndexOf('$') + 1, eName.length());
+					if (eltName.endsWith(".class")) { //$NON-NLS-1$
+						eltName = eltName.substring(0, eltName.length() - ".class".length()); //$NON-NLS-1$
+					}
 					try {
 						this.anonymousClassName = Integer.parseInt(eltName);
 					} catch(NumberFormatException e) {
-						// ignore
+						throw new RuntimeException("Failed to parse anonymous class name '" //$NON-NLS-1$
+								+ eName + "'", e); //$NON-NLS-1$
 					}
 				}
 				doFullParse = hasToRetrieveSourceRangesForLocalClass(fullName);


jukzi:
List of tests where the change is executed:
```
 org.eclipse.jdt.core.tests.dom.ASTModelBridgeTests.testBinaryMemberTypeFromAnonymousClassFile1
 org.eclipse.jdt.core.tests.dom.ASTModelBridgeTests.testBinaryMemberTypeFromAnonymousClassFile2
 org.eclipse.jdt.core.tests.model.AttachSourceTests.testInnerClass2
 org.eclipse.jdt.core.tests.model.AttachSourceTests.testInnerClass3
 org.eclipse.jdt.core.tests.model.AttachSourceTests.testInnerClass4
 org.eclipse.jdt.core.tests.model.AttachSourceTests.testInnerClass7
 org.eclipse.jdt.core.tests.model.JavaSearchTests.testLocalVariableReference3
 org.eclipse.jdt.core.tests.model.ResolveTests.testLocalNameForClassFile
 org.eclipse.jdt.core.tests.model.ResolveTests.testMethodDeclarationInAnonymous4
 org.eclipse.jdt.core.tests.model.TypeResolveTests.testResolveTypeInBinary3
```

stephan-herrmann:
Number-parsing part of the name of a class*file* looks wrong, indeed, but let's dig a bit deeper:
* while it's true that I made changes in this area via https://bugs.eclipse.org/521058 this didn't touch the case of anonymous classes, but only `module-info.class`
* the block in question ..
   * was introduced via https://bugs.eclipse.org/3208 (the oldest JDT bug I ever looked at :) )
   * got its handling for anonymous classes via https://bugs.eclipse.org/75885
* the same https://bugs.eclipse.org/75885 also made `SourceMapper.IType getType(String)` fit for anonymous classes
* given that the block in `mapSource(NamedMember,char[],IBinaryType,IJavaElement)` has been wrong for more than 20 years, what purpose would it serve anyway?
* I couldn't find any test failing when I simply remove the block `if (isAnonymousClass)` :smile:

So I suggest not to fix unnecessary code, but let's give @jarthana  also a chance to chime in, if he has an idea what this code might be needed for.

stephan-herrmann:
One more observation: with or without the block in question, `SourceMapper.anonymousClassName` will always be `0`.

Consider this block in `enterType()`:

```java
			this.anonymousCounter++;
			if (this.anonymousCounter == this.anonymousClassName) {
				this.types[this.typeDepth] = getType(this.binaryTypeOrModule.getElementName());
			} else {
				this.types[this.typeDepth] = getType(DeduplicationUtil.toString(typeInfo.name));
			}
```

if `ananymousClassName` is always `0`, the then-branch of the above `if` will never be triggered, i.e., we always retrieve the type  using an empty name. To check:
* do we have tests challenging this code with more than one anonymous class (to be sure that int parsing is actually tested)?
* does `getType("")` actually supersede (since [bug 75885](https://bugs.eclipse.org/75885)) the dead then-block?
* perhaps we can completely remove field `anonymousClassName` and all code accessing it?

jukzi:
> * do we have tests challenging this code with more than one anonymous class (to be sure that int parsing is actually tested)?

AttachSourceTests.testInnerClass3 did evaluate the number to "2"

jukzi:
@stephan-herrmann what do we need to continue?

stephan-herrmann:
> @stephan-herrmann what do we need to continue?

Do we have any motivation to keep that code section?

See my hint:
> perhaps we can completely remove field anonymousClassName and all code accessing it?

Debugging AttachSourceTests.testInnerClass3 I see the field used only to distinguish between using "" or "" as the classname :)  -- The name of the classfile will always be determined inside `getType()`.

stephan-herrmann:
> perhaps we can completely remove field anonymousClassName and all code accessing it?

RunModelTests confirms my theory.

jukzi:
> perhaps we can completely remove field anonymousClassName and all code accessing it?

I created a new PR for that: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3583

PR #3545, opened 2025-01-09
Fire CHange events of an operation in a new job

laeubi:
Currently events are fired in sync with the operation, this can lead to code executed as part of an operation.

This now fire events in an own job.

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/DeltaProcessor.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/DeltaProcessor.java
@@ -38,6 +38,7 @@ import org.eclipse.core.runtime.OperationCanceledException;
 import org.eclipse.core.runtime.Path;
 import org.eclipse.core.runtime.PerformanceStats;
 import org.eclipse.core.runtime.SafeRunner;
+import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.jdt.core.*;
 import org.eclipse.jdt.core.compiler.CharOperation;
 import org.eclipse.jdt.internal.compiler.SourceElementParser;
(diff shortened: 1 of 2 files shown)


org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/DeltaProcessor.java
```diff
@@ -1541,14 +1542,15 @@ public void fire(IJavaElementDelta customDelta, int eventType) {
...
-		}
+		Job.create("Fire Change", m->{ //$NON-NLS-1$
+			switch (eventType) {
+				case DEFAULT_CHANGE_EVENT:
+				case ElementChangedEvent.POST_CHANGE:
+					firePostChangeDelta(deltaToNotify, listeners, listenerMask, listenerCount);
+					fireReconcileDelta(listeners, listenerMask, listenerCount);
+					break;
+			}
+		}).schedule();
```
iloveeclipse, line 1553: This is a "no go". There will be N jobs running in arbitrary order after this change.

iloveeclipse (changes requested):

laeubi:
> > Currently events are fired in sync with the operation, this can lead to code executed as part of an operation.
> 
> Might be that **was** the intent?

I have no clue, javadoc says nothing about what one should expect here:

> Notifies that one or more attributes of one or more Java elements have changed.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/DeltaProcessor.java
szarnekow: Given that `flush` is called on the `this` instance from within `firePostChangeDelta` I feel that this will break all sorts of invariants. Random async changes to a code base that is not designed for asynchronous processing is not trivial.

laeubi:
Does not seem to work very well...

Issue #3116, opened 2024-10-18
[23] Flow analysis ignores assigment in pre constructor context


jarthana:
Consider this code:
```
class X {
    final int final_field;
    int x;
    { x = final_field; } // Error: The blank final field final_field may not have been initialized
    X() {
        final_field = -1;
        super();
    }
}
```

Looks like the assignment in the constructor before super() call has no effect on flow analysis.

Copying @mpalat and @stephan-herrmann

stephan-herrmann:
Resolved by https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3557

PR #3557, opened 2025-01-14
[23] Flow analysis ignores assigment in pre constructor context

stephan-herrmann:
+ analyse ctor statements in two segments (if needed)
  - this allows field initializers to see inits from ctor prologues
+ generate field inits only after explicit ctor (if prologue present)

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3116


stephan-herrmann:
@srikanth-sankaran The change proposed here, looks a bit invasive. Let me know if you want to weigh in on the design.

The essence is: both analyseCode and generateCode need to interleave constructor prologues (before the explicit constructor call) and field initializers and then the rest of constructor bodies in funny ways.
* during analyseCode we need to accumulate the prologue info from all constructors
* during generatedCode each constructor can handle the interleaving on its own.

Corresponding changes to the rules of definite assignment can be found in https://cr.openjdk.org/~iris/se/24/spec/draft/java-se-24-draft-spec-29/specs/flexible-constructor-bodies-jls.html#jls-16

Note: there might still be fine points not yet covered by this change ...

Issue #3298, opened 2024-11-12
"Missing System Library" exception in ASTParser


ptziegler:
This seems to be a regression introduced by https://github.com/eclipse-jdt/eclipse.jdt.core/commit/74d062b38d3003638127127cc5358dd9f79df0f4

To reproduce: Make sure compiler compliance is set to > 1.8 and open a Java file from a project without Java nature.

![image](https://github.com/user-attachments/assets/5c8a6187-1d02-4534-832a-175b0b7bf02d)

Stack Trace:
```
java.lang.IllegalStateException: Missing system library
	at org.eclipse.jdt.core.dom.ASTParser.checkForSystemLibrary(ASTParser.java:311)
	at org.eclipse.jdt.core.dom.ASTParser.getClasspath(ASTParser.java:272)
	at org.eclipse.jdt.core.dom.ASTParser.internalCreateASTCached(ASTParser.java:1299)
	at org.eclipse.jdt.core.dom.ASTParser.lambda$1(ASTParser.java:1178)
	at org.eclipse.jdt.internal.core.JavaModelManager.cacheZipFiles(JavaModelManager.java:5694)
	at org.eclipse.jdt.core.dom.ASTParser.internalCreateAST(ASTParser.java:1178)
	at org.eclipse.jdt.core.dom.ASTParser.createAST(ASTParser.java:918)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider$1.run(CoreASTProvider.java:294)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:47)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.createAST(CoreASTProvider.java:286)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:199)
	at org.eclipse.jdt.core.manipulation.CoreASTProvider.getAST(CoreASTProvider.java:189)
	at org.eclipse.jdt.core.manipulation.SharedASTProviderCore.getAST(SharedASTProviderCore.java:138)
	at org.eclipse.jdt.internal.ui.viewsupport.SelectionListenerWithASTManager$PartListenerGroup.calculateASTandInform(SelectionListenerWithASTManager.java:167)
	at org.eclipse.jdt.internal.ui.viewsupport.SelectionListenerWithASTManager$PartListenerGroup$1.lambda$0(SelectionListenerWithASTManager.java:149)
[...]

PR #3560, opened 2025-01-15
Restrict system library check to require Java nature

ptziegler:
The validation of the project classpath introduced with 74d062b38d3003638127127cc5358dd9f79df0f4 always fails, if the project doesn't have a Java nature (and thus usually no .classpath file).

This causes issues when e.g. trying to open a source file with the Java editor or simply when trying to use the AST parser outside of a Java project. With this change, the check becomes more conservative and only fails, if both the Java nature exists and no system library is configured.

Note that one can avoid this exception by disabling the binding resolution. But given that this requires changes in user code, it is more prudent to only fail in cases where this is the desired behavior.

Resolves https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3298

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
+++ b/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
@@ -272,6 +272,9 @@ public class ASTParser {
 	}
 
 	private void checkForSystemLibrary(List<Classpath> allClasspaths) {
+		if (!hasJavaNature()) {
+			return;
+		}
 		boolean hasSystemLibrary = true; // default for 1.8 setting without a valid project
 		boolean hasModule = false;
 		Throwable exception = null;
@@ -1655,5 +1658,9 @@ public class ASTParser {
 					compilationUnit.types().add(typeDeclaration);
 				}
 		}
+	}
+
+	private boolean hasJavaNature() {
+		return this.project == null || JavaProject.hasJavaNature(this.project.getProject());
 	}
 }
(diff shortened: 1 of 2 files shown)


ptziegler:
I've added an additional test case that also checks the situation where the AST parser is used without any project.

Note that this currently causes `testGH3047()` to fail because the source file is parsed without the Java project being set (and also because no requestor is set). This hasn't thrown an exception up until now and therefore, at least from my perspective, shouldn't.

ptziegler:
After looking closer _testGH3047()_ and _testGH3047_2()_ are effectively copies of one another, except that the latter is passing the `IJavaProject` to the ASTParser. The problem is that _testGH3047()_ is effectively the setup we have in our application and what's causing the regression.

ptziegler:
Any chance this can make it into the M3? I've adapted the code to only avoid the exception when opening the Java editor in a non-Java project.

There is still the issue that you can no longer resolve method bindings when parsing Java source code (without a project), which after 74d062b38d3003638127127cc5358dd9f79df0f4 is the now expected behavior.

We have disabled binding resolution for our application. So at least for me, this is no longer a real concern.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/ASTParser.java
```diff
@@ -276,7 +276,7 @@ private void checkForSystemLibrary(List<Classpath> allClasspaths) {
 		boolean hasModule = false;
 		Throwable exception = null;
 		String compliance = this.compilerOptions.get(JavaCore.COMPILER_COMPLIANCE);
-		if (CompilerOptions.versionToJdkLevel(compliance) >= ClassFileConstants.JDK9) {
+		if (CompilerOptions.versionToJdkLevel(compliance) >= ClassFileConstants.JDK9 && hasJavaNature()) {
```
iloveeclipse, line 279: May I ask you to move this check before this if and make it independent on Java version?
```
if(! hasJavaNature()){
return;
}
```
So it is clear that we don't want to resolve anything independently on Java version.
Currently the code reads as if we would support Java 8 without Java project, which is not correct.

ptziegler: Makes sense. I've just updated the PR.

iloveeclipse: Thanks.

PR #3561, opened 2025-01-16
Don't evaluate the same type twice when renaming a method

fedejeanne:
Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1945

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/hierarchy/HierarchyResolver.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/hierarchy/HierarchyResolver.java
@@ -29,6 +29,7 @@ package org.eclipse.jdt.internal.core.hierarchy;
 
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.IdentityHashMap;
 import java.util.Map;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IResource;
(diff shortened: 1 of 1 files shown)


jukzi (changes requested):
Theoretically the PR looks good, but unfortunatly ReferenceBinding may have unexpected hash/equals contract, see https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3412. It might work to use a IdentityHashSet instead.

fedejeanne:
`IdentityHashSet` is a test class and moving it to a production plugin might be an overkill. I emulated the behavior in the internal `Wrapper` class in my 2nd commit (which should be squashed before merging).

WDYT @jukzi ?

jukzi:
> `IdentityHashSet` is a test clas

use java.util.IdentityHashMap or org.eclipse.jdt.internal.compiler.lookup.ReferenceBindingSetWrapper

fedejeanne:
Thank you for the hint! I used `java.util.IdentityHashMap`

jukzi (approved):

fedejeanne:
I found these:
- `org.eclipse.jdt.core.tests.model.TypeHierarchyTests`
- `org.eclipse.jdt.ui.tests.core.TypeHierarchyTest`
- `org.eclipse.jdt.core.tests.model.TypeHierarchyNotificationTests`

They run through the code, they run fast and they all pass.

There may be more (I stopped the search after these 3 findings).

jukzi:
ok, org.eclipse.jdt.core.tests.model\workspace\TypeHierarchy\src\p2\I1.java is extended by both I3 and X, and Y extends both X and I3, which forms a type hierarchy with redundant superinterfaces.

PR #3567, opened 2025-01-17
Add AST getAll[Supported]Versions and isSupportedVersion

akurtakov:
New methods to ease finding all/supported JLS versions. Supported JLS versions helpers are needed in order ease limiting AST support to what compiler supports.

### DIFF
--- a/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/AST.java
+++ b/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/AST.java
@@ -558,6 +558,17 @@ public final class AST {
 	@Deprecated
 	public static final int JLS_Latest = JLS_INTERNAL_Latest;
 
+	private static final List<Integer> ALL_VERSIONS = List.of(JLS2, JLS3, JLS4, JLS8, JLS9, JLS10, JLS11, JLS12, JLS13, JLS14, JLS15, JLS16, JLS17, JLS18, JLS19, JLS20, JLS21, JLS22, JLS23);
+	private static final List<Integer> UNSUPPORTED_VERSIONS = List.of(JLS2, JLS3, JLS4);
+	private static final List<Integer> SUPPORTED_VERSIONS;
+	static {
+		List<Integer> temp = new ArrayList<>();
+		temp.addAll(ALL_VERSIONS);
+		temp.removeAll(UNSUPPORTED_VERSIONS);
+		SUPPORTED_VERSIONS = Collections.unmodifiableList(temp);
+	}
+
+
 	/*
 	 * Must not collide with a value for ICompilationUnit constants
 	 */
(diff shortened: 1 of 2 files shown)


akurtakov:
https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1948 is an example of usage of new API.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/AST.java
```diff
@@ -1193,7 +1204,7 @@ private AST(int level, boolean previewEnabled) {
 						false/*isPreviewEnabled*/);
 				break;
 			default:
-				if (level < JLS2_INTERNAL && level > JLS_Latest) {
+				if (!ALL_VERSIONS.contains(level)) {
```
iloveeclipse, line 1207: Should it be `SUPPORTED_VERSIONS` ?

akurtakov: Using SUPPORTED_VERSIONS should happen in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3533 together with switching the mapping to prevent inconsistencies.

iloveeclipse (approved):
LGTM, but I've never worked within AST area, so no idea what is supposed to be supported or not. So would be nice if someone familiar with AST could review too,

akurtakov:
I'll wait Monday and push it if no one jumps in.

stephan-herrmann:
I found this only now. 

@akurtakov this adds to the set of locations that need to be touched whenever a new Java version is supported. We already have `org.eclipse.jdt.core.dom.AST.JLS_INTERNAL_Latest` and `getJLSLatest()` which should encapsulate the moving target. Why do we now need another explicit list?

akurtakov:
@stephan-herrmann Here yes but e.g. in https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1966 it removes one such place (the goal is to have jdt.ui/debug need less or even no changes for adding constants).  
Supported versions is also used in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3533 and simplifies AST constructor and removes(automates) one more place in ASTTest(which was stuck to Java 20 !). Overall, for the JDT project it would be a decrease.

stephan-herrmann:
> @stephan-herrmann Here yes but e.g. in [eclipse-jdt/eclipse.jdt.ui#1966](https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/1966) it removes one such place (the goal is to have jdt.ui/debug need less or even no changes for adding constants). Supported versions is also used in #3533 and simplifies AST constructor and removes(automates) one more place in ASTTest(which was stuck to Java 20 !). Overall, for the JDT project it would be a decrease.

I didn't argue against providing this API, I only wonder why it needs to be implemented using an explicit list.

akurtakov:
My reasons were:
* match JavaCore SUPPORTED_VERSION implementation so codebase has a more unified way of handling similar problems
* keep possibility  to e.g. keep only LTS versions support (unlikely but still) e.g at some point 8, 11, 17, 18....n are supported.

stephan-herrmann:
> My reasons were: ...

Thanks. 

I had hoped we could do with less duplication, but for now I added the new locations needing updates to https://github.com/eclipse-jdt/eclipse.jdt.core/wiki/Adding-support-for-new-Java-version-and-Features-in-JDT  (also `JavaCore.allVersions` was missing) . It's a pity that this list seems to be growing monotonously, never shrinking.

PR #3568, opened 2025-01-17
do not call isPackage recursive on subdirectories

jukzi:
java packages are not hierarchical


jukzi:
isPackage() is used in a hierarchical way here :-(

PR #3580, opened 2025-01-20
Jenkinsfile: add error stacktrace

jukzi:
if any upon
"org.eclipse.jdt.core.compiler.batch: Execute ApiApplication failed"


jukzi:
for example during errors like https://github.com/eclipse-pde/eclipse.pde/issues/1575

PR #3581, opened 2025-01-20
Replace some colossal strings in the compiler tests

datho7561:
There are some very large string literals in the compiler tests. These string literals make the code hard to read, since you have to scroll down thousands of lines to see the rest of the test case. Since these strings are very repetitive, they can be easily generated with a `StringBuilder` and for loops.

After this change, we will still be testing that ECJ can compile colossal strings, since there is a test case for that.

## What it does
Replace some colossal string literals in the compiler tests with strings built using `StringBuilder` and for loops.

## How to test
Run the compiler test suite. The tests should compile, run, and pass.

## Author checklist


datho7561:
One of the test cases had imbalanced brackets; I think I've fixed it now but I want to see the test results first but I'll wait for CI before marking this as ready for review.

rgrunber (commented):
Overall looks good. There's nothing that explicitly states the file being maintained must also be "very large", so as long as the testcase is still testing a large source file, it should be fine.

Just wondering why you left out some cases of `test00NM` in `XLargeTest`. Not that halving the size of the source file isn't great.

datho7561:
> Just wondering why you left out some cases

I didn't have a good reason; I'll look into reducing the size of some of them right now.

datho7561:
I've replaced the strings that are easy to build with for loops. There are still some large strings, but they would be difficult or impossible to generate with a simple loop.

rgrunber (approved):
Looks good. Confirmed tests all pass and did a basic inspection of the duplicated lines.

PR #3584, opened 2025-01-21
Remove custom ReferenceBinding.hashCode()

jukzi:
To avoid possible hash collisions.

Instead of manipulating the existing key in a map remove exiting entry and put the updated entry.
Use a HashMap instead of SimpleLookupTable.


jukzi:
i did not get it running

PR #3586, opened 2025-01-21
Move the bundles that depend on jdt.debug to JavaSE-21 #3574

### LINKED ISSUE #3574: Move the bundles that depend on jdt.debug to JavaSE-21
There are some test bundles that depend on org.eclipse.jdt.debug that has already moved to JavaSE-21. These test bundles should now be moved too to get rid of the warnings. I just took a glance and looks like there's quite a bit of deprecation that happened between 17 and 21. And it might not be possible to fix all the relevant code for this release, esp. those that use `System.runFinalization()`. In some cases, we might opt to fix the warnings with a `@SuppressWarning`, but I see some straight forward ones too.

jarthana:
## What it does
Fixes #3574

## Author checklist


jarthana:
Just a note that jdt.core and compiler.batch will still remain at 17, especially the latter since it is also the standalone compiler.

PR #3587, opened 2025-01-21
Adjust tests that expect jmod files to be part of all JDKs

### LINKED ISSUE #3549: Adjust tests that expect jmod files to be part of all JDKs
> > @jarthana Java9ElementTests.test515342b (and possibly others) explicitly tries to read _jdk_/jmods/java.base.jmod, but the temurin JDK used during production builds does not have any jmods.
> > https://bugs.eclipse.org/bugs/show_bug.cgi?id=515342 doesn't mention jmods, so do you by any chance remember (after more than 7 years ...) why the test was written like this?
> 
> This may have been some leftover from the effort of [bug 490103](https://bugs.eclipse.org/bugs/show_bug.cgi?id=490103) that tried to get rid of the bogus jrt-fs.jar that represents the JRE container. But in the end, we decided it depended too much on the implementation details of the JDK.

Trying to refresh memory of old discussions I came across https://bugs.eclipse.org/bugs/show_bug.cgi?id=490103#c65

Do you think we should remove the implementation for jmod support or only disable / remove those tests?

_Originally posted by @stephan-herrmann in https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3228#issuecomment-2582072416_

jarthana:
## What it does
Fixes #3549

## Author checklist


jarthana:
After another look at the test failures, I am having second thoughts about this. Oracle still recognizes this format, specifies as one of the valid formats to be part of the classpath and even provides utilities to create a jmod file.

Issue #1907, opened 2024-01-25
[Patterns] Withdraw support for record pattern element variable in enhanced for loop


srikanth-sankaran:
https://openjdk.org/jeps/440 withdrew support for record patterns to appear in the header of an enhanced for statement - ECJ has left in the code added for this during preview phases. This is just code clutter and would be better removed. Should the feature get reintroduced at a later time, we can always restore the portions of code that would make sense in its new avatar.

PR #3595, opened 2025-01-22
Rewrite for-each pattern variable completion tests to use regular for

### LINKED ISSUE #3589: Invalid syntax in completion test for record pattern destructuring
In a few of the completion tests for record pattern destructuring, a record pattern is placed in the first part of an enhanced for statement, eg:

```java
import java.util.List;
public class Main {
  private static record MyRecord(String a, int b) {}
    List<MyRecord> myList = List.of(new MyRecord("hello", 12), new MyRecord("world", 42));
    for (MyRecord(String j, int k) : myList) {
      System.out.println(j);
    }
  }
}
```

Here is an example of such a test:
https://github.com/eclipse-jdt/eclipse.jdt.core/blob/57ae0e908f9aa809ddd40442d36c8a22b34fb4b8/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTestsForRecordPattern.java#L461

After reading the specification, I believe this is invalid syntax. I confirmed the given snippet doesn't compile under ECJ or javac.

I reread [6.3.2.5](https://docs.oracle.com/javase/specs/jls/se23/html/jls-6.html#jls-6.3.2.5), the section that seems the most relevant to this test. My understanding is that the scope for `for` statements only applies to the condition portion of regular for loops, and there is no special syntax for pattern variables in enhanced for loops. eg. the following is allowed:

```java
import java.util.List;
public class Main {
  private static record MyRecord(String a, int b) {}
    List myList = List.of(new MyRecord("hello", 12), new Object());
    for (int i = 0; i < myList.size() && myList.get(i) instanceof MyRecord(String j, int k); i++) {
      System.out.println(j);
    }
  }
}
```

I think the wording of 6.3.2.5 could have caused some confusion, which lead to this test to be implemented. (After all, I think the "erroneous" syntax looks legitimate at first glance. If it was valid syntax, it would be very helpful when iterating over collections of records). It's also possible the spec changed after the feature was developed and this was a leftover from that.

Either way, I think we should adjust the tests to use valid syntax and test for the pattern variable scoping as described in 6.3.2.5.

datho7561:
## What it does
The syntax used in these tests was withdrawn before the feature was delivered (see #1907). I've altered the tests to use a regular for loop with an `instanceof` in it, so that the tests still check that the scoping of the pattern variables is working properly for completion.

Fixes #3589

## How to test
Make sure that the modified integration tests still pass.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTestsForRecordPattern.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTestsForRecordPattern.java
@@ -466,7 +466,7 @@ public class CompletionTestsForRecordPattern extends AbstractJavaModelCompletion
 					+ "    public static boolean foo(Object o) {\n"
 					+ "        boolean ret = false;\n"
 					+ "        R[] recArray = {new R(0)};\n"
-					+ "        for (R(int x_1) : recArray) {\n"
+					+ "        for (int i = 0; i < recArray.length && recArray[i] instanceof R(int x_1); i++) {\n"
 					+ "            System.out.println(x_);  \n"
 					+ "            ret = true;\n"
 					+ "        }\n"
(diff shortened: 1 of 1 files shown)


datho7561:
@srikanth-sankaran if you have some time, could you please take a look at this?

srikanth-sankaran (approved):
LGTM, Thanks!

akurtakov:
Merge as per previous comment.

Issue #3590, opened 2025-01-21
Performance issues of type inferece, and tricks


stephan-herrmann:
There's a [discussion](https://mail.openjdk.org/pipermail/compiler-dev/2025-January/029197.html) on https://mail.openjdk.org/mailman/listinfo/compiler-dev under the heading "Exponential compilation time for List.of(Map.entry) expressions"

Someone presented an [example program](https://gist.github.com/dweiss/93caf579a89f3dbbfac5c292048e42ad) which makes javac, intellij and also ecj go into deep thought mode for a long time.

Some take-aways from that discussion:
* the core problem is inherent in JLS, or even in the father of all similar type systems: ML. The problem **is** exponential. Period.
* IntelliJ has an [inspection + quickfix](https://www.jetbrains.com/help/inspectopedia/JavacQuirks.html) that let's users add explicit type arguments where it helps to avoid the problem
* javac already performs some tricks to cut down the number of inference variables, see https://mail.openjdk.org/pipermail/compiler-dev/2025-January/029239.html

Items (2) and (3) are worth considering also in JDT.

jukzi:
jdt.ui does not like to edit such source either resulting in major UI freeze:

```
"main" #1 [27096] prio=6 os_prio=0 cpu=94609.38ms elapsed=1134.11s tid=0x000002663b9ea9e0 nid=27096 runnable  [0x000000713d8fb000]
   java.lang.Thread.State: RUNNABLE
        at org.eclipse.jdt.internal.compiler.lookup.InferenceVariable.hashCode(InferenceVariable.java:216)
        at java.util.HashMap.hash(java.base@22.0.2/HashMap.java:338)
        at java.util.HashMap.getNode(java.base@22.0.2/HashMap.java:576)
        at java.util.LinkedHashMap.get(java.base@22.0.2/LinkedHashMap.java:536)
        at org.eclipse.jdt.internal.compiler.lookup.BoundSet.dependsOnResolutionOf(BoundSet.java:1058)
        at org.eclipse.jdt.internal.compiler.lookup.InferenceContext18.addDependencies(InferenceContext18.java:1396)
        at org.eclipse.jdt.internal.compiler.lookup.InferenceContext18.addDependencies(InferenceContext18.java:1397)
        at org.eclipse.jdt.internal.compiler.lookup.InferenceContext18.getSmallestVariableSet(InferenceContext18.java:1362)
[...]

PR #3594, opened 2025-01-22
[performance] InferenceVariable.hashCode: avoid Mersenne prime# 3593

jukzi:
```
  something  + 31*sourceStart + (sourceStart + sourceLength)
= something  + 32*sourceStart + sourceLength
-> almost constant low bits
```

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3593

PR #3597, opened 2025-01-22
[performance] BoundSet: move properTypesByInferenceVariable before loop

jukzi:
as it does not depend on the loop variables

improves https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3590 drastically

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/BoundSet.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/BoundSet.java
@@ -531,6 +531,7 @@ class BoundSet {
 		boolean analyzeNull = context.environment.globalOptions.isAnnotationBasedNullAnalysisEnabled;
 		ConstraintTypeFormula [] mostRecentFormulas = new ConstraintTypeFormula[4]; // poor man's cache to toss out duplicates, in pathological cases there are a good quarter million of them.
 		// check each pair, in each way.
+		Map<InferenceVariable,TypeBound> properTypesByInferenceVariable = properTypesByInferenceVariable(first, next);
 		for (TypeBound bound1 : first) {
 			for (TypeBound bound2 : next) {
 				if (bound1 == bound2)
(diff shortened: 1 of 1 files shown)


jukzi:
reduces reproducer https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3594#issuecomment-2607265264 from 17.7s to 5s

after:
image

stephan-herrmann:
I'll take a look, though not right now ...

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/BoundSet.java
```diff
@@ -871,20 +872,19 @@ private boolean isParameterizedDependency(TypeBound typeBound) {
...
 		return null;
 	}
 
 	private Collection<TypeBound> getProperTypesForAllInferenceVariablesOrNull(ParameterizedTypeBinding parameterizedType,
-			TypeBound[] firstBounds, TypeBound[] nextBounds) {
-		final Map<InferenceVariable,TypeBound> properTypesByInferenceVariable = properTypesByInferenceVariable(firstBounds, nextBounds);
-		if(properTypesByInferenceVariable.size() == 0) {
-			return null;
-		}
+			TypeBound[] firstBounds, TypeBound[] nextBounds, Map<InferenceVariable,TypeBound> properTypesByInferenceVariable) {
```
stephan-herrmann, line 887: `firstBounds` and `nextBounds` are no longer used in this method. Removal should propagate to callers several levels out (thus reverting their introduction in 372c6048)

stephan-herrmann (changes requested):
The change is technically sound.

In addition to the cleanup request, I'd like to see one more measurement:

You reported that the extreme case from https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3594#issuecomment-2607265264 benefits from the change, good.

What about "normal" usage? The computation `properTypesByInferenceVariable()` will now be performed always, even if its result is never used. Does this method appear in any measurements of, e.g., `GenericsRegressionTest_1_8`?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/BoundSet.java
```diff
@@ -871,20 +872,19 @@ private boolean isParameterizedDependency(TypeBound typeBound) {
 	}
 
 	private ConstraintTypeFormula incorporateIntoParameterizedDependencyIfAllArgumentsAreProperTypes(TypeBound typeBound,
-			TypeBound[] firstBounds, TypeBound[] nextBounds) {
-		Collection<TypeBound> properTypesForAllInferenceVariables = getProperTypesForAllInferenceVariablesOrNull((ParameterizedTypeBinding)typeBound.right, firstBounds, nextBounds);
+			TypeBound[] firstBounds, TypeBound[] nextBounds, Map<InferenceVariable,TypeBound> properTypesByInferenceVariable) {
+		if(properTypesByInferenceVariable.isEmpty()) {
+			return null;
+		}
```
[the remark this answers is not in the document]
jukzi: not anymore, i moved it back now.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/BoundSet.java
jukzi: good. done.

jukzi:
> What about "normal" usage? The computation `properTypesByInferenceVariable()` will now be performed always, even if its result is never used. Does this method appear in any measurements of, e.g., `GenericsRegressionTest_1_8`?

I executed GenericsRegressionTest_1_8 with and without this PR and could not see a performance difference above the normal jitter. Both cases take ~ 1:22 +/-1sec  to run. The method is used during that test for ~200ms. Its caller, BoundSet.incorporate(), is dominated by another method reduceOneConstraint().
As far as i understand properTypesByInferenceVariable() is super fast on cases that would not require to compute it and would only take significant time if the result will be used.
image

Issue #3559, opened 2025-01-15
Boolean switch inconsistency between ECJ and javac


nlisker:
```java
public class Test {

	public static void main(String[] args) {
		new Test().d(true);
	}

	void d(Boolean b) {
		switch (b) {
			case true  -> System.out.println("1");
			case false -> System.out.println("2");
		};
	}
}
```
In Java 22, this compiles and runs in ECJ, but javac reports
```
Test.java:11: error: constant label of type boolean is not compatible with switch selector type Boolean
			case true -> System.out.println("1");
			     ^
Test.java:12: error: constant label of type boolean is not compatible with switch selector type Boolean
			case false -> System.out.println("2");
			     ^
```

However, adding a guard pattern makes it not compile anymore in ECJ:
```java
case true when b.toString().length() == 1 -> System.out.println("1");
```

Perhaps related to https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3558.

srikanth-sankaran:
@stephan-herrmann - do you want to look at these issues with Bollean switches? If not please assign to me. Thanks.

srikanth-sankaran:
Behavior change happened at 4.34 M1. Upto 4.33 we reject the code with:

Description	Resource	Path	Location	Type
Type mismatch: cannot convert from boolean to Boolean	Test.java	/_pasted_code_/src	line 9	Java Problem
Type mismatch: cannot convert from boolean to Boolean	Test.java	/_pasted_code_/src	line 10	Java Problem

srikanth-sankaran:
Confirmed that this is a regression brought on by BETA_JAVA23 branch being merged onto master via commit f8fb58210e5a04aa17773fe757e7e02b2b9ac5cd

srikanth-sankaran:
JDK23 compiles the code (without any preview flags) - that looks certainly suspect!

And I don't think the problem is particular to booleans.

For example, the following was rejected by JDK22, but accepted by JDK23 (without any preview flags):

```
public class X {

	public static void main(String[] args) {
		new X().d(10L);
	}

	void d(Long l) {
		switch (l) {
			case 10L  -> System.out.println("1");
			case 20L -> System.out.println("2");
                       default -> System.out.println("Default");
		};
	}
}

```

@stephan-herrmann - do you agree that JDK23 behavior is incorrect ?

stephan-herrmann:
Sorry for late response. 

Some more fun facts:
[...]

PR #3604, opened 2025-01-23
[24] Check if preview flag is set in classfiles where needed

stephan-herrmann:
+ unify verifyClassFile methods
+ sprinkle actual checks into a few current preview test classes

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/CompilationResult.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/CompilationResult.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2023 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -7,6 +7,10 @@
  * https://www.eclipse.org/legal/epl-2.0/
  *
  * SPDX-License-Identifier: EPL-2.0
+ *
+ * This is an implementation of an early-draft specification developed under the Java
+ * Community Process (JCP) and is made available for testing and evaluation purposes
+ * only. The code is not compatible with any specification of the JCP.
  *
  * Contributors:
  *     IBM Corporation - initial API and implementation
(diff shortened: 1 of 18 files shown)


stephan-herrmann:
@jarthana @mpalat  I'm kindly asking for your review, since the second commit in this PR changes a design made way back in https://bugs.eclipse.org/bugs/show_bug.cgi?id=533619 : originally, classfiles are tagged as requiring preview as soon as `--enable-preview` has been _requested_.

The way I read https://openjdk.org/jeps/12 , however, the version increment of `0xFFFF` should only be set on classfiles that actually _use/require_ a preview feature or API.

This change revealed that for `requires transitive java.base` (from feature MODULE_IMPORT) I forgot to raise the preview warning.

OTOH, it suggests that we may want to add the check from commit 1 in more of our tests.

mpalat:
> 
> The way I read https://openjdk.org/jeps/12 , however, the version increment of `0xFFFF` should only be set on classfiles that actually _use/require_ a preview feature or API.

@stephan-herrmann ..hmm this subtle point seems to be true:
"""
If Java source code uses (i) a preview language feature of Java SE $N, or (ii) a class or interface declared using a preview language feature of Java SE $N, or (iii) a normal preview API of Java SE $N (whether referenced by name, or invoked, or overridden), then a Java compiler must record that the emitted class file depends on the preview features of Java SE $N.
 """

Will review the code

stephan-herrmann:
Unfortunately, my first attempt was incomplete. 

[f0b12f1](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3604/commits/f0b12f173b3d58aa1cf7a93a297ffc193462b442) tries to fill the gap, but it uncovered a few tests where the flag is not stored in the classfile with the new strategy. Part of this points to https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3559 (not yet merged in BETA_JAVA24), but there might be more to it.

Sorry for adding more changes after requesting review already...

OTOH, had we resolved this earlier, then #3559 would have been discovered much earlier :)

mpalat:
@stephan-herrmann will complete the review by EOD (Saturday 25) today

mpalat (approved):
+1
@stephan-herrmann The code looks good. I looked into all the four preview features that affects the source code and all of them look covered with testcases also updated. Code looks complete (as usual :) ).  We can go ahead with this merge.

mpalat:
@stephan-herrmann I do have a further comment on the points 2 and 3 of [JEP points](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3604#issuecomment-2609967884). While testing for features not affecting source, eg: Scoped Values,  there are two issues:
(1) ecj does not give an error if the preview feature is disabled while javac gives. 
(2) Also, the minor version of 0xFFFF is not set in .class file even if the preview features is enabled.

As mentioned in the review comment, this PR can be merged and  these can be taken up for discussion and resolution in a new issue - issue #3614 . Thanks!

PR #3606, opened 2025-01-23
Forward release setting to AbstractImageBuilder

mickaelistria:
## What it does

Honor the OPTION_release setting in CompilerOptions so it can easily be used in downstream AbstractImageBuilder.

Note that there is currently no practical consumer for this option in ECJ-based builder; but it still seems interesting to allow proper access to this setting anyway.

## How to test

Run tests, see no regression.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/impl/CompilerOptions.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/impl/CompilerOptions.java
@@ -434,6 +434,11 @@ public class CompilerOptions {
 	/** Java source level, refers to a JDK version, e.g. {@link ClassFileConstants#JDK1_4} */
 	public long sourceLevel;
 	/**
+	 * Use <code>-release</code> setting to pass compliance version and enable checking for
+	 * availability of system APIs for the compliance version.
+	 */
+	public boolean release;
+	/**
 	 * Initially requested source version, not necessarily consistent with {@link #sourceLevel} as
 	 * sourceLevel forcibly contain a version that is compatible with ECJ.
 	 * <p>Consumers are free to use {@code sourceLevel} or this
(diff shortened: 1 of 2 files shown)


jukzi:
mind to add any test for it so it is not accidentally removed?

stephan-herrmann:
what's wrong with asking `this.javaBuilder.javaProject.getOption(JavaCore.COMPILER_RELEASE)` or similar?

mickaelistria:
> what's wrong with asking this.javaBuilder.javaProject.getOption(JavaCore.COMPILER_RELEASE) or similar?

Nothing too wrong, but since the non-deprecated constructors for Compiler and the AbstractImageBuilder.newCompiler() do use CompilerOptions as input without easy access to a particular project, it seems like honoring the release in CompilerOptions would make things much more consistent.

> mind to add any test for it so it is not accidentally removed?

I'll try it soon.

stephan-herrmann:
> ... AbstractImageBuilder.newCompiler() do use CompilerOptions as input without easy access to a particular project

`AbstractImageBuilder.newCompiler()` certainly has no issues accessing the project:

```
protected Compiler newCompiler() {
	// disable entire javadoc support if not interested in diagnostics
	Map projectOptions = this.javaBuilder.javaProject.getOptions(true);
...
```
I simply don't get which problem you are solving... you may have a relevant usecase in mind, but I don't see it.  `--release` is interesting for name environments not for the compiler.

mickaelistria:
But then `Compiler newCompiler = compilerFactory.newCompiler(..)` looses track of the project, but has access to the CompilerOptions.
Another argument is the one can expect that the value returned by `new CompilerOptions(this.javaBuilder.javaProject.getOptions(true)).getMap()` contains no different value (not contain all values, but all the contained ones are equivalent) than initial `this.javaBuilder.javaProject.getOptions(true)`. Some data may be lost in translation, but none should be changed in translation, as it's currently the case for `release`.

And as you probably have guess, when working on an alternative compiler (eg Javac) we need to have access to information about whether `--release` is to be used; and CompilerOptions are used to access the relevant options.

stephan-herrmann:
> ... as you probably have guess ...

... I prefer explicit communication over guessing ... question answered.

mickaelistria:
> mind to add any test for it so it is not accidentally removed?

Done in latest patch set.

jukzi:
LGTM @stephan-herrmann still concerns? looks like it won't harm anybody

mickaelistria:
Thank you all for the review and merge.

PR #3613, opened 2025-01-24
Extension point to replace `CompletionEngine` implementation

datho7561:
## What it does
This PR allows plugins to replace the implementation of `CompletionEngine` used in `Openable`. This covers most completion cases.

In order to accomplish this, this PR introduces the extension point `completionEngineProvider`, which allows plugins to specify an implementation of `ICompletionEngineProvider`. `ICompletionEngineProvider` consists of a method with the same parameters as the `CompletionEngine` constructor that returns an `ICompletionEngine`.
`ICompletionEngine` consists of one method that is identical to the existing method `CompletionEngine.complete(ICompilationUnit sourceUnit, int completionPosition, int pos, ITypeRoot root)`.

In order to configure a new `ICompletionEngineProvider`, register the class in `plugin.xml` using the newly introduced extension point. Also, set the Java argument `ICompletionEngineProvider` to the implementation, eg. `-DICompletionEngineProvider=dev.datho7561.completion.MyCompletionEngineProvider`.

## How to test
I've added a regression test in which I configure a simple test implementation of `ICompletionEngineProvider` that provides an `ICompletionEngine` that always returns the same result. I make sure the result is the one provided by the test implementation, instead of the one returned by the default `CompletionEngine`.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionEngineProviderTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionEngineProviderTests.java
@@ -1,0 +1,91 @@
+/*******************************************************************************
+ * Copyright (c) 2025 Red Hat, Inc. and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.jdt.core.tests.model;
+
+import junit.framework.Test;
+import org.eclipse.jdt.core.ICompilationUnit;
+import org.eclipse.jdt.core.JavaModelException;
+import org.eclipse.jdt.internal.compiler.impl.CompilerOptions;
+
+/**
+ * Test for the `org.eclipse.jdt.core.completionEngineProvider` extension point
+ * and {@link org.eclipse.jdt.internal.core.CompletionEngineProviderDiscovery}.
+ */
+@SuppressWarnings("javadoc")
+public class CompletionEngineProviderTests extends AbstractJavaModelCompletionTests {
+
+	private String oldSystemProperty;
+	private static final String ENGINE_PROPERTY = "ICompletionEngineProvider";
+
+	public CompletionEngineProviderTests(String name) {
+		super(name);
+	}
+
+	public static Test suite() {
+		return buildModelTestSuite(CompletionEngineProviderTests.class, ALPHABETICAL_SORT);
+	}
+
+	@Override
+	public void setUpSuite() throws Exception {
+		super.setUpSuite();
+		setupExternalJCL("jclMin");
+		if (COMPLETION_PROJECT == null)  {
+			COMPLETION_PROJECT = setUpJavaProject("Completion");
+		} else {
+			setUpProjectCompliance(COMPLETION_PROJECT, CompilerOptions.getFirstSupportedJavaVersion());
+		}
+		this.currentProject = COMPLETION_PROJECT;
+	}
+
+	@Override
+	protected void setUp() throws Exception {
+		super.setUp();
+		this.oldSystemProperty = System.getProperty(ENGINE_PROPERTY);
+		System.setProperty(ENGINE_PROPERTY, "org.eclipse.jdt.core.tests.model.TestCompletionEngineProvider");
+	}
+
+	@Override
+	protected void tearDown() throws Exception {
+		super.tearDown();
+		if (this.oldSystemProperty != null) {
+			System.setProperty(ENGINE_PROPERTY, this.oldSystemProperty);
+		} else {
+			System.clearProperty(ENGINE_PROPERTY);
+		}
+	}
+
+	public void testAlwaysProvidesSameResult() throws JavaModelException {
+		this.workingCopies = new ICompilationUnit[1];
+		this.workingCopies[0] = getWorkingCopy(
+			"/Completion/src/ArrayInitializer.java",
+			"public class ArrayInitializer {\n"+
+			"	int bar() {return 0;}\n"+
+			"	void foo(int[] i) {\n"+
+			"		i = new int[] {\n"+
+			"			bar()\n"+
+			"		};\n"+
+			"	}\n"+
+			"}\n");
+
+		CompletionTestsRequestor2 requestor = new CompletionTestsRequestor2();
+
+		String str = this.workingCopies[0].getSource();
+		String completeBehind = "bar(";
+		int cursorLocation = str.lastIndexOf(completeBehind) + completeBehind.length();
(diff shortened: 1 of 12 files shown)


org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/CompletionEngineProviderDiscovery.java
```diff
@@ -0,0 +1,91 @@
...
+
+	public static ICompletionEngineProvider getInstance() {
+		String id = System.getProperty(SELECTED_SYSPROP);
+		IConfigurationElement configElement = getConfigurationElement(id);
+		lastId = id;
+		lastExtension = configElement;
+		if (configElement != null) {
+			try {
+				// We do prefer creating a new instance on each call, as it can allow the extension
+				// to store some state more easily than by using a singleton.
```
rgrunber, line 40: I would maybe test roughly how long this part takes in regular usage. I would have thought caching after the first invocation would be a way to avoid any surprises.

rgrunber: On second though, I guess this isn't so bad because in the normal case, there's no extension defined for the standard use case, so none of that runs.

datho7561: Takes around a millisecond. I could do some caching with a HashMap to avoid recreating it. To be honest, I copied this code from the CompilationUnitResolver extension point, so we should be able to cache it safely

datho7561: I've implemented a caching mechanism

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/CompletionEngineProviderDiscovery.java
```diff
@@ -0,0 +1,97 @@
...
+			for (IExtension ext : extensions) {
+				IConfigurationElement[] configElements = ext.getConfigurationElements();
+				for (final IConfigurationElement configElement : configElements) {
+					String elementId = configElement.getAttribute("id"); //$NON-NLS-1$
+					if (id.equals(elementId) && "resolver".equals(configElement.getName())) { //$NON-NLS-1$
+						return configElement;
+					}
+				}
+			}
+		}
```
rgrunber, line 84: Just wanted to confirm. The extension point framework handles correctly starting the bundle that contains the contributed extension right ? That's why you didn't just directly do `Class.forName(...).getMethod(...).invoke(...)` on the system property right ? I'm asking because otherwise it seems simpler to bypass all this.

datho7561: Honestly this class started out as a copy of [CompilationUnitResolverDiscovery](https://github.com/eclipse-jdt/eclipse.jdt.core/blob/ca77e4c56980ac52d225595273979dfd1316cfd4/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolverDiscovery.java#L23). I don't fully understand it, but it was the reference I was working off of.

rgrunber: Ok fair, if it's based off of that, and that was merged, then let's stick to that approach since obviously there hasn't been any opposition to it.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionEngineProviderTests.java
```diff
@@ -0,0 +1,90 @@
...
+import junit.framework.Test;
+import org.eclipse.jdt.core.ICompilationUnit;
+import org.eclipse.jdt.core.JavaModelException;
+import org.eclipse.jdt.internal.compiler.impl.CompilerOptions;
+
+/**
+ * Test for the `org.eclipse.jdt.core.completionEngineProvider` extension point
+ * and {@link org.eclipse.jdt.internal.core.CompletionEngineProviderDiscovery}.
+ */
+public class CompletionEngineProviderTests extends AbstractJavaModelCompletionTests {
```
rgrunber, line 22: Is it worth adding a `@SuppressWarnings("javadoc")` here to silence the warning, and just avoid increasing them ?

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/ICompletionEngine.java
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Represents a class that can provide completion for a compilation unit.
+ */
+public interface ICompletionEngine {
+
+	/**
+	 * Provide completion items based on the given source unit and completion position to the configured requestor.
+	 *
+	 * @param sourceUnit the compilation unit for which completion was triggered
+	 * @param ast the AST of the compilation unit to use for completion. can be null, in which case the completion engine is expected to build it
```
rgrunber, line 25: Looks like `ast` isn't used. Safe to remove ?

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/Openable.java
```diff
@@ -132,8 +134,10 @@ protected void codeComplete(
 	environment.unitToSkip = unitToSkip;
 
 	// code complete
-	CompletionEngine engine = new CompletionEngine(environment, requestor, project.getOptions(true), project, owner, monitor);
-	engine.complete(cu, position, 0, typeRoot);
```
rgrunber, line 136: Was the choice to create a helper method just to avoid that 0 parameter for "offset" ? Any reason not to use the 4-arg method in the interface ?

rgrunber (commented):
Just a few minor things and a question.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/Openable.java
rgrunber: No need to change, was just curious, and confused by what `pos` (undocumented) did given that `completionPosition` was the cursor position.

datho7561: I looked into it. The places that use this parameter (ie. set it to something other than 0) often reach into the internals of `CompletionEngine`, or also use a different constructor on another codepath. For the purposes of replacing the implementation in `Openable`, it seems this adjustment feature isn't used, which is why I wanted to drop the parameter from the interface.

datho7561: > It would look a bit less sketchy than hiding the mystery parameter

Alright, makes sense. I've added the parameter to the interface. I added Javadoc explaining the purpose of `pos`  in `CompletionEngine` and renamed it to `adjustment`.

rgrunber (approved):
LGTM. Will be merging soon.

PR #3616, opened 2025-01-25
Preview Flag not set with runtime only preview features #3614

### LINKED ISSUE #3614: Preview Flag not set with runtime only preview features
ref: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3604#issuecomment-2609622313

Ref: JEP 12 https://openjdk.org/jeps/12 :
This issues is primarily about the point no: 2 (and maybe 3) from the following comment:
"If Java source code uses (i) a preview language feature of Java SE $N, or (ii) a class or interface declared using a preview language feature of Java SE $N, or (iii) a normal preview API of Java SE $N (whether referenced by name, or invoked, or overridden), then a Java compiler must record that the emitted class file depends on the preview features of Java SE $N."

From points 2 and 3 of https://openjdk.org/jeps/12: for features not affecting source, for eg: Scoped Values, there are two issues:
(1) ecj does not give an error if the preview feature is disabled while javac gives.
(2) Also, the minor version of 0xFFFF is not set in .class file even if the preview features is enabled.
eg:

```
public class X {
		ScopedValue<Integer> si = ScopedValue.newInstance();
		System.out.println(si == null ? "hello" : "world");
	}
}

```

**Problem No:1**

use java24 (I used ea31) to compile without any flag 
$javac X.java
to get the following output:
"""
X.java:3: error: ScopedValue is a preview API and is disabled by default.
		ScopedValue<Integer> si = ScopedValue.newInstance();
		^
  (use --enable-preview to enable preview APIs)
X.java:3: error: ScopedValue is a preview API and is disabled by default.
		ScopedValue<Integer> si = ScopedValue.newInstance();
		                          ^
  (use --enable-preview to enable preview APIs)
2 errors
"""

Now use ecj (24) without --enable-preview flag, it just compiles (java used here is 24 ea31 just fyi)
$java -jar ./ecj.jar --release 24  X.java 
And we get X.class file
Expected Behaviour: ecj should throw the preview not enabled error

**Problem No:2**

use java24  to compile with --enable-preview flag 
$javac --enable-preview --release 24  X.java
X.class file is generated.
Inspect using javap,
$javap -p -v -c X|grep minor
we get ==>   _minor version: 65535_

Now, use ecj with the same set of flags:
$ java -jar ./ecj.jar --enable-preview --release 24  X.java
Inspect using javap,
we get ==>  _minor version: 0_

Expected Behaviour: ecj should have _minor version: 65535_

mpalat:
+ Improve analysis and reporting for usage of preview API.
+ Ensure that usage of preview API is also recorded in the resulting class file.

Fixes #3614

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/core/compiler/IProblem.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/core/compiler/IProblem.java
@@ -2250,6 +2250,8 @@ void setSourceStart(int sourceStart);
 	int JavaVersionNotSupported = Compliance + 1109;
 	/** @since 3.40*/
 	int JavaVersionTooRecent = Compliance + 1110;
+	/** @since 3.41 */
+	int PreviewAPIDisabled = Compliance + 1111;
 
 	/** @since 3.13 */
 	int UnlikelyCollectionMethodArgumentType = 1200;
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ASTNode.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ASTNode.java
@@ -58,6 +58,7 @@ import org.eclipse.jdt.internal.compiler.ASTVisitor;
 import org.eclipse.jdt.internal.compiler.ast.TypeReference.AnnotationPosition;
 import org.eclipse.jdt.internal.compiler.classfmt.ClassFileConstants;
 import org.eclipse.jdt.internal.compiler.env.AccessRestriction;
+import org.eclipse.jdt.internal.compiler.env.IBinaryAnnotation;
 import org.eclipse.jdt.internal.compiler.lookup.*;
 
 @SuppressWarnings({"rawtypes", "unchecked"})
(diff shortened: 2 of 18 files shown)


stephan-herrmann:
that's a trick: add a new test class without linking it to the suite and see jenkins build succeed :)

stephan-herrmann:
Updates coming in a minute, some comments about changes:

Things that looked bogus / outdated:
* there was a mix of using `tagBits` and `extendedTagBits`: write to the former, try to read the value from that latter and call it "modifiers"  ;-P
* some code looked outdated:
  * I can't find an attribute `essentialAPI` in annotation `PreviewFeature`, true the JEP mentioned "essential" as one of the categories, but this doesn't seem to be reflected in code, nor do I see impact on compilation
  * instead there is the attribute `reflective` which indeed is the basis for an exception, downgrading errors to warnings as outlined in the JEP under "Special rules for use of reflective preview APIs"
* I couldn't see a strong reason to match "PreviewFeature" only by the simple name, changed to qualified name

New implementation:
* Remove usage of tag bits
  * given that we detect application of `@PreviewFeature` based on the qualified name in binary classes (and no need to detect it in source types), I'm removing `TagBits.AnnotationPreviewFeature` and `TypeIds.T_JdkInternalPreviewFeature`. Am happy that this slightly relaxes the overloaded space of bits in `TagBits`.
* When seeing `@PreviewFeature` on any type, field or method, keep the `IBinaryAnnotation` for later use (field `binaryPreviewAnnotation`)
* When reading a binary annotation of type `jdk.internal.javac.PreviewFeature.JEP` force storing of this annotation so we have access to its members (these are annotations on the enum constants in `PreviewFeature.Feature`).
* All further logic is in `ProblemReporter.previewAPIUsed()`:
  * evaluate the annotation so we know
    * the name of the preview feature
    * value of the flag `reflective`
  * with this information we can select:
     * messages (disabled, or informative)
     * severity (error if disabled unless reflective)

PR #3619, opened 2025-01-26
Fix new failure in run.javac comparison from verifyClassFile()

stephan-herrmann:
After recent additions (#3604) ModuleImportTests had 8 failures in run.javac mode, because the branch for javac compilation wiped those class files that should be inspected for proper setting of the preview indicator.


stephan-herrmann:
Build is green again: https://ci.eclipse.org/jdt/job/eclipse.jdt.core-run.javac-24/19/

Issue #2877, opened 2024-08-30
randomly Broken imports - type from JDK cannot be resolved


andreasdc:
I often get errors about not detected classes, for example:
java.util.IdentityHashMap cannot be resolved to a type
today I even had error on System.out.
Cleaning or refreshing doesn't help, but restarting Eclipse fixes that.

jukzi:
I experienced the same in the last months some times. However i do not know how to reproduce :-( For me it was always only about the classes in JDK.  - and yes only restarting helped - reliably.

andreasdc:
> I experienced the same in the last months some times. However i do not know how to reproduce :-( For me it was always only about the classes in JDK. - and yes only restarting helped - reliably.

"The type java.io.PrintStream cannot be resolved. It is indirectly referenced from required type java.lang.System"
I just did something in another project after restart and I see this error in couple projects.

jnsmiles:
I also have been confronted to that random bug since a long time (more than 3 years).

Here is a protocol to try to reproduce that bug:

Code context
project0
package pkg0

unit MyNestClassA.java
public class MyNestClassA {
  public static class MyMemberClassB {
    // MyMemberClassB is widely used across multiple projects
  }
}

unit MyNestClassC.java
public class MyNestClassC {
}


project1
package pkg1

unit D.java
import javax.annotations.Generated


@Generated("","")
public class D {
  // ...
  public void myIndirectUsageOfClassB() {
    // ...
  }
}

unit E.java
public class E extends class D {
}

Update operations:
1) cut code of MyMemberClassB from MyNestClassA and paste it into MyNestClassC
2) run control+B command (rebuild all projects in workspace)
3) look for classes that have compilation error flags

The result:
- Class E has a compilation error flag with a non explicit message
- class D has no error flag but when I open unit in editor error flags appear
- all classes that use Generated class have a compilation error reporting that this class is missing
- I try to open Generated class in the Java Editor it is unavailable
  and a get the exception log reported below
- Generated class appears normally in the package explorer
- other classes from Generated package like Filler are all normally available
- the missing a API may be recovered when all import clauses have been corrected and after a full workspace clean and rebuild 
  and some times multiple eclipse platform restart

  
This bug appears randomly and with potentially any JRE API (I got it with Float, Boolean, StandardCharsets, IllegalArgumentException) and only JRE APIs.

jnsmiles:
Console error log:

[...]

PR #3620, opened 2025-01-27
Handle ClosedByInterruptException in JRTUtil#safeReadBytes only once

laeubi:
Currently the method JRTUtil#safeReadBytes handles
ClosedByInterruptException in a problematic way:

1. The API states that all I/O errors are rethrown except
NoSuchFileException
2. There is no way for the caller to distinguish the case of notfound
and interrupted given the method return null in both cases
3. Some callers even expect a ClosedByInterruptException to function
properly, all other caller either rethrow or handle I/O errors already
4. ClosedByInterruptException is documented to set set interupt flag of
the thread, so reading it again will just again throw (and ignore) the
exception, there is not any documentation that this can ever throw
ClosedByInterruptException 'randomly' or in a recoverable way other than
maybe reset the interrupt flag of the thread

This now clear the interrupt flag when ClosedByInterruptException is
found and retry once again, if this still fails rethrow the exception to
the caller.

See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2877

### DIFF
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/util/JrtUtilTest.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/util/JrtUtilTest.java
@@ -14,6 +14,10 @@
 package org.eclipse.jdt.core.tests.compiler.util;
 
 import java.io.File;
+import java.io.IOException;
+import java.net.URI;
+import java.nio.file.FileSystem;
+import java.nio.file.FileSystems;
 import java.nio.file.Paths;
 import org.eclipse.jdt.core.JavaCore;
 import org.eclipse.jdt.core.tests.junit.extension.TestCase;
@@ -68,6 +72,20 @@ public class JrtUtilTest extends TestCase {
 
 		Object jrtSystem3 = JRTUtil.getJrtSystem(this.image, null);
 		assertSame(jrtSystem, jrtSystem3);
+	}
+
+	@Test
+	public void testWorksUnderInterruptCondition() throws IOException {
+		FileSystem fs = FileSystems.getFileSystem(URI.create("jrt:/"));
+		try {
+			Thread.currentThread().interrupt();
+			assertNotNull(JRTUtil.safeReadBytes(fs.getPath("modules", "java.base", "java/lang/Object.class")));
+			assertTrue(Thread.currentThread().isInterrupted());
+			assertNull(JRTUtil.safeReadBytes(fs.getPath("xxxxxxxxx")));
+			assertTrue(Thread.currentThread().isInterrupted());
+		} finally {
+			Thread.interrupted();
+		}
 	}
 
 	private static int getMajorVersionSegment(String releaseVersion) {
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/util/JRTUtil.java
@@ -309,25 +309,26 @@ public class JRTUtil {
 	}
 
 	/**
-	 * Tries to read all bytes of the file denoted by path,
-	 * returns null if the file could not be found or if the read was interrupted.
+	 * Tries to read all bytes of the file denoted by path, returns null if the file could not be found. If
+	 * {@link ClosedByInterruptException} occurs, one more attempt is made to read the file before the error is
+	 * rethrown.
+	 *
 	 * @return bytes or null
-	 * @throws IOException any IO exception other than NoSuchFileException
+	 * @throws IOException
+	 *             any IO exception other than NoSuchFileException
 	 */
 	public static byte[] safeReadBytes(Path path) throws IOException {
 		try {
 			return Files.readAllBytes(path);
 		} catch (ClosedByInterruptException e) {
-			// retry once again
+			Thread.interrupted(); // clear the interrupt flag to simulate non interruptible i/o
 			try {
 				return Files.readAllBytes(path);
 			} catch (NoSuchFileException e2) {
 				return null;
-			} catch (ClosedByInterruptException e2) {
-				if (PROPAGATE_IO_ERRORS) {
-					throw e2;
-				}
-				return null;
+			} finally {
+				// restore interrupt flag
+				Thread.currentThread().interrupt();
 			}
 		} catch (NoSuchFileException e) {
 			return null;


laeubi:
@stephan-herrmann can you review / merge?

stephan-herrmann:
I'm hesitant to say top or flop. Right when I was about to say that I know very little about this code area, I found https://bugs.eclipse.org/bugs/show_bug.cgi?id=525842 demonstrating that I was indeed involved in the current broken state.

Perhaps @iloveeclipse (who certainly remembers the discussion in bug 525842 :smile:) or @jukzi  have a clearer view of how things should be?

Also our bugzilla links to https://bugs.openjdk.org/browse/JDK-8156602 which interestingly speaks of making the read non-interruptible ... does that give a hint for us?

stephan-herrmann:
Specifically intriguing in the JDK bug: `-Djdk.image.map.all=false` is supposed to trigger the problem quite reliably. Will it trigger also our current problem? That would remove any doubt about the cause, wouldn't it?

laeubi:
> Also our bugzilla links to https://bugs.openjdk.org/browse/JDK-8156602 which interestingly speaks of making the read non-interruptible ... does that give a hint for us?

As described here, its really hard to tell with the current state because two cases are not possible to distinguish so safereadbytes is not really the right place to handle this.

Looking at the code of the JDK it is like this:

```
    public static byte[] readAllBytes(Path path) throws IOException {
        try (SeekableByteChannel sbc = Files.newByteChannel(path);
             InputStream in = Channels.newInputStream(sbc)) {
            if (sbc instanceof FileChannelImpl)
                ((FileChannelImpl) sbc).setUninterruptible();
            long size = sbc.size();
            if (size > (long) Integer.MAX_VALUE)
                throw new OutOfMemoryError("Required array size too large");
            return read(in, (int)size);
        }
    }
```

and I tried with this testcase that one actually can't interrupt a file read:

```
package org.eclipse.jdt.internal.compiler.util;

import java.io.IOException;
import java.nio.file.Files;
import java.nio.file.Path;
import java.util.UUID;

public class InterruptMe {
	public static void main(String[] args) throws IOException {
		Path gibsNicht = Path.of(UUID.randomUUID().toString());
		assertThat(!Files.isRegularFile(gibsNicht), "A regular file?!?");
		assertThat(JRTUtil.safeReadBytes(gibsNicht) == null, "Should return null");
		Path tempFile = Files.createTempFile("testme", ".tmp");
		assertThat(JRTUtil.safeReadBytes(tempFile) != null, "Should not return null");
		Thread.currentThread().interrupt();
		assertThat(JRTUtil.safeReadBytes(tempFile) != null, "Should not return null");
		assertThat(Thread.currentThread().isInterrupted(), "thread must be intterupted");
	}

	private static void assertThat(boolean condition, String message) {
		if (!condition) {
			System.err.println(message);
		}

	}
}
```

but we are at the JRTFileSystem here and there it is like this:

```
	public static void main(String[] args) throws IOException {
		FileSystem fs = FileSystems.getFileSystem(URI.create("jrt:/"));
		Thread.currentThread().interrupt();
		Path p = fs.getPath("modules", "java.base", "java/lang/Object.class");
		assertThat(JRTUtil.safeReadBytes(p) != null, "Should not return null");
		assertThat(Thread.currentThread().isInterrupted(), "thread must be intterupted");
	}
```

the console prints:

`Should not return null`

and in the debugger I see that the catching of the exception simply don't work as it fails 2 times here, if I enable `org.eclipse.jdt.propagate_io_errors` then I get:

```
Exception in thread "main" java.nio.channels.ClosedByInterruptException
	at java.base/java.nio.channels.spi.AbstractInterruptibleChannel.end(AbstractInterruptibleChannel.java:199)
	at java.base/java.nio.channels.Channels$ReadableByteChannelImpl.read(Channels.java:389)
	at java.base/jdk.internal.jrtfs.JrtFileSystem$1.read(JrtFileSystem.java:368)
	at java.base/sun.nio.ch.ChannelInputStream.read(ChannelInputStream.java:65)
	at java.base/sun.nio.ch.ChannelInputStream.read(ChannelInputStream.java:107)
	at java.base/sun.nio.ch.ChannelInputStream.read(ChannelInputStream.java:101)
	at java.base/java.nio.file.Files.read(Files.java:3244)
	at java.base/java.nio.file.Files.readAllBytes(Files.java:3295)
	at org.eclipse.jdt.internal.compiler.util.JRTUtil.safeReadBytes(JRTUtil.java:323)
	at org.eclipse.jdt.internal.compiler.util.InterruptMe.main(InterruptMe.java:17)
```

I can make the code work by change the method like this:

```
	public static byte[] safeReadBytes(Path path) throws IOException {
		try {
			return Files.readAllBytes(path);
		} catch (ClosedByInterruptException e) {
			Thread.interrupted(); // clear the interrupt flag to simulate non interruptible i/o
			try {
				byte[] bytes = Files.readAllBytes(path);
				//restore interrupt flag
				Thread.currentThread().interrupt();
				return bytes;
			} catch (NoSuchFileException e2) {
				return null;
			} catch (ClosedByInterruptException e2) {
				throw e2;
			}
		} catch (NoSuchFileException e) {
			return null;
		}
	}
```

jukzi (changes requested):
i vote jdt should retry on ClosedByInterruptException (i.e. try to ignore interrupt) at least once but should not return null if it happens again but fail with ClosedByInterruptException  if it happens again.

laeubi:
I have now adjusted code to retry once and added a testcase that the method works under interrupt conditions

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/util/JrtUtilTest.java
```diff
@@ -70,6 +74,19 @@ public void testGetNewJrtFileSystem() throws Exception {
 		assertSame(jrtSystem, jrtSystem3);
 	}
 
+	@Test
+	public void testWorksUnderInterruptCondition() throws IOException {
+		try (FileSystem fs = FileSystems.getFileSystem(URI.create("jrt:/"))) {
```
jukzi, line 79: java.lang.UnsupportedOperationException
	at java.base/jdk.internal.jrtfs.JrtFileSystem.close(JrtFileSystem.java:104)
	at org.eclipse.jdt.core.tests.compiler.util.JrtUtilTest.testWorksUnderInterruptCondition(JrtUtilTest.java:85)

laeubi: Fixed, even though it should only throw it for "the default file system" according to the javadoc...

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/util/JrtUtilTest.java
```diff
@@ -70,6 +74,20 @@ public void testGetNewJrtFileSystem() throws Exception {
 		assertSame(jrtSystem, jrtSystem3);
 	}
 
+	@Test
+	public void testWorksUnderInterruptCondition() throws IOException {
+		FileSystem fs = FileSystems.getFileSystem(URI.create("jrt:/"));
+		try {
+			Thread.currentThread().interrupt();
+			assertNotNull(JRTUtil.safeReadBytes(fs.getPath("modules", "java.base", "java/lang/Object.class")));
```
jukzi, line 82: this assertNotNull is the only condition that failed with the previous implementation

laeubi: > this assertNotNull is the only condition that failed with the previous implementation

That is correct and expected.

jukzi (approved):
I am unsure of the introduction of usage of the interrupted flag in JDT, as normally it is just ignored in JDT. However i am tempted to submit as is to get the null away, which is a substantial fix.

jukzi:
i'll wait with submit until tomorrow to give @iloveeclipse @stephan-herrmann  another chance to review.

Issue #3601, opened 2025-01-23
performance: ReferenceBinding.superInterfaces() with redundant super interfaces


jukzi:
for example
org.eclipse.jdt.internal.compiler.lookup.BinaryTypeBinding.superInterfaces() recursively calls ReferenceBinding.superInterfaces() (without even using the returned type (`// make super-type resolving recursive for propagating typeBits downwards`) and does not protect against redundant super interfaces
sampling data from https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3387#issuecomment-2606816518:
![Image](https://github.com/user-attachments/assets/69621b4b-0c20-427c-83aa-e4046ba42c9b)
@fedejeanne you may want to protect this against already visited types too.

stephan-herrmann:
>  @fedejeanne you may want to protect this against already visited types too.

Can you explain which calls exactly cause unnecessary work?

Each BinaryTypeBinding performs computation inside `superInterfaces()` at most once:
```java
	if ((this.tagBits & TagBits.HasUnresolvedSuperinterfaces) == 0)
		return this.superInterfaces;
...
	this.tagBits &= ~TagBits.HasUnresolvedSuperinterfaces;
	return this.superInterfaces;
```

We're not talking about cyclic inheritance, are we?

fedejeanne:
It's been a while since this issue but if I recall correctly...

> We're not talking about cyclic inheritance, are we?

... I wouldn't call it _cyclic_ but definitely a _diamond_ hierarchy (through interfaces) e.g.

- `interface A`
- `interface B extends A`
- `interface C extends A`
- `interface D extends B, C`

... and similar

stephan-herrmann:
> ... I wouldn't call it _cyclic_ but definitely a _diamond_ hierarchy (through interfaces) e.g.

That's fine. In this case no binary type should perform the superInterface _computation_ more than once, right? When this computation completes it clears the flag `HasUnresolvedSuperinterfaces` and the next invocation will directly answer the field value.

Things would be bad, if we had re-entrance before the flag is cleared, but that would need a true cycle.

PR #3622, opened 2025-01-27
Avoid duplicate calls in ReferenceBinding::superInterfaces #3601

fedejeanne:
- Make the method final
- Add new recursive method ReferenceBinding.superInterfacesRecursive that takes an IndentityHashMap to keep the visited types

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3601

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/BinaryTypeBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/BinaryTypeBinding.java
@@ -2524,10 +2524,12 @@ private void detectCircularHierarchy() {
 
 // NOTE: superInterfaces of binary types are resolved when needed
 @Override
-public ReferenceBinding[] superInterfaces() {
-
+protected ReferenceBinding[] superInterfacesRecursive(Map<ReferenceBinding, Object> visited) {
+	if (visited.put(this, this) != null) { // do not evaluate the same type twice
+		return this.superInterfaces;
+	}
 	if (!isPrototype()) {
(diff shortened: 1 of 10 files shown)


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/BinaryTypeBinding.java
```diff
@@ -2524,10 +2524,12 @@ private void detectCircularHierarchy() {
 
 // NOTE: superInterfaces of binary types are resolved when needed
 @Override
-public ReferenceBinding[] superInterfaces() {
-
+protected ReferenceBinding[] superInterfacesRecursive(Map<ReferenceBinding, Object> visited) {
+	if (visited.put(this, this) != null) { // do not evaluate the same type twice
+		return this.superInterfaces;
```
jukzi, line 2529: Does it ever return here? If not - why? if yes - why doesn't it help? Maybe different instances for entries that should equal? If you could create a junit testcase that shows the bad behavior i could debug. Otherwise some tracing would help

fedejeanne: Thank you for your help, @jukzi . Just a heads up, I'll be leaving this issue to @christianstaib

fedejeanne: superseded by #3710

fedejeanne:
superseded by https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3710

Issue #3624, opened 2025-01-27
Internal compiler error: java.lang.AssertionError: array store with invalid types at org.eclipse.jdt.internal.compiler.codegen.OperandStack.xastore(OperandStack.java:168)


Albex:
Consider this snippet:

```java
import java.util.Arrays;

public class FailToCompile
{
    public static void main(String[] args)
    {
        D<Object> a = new ACDClass();
        C<Object, Object> c = new ACDClass();
        Arrays.asList(a, c); // fail to compile
        Arrays.<A<Object>>asList(a, c);// compile
    }

    static class ACDClass implements A, C, D {}
    interface A<AA> {}
    interface B<AA> {}
    interface C<AA, BB> extends A<AA>, B<BB> {}
    interface D<AA> extends A<AA>, B<Object> {}
}
```

It fails to compile from version 3.39.0+ with error:
```
Internal compiler error: java.lang.AssertionError: array store with invalid types at org.eclipse.jdt.internal.compiler.codegen.OperandStack.xastore(OperandStack.java:168)
```
It compiles using javac 21 or ecj version 3.38.0 or prior.

srikanth-sankaran:
I'll take a look

srikanth-sankaran:
Reproduced on master/HEAD

PR #3644, opened 2025-01-31
AssertionError: array store with invalid types at org.eclipse.jdt.internal.compiler.codegen.OperandStack.xastore

srikanth-sankaran:
Do not admit arrays with non-reifiable element types into book keeping. 

* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3624

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/codegen/OperandStack.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/codegen/OperandStack.java
@@ -18,6 +18,7 @@ import java.util.Stack;
 import java.util.function.Supplier;
 import org.eclipse.jdt.internal.compiler.ClassFile;
 import org.eclipse.jdt.internal.compiler.lookup.ArrayBinding;
+import org.eclipse.jdt.internal.compiler.lookup.IntersectionTypeBinding18;
 import org.eclipse.jdt.internal.compiler.lookup.ReferenceBinding;
 import org.eclipse.jdt.internal.compiler.lookup.Scope;
 import org.eclipse.jdt.internal.compiler.lookup.TypeBinding;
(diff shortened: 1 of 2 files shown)


srikanth-sankaran:
@stephan-herrmann - I need some consulting help on this one.

It appears that the type verification assertions done by `org.eclipse.jdt.internal.compiler.codegen.OperandStack.xastore()` while well intended, backfire in some cases. 

In this PR I have captured the test case and if you pull the changes and debug the test `org.eclipse.jdt.core.tests.compiler.regression.GenericsRegressionTest.testIssue3624()`
with a breakpoint in `org.eclipse.jdt.internal.compiler.codegen.CodeStream.anewarray(TypeBinding)`
you can see that we are trying to create an array with the element type being `java.lang.Object & A<java.lang.Object> & B<java.lang.Object>`

Since Java doesn't allow generic array creation, this problem is unique to the internal array created to hold variable arity arguments.

From the operand stack types tracking point of view, I really want to drop the parameterizations and say that the array is of the erased element type. 

But if I modify the line

`this.operandStack.push(scope.createArrayType(typeBinding, 1));`

in `org.eclipse.jdt.internal.compiler.codegen.CodeStream.anewarray(TypeBinding)` to be 

`this.operandStack.push(scope.createArrayType(typeBinding.erasure(), 1));`

I get back the receiver itself from the erasure call. 

It appears that for https://bugs.eclipse.org/bugs/show_bug.cgi?id=537089 you defined `org.eclipse.jdt.internal.compiler.lookup.IntersectionTypeBinding18.erasure()` to ignore j.l.O appearing as the class in an `IntersectionTypeBinding18`

When I called `erasure()` on `IntersectionTypeBinding18` typed receiver, I kind of sort of expected the behavior similar to `org.eclipse.jdt.internal.compiler.lookup.WildcardBinding.erasure()` for intersection types and was not prepared for the receiver being returned as is. 

What I want to do is:

If the element type of anewarray happens to have parameterizations, reach for the erasure which could be the class (jlO in this case) or the first interface's erasure so that the sanity checks in OperandStack are not overstrict and fail.

(Aside: Why they currently blow up itself is a good question. Where `org.eclipse.jdt.internal.compiler.codegen.OperandStack.xastore()` blows up currently with the test case, the array is seen to be of element type `java.lang.Object & A<java.lang.Object> & B<java.lang.Object>` and the stored value is seen to be of type `D` which implements `implements : A<AA>, B<java.lang.Object>`

See the parameterization mismatch for A (object in array and AA in value being stored) - I don't know if this difference itself is legitimate. )

What do you advise ?

srikanth-sankaran:
A strawman fix - a bit coarse.

`org.eclipse.jdt.internal.compiler.lookup.TypeBinding.isReifiable()` answers true for `IntersectionTypeBinding18` ! Corrected.

srikanth-sankaran:
I think I am happy with this fix + a follow on defect against `IntersectionTypeBinding18.erasure()` returning the receiver under some circumstances making the return value non-reifiable.

PR #3673, opened 2025-02-04
Stop using JLS3 in converter tests

akurtakov:
AST overrides to JLS8 so this change makes the tests easier to understand.
By using first supported JLS version explicitly (parser creation) or implicitly (calling runConversion without astLevel) tests have less silently overriden parts.

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverter15Test.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverter15Test.java
@@ -48,7 +48,7 @@ public class ASTConverter15Test extends ConverterTestSetup {
 	@Override
 	public void setUpSuite() throws Exception {
 		super.setUpSuite();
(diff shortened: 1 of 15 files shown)


srikanth-sankaran:
@akurtakov - this commit seems to cause a number of failures in `org.eclipse.jdt.core.tests.dom.ASTConverterRecoveryTest` - can you take a look ? Many (but many not be all) failures from https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3675 are connected to this. Thanks in advance

srikanth-sankaran:
> @akurtakov - this commit seems to cause a number of failures in `org.eclipse.jdt.core.tests.dom.ASTConverterRecoveryTest` - can you take a look ? Many (but many not be all) failures from #3675 are connected to this. Thanks in advance

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3678

srikanth-sankaran:
This jdt.core PR looks strange to me. It says 6 checks passed. How come a full tests runs is not there ??!

akurtakov:
I would guess that Jenkins was temporary down and thus missed the event to verify this PR.

Issue #3665, opened 2025-02-03
[Records] Simplify ECJ's grammar for record classes 


srikanth-sankaran:
Consider this grammar fragment:

```
RecordBody ::= '{' RecordBodyDeclarationopt '}'
/.$putCase consumeRecordBody(); $break ./
/:$readableName RecordBody:/
/:$compliance 14:/

RecordBodyDeclarationopt ::= $empty
/.$putCase consumeEmptyRecordBodyDeclaration(); $break ./
RecordBodyDeclarationopt -> RecordBodyDeclarations
/:$readableName RecordBodyDeclarationopt:/
/:$compliance 14:/

RecordBodyDeclarations ::= RecordBodyDeclaration
RecordBodyDeclarations ::= RecordBodyDeclarations RecordBodyDeclaration
/.$putCase consumeRecordBodyDeclarations(); $break ./
/:$readableName RecordBodyDeclarations:/
/:$compliance 14:/

RecordBodyDeclaration ::=  ClassBodyDeclaration
/.$putCase consumeRecordBodyDeclaration(); $break ./
RecordBodyDeclaration ::=  CompactConstructorDeclaration
/.$putCase consumeRecordBodyDeclaration(); $break ./
/:$readableName RecordBodyDeclaration:/
/:$compliance 14:/

CompactConstructorDeclaration ::= CompactConstructorHeader MethodBody
/.$putCase consumeCompactConstructorDeclaration(); $break ./
/:$readableName CompactConstructorDeclaration:/
/:$compliance 14:/

CompactConstructorHeader ::= CompactConstructorHeaderName MethodHeaderThrowsClauseopt
/.$putCase consumeCompactConstructorHeader(); $break ./
/:$readableName CompactConstructorDeclaration:/
/:$compliance 14:/

CompactConstructorHeaderName ::= Modifiersopt 'Identifier'
/.$putCase consumeCompactConstructorHeaderName(); $break ./
CompactConstructorHeaderName ::= Modifiersopt TypeParameters 'Identifier'
/.$putCase consumeCompactConstructorHeaderNameWithTypeParameters(); $break ./
/:$readableName CompactConstructorHeaderName:/
/:$compliance 14:/
```

The key difference between a record's body and a class's body is that the former may have compact constructor declarations.
It would seem this whole fragment could be eliminated and we could simply use `ClassBody` non-terminal in place of `RecordBody` if
[...]

PR #3675, opened 2025-02-04
Grammar simplification for Record classes: Merge RecordBody with ClassBody

srikanth-sankaran:
* Partial fix for https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3665


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/problem/messages.properties
```diff
@@ -1102,6 +1102,7 @@
 1766 = Illegal modifier for local enum {0}; no explicit modifier is permitted
 1767 = The record {0} cannot be the superclass of {1}; a record is final and cannot be extended
 1768 = Erasure incompatibility in argument {0} of canonical constructor in record
+1769 = A canonical constructor is allowed only in record classes
```
stephan-herrmann, line 1105: @srikanth-sankaran  shouldn't this read "A compact constructor ..."? 
(Canonical constructors don't have specific syntax, right?)

srikanth-sankaran: Indeed, thanks for catching it - will include a fix soon,

PR #3697, opened 2025-02-06
Fix source type binding key on windows

BoykoAlex:
On Windows since path separator is \ the binding key ends up being Lp/<path to the java file>~X, i.e. includes file path. On Linux and Mac due to separator being / the key is Lp/X; as expected.
PR should fix the binding key on Windows


BoykoAlex:
@jukzi @martinlippert I have created a new PR since #3416 is corrupted somehow now... Could you please review it?

jukzi (approved):

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/dom/StandAloneASTParserTest.java
```diff
@@ -16,6 +16,11 @@
  *******************************************************************************/
 package org.eclipse.jdt.core.tests.dom;
 
+import static org.junit.Assert.assertFalse;
```
jukzi, line 19: 00:06:23.884  [ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:4.0.11:compile (default-compile) on project org.eclipse.jdt.core.tests.compiler: Compilation failure: Compilation failure: 
00:06:23.884  [ERROR] /home/jenkins/agent/workspace/eclipse.jdt.core-Github_PR-3697/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/dom/StandAloneASTParserTest.java:[19] 
00:06:23.884  [ERROR] 	import static org.junit.Assert.assertFalse;
00:06:23.884  [ERROR] 	              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
00:06:23.884  [ERROR] The import org.junit.Assert.assertFalse is never used
00:06:23.884  [ERROR] /home/jenkins/agent/workspace/eclipse.jdt.core-Github_PR-3697/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/dom/StandAloneASTParserTest.java:[20] 
00:06:23.884  [ERROR] 	import static org.junit.Assert.assertNotNull;
00:06:23.884  [ERROR] 	              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
00:06:23.884  [ERROR] The import org.junit.Assert.assertNotNull is never used
00:06:23.884  [ERROR] /home/jenkins/agent/workspace/eclipse.jdt.core-Github_PR-3697/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/dom/StandAloneASTParserTest.java:[21] 
00:06:23.884  [ERROR] 	import static org.junit.Assert.assertNull;
00:06:23.884  [ERROR] 	              ^^^^^^^^^^^^^^^^^^^^^^^^^^^
00:06:23.884  [ERROR] The import org.junit.Assert.assertNull is never used
00:06:23.884  [ERROR] /home/jenkins/agent/workspace/eclipse.jdt.core-Github_PR-3697/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/dom/StandAloneASTParserTest.java:[22] 
00:06:23.884  [ERROR] 	import static org.junit.Assert.assertTrue;
00:06:23.884  [ERROR] 	              ^^^^^^^^^^^^^^^^^^^^^^^^^^^
00:06:23.884  [ERROR] The import org.junit.Assert.assertTrue is never used
00:06:23.884  [ERROR] 4 problems (4 errors)
00:06:23.884  [ERROR] -> [Help 1]

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/dom/StandAloneASTParserTest.java
```diff
@@ -1943,7 +1943,63 @@ class com {
...
+
+			parser.createASTs(new String[] {absPath}, null, new String[0], requestor, null);
+
+			assertNotNull("No binding", bindings[0]);
+			assertEquals("Wrong type of binding", IBinding.TYPE, bindings[0].getKind());
+			ITypeBinding typeBinding = (ITypeBinding) bindings[0];
+			assertEquals("Wrong binding", "p.X", typeBinding.getQualifiedName());
+			assertEquals("Wrong binding key", "Lp/X;", typeBinding.getKey());
+		} finally {
+			file.delete();
```
jukzi, line 2000: please also clean up packageDir, rootDir

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/dom/StandAloneASTParserTest.java
```diff
@@ -1943,7 +1943,63 @@ class com {
         assertNotNull(ctorBinding);
         IAnnotationBinding[] annotations = ctorBinding.getAnnotations();
         assertNotNull(annotations);
-}
+	}
+
+	public void testBindingKey() throws IOException {
+		File rootDir = new File(System.getProperty("java.io.tmpdir"));
```
jukzi, line 1949: You could shrink the code by using java.nio.file.Files.createTempDirectory

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/dom/StandAloneASTParserTest.java
```diff
@@ -1943,7 +1943,63 @@ class com {
...
+		final IBinding[] bindings = new IBinding[1];
+
+		String contents =
+			"package p;\n" +
+			"public class X {\n" +
+			"}";
+
+		File packageDir = new File(rootDir, "p");
+		packageDir.mkdir();
+		File file = new File(packageDir, "X.java");
```
jukzi, line 1966: would be shorter to use java.nio.file.Files.writeString

jukzi:
using File.separatorChar matches org.eclipse.jdt.internal.core.NameLookup.findPackageFragment(IPath) to this looks like a good change!

stephan-herrmann:
Please explain how the binding key relates to filesystem / operating system. 

From the top of my head I only gather these contracts:
* keys should be unique
* keys should harmonize with the implementation of BindingKeyResolver / BindingKeyParser

I guess the central documentation about these things can be found at `org.eclipse.jdt.core.dom.IBinding.getKey()`, which in particular says:

> The exact details of how the keys are generated is unspecified. However, it is a function of the following information:

So which clients make assumptions about the meaning of `/` vs. `\`?

BoykoAlex:
Without the fix the binding from the added unit test looks like below:
`Lp/C:\Users\sasha\AppData\Local\Temp\jdt-test-binding-key17138182034659599285\p\X~X;`
Rather than:
`Lp/X;`

Thus it clearly looks at the OS file path when computing the binding key at the moment. The code I'm correcting used to look like this:
```java
int start = CharOperation.lastIndexOf('/', this.fileName) + 1;
```
From the above I've concluded that we do expect file path separators to be the `fileName` field...

Therefore, I felt that OS path separator would be an obvious fix as it would make the binding key platform independent.

Perhaps the expectation is that `fileName` field suppose to indeed hold just the file name rather than a file path? It'd probably be a riskier fix then...

stephan-herrmann:
> Without the fix the binding from the added unit test looks like below: `Lp/C:\Users\sasha\AppData\Local\Temp\jdt-test-binding-key17138182034659599285\p\X~X;` Rather than: `Lp/X;`
> 
> Thus it clearly looks at the OS file path when computing the binding key at the moment. The code I'm correcting used to look like this:
> 
> ```java
> int start = CharOperation.lastIndexOf('/', this.fileName) + 1;
> ```
> 
> From the above I've concluded that we do expect file path separators to be the `fileName` field...
> 
> Therefore, I felt that OS path separator would be an obvious fix as it would make the binding key platform independent.
> 
> Perhaps the expectation is that `fileName` field suppose to indeed hold just the file name rather than a file path? It'd probably be a riskier fix then...

Thanks, this helps to see where you are coming from.

Unfortunately, the value of `SourceTypeBinding.fileName` could mean a lot of different things, depending on the context where this thing was created. The immediate source is always `CompilationResult.getFileName()`, but if you look for how `CompilationResult` is instantiated, you'll find a variety of different inputs. One possible source is `IDependent.getFileName()` which *admits* (but doesn't require) OS-specific paths (by way of saying "e.g."), but then `org.eclipse.jdt.internal.core.builder.SourceFile.getFileName()` uses `IPath.toString()` which explicitly specifies 
> The same separator, "/", is used on all platforms.

So, if we get a file name from a SourceFile, then looking for `File.separatorChar` will work only on *nix.

For that reason I'm afraid this PR only moves a problem from one place to another.

stephan-herrmann:
For the p.o.v. of the batch compiler see #1458

IMHO, using `\` in more locations takes us farther away from any consistent strategy.

stephan-herrmann (changes requested):
Blocking the PR against accidental merge, until we have a better understanding of the risks.

stephan-herrmann:
I have instrumented the code to print out names of those classes that contribute the fileName into `CompilationResult` (from where it is further transferred into SourceTypeBinding). This is my (possibly incomplete) result from running AllJavaModelTests:
```
org.eclipse.jdt.internal.compiler.batch.CompilationUnit
org.eclipse.jdt.internal.compiler.classfmt.ClassFileReader
org.eclipse.jdt.internal.core.BasicCompilationUnit
org.eclipse.jdt.internal.core.builder.SourceFile
org.eclipse.jdt.internal.core.ClassFile
org.eclipse.jdt.internal.core.CompilationUnit
org.eclipse.jdt.internal.core.CompilationUnit$1
org.eclipse.jdt.internal.core.hierarchy.HierarchyBuilder$1
org.eclipse.jdt.internal.core.jdom.CompilationUnit
org.eclipse.jdt.internal.core.search.matching.PossibleMatch
org.eclipse.jdt.internal.core.SourceTypeElementInfo
org.eclipse.jdt.internal.core.util.ResourceCompilationUnit
org.eclipse.jdt.internal.eval.EvaluationContext$2
org.eclipse.jdt.internal.eval.Evaluator$1
```
Some of these indeed use the platform specific separator, but (at least) the following use `/` on all platforms:
* org.eclipse.jdt.internal.core.builder.SourceFile
* org.eclipse.jdt.internal.core.CompilationUnit
* org.eclipse.jdt.internal.core.util.ResourceCompilationUnit

ClassFileReader even seems to be used with inconsistent strategies depending on call path. ClassFileReader, representing a class file, looks irrelevant wrt SourceTypeBinding, but actually a ClassFileReader plus a BinaryTypeConverter indeed create a TypeDeclaration and then a SourceTypeBinding.

Another inconsistency: The field `ReferenceBinding.fileName` is shared with SourceTypeBinding and BinaryTypeBinding. At a cursory look BinaryTypeBinding leans towards platform independent `/`.

Actually I'm surprised these inconsistencies don't blow up more frequently.

While ideally we should have a clear strategy, I could think that for the issue in SourceTypeBinding the best solution would be to **accept both styles**, i.e., check for presence of `/` and `\`.  Cf. what we already do in `org.eclipse.jdt.internal.core.BasicCompilationUnit.getMainTypeName()`. 
Does Windows allow `/` as a part of a name segment (i.e., with no separator semantics)? If you are crude you can create files on unix using `\` within a segment.

That method in BasicCompilationUnit actually gives a valuable hint: it was added with this comment:
> Handle main type name in compilation unit from .class file in jar
File names inside a jar always use `/`, no?

This leads to the question: do any clients rely on absolute paths, or would it help to "normalize" all fileNames to those paths that correspond to packages, smth like source-folder-relative paths?

Issue #3700, opened 2025-02-06
ecj complains about intermediate constructor invocation inside lambda


stephan-herrmann:
Given this test case from https://bugs.openjdk.org/browse/JDK-8333313
```java
 interface Foo {
    void foo();
}

public class Main {
    static int check = 0;

    class Test {
        Test() {}
        Test(int a) {
            class InnerLocal {
                int a = 1;
            }
            Foo lmb = () -> {
                Main.check = new InnerLocal() {
                    public int a() {
                        return this.a;
                    }
                }.a();
            };
            lmb.foo();
            this();
        }
    }
} 
```

While javac 24 ea accepts this, ecj complains:
```
        Main.check = new InnerLocal() {
                    public int a() {
                        return this.a;
                    }
                }.a();
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
No enclosing instance of type Main is available due to some intermediate constructor invocation
```

stephan-herrmann:
Fixed by https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3702

PR #3702, opened 2025-02-06
ecj complains about intermediate constructor invocation inside lambda

stephan-herrmann:
Shy start to unify type declaration and lambda in some respect:
+ both need to manage access to enclosing instances
+ hook the pulled-up method into LE.analyseCode() to fix the bug

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3700

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeOrLambda.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeOrLambda.java
@@ -1,0 +1,65 @@
+/*******************************************************************************
+ * Copyright (c) 2025 GK Software SE, and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * This is an implementation of an early-draft specification developed under the Java
+ * Community Process (JCP) and is made available for testing and evaluation purposes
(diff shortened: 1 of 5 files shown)


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeOrLambda.java
```diff
@@ -0,0 +1,65 @@
...
+import org.eclipse.jdt.internal.compiler.lookup.SourceTypeBinding;
+
+/**
+ * Common properties of type declarations and lambda expressions.
+ * <p>
+ * In terms of access to enclosing instances both type declarations and lambda expressions
+ * are barriers which need to be bridged with synthetic arguments (and fields in the case of types).
+ * </p>
+ */
+public interface TypeOrLambda {
```
stephan-herrmann, line 33: @srikanth-sankaran what do you think of this new union type? :smile:

When dealing with enclosing instances time and again I forgot about lambdas and then I had to figure out how to re-implement some logic that already exists _very_ similarly for type declarations, but then ended up looking quite differently for lambdas.

Have you, too, ever wondered where to put common code of these two constructs? I mean, haven't lambdas been presented as an evolution of anonymous classes?

Right now there's not a whole lot of code here, but perhaps this will help us factor out some more stuff.

Just wanted to let you know. Perhaps you had different plans with a similar effect?

srikanth-sankaran: > Just wanted to let you know. Perhaps you had different plans with a similar effect?

Thanks, this cane be useful, I will bear this in mind. Haven't dealt with these concerns in a while, so don't recall how sorely such was missed.

Issue #3370, opened 2024-11-29
AbstractImageBuilder and ICompilerFactory not passing a CompilationProgress


mickaelistria:
From an implementation of compiler (either with ECJ or Javac as backend), it's not possible to check the cancellation status directly because no CompilationProgress is passed to the compiler. It seems possible to easily-ish implementation such a CompilationProgress in AbstractImageBuilder delegating to BuildNotifier, and to improve the ICompilerFactory interface so it forwards such CompilationProgress object.

mickaelistria:
cc @testforstephen

jukzi:
relates to https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3317
The Compiler uses to call org.eclipse.jdt.internal.core.builder.BuildNotifier.checkCancelWithinCompiler()
![image](https://github.com/user-attachments/assets/dd20d568-d25a-4d7b-ad5b-f80c944658e2)

mickaelistria:
Actually, even that last call as part of `acceptResult` is not usually the one that cancel the *compilation* (`AbstractImageBuilder` orchestrates the compilation, but the compiler doesn't really know of whether it's called by a builder or some other way to fully rely on this assumption). The operation that is most often aborting the build is the one from `NameEnvironment.findClass()`. So in most cases, the name lookup is responsible for cancellation of the compiler, which seems

jukzi:
Compilation of a single source takes usually only milliseconds. Why would you like to check for cancel more often?

mickaelistria:
Because not all cases are Compilation of a single source. Import jdt.core in a fresh workspace and you get more than a thousand files built together. And this takes about a dozen of seconds with ECJ.

PR #3711, opened 2025-02-07
extends ICompilerFactory by CompilationProgress #3370

jukzi:
- Allows compiler implementations to check for cancel and to report progress.
- Implements progress report and explicit cancel check before progressing AnnotationProcessor

implements https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3370
image

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/compiler/ICompilerFactory.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/compiler/ICompilerFactory.java
@@ -13,6 +13,7 @@
 
 package org.eclipse.jdt.internal.compiler;
 
+import org.eclipse.jdt.core.compiler.CompilationProgress;
 import org.eclipse.jdt.internal.compiler.env.INameEnvironment;
 
 /**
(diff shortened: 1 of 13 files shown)


jukzi:
@mickaelistria i think this is what you requested and we should synchronize for a date

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/compiler/ICompilerFactory.java
```diff
@@ -28,11 +29,15 @@ public interface ICompilerFactory {
...
+	 * @param compilationProgress - the CompilationProgress to be used
 	 * @return the new compiler instance
+	 * @see CompilationProgress
 	 */
-	public Compiler newCompiler(final INameEnvironment environment,
-			final IErrorHandlingPolicy policy,
-			final CompilerConfiguration compilerConfig,
-			final ICompilerRequestor requestor,
-			final IProblemFactory problemFactory);
+	public Compiler newCompiler(INameEnvironment environment,
```
mickaelistria, line 36: Would it be possible for a 1st iteration to add the new construtor and keep the other as deprecated, and plan removal for ~4 days later?
We should be able to update promptly, but we'd rather avoid broken builds if possible.

mickaelistria (commented):
Thanks. This proposal is very welcome and good for us!
Please just see proposal inline to make adoption smoother.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/compiler/ICompilerFactory.java
jukzi: The workflow would be ok for me, but it feels like too much effort to actually call this or that constructor depending on availability. So if you insist on it would it be possible that you add that extra code?

mickaelistria: OK, I understand the trouble and agree it's probably not worth it. I'm fine merging it at your own convenience; we'll easily spot the issue in the javac-based branch afterwards, and even if build is broken for a couple of days, that wouldn't be a strong issue; so let's not make it a cause of a delay upstream.

jukzi: would you prefer a merge asap or on m1?

mickaelistria: ASAP would work fine.

Issue #3369, opened 2024-11-29
[JEP 488][Preview][Primitive Patterns] ECJ crashes compiling switch over void typed selector expression


srikanth-sankaran:
Found by code inspection and white box testing:

ECJ crashes compiling the following code; javac complains `error: the switch statement does not cover all possible input values`

```
public class X {
	public static void main(String[] args) {
		switch (main(null)) {
		
		}
	}
}
```

Stack trace on master (problem can be seen from 4.34 M2) 

```
Caused by: java.util.EmptyStackException
	at java.base/java.util.Stack.peek(Stack.java:103)
	at org.eclipse.jdt.internal.compiler.codegen.OperandStack.peek(OperandStack.java:123)
	at org.eclipse.jdt.internal.compiler.codegen.CodeStream.dup(CodeStream.java:1082)
	at org.eclipse.jdt.internal.compiler.ast.SwitchStatement.generateCodeSwitchPatternPrologue(SwitchStatement.java:850)
	at org.eclipse.jdt.internal.compiler.ast.SwitchStatement.generateCode(SwitchStatement.java:708)
	at org.eclipse.jdt.internal.compiler.ast.AbstractMethodDeclaration.generateCode(AbstractMethodDeclaration.java:395)
	at org.eclipse.jdt.internal.compiler.ast.AbstractMethodDeclaration.generateCode(AbstractMethodDeclaration.java:331)
	at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.generateCode(TypeDeclaration.java:773)
	at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.generateCode(TypeDeclaration.java:843)
	at org.eclipse.jdt.internal.compiler.ast.CompilationUnitDeclaration.generateCode(CompilationUnitDeclaration.java:403)
	at org.eclipse.jdt.internal.compiler.Compiler.resolve(Compiler.java:1119)
	at org.eclipse.jdt.internal.compiler.Compiler.resolve(Compiler.java:1155)
	at org.eclipse.jdt.internal.core.CompilationUnitProblemFinder.process(CompilationUnitProblemFinder.java:292)
```

srikanth-sankaran:
[...]

PR #3722, opened 2025-02-11
[JEP 488][Preview][Primitive Patterns] ECJ crashes compiling switch over void typed selector expression

stephan-herrmann:
+ report error rather than crashing

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3369

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/SwitchStatement.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/SwitchStatement.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2023 IBM Corporation and others.
+ * Copyright (c) 2000, 2025 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -7,6 +7,10 @@
  * https://www.eclipse.org/legal/epl-2.0/
  *
  * SPDX-License-Identifier: EPL-2.0
+ *
+ * This is an implementation of an early-draft specification developed under the Java
+ * Community Process (JCP) and is made available for testing and evaluation purposes
+ * only. The code is not compatible with any specification of the JCP.
  *
  * Contributors:
  *     IBM Corporation - initial API and implementation
@@ -552,6 +556,10 @@ public class SwitchStatement extends Expression {
 				this.expression.computeConversion(upperScope, expressionType, expressionType);
 				checkType: {
 					if (expressionType.isBaseType()) {
+						if (expressionType.id == TypeIds.T_void) {
+							upperScope.problemReporter().illegalVoidExpression(this.expression);
+							break checkType;
+						}
 						if (JavaFeature.PRIMITIVES_IN_PATTERNS.isSupported(compilerOptions)) {
 							upperScope.referenceContext().compilationResult().usesPreview = true;
 							this.isPrimitiveSwitch = true;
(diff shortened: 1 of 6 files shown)


org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/SwitchExpressionsYieldTest.java
```diff
@@ -7679,14 +7679,14 @@ public static void main(String[] args) {
...
-						+ "	case 1.0 -> System.out.println(d);\n"
-						+ "	     ^^^\n"
-						+ "Case constant of type double is incompatible with switch selector type void\n"
-						+ "----------\n");
+						"----------\n" +
+						"1. ERROR in X.java (at line 7)\n" +
+						"	switch (foo()) {\n" +
+						"	        ^^^^^\n" +
+						"Expression must return a value\n" +
+						"----------\n");
```
srikanth-sankaran, line 7689: 1. Are both arms of the ternary identical now ? They appear so. Could they be collapsed together ? 
2. With the message `Expression must return a value` I feel the quality of the diagnostic actually degrades in this context. How about we change the message to `Expression has no value` - ("to switch on", "to synchronize with" etc left to be understood) ?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/SwitchStatement.java
```diff
@@ -552,6 +556,10 @@ public void resolve(BlockScope upperScope) {
 				this.expression.computeConversion(upperScope, expressionType, expressionType);
 				checkType: {
 					if (expressionType.isBaseType()) {
+						if (expressionType.id == TypeIds.T_void) {
+							upperScope.problemReporter().illegalVoidExpression(this.expression);
+							break checkType;
+						}
```
srikanth-sankaran, line 562: Admittedly, this is even more corner case: What about 'V'oid methods ? should they be handled here too ?  (Actually what can a Void method return other than null ? I have poor understanding here)

srikanth-sankaran (approved):
No major issues. Three minor potential follow up comments.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/SwitchStatement.java
stephan-herrmann: Indeed I can't see how switching over `Void` could ever do anything useful (you'd get NPE at best). Yet doing so would be legal code, in contrast to `void`.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/SwitchExpressionsYieldTest.java
stephan-herrmann: ad (1): your eagle eyes spotted correctly :)

stephan-herrmann: ad (2): I agree, that message is a bit odd. 
"has no value" makes me think of "worthless" :) 
Perhaps we can agree on "yields no value"?

Let me see if that change will cause a landslide in our tests.

stephan-herrmann: > Let me see if that change will cause a landslide in our tests.

Quite to the contrary. I can't find any tests covering the 2 pre-existing uses of this error message. I'll add a couple.

PR #3730, opened 2025-02-12
CompilationUnit does not contains modifier for the ImplicitTypeDeclaration

### LINKED ISSUE #3696: [24] CompilationUnit does not contains modifier for the ImplicitTypeDeclaration
ref: https://openjdk.org/jeps/495
JLS Changes : https://cr.openjdk.org/~gbierman/jep495/jep495-20241101/specs/simple-source-files-instance-main-methods-jls.html#jls-8.1.8

Summery
The `CompilationUnit.types` of `ImplicitTypeDeclaration` does not contain modifier. In light of JEP-495, the `ImplicitTypeDeclaration` can have the `final` modifier. However, the modifier is accessible through `ITypeBinding`.

Reproducer
========

```
public void test001() throws CoreException {
		ASTParser astParser = ASTParser.newParser(getAST24());
	    Map<String, String> options = new HashMap<>();
	    options.put(JavaCore.COMPILER_COMPLIANCE, "24");
	    options.put(JavaCore.COMPILER_SOURCE, "24");

	    astParser.setCompilerOptions(options);
	    astParser.setEnvironment(new String[] {}, new String[] {}, new String[] {}, true);
	    astParser.setUnitName("Example.java");
	    astParser.setResolveBindings(true);
	    astParser.setBindingsRecovery(true);

	    String source ="""
		    		String greeting() {
						return "Hello, World!";
					}
					void main() {
					    System.out.println(greeting());
					}
	    		""";

	    astParser.setSource(source.toCharArray());

	    CompilationUnit compilationUnit = (CompilationUnit) astParser.createAST(null);
	    ImplicitTypeDeclaration a = (ImplicitTypeDeclaration) compilationUnit.types().get(0);
	    assertTrue(Modifier.isFinal(a.getModifiers()));

	    ITypeBinding aBinding = a.resolveBinding();
	    assertTrue(Modifier.isFinal(aBinding.getModifiers()));
	}
```

subyssurendran666:
## What it does
All the necessary informations has beed added to the parent ticket.
This PR closes #3696 

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverter_24Test.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverter_24Test.java
@@ -1,0 +1,92 @@
+/*******************************************************************************
+ * Copyright (c) 2025 IBM Corporation and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.jdt.core.tests.dom;
+
+import java.util.HashMap;
+import java.util.Map;
+import junit.framework.Test;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.jdt.core.ICompilationUnit;
+import org.eclipse.jdt.core.JavaCore;
+import org.eclipse.jdt.core.dom.AST;
+import org.eclipse.jdt.core.dom.ASTParser;
+import org.eclipse.jdt.core.dom.CompilationUnit;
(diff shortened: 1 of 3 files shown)


mpalat (commented):
@subyssurendran666 please take a look at Record also - here final is implicit. Let's check how its implemented there

subyssurendran666:
> @subyssurendran666 please take a look at Record also - here final is implicit. Let's check how its implemented there

I have analysed the Record and which is Implicitly "public final", ie `record Eclipse(String name, int age) {} ` which is equivalent to `public final record Eclipse(String name, int age) {}`.

There is no modifiers property attached to the RecordDeclaration if it is set implicitly. Therefore, the same approach can be followed here. We can modify the Javadoc of the ImplicitTypeDeclaration node.

mpalat:
> > @subyssurendran666 please take a look at Record also - here final is implicit. Let's check how its implemented there
> 
> I have analysed the Record and which is Implicitly "public final", ie `record Eclipse(String name, int age) {} ` which is equivalent to `public final record Eclipse(String name, int age) {}`.
> 
> There is no modifiers property attached to the RecordDeclaration if it is set implicitly. Therefore, the same approach can be followed here. We can modify the Javadoc of the ImplicitTypeDeclaration node.

Great - I agree

PR #3731, opened 2025-02-12
Prevent initial build-path errors on new external-folders in container

HannesWell:
## What it does
When adding an external-folder ('library-folder') to a classpath-container for the first time the initial build (without a rebuild, e.g. through auto-build) leads to a build-path (validation) error.
This happens for example when adding to a MANIFEST.MF a new requirement on a bundle that's a directory and not a jar and the PDE 'required-Plugins' classpath container is updated.

The reason is that the validation checks the linked folders created for external-folders. Calling `ExternalFolderChange.updateExternalFoldersIfNecessary` before the validation avoids these false-positive errors.

## How to test

The use-case in our code-base is non-trivial, but I can try to derive a dedicated test-case for it once I know that this doesn't break anything and is generally a reasonable change.

## Author checklist


HannesWell:
@iloveeclipse or @jukzi can you have a look at this?

jjohnstn (approved):
LGTM

PR #3741, opened 2025-02-15
Merge master (4.35 M3)  into BETA_JAVA24

stephan-herrmann:
Merge needed in reaction to version bump in #3711


stephan-herrmann:
Pushed directly

PR #3771, opened 2025-03-06
Problem arguments must be consistent

### LINKED ISSUE #3556: Problem arguments must be consistent
Several methods in `ProblemReporter` pass inconsistent problem arguments to `handle(..)`
* one variant of problem arguments is passed as `NoArgument` or similar
* the other variant passes a non-empty array

This will blow up whenever the error message (containing placeholders `{0}` ...) should be rendered using `NoArgument`: will try to read `[0]` from the 0-length array => AIOOBE. One case was observed in the context in JSR 199, where each and every such occurrence can throw.

subyssurendran666:
## What it does
Due to Problem argument in consistency, AIOOBE occurred sometimes.  This handles the AIOOBE in the ProblemHandler.

This PR closes #3556 

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/problem/ProblemHandler.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/problem/ProblemHandler.java
@@ -223,6 +223,13 @@ public void handle(
 	ReferenceContext referenceContext,
 	CompilationResult unitResult) {
 
+	if (problemArguments.length == 0) {
+		problemArguments = new String[] { "N/A" }; //$NON-NLS-1$
+	}
+
+	if (messageArguments.length == 0) {
+		messageArguments =new String[] { "N/A" }; //$NON-NLS-1$
+	}
 	this.handle(
 		problemId,
 		problemArguments,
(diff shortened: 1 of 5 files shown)


mpalat (changes requested):
@subyssurendran666 Even if I comment the fix, the tests pass. You may want to recheck

subyssurendran666:
> @subyssurendran666 Even if I comment the fix, the tests pass. You may want to recheck

Could you please re-spin the PR ?

stephan-herrmann:
The intention behind https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3556 was to ensure that for every invocation of `handle(..)` the lengths of `problemArguments` and `messageArguments` match. Inventing any missing values inside `handle(..)` is not a good solution.

Note for testing: we first need to determine which client uses `problemArguments` and which one uses `messageArguments`.

subyssurendran666:
Thank you for taking the time to look into this and for clarifying the intent behind #3556. That makes sense. Patching the mismatch inside `handle(..)` would only mask the symptom rather than address the actual root cause.

I will go back and trace which caller is passing `problemArguments` and which one is passing `messageArguments`. This should help me identify the actual call site where the argument lengths are getting out of sync, and I will fix it there instead.
I will update the PR once I have a proper root cause fix and will also revisit accordingly.
Thanks again for the guidance and for pointing me in the right direction.

PR #3796, opened 2025-03-14
ecj-javac difference when var hits diamond with capture

mpalat:
WIP - DRAFT

				interface I<T> {}
				record R<T>(T t) {}

				public class X  {

				    static <T> boolean foo(R<T> y) {
				       if (y.t() instanceof R(var v)) {
				            var t = new R<>(v);
				            R<Object> r = t; //error  shown [Type mismatch: cannot convert from R<capture#2-of ?> to R<Object>]
				            return true;
				        }
				        return false;
				    }
				        System.out.println(foo(new R<R<Object>>(new R<>(new X()))));
				    }
				}

## What it does
Test case that shows difference between javac and ecj behaviour.  javac compiles and prints true while running while ecj gives the following error:
"Type mismatch: cannot convert from R<capture#2-of ?> to R<Object>"

## How to test
The code above is a reproducible testcase. Currently the fix is in the works. The patch contains the test case only as of now.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/RecordPatternTest.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/RecordPatternTest.java
@@ -28,7 +28,7 @@ public class RecordPatternTest extends AbstractRegressionTest9 {
 	static {
 //		TESTS_NUMBERS = new int [] { 40 };
 //		TESTS_RANGE = new int[] { 1, -1 };
-//		TESTS_NAMES = new String[] { "testRecPatExhaust018" };
+//		TESTS_NAMES = new String[] { "testRecPatTypeInfer_001" };
 	}
 	private String extraLibPath;
 	public static Class<?> testClass() {
@@ -4920,5 +4920,30 @@ public class RecordPatternTest extends AbstractRegressionTest9 {
 				"""
 			},
 			"012012345678");
+	}
+	public void testRecPatTypeInfer_001() {
+		runConformTest(new String[] {
+				"X.java",
+				"""
+				interface I<T> {}
+				record R<T>(T t) {}
+
+				public class X  {
+
+				    static <T> boolean foo(R<T> y) {
+				       if (y.t() instanceof R(var v)) {
+				            var t = new R<>(v);
+				            R<Object> r = t;
+				            return true;
+				        }
+				        return false;
+				    }
+				    public static void main(String argv[]) {
+				        System.out.println(foo(new R<R<Object>>(new R<>(new X()))));
+				    }
+				}
+				"""
+				},
+				"true");
 	}
 }


stephan-herrmann:
In the test case I see three different things called "T". 
Has it been written somewhere that all type parameters are called "T"?
Do yourself a favor and invest those 10 seconds to 'invent' different names. 
How else do you want to keep track which is which?

stephan-herrmann:
Preliminary comments:

We perform record type inference with an expression of type `<T>`, an unbounded type variable.

From the items of 18.5.5 I see only these as applicable:
* 2: so an initial bound set is created (which however is empty due to lack of declared bounds)
* 5: resolution without sufficient information creates a CaptureBinding18 `Z#0-T#0`
* 6: I see upwards projection converting `R<Z#0-T#0>` into `R<?>`.

Note in particular that items 3 and 4 are not really applicable. There's one oddity in item 3: it's all about the case of T being a parameterized type, from which betas are derived and how further constraints are created for them. Hidden in that list of inapplicable steps there are two bullets for the "outer else", i.e., when T is not a parameterized type.  But even that bullet is moot for our case, as it only derives further constraints from a type variable's bound, but our type variable has no bounds.

Let's assume for a moment that `R<?>` is the correct result from 6.  Since 9e5215a we immediately capture that type to the `R<capture#2-of ?>` we see in the error message. We might discuss with @srikanth-sankaran  the exact capturing story. See, e.g., that a few lines down in `RecordPattern.resolveType()` we say `this.resolvedType.capture(..)` on the already captured type :stuck_out_tongue: 

But none of this would transform the wildcard (captured or not) into type `j.l.Object` as required for accepting the program.

Next idea: perhaps somewhere down the line we treat `<T>` as having no bound, where perhaps we should regard it as `<T extends Object>`, but even adding explicit `extends Object` to all type parameters, doesn't help.

Ergo: for the moment I have no idea, what could lead us to inferring `R<Object>`.

iloveeclipse:
This seem to be forgotten?
Removing 4.36 milestone.

iloveeclipse:
Moving out of 4.38 milestone.

stephan-herrmann:
Using instrumented javac I don't see any type inference working on the record pattern `R(var v)`. I can only assume that `v` is resolved as `j.l.Object`.

This is a different entry into inference as usual. @srikanth-sankaran could you easily wire it up with your tracing?

srikanth-sankaran:
> This is a different entry into inference as usual. @srikanth-sankaran could you easily wire it up with your tracing?

Will do - yes, we have tracing of generic method invocations. Functional type inference and record pattern inference are different entry points.

stephan-herrmann:
> > This is a different entry into inference as usual. @srikanth-sankaran could you easily wire it up with your tracing?
> 
> Will do - yes, we have tracing of generic method invocations. Functional type inference and record pattern inference are different entry points.

* Meanwhile I see https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5209 :)

srikanth-sankaran:
> Let's assume for a moment that `R<?>` is the correct result from 6

I can confirm that is what javac computes. I am having a bit of bother sprinkling in instrumentation into `com.sun.tools.javac.comp.Infer#instantiatePatternType` - the code there does not map neatly to JLS 18.5.5 - plus even where it maps, some of entities are anonymous and so don't leave a trace in the output: (see `Instantiated pattern variables : ` below, there is actually one entry, but that is nameless so doesn't show up)

This is what I have at the moment which is not super useful:

```

---------------- Inferring types for record pattern: R<T>----------------



---------------- Initial Bound Set (B0): ---------------

        Inference variable = T 

Instantiated pattern variables : 
Substituted type : R<>
Inferred record pattern type (upward projection(R<>,)) = R<?>
```

This is emitting information at specific points in the code where something major seems to have been accomplished. 

I don't see a way to emit the bound set at the end of each of the steps in 18.5.5 as I can't readily demarcate the boundaries of such steps. This calls for deeper understanding ...

stephan-herrmann:
> > Let's assume for a moment that `R<?>` is the correct result from 6
> 
> I can confirm that is what javac computes.

That would mean the ecj and javac actually agree for the record pattern inference.

In that case we should search for a difference regarding the next statement:
```java
var t = new R<>(v);
```

Indeed when I reduce the snippet to:
```java
			    static <T> boolean foo2(Object y) {
			       if (y instanceof R(var v)) {
			            R<Object> r = new R<>(v);
			            return true;
			        }
			        return false;
			    }
```
javac and ecj accept it alike.

Re-adding the `var` local re-introduces the difference:
```java
			    static <T> boolean foo3(Object y) {
			       if (y instanceof R(var v)) {
			            var t = new R<>(v);
			            R<Object> r = t; 
			            return true;
			        }
			        return false;
			    }
```

_the expression `y.t()` as the instanceof LHS as shown in the original snippet has no impact_

OK, the record pattern can be omitted, too:
```java
			    static <T> boolean foo4(R<?> y) {
					var t = new R<>(y.t());
					R<Object> r = t; 
					return true;
			    }
```

These seem to be the essential ingredients:
* a `var` typed local
   *  initialized from a diamond allocation
      * having a captured argument to derive its type argument from

To remove all traces of records:
```java
class I<T> {
	T t() { return null; }
}
class J<T> {
	J(T t) {}
}
class Test {
	void test(I<?> i) {
		var t = new J<>(i.t());
		J<Object> r = t; 
	}
}
```

javac: accept.

ecj:
```
----------
1. ERROR in /tmp/Test.java (at line 11)
        J<Object> r = t; 
                      ^
Type mismatch: cannot convert from J<capture#2-of ?> to J<Object>
----------
1 problem (1 error)
```

stephan-herrmann:
For `var` locals we consult JLS 14.4.1. :
> If the LocalVariableType is var, then let T be the type of the initializer expression when treated as if it did not appear in an assignment context, and were thus a standalone expression ([§15.2](https://docs.oracle.com/javase/specs/jls/se26/html/jls-15.html#jls-15.2)). The type of the local variable is the upward projection of T with respect to all synthetic type variables mentioned by T ([§4.10.5](https://docs.oracle.com/javase/specs/jls/se26/html/jls-4.html#jls-4.10.5)).

So we need to resolve the diamond `new J<>(i.t());` as a standalone expression.

In ecj this has type `J<capture#1-of ?>`. To mimic javac's behavior we might convert that type to raw, which would let us accept the snippet. 

Converting to raw could, e.g., happen, if unchecked conversion were needed to show applicability of the constructor. But I don't see any unchecked conversion.

I thinks this is enough material to ask on compiler-dev@openjdk

stephan-herrmann:
> I thinks this is enough material to ask on compiler-dev@openjdk

Let's see if this bears fruit:
* https://mail.openjdk.org/archives/list/compiler-dev@openjdk.org/thread/GUUK6OCUCWG6P6GDFMM4ZSA6UYTYOPCQ/

stephan-herrmann:
> > I thinks this is enough material to ask on compiler-dev@openjdk
> 
> Let's see if this bears fruit:
> 
> * https://mail.openjdk.org/archives/list/compiler-dev@openjdk.org/thread/GUUK6OCUCWG6P6GDFMM4ZSA6UYTYOPCQ/

I'm tracking this via https://github.com/eclipse-jdt/eclipse.jdt.core/wiki/ECJ-Relevant-JDK-Issues#requests-for-clarification

Issue #3303, opened 2024-11-13
[Modernization][4.35+] What would it take to make TerminalTokens an enumeration ? 


srikanth-sankaran:
It would make debugging the parser and scanner SO MUCH easier. What are the complications if any ??

iloveeclipse:
I have no IDE at hand, but I believe there was some code that copied or translated these constants in some "upper" JDT layer constants, see https://github.com/eclipse-jdt/eclipse.jdt.core/pull/10.
If that also wkrks with enums, why not.

jespersm:
I'm willing to give this one a try,

srikanth-sankaran:
That would be very welcome, thanks @jespersm

jespersm:
@srikanth-sankaran and @iloveeclipse  : I've pushed an initial draft for your review. If you approve of the approach, I will improve the PR w.r.t. he warnings and test failures found by the build.

PR #3819, opened 2025-03-18
Fix null-status of token enums (#3303)

jespersm:
## What it does

With the change from using an int to using an enum to represent a token, the default value of a
token-holding variable has changed from `0` (meaning TokenNameNotAToken, and works like that in a switch) to `null` which is bad in a switch statement.

This was largely ignored in the recent patch, and not found in testing.

This patch rectifies this omission. I've checked every field initialization and local variable holding a `TerminalToken`, and initialized the ones where they appeared to be missing. I also put in 

And by the way: Is there a good reason not to use JDT's null-checking for ECJ itself?

## How to test
Since the existing `null`-handling didn't trigger any errors in the suite run by CI, I don't expect it to be easy to to trigger an error scenario.

I didn't do fuzz-testing, but did run this change through ECJ's null-analysis.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
@@ -167,7 +167,7 @@ public class Parser implements ParserBasicInformation, ConflictedParser, Operato
 	private static final short TYPE_CLASS = 1;
 
 	public Scanner scanner;
-	public TerminalToken currentToken;
+	public TerminalToken currentToken = TokenNameNotAToken;
 
 	static {
 		try{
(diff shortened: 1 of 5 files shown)


srikanth-sankaran (changes requested):
Looks good!

iloveeclipse:
@jespersm : could you please rebase instead of merge? Merge commits are not allowed.

jespersm:
> @jespersm : could you please rebase instead of merge? Merge commits are not allowed.

My bad, I tried the "update"-button in the GitHub app, expecting the same choices as the web interface.
So I've learned not to do that now. Will rebase and force push shortly!

jespersm:
@srikanth-sankaran : The GitHub UI claimed you requested changes, but I can't see which.

I've rebased onto master and am waiting for CI checks.

srikanth-sankaran:
Yes, the "requesting changes: appears to be a glitch.

PR #3836, opened 2025-03-20
Do not suggest `import` keyword before package declaration

### LINKED ISSUE #3799: Completion result in `CompletionTests.testCompletionKeywordImport7` doesn't make sense
Here is the code being completed (completion opened at the `|`)

```java
#
imp|

package bbb;
public class CompletionKeywordImport7 {

}
```

Given import declarations cannot appear before the package declaration, I don't think that the `import` keyword should be suggested. However the test asserts that `import` should be suggested.

datho7561:
## What it does
eg. `import` should not be suggested as a completion at the `|`:

```java
imp|

package dev.datho7561.utils;

public class HelloWorld {

}
```

Fixes #3799

## How to test
Try completion with the code snippet above.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests.java
@@ -9177,7 +9177,7 @@ public void testCompletionKeywordImport7() throws JavaModelException {
 		cu.codeComplete(cursorLocation, requestor);
 
 		assertEquals(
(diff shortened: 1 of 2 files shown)


datho7561:
It seems CI hasn't run? I wonder if this is because I marked this as work in progress. I will probably need to rewrite the results for some of the other completion tests, I just don't know which yet. I'll run the tests locally to find those, I guess.

rgrunber:
Was just looking at this and wondering. Is there any value in returning the (non-applicable at that location) completion results, having the diagnostics emit errors, and then user realizes they need probably need to move the imports as opposed to not returning anything and having the user wonder "why does't completion work" ?

rgrunber (approved):
LGTM. It's not excluded as part of parsing (as is the case for placing `imp|` after a type declaration or within one)  but this seems pretty reasonable.

PR #3852, opened 2025-03-25
Add support for MaxJ syntax

mbazley2:
Derived from https://github.com/maxeler/eclipse.jdt.core/tree/MAXJ_4_4_maintenance


mbazley2:
Goddammit GitHub.

Issue #3891, opened 2025-04-03
Misplaced constructor call causes NPE


stephan-herrmann:
```java
public class Test {
	{
		super();
	}
}
```

Compile with ecj from 4.35:
```
----------
1. ERROR in /tmp/Test.java (at line 3)
        super();
        ^^^^^^^^
Constructor call must be the first statement in a constructor
----------
1 problem (1 error)
```

Regression in HEAD:
```
----------
1. ERROR in /tmp/Test.java (at line 0)
        public class Test {
        ^
Internal compiler error: java.lang.RuntimeException: Internal Error compiling /tmp/Test.java at org.eclipse.jdt.internal.compiler.ProcessTaskManager.processing(ProcessTaskManager.java:139)
----------
Internal Error compiling /tmp/Test.java
```

PR #3728, opened 2025-02-12
[Records] Code generation for compact constructor is needlessly roundabout

srikanth-sankaran:
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3727

PR #2685, opened 2024-07-07
[23] JEP 482: Flexible Constructor Bodies (Second Preview)

stephan-herrmann:
Preparation
+ renamed the feature from "Statements before Super" to "Flexible Constructor Bodies"
+ move the main flag to let ClassScope manage early construction contexts

Implement updates for JEP 482:
+ change wording towards "early construction context"
+ determine early-ctor-ctx based on target class
  fine tuning:
  - where should enclosing types be considered?
  - search for enclosings first but also consider their supers in a second round
+ reduce # mechanisms for detecting early construction context
+ allow assignment to fields of uninitialized instance
+ differentiate new errors: 
   - FieldReadInEarlyConstructionContext
   - ThisInEarlyConstructionContext
   - AllocationInEarlyConstructionContext
   - MessageSendInEarlyConstructionContext
   - DuplicateExplicitConstructorCall
   - ConstructorCallNotAllowedHere
   - SuperFieldAssignInEarlyConstructionContext
   - AssignFieldWithInitializerInEarlyConstructionContext
+ fine tune generation of synthetic arguments and fields for enclosing instances
   - such synthetics allow to skip over unavailable enclosing instances towards outer instances that are available

Documentation of key strategies as javadoc / code comments:
+ enablement: `JavaFeature.FLEXIBLE_CONSTRUCTOR_BODIES`
  - this is relevant because effects of this feature spread wide into the code, explains which invariants still help to keep the number of checks low
+ detection / analysis: `ClassScope.insideEarlyConstructionContext`
   - describes the combination of structural and temporary contexts
+ explicit vs implicit constructor calls: `ConstructorDeclaration.resolveStatements()`
   - code comment describing updated contract of `ConstructorDeclaration.constructorCall`
+ code gen: `TypeDeclaration.manageEnclosingInstanceAccessIfNecessary(BlockScope, FlowInfo)`
   - code comments describing the strategy to generate synth args & fields for various outer classes, which orchestrates code generation

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2472

PR #3892, opened 2025-04-03
Misplaced constructor call causes NPE

stephan-herrmann:
Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3891

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ExplicitConstructorCall.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ExplicitConstructorCall.java
@@ -329,7 +329,9 @@ public class ExplicitConstructorCall extends Statement implements Invocation {
 				}
 			}
 			if (hasError) {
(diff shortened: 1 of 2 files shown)


stephan-herrmann:
Fixes regression caused by #3728 -  see https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3728/files#diff-4877cf7583679781e310837d3fdedfaf8886f44c2832b511acc74df39ab962eb
@srikanth-sankaran  fyi :)

stephan-herrmann:
Actually a similar NPE could be triggered since #2685 if `--enable-preview` is enabled.

srikanth-sankaran:
> Fixes regression caused by #3728 - see https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3728/files#diff-4877cf7583679781e310837d3fdedfaf8886f44c2832b511acc74df39ab962eb @srikanth-sankaran fyi :)

Thanks for catching and fixing it @stephan-herrmann - Obviously `methodDeclaration instanceof CompactConstructorDeclaration` isn't replaceable by `methodDeclaration.isCompactConstructor()` without a null check - I went over all the `instanceof CompactConstructorDeclaration` checks in that PR and other places where this was changed to a call to `isCompactConstructor()` look safe.

Issue #3918, opened 2025-04-10
Organize imports in post save actions causes unsync timestamp of IResource and CompilationUnitElementInfo


mehmet-karaman:
Hello,

I've stumbled over a problem in the BasicSearchEngine. If the user has edited a Java File and saves the Java File with unused imports the BasicSearchEngine with using a HierarchyScope, it won't find any searched method in that Java type.

This happens because the organize imports is executed by the post save listener. In this case the post save listener is modifying the IResource and in that time the IResources modification timestamp and the ICompilationUnits ElementInfo timestamp gets unsynced. 

The fix is provided in two repositories: 
- eclipse.jdt.core => the new interface method which can be called by the UI Code. The new method has a empty default implementation to not force other clients to implement it. This will be used to synchronize the timestamps between the IResource and ICompilationUnits ElementInfo inside of the post save listener execution after the listener was executed.
Link to the PR: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3919
 
- eclipse.jdt.ui => Calls the provided method.. Added two tests which ensure that the timestamps are in sync and the other one ensures that the BasicSearchEngine can find results from the search.
Link to the PR: https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2155

Best regards
Mehmet

mehmet-karaman:
After several hours of debugging and trying to understand why this happens in that curtain case, i stumbled over the working copies array that was sometimes contained 1 object and sometimes nothing.. Now I understand that there are two different behaviors in saving files. The difference comes that in that curtain cases the timestamp of the ICompilationUnit and the IResource wasn't the same. 

Things I've done to reproduce:

I've opened a class modified it, added necessary imports, removed the unnecessary imports and it worked as expected. The timestamp of the resource and the compilation unit was the same.. and the IComplilationUnit wasn't in the copies array.

The second time I've added the necessary imports, didn't removed the unnecessary imports and saved. Inbetween that the automated imports organizer changed the timestamp in the resource but didn't updated the time in the compilation unit. 

I am going to understand how it has to be done..

srikanth-sankaran:
@jarthana - is this simply a case of a missing reconcile operation ? Is the new API achieving the same things a reconcile operation would have achieved ? (I could be way off here - if so apologies)

PR #3919, opened 2025-04-10
Provide new ICompilationUnit.updateTimeStamp() API

mehmet-karaman:
Provides new `ICompilationUnit.updateTimeStamp() API` required by JDT UI to fix the problem with Java search affected by improper `CompilationUnitDocumentProvider.commitWorkingCopy()` design.

Basically, the JDT Java search has a special case for working copies but the working copies weren't properly updated by UI during "Save" actions...

Fixes the problem described here: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3918

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/core/ICompilationUnit.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/core/ICompilationUnit.java
@@ -17,6 +17,7 @@ package org.eclipse.jdt.core;
 
 import java.util.Collections;
 import java.util.Map;
+import org.eclipse.core.resources.IResource;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.jdt.core.dom.AST;
 import org.eclipse.jdt.core.dom.ASTParser;
@@ -845,4 +846,19 @@ CompilationUnit reconcile(int astLevel, int reconcileFlags, WorkingCopyOwner own
  */
 @Override
 void restore() throws JavaModelException;
+
+/**
+ * Can be used to synchronize the internally used modification stamp to its corresponding {@link IResource}s
+ * modification stamp. <br>
+ * Default implementation does nothing.
+ *
+ * @throws JavaModelException
+ *                                can be thrown if the last modification stamp of the corresponding {@link IResource}
+ *                                object isn't set.
+ * @see IResource#getModificationStamp()
+ * @since 3.42
+ */
+default void updateTimeStamp() throws JavaModelException {
+	// does nothing by default
+}
 }
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/CompilationUnit.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/CompilationUnit.java
@@ -1513,6 +1513,11 @@ protected void updateTimeStamp(CompilationUnit original) throws JavaModelExcepti
 }
 
 @Override
+public void updateTimeStamp() throws JavaModelException {
+	updateTimeStamp(this);
+}
+
+@Override
 protected IStatus validateExistence(IResource underlyingResource) {
 	// check if this compilation unit can be opened
 	if (!isWorkingCopy()) { // no check is done on root kind or exclusion pattern for working copies


org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/BasicSearchEngine.java
```diff
@@ -1950,7 +1950,11 @@ public boolean acceptIndexMatch(String documentPath, SearchPattern indexRecord,
 				for (int i = 0; i < copiesLength; i++) {
 					SubMonitor iterationMonitor = subMonitor.split(i);
 					final ICompilationUnit workingCopy = copies[i];
-					if (scope instanceof HierarchyScope) {
+					if (scope instanceof HierarchyScope && !(workingCopy instanceof IType && workingCopy instanceof IMember)) {
+						String javaTypeName = workingCopy.getElementName().replace(".java", "");
```
iloveeclipse, line 1954: As discussed, this is a hack and we need to understand where the original search issue is coming from before we will start patching.

iloveeclipse (changes requested):

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/BasicSearchEngine.java
mehmet-karaman: I will try to isolate this problem with plain java editors, so we will have an easier example to explain the problem. Then we might be able to understand the case and it would be easier to create a testcase after the bugfix.

mehmet-karaman: I can't reproduce this bug with plain Java Editors, because the JDT Completion proposals for types don't use the BasicSearchEngine like the Xtext completions.. Could be that the Xtext Completions should use a different way? Maybe its better to go that way? I am going to check how JDT does propose the Types, maybe I could use the same for Xtext.

mehmet-karaman: Tried that too, but it never entered the acceptIndexMatch method. In the Call hierarchy I've seen that the MissingTypeGuesser is also calling somehow (not directly) the method.. I will do some experiments with the search and code completion, there must be a way to reach this..

image

mehmet-karaman: Accidentally caused that call.. The hover went into this method.. Discovered that only the current active CompilationUnit is contained in that ICompilationUnit arrays ("copies") ... If i recall correct, in the Xtext Code completion I've got all open editors in that list.. Will recheck that..

mehmet-karaman:
@iloveeclipse The first part of the fix is done. The second part of the fix will be done in the eclipse.jdt.ui repository and will contain also the test. Need to work on the test implementation..

mehmet-karaman:
After this PR is merged, the second PR can be build.

https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2155

org.eclipse.jdt.core/model/org/eclipse/jdt/core/ICompilationUnit.java
```diff
@@ -845,4 +847,12 @@ default Map<String, String> getOptions(boolean inheritJavaCoreOptions) {
  */
 @Override
 void restore() throws JavaModelException;
+
+/**
+ * Can be used to synchronize the {@link CompilationUnitElementInfo} timestamp to its corressponding {@link IResource}s timestamp.
+ * @throws JavaModelException if the updateTimeStamp(CompilationUnit)
```
stephan-herrmann, line 853: Hi, Mehmet :)

I just happened to see this incomplete javadoc.

mehmet-karaman: Hi Stephan :),

thanks.. I've adjusted the javadoc.

mehmet-karaman: Seems that @since tag is necessary for new interface methods .. I hope 4.36 M1 was correct.. or should i use 4.37?

iloveeclipse: > I hope 4.36 M1 was correct.. or should i use 4.37?

Neither one. You should use version of the bundle that introduced the API, not SDK.

org.eclipse.jdt.core/model/org/eclipse/jdt/core/ICompilationUnit.java
```diff
@@ -845,4 +847,13 @@ default Map<String, String> getOptions(boolean inheritJavaCoreOptions) {
...
 @Override
 void restore() throws JavaModelException;
+
+/**
+ * Can be used to synchronize the {@link CompilationUnitElementInfo} timestamp to its corressponding {@link IResource}s timestamp.
+ * @throws JavaModelException can be thrown if the last modification timestamp of the coressponding {@link IResource} object isn't set.
+ * @since 3.42
+ */
+default void updateTimeStamp() throws JavaModelException {
+
```
iloveeclipse: > Code check says "Empty block should be documented"

org.eclipse.jdt.core/model/org/eclipse/jdt/core/ICompilationUnit.java
```diff
@@ -845,4 +847,13 @@ default Map<String, String> getOptions(boolean inheritJavaCoreOptions) {
  */
 @Override
 void restore() throws JavaModelException;
+
+/**
+ * Can be used to synchronize the {@link CompilationUnitElementInfo} timestamp to its corressponding {@link IResource}s timestamp.
```
iloveeclipse: `CompilationUnitElementInfo` is internal non API type, it shouldn't appear in API javadoc

iloveeclipse (changes requested):
Commit title should not start with "First part of". There will be no second part in this repo.

The commit message body could explain that the full fix will require JDT UI fix as well, but title should be self contained, something like "Provide new ICompilationUnit.updateTimeStamp()" API and after a new line you could explain in the commit message why this API method is needed and why it is made default etc.

I've pushed to your branch the changes needed.

iloveeclipse:
Waiting now on improved https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2155 patch.

iloveeclipse (approved):

iloveeclipse:
@mehmet-karaman : I've updated **this** PR title and description, please update https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3918 original description to reflect our latest understanding of the problem. Currently the original ticket describes the original attempt with the "dirty hack", and that is only confusing.

Issue #3929, opened 2025-04-15
NPE when project's source level is higher then supported levels


stephan-herrmann:
Seen in 4.36 M1 (Version: 2025-06 (4.36) Build id: I20250403-1800):

Many NPEs like this:
```
java.lang.NullPointerException: Cannot invoke "java.lang.Integer.intValue()" because the return value of "java.util.Map.get(Object)" is null
	at org.eclipse.jdt.core.dom.AST.<init>(AST.java:1246)
	at org.eclipse.jdt.core.dom.AST.newAST(AST.java:767)
	at org.eclipse.jdt.internal.core.CompilationUnit.getCustomOptions(CompilationUnit.java:1581)
	at org.eclipse.jdt.core.ICompilationUnit.getOptions(ICompilationUnit.java:606)
	at org.eclipse.jdt.internal.core.CompilationUnit.buildStructure(CompilationUnit.java:135)
	at org.eclipse.jdt.internal.core.Openable.generateInfos(Openable.java:250)
	at org.eclipse.jdt.internal.core.JavaElement.openWhenClosed(JavaElement.java:569)
	at org.eclipse.jdt.internal.core.JavaElement.getElementInfo(JavaElement.java:292)
	at org.eclipse.jdt.internal.core.JavaElement.getElementInfo(JavaElement.java:278)
	at org.eclipse.jdt.internal.core.PackageFragmentRoot.getSourceModuleDescription(PackageFragmentRoot.java:922)
	at org.eclipse.jdt.internal.core.PackageFragmentRoot.getModuleDescription(PackageFragmentRoot.java:905)
	at org.eclipse.jdt.internal.core.JavaProject.buildStructure(JavaProject.java:498)
	at org.eclipse.jdt.internal.core.Openable.generateInfos(Openable.java:250)
	at org.eclipse.jdt.internal.core.JavaElement.openWhenClosed(JavaElement.java:569)
	at org.eclipse.jdt.internal.core.JavaElement.getElementInfo(JavaElement.java:292)
	at org.eclipse.jdt.internal.core.JavaElement.getElementInfo(JavaElement.java:278)
	at org.eclipse.jdt.internal.core.JavaProject.getModuleDescription(JavaProject.java:3746)
	at org.eclipse.jdt.internal.core.builder.NameEnvironment.computeClasspathLocations(NameEnvironment.java:155)
[...]

PR #3938, opened 2025-04-17
Allow creating AST even with unknown source version

mickaelistria:
Will default to latest known version in that case.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3929

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/CompilationUnit.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/CompilationUnit.java
@@ -46,6 +46,7 @@ import org.eclipse.jdt.internal.codeassist.DOMCodeSelector;
 import org.eclipse.jdt.internal.compiler.IProblemFactory;
 import org.eclipse.jdt.internal.compiler.SourceElementParser;
 import org.eclipse.jdt.internal.compiler.ast.CompilationUnitDeclaration;
+import org.eclipse.jdt.internal.compiler.classfmt.ClassFileConstants;
 import org.eclipse.jdt.internal.compiler.env.IElementInfo;
 import org.eclipse.jdt.internal.compiler.impl.CompilerOptions;
 import org.eclipse.jdt.internal.compiler.lookup.TypeConstants;
@@ -1583,7 +1584,7 @@ public Map<String, String> getCustomOptions() {
 			IJavaProject parentProject = getJavaProject();
 			Map<String, String> parentOptions = parentProject == null ? JavaCore.getOptions() : parentProject.getOptions(true);
 			if (JavaCore.ENABLED.equals(parentOptions.get(JavaCore.COMPILER_PB_ENABLE_PREVIEW_FEATURES)) &&
-				AST.newAST(parentOptions).apiLevel() < AST.getJLSLatest()) {
+				CompilerOptions.versionToJdkLevel(parentOptions.getOrDefault(JavaCore.COMPILER_SOURCE, JavaCore.latestSupportedJavaVersion())) < ClassFileConstants.getLatestJDKLevel()) {
 				// Disable preview features for older Java releases as it causes the compiler to fail later
 				if (customOptions != null) {
 					customOptions.put(JavaCore.COMPILER_PB_ENABLE_PREVIEW_FEATURES, JavaCore.DISABLED);


org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/AST.java
```diff
@@ -1243,7 +1243,7 @@ private AST(int level, boolean previewEnabled) {
 	 * @see JavaCore#getDefaultOptions()
 	 */
 	public AST(Map options) {
-		this(apiLevelMap.get(options.get(JavaCore.COMPILER_SOURCE)),
+		this(apiLevelMap.getOrDefault(JavaCore.COMPILER_SOURCE, JLS_INTERNAL_Latest),
```
laeubi, line 1246: Should it not better default to the _lowest_ known value? I can only think about that values become "illegal" after a while when older java versions are dropped, and while development time it would be better to get possibly an error for an unsupported java feature when specify a too high value.

mickaelistria: > Should it not better default to the lowest known value? 

If using the lowest version, then every more modern structure will cause exceptions later and result in a non usable AST/editor. And as the initial report is about using a too new version, going to latest is the least breaking default.

laeubi: > If using the lowest version, then every more modern structure will cause exceptions later and result in a non usable AST/editor

I think setting a "too high" version manually (or by code) is an error and one not really gets "unusable" AST/editor but just complaining about unusupported things and one then might check this.

>  And as the initial report is about using a too new version, going to latest is the least breaking default.

The report mainly complains about a NPE so I'm not sure what was the behavior before, but if it set e.g. java version =  99 and get no errors my expectation would be that this value is used. but maybe @stephan-herrmann can tell what is the actual expectation from compiler point of view.

mickaelistria: > I think setting a "too high" version manually (or by code) is an error

Not necessarily. Opening a recent project with source version set in metadata in an older Eclipse version leads to this state and is not so much of an error.

laeubi: > > I think setting a "too high" version manually (or by code) is an error
> 
> Not necessarily. Opening a recent project with source version set in metadata in an older Eclipse version leads to this state and is not so much of an error.

But if my "recent project" requires Java 24 but my "older Eclipse" only supports Java 21, then it does not help much if if get Java 21 as the project settings that complains about Java 24 features.

So I still think this is an error open a project requiring a higher java version in an Eclipse not supporting that.

mickaelistria: To be honest, I don't really care about the chosen solution in the end as it's not a problem that is in my current scope of interest, so please submit an alternative PR withou your proposal and if it's better, let's just merge it.

stephan-herrmann: @mickaelistria  @laeubi  I can see where the disagreement might come from: creating an `AST` for an unsupported version is "unsupported" - so there may not be a perfect solution.

For the original problem #3929  it may still be better to avoid (needlessly) using this API for #2700 as mentioned in https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3929#issuecomment-2805147096

stephan-herrmann: BTW:
> it's not a problem that is in my current scope of interest

it may well enter your scope of interest if #3929 is fixed by reverting #2700 which triggered the NPE :)

mickaelistria:
@stephan-herrmann Does this new version better match your expectations?

stephan-herrmann:
> @stephan-herrmann Does this new version better match your expectations?

No.

I didn't mean the `CompilationUnit.getCustomOptions()` should be crippled if `!DOM_BASED_OPERATIONS`.

Instead I meant that this condition introduced in #2700 
```java
if (JavaCore.ENABLED.equals(parentOptions.get(JavaCore.COMPILER_PB_ENABLE_PREVIEW_FEATURES)) &&
		AST.newAST(parentOptions).apiLevel() < AST.getJLSLatest()) {
```
should be written without any DOM API, but using fundamental implementations like `CompilerOptions` and `JavaCore`. See, e.g., 
* `JavaCore.latestSupportedJavaVersion()`
* `CompilerOptions.versionToJdkLevel()`

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/CompilationUnit.java
```diff
@@ -1577,13 +1577,13 @@ public void setOptions(Map<String, String> newOptions) {
...
 public Map<String, String> getCustomOptions() {
-	if (this.owner != null) {
+	if (this.owner != null && DOM_BASED_OPERATIONS) {
 		try {
 			Map<String, String> customOptions = this.getCompilationUnitElementInfo().getCustomOptions();
 			IJavaProject parentProject = getJavaProject();
 			Map<String, String> parentOptions = parentProject == null ? JavaCore.getOptions() : parentProject.getOptions(true);
 			if (JavaCore.ENABLED.equals(parentOptions.get(JavaCore.COMPILER_PB_ENABLE_PREVIEW_FEATURES)) &&
-				AST.newAST(parentOptions).apiLevel() < AST.getJLSLatest()) {
+				CompilerOptions.versionToJdkLevel(parentOptions.getOrDefault(JavaCore.COMPILER_SOURCE, JavaCore.latestSupportedJavaVersion())) < CompilerOptions.versionToJdkLevel(JavaCore.latestSupportedJavaVersion())) {
```
stephan-herrmann, line 1586: In case you want to keep it a bit shorter have a look at `ClassFileConstants.getLatestJDKLevel()`

With fallbacks explicit on both sides, we now see that you allow enabling preview features also for unspecified and unsupported Java versions (where the LHS will fall back to _latest_). Is that by intention? What should be the outcome?

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/CompilationUnit.java
```diff
@@ -1577,13 +1577,13 @@ public void setOptions(Map<String, String> newOptions) {
 
 @Override
 public Map<String, String> getCustomOptions() {
-	if (this.owner != null) {
+	if (this.owner != null && DOM_BASED_OPERATIONS) {
```
stephan-herrmann, line 1580: this addition is wrong

stephan-herrmann (changes requested):

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/CompilationUnit.java
mickaelistria: > Is that by intention? What should be the outcome?

Yes, but it's not a strong requirement. Basically, the goal is that for an unknown version, we failback in most cases where AST is used to the "most capable" configuration set to get an AST as helpful as possible. My interpretation is that, usually, the highest version + preview feature is the most capable set.

stephan-herrmann:
Code changes now look good to me. Once we have a commit message that matches the change and a successful build we can release this.

stephan-herrmann:
> [Avoid crash when creating AST even with unknown source version](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3938/commits/741c1ada09c52270bbfcd35eff42d9ca52563d70)

There is no AST in the picture. The NPE in #3929 happened when computing a classpath / opening an IJavaProject on behave of the builder. The fix is in CompilationUnit from Java model.

Should I explain the difference between AST and java model?

stephan-herrmann (approved):
OK let's just wait for build machines to recover.

Issue #4006, opened 2025-05-15
NullPointerException in ProblemReporter.moduleNotFound


ptziegler:
I'm currently trying out the M2 for the 2025-06 release and in our application, I'm now seeing the following exception:

```
de.dsa.rcp.plugins.ts.generate.TSCompileException: Cannot read the array length because "value" is null
	at de.dsa.rcp.plugins.ts.runtimeapi.AbstractRuntimeModelGenerator.compilationResults(AbstractRuntimeModelGenerator.java:167) ~[de.dsa.rcp.plugins.editor.debug/:?]
	at de.dsa.rcp.plugins.ts.runtimeapi.AbstractRuntimeModelGenerator.compilationResults(AbstractRuntimeModelGenerator.java:116) ~[de.dsa.rcp.plugins.editor.debug/:?]
	at de.dsa.rcp.plugins.ts.runtimeapi.AbstractRuntimeModelGenerator.compilationResults(AbstractRuntimeModelGenerator.java:474) ~[de.dsa.rcp.plugins.editor.debug/:?]
	at de.dsa.rcp.plugins.ts.runtimeapi.JaninoGenerator.findCompileProblems(JaninoGenerator.java:131) ~[de.dsa.rcp.plugins.editor.debug/:?]
	at de.dsa.rcp.plugins.ts.runtimeapi.JaninoGenerator.analyze(JaninoGenerator.java:125) ~[de.dsa.rcp.plugins.editor.debug/:?]
	at de.dsa.rcp.plugins.ts.runtimeapi.JaninoGenerator.byteCodeForAuthoringCompiler(JaninoGenerator.java:59) ~[de.dsa.rcp.plugins.editor.debug/:?]
...
Caused by: java.lang.NullPointerException: Cannot read the array length because "value" is null
	at java.base/java.lang.String.<init>(String.java:278) ~[?:?]
	at java.base/java.lang.String.valueOf(String.java:4479) ~[?:?]
	at org.eclipse.jdt.internal.compiler.problem.ProblemReporter.moduleNotFound(ProblemReporter.java:11403) ~[org.eclipse.jdt.core.compiler.batch_3.42.0.v20250507-1036.jar:?]
	at org.eclipse.jdt.internal.compiler.ast.CompilationUnitDeclaration.module(CompilationUnitDeclaration.java:842) ~[org.eclipse.jdt.core.compiler.batch_3.42.0.v20250507-1036.jar:?]
[...]

PR #4007, opened 2025-05-15
Add fallback to unnamed module if INameEnvironment is used

ptziegler:
## What it does

If an ICompilationUnit is compiled with a plain INameEnvironment, a NullPointerException is thrown when the source code is also in an unnamed module.

In such a case, the ICompilationUnit should return the unnamed module, rather than null, thus restoring the previous behavior that was changed as part of 681dc0aff701f3bfe3109d43b838b0894c6d2cdd.

Closes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4006

## How to test

Execute the following snippet:
```java
package test;

import java.util.HashMap;
import java.util.Locale;
import java.util.Map;

import org.eclipse.jdt.internal.compiler.Compiler;
import org.eclipse.jdt.internal.compiler.DefaultErrorHandlingPolicies;
import org.eclipse.jdt.internal.compiler.ICompilerRequestor;
import org.eclipse.jdt.internal.compiler.IErrorHandlingPolicy;
import org.eclipse.jdt.internal.compiler.IProblemFactory;
import org.eclipse.jdt.internal.compiler.env.ICompilationUnit;
import org.eclipse.jdt.internal.compiler.env.INameEnvironment;
import org.eclipse.jdt.internal.compiler.env.NameEnvironmentAnswer;
import org.eclipse.jdt.internal.compiler.impl.CompilerOptions;
import org.eclipse.jdt.internal.compiler.problem.DefaultProblemFactory;

public class Test {
	private static String content = """
			public class Foo {
			}
			""";
	
		ICompilationUnit cu = new CompilationUnit(content, "Foo.java");
		final INameEnvironment nameEnvironment = new NameEnvironmentStub();
		final IErrorHandlingPolicy policy = DefaultErrorHandlingPolicies.proceedWithAllProblems();
		Map<String, String> defaultSettings = new HashMap<>();
		defaultSettings.put(CompilerOptions.OPTION_Source, CompilerOptions.VERSION_17);
		defaultSettings.put(CompilerOptions.OPTION_Compliance, CompilerOptions.VERSION_17);
		final CompilerOptions compilerOptions = new CompilerOptions(defaultSettings);
		final ICompilerRequestor compilerRequestor = e -> {};
		final IProblemFactory problemFactory = new DefaultProblemFactory(Locale.getDefault());
		
		Compiler c = new Compiler(nameEnvironment, policy, compilerOptions, compilerRequestor, problemFactory);
		c.compile(new ICompilationUnit[] {cu});
	}
	
	private static class CompilationUnit implements ICompilationUnit {
		private String contents;
		private String fileName;
		
		public CompilationUnit(String contents, String fileName) {
			this.contents = contents;
			this.fileName = fileName;
		}

		@Override
		public char[] getFileName() {
			return fileName.toCharArray();
		}

		@Override
		public char[] getContents() {
			return contents.toCharArray();
		}

		@Override
		public char[] getMainTypeName() {
			return fileName.substring(0, fileName.indexOf('.')).toCharArray();
		}

		@Override
		public char[][] getPackageName() {
			return null;
		}
		
	}
	
	private static class NameEnvironmentStub implements INameEnvironment {
		@Override
		public NameEnvironmentAnswer findType(char[][] compoundTypeName) {
			return null;
		}

		@Override
		public NameEnvironmentAnswer findType(char[] typeName, char[][] packageName) {
			return null;
		}

		@Override
		public boolean isPackage(char[][] parentPackageName, char[] packageName) {
			return false;
		}

		@Override
		public void cleanup() {
		}
	}
}
```

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/env/ICompilationUnit.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/env/ICompilationUnit.java
@@ -65,7 +65,7 @@ default ModuleBinding module(LookupEnvironment environment) {
 		char[] moduleName = modEnv.isOnModulePath(this) ? getModuleName() : ModuleBinding.UNNAMED;
 		return environment.getModule(moduleName);
 	}
-	return null;
+	return environment.getModule(ModuleBinding.UNNAMED);
 }
 /**
  * Returns the name of the module to which this compilation unit is associated.


ptziegler:
Any thoughts on this?

Based on https://github.com/eclipse-jdt/eclipse.jdt.core/blob/2eb0b1ddcd0fb3fff3c0b78c8cfdedce8681eb85/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/CompilationUnitDeclaration.java#L837-L847 I think that `compilationUnit.module(environment)` is only supposed to return `null` if used in a modular environment and if the module containing the compilation unit can't be found.

The current behavior makes the 2025-06 effectively unusable, which is why any feedback is greatly appreciated.

stephan-herrmann:
@ptziegler Thanks for the test class. If someone turns that into a regular JUnit test, then I'm fine to support this PR, in which case we only need the additional approval for releasing to RC1 (right now) or even RC2.

ptziegler:
> @ptziegler Thanks for the test class. If someone turns that into a regular JUnit test, then I'm fine to support this PR, in which case we only need the additional approval for releasing to RC1 (right now) or even RC2.

Thanks for the feedback. Together with your https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4006#issuecomment-2894458793 I understand this as that the `INameEnviromnent` is primarily kept for legacy reason and might be removed at any point and will most likely the moment support for Java 8 is dropped.

Or from our side: Even if this exception is "fixed" now, we effectively only delay the inevitable.

> So if avoiding the illegal implementation is not an option, at least it should be updated to be compatible with recent Java versions.

I fear that this is unavoidable. At the very least I don't think there is any way our specific use case could be with what's currently API (Just as an example: We don't have a workspace). I've tried implementing the `IModuleAwareNameEnvironment` for our product and on a surface-level this seems to work.

I'll have to run some additional tests and if there are no other problems, I think that this will be the "best" solution that won't require any adaptations in JDT.

ptziegler:
Closing, as the underlying issue is no longer relevant.

Issue #4015, opened 2025-05-19
[Records] Incorrect error "Cannot instantiate local class 'ARecord' in a static context"


srikanth-sankaran:
Brought forward from https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3831#issuecomment-2890446708

Appears to be a defect introduced by Java 24 work - need to double check

```
import java.util.function.Supplier;

public class X {

	
	public void staticMethod() {
		
		record ARecord() {
			public static ARecord of() {
				return new ARecord();
			}
		}
	}
	
	public static Supplier<Object> test() {
		class A {}
		return () -> new A();
	}

}
```

srikanth-sankaran:
Yet another problematic case put together by studying other call sites of `org.eclipse.jdt.internal.compiler.problem.ProblemReporter.allocationInStaticContext(ASTNode, LocalTypeBinding)`

```
import java.util.function.Supplier;

public class X {

	interface I {}
	
	public void staticMethod() {
		
		record ARecord() implements I {
			public static ARecord of() {
				Supplier<I> si = ARecord::new;
				return (ARecord) si.get();
			}	
		}
	}
}
```

This compiles with javac but not with ECJ

srikanth-sankaran:
I also checked whether the call site of `org.eclipse.jdt.internal.compiler.problem.ProblemReporter.allocationInStaticContext(ASTNode, LocalTypeBinding)` in `org.eclipse.jdt.internal.compiler.ast.ExplicitConstructorCall.resolve(BlockScope)` may suffer from the same problem. 

Answer is No.

Here is the reasoning: 

The reporting of `allocationInStaticContext` in `org.eclipse.jdt.internal.compiler.ast.ExplicitConstructorCall.resolve(BlockScope)`  is guarded by the predicate `this.accessMode <= ExplicitConstructorCall.Super`  so any chaining via `this(...)` does not count. All non-canonical constructors must start with chaining via `this(...)` and are not impacted. Canonical constructors may not contains an explicit super call, but the implicit super case is not problematic since `allocationInStaticContext` reporting happens only for local types and just above that place we have this:

```
if (this.accessMode != ExplicitConstructorCall.This) {
				receiverType = receiverType.superclass();
```

which will make `receiverType ` to be java.lang.Record which is not a local class. So we won't issue `allocationInStaticContext` diagnostic ever for record class's constructors chaining explicit or implicit

srikanth-sankaran:
[...]

PR #4020, opened 2025-05-20
[Records] Incorrect error "Cannot instantiate local class 'ARecord' in a static context"

srikanth-sankaran:
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4015

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ExplicitConstructorCall.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ExplicitConstructorCall.java
@@ -377,7 +377,7 @@ public class ExplicitConstructorCall extends Statement implements Invocation {
 				}
 				if (!receiverType.isEnum() &&
 						this.accessMode <= ExplicitConstructorCall.Super &&
-						receiverType instanceof LocalTypeBinding local) {
+						receiverType instanceof LocalTypeBinding local) { // local cannot be a record class
 					MethodScope allocationStaticEnclosing = scope.parent.nearestEnclosingStaticScope(); // Constructor scope already has static, start from parent scope
 					MethodScope typesEnclosingStaticScope = local.scope.nearestEnclosingStaticScope();
 					if (allocationStaticEnclosing != null && typesEnclosingStaticScope != null && allocationStaticEnclosing != typesEnclosingStaticScope)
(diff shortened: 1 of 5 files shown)


srikanth-sankaran:
JLS 8.1.3 Inner Classes and Enclosing Instances

The following nested classes are implicitly static, so are not inner classes:
• a member enum class (§8.9)
• a local enum class (§14.3)
• a member record class (§8.10)
• a local record class (§14.3)
• a member class of an interface (§9.5)

...

An instance i of a direct inner class C of a class or interface O is associated with an
instance of O, known as the immediately enclosing instance of i. The immediately
enclosing instance of an object, if any, is determined when the object is created
....

Ergo, a record class local or member is not an inner class and is not associated with an enclosing instance - (which if it were allowed, would silently add state to the record which would be design violation), so it doesn't make sense to complain of them being instantiated in static contexts where enclosing instances are unavailable

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ExplicitConstructorCall.java
```diff
@@ -377,7 +377,7 @@ public void resolve(BlockScope scope) {
 				}
 				if (!receiverType.isEnum() &&
 						this.accessMode <= ExplicitConstructorCall.Super &&
-						receiverType instanceof LocalTypeBinding local) {
+						receiverType instanceof LocalTypeBinding local) { // local cannot be a record class
```
srikanth-sankaran, line 380: See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4015#issuecomment-2893003661 for a reasoning

stephan-herrmann:
I'd prefer if we implement JLS 8.1.3 as closely as possible, i.e., mark those classes as static, so `AllocationExpression.resolveType()` only needs to check static, not a list of "special cases".

Is it relevant? Yes, see:
```java
public class C {
	void q() {
		interface I {
			class IInner {}
			static void p() {
				new IInner();
			}
		}
	}
}
```

ecj: reject, javac accept. Same reasoning as the record case.

srikanth-sankaran:
> I'd prefer if we implement JLS 8.1.3 as closely as possible, i.e., mark those classes as static

I agree with what you say. But given we are in 4.36 end game, a point fix for a problem encountered in the field may be a good idea with a follow up ticket that is more comprehensively tackled - what do you say ??

stephan-herrmann (approved):
> > I'd prefer if we implement JLS 8.1.3 as closely as possible, i.e., mark those classes as static
> 
> I agree with what you say. But given we are in 4.36 end game, a point fix for a problem encountered in the field may be a good idea with a follow up ticket that is more comprehensively tackled - what do you say ??

No objections from me, so I'm giving +1 under that condition :)

You still need a project lead's +1, though.

srikanth-sankaran:
@jarthana - agree with integrating this for RC1 ??

jarthana:
> @jarthana - agree with integrating this for RC1 ??

Yes, Ok for RC1.

Issue #4022, opened 2025-05-20
Compile error involving wildcards


brychcy:
Hi,
I upgrade to 4.36M3 and noted a new compile error in our sources.

I narrowed it down to this commit: https://github.com/eclipse-jdt/eclipse.jdt.core/commit/d9d550f8ddeda45cf4d1b803a99afbd73abf57e4 
Seeing wildcards, I know it is not trivial to say what would be correct according to the JLS, but javac compiles it and the code looks resonable.

As workaround, "extract local variable" helps as shown, but I'd have to add a comment pointing to Eclipse and I'd rather avoid that.

```
package compilebug;

import java.util.Collection;
import java.util.Set;

public abstract class Bug {
	// real code uses some guava method, so method signature  cannot be changed
	public abstract <T2> Collection<T2> toCollection(Iterable<? extends T2> elements);
	
	public abstract <T> Set<Class<? extends T>> getSubTypesAsSet(Class<T> type);

	public <T> Collection<Class<? extends T>> getSubTypesAsCollection(Class<T> superclassOrInterface)    {
		// compiles on older eclipse and javac 21
		// fails since d9d550f8ddeda45cf4d1b803a99afbd73abf57e4
		return toCollection(getSubTypesAsSet(superclassOrInterface));
	}

	public <T> Collection<Class<? extends T>> getSubTypesAsCollectionFixed(Class<T> superclassOrInterface) {
		// type as suggested by eclipse when using "extract local variable"
		Set<Class<? extends T>> subTypesAsSet = getSubTypesAsSet(superclassOrInterface);
		
		// compiles
		return toCollection(subTypesAsSet);
	}
}
```

brychcy:
@stephan-herrmann , the commit in question is by you, maybe you can have a look.

stephan-herrmann:
Hi @brychcy  I'd like to help if possible. Unfortunately my first glance at your first examples leaves little doubt that rejecting is in line with JLS. I'll be back with more details soon.

stephan-herrmann:
> Hi [@brychcy](https://github.com/brychcy) I'd like to help if possible. Unfortunately my first glance at your first examples leaves little doubt that rejecting is in line with JLS. I'll be back with more details soon.

I'm glad I didn't say "no doubt". 😎 

[...]

PR #4023, opened 2025-05-20
Compile error involving wildcards

stephan-herrmann:
PTB.findSuperTypeOriginatingFrom() should capture only when actually performing supertype navigation.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4022

### DIFF
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/NegativeLambdaExpressionsTest.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/NegativeLambdaExpressionsTest.java
@@ -37,7 +37,7 @@ import org.eclipse.jdt.internal.compiler.impl.CompilerOptions;
 public class NegativeLambdaExpressionsTest extends AbstractRegressionTest {
 
 static {
-	TESTS_NAMES = new String[] { "testIssue3956"};
+//	TESTS_NAMES = new String[] { "testIssue3956"};
 //	TESTS_NUMBERS = new int[] { 50 };
 //	TESTS_RANGE = new int[] { 11, -1 };
 }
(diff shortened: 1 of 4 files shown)


stephan-herrmann:
@jarthana @mpalat @srikanth-sankaran  a quick heads-up, that I'd like to help @brychcy with this fix, which would require review and PL +1 for RC1. Anyone available?

stephan-herrmann:
To copy the conceptual backing of this change [from the issue](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4022#issuecomment-2895172445):
> I think I found the conceptual leak: to determine the supertype of a parameterized type we need to capture (JLS §4.10.2), but during findSuperTypeOriginatingFrom() we are currently capturing even if the generic type is the same, i.e., no navigation to a supertype is involved. That's wrong and fixing this leak fixes at least example 1.

Fixes actually both provided examples.

stephan-herrmann:
I believe test failures in ClasspathTests were unrelated.

srikanth-sankaran:
@jarthana - I'll review this, this will also require lead approval.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/NegativeLambdaExpressionsTest.java
```diff
@@ -37,7 +37,7 @@
 public class NegativeLambdaExpressionsTest extends AbstractRegressionTest {
 
 static {
-//	TESTS_NAMES = new String[] { "test404657_loop"};
+	TESTS_NAMES = new String[] { "testIssue3956"};
 //	TESTS_NUMBERS = new int[] { 50 };
```
srikanth-sankaran, line 41: Please recomment this!

srikanth-sankaran (approved):
Please fix the NLET - so it runs all tests

jarthana:
+1 for RC1

stephan-herrmann:
Thanks to both for your help.

stephan-herrmann:
> Please fix the NLET - so it runs all tests

Of course we want to stay clean, sorry for having rushed this one. 

Just as a remark I believe those test filters are not active when running the entire suite, so a forgotten filter should not necessarily block merging, IMHO. And removing a test filter actually doesn't need review & PL approval :smile:

PR #4028, opened 2025-05-21
fix for null analysis issue for conditional expressions 

### LINKED ISSUE #3990: Bug 195757 - Missing redundant null check in if condition through ternary operator
From https://bugs.eclipse.org/bugs/show_bug.cgi?id=195757

```java
public class X {

  boolean test() {
    return true;
  }

  int foo() {
    if ((test() ? null : null) == null)
      return 1;
    return 0;
  }

  // more complex

  int bar() {
    String s = null;
    if ((test() ? s : null) == null)
      return 1;
    return 0;
  }
}
```
In both case, a "redundant null check" should be reported.

For initial discussion see bugzilla.

mpalat:
## What it does
fixes #3990 

Null in Conditional Expressions are now reported with this fix. 
General algorithm:
At EqualExpression, check for conditional expression in checkNullComparison and then the logic followed is similar to the flow analysis of checkVariableComparison. Additional error code of 990 error code [IProblem.RedundantNullCheckNullValueExpression] is introduced - see below as well.

A few points:

- 671 [IProblem.RedundantNullCheckOnNonNullExpression] has been reused in the fix but could not use 672 as it meant NullExpressionReference though the error message is similar to the newly introduced 990 error code [IProblem.RedundantNullCheckNullValueExpression]
- Have not optimised the code - or done much common method refactorings - atleast for the first round of PR for the sake of clarity.

## How to test
Test cases are part of the PR

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/META-INF/MANIFEST.MF
+++ b/org.eclipse.jdt.core.compiler.batch/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Main-Class: org.eclipse.jdt.internal.compiler.batch.Main
 Bundle-ManifestVersion: 2
 Bundle-Name: Eclipse Compiler for Java(TM)
 Bundle-SymbolicName: org.eclipse.jdt.core.compiler.batch
-Bundle-Version: 3.42.50.qualifier
+Bundle-Version: 3.43.0.qualifier
 Bundle-ClassPath: .
 Bundle-Vendor: Eclipse.org
 Automatic-Module-Name: org.eclipse.jdt.core.compiler.batch
--- a/org.eclipse.jdt.core.compiler.batch/pom.xml
+++ b/org.eclipse.jdt.core.compiler.batch/pom.xml
@@ -17,7 +17,7 @@
     <version>4.37.0-SNAPSHOT</version>
   </parent>
   <artifactId>org.eclipse.jdt.core.compiler.batch</artifactId>
-  <version>3.42.50-SNAPSHOT</version>
+  <version>3.43.0-SNAPSHOT</version>
   <packaging>eclipse-plugin</packaging>
 
   <properties>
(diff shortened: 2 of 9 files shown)


mpalat:
@stephan-herrmann - can you please review when you get time? TIA. [I will wait until the tests pass]

stephan-herrmann:
At a *very brief* look from a distance I was a bit surprised about the size of the change, but maybe all this is needed, so...

Two things you might want to check:
* Does `ConditionalExpression.nullStatus` provide sufficient information right when analysing the `EqualExpression`? Would using that simplify the implementation?
* OTOH, seeing changes in `FlowContext` lets me think, yes perhaps those are necessary, but to prove the relevance of FlowContext it would be great to have test cases where this expression is embedded in some control structure like a loop.  Think of a loop where a statement after that if statement assigns a non-null value so the null check actually isn't redundant (in a subsequent iteration of the loop)! Does your implementation handle this correctly?

mpalat:
> * OTOH, seeing changes in `FlowContext` lets me think, yes perhaps those are necessary, but to prove the relevance of FlowContext it would be great to have test cases where this expression is embedded in some control structure like a loop.  Think of a loop where a statement after that if statement assigns a non-null value so the null check actually isn't redundant (in a subsequent iteration of the loop)! Does your implementation handle this correctly?

Thanks @stephan-herrmann for the input. The following gives an erroneous message - investigating the same:
```
	int bar() {
		int ret = 0; 
		String s = null;

		for (int i = 0; i < 2; i++) {
			if ((test() ? s : null) == null) // wrong null warning here
				ret = 1;
			s = "hello"; 
		}
		return ret;
	}

```

mpalat:
This doesn't flag the "definitely null" for the conditional inside a loop yet
`	
int foo3() {
		int ret = 0;
		String s = null;

		for (int i = 0; i < 2; i++) {
			if (s == null) //  null error is flagged
				ret = 2;

			if ((test() ? s : null) ==  null) // Flag null error - it's not flagged now.
				ret = 1;
		}	   
		return ret;    
	}       
`

mpalat:
> >   Think of a loop where a statement after that if statement assigns a non-null value so the null check actually isn't redundant (in a subsequent iteration of the loop)! Does your implementation handle this correctly?

 @stephan-herrmann  - This  particular case is corrected .However, the code [here](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4028#issuecomment-2982911949) doesn't flag a null error in the conditional expression. Looking through the LoopingFlowContext.checkOnDeferredNullChecks(), I see that each variable is checked for nullity and at that point the null error is flagged in the case of if statement. Investigating whether its pragmatic (it is possible for sure) to propagate this nullity property upwards to the parent as appropriate (in situations similar to that of conditional expression involving variables).

Issue #2856, opened 2024-08-23
AddJarFileToIndex.java references java.util.zip.ZipError


eirbjo:
The class ` java.util.zip.ZipError` has been unused in OpenJDK since JDK 9. Efforts are underway to mark this class as deprecated with the intention to eventually remove it in a future release.

This deprecation issue is being tracked here: https://bugs.openjdk.org/browse/JDK-8336843

To help minimize ecosystem  impact, I'm reaching out to Eclipse JDT because we identified that the `AddJarFileToIndex.java` file references `ZipError`.

If the deprecation suggestion is accepted by OpenJDK in a future release, then this reference will cause a compiler warning. In a later release, the removal of the class will cause a compilation error.

If JDT needs to compile across Java 8 and newer releases, then this code could be updated to catch the superclass `InternalError` instead, and check the class name at runtime to determine if the exception was indeed a `ZipError`. If Java 8 support is no longer needed, the ZipError catch could simply be removed from the class.

This issue is mostly a heads-up. If you want to take action now and remove/rewrite the referencing code, that would be great. Otherwise, you might start seeing compiler warnings about the deprecation in a not-so-far future..

Thanks!

PR #4029, opened 2025-05-21
Don't use java.util.zip.ZipError (not used in JDK since Java 9)

iloveeclipse:
This class is deprecated for removal since JDK 24. 
See https://bugs.openjdk.org/browse/JDK-8336843

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2856


srikanth-sankaran (approved):
LGTM

Issue #3920, opened 2025-04-10
Switch expression and type inference generates wrong/unsecure bytecode


jbyunes:
Hi,

I am currently using Version: 2025-03 (4.35.0) Build id: 20250306-0812.
Project is Java 24 JDK/JRE.
The program is:
```
import java.util.List;
import java.util.stream.Collectors;

public class Main {
	public static void main(String[] args) {
		var l = List.of("A","B");
		int i = 12;
		var t = switch(i) {
		case 1 -> l.stream().collect(Collectors.joining(" "));
		default -> "";
		};
		System.out.println(t);
	}
}

```
Compiling is ok, but running says:
- under Eclipse:
   ```
   Error: Unable to load main class test.Main in module test
   Caused by: java.lang.VerifyError
   ```
- using external Oracle's JVM java 24 2025-03-18, Java(TM) SE Runtime Environment (build 24+36-3646), Java HotSpot(TM) 64-Bit Server VM (build 24+36-3646, mixed mode, sharing):
   ```
   Error: Unable to initialize main class Main
   Caused by: java.lang.VerifyError: Bad type on operand stack
   Exception Details:
     Location:
       test/Main.main([Ljava/lang/String;)V @58: invokevirtual
     Reason:
       Type 'java/lang/Object' (current frame, stack[1]) is not assignable to 'java/lang/String'
     Current Frame:
       bci: @58
       flags: { }
       locals: { '[Ljava/lang/String;', 'java/util/List', integer, 'java/lang/Object' }
       stack: { 'java/io/PrintStream', 'java/lang/Object' }
     Bytecode:
       0000000: 1210 1212 b800 144c 100c 3d1c aa00 0000
       0000010: 0000 0027 0000 0001 0000 0001 0000 0014
       0000020: 2bb9 001a 0100 121e b800 20b9 0026 0200
       0000030: a700 0512 2c4e b200 2e2d b600 34b1     
     Stackmap Table:
       append_frame(@32,Object[#21],Integer)
       same_frame(@51)
       same_locals_1_stack_item_frame(@53,Object[#3])
   ```

Note that compiling and running with external Oracle JDK/JRE is ok.

Also note that rewriting the `switch` as:
```
		var t = switch(i) {
		case 12 -> {
			var r = l.stream().collect(Collectors.joining(" "));
			yield r;
		}
		default -> "";
		};
```
is perfectly ok under Eclipse.

Best, JBY

[...]

PR #4037, opened 2025-05-23
Switch expression and type inference generates wrong/unsecure bytecode which fails to verify

srikanth-sankaran:
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3920


srikanth-sankaran:
One of the return sites of `org.eclipse.jdt.internal.compiler.ast.SwitchExpression.Results.resultType()` inadvertently fails to call org.eclipse.jdt.internal.compiler.ast.SwitchExpression.resolveAsType(TypeBinding) on the return value - which results in required casts not being computed.

iloveeclipse (approved):

Issue #4039, opened 2025-05-23
ECJ Compile Error: "The method sort(List<T>) in the type Collections is not applicable for the arguments ..."


fedejeanne:
Some time **after** the I-Build _I20250515-1800_ (_i.e._ after 4.36 M3), this code stopped compiling:

```java
import java.util.Collection;
import java.util.Collections;
import java.util.Iterator;
import java.util.List;

public class CollectionsSortReproducer {
	class Cranberry<T_Value> implements Comparable<Cranberry<T_Value>> {
		@Override
		public int compareTo(Cranberry<T_Value> o) { return 0; }
	}

	public static void main(String[] args) {
		List<Cranberry<?>> l = null;
	    Collections.sort(l);
	}
}
```

The error says: 

```
The method sort(List<T>) in the type Collections is not applicable for the arguments (List<CollectionsSortReproducer.Cranberry<?>>)
```

Here's yesterday's I-Build (_I20250522-1800_) showing the error:

![Image](https://github.com/user-attachments/assets/8c239da9-acaa-4165-a2ff-666d860ca0d3)

Here's the same code in 20250515-1800 (No error)

![Image](https://github.com/user-attachments/assets/dad4fd10-742e-4b6b-880e-8088528ea485)

srikanth-sankaran:
Only relevant commit in that window is from PR https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4023 and indeed commenting out that changed line in PTB resolves this issue -

iloveeclipse:
> Only relevant commit in that window is from PR [#4023](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4023) and indeed commenting out that changed line in PTB resolves this issue -

That was a fix for https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4022.
If we revert it, we will have https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4022, if we keep it, we will have this issue.

So ideally we would need a fix for both. @srikanth-sankaran, @stephan-herrmann : any chance to get something for RC2?

srikanth-sankaran:
> > Only relevant commit in that window is from PR [#4023](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4023) and indeed commenting out that changed line in PTB resolves this issue -
> 
[...]

PR #4035, opened 2025-05-22
ECJ compile error: "Bound mismatch: The generic method ... is not applicable for the arguments"

stephan-herrmann:
+ "normalize" `? extends capture#1-of X` to `? extends X` by uncapture
+ extend the scope of when IC18 is "active" for (un)capture
+ test adjustments:
  - adjust type display in test expectations
  - expect one more error (also reported by javac)

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4033

PR #4040, opened 2025-05-23
ECJ Compile Error: "The method sort(List<T>) in the type Collections is not applicable for the arguments ..."

stephan-herrmann:
For now reduce the effect of #4023

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4039

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ParameterizedTypeBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ParameterizedTypeBinding.java
@@ -169,10 +169,15 @@ public class ParameterizedTypeBinding extends ReferenceBinding implements Substi
 	}
 	@Override
 	public TypeBinding findSuperTypeOriginatingFrom(TypeBinding otherType) {
+		// see https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4039 :
+		// if capture makes this and otherType equal, then treat that as the sought super type (for now)
+		TypeBinding capture = InferenceContext18.maybeCapture(this);
+		if (TypeBinding.equalsEquals(capture, otherType))
+			return capture;
+
 		if (otherType instanceof ReferenceBinding otherRef && TypeBinding.equalsEquals(this.type, otherRef.actualType()))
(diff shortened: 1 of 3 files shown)


stephan-herrmann:
Here's a generous proposal :)

This indeed PR reduces the effect of #4023 to "fix" the "regression" #4039, *but* we're not on very firm ground here, to tell if that is a regression or a progression in the first place.

Looking at the code I can give neither JLS backing nor common sense reasoning, why this change is "good". The best I can say is, that it is a mean ground between behavior of M3 and current HEAD.

In fact the change goes against what javac says.

Still you may decide to merge this change for now because I also don't have a good explanation for users, why code like CollectionsSortReproducer is illegal.

Meanwhile I will ask Oracle, if they think javac's rejecting this case is correct.

stephan-herrmann:
BTW, should this be merged, the exact error message in `GenericsRegressionTest_1_8.testBug472851()` also depends on  whether or not #4035 is included.

srikanth-sankaran:
@stephan-herrmann - could you please outline the preferred plan ? Are you suggesting we go with https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4040 and https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4041 and ignore https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4035 ? 

Does the combination of https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4040 and https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4041 require any remastering ?

srikanth-sankaran:
> @stephan-herrmann - could you please outline the preferred plan ? Are you suggesting we go with #4040 and #4041 and ignore #4035 ?
> 
> Does the combination of #4040 and #4041 require any remastering of tests ?
> 
> Can you recommend a preferred order of review and merge ?

I am also assuming that we don't have a situation where we are considering all of these three PRs for integration - right ?

srikanth-sankaran:
Please wait for a +1 from @mpalat or @jarthana before integrating.

mpalat:
+1

fedejeanne:
Thank you very much again for the speedy resolution and approval, @stephan-herrmann , @srikanth-sankaran and @mpalat ! 🚀 💪

Issue #4058, opened 2025-05-28
[Cleanup] Vestiges of CLDC support still remain in ECJ


srikanth-sankaran:
Inspect all uses of `org.eclipse.jdt.internal.compiler.classfmt.ClassFileConstants.ATTR_STACK_MAP`

`org.eclipse.jdt.internal.compiler.ClassFile.produceAttributes` never gets assigned this bit flag - so all blocks that read

```
if ((this.produceAttributes & ClassFileConstants.ATTR_STACK_MAP) != 0) {
			attributesNumber += generateStackMapAttribute(
					this.codeStream.methodDeclaration != null ? this.codeStream.methodDeclaration.binding : this.codeStream.lambdaExpression.binding,
					code_length,
					codeAttributeOffset,
					max_locals,
					false,
					scope);
		}
```

are dead code.

`org.eclipse.jdt.internal.compiler.codegen.CodeStream.generateAttributes` is always assigned this bit flag. So all code blocks that read like:

```
if ((this.generateAttributes & (ClassFileConstants.ATTR_VARS
			| ClassFileConstants.ATTR_STACK_MAP_TABLE
			| ClassFileConstants.ATTR_STACK_MAP)) == 0)
		return;
```

are dead code.

Issue #4055, opened 2025-05-28
[Internal][cleanup] org.eclipse.jdt.internal.compiler.ClassFile.initializeDefaultLocals has some weird code in it.


srikanth-sankaran:
The following lines appear bizarre and completely out of place in the loop in which they feature. This needs to be studied if this arose from some merge error:

```
                                                // For the branching complexities in the generated $deserializeLambda$ we need the local variable
						LocalVariableBinding localVariableBinding = new LocalVariableBinding((" synthetic"+i).toCharArray(), typeBinding, 0, true); //$NON-NLS-1$
						localVariableBinding.resolvedPosition = i;
						this.codeStream.record(localVariableBinding);
						localVariableBinding.recordInitializationStartPC(0);
						localVariableBinding.recordInitializationEndPC(codeLength);
```

Issue #4038, opened 2025-05-23
[Record patterns] Verify error with record patterns when case body starts with while loop


srikanth-sankaran:
Brought forward from https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3920#issuecomment-2825234646

This program fails with a verify error at runtime:

```
public class Problem {

    record R(Object o)  {}
 
    static void execute(R stmt) {
        switch (stmt) {
            case R(var _) -> {
                while (cond(null)) {
                    execute(null);
                }
            }
        }
    }
 
    static boolean cond(Object object) {
        return true;
    }
    
    public static void main(String[] args) {}
}
```

Error: Unable to initialize main class X
Caused by: java.lang.VerifyError: Instruction type does not match stack map
Exception Details:
  Location:
    X.execute(LX$R;)V @52: aconst_null
  Reason:
    Current frame's stack size doesn't match stackmap.
  Current Frame:
    bci: @52
    flags: { }
    locals: { 'X$R', 'X$R' }
    stack: { 'X$R', integer }
  Stackmap Frame:
    bci: @52
    flags: { }
    locals: { 'X$R', 'X$R' }
    stack: { }
  Bytecode:
    0000000: 2a59 b800 1057 4c2b 03ba 0016 0000 aa00
    0000010: 0000 0030 0000 0000 0000 0000 0000 0012
    0000020: 2bc0 001a b600 1c57 a700 0c2b 04a7 ffdc
    0000030: 01b8 0020 01b8 0022 9aff f8a7 000d bb00
    0000040: 2659 0101 b700 28bf b1bb 0026 5a5f 59b6
    0000050: 002b 5fb7 0028 bf                      
  Exception Handler Table:
    bci [36, 39] => handler: 73
  Stackmap Table:
    full_frame(@9,{Object[#26],Object[#26]},{Object[#26],Integer})
    same_frame(@32)
    same_frame(@43)
    full_frame(@48,{Object[#26],Object[#26]},{Object[#26],Integer})
    same_frame(@52)
    same_frame(@62)
    chop_frame(@72,1)
    same_locals_1_stack_item_frame(@73,Object[#44])

srikanth-sankaran:
This is targeted for 4.37 - as soon as 4.36 is cooked and the master branch is open I'll release it - if no issues are found.

danielkroeni:
@srikanth-sankaran Thank you!

[...]

Issue #3897, opened 2025-04-04
Compilation error on ECJ but not with JavaC


RoiSoleil:
This code doesn't compile with ECJ :

```
import java.util.ArrayList;

public class CompilationErrorWithECJButNotWithJavac {

	public void main(String[] args) {
		var o = new Object();
		if (o instanceof ListInteger<? extends Integer> list) {
			System.out.println("List");
		}
		System.out.println("Not List");
	}

	class ListInteger<T extends Integer> extends ArrayList<T> {

	}
}
```

![Image](https://github.com/user-attachments/assets/dc9a3a30-1e07-4679-a546-3299cdb9afaa)

But with javac ==> no error

srikanth-sankaran:
Not a recent behavior change - This behavior can be observed at 4.20, probably dates back even earlier,

srikanth-sankaran:
This code doesn't compile with javac:

```
import java.util.List;
import java.util.ArrayList;

public class X {

	public static void main(String[] args) {
		Object o = new Object();
        if (o instanceof List<String> s) {
            System.out.println("Matched!");
        }
		if (o instanceof ListInteger<? extends Integer> list) {
			System.out.println("List");
		}
		System.out.println("Not List");
	}

	class ListInteger<T extends Integer> extends ArrayList<T> {

	}
}
```

I think ECJ behavior is correct - there is a rule about reifiability requirement that javac is not enforcing properly I think - will confirm later - Friday end of the day is not the time to pore over JLS! :)

RoiSoleil:
For me it looks like more a intended feature beacause if i do remove de " extends Integer" on ListInteger, i doesn't compile on Javac anymore. So it looks like there is a check on generic compatibility on javac but not on ECJ. Have a nive Friday ;)

jensli:
I think this difference between ECJ and Javac is because they treat these two things in different ways: 

(1) Wildcards where the bound is the same as the type parameter bound, and
(2) unbounded windcards.

Javac considers (1) and (2) to be the same. ECJ considers them to be different. At least in the context of casts and `instanceof`.

Notice how for ECJ this behaviour generates both an unchecked cast warnings at 1. and an error at 3.

```java
interface I {}

class A<T extends I> {
    T i;
}

public class InstanceOfCastBug {
    public static void main(String[] args) {
        Object o = new A<>();

        // 1. No warning in Javac; "Unchecked cast" warning in ECJ, even if the cast can never fail.
[...]

PR #4061, opened 2025-05-29
Defect fixes for 4.37 

srikanth-sankaran:
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4038
[Record patterns] Verify error with record patterns when case body starts with while loop

* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4058
 [Cleanup] Vestiges of CLDC support still remain in ECJ 

* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3798
BootstrapMethodError / LambdaConversionException using ECJ

* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3895
Reclaim unused bit-flag ExtraCompilerModifiers.AccModifierProblem

* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3897
Compilation error on ECJ but not with JavaC 

* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4065
[Null][Record] Invalid "dead code" warning for record pattern with null-guard on component

* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4055
[Internal][cleanup] org.eclipse.jdt.internal.compiler.ClassFile.initializeDefaultLocals has some weird code in it. 

* Hoist record component discovery into a separate step under `CompleteTypeBindingsSteps` in order to disentangle the design

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/CastExpression.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/CastExpression.java
@@ -458,9 +458,37 @@ public static boolean checkUnsafeCast(Expression expression, Scope scope, TypeBi
 						expression.bits |= ASTNode.UnsafeCast; // upcast since castType is known to be bound paramType
 						return true;
 					default :
-						if (isNarrowing){
-							// match is not parameterized or raw, then any other subtype of match will erase  to |T|
-							expression.bits |= ASTNode.UnsafeCast;
+						if (isNarrowing) {
+							/* So we have a cast `(G<T1, ... Tn>) match` where at least one of the type arguments is NOT an unbounded wildcard
+							   as castType would have been reifiable otherwise and we won't be here. Since `match` is not parameterized or raw,
+							   should the runtime checkcast succeed, match will amount to G<?, ... ?> and so the question distills to whether
+							   the cast `(G<T1, ... Tn>) G<?,...?>` is unsafe
+
+							   JLS 5.1.6.2 states:
+									• A narrowing reference conversion from a type S to a parameterized class or
+									interface type T is unchecked, unless at least one of the following is true:
+									– All of the type arguments of T are unbounded wildcards.
+									– T <: S, and S has no subtype X other than T where the type arguments of X are
+									not contained in the type arguments of T.
+							*/
+
+							ParameterizedTypeBinding ParameterizedCastType = (ParameterizedTypeBinding) castType;
+							TypeBinding [] typeArguments = ParameterizedCastType.typeArgumentsIncludingEnclosing();
+							TypeVariableBinding [] typeVariables = ParameterizedCastType.genericType().typeVariablesIncludingEnclosing();
+
+							if (typeArguments.length != typeVariables.length) { // broken type.
+								expression.bits |= ASTNode.UnsafeCast;
+								return true;
+							}
+							for (int i = 0, length = typeArguments.length; i < length; i++) {
+								TypeVariableBinding tvb = typeVariables[i];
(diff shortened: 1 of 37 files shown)


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/CastExpression.java
```diff
@@ -458,9 +458,37 @@ public static boolean checkUnsafeCast(Expression expression, Scope scope, TypeBi
...
+							   JLS 5.1.6.2 states:
+									• A narrowing reference conversion from a type S to a parameterized class or
+									interface type T is unchecked, unless at least one of the following is true:
+									– All of the type arguments of T are unbounded wildcards.
+									– T <: S, and S has no subtype X other than T where the type arguments of X are
+									not contained in the type arguments of T.
+							*/
+
+							ParameterizedTypeBinding ParameterizedCastType = (ParameterizedTypeBinding) castType;
+							TypeBinding [] typeArguments = ParameterizedCastType.typeArgumentsIncludingEnclosing();
```
stephan-herrmann, line 476: Visually this looks like a static method call - but only because someone is trying to confuse us ;-P

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/CastExpression.java
```diff
@@ -458,9 +458,37 @@ public static boolean checkUnsafeCast(Expression expression, Scope scope, TypeBi
...
+							TypeVariableBinding [] typeVariables = ParameterizedCastType.genericType().typeVariablesIncludingEnclosing();
+
+							if (typeArguments.length != typeVariables.length) { // broken type.
+								expression.bits |= ASTNode.UnsafeCast;
+								return true;
+							}
+							for (int i = 0, length = typeArguments.length; i < length; i++) {
+								TypeVariableBinding tvb = typeVariables[i];
+								TypeBinding checkedTopOfStackTypeArgument =
```
stephan-herrmann, line 485: If at some point I'll need to work on this code, I'll need to ask: how does a stack come into the picture?

srikanth-sankaran: > If at some point I'll need to work on this code, I'll need to ask: how does a stack come into the picture?

It is simply a variable name. The expression that is being casted is on the top of the stack - no ? Perhaps just `expression` would have sufficed ... but is taken for the AST form already. `match` is the super type instance and so I need a new variable to represent the potential subtype instance of type `G<?,...?>`

So in that chunk of code:

`ParameterizedCastType` is the target type of the cast, so `(G<T1, ... Tn>)`
`checkedTopOfStackTypeArgument` is iteratively the type argument of the subtype instance `G<?,...?>`

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/CastExpression.java
srikanth-sankaran: > Visually this looks like a static method call - but only because someone is trying to confuse us ;-P

Sorry, will clean it up.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/CastExpression.java
srikanth-sankaran: If you have suggestions for a better name, I will use that and make the change along with changing the other variable name to have lower case start.

Issue #4011, opened 2025-05-16
Annotation based null analysis with custom annotations leads to misleading info message


marcelziegler:
When annotation based null analysis with custom annotations is activated in eclipse and a function outside of current project (different plugin/3-party-lib) is called and the called function contains a typed parameter and the function returns an instance of this type and the retured value is used (variable assignment/returned etc.) the follwing info message is reported:

![Image](https://github.com/user-attachments/assets/b281bce1-af06-405f-a452-0cacb7497c2d)

The calling project does not use any of the custom annotations. 

Also the reported massage cannot be supressed by the option "Unsafe conversion of annotated parameterized type to less-annotated type".

![Image](https://github.com/user-attachments/assets/a0363264-fc3a-48ee-a6f6-e59dd3ccc444)

![Image](https://github.com/user-attachments/assets/26f75e66-586b-4222-8332-fa3fb1184fac)

E.g. calling java.nio.file.Files.readAttributes(Path, Class<A>, LinkOption...) or org.mockito.Mockito.mock(Class<T>). 

Sample project for reproduction: 
[NullAnalysisTest.zip](https://github.com/user-attachments/files/20243659/NullAnalysisTest.zip)

We got over 50k messages in our project of that kind so any help is appreciated.

Thanks, 
Greetings Marcel

stephan-herrmann:
> The calling project does not use any of the custom annotations.

Did you mean to write "the project containing the **called** method"? Otherwise I wouldn't understand where those custom annotations are used at all.

stephan-herrmann:
Let's see what is going on here:

Regarding this method ...
```
  public BasicFileAttributes test() throws IOException {
    return Files.readAttributes(Path.of("file://xxx"), BasicFileAttributes.class); //$NON-NLS-1$
  }
```
... you could **avoid the warning by marking the method return type as `@Nullable`.** This would tell the analysis that we don't need to worry about nullness of what `readAttributes()` returns.

----

I'm slightly surprised to see the warning against an **unannotated method return type**. More on that later.

----

[...]

PR #4073, opened 2025-06-06
Annotation based null analysis with custom annotations leads to misleading info message

stephan-herrmann:
+ Avoid this warning if nonnull is not required

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4011


stephan-herrmann:
The situation is a bit trickier than I thought: when the value with unsure nullness is assigned to a local variable, we want local flow analysis to take over and hence we typically don't have null annotations on the local's type. Should this mean we discard the nullness inferred during type inference??

Otherwise we need to restrict the change to cases where the value in question crosses a method boundary (either as return value, or as argument to another method call or as assignment to a field).

Issue #3889, opened 2025-04-03
[25] JEP 507 - Primitive Type in patterns, instanceof, and switch (Third Preview)


mpalat:
JEP: https://openjdk.org/jeps/507
Also tracked via https://bugs.openjdk.org/browse/JDK-8349215

stephan-herrmann:
As per https://bugs.openjdk.org/browse/JDK-8354325 spec changes are here: https://cr.openjdk.org/~abimpoudis/instanceof/latest

Current Changelog:
> 2025-04-29: Update links and dates. No changes.

😎

stephan-herrmann:
Nothing to be done here, the JEP has been marked Closed / Delivered with no changes to 24.

stephan-herrmann:
> Nothing to be done here, the JEP has been marked Closed / Delivered with no changes to 24.

not really done, see #4254

PR #4077, opened 2025-06-07
[25] JEP 511 Module Import Declarations

stephan-herrmann:
+ feature is graduating
+ remove tests only challenging the preview-aspect
+ adjust others to no longer enable preview

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3886

PR #4078, opened 2025-06-07
[25] JEP 507 - Primitive Type in patterns, instanceof, and switch (Third Preview)

stephan-herrmann:
+ https://bugs.openjdk.org/browse/JDK-8348410 has been fixed
  - define upper limit for excuse JavacBug8348410
  - actually pass --enable-preview to javac in one more test

Relates to https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3889


stephan-herrmann:
The test failures seen here are being addressed via #4077

Issue #4081, opened 2025-06-09
ecj doesn't properly apply the comb rule from 15.12.1


stephan-herrmann:
Here's org.eclipse.jdt.core.tests.compiler.regression.InnerEmulationTest.test172() from https://bugs.eclipse.org/197271:
```
public class X {
         void a() {}
         private static void a(String s) {}
         private void c() {}
         private static void c(String s) {}
         static class M1 extends X {
                 public void x() {
                         a(null);
                         c(null);
                 }
         }
         static class M2 {
                 public void x() {
                         a(null);
                         c(null);
                 }
         }
}
```

According to the comb rule from [15.12.1](https://docs.oracle.com/javase/specs/jls/se24/html/jls-15.html#jls-15.12.1) the first invocation `a(null)` should first look at members from the hierarchy of M1. Here we find `void a() {}` (inherited from X), so M1 is the type to search in, not X.

Also considering that private members are not inherited, none of the other methods are available via M1.

Ergo, despite the "matching" method `private static void a(String s) {}` resolution is stuck with the arg-less method and must then fail due to mismatching number of formal vs. actual arguments.

PR #4087, opened 2025-06-10
ecj doesn't properly apply the comb rule from 15.12.1

stephan-herrmann:
+ don't find private methods from super types
  + unless there is only one private method, to be reported as invisible

+ support running with run.javac on JDK 24.0.1

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4081

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
@@ -1684,6 +1684,9 @@ public abstract class Scope {
 		// superclass lookup
 		ReferenceBinding classHierarchyStart = currentType;
 		MethodVerifier verifier = environment().methodVerifier();
+		boolean currentIsSuper = false;
+		MethodBinding singlePrivateMethod = null;
+		boolean multiplePrivateMethods = false;
 		while (currentType != null) {
 			unitScope.recordTypeReference(currentType);
 			currentType = (ReferenceBinding) currentType.capture(this, invocationSite == null ? 0 : invocationSite.sourceStart(), invocationSite == null ? 0 : invocationSite.sourceEnd());
@@ -1718,6 +1721,20 @@ public abstract class Scope {
 							}
 						}
 					}
+				} else if (currentIsSuper) {
+					for (int i = 0, l = currentLength; i < l; i++) { // currentLength can be modified inside the loop
+						MethodBinding currentMethod = currentMethods[i];
+						if (currentMethod.isPrivate()) {
+							if (singlePrivateMethod == null && !multiplePrivateMethods) {
+								singlePrivateMethod = currentMethod;
+							} else {
+								singlePrivateMethod = null;
+								multiplePrivateMethods = true;
+							}
+							currentLength--;
+							currentMethods[i] = null;
+						}
+					}
 				}
 
 				if (currentLength > 0) {
@@ -1733,6 +1750,7 @@ public abstract class Scope {
 				}
 			}
 			currentType = currentType.superclass();
+			currentIsSuper = true;
 		}
 
 		// if found several candidates, then eliminate those not matching argument types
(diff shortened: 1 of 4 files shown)


stephan-herrmann:
@srikanth-sankaran do you want to comment on this change?

I think this case is interesting because:
* it modifies our fundamental method `Scope.findMethod0()`
* the current behavior decided upon in 2008 (https://bugs.eclipse.org/197271) turned out to be wrong only now, see https://mail.openjdk.org/pipermail/compiler-dev/2025-June/030746.html
* that old bug has a patch proposed by Kent but withdrawn based on the (wrong) decision to keep behavior unchanged. The outcome expected in the test in that patch, however, also looks wrong from today's perspective.
* in my change I simply test `isPrivate()`, but haven't verified if this is sufficient, or if more needs to be done to check if a given method from a supertype is actually _inherited_.
* the change influences also a few existing error messages (like going from cannot be resolved to is not visible).

srikanth-sankaran:
Sorry for the delay - will help wrap this up this week.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/JavadocTestForMethod.java
```diff
@@ -4130,7 +4130,7 @@ public void test147() {
 				+ "1. ERROR in test\\X.java (at line 6)\n"
 				+ "	* @see Visibility#avm_private(int) Invalid ref: non-applicable inherited method\n"
 				+ "	                  ^^^^^^^^^^^\n"
-				+ "Javadoc: The method avm_private() in the type AbstractVisibility is not applicable for the arguments (int)\n"
+				+ "Javadoc: The method avm_private() from the type AbstractVisibility is not visible\n"
```
srikanth-sankaran, line 4133: FWIW: Javac issues a bland `error: reference not found` for all 4 cases here.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
```diff
@@ -1718,6 +1721,20 @@ public MethodBinding findMethod0(ReferenceBinding receiverType, char[] selector,
 							}
 						}
 					}
+				} else if (currentIsSuper) {
+					for (int i = 0, l = currentLength; i < l; i++) { // currentLength can be modified inside the loop
```
srikanth-sankaran: I have stared at this code for a while trying to convince myself that it is correct that this new block features as the `else` of `if ((receiverTypeIsInterface || found.size > 0)) {}` - I am lost!

I can't construct a case where it can lead to a problem, nor am I able to feel all good about it.

Perhaps you could state a few words why ?

srikanth-sankaran: > I can't construct a case where it can lead to a problem, nor am I able to feel all good about it.

Actually, see test case above where our patched behavior differs from javac.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/JavadocTestForMethod.java
stephan-herrmann: Isn't it nicer to tell "not visible" rather than "not found"?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
stephan-herrmann: Thanks for catching. The `else` was indeed bogus.

On a second look I figured it's best to integrate the new filtering (for private from super) with the existing filtering for things that have been nulled out on the other branch.

Does it look better, now?

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/JavadocTestForMethod.java
srikanth-sankaran: > Isn't it nicer to tell "not visible" rather than "not found"?

Yes, hence the adjective bland :)

srikanth-sankaran (approved):
Latest changes look good to me.

Do the newly added/modified tests capture all nuances from JLS Example 6.5.7.1-1 ? If not, we may add that too.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/JavadocTestForMethod.java
stephan-herrmann: > Yes, hence the adjective bland :)

I should have consulted my dictionary :)  -- learning new words all the time.

Issue #3976, opened 2025-05-06
[25] JavaSearchBugs16Tests.testAIOOBEForRecordClassGh790() throws AssertionError


stephan-herrmann:
Running the given test on JDK 25 raises
`java.lang.AssertionError: PlainPackageBinding cannot be parent of split package sun.text.spi (from java.base, java.base)`

This happens when we try to resolve a PossibleMatch for `module-info.class [in <default> [in <module:java.base>]]`

At that point we have both a SourceModuleBinding and a BinaryModuleBinding for java.base, producing a pseudo split package.

I'm not sure if we can accept this schizophrenia of a module, but then we probably cannot avoid the BinaryModuleBinding of the premordial module, nor can we use it for reference search, so we need the source variant, too :-/

PR #4095, opened 2025-06-12
[25] JavaSearchBugs16Tests.testAIOOBEForRecordClassGh790() throws AssertionError

stephan-herrmann:
+ boldly reduce split package from source & binary versions of the same module to just one constituent.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3976


stephan-herrmann:
This change is a mere band-aid to avoid creation of an unwanted SplitPackageBinding, by reducing it to one of its constituents. I don't even want to think about what happens downstream with this arbitrarily chosen PlainPackageBinding. 

The question is not, whether the fix is correct (it is not), but rather: how much should we care about the initial situation in the first place? How much do we want to invest in its solution?

PR #4105, opened 2025-06-17
Fix super interfaces for captures using interfaces from the type variable when the super class has more specific interfaces

### LINKED ISSUE #2523: Method call compiles with javac but not ecj with upper bound of type variable
When attempting to compile this:
```java
    public void test() {
    	C<? extends D> c = null;
        optional(c, null);
    }

    private <I extends F, T extends E<I>> void optional(C<T> l, I i) {}

    public static interface C<T extends E<?>> {}
    public static class D implements E<F> {}
    public static interface E<I extends F> {}
    public static class F {}
```
ecj fails with the error: `The method optional(Outer.C<T>, I) in the type Outer is not applicable for the arguments (Outer.C<capture#1-of ? extends Outer.D>, null)` while javac successfully compiles.
If I remove the ` extends E<?>` from the `C` interface ecj also successfully compiles.

This seems to be due to:
https://github.com/eclipse-jdt/eclipse.jdt.core/blob/41d22845ac45f2a4179aa52b15bf0faa77ef6857/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding.java#L257
setting the super interfaces from the type variable allowing for `I#0 = ?` during inference where the wildcard is obtained from the super interface of the capture instead of checking if the super interface of the class has more specific interfaces.

coehlrich:
## What it does
Fixes #2523 by only setting super interfaces that aren't already an interface of the super class.

## How to test
Attempt to compile the test case.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding.java
@@ -247,6 +247,23 @@ public class CaptureBinding extends TypeVariableBinding {
 						// TODO: there are cases were we need to compute glb(capturedWildcardBound, substitutedVariableSuperclass)
 						//       but then when glb (perhaps triggered inside setFirstBound()) fails, how to report the error??
 					}
+					if (this.superclass.id != TypeIds.T_JavaLangObject) {
+						int length = substitutedVariableInterfaces.length;
+						ReferenceBinding[] superTypes = new ReferenceBinding[length + 1];
+						System.arraycopy(substitutedVariableInterfaces, 0, superTypes, 1, length);
+						superTypes[0] = this.superclass;
+						superTypes = Scope.greaterLowerBound(superTypes);
+						substitutedVariableInterfaces = new ReferenceBinding[superTypes.length - 1];
+						boolean foundSuper = false;
+						for (int i = 0; i < superTypes.length; i++) {
+							ReferenceBinding current = superTypes[i];
+							if (current.isInterface()) {
+								substitutedVariableInterfaces[i - (foundSuper ? 1 : 0)] = superTypes[i];
+							} else {
+								foundSuper = true;
+							}
+						}
+					}
 					this.setSuperInterfaces(substitutedVariableInterfaces);
 				}
 				this.setFirstBound(capturedWildcardBound);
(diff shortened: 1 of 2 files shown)


stephan-herrmann:
Cool, someone really is in a positive flow here.

Did you see https://bugs.eclipse.org/bugs/show_bug.cgi?id=521212#c5? This is when I added the code comment just above the paragraph you inserted:
>// TODO: there are cases were we need to compute glb(capturedWildcardBound, substitutedVariableSuperclass)
>//              but then when glb (perhaps triggered inside setFirstBound()) fails, how to report the error??

This looks very related to your change :smile:

coehlrich:
I added a check to bypass the `greaterLowerBound` check if the `substitutedVariableInterfaces` is empty, added handling for the case if the super class gets removed, set the `substitutedVariableInterfaces` to `Binding.NO_SUPERINTERFACES` when there are no remaining interfaces and incase `greaterLowerBound` returns null (even though for the reasons mentioned below I don't see any way for `greaterLowerBound` to return null).

The comment seems more targeted towards the super class of the capture (which would have to error if the returned array contains more than 1 entry) while this sets the super interfaces more similarly to when the `originalWildcardBound` is an interface and `substitutedVariableInterfaces` is not empty except with the super class included in the `greaterLowerBound` call.

glb can only return null in 4 cases:
- a null array is passed in
- an empty array is passed in
- an incompatible class and type variable is passed in
- all entries get removed

I don't see any way for any of those conditions to be met because:
- `superTypes` cannot be null
- `superTypes` always contains atleast the super class and `filterValidTypes` always keeps atleast 1 entry
- the super class is the only class that is passed in and type variables are only included as super classes so there can never be a regular class and a type variable passed in
- I don't see any way for `greaterLowerBound` to remove all entries since it needs atleast 2 entries to remove 1 entry

stephan-herrmann:
@coehlrich this looks solid, much appreciated!

Next, your change inspired me to question the current implementation of `CaptureBinding.initializeBounds()` more broadly. In previous years I was always afraid of touching this method due to its "colorful" history, including things like conditionally applying capture *recursively*, which definitely is against JLS, but previous JDT committers saw indications that this is needed sometimes anyway, etc. Some of this got simplified when support for version < 8 was dropped. Good.

Now with a fresh look also at JLS I wonder if glb should be applied more consistently, see [§5.1.10](https://docs.oracle.com/javase/specs/jls/se24/html/jls-5.html#jls-5.1.10):
> If Ti is a wildcard type argument of the form ? extends Bi, then Si is a fresh type variable whose upper bound is glb(Bi, Ui[A1:=S1,...,An:=Sn]) and whose lower bound is the null type. 

Ui is the bound of the corresponding type parameter, where this bound can in fact be an intersection type.

I will upload an experiment in this direction.

stephan-herrmann:
I selected the fix in #4116 as it generally simplifies this code block. I kept @coehlrich as a co-author, since the use of glb in this PR was the central idea for improvement. Thanks!

PR #4116, opened 2025-06-19
Method call compiles with javac but not ecj with upper bound of type variable

stephan-herrmann:
Fix CaptureBinding.initializeBounds() by more closely adhering to JLS

Fix a bug in glb concerning T#RAW vs T<X> or T<?>
+ this fix includes making RawTypeBinding.isSubtypeOf() stricter

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2523

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding.java
@@ -22,6 +22,7 @@
  *******************************************************************************/
 package org.eclipse.jdt.internal.compiler.lookup;
 
+import java.util.Arrays;
 import org.eclipse.jdt.core.compiler.CharOperation;
 import org.eclipse.jdt.internal.compiler.ast.ASTNode;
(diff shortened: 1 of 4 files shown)


stephan-herrmann:
@coehlrich  do you want to join me discussing which solution for #2523 is more suitable?

stephan-herrmann:
The basic idea here is to collect all relevant bounds from the wildcard and the corresponding type variable, perform substitution and then compute glb of them all.

If glb is defined, simply pick the class, if any, make it the superclass, and put the rest into superInterfaces.

The else-branch (the case of `fullGlb==null`) is less rigorously motivated. In many cases this branch could actually be avoided if `getAllAllUpperBounds()` would filter out any reference to `java.lang.Object`. There seems to be more to it, though. Could null glb indicate an actual bound error? Would this (always) correlate to a bound error of the parameterized type being captured?

Interestingly, this experiment triggered one regression which could be traced back to a bogus result of glb: we were computing `glb(T#RAW, T<?>)` as `T#RAW`. But `T#RAW` is not a subtype of `T<?>`, it is only convertible to it. This could be fixed by small changes in `greaterLowerBound()` itself plus an addition in `RawTypeBinding.isSuperTypeOf()`. These changes might be valuable in their own right.

Issue #4043, opened 2025-05-26
[25] JEP 506: Scoped Values


mpalat:
ref: https://openjdk.org/jeps/506

Summary
Introduce scoped values, which enable a method to share immutable data both with its callees within a thread, and with child threads. Scoped values are easier to reason about than thread-local variables. They also have lower space and time costs, especially when used together with virtual threads ([JEP 444](https://openjdk.org/jeps/444)) and structured concurrency ([JEP 505](https://openjdk.org/jeps/505)).

ref: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3292 for en ealier issue.

subyssurendran666:
This does not affect JDT.core, therefore we can close this.

PR #4119, opened 2025-06-19
[25] JEP 506: Scoped Values

stephan-herrmann:
Feature has graduated as of 25-ea+27

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4043


stephan-herrmann:
Test fail since jdt ci.eclipse.org/jdt uses jdk-25/25-ea+**25** whereas Y-builds are already at jdk-25/25-ea+**27**.

I guess we can live with 3 known failures in Y-build tests on 25 until we next synchronize ea builds.

PR #4127, opened 2025-06-21
PR-2107 follow-up: avoid sharing AST nodes

stephan-herrmann:
PR 2107 copies a reference to an AST node (local declaration) which implies that the same node will be visited twice for all phases.

Let's play safe and _move_ instead of _copy_ in order to maintain that AST is indeed a tree :)


stephan-herrmann:
@snjeza @rgrunber FYI: this change improves on PR #2107, where this line:
```java
this.currentElement.add(pattern.local, 0);
```
actually caused the node `local` to be contained _twice_ in the overall AST. This violates the AST's tree-property :smile: and should always be avoided. In fact by lifting the local outside the if statement, the instanceof pattern is no longer needed, so I made it a _move_ by deleting the old reference.

In current ecj this doesn't seem to harm, but resolving, analyzing ... any node more than once is not guaranteed to behave well.

Indeed I noticed this not during JDT work, but in the fork for Object Teams, where just a slight modification in other parts of the code created a conflict with this "shared" AST node.

PR #3900, opened 2025-04-05
Add multi-release compilation support to JDT

laeubi:
Currently JDT (like other IDEs) lacks good support for compilation of so called [multi-release-jar](https://openjdk.org/jeps/238) setups. Maven has quite good support for this by setting a `${release}` property per source folder and [enable an output flag](https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#multiReleaseOutput) but often needs some ugly workarounds due to missing support in IDEs.

This is an attempt to bring multi-release-compilation to JDT to finally have this feature supported at least in one of the major IDEs and make this powerful feature more widely adopted.

In its current form it does the following:

1. One can set a `release` attribute in the classpath of a source folder, this can look like this:
```
<?xml version="1.0" encoding="UTF-8"?>
<classpath>
	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-11">
		<attributes>
			<attribute name="module" value="true"/>
		</attributes>
	</classpathentry>
	<classpathentry kind="src" path="src"/>
	<classpathentry kind="src" path="src9">
		<attributes>
			<attribute name="release" value="9"/>
		</attributes>
	</classpathentry>
	<classpathentry kind="output" path="bin"/>
</classpath>
```
2. if such attribute is found JDT enables "multi-release-output" as per specification to place the compiled output of this source folder to `<outputfolder>/META-INF/versions/<release>`

The next step would be to further pass the release option to the compiler itself.

PR #4143, opened 2025-06-25
Make release dependent module loading static

laeubi:
Currently the ClasspathJrt uses a static loadModules method while the ClasspathJrtWithReleaseOption uses an instance method (that effectively modifies static state). This makes it harder to reuse and is prone to changes in the handling of instance fields.

This now extracts the part into the base class ClasspathJrt and passing the CtSym and release code as a parameter instead and streamlining the logic if modules should be loaded or not to allow reuse and easier debugging.

I came across this while working on 
- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3900

 where it will be needed to more dynamically load release specific classes/information from the Jrt.

Once this is done I also plan to further improve the caching here, that currently caches all loaded modules forever in a static map without any way for the GC to clear things, but this is another topic.

@stephan-herrmann can you review / merge this one as it would make the changes in my PR much more streamlined.

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/ClasspathJrt.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/ClasspathJrt.java
@@ -16,7 +16,10 @@ package org.eclipse.jdt.internal.core.builder;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.FileVisitResult;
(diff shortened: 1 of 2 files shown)


laeubi:
Hm... need to look into the test failure

laeubi:
Closing for now as there is already a better variant in the PR... still currently loading the modules eagerly seems not the best choice.

Issue #4146, opened 2025-06-26
Unable to build Record with components annotated with Qualified name reference


RoiSoleil:
This kind of Record throws a NPE when trying to build :

```
package repro;

import com.fasterxml.jackson.annotation.JsonInclude;
import com.fasterxml.jackson.dataformat.xml.annotation.JacksonXmlProperty;

public record Segment(@JacksonXmlProperty(isAttribute = true) String id,
                      String source,
                      @JsonInclude(JsonInclude.Include.NON_NULL) String target) {

}
```

with the stacktrace of the NPE being :

```
java.lang.NullPointerException: Cannot invoke "org.eclipse.jdt.internal.compiler.ast.AbstractVariableDeclaration.getKind()" because "variableDeclaration" is null
	at org.eclipse.jdt.internal.compiler.lookup.SourceTypeBinding.resolveTypeFor(SourceTypeBinding.java:929)
	at org.eclipse.jdt.internal.compiler.lookup.SourceTypeBinding.resolveTypeFor(SourceTypeBinding.java:1013)
	at org.eclipse.jdt.internal.compiler.lookup.ClassScope.collateRecordComponents(ClassScope.java:1407)
	at org.eclipse.jdt.internal.compiler.lookup.CompilationUnitScope.collateRecordComponents(CompilationUnitScope.java:417)
	at org.eclipse.jdt.internal.compiler.lookup.LookupEnvironment$CompleteTypeBindingsSteps.perform(LookupEnvironment.java:192)
	at org.eclipse.jdt.internal.compiler.lookup.LookupEnvironment.completeTypeBindings(LookupEnvironment.java:564)
	at org.eclipse.jdt.internal.compiler.Compiler.internalBeginToCompile(Compiler.java:787)
	at org.eclipse.jdt.internal.compiler.Compiler.beginToCompile(Compiler.java:277)
	at org.eclipse.jdt.internal.compiler.Compiler.compile(Compiler.java:327)
	at org.eclipse.jdt.internal.compiler.Compiler.compile(Compiler.java:309)
	at org.eclipse.jdt.internal.core.builder.AbstractImageBuilder.compile(AbstractImageBuilder.java:441)
	at org.eclipse.jdt.internal.core.builder.BatchImageBuilder.compile(BatchImageBuilder.java:250)
	at org.eclipse.jdt.internal.core.builder.AbstractImageBuilder.compile(AbstractImageBuilder.java:372)
[...]

PR #4148, opened 2025-06-26
Unable to build Record with components annotated with Qualified name reference

srikanth-sankaran:
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4146


RoiSoleil:
On my code base it's ok now. Thx.

srikanth-sankaran:
> On my code base it's ok now. Thx.

Thanks for checking, actually there is a follow on commit that is on its way that ensures annotations are propagated even when there is a phase ordering problem due to reentrancy,

Issue #4149, opened 2025-06-26
ECJ Compile error: "The method ... from the type ... is not visible"


fedejeanne:
Somewhere between 4.36 M3 (**I20250515-1800**) and 4.37 M1 (**I20250625-2330**), this code stopped compiling in Eclipse:

```java
public class ReproducerDiamondMethod extends A implements B {
	void callIt() {
		init();
	}
}

class A {
	private void init() {};
}

interface B {
	default void init() {};
}
```

I know this is an odd constellation but the fact that it stopped working makes it relevant and I wanted to ask: is this an error?

**Additional information**
- `javac` compiles the code just fine

```
$> javac .\ReproducerDiamondMethod.java
$> javac --version
javac 24.0.1
```

srikanth-sankaran:
@stephan-herrmann can you take look ? Is this a side effect of the comb rule fix ??

fedejeanne:
Thank you for the hint @srikanth-sankaran ! I can confirm that **commit 3a87bcb11072bd7c10b97d4a8c55dc6b3c6d6e66 introduced the issue.** I simply reverted `Scope` to the previous state, dropped the generated `.class` files into the appropriate JAR in my P2 bundle (`org.eclipse.jdt.core.compiler.batch_3.42.50.v20250628-1110`) and the snippet compiled just fine.

fedejeanne:
🚀

PR #4156, opened 2025-07-01
ECJ Compile error: "The method ... from the type ... is not visible"

stephan-herrmann:
defer returning invisible private method until after interfaces lookup

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4149


fedejeanne:
Thank you very much for the fix!

stephan-herrmann:
> Thank you very much for the fix!

Thanks for the report!

PR #4168, opened 2025-07-07
Improve Performance of getInterfaceAbstractContracts() for interfaces with many inherited Methods

flori-schwa:
Our data model uses interface with complex type hierarchies and large amounts of methods. During Code-Completion the "isFunctionalInterface" Call in `org.eclipse.jdt.internal.codeassist.CompletionEngine.findLambdaExpressions(Scope)` can take up several seconds when an instance of one such interface is expected.

After further analysis I found out, that the `org.eclipse.jdt.internal.compiler.lookup.ReferenceBinding.getInterfaceAbstractContracts(Scope, boolean, boolean)` Method recursively collects all abstract Methods of the Superinterfaces and checks if any of them are overriden. 

## What it does
This PR improves the Performance of the `getInterfaceAbstractContracts()` Method (which is used by `getSingleAbstractMethod()`, which itself is used by `isFunctionalInterface()`) for interfaces with many inherited Methods by grouping the Methods by their Name (`org.eclipse.jdt.internal.compiler.lookup.MethodBinding.selector`) and only checking for overriding Methods in these groupings. This should not change which Methods are contained in the final array. However, due to the grouping, the order can be different.

## How to test
Import the java file in the attached zip file ([Main.zip](https://github.com/user-attachments/files/21102092/Main.zip)) in a project and perform Code Completion (CTRL+Space at the commented line).
There should be a measurable Performance Improvement. Although it is difficult to reproduce this issue using sample code, the code changes noticably improve the performance of the code completion in a productive workspace with our data model:
Before applying the patch in a productive workspace
grafik
After applying the patch in a productive workspace (Code completion performed on the same line of code)
grafik

## Author checklist

  - I have ran the `org.eclipse.jdt.core.tests.compiler.regression.TestAll` Testsuite and no problems regarding (functional) interfaces were found.
    - Only two Testcases regarding the Activation of Java Preview-Features failed (in `org.eclipse.jdt.core.tests.compiler.regression.PreviewFeatureTest`). I am not sure if or how my changes could have caused these Tests to fail

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
@@ -54,9 +54,12 @@ package org.eclipse.jdt.internal.compiler.lookup;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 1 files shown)


srikanth-sankaran:
I'll review this certainly for M3, but will try to do so opportunistically in the M2 time frame itself. Thanks for the PR @flori-schwa

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
```diff
@@ -2332,91 +2335,88 @@ public void detectWrapperResource() {
...
-			throw new InvalidBindingException("Not a functional interface"); //$NON-NLS-1$
-		for (int j = 0; j < contractsCount;) {
-			if ( contracts[j] != null && MethodVerifier.doesMethodOverride(method, contracts[j], environment)) {
-				contractsCount--;
-				// abstract method from super type overridden by present interface ==> contracts[j] = null;
-				if (j < contractsCount) {
-					System.arraycopy(contracts, j+1, contracts, j, contractsCount - j);
-					continue;
+	if (methods != null && methods.length > 0) {
+		for (final MethodBinding method : methods) {
```
fedejeanne: NIT: how about reverting back to the previous version so that the PR is a bit easier to review? Maybe @srikanth-sankaran appreciates that?

fedejeanne (commented):
Thank you for tackling this issue, @flori-schwa ! 

I'm not a JDT committer but I do have some tips regarding PRs, just to make it easier for the committers to do their part (based on my own experience as a committer in Platform :-) )

1. Please try to _squash_ your commits so that the PR has 1 commit (you currently have 3). An exception to this rule is when you have to bump versions in the manifests/poms
2. Maybe you could reduce the amount of changes in the coding style and stick to the factual changes? This one is my very own personal opinion so check with @srikanth-sankaran , but if I were to review/merge this one, I'd find it easier to do so if there were less (or no) changes like the one in lines 2361-2362, where you split a `for` into and `if` and a `for`. If I revert back to the previous style, I can clearly see that you made changes in the line starting with `List<MethodBinding> contractsOfSelector = ...`, which is easier to spot.

Again, that last one is my personal opinion. Ask the committer (@srikanth-sankaran in this case) to see if such nit-picky details are appreciated :-) 

HTH!

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
flori-schwa: I personally think this change makes the code more readable here. It makes it clear, that the for loop is not executed, if the Array is null. I just extracted this logic from the old for loop initializer.

I could, however, remove the redundant length check

srikanth-sankaran:
Hello @flori-schwa - I expect to start reviewing this early next week.  

Could you share the initial and final performance numbers please ? Right now I am unable to click on the images you have shared at the top. I get this error: `This private-user-images.githubusercontent.com page can’t be found`

I assume disabling lambda proposals (if that were possible via a knob) is not what you want - right ?

flori-schwa:
Hi @srikanth-sankaran 

When an instance of an interface from our data model is expected, Code Completion currently takes approximately 3-4 seconds. With the patch, Code Completion finishes in under a second.
(The Total time of the `isFunctionalInterface` Method went down to 500ms from 3100ms)

Disabling Lambda proposals would've also fixed the performance issue for us here. But as far as I can tell, it is not possible to configure this in the Content-Assist settings. (And would've also disabled them when working with functional interfaces)

srikanth-sankaran:
I think the problem is that `org.eclipse.jdt.internal.compiler.lookup.ReferenceBinding.getInterfaceAbstractContracts(Scope, boolean, boolean)` is too much of a general purpose hammer. Its only client is `org.eclipse.jdt.internal.compiler.lookup.ReferenceBinding.getSingleAbstractMethod(Scope, boolean)` which could benefit from a fail fast mode which `getInterfaceAbstractContracts` does not provide.

`getInterfaceAbstractContracts` works hardly but not smartly in doing its job 

I worry that the present patch - which I am still in the beginning stages of understanding - attempts to _speed up the hard work_ rather than coming up with a fail fast mode.

To fail fast, I wonder if we should _rotate_ the method - by that I mean before walking the super interfaces, walk own contract - if there are two abstract methods already - we can fail fast already without ever walking the super types.

I'll propose a patch shortly that does this and we can discuss further.

srikanth-sankaran:
> I'll propose a patch shortly that does this and we can discuss further.

PR here https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4228 - passes some smoke tests locally. Let us wait for the full run.

With this patch, when I attempt code completion on the provided test case from Main.zip, it doesn't even go into `getInterfaceAbstractContracts()` or `getSingleAbstractMethod()`

I took some numbers wrapping the call to `isFunctionalInterface` from `org.eclipse.jdt.internal.codeassist.CompletionEngine.findLambdaExpressions(Scope)` using `System.nanoTime();`  and this is what I saw (operation repeated a few times and first couple of numbers discarded)

Without the patch:

Duration: 3900 ns
Duration: 4200 ns
Duration: 2700 ns
Duration: 3500 ns


With the patch 

Duration: 1000 ns
Duration: 900 ns
Duration: 900 ns
Duration: 1100 ns

srikanth-sankaran:
We should still look at rotating the implementation of `getInterfaceAbstractContracts()` so it inspects own contract first before looking at inherited contract

srikanth-sankaran:
> PR here #4228 - passes some smoke tests locally. Let us wait for the full run

The test run is green. @flori-schwa can you try this patch to see if it addresses your concerns. 

If it does, I'll release this and raise a follow up ticket to implement fail fast mode into getInferfaceAbstactContracts by rotating the implementation and also providing  other smarts and follow up leisurely

srikanth-sankaran:
> can you try this patch to see if it addresses your concerns.

I have integrated this with master just now. You can try a subsequent I-Build

> I took some numbers wrapping the call to `isFunctionalInterface` from `org.eclipse.jdt.internal.codeassist.CompletionEngine.findLambdaExpressions(Scope)` using `System.nanoTime();` and this is what I saw (operation repeated a few times and first couple of numbers discarded)
> 
> Without the patch:
> 
> Duration: 3900 ns Duration: 4200 ns Duration: 2700 ns Duration: 3500 ns
> 
> With the patch
> 
> Duration: 1000 ns Duration: 900 ns Duration: 900 ns Duration: 1100 ns

Something wrong with the above. I don't know what exactly -

Here are fresh numbers on master and before commit https://github.com/eclipse-jdt/eclipse.jdt.core/commit/de4eb32cf081f82f739b599d5bdc2e5ce670aac3 respectively: (once again ignoring first few runs)

master (with https://github.com/eclipse-jdt/eclipse.jdt.core/commit/de4eb32cf081f82f739b599d5bdc2e5ce670aac3)

Duration = 6200
Duration = 4700
Duration = 4200
Duration = 6300
Duration = 4300

Without https://github.com/eclipse-jdt/eclipse.jdt.core/commit/de4eb32cf081f82f739b599d5bdc2e5ce670aac3

Duration = 195260800
Duration = 199898300
Duration = 200422600
Duration = 255901100
Duration = 178698100

All numbers in nanoseconds.

FTR, here is the patch I used for benchmarking:

```
diff --git a/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java b/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
index a6aeec8..dc1d2bc 100644
--- a/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
+++ b/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
@@ -14633,7 +14633,11 @@
 			for (int i = 0; i <= this.expectedTypesPtr; i++) {
 				TypeBinding type = this.expectedTypes[i];
 
-				if (type.isFunctionalInterface(scope)) {
+				long start = System.nanoTime();
+				final boolean functionalInterface = type.isFunctionalInterface(scope);
+				long stop = System.nanoTime();
+				System.out.println("Duration = " + (stop - start));
+				if (functionalInterface) {
 					int relevance = computeBaseRelevance();
 					relevance += R_EXACT_EXPECTED_TYPE;
 					relevance += R_ABSTRACT_METHOD;
```

srikanth-sankaran:
> `getInterfaceAbstractContracts` works hardly but not smartly in doing its job
> 
> I worry that the present patch - which I am still in the beginning stages of understanding - attempts to _speed up the hard work_ rather than coming up with a fail fast mode.
> 
> To fail fast, I wonder if we should _rotate_ the method - by that I mean before walking the super interfaces, walk own contract - if there are two abstract methods already - we can fail fast already without ever walking the super types.

Interesting revelations by software archeology:

The original implementation of `getInterfaceAbstractContracts`  did have a fail fast mechanism implemented in commit 5138a70372af4817aefdd3da44dfadf7f7557bf3 for https://bugs.eclipse.org/bugs/show_bug.cgi?id=399224. But this got undone during the fix for https://bugs.eclipse.org/bugs/show_bug.cgi?id=514105 via the commit 823c9ddd71537118b17b028f2c504b1e97c6e35c - the latter commit ends up rotating the method, so super types are inspected ahead of own contract.

(Edit: Wrong, the rotation seems to have happened earlier - need to study in detail)

While with the PR #4228, the performance problem reported here indeed would indeed go away, it may be worth a relook what restoring the original architecture of this method would mean for improving performance for both more general and pathological cases. I think this should/could be addressed in a follow up ticket for 4.38 M1 time frame.

The tweaks in the current PR could be considered for that effort, although I think it is going to look fundamentally very different with fail fast smarts built-in all the way.

srikanth-sankaran:
I have raised https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4230 to be looked at in 4.38 M1 time frame, possibly incorporating elements of the PR here. 

This is all assuming PR https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4228 fixes your current problem of course.

srikanth-sankaran:
An alternate approach to SAM computation is here: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4232 - This passes LET and NLET locally - let us wait for a full test run

You will be excused if you don't recognize a line of code from before 😆

flori-schwa:
I will definitely try out the fail-fast Check (#4228) tomorrow in our productive workspace. 
Maybe the methods could also be grouped by their name in #4232, though I'm not sure if that is necessary after the fail-fast check...

srikanth-sankaran:
> I will definitely try out the fail-fast Check (#4228) tomorrow in our productive workspace. Maybe the methods could also be grouped by their name in #4232, though I'm not sure if that is necessary after the fail-fast check...

I am hoping your performance problems would be addressed by the fail fast mode already released. 

While I have been experimenting with alternate approaches (https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4232)  - I don't want to rush - this code was written long ago and we haven't had complaints so far ...

srikanth-sankaran:
FTR, some benchmark numbers are posted here: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4232#issuecomment-3095075691

@flori-schwa - could you try both (a) the I-Build https://download.eclipse.org/eclipse/downloads/drops4/I20250720-1800/ which has the fail fast fix implemented via https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4228 and (b) the PR https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4232 which is not yet integrated into master please ? Thanks.

flori-schwa:
> FTR, some benchmark numbers are posted here: [#4232 (comment)](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4232#issuecomment-3095075691)
> 
> @flori-schwa - could you try both (a) the I-Build https://download.eclipse.org/eclipse/downloads/drops4/I20250720-1800/ which has the fail fast fix implemented via #4228 and (b) the PR #4232 which is not yet integrated into master please ? Thanks.

@srikanth-sankaran the fail-fast check definitely fixes the Performance Issue, the Code Completion (which previously took a few seconds) finishes in under half a second with the isFunctionalInterface() Check only taking around 9ms:

grafik

Since every interface from our data model always overrides two methods without providing a default implementation, this already causes them all to fail the fast check. Although it is not a usecase, I can - however - define my own interface, which extends an interface from our data model and only defines one new abstract method.

Running Code Completion when an instance of this interface is expected, triggers a "full run" of getInterfaceAbstractContracts(): (Code Completion taking Approximately 2 Seconds)
grafik

The changes from #4232 then fix this Problem: (Code Completion under half a second again)
grafik

srikanth-sankaran:
Thanks very much for the confirmation @flori-schwa - I expect that after multiple reviews #4232 will be released for 4.37 M3.

This was a worthwhile good mini project. It is interesting from academic perspective how with each defect fix the original small(er) implementation grew and grew in complexity and lost the original fail fast strategy somewhere along the way.

Thanks for alerting us about this.

srikanth-sankaran:
I am closing this as the work has shifted to https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4232 - I have just pushed a commit with additional restructuring & improvement provided to that PR by @flori-schwa - Thanks!

PR #4169, opened 2025-07-07
Completion does not offer record components from Javadoc

### LINKED ISSUE #4158: Completion does not offer record components from Javadoc
Originally reported here: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4102#issuecomment-2973441405

```
/**
 * A foo bar.
 * @param f| // <-- this is the cursor
 */
public record FooBar(String foo, String bar) {}

Pressing Ctrl+Space for Content Assist on any function should give me @param foo as an option. 
But for records it doesn't give me any choice at all. Eclipse doesn't understand that the parameters 
to a record's constructor are in the Javadoc for the class.
```

jarthana:
Add missing resolution and completion code for records inside Javadoc

## What it does
Fixes #4158

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Javadoc.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Javadoc.java
@@ -16,6 +16,7 @@
  *******************************************************************************/
 package org.eclipse.jdt.internal.compiler.ast;
 
+import java.util.function.Function;
 import org.eclipse.jdt.core.compiler.CharOperation;
 import org.eclipse.jdt.internal.compiler.ASTVisitor;
 import org.eclipse.jdt.internal.compiler.classfmt.ClassFileConstants;
@@ -241,6 +242,7 @@ public class Javadoc extends ASTNode {
 			JavadocSingleNameReference param = this.paramReferences[i];
 			scope.problemReporter().javadocUnexpectedTag(param.tagSourceStart, param.tagSourceEnd);
 		}
+		resolveParamTags(scope, true);
 		resolveTypeParameterTags(scope, true);
 
 		// @return tags
@@ -542,9 +544,35 @@ public class Javadoc extends ASTNode {
 			scope.problemReporter().javadocInvalidReference(reference.sourceStart, reference.sourceEnd);
 		}
 	}
-
 	/*
-	 * Resolve @param tags while method scope
+	 * Resolve @param tags for records
+	 */
+	private void resolveParamTags(ClassScope scope, boolean reportMissing) {
+		TypeDeclaration typeDecl = scope.referenceContext;
+		if (!typeDecl.isRecord())
+			return;
+		Function<JavadocSingleNameReference, Binding>  resolveNameRef =
+						(nameRef) -> {
+							nameRef.resolve(scope);
+							return nameRef.binding;
+						};
+		Function<AbstractVariableDeclaration, Binding>  resolveArgumentOrComponent =
+						(arg) -> {
+							if (arg instanceof RecordComponent) {
+								return typeDecl.binding.getField(arg.name, false);
+							}
+							return (VariableBinding) arg.getBinding();
+						};
+		resolveParamTags(typeDecl.initializerScope,
+				reportMissing,
+				typeDecl.recordComponents,
+				true,
+				typeDecl.modifiers, // not used
+				resolveNameRef,
+				resolveArgumentOrComponent);
+	}
+	/*
+	 * Resolve @param tags for methods
 	 */
 	private void resolveParamTags(MethodScope scope, boolean reportMissing, boolean considerParamRefAsUsage) {
 		AbstractMethodDeclaration methodDecl = scope.referenceMethod();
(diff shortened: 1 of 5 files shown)


jarthana:
Srikanth, can you glance over the changes please? The fix is in Javadoc and CompletionJavadoc. CompletionJavadoc#missingRecordComponentNames() has been added to compute the missing param tags for completion. The changes in Javadoc ensure that Javadoc name references from @param tags are resolved properly. These bindings are compared with the synthetic bindings later in CompletionJavadoc. I have removed some of the existing code in favor of reusing existing methods.

srikanth-sankaran:
> Srikanth, can you glance over the changes please?

Yes, I will. Targeting M2, will get it to it before end of this week

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/JavadocTestForRecord.java
```diff
@@ -286,5 +283,84 @@ public void test_bug572367() {
 				"		}" },
 				"0");
 	}
-
+	public void testGHIssue4158_1() {
+		if(this.complianceLevel < ClassFileConstants.JDK14) {
```
srikanth-sankaran: 1. if ( please!
2. records got introduced at JDK16 (ignoring previews.) we should check for JDK16

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/JavadocTestForRecord.java
```diff
@@ -286,5 +283,84 @@ public void test_bug572367() {
...
+				"4. ERROR in X.java (at line 6)\n" +
+				"	* @return no return\n" +
+				"	   ^^^^^^\n" +
+				"Javadoc: Unexpected tag\n" +
+				"----------\n",
+				JavacTestOptions.Excuse.EclipseWarningConfiguredAsError);
+	}
+	public void testGHIssue4158_2() {
+		if(this.complianceLevel < ClassFileConstants.JDK14) {
+			return;
```
srikanth-sankaran: JDK16

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/JavadocTestForRecord.java
```diff
@@ -286,5 +283,84 @@ public void test_bug572367() {
...
+		}
+		runNegativeTest(new String[] { "X.java",
+					"""
+					/**
+					 * @param <T> a type param
+					 * @param <S> a type param
```
srikanth-sankaran, line 299: The test is better with one valid and one invalid type parameter

srikanth-sankaran (changes requested):
I have reviewed the tests and have some improvement suggestions.
Lets us walk through the code changes together as planned

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/JavadocTestForRecord.java
jarthana: Unfortunately all tests have similar format for if(. I will have to touch all of them.

srikanth-sankaran: > Unfortunately all tests have similar format for if(. I will have to touch all of them.

No, let us do the right thing for new code :)

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests16_2.java
```diff
@@ -642,4 +642,79 @@ public static void main(String[] args) {
...
+		this.workingCopies = new ICompilationUnit[2];
+		this.workingCopies[0] = getWorkingCopy(
+			"/Completion/src/test/FooBar.java",
+			"""
+				/**
+				 *
+				 * @param <X
+				 */
+				public record X<XYZ>(int abc, int XYZ) {
+				}
```
srikanth-sankaran, line 709: Suggest rename the type parameter to be XTZ and the component to be XPZ in this test and the subsequent one so that it stands out we are offering the right proposals.

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/complete/CompletionJavadoc.java
```diff
@@ -66,7 +68,12 @@ private void internalResolve(Scope scope) {
 				if (resolve) {
 					switch (scope.kind) {
 						case Scope.CLASS_SCOPE:
-							this.completionNode.resolveType((ClassScope)scope);
+							if (scope.referenceContext() instanceof TypeDeclaration type && type.isRecord()) {
+								this.completionNode.resolveType(type.initializerScope);
+							}
+							else {
+								this.completionNode.resolveType((ClassScope)scope);
+							}
```
srikanth-sankaran, line 76: What necessitates this change ? If I back this change, none of the modified tests fail

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/complete/CompletionJavadoc.java
```diff
@@ -196,6 +205,58 @@ public void resolve(MethodScope scope) {
...
+		for (int i = 0; i < componentSize; i++) {
+			RecordComponent arg = components[i];
+			boolean found = false;
+			int paramNameRefCount = 0;
+			for (int j = 0; j < paramTagsSize && !found; j++) {
+				JavadocSingleNameReference param = this.paramReferences[j];
+				VariableBinding field = type.binding.getField(arg.name, false);
+				if (field == param.binding) {
+					if (param.binding == paramNameRefBinding) { // do not count first occurrence of component name reference
+						paramNameRefCount++;
```
srikanth-sankaran, line 241: What exactly is handled here ? Why do we skip the first occurrence ?? Given this is a copy, it is likely doing something valid - but unable to readily imagine what

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Javadoc.java
```diff
@@ -542,9 +544,35 @@ else if (reference instanceof JavadocSingleTypeReference && reference.resolvedTy
...
+		Function<JavadocSingleNameReference, Binding>  resolveNameRef =
+						(nameRef) -> {
+							nameRef.resolve(scope);
+							return nameRef.binding;
+						};
+		Function<AbstractVariableDeclaration, Binding>  resolveArgumentOrComponent =
+						(arg) -> {
+							if (arg instanceof RecordComponent) {
+								return typeDecl.binding.getField(arg.name, false);
+							}
```
srikanth-sankaran: Looks like the instanceof will always succeed ??

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/complete/CompletionJavadoc.java
jarthana: @param bar
@param bar|

We have two name references both with the same name. We are to offer the completion of bar only when there's only one "@param bar"

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/complete/CompletionJavadoc.java
jarthana: Without this, the completion node name reference doesn't get resolved (the result of which is captured in paramNameRefBinding) and as a result the binding comparison `if (param.binding == paramNameRefBinding)` in line 240 will fail. I have added a new test to cover this in CompletionTests16_2.testGHIssue4158_2().

PR #4173, opened 2025-07-08
'var' typed locals are not suggested as auto-completion method arguments

srikanth-sankaran:
* Addresses https://github.com/eclipse-jdt/eclipse.jdt.core/issues/228


srikanth-sankaran:
Oops, I meant to mention @hotspotoq as a co-author but ended up integrating the PR without attribution. Sorry about that

Issue #2347, opened 2024-04-16
Add -proc:full to batch compiler


cpfeiffer:
https://bugs.openjdk.org/browse/JDK-8308245 adds the `-proc:full` option value to javac for versions 8, 11 and 17, and 21. maven-compiler-plugin already [supports this value as well](https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#proc).

It would be nice if JDT's batch compiler would also support this.

See https://github.com/openjdk/jdk/pull/14432 for the discussion around this.

jukzi:
Do you miss the function to annotation processing/compilation or do you just want ecj not to complain about the parameter?

cpfeiffer:
I would expect no functional changes other than accepting the new `full` value, which is supposed to reflect the current behavior wrt annotation processing.

PR #4208, opened 2025-07-16
Problem during restore of ClassFileEditor #2347

mehmet-karaman:
- Added a the new IAsyncClasspathContainer type for asynchronously loaded classpath containers.
- Added a new kind for the failure containers (for beeing able to recognize the container)
- Added a method to check if all of the classpath containers are resolved.
- Added the initialization of the IAsyncClasspathContainers

**The reproduction scenario:**
- Create a plugin project
- Add any plugin as a plugin dependency to the MANIFEST.MF (for example: org.eclipse.jface)
- Navigate to the newly created Project in the Package Explorer
- Expand "Plug-in Dependencies"
- Find the newly introduced depencies entry and open any class file in the ClassFileEditor (it just must be a class inside the Plug-in Dependencies container)
- Restart Eclipse

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding.java
@@ -22,6 +22,7 @@
  *******************************************************************************/
 package org.eclipse.jdt.internal.compiler.lookup;
 
+import java.util.Arrays;
 import org.eclipse.jdt.core.compiler.CharOperation;
 import org.eclipse.jdt.internal.compiler.ast.ASTNode;
 import org.eclipse.jdt.internal.compiler.ast.Wildcard;
@@ -203,6 +204,12 @@ public class CaptureBinding extends TypeVariableBinding {
 			}
 			return;
 		}
+		// JLS 5.1.10
+		//   Let G name a generic type declaration (§8.1.2, §9.1.2) with n type parameters A1,...,An with corresponding bounds U1,...,Un.
+		// our equivalents:
+		//   G = capturedParameterizedType.type
+		//   Ai = wildcardVariable
+		//   Ui = wildcardVariable.superclass & wildcardVariable.superInterfaces()
 		ReferenceBinding originalVariableSuperclass = wildcardVariable.superclass;
 		ReferenceBinding substitutedVariableSuperclass = (ReferenceBinding) Scope.substitute(capturedParameterizedType, originalVariableSuperclass);
 		// prevent cyclic capture: given X<T>, capture(X<? extends T> could yield a circular type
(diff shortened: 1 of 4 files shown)


iloveeclipse:
@mehmet-karaman : please rebase all your branches to latest master state. This PR is based on an old JDT state and will never build.

iloveeclipse:
Please also enable API tools in your workspace and check API errors.
The new class & field need `since` annotations.
Please do **NOT** change JDT core version to 4.0.0 as API tools suggests, instead a filter need to be added.

After rebase on master, apply this patch:

```
diff --git a/org.eclipse.jdt.core/.settings/.api_filters b/org.eclipse.jdt.core/.settings/.api_filters
index f9c56a5..c17576a 100644
--- a/org.eclipse.jdt.core/.settings/.api_filters
+++ b/org.eclipse.jdt.core/.settings/.api_filters
@@ -224,2 +224,10 @@
     </resource>
+    <resource path="model/org/eclipse/jdt/core/IClasspathContainer.java" type="org.eclipse.jdt.core.IClasspathContainer">
+        <filter id="403767336">
+            <message_arguments>
+                <message_argument value="org.eclipse.jdt.core.IClasspathContainer"/>
+                <message_argument value="K_UNINITIALIZED_CONTAINER"/>
+            </message_arguments>
+        </filter>
+    </resource>
     <resource path="model/org/eclipse/jdt/core/SourceRange.java" type="org.eclipse.jdt.core.SourceRange">
diff --git a/org.eclipse.jdt.core/model/org/eclipse/jdt/core/IAsyncClasspathContainer.java b/org.eclipse.jdt.core/model/org/eclipse/jdt/core/IAsyncClasspathContainer.java
index c04a9cd..8c5fe6e 100644
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/core/IAsyncClasspathContainer.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/core/IAsyncClasspathContainer.java
@@ -4,2 +4,3 @@
  * Classpath container implementations which are initialized asynchronously should implement this interface.
+ * @since 3.43
  */
diff --git a/org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java b/org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
index 3759e7c..18982c4 100644
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
@@ -6681,2 +6681,3 @@
 	 * @see #initializeAfterLoad(IProgressMonitor)
+	 * @since 3.43
 	 */
```

org.eclipse.jdt.core/model/org/eclipse/jdt/core/IAsyncClasspathContainer.java
```diff
@@ -0,0 +1,23 @@
+package org.eclipse.jdt.core;
+
+/**
+ * Classpath container implementations which are initialized asynchronously should implement this interface.
+ */
+public interface IAsyncClasspathContainer extends IClasspathContainer {
```
laeubi: I must confess I don't understand the need for this... The PDE container is NOT initialized asynchronously per se, and it is not forbidden to update the Java classpath later on so I really doubt such interface is ever needed.

org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
```diff
@@ -4803,6 +4803,19 @@ public static void initializeAfterLoad(IProgressMonitor monitor) throws CoreExce
...
+
+					if (entry.getEntryKind() == IClasspathEntry.CPE_CONTAINER) {
+						IClasspathContainer container = getClasspathContainer(entry.getPath(), project);
+						if (container == null || container.getKind() == IClasspathContainer.K_UNINITIALIZED_CONTAINER) {
+							continue;
+						}
+						if (container instanceof IAsyncClasspathContainer aContainer) {
+							if (!aContainer.isInitialized()) {
+								// Trigger lazy getClasspathEntries() implementations
+								aContainer.initialize();
```
laeubi: This is completely off the shelf, so if a container is async, you now force it to be in sync again, that absolutely makes no sense. So now PDE must be async again in initialize to protect from bad callers like JDT.

iloveeclipse:
> It is completely unclear why and how this fixes the issue

Probably because Mehmet did not explained that yet? Note, there are 3 repositories involved.

@mehmet-karaman : please provide a high level summary of the fix on https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2245 and also exact steps / preconditions required to reproduce.

merks:
Actually I do see errors again in M1 after the "use jobs" design was reverted:

image

If the thread was delayed until the job manager started that would probably be an easy fix...  But I'm not sure the current state of the reconciler.

iloveeclipse:
> Actually I do see errors again in M1 after the "use jobs" design was reverted:

You probably mean https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2069? Should I reopen that?

laeubi:
The problem can be reproduced with a plain JDT project, no PDE involved:

- Open this project [testclasseditor.zip](https://github.com/user-attachments/files/21261949/testclasseditor.zip)
- Got to the Project Explorer and disable the hiding of output folders
- Open the classfile `Blablabla.class`
- Close Eclipse and delete the classfile
- Open Eclipse and you get

grafik

even if JDT has restored the file in the build.

iloveeclipse:
> The problem can be reproduced with a plain JDT project, no PDE involved:
> 
> * Open this project [testclasseditor.zip](https://github.com/user-attachments/files/21261949/testclasseditor.zip)
> * Got to the Project Explorer and disable the hiding of output folders
> * Open the classfile `Blablabla.class`
> * Close Eclipse and delete the classfile

Well, this shows **similar** symptom of a **different** problem, fully unrelated to the regression introduced in https://github.com/eclipse-pde/eclipse.pde/pull/1552, and also unrelated to https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2245 and this PR. 
And of course only very limited number of users (n==1?) open editors on generated class files and  delete generated class files every time before starting Eclipse. Opposite is the case for the problem we are solving here. Lot of users have opened regular class files from PDE provided libraries and expect that on next restart same editors are restored properly.

laeubi:
I strongly believe you understand the concept of a minimal simple reproducer for a problem.

The same occurs when you open the edit from a classpath jar and that jar is later updated, on debugging sessions, when external tools delete a classfile, renaming types, switching branches and so on ... so if you like more complex ones choose any of those.

Fact is the ClassFileEditor currently does not behave good in case of dynamic changes to the underlying editor input of any kind and that problem exist since a long time. And since a long time users where able to simply close and reopen the classfile in such situations because no one bothered to improve the Editor.
Another fact is that classpath is allowed to change anytime (sooner or later after the startup) so any code should work with that.

So addressing that issue in the classfile editor would solve not only one singular problem but a whole bunch of things but it seems easier to insist it is PDE who broke what already was broken but less visible...

iloveeclipse:
We provide a fix for a [regression](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2245) caused by https://github.com/eclipse-pde/eclipse.pde/pull/1552 and that fix (and regression) affects most often occured use case.

Similar (but not same) symptoms of a broken class editor can be caused by different reasons.

Nowhere in the [ticket](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2245) or fix we claim to fix all and every possible problem that can happen with the class file editor, nor we believe a generic fix for all possible input problems is possible. If you believe there is one, please set a PR and we can review that.

Issue #4209, opened 2025-07-16
ECJ should not warn about missing serialVersionUID for anonymous inner classes


garretwilson:
If the "Serializable class without serialVersionUID" option is enabled, Eclipse provide a warning inside the IDE:

> The serializable class does not declare a static final serialVersionUID field of type long

Most of the time this is correct and desired behavior, as it was requested by the setting.

This setting mirrors the `serial` setting of `javac` with `-Xlint` (see the [Java 21 `javac` docs](https://docs.oracle.com/en/java/javase/21/docs/specs/man/javac.html)):

> `serial`: Warns about the serializable classes that do not provide a serial version ID. Also warns about access to non-public members from a serializable element.

However in [JDK-7152104](https://bugs.openjdk.org/browse/JDK-7152104) `javac` as of Java 8 no longer gives a warning for missing `serialVersionUID` for anonymous inner classes. It would be great if Eclipse followed suite; it is desirable that the Eclipse behavior mirror that of `javac`. It is an added burden to place a `serialVersionUID` on anonymous inner classes just to satisfy Eclipse, and the only alternative to avoid the warning is to turn off the warning altogether in Eclipse, which removes a lot of valid warnings.

The following is a test case:

```java
package com.example;

public class Test {
  void test() {
    new Number() {
      public double doubleValue() {return 0.0;}
      public float floatValue() {return 0.0f;}
      public int intValue() {return 0;}
      public long longValue() {return 0L;}
    };
  }
}
```

Eclipse shows a warning for a missing `serialVersionUID` for the anonymous subclass of `Number`, but `javac` shows no such warning. This was tested with Eclipse EE 2025-06 and Java 24.

Please bring Eclipse in line with `javac` behavior and forgo showing a warning for a missing `serialVersionUID` on anonymous inner classes.

srikanth-sankaran:
I'll follow up.

srikanth-sankaran:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=343858 got closed as stale bug - Grrr.

garretwilson:
Thank you!

garretwilson:
[...]

PR #4210, opened 2025-07-17
ECJ should not warn about missing serialVersionUID for anonymous inner classes

srikanth-sankaran:
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4209


HannesWell:
This has the potential to break the next I-build (at least the verification build of `eclipse.platform.releng.aggregator`) is already broken, because new we have at least one unnecessarily suppressed warning in 

https://github.com/eclipse-jdt/eclipse.jdt.debug/blob/8a448eb1bd4b9f5fd63984628ae727557ccf51d7/org.eclipse.jdt.launching/launching/org/eclipse/jdt/internal/launching/sourcelookup/advanced/FileHashing.java#L131

Issue #4217, opened 2025-07-17
"NPE: Cannot invoke `org.eclipse.jdt.internal.compiler.lookup.BlockScope.getBinding(char[], int, org.eclipse.jdt.internal.compiler.lookup.InvocationSite, boolean)` because `scope` is null" on hyperlink request


nlisker:
Split from https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2286.
```java
import java.util.concurrent.Callable;

public class Test {

	public static <T> void createObjectBinding(final Callable<T> func) {
		return;
	}

	sealed interface Index {
		enum SS implements Index {}
		enum TS implements Index {}
	}
	
	public abstract sealed class Entity<S extends Index> permits Struct, Time {}
	
	final class Struct extends Entity<Index.SS> {}
	
	final class Time extends Entity<Index.TS> {
		
		Struct getStruct() {
			return null;
		}
	}
	
	private void setMaterials(Time entity) {
		ObjectBinding<Entity<?>> selfIllumImage = Bindings.<Entity<?>>createObjectBinding(() -> {
				var entity2 = entity.getStruct() == null ? entity : entity.getStruct();
				return createSlefIlluminationMap(entity.getSpells());
		});
	}
}
```
Hold ctrl and hover over `entity2`:
```
java.lang.NullPointerException: Cannot invoke "org.eclipse.jdt.internal.compiler.lookup.BlockScope.getBinding(char[], int, org.eclipse.jdt.internal.compiler.lookup.InvocationSite, boolean)" because "scope" is null
	at org.eclipse.jdt.internal.compiler.ast.LocalDeclaration.resolve(LocalDeclaration.java:304)
	at org.eclipse.jdt.internal.compiler.ast.LocalDeclaration.resolve(LocalDeclaration.java:258)
	at org.eclipse.jdt.internal.codeassist.select.SelectionOnLocalName.resolve(SelectionOnLocalName.java:32)
	at org.eclipse.jdt.internal.codeassist.SelectionEngine$1Visitor.visit(SelectionEngine.java:1413)
	at org.eclipse.jdt.internal.compiler.ast.LocalDeclaration.traverse(LocalDeclaration.java:488)
	at org.eclipse.jdt.internal.compiler.ast.Block.traverse(Block.java:163)
	at org.eclipse.jdt.internal.compiler.ast.LambdaExpression.traverse(LambdaExpression.java:778)
[...]

PR #4225, opened 2025-07-18
NPE: Cannot invoke org.eclipse.jdt.internal.compiler.lookup.BlockScope.getBinding() because scope is null" on hyperlink request

josediazfer:
…e.getBinding() because scope is null" on hyperlink request (#4217)

* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4217

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ResolveTests10.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ResolveTests10.java
@@ -178,4 +178,49 @@ public class ResolveTests10 extends AbstractJavaModelTests {
 				selected
 			);
 	}
+
+	// https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4217
+	// NPE: Cannot invoke org.eclipse.jdt.internal.compiler.lookup.BlockScope.getBinding(char[], int, org.eclipse.jdt.internal.compiler.lookup.InvocationSite, boolean) because scope is null" on hyperlink request 
+	public void testIssue4217() throws CoreException {
+		this.wc = getWorkingCopy("/Resolve/src/X.java",
+		"""
+		import java.util.concurrent.Callable;
+
+		public class Test {
+
+			public static <T> void createObjectBinding(final Callable<T> func) {
+				return;
+			}
+
+			sealed interface Index {
+				enum SS implements Index {}
+				enum TS implements Index {}
+			}
+
+			public abstract sealed class Entity<S extends Index> permits Struct, Time {}
+
+			final class Struct extends Entity<Index.SS> {}
+
+			final class Time extends Entity<Index.TS> {
+				Struct getStruct() {
+					return null;
+				}
+			}
+
+			private void setMaterials(Time entity) {
+				ObjectBinding<Entity<?>> selfIllumImage = Bindings.<Entity<?>>createObjectBinding(() -> {
+					var entity2/*here*/ = entity.getStruct() == null ? entity : entity.getStruct();
+					return createSlefIlluminationMap(entity.getSpells());
+				});
+			}
+		}
+		""");
+		String str = this.wc.getSource();
+		String selection = "entity2/*here*/";
+		int start = str.lastIndexOf(selection);
+		int length = selection.length();
+
+		IJavaElement[] selected = this.wc.codeSelect(start, length);
+		assertEquals(0, selected.length);
+	}
 }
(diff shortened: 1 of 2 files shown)


srikanth-sankaran:
Have you found out why scope is null there ? Slightly wary of null checks as solution unless fuller analysis justifies it.

josediazfer:
@srikanth-sankaran 

A smaller example of the problem:
```java
public void test() {
	ArrayList<String> testArray = new ArrayList<>();
	
	testArray.testMethod(e -> {
		int valueVar = 1;
	});
}
```

The problem occurs when the "valueVar" variable  is selected within the lambda passed as a parameter to the "testMethod" method.The "testMethod" method not be resolved because it does not exist in the ArrayList class.

The scope is null because it hasn't been created in the resolveType method of the LambdaExpression class. Specifically, it is created when calling resolveType of the LambdaExpression class from the resolvePolyExpressionArguments0 method of the ASTNode class but since the method "testMethod" does not exist ("**candidateMethod**" is null), the resolveType of LambaExpression is not called from the resolvePolyExpressionArguments0 method. 

I put a piece of source code of ASTNode.resolvePolyExpressionArguments0:
```java
	private static MethodBinding resolvePolyExpressionArguments0(Invocation invocation, MethodBinding method, TypeBinding[] argumentTypes, BlockScope scope) {
		MethodBinding candidateMethod = method.isValidBinding() ? method : method instanceof ProblemMethodBinding ? ((ProblemMethodBinding) method).closestMatch : null;
		if (candidateMethod == null)
			return method;
			.....
		if (argumentTypes[i] != null && argumentTypes[i].isPolyType()) {
			argument.setExpectedType(parameterType);
			TypeBinding updatedArgumentType;
			if (argument instanceof LambdaExpression) {
				LambdaExpression lambda = (LambdaExpression) argument;
				// avoid complaining about non-kosher descriptor as secondary problem
				boolean skipKosherCheck = method.problemId() == ProblemReasons.Ambiguous;
				updatedArgumentType = lambda.resolveType(scope, skipKosherCheck);
```

If it were an existing method, for example, the forEach method of testArray, lamba.resolveType() would be called from resolvePolyExpressionArguments0.
Then, the lamba scoped method would be created within the resolveType method of the LambaExpression class, and therefore the type of the "valueVar" variable would be resolved, generating the SelectionNodeFound exception with its type.

In the above example, the node tree associated with the "test" method will be traversed, and when the selected node "valueVar" is visited, an attempt will be made to resolve its type, but since the scope of the lambda is not initialized, it will fail with a NullPointerException

stephan-herrmann:
> ... generating the SelectionNodeFound exception ...

should we assume that the problem occurs only during selection, not during regular compilation?

josediazfer:
@stephan-herrmann Yes, exactly when you hover over the variable name, the problem does not appear during compilation.

stephan-herrmann:
The problem vaguely reminds me of https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1077/commits/991c1db1f714c0dc06f6fc658b506d0c0bb67e7b

The general topic would be: completion and selection similarly abort compilation by throwing {Completion,Selection}NodeFound. If this exception will leave AST & bindings in an incomplete / inconsistent state, it may be better to stash the exception, continue regular compilation, and only pull out the exception when all is properly resolved.

Just mentioning.

josediazfer:
@stephan-herrmann 

I fully agree with your proposal to unify the same functionality for both selection and completion, but I don't think it's useful for this specific case. Both CompletionEngine and SelectionEngine don't generate any type of (Selection/Completion)NodeFound exception after calling resolve(). The difference between SelectionEngine and Completion is that after calling resolve() in SelectionEngine, the AST is traversed, causing a NullPointerException. Another possible fix, but more dangerous, would be to remove the following lines:
```
							if (node != null) {
								selectLocalDeclaration(node);
							}
```

I'm not sure if it still makes sense to traverse the AST if a SelectionNodeFound exception wasn't previously generated. In CompletionEngine, nothing similar is done after calling resolve().
What do you think is better?

srikanth-sankaran:
See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4217#issuecomment-3883663113. The current PR is not suitable. It is not just a matter of a null check. I'll propose an alternate PR.

srikanth-sankaran:
I have raised PR https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4836 - This passes `RunModelTests` locally.

@josediazfer - I want to acknowledge for the record that much of your analysis is spot on (great job). I have opted for a more general fix that avoids descending into an unresolved lambda as it is potentially full of landmines. The hover behavior won't magically start working, but we won't trigger a runtime error and that is as good as it gets: The AST nodes in the subtrees are completely undecorated and so are unsuitable for visitation. In a broken program, where no name resolution hasn't happened in significant regions of code, not much more can be expected.

Issue #4230, opened 2025-07-20
[Performance] ReferenceBinding.getInterfaceAbstractContracts works too hard, can use some smarts


srikanth-sankaran:
Problem came to our radar with `https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4168`

https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4228 has been integrated into master and addresses the reported problem by providing a fail fast mode.

However, `ReferenceBinding.getInterfaceAbstractContracts` can be reworked to be smart for both the general and pathological scenarios.

See https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4168#issuecomment-3092999680 for some background.

We should also look at implementing elements of the fix in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4168 in the new work if they would fit in.

PR #4232, opened 2025-07-20
More efficient, more modern, fast failing version of computation of the abstract contracts of an interface while deriving SAM 

srikanth-sankaran:
PR https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4168 raised alert about serious performance bottlenecks during code assist that under the hood attempt to check if a given interface is a functional interface.

The fail fast hacks in the original implementation seem to have withered away with various incremental changes to `getInterfaceAbstractContracts()` and the version we have now seems to work very hard without being smart.

https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4228 implements a very fast failing check that would address _many but not all_ common scenarios. This PR is a follow up to clean up and speed up the implementation of  `getInterfaceAbstractContracts()` for the general case.

**Speed up is achieved by:**

    *  Not visiting an interface more than once
    *  While checking possible overrides, greatly reduce the loop iterations by first soring on selector and checking overrides only between methods with identical selector
    *  Failing fast - if we have seen two distinct abstract methods already, just punt right away.
 
**Cleanup is achieved by**

    * Use of streams and collectors in lieu of low level array management ops.
    * Clarifying intent by opting for method names that clarify that this code is specific to sam computation as opposed to a general purpose hammer (`getInterfaceAbstractContracts()` split and renamed to `getFunctionalInterfaceAbstractContracts()` and `collateFunctionalInterfaceContracts()`) 

**No correctness issue is addressed here, only performance improvement is attempted**

* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4230

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/IntersectionTypeBinding18.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/IntersectionTypeBinding18.java
@@ -24,6 +24,7 @@
 package org.eclipse.jdt.internal.compiler.lookup;
 
 import java.util.Set;
+import java.util.stream.Stream;
 import org.eclipse.jdt.internal.compiler.ast.Wildcard;
 import org.eclipse.jdt.internal.compiler.classfmt.ClassFileConstants;
 
@@ -67,22 +68,10 @@ public class IntersectionTypeBinding18 extends ReferenceBinding {
 	}
 
 	@Override
-	protected MethodBinding[] getInterfaceAbstractContracts(Scope scope, boolean replaceWildcards, boolean filterDefaultMethods) throws InvalidBindingException {
-		int typesLength = this.intersectingTypes.length;
-		MethodBinding[][] methods = new MethodBinding[typesLength][];
-		int contractsLength = 0;
-		for (int i = 0; i < typesLength; i++) {
-			methods[i] = this.intersectingTypes[i].getInterfaceAbstractContracts(scope, replaceWildcards, true);
-			contractsLength += methods[i].length;
-		}
-		MethodBinding[] contracts = new MethodBinding[contractsLength];
-		int idx = 0;
-		for (int i = 0; i < typesLength; i++) {
-			int len = methods[i].length;
-			System.arraycopy(methods[i], 0, contracts, idx, len);
-			idx += len;
-		}
-		return contracts;
+	protected Stream<MethodBinding> collateFunctionalInterfaceContracts(Scope scope, boolean replaceWildcards, Set<MethodBinding> contracts) throws DysfunctionalInterfaceException {
+		for (ReferenceBinding intersectingType : this.intersectingTypes)
+			intersectingType.collateFunctionalInterfaceContracts(scope, replaceWildcards, contracts);
+		return contracts.stream();
 	}
 
 	@Override
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/IntersectionTypeBinding18.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/IntersectionTypeBinding18.java
@@ -23,6 +23,7 @@
 
 package org.eclipse.jdt.internal.compiler.lookup;
 
+import java.util.Arrays;
 import java.util.Set;
 import java.util.stream.Stream;
 import org.eclipse.jdt.internal.compiler.ast.Wildcard;
@@ -68,10 +69,10 @@ public class IntersectionTypeBinding18 extends ReferenceBinding {
 	}
 
 	@Override
-	protected Stream<MethodBinding> collateFunctionalInterfaceContracts(Scope scope, boolean replaceWildcards, Set<MethodBinding> contracts) throws DysfunctionalInterfaceException {
-		for (ReferenceBinding intersectingType : this.intersectingTypes)
-			intersectingType.collateFunctionalInterfaceContracts(scope, replaceWildcards, contracts);
-		return contracts.stream();
+	protected Stream<MethodBinding> collateFunctionalInterfaceContracts(Scope scope, boolean replaceWildcards) throws DysfunctionalInterfaceException {
+		return Arrays.stream(this.intersectingTypes).flatMap( //
+						intersectingType -> intersectingType.collateFunctionalInterfaceContracts(scope, replaceWildcards) //
+		);
 	}
 
 	@Override
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/IntersectionTypeBinding18.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/IntersectionTypeBinding18.java
@@ -69,9 +69,9 @@ public class IntersectionTypeBinding18 extends ReferenceBinding {
 	}
 
 	@Override
-	protected Stream<MethodBinding> collateFunctionalInterfaceContracts(Scope scope, boolean replaceWildcards) throws DysfunctionalInterfaceException {
+	protected Stream<MethodBinding> collateFunctionalInterfaceContracts(Scope scope, boolean replaceWildcards, Set<ReferenceBinding> visitedInterfaces) throws DysfunctionalInterfaceException {
 		return Arrays.stream(this.intersectingTypes).flatMap( //
-						intersectingType -> intersectingType.collateFunctionalInterfaceContracts(scope, replaceWildcards) //
+						intersectingType -> intersectingType.collateFunctionalInterfaceContracts(scope, replaceWildcards, visitedInterfaces) //
 		);
 	}
 
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
@@ -57,12 +57,10 @@ import java.util.Arrays;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.Predicate;
-import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.eclipse.jdt.core.compiler.CharOperation;
 import org.eclipse.jdt.internal.compiler.ast.ASTNode;
@@ -2334,8 +2332,9 @@ private MethodBinding[] getFunctionalInterfaceAbstractContracts(Scope scope, boo
 	boolean isAnnotationBasedNullAnalysisEnabled = environment.globalOptions.isAnnotationBasedNullAnalysisEnabled;
 	MethodBinding samCandidate = null;
 
-	MethodBinding[] methods = collateFunctionalInterfaceContracts(scope, replaceWildcards, new LinkedHashSet<>())
+	MethodBinding[] methods = collateFunctionalInterfaceContracts(scope, replaceWildcards)
 		.sorted((m1, m2) -> CharOperation.compareTo(m1.selector, m2.selector))
+		.distinct()
 		.toArray(MethodBinding []::new);
 
 	for (int i = 0, length = methods.length; i < length; i++) {
@@ -2361,23 +2360,23 @@ private MethodBinding[] getFunctionalInterfaceAbstractContracts(Scope scope, boo
 	return Arrays.stream(methods).filter(m -> m != null && m.isAbstract()).toArray(MethodBinding []::new);
 }
 
-protected Stream<MethodBinding> collateFunctionalInterfaceContracts(Scope scope, boolean replaceWildcards, Set<MethodBinding> contracts) throws DysfunctionalInterfaceException {
+protected Stream<MethodBinding> collateFunctionalInterfaceContracts(Scope scope, boolean replaceWildcards) throws DysfunctionalInterfaceException {
 
 	if (!isInterface() || !isValidBinding())
 		throw DYSFUNCTIONAL_INTERFACE_EXCEPTION;
 
-	for (ReferenceBinding superInterface : superInterfaces())
-		superInterface.collateFunctionalInterfaceContracts(scope, replaceWildcards, contracts);
-
 	Predicate<MethodBinding> isContractual = m -> { // select valid public abstract || default methods
-									if (m == null || m.isStatic() || m.redeclaresPublicObjectMethod(scope) || m.isPrivate())
-										return false;
-									if (!m.isValidBinding())
-										throw DYSFUNCTIONAL_INTERFACE_EXCEPTION;
-									return true;
-							};
-	contracts.addAll(Arrays.stream(methods()).filter(isContractual).collect(Collectors.toList()));
-	return contracts.stream();
+		if (m == null || m.isStatic() || m.redeclaresPublicObjectMethod(scope) || m.isPrivate())
+			return false;
+		if (!m.isValidBinding())
+			throw DYSFUNCTIONAL_INTERFACE_EXCEPTION;
+		return true;
+	};
+
+	return Stream.concat( //
+			Arrays.stream(superInterfaces()).flatMap(superInterface -> superInterface.collateFunctionalInterfaceContracts(scope, replaceWildcards)), //
+			Arrays.stream(methods()).filter(isContractual)
+	);
 }
 
 @Override
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
@@ -2341,10 +2341,10 @@ private MethodBinding[] getFunctionalInterfaceAbstractContracts(Scope scope, boo
 		for (int j = i + 1; j < length; j++) {
 			MethodBinding otherMethod = methods[j];
 			if (method.selector.length != otherMethod.selector.length || !CharOperation.equals(method.selector, otherMethod.selector))
-				break; // Skip comparing apple to oranges.
+				break; // Skip comparing apple to oranges (input sorted by selector, so no viable overrides past this point)
 			if (MethodVerifier.doesMethodOverride(otherMethod, method, environment)) {
 				methods[i] = method = null;
-				break; // Skip comparing rotten apple with remaining lot.
+				break; // Skip comparing rotten apple with remaining lot: is overridden already, no use checking if is overridden also by another
 			}
 		}
 		if (method != null && method.isAbstract()) {
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
@@ -2332,9 +2332,8 @@ private MethodBinding[] getFunctionalInterfaceAbstractContracts(Scope scope, boo
 	boolean isAnnotationBasedNullAnalysisEnabled = environment.globalOptions.isAnnotationBasedNullAnalysisEnabled;
 	MethodBinding samCandidate = null;
 
-	MethodBinding[] methods = collateFunctionalInterfaceContracts(scope, replaceWildcards)
+	MethodBinding[] methods = collateFunctionalInterfaceContracts(scope, replaceWildcards, new HashSet<>())
 		.sorted((m1, m2) -> CharOperation.compareTo(m1.selector, m2.selector))
-		.distinct()
 		.toArray(MethodBinding []::new);
 
 	for (int i = 0, length = methods.length; i < length; i++) {
@@ -2360,7 +2359,7 @@ private MethodBinding[] getFunctionalInterfaceAbstractContracts(Scope scope, boo
 	return Arrays.stream(methods).filter(m -> m != null && m.isAbstract()).toArray(MethodBinding []::new);
 }
 
-protected Stream<MethodBinding> collateFunctionalInterfaceContracts(Scope scope, boolean replaceWildcards) throws DysfunctionalInterfaceException {
+protected Stream<MethodBinding> collateFunctionalInterfaceContracts(Scope scope, boolean replaceWildcards, Set<ReferenceBinding> visitedInterfaces) throws DysfunctionalInterfaceException {
 
 	if (!isInterface() || !isValidBinding())
 		throw DYSFUNCTIONAL_INTERFACE_EXCEPTION;
@@ -2374,7 +2373,7 @@ protected Stream<MethodBinding> collateFunctionalInterfaceContracts(Scope scope,
 	};
 
 	return Stream.concat( //
-			Arrays.stream(superInterfaces()).flatMap(superInterface -> superInterface.collateFunctionalInterfaceContracts(scope, replaceWildcards)), //
+			Arrays.stream(superInterfaces()).filter(iface->visitedInterfaces.add(iface)).flatMap(superInterface -> superInterface.collateFunctionalInterfaceContracts(scope, replaceWildcards, visitedInterfaces)), //
 			Arrays.stream(methods()).filter(isContractual)
 	);
 }
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
@@ -57,9 +57,13 @@ import java.util.Arrays;
 import java.util.Comparator;
 import java.util.HashMap;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.eclipse.jdt.core.compiler.CharOperation;
 import org.eclipse.jdt.internal.compiler.ast.ASTNode;
 import org.eclipse.jdt.internal.compiler.ast.LambdaExpression;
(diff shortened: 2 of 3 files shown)


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
```diff
@@ -2322,93 +2325,51 @@ public void detectWrapperResource() {
...
-			contractsCount--;
-			// remove default method after it has eliminated any matching abstract methods from contracts
-			if (i < contractsCount) {
-				System.arraycopy(contracts, i+1, contracts, i, contractsCount - i);
+	Iterator<MethodBinding> iterator = contracts.iterator();
+outer:
+	while (iterator.hasNext()) {
+		MethodBinding method = iterator.next();
+		for (MethodBinding otherMethod : contracts) {
+			if (MethodVerifier.doesMethodOverride(otherMethod, method, environment)) {
```
szarnekow, line 2345: Would you think that grouping/partitioning the methods by name and arity would help here, too? 
It'd bring down the complexity of this impl from n-squared with n=number of methods to something m*(max(inheritance-depth with overrides) squared) where m is the number of partitions.

srikanth-sankaran: > Would you think that grouping/partitioning the methods by name and arity would help here, too? It'd bring down the complexity of this impl from n-squared with n=number of methods to something m*(max(inheritance-depth with overrides) squared) where m is the number of partitions.

I would wait for real world data to drive that decision. More than loop count, it is what transpires inside the loop that matters. 

The program below runs in about 40ms even though it is a 10000 x 10000 loop.

```
public class X {
		long start = System.currentTimeMillis();
		int k = 0;
		for (int i = 0; i < 10000; i++) {
			for (int j = 0; j < 10000; j++) {
				if ((i % 1000) == 0 && i == j) {
					k++;
				}
			}
		}
		long stop = System.currentTimeMillis();
		System.out.println("k = " + k);
		System.out.println("Duration " + (stop - start));
	}
}
```

In this case the nested loop is tiny:

```
    if (MethodVerifier.doesMethodOverride(otherMethod, method, environment)) {
        iterator.remove();
	continue outer;
    }
```

`doesMethodOverride` would fail fast in most cases. In any case I have now included a few additional checks to avoid the override check.

srikanth-sankaran:
The latest evolved and squashed consolidated patch incorporates the idea (though not code) from https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4168 that we need override checks only within the same named groups of methods.

It implements further fail fast checks and is **125-150 times faster** than master implementation (with the fail fast approach in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4228 **disabled** for experimental purposes)

The PR https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4168 is about **20-40 times faster** than master implementation (with the fail fast approach in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4228 **disabled** for experimental purposes)

Just master (with the fail fast approach in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4228 **INTACT**) is **40000 times (sic) faster** than **without** https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4228

srikanth-sankaran:
@stephan-herrmann @jarthana @mpalat and @jespersm - I have marked you as reviewers, hoping to get as many independent eyes to look at this as possible - given how fundamental a change this is.

It is revealing how much of a ball and chain (a) JDT's original restrictions around using only 1.4 code for development and (b) its insistence on pervasive use of arrays - are! I feel the new code is much more modern and is so much simpler to understand. 

A reviewer may take a diff based review approach or choose to review the new implementation by itself given how drastically things are changed. For the latter approach, I call out the methods of interest.

    - org.eclipse.jdt.internal.compiler.lookup.ReferenceBinding.getFunctionalInterfaceAbstractContracts
    - org.eclipse.jdt.internal.compiler.lookup.ReferenceBinding.collateFunctionalInterfaceContracts
    - org.eclipse.jdt.internal.compiler.lookup.ReferenceBinding.isPatentlyDysfunctional
    - org.eclipse.jdt.internal.compiler.lookup.IntersectionTypeBinding18.collateFunctionalInterfaceContracts
    - org.eclipse.jdt.internal.compiler.lookup.ParameterizedTypeBinding.collateFunctionalInterfaceContracts

Thank you.

srikanth-sankaran:
@flori-schwa - you are welcome to review this too, I couldn't officially mark you as reviewer.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
srikanth-sankaran: > Would you think that grouping/partitioning the methods by name and arity would help here, too?

@szarnekow - please study the latest PR which achieves this but by a different mechanism. Thanks

srikanth-sankaran:
Also @stephan-herrmann - I think the call to `ImplicitNullAnnotationVerifier.ensureNullnessIsKnown(method, scope);` introduced by commit 14c16cd0fb702e4d54250d98c2947d1411fbb696 for https://bugs.eclipse.org/bugs/show_bug.cgi?id=514091 looks rather odd in this place! Does it really belong here ??

mpalat:
> @stephan-herrmann @jarthana @mpalat and @jespersm - I have marked you as reviewers, hoping to get as many independent eyes to look at this as possible - given how fundamental a change this is.
> 

@srikanth-sankaran -First, I am yet to delve into changes. However, checked with Co-pilot for fun - summary may be too nascent for you, but still here it is:

-----Start 1 -----

The changes you posted **do not provide exactly the same output** as before. Here’s why:

### Key Functional Changes

1. **API Change:**  
   - **Old:** Used `getInterfaceAbstractContracts` returning a `MethodBinding[]`.
   - **New:** Uses `collateFunctionalInterfaceContracts` returning a `Stream<MethodBinding>` and a new internal helper `getFunctionalInterfaceAbstractContracts`.
   - **Effect:** The new API is stream-based, which changes how methods are collected, filtered, and possibly presented.

2. **Exception Handling:**  
   - **Old:** Threw and caught `InvalidBindingException`.
   - **New:** Throws and catches `DysfunctionalInterfaceException` (now a `RuntimeException`).
   - **Effect:** The type of exceptions and their handling is changed, which may affect stack traces or error propagation.

3. **Collection and Filtering Logic:**  
   - **Old:** Used arrays and manual loop/filter logic.
   - **New:** Uses Java Streams, sets, and predicates for collection and filtering.
   - **Effect:** The filtering and ordering may differ slightly, especially regarding method duplicates, overrides, or ordering (now sorted by selector).

4. **Default Methods Filtering:**  
   - **Old:** Had a `filterDefaultMethods` parameter that could omit default methods.
   - **New:** This parameter is removed; the filtering logic is now encapsulated in the stream/predicate.
   - **Effect:** The behavior regarding default methods may change (e.g., whether or not default methods are included).

### **Practical Output Differences**

- **Order of Results:**  
  Streams are sorted by selector, which may change the order of methods returned compared to the previous implementation.
- **Inclusion/Exclusion of Methods:**  
  The filtering predicate may behave differently, especially for default/static/private methods.
- **Error Reporting:**  
  The exception type and handling may change what is reported if the interface is not functional.

### **Summary Table**

| Aspect                 | Before                                  | After                                      |
|------------------------|-----------------------------------------|--------------------------------------------|
| Return Type            | `MethodBinding[]`                       | `Stream<MethodBinding>`/`MethodBinding[]`  |
| Filtering Logic        | Manual, array-based                     | Stream-based, predicate                    |
| Exception              | `InvalidBindingException` (checked)     | `DysfunctionalInterfaceException` (runtime)|
| Default Methods        | Explicit filter param                   | Predicate inside stream                    |
| Ordering               | As inserted                             | Sorted by selector                         |

---

**In summary:**  
**No, the output is not guaranteed to be the same.**  
If you need exact compatibility, especially for downstream consumers, you’ll need to review how results are consumed and possibly adapt them. The new implementation is likely cleaner and more modern, but it changes details that can affect output and error handling. If you want to discuss a specific case or test, let me know!

-----End 1 -----

I had asked further questions as well- new job opportunities in prompt engineering :). Let me know if the above if the above added any value. Will take a look actually meanwhile.

srikanth-sankaran:
> However, checked with Co-pilot for fun - summary may be too nascent for you, but still here it is:

:) You can probe those observations further during your review and see if any of them lead to a defect. 

Some points to bear in mind: 

1. A default method **declared** in an interface I may override an inherited abstract method

2. Only public abstract and default are collated. 

3. It is guaranteed by the visitation that a super type's methods would always precede a subtypes methods 

4. Because we use a set, duplicates are not possible.

5. Because we use a linked hash set, insertion order is preserved during sorting and streaming.

6. While a given interface may be encountered many times during a super hierarchy walk, it is never the case that such an interface can have different parameterizations - which is illegal.  

With these points as the backdrop you can investigate co-pilot's observations to see if there could be a defect.

srikanth-sankaran:
> collateFunctionalInterfaceContracts returning a Stream<MethodBinding> and a new internal helper getFunctionalInterfaceAbstractContracts.

No. Co-pilot is confusing who is helping who. It is the latter that calls the former.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/IntersectionTypeBinding18.java
```diff
@@ -67,22 +68,10 @@ public TypeBinding clone(TypeBinding enclosingType) {
...
-		}
-		MethodBinding[] contracts = new MethodBinding[contractsLength];
-		int idx = 0;
-		for (int i = 0; i < typesLength; i++) {
-			int len = methods[i].length;
-			System.arraycopy(methods[i], 0, contracts, idx, len);
-			idx += len;
-		}
-		return contracts;
+	protected Stream<MethodBinding> collateFunctionalInterfaceContracts(Scope scope, boolean replaceWildcards, Set<MethodBinding> contracts) throws DysfunctionalInterfaceException {
```
flori-schwa: I'm not sure what exactly the purpose of the "contracts" Set is. 
IMO it could be omitted, using Stream.concat() and flatMap() in the implementations.
If there is a way for duplicate MethodBindings to end up in the Stream, we can also use .distinct() in getFunctionalInterfaceAbstractContracts().
I'll try it out locally and let the Compiler Regression Tests run

flori-schwa: Seems to work, here's the patch: [NoSetParameter.patch](https://github.com/user-attachments/files/21344531/NoSetParameter.patch)

srikanth-sankaran: > Seems to work, here's the patch: [NoSetParameter.patch](https://github.com/user-attachments/files/21344531/NoSetParameter.patch)

At a first glance looks ok, I will study this in detail tomorrow and also run some numbers. Yes, duplicates are very much possible and in fact there are plenty of duplicates in your example (same interface gets visited multiple times in the walk - we could of course use a different approach to ensure methods from an interface are collated only once but I used a set to weed out duplicates. And it was a Linked Hash set to ensure insertion order is preserved in streaming and sorting. The nested loop over arrays on the collated contracts **relies** on methods from super types preceding methods from subtypes for failing fast). 

I need to convince myself this invariant is maintained in your patch - looks to be so.

 And I suspect we should call distinct() before sorted() for better performance - thoughts ?

srikanth-sankaran: I think if we pass around a set of visited interfaces and use that to prevent multiple visits to the same interface then the distinct() operation would be unnecessary and also it would speed up overall work. Why collate duplicates in the first place and then weed them out finally ??

flori-schwa: Passing a set of visited interfaces sounds good 👍

srikanth-sankaran:
21462 - 1462 

> The latest evolved and squashed consolidated patch incorporates the idea (though not code) from #4168 that we need override checks only within the same named groups of methods.
> 
> It implements further fail fast checks and is **125-150 times faster** than master implementation (with the fail fast approach in #4228 **disabled** for experimental purposes)
> 
> The PR #4168 is about **20-40 times faster** than master implementation (with the fail fast approach in #4228 **disabled** for experimental purposes)
> 
> Just master (with the fail fast approach in #4228 **INTACT**) is **40000 times (sic) faster** than **without** #4228

The patch has been tuned to death - as of now it is 750 times faster than when we started - I am ready to call halt on the work and target it for M3 assuming the tests after the latest updates are green and reviews are good to go

iloveeclipse:
> The patch has been tuned to death - as of now it is 750 times faster than when we started

image

Cool. Less code, more fun...

stephan-herrmann:
> @stephan-herrmann @jarthana @mpalat and @jespersm - I have marked you as reviewers, hoping to get as many independent eyes to look at this as possible - given how fundamental a change this is.

Could you help me understand whether this is intended as pure performance improvement, or did you fix any semantic bug on the way? "Alternative computation" isn't very explicit :)

stephan-herrmann:
> Also @stephan-herrmann - I think the call to `ImplicitNullAnnotationVerifier.ensureNullnessIsKnown(method, scope);` introduced by commit [14c16cd](https://github.com/eclipse-jdt/eclipse.jdt.core/commit/14c16cd0fb702e4d54250d98c2947d1411fbb696) for https://bugs.eclipse.org/bugs/show_bug.cgi?id=514091 looks rather odd in this place! Does it really belong here ??

In master this is necessary for NullTypeAnnotationTest to pass.  Running (on master) NullTypeAnnotationTest.testBug514091() needs that call when `getInterfaceAbstractContracts()` is indirectly called on behalf of `LE.resolveType()`, where we need to ensure that `@NonNullByDefault` has been applied to the SAM before invoking `LE.mergeParameterNullAnnotations()`. Omitting the call in question produces a bogus compile error.

What exactly do you regard as odd?

srikanth-sankaran:
> Could you help me understand whether this is intended as pure performance improvement, or did you fix any semantic bug on the way? "Alternative computation" isn't very explicit :)

Sorry, my bad. This was an "alternative" for https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4168 which started this thread, which furnishes the context, but I have updated the top level topic comment with details.

srikanth-sankaran:
> > Also @stephan-herrmann - I think the call to `ImplicitNullAnnotationVerifier.ensureNullnessIsKnown(method, scope);` introduced by commit [14c16cd](https://github.com/eclipse-jdt/eclipse.jdt.core/commit/14c16cd0fb702e4d54250d98c2947d1411fbb696) for https://bugs.eclipse.org/bugs/show_bug.cgi?id=514091 looks rather odd in this place! Does it really belong here ??
> 
> In master this is necessary for NullTypeAnnotationTest to pass. Running (on master) NullTypeAnnotationTest.testBug514091() needs that call when `getInterfaceAbstractContracts()` is indirectly called on behalf of `LE.resolveType()`, where we need to ensure that `@NonNullByDefault` has been applied to the SAM before invoking `LE.mergeParameterNullAnnotations()`. Omitting the call in question produces a bogus compile error.
> 
> What exactly do you regard as odd?

At first blush, this looks like a side effect stapled to a function intended for a different purpose - I don't know and I could be wrong. I saw that it is needed for NullTypeAnnotationTest  to pass and I have retained this in this PR too. I am ok with it being where it is - if you are ok with it.

stephan-herrmann:
> > > Also @stephan-herrmann - I think the call to `ImplicitNullAnnotationVerifier.ensureNullnessIsKnown(method, scope);` introduced by commit [14c16cd](https://github.com/eclipse-jdt/eclipse.jdt.core/commit/14c16cd0fb702e4d54250d98c2947d1411fbb696) for https://bugs.eclipse.org/bugs/show_bug.cgi?id=514091 looks rather odd in this place! Does it really belong here ??
> > 
> > 
> > In master this is necessary for NullTypeAnnotationTest to pass. Running (on master) NullTypeAnnotationTest.testBug514091() needs that call when `getInterfaceAbstractContracts()` is indirectly called on behalf of `LE.resolveType()`, where we need to ensure that `@NonNullByDefault` has been applied to the SAM before invoking `LE.mergeParameterNullAnnotations()`. Omitting the call in question produces a bogus compile error.
> > What exactly do you regard as odd?
> 
> At first blush, this looks like a side effect stapled to a function intended for a different purpose - I don't know and I could be wrong. I saw that it is needed for NullTypeAnnotationTest to pass and I have retained this in this PR too. I am ok with it being where it is - if you are ok with it.

In essence, every method that feeds MethodBindings into further processing, where null analysis may be relevant, must ensure, that any `@NonNullByDefault` has been applied where applicable. 

Theoretically, this could perhaps be pushed upstream to `SourceTypeBinding.methods()`, but I could imagine, that in some situations this might trigger resolving of annotations before we are ready to do so. Perhaps that's why 14c16cd0f placed it closer to the consumer who needs the fully annotated method binding.

stephan-herrmann:
> The patch has been tuned to death - as of now it is 750 times faster than when we started

Could you expand a little for which scenario and under which conditions your improvements have been observed?

My typical question here is: if the huge performance gains have been observed on particularly complex / large input, what is the impact on simpler use cases? E.g. if problems occurred with a single class with a large supertype hierarchy, how does the change impact compilation of gazillions of classes with empty or simple supertype hierarchy? I.e., is there a measurable added effort (e.g., allocating and maintaining `visitedInterfaces`) that could become significant if large input comes in a different shape?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
```diff
@@ -2322,93 +2326,58 @@ public void detectWrapperResource() {
...
+	MethodBinding[] methods = collateFunctionalInterfaceContracts(scope, replaceWildcards, new HashSet<>())
+		.sorted((m1, m2) -> CharOperation.compareTo(m1.selector, m2.selector))
+		.toArray(MethodBinding []::new);
+
+	for (int i = 0, length = methods.length; i < length; i++) {
+		MethodBinding method = methods[i];
+		for (int j = i + 1; j < length; j++) {
+			MethodBinding otherMethod = methods[j];
+			if (method.selector.length != otherMethod.selector.length || !CharOperation.equals(method.selector, otherMethod.selector))
+				break; // Skip comparing apple to oranges.
```
stephan-herrmann: You have two kinds of "skipping": here you keep the orange, whereas below the rotten apple is removed from `methods`. Perhaps a one-line comment about the purpose of this loop might clarify?

srikanth-sankaran:
> Could you expand a little for which scenario and under which conditions your improvements have been observed?

This is on the workload in PR #4168

srikanth-sankaran:
> what is the impact on simpler use cases? E.g. if problems occurred with a single class with a large supertype hierarchy, how does the change impact compilation of gazillions of classes with empty or simple supertype hierarchy?

Good question! I should have thought of gathering some numbers myself without going with the gut feel that this is a superior implementation. 

Can we agree on a work load ?? Would every interface in JDK - functional and non functional from JDK 24 - sequentially checked for isFunctionalInterface() - all wrapped up in a loop that iterates 100 times be a work load for this ?

srikanth-sankaran:
> is there a measurable added effort (e.g., allocating and maintaining visitedInterfaces) that could become significant if large input comes in a different shape?

Allocating the set is O(1). 

Maintaining the set is n log(n) where n in the number of interfaces involved. Every duplicated interface is a win in not drilling down that path. Every new interface is a hit in terms on log(n) on the add() call. 

Can we agree on a workload for this ?? For worst case scenario it would be no duplications at all. 

How about an interface I that has a half a dozen supers I1 .. I6 each of which has half a dozen supers I11 .. I66 etc until I111111 .. I666666

Feel free to suggest workloads. I'll gather numbers and share here

srikanth-sankaran:
> this could perhaps be pushed upstream to SourceTypeBinding.methods()

This was the automatic question. I actually think it is safer there for annotation processing but I am fine with burying this side matter.

srikanth-sankaran:
> it is safer there for annotation processing

I meant safe not safer.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
srikanth-sankaran: Will do. 

For now and here, the distinction is:

Given the input is sorted by method name, we can stop the inner loop if the selectors differ. No further interaction will answer true to doesMethodOverride - so fail early by skipping the rest of inner loop. 

And 

If the outer loop's cursor has been seen to be overridden by the cursor method of the inner loop the outer cursor is weeded out and no point checking if it is overridden by any other methods. That is wasted effort.

srikanth-sankaran:
> In essence, every method that feeds MethodBindings into further processing, where null analysis may be relevant, must ensure, that any @NonNullByDefault has been applied where applicable.

Where it is on master could we be applying it on methods that don't eventually end up as the SAM ? OIOW, should this application be hoisted to the getSAM call rather than in the bowels ? All this with saying it is ok to retain as is of course :)

flori-schwa:
As for testdata, here is some sample code where the Performance Improvements by #4228 and #4232 (with disabled fail-fast Check) are definitely noticable: [repro.zip](https://github.com/user-attachments/files/21365279/repro.zip)
It's basically the anonymized Supertype Hierarchy (Including Methods) of the interface from our data model where the problem was observed.
The expanded Type Hierarchy does not Fit on one screen and several super-interface are encountered many different times: 
grafik

srikanth-sankaran:
[X.java.txt](https://github.com/user-attachments/files/21466225/X.java.txt)

> My typical question here is: if the huge performance gains have been observed on particularly complex / large input, what is the impact on simpler use cases?

@stephan-herrmann : Using the program above I ran some experiments. The program declares a class with 1363 interfaces from Java 24's public API. Checking each of these interface for being functional interface or not looped six times, timed and averaged shows the new implementation is slightly faster (mind you the number vary quite a bit run to run ... but it trends lower with the fix)

Also taking a relook at the code involved, I would say it is not far from intrinsic complexity and some of the optimizations such as the set while may seem like additional work pay back in fairly common cases - it is not uncommon to run into the same interface in super type walks for example. 

Let me know if you would want me to look into any additional workloads.

srikanth-sankaran:
```
$ diff new old
1364c1364,1365
< Average based on last six iterations = 234616
---
> Average based on last six iterations = 360683
> 
```

```
$ diff new2 old2
1364c1364,1365
< Average based on last six iterations = 321366
---
> Average based on last six iterations = 360683
> 
```

Some sample numbers (elasped nanoseconds to check functional interfaceness of 1363 interfaces)

where `new` numbers are with this PR and `old` are with the commit c29eb6f82f45b2a07fe663fcec054c7e8fa9ec98 which is before the PR https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4228

srikanth-sankaran:
> to check functional interfaceness of 1363 interfaces

I should also mention the results of these checks match with and without the PR for these 1363 interfaces.

srikanth-sankaran:
> How about an interface I that has a half a dozen supers I1 .. I6 each of which has half a dozen supers I11 .. I66 etc until I111111 .. I666666

I tried this which would be the worst case scenario for the current implementation - with absolutely no duplicates in super interfaces, the set maintenance is all overhead.

I generated a file I.java using this program:

```


import java.util.ArrayList;
import java.util.HashSet;
import java.util.List;
import java.util.ListIterator;
import java.util.Set;

public class X {

	static List<String> ls = new ArrayList<>();

		ls.add("I");
		Set<String> emit = new HashSet<>();
		boolean changed;
		do {
			changed = false;
			ListIterator<String> iterator = ls.listIterator();
			while (iterator.hasNext()) {
				String iface = iterator.next();
				if (emit.add(iface)) {
					System.out.print("    interface " + iface);
					if (iface.length() < 6) {
						System.out.print(" extends");
						for (int i = 1; i < 7; i++) {
							String siface = iface + String.valueOf(i);
							System.out.print((i == 1 ? " " : ", ") + siface);
							iterator.add(siface);
							changed = true;
						}
					}
					System.out.println(" {\n    }\n");
				}
			}
		} while (changed);
	}
}
```

and timed the call to `org.eclipse.jdt.internal.compiler.lookup.ReferenceBinding.isFunctionalInterface(Scope)` using this patch: (by loading I.java into an inner SDK IDE instance using this PR and with the commit  https://github.com/eclipse-jdt/eclipse.jdt.core/commit/c29eb6f82f45b2a07fe663fcec054c7e8fa9ec98 which is before any of the fail fast smarts were added.

```
diff --git a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeDeclaration.java b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeDeclaration.java
index 6d73d4a..8f647d5 100644
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeDeclaration.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeDeclaration.java
@@ -1248,12 +1248,15 @@
 				&& (sourceType.modifiers & ClassFileConstants.AccDeprecated) != 0) {
 			this.scope.problemReporter().missingDeprecatedAnnotationForType(this);
 		}
-		if ((annotationTagBits & TagBits.AnnotationFunctionalInterface) != 0) {
+		if (CharOperation.equals(this.name, "I".toCharArray()) || (annotationTagBits & TagBits.AnnotationFunctionalInterface) != 0) { //$NON-NLS-1$
+			long start = System.nanoTime();
 			if (this.binding.isSealed()) {
 				this.scope.problemReporter().functionalInterfaceMayNotBeSealed(this);
 			} else if (!this.binding.isFunctionalInterface(this.scope)) {
 				this.scope.problemReporter().notAFunctionalInterface(this);
 			}
+			long stop = System.nanoTime();
+			System.out.println("Duration of isFI : " + (stop - start)); //$NON-NLS-1$
 		}
 
 		if (!sourceType.isRecord() && (this.bits & ASTNode.UndocumentedEmptyBlock) != 0)

```

and I find that new implementation is about 8.7x slower (4180300 nano seconds vs 478800) but we are still in the 0.0041803 seconds territory. On the other hand you can see the IDE overall being very sluggish with 9331 interface types albeit each one a marker 😆 

Ergo - I don't find a performance issue with garden variety JDK interfaces 1363 of them or with a hand crafted pathological case. 

I'll proceed to integrate on that basis - with the promise that any issues found will be dealt with promptly.

srikanth-sankaran:
Thanks @flori-schwa for the problem report & contributions to the PR, @stephan-herrmann for studying the patch and forcing additional benchmarking and @jarthana for your review.

Issue #4235, opened 2025-07-21
NullPointerException during AST-Parsing


flori-schwa:
A change introduced between [I20250528-1830 and I20250720-1800](https://github.com/eclipse-jdt/eclipse.jdt.core/compare/I20250528-1830...I20250720-1800) is causing an NPE during AST-Parsing of one of our custom assertj Assertions.

Stacktrace:
```
java.lang.NullPointerException: Cannot invoke "org.eclipse.jdt.internal.compiler.lookup.TypeBinding.isUnresolvedType()" because "type" is null
	at org.eclipse.jdt.internal.compiler.lookup.TypeSystem.getUnannotatedType(TypeSystem.java:234)
	at org.eclipse.jdt.internal.compiler.lookup.TypeSystem$HashedParameterizedTypes.get(TypeSystem.java:164)
	at org.eclipse.jdt.internal.compiler.lookup.AnnotatableTypeSystem.getParameterizedType(AnnotatableTypeSystem.java:123)
	at org.eclipse.jdt.internal.compiler.lookup.AnnotatableTypeSystem.getParameterizedType(AnnotatableTypeSystem.java:141)
	at org.eclipse.jdt.internal.compiler.lookup.LookupEnvironment.createParameterizedType(LookupEnvironment.java:1401)
	at org.eclipse.jdt.internal.compiler.lookup.Scope.greaterLowerBound(Scope.java:461)
	at org.eclipse.jdt.internal.compiler.lookup.CaptureBinding.initializeBounds(CaptureBinding.java:253)
	at org.eclipse.jdt.internal.compiler.lookup.ParameterizedTypeBinding.capture(ParameterizedTypeBinding.java:212)
	at org.eclipse.jdt.internal.compiler.lookup.ParameterizedTypeBinding.capture(ParameterizedTypeBinding.java:1)
	at org.eclipse.jdt.internal.compiler.ast.MessageSend.resolveType(MessageSend.java:1018)
	at org.eclipse.jdt.internal.compiler.ast.MessageSend.resolveType(MessageSend.java:780)
	at org.eclipse.jdt.internal.compiler.ast.Expression.resolve(Expression.java:1057)
	at org.eclipse.jdt.internal.compiler.ast.Statement.resolveWithBindings(Statement.java:523)
	at org.eclipse.jdt.internal.compiler.ast.ASTNode.resolveStatements(ASTNode.java:696)
[...]

PR #4238, opened 2025-07-22
NullPointerException during AST-Parsing #4235

stephan-herrmann:
Account for null CaptureBinding.upperBound() when still initializing.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4235

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
@@ -428,50 +428,6 @@ public abstract class Scope {
 					}
 					result[j] = null;
 					removed ++;
-				} else if (!jType.isCompatibleWith(iType, scope)) {
-					// avoid creating unsatisfiable intersection types (see https://bugs.eclipse.org/405706):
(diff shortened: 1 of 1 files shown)


stephan-herrmann:
So the original problem is that we had a cyclic dependency: `CaptureBinding.initializeBounds()` calls `greaterLowerBound()` (since #4224), which in turn expected `TypeVariableBinding.upperBound()` to answer non-null for all type variables (incl. captures). If the current capture appears also in some type argument position, that guy does not have its upper bound initialized at that point.

Attempt 1 simply ignored a null upper bound, which solved the bug but left us wondering, what is the effect of opportunistically short-circuiting that code block.

Next I researched the origin of the code section (in `greaterLowerBound()`) and found it to be guess work of my own doing in order to solve some problems that got introduced with 1.7 (https://bugs.eclipse.org/405706 and
https://bugs.eclipse.org/413958). According to our tests, that entire code block is not relevant in 1.8+. Removing the old stuff feels safer in several ways: less guess work side stepping the spec, and fewer assumptions about that status of bindings during `greaterLowerBound()`.

jarthana:
I am seeing a new problem, which I believe is caused by this change. I have reported it here - https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4263

Issue #4245, opened 2025-07-25
[Formatter] Support adding blank lines between tags of different type in Markdown


SougandhS:
**Blank lines between tags of different type** doesn't work with markdown comments

<img width="491" height="319" alt="Image" src="https://github.com/user-attachments/assets/a1d9205e-f108-41a2-862d-0265772ee585" />

https://github.com/user-attachments/assets/c14cca48-e97f-4e87-abd9-cf0a9e25ad3a

PR #4246, opened 2025-07-25
Enable Blank Lines Between Different Tags in Markdown Comments via Formatter

SougandhS:
This commit enhances the formatter to correctly tokenize markdown comments and allows the insertion of blank lines between different markdown tags

Fixes : https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4245

https://github.com/user-attachments/assets/740d354b-b6d5-4f1e-b9dd-ce1671d6b904

## How to test

Enable the **_Blank lines between different markdown tags_** opt and try formating

```
class Mark {
	/// @param param1
	/// @return int
	public int sample(String param1) {
		return 0;
	}
}

```

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/formatter/FormatterRegressionTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/formatter/FormatterRegressionTests.java
@@ -16490,4 +16490,50 @@ public void testIssue2977() {
 		""",
 		CodeFormatter.K_COMPILATION_UNIT);
 }
+	public void testMarkdownSpacingFormat() throws JavaModelException {
+		setComplianceLevel(CompilerOptions.VERSION_23);
+		String input = """
+				class Mark {
+					/// @param 		param
+					///  @return 			int
+					public int sample(String param) {
+						return 0;
+					}
+				}
+				""";
+		String expected = """
+				class Mark {
+					/// @param param
+					/// @return int
+					public int sample(String param) {
+						return 0;
+					}
+				}
+				""";
+		formatSource(input, expected);
+	}
+	public void testMarkdownEmptyLinesBtwnDiffTags() throws JavaModelException {
+		setComplianceLevel(CompilerOptions.VERSION_23);
+		this.formatterPrefs.comment_insert_empty_line_between_different_tags = true;
+		String input = """
+				class Mark {
+					/// @param param1
+					/// @return int
+					public int sample(String param1) {
+						return 0;
+					}
+				}
+				""";
+		String expected = """
+				class Mark {
+					/// @param param1
+					///\s
+					/// @return int
+					public int sample(String param1) {
+						return 0;
+					}
+				}
+				""";
+		formatSource(input, expected);
+	}
 }
--- a/org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/CommentsPreparator.java
+++ b/org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/CommentsPreparator.java
@@ -14,6 +14,7 @@
  *     Mateusz Matela <mateusz.matela@gmail.com> - [formatter] Bad line breaking in Eclipse javadoc comments - https://bugs.eclipse.org/348338
  *     Lars Vogel <Lars.Vogel@vogella.com> - Contributions for
  *     						Bug 473178
+ *     IBM Corporation - Markdown support
  *******************************************************************************/
 package org.eclipse.jdt.internal.formatter;
 
@@ -603,7 +604,9 @@ public class CommentsPreparator extends ASTVisitor {
 			return false;
 
 		boolean isHeader = this.tm.isInHeader(commentIndex);
-		boolean formattingEnabled = (isHeader ? this.options.comment_format_header : this.options.comment_format_javadoc_comment);
+		boolean formattingEnabled = isHeader ? this.options.comment_format_header
+				: node.isMarkdown() ? this.options.comment_format_markdown_comment
+						: this.options.comment_format_javadoc_comment;
 		if (!formattingEnabled || !tokenizeMultilineComment(commentToken)) {
 			commentToken.setInternalStructure(commentToLines(commentToken, -1));
 			return false;
@@ -632,8 +635,9 @@ public class CommentsPreparator extends ASTVisitor {
 			return true;
 
 		int startIndex = tokenStartingAt(node.getStartPosition());
-		this.ctm.get(startIndex + 1).setWrapPolicy(WrapPolicy.DISABLE_WRAP);
-
+		if (this.ctm.size() > startIndex + 1) {
+			this.ctm.get(startIndex + 1).setWrapPolicy(WrapPolicy.DISABLE_WRAP);
+		}
 		if (node.getParent() instanceof Javadoc) {
 			assert this.ctm.toString(startIndex).startsWith(tagName);
 
@@ -1213,7 +1217,7 @@ public class CommentsPreparator extends ASTVisitor {
 			this.allowSubstituteWrapping = new boolean[commentToken.countChars()];
 		}
 		boolean isMarkdown = commentToken.tokenType == TokenNameCOMMENT_MARKDOWN;
-		boolean isJavadoc = commentToken.tokenType == TokenNameCOMMENT_JAVADOC || isMarkdown;
+		boolean isJavadoc = commentToken.tokenType == TokenNameCOMMENT_JAVADOC;
 		Arrays.fill(this.allowSubstituteWrapping, 0, commentToken.countChars(), !isJavadoc);
 
 		final boolean cleanBlankLines = isJavadoc ? this.options.comment_clear_blank_lines_in_javadoc_comment
@@ -1229,15 +1233,15 @@ public class CommentsPreparator extends ASTVisitor {
 		first.spaceAfter();
 		structure.add(first);
 
-		int lastTokenStart = commentToken.originalEnd - 1;
-		while (lastTokenStart - 1 > firstTokenEnd && this.tm.charAt(lastTokenStart - 1) == markerChar)
-			lastTokenStart--;
+		int positionLimit = isMarkdown ? commentToken.originalEnd + 1: commentToken.originalEnd - 1;
+		while (positionLimit - 1 > firstTokenEnd && this.tm.charAt(positionLimit - 1) == markerChar)
+			positionLimit--;
 
 		int position = firstTokenEnd + 1;
 		int lineBreaks = 0;
 		while (position <= commentToken.originalEnd) {
 			// find line start
-			for (int i = position; i < lastTokenStart; i++) {
+			for (int i = position; i < positionLimit; i++) {
 				char c = this.tm.charAt(i);
 				if (c == '\r' || c == '\n') {
 					lineBreaks++;
@@ -1256,22 +1260,26 @@ public class CommentsPreparator extends ASTVisitor {
 			int tokenStart = position;
 			while (position <= commentToken.originalEnd + 1) {
 				char c = 0;
-				if (position == commentToken.originalEnd + 1 || position == lastTokenStart
+				if (position == commentToken.originalEnd + 1 || position == positionLimit
 						|| ScannerHelper.isWhitespace(c = this.tm.charAt(position))) {
 					if (tokenStart < position) {
 						Token outputToken = new Token(tokenStart, position - 1, commentToken.tokenType);
 						outputToken.spaceBefore();
 						if (lineBreaks > 0) {
-							if (cleanBlankLines)
+							if (cleanBlankLines && !isMarkdown)
 								lineBreaks = 1;
-							if (lineBreaks > 1 || !this.options.join_lines_in_comments)
+							if (lineBreaks > 1 || (!this.options.join_lines_in_comments && !isMarkdown))
 								outputToken.putLineBreaksBefore(lineBreaks);
 						}
 						if (this.tm.charAt(tokenStart) == '@') {
 							outputToken.setWrapPolicy(WrapPolicy.DISABLE_WRAP);
-							if (commentToken.tokenType == TokenNameCOMMENT_BLOCK && lineBreaks == 1
-									&& structure.size() > 1) {
-								outputToken.putLineBreaksBefore(cleanBlankLines ? 1 : 2);
+							if (lineBreaks == 1 && structure.size() > 1) {
+								if (commentToken.tokenType == TokenNameCOMMENT_BLOCK) {
+									outputToken.putLineBreaksBefore(cleanBlankLines ? 1 : 2);
+								}
+								if (isMarkdown) {
+									outputToken.breakBefore();
+								}
 							}
 							if (lineBreaks > 0 && isCommonsAttributeAnnotation(this.tm.toString(outputToken))) {
 								outputToken.breakBefore();
@@ -1283,25 +1291,26 @@ public class CommentsPreparator extends ASTVisitor {
 					}
 					if (c == '\r' || c == '\n')
 						break;
-					tokenStart = position == lastTokenStart ? position : position + 1;
+					tokenStart = position == positionLimit ? position : position + 1;
 				}
 				position++;
 			}
 		}
 
 		Token last = structure.get(structure.size() - 1);
-		boolean newLinesAtBoundries = commentToken.tokenType == TokenNameCOMMENT_JAVADOC
-				? this.options.comment_new_lines_at_javadoc_boundaries
-				: this.options.comment_new_lines_at_block_boundaries;
-		if (!newLinesAtBoundries) {
-			structure.get(1).clearLineBreaksBefore();
-			last.clearLineBreaksBefore();
-		} else if (this.tm.countLineBreaksBetween(first, last) > 0) {
-			first.breakAfter();
-			last.breakBefore();
+		if (!isMarkdown) {
+			boolean newLinesAtBoundries = commentToken.tokenType == TokenNameCOMMENT_JAVADOC
+					? this.options.comment_new_lines_at_javadoc_boundaries
+					: this.options.comment_new_lines_at_block_boundaries;
+			if (!newLinesAtBoundries) {
+				structure.get(1).clearLineBreaksBefore();
+				last.clearLineBreaksBefore();
+			} else if (this.tm.countLineBreaksBetween(first, last) > 0) {
+				first.breakAfter();
+				last.breakBefore();
+			}
+			last.setAlign(1);
 		}
-		last.setAlign(1);
-
 		if (structure.size() == 2)
 			return false;
 		commentToken.setInternalStructure(structure);
--- a/org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
+++ b/org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
@@ -10,6 +10,7 @@
  *
  * Contributors:
  *     Mateusz Matela <mateusz.matela@gmail.com> - [formatter] Formatter does not format Java code correctly, especially when max line width is set - https://bugs.eclipse.org/303519
+ *     IBM Corporation - Markdown support
  *******************************************************************************/
 package org.eclipse.jdt.internal.formatter;
 
@@ -90,7 +91,8 @@ public class TextEditsBuilder extends TokenTraverser {
 
 			if (start > sourceStart) {
 				Token token = this.tm.get(this.tm.findIndex(start, ANY, false));
-				if ((token.tokenType == TokenNameCOMMENT_BLOCK || token.tokenType == TokenNameCOMMENT_JAVADOC)
+				if ((token.tokenType == TokenNameCOMMENT_BLOCK || token.tokenType == TokenNameCOMMENT_JAVADOC
+						|| token.tokenType == TokenNameCOMMENT_MARKDOWN)
 						&& start <= token.originalEnd) {
 					start = token.originalStart;
 				}
@@ -98,7 +100,8 @@ public class TextEditsBuilder extends TokenTraverser {
 
 			if (end > start && end > sourceStart) {
 				Token token = this.tm.get(this.tm.findIndex(end, ANY, false));
-				if ((token.tokenType == TokenNameCOMMENT_BLOCK || token.tokenType == TokenNameCOMMENT_JAVADOC)
+				if ((token.tokenType == TokenNameCOMMENT_BLOCK || token.tokenType == TokenNameCOMMENT_JAVADOC
+						|| token.tokenType == TokenNameCOMMENT_MARKDOWN )
 						&& end < token.originalEnd) {
 					end = token.originalEnd;
 				}
@@ -121,12 +124,6 @@ public class TextEditsBuilder extends TokenTraverser {
 	protected boolean token(Token token, int index) {
 
 		bufferWhitespaceBefore(token, index);
-
-		if (token.tokenType == TokenNameCOMMENT_MARKDOWN) {
-			flushBuffer(token.originalStart);
-			this.counter = token.originalEnd + 1;
-			return true; // don't touch markdown format for now.
-		}
 
 		List<Token> structure = token.getInternalStructure();
 		if (token.tokenType == TokenNameCOMMENT_LINE) {
@@ -196,6 +193,7 @@ public class TextEditsBuilder extends TokenTraverser {
 		}
 
 		boolean isTextBlock = token != null && token.tokenType == TokenNameTextBlock;
+		boolean isMarkdown = token != null && token.tokenType == TokenNameCOMMENT_MARKDOWN;
 		this.parent.counter = this.counter;
 		this.parent.bufferLineSeparator(null, false);
 		if (!(isTextBlock && emptyLine && !this.options.indent_empty_lines))
@@ -206,30 +204,39 @@ public class TextEditsBuilder extends TokenTraverser {
 			return;
 		if (token != null && token.tokenType == TokenNameNotAToken)
 			return; // this is an unformatted block comment, don't force asterisk
-		if (getNext() == null && !emptyLine)
-			return; // this is the last token of block comment, asterisk is included
-
-		boolean asteriskFound = false;
+		if (getNext() == null && !emptyLine) {
+			if (isMarkdown) {
+				this.buffer.append("/// "); //$NON-NLS-1$
+			}
+			return;
+		}
+		boolean markerCharFound = false;
 		int searchLimit = token != null ? token.originalStart : this.sourceLimit;
+
+		char markerChar = isMarkdown ? '/' : '*';
 		for (int i = this.counter; i < searchLimit; i++) {
 			char c = this.source.charAt(i);
-			if (c == '*') {
-				this.buffer.append(' ');
+			if (c == markerChar) {
+				if (!isMarkdown)
+					this.buffer.append(' ');
 				flushBuffer(i);
-				while (i + 1 < this.sourceLimit && this.source.charAt(i + 1) == '*')
+				while (i + 1 < this.sourceLimit && this.source.charAt(i + 1) == markerChar)
 					i++;
 				this.counter = i + 1;
 				c = this.source.charAt(i + 1);
 				if ((c != '\r' && c != '\n') || !emptyLine)
 					this.buffer.append(' ');
-				asteriskFound = true;
+				markerCharFound = true;
 				break;
 			}
 			if (!ScannerHelper.isWhitespace(c))
 				break;
 		}
-		if (!asteriskFound)
-			this.buffer.append(" * "); //$NON-NLS-1$
+
+		if (!markerCharFound) {
+			this.buffer.append(isMarkdown ? "/// " : " * "); //$NON-NLS-1$ //$NON-NLS-2$
+		}
+
 	}
 
 	private void bufferIndent(Token token, int index) {
--- a/org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/linewrap/CommentWrapExecutor.java
+++ b/org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/linewrap/CommentWrapExecutor.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2014, 2018 Mateusz Matela and others.
+ * Copyright (c) 2014, 2025 Mateusz Matela and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -12,11 +12,13 @@
  *     Mateusz Matela <mateusz.matela@gmail.com> - [formatter] Formatter does not format Java code correctly, especially when max line width is set - https://bugs.eclipse.org/303519
  *     Lars Vogel <Lars.Vogel@vogella.com> - Contributions for
  *     						Bug 473178
+ *     IBM Corporation - Markdown support
  *******************************************************************************/
 package org.eclipse.jdt.internal.formatter.linewrap;
 
 import static org.eclipse.jdt.internal.compiler.parser.TerminalToken.TokenNameCOMMENT_JAVADOC;
 import static org.eclipse.jdt.internal.compiler.parser.TerminalToken.TokenNameCOMMENT_LINE;
+import static org.eclipse.jdt.internal.compiler.parser.TerminalToken.TokenNameCOMMENT_MARKDOWN;
 import static org.eclipse.jdt.internal.compiler.parser.TerminalToken.TokenNameNotAToken;
 import static org.eclipse.jdt.internal.compiler.parser.TerminalToken.TokenNameWHITESPACE;
 import static org.eclipse.jdt.internal.formatter.CommentsPreparator.COMMENT_LINE_SEPARATOR_LENGTH;
@@ -70,9 +72,14 @@ public class CommentWrapExecutor extends TokenTraverser {
 		this.simulation = simulate;
 		this.wrapDisabled = noWrap;
 		this.potentialWrapToken = this.potentialWrapTokenSubstitute = null;
-		this.newLinesAtBoundries = commentToken.tokenType == TokenNameCOMMENT_JAVADOC
-				? this.options.comment_new_lines_at_javadoc_boundaries
-				: this.options.comment_new_lines_at_block_boundaries;
+
+		if (commentToken.tokenType != TokenNameCOMMENT_MARKDOWN) {
+			this.newLinesAtBoundries = commentToken.tokenType == TokenNameCOMMENT_JAVADOC
+					? this.options.comment_new_lines_at_javadoc_boundaries
+					: this.options.comment_new_lines_at_block_boundaries;
+		} else {
+			this.newLinesAtBoundries = false;
+		}
 
 		List<Token> structure = commentToken.getInternalStructure();
 		if (structure == null || structure.isEmpty())
@@ -211,7 +218,8 @@ public class CommentWrapExecutor extends TokenTraverser {
 		new TokenTraverser() {
 			@Override
 			protected boolean token(Token token, int index) {
-				if (token.tokenType == TokenNameCOMMENT_JAVADOC && token.getInternalStructure() == null) {
+				if ((token.tokenType == TokenNameCOMMENT_JAVADOC || token.tokenType == TokenNameCOMMENT_MARKDOWN)
+						&& token.getInternalStructure() == null) {
 					if (getLineBreaksBefore() > 0)
 						token.setAlign(token.getAlign() + token.getIndent());
 					token.setIndent(0);
--- a/org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/linewrap/WrapPreparator.java
+++ b/org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/linewrap/WrapPreparator.java
@@ -11,6 +11,7 @@
  * Contributors:
  *     Mateusz Matela <mateusz.matela@gmail.com> - [formatter] Formatter does not format Java code correctly, especially when max line width is set - https://bugs.eclipse.org/303519
  *     Mateusz Matela <mateusz.matela@gmail.com> - [formatter] follow up bug for comments - https://bugs.eclipse.org/458208
+ *     IBM Corporation - Markdown support
  *******************************************************************************/
 package org.eclipse.jdt.internal.formatter.linewrap;
 
@@ -1491,7 +1492,8 @@ public class WrapPreparator extends ASTVisitor {
 				if (token.tokenType == TokenNameCOMMENT_LINE) {
 					commentWrapper.wrapLineComment(token, startPosition);
 				} else {
-					assert token.tokenType == TokenNameCOMMENT_BLOCK || token.tokenType == TokenNameCOMMENT_JAVADOC;
+					assert token.tokenType == TokenNameCOMMENT_BLOCK || token.tokenType == TokenNameCOMMENT_JAVADOC
+							|| token.tokenType == TokenNameCOMMENT_MARKDOWN;
 					commentWrapper.wrapMultiLineComment(token, startPosition, false, false);
 				}
 			}
--- a/org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/Token.java
+++ b/org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/Token.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2014, 2021 Mateusz Matela and others.
+ * Copyright (c) 2014, 2025 Mateusz Matela and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -11,6 +11,7 @@
  * Contributors:
  *     Mateusz Matela <mateusz.matela@gmail.com> - [formatter] Formatter does not format Java code correctly, especially when max line width is set - https://bugs.eclipse.org/303519
  *     Till Brychcy - Java Code Formatter breaks code if single line comments contain unicode escape - https://bugs.eclipse.org/471090
+ *     IBM Corporation - Markdown support
  *******************************************************************************/
 package org.eclipse.jdt.internal.formatter;
 
@@ -323,6 +324,7 @@ public class Token {
 			case TokenNameCOMMENT_BLOCK:
 			case TokenNameCOMMENT_JAVADOC:
 			case TokenNameCOMMENT_LINE:
+			case TokenNameCOMMENT_MARKDOWN:
 				return true;
 			default:
 				return false;


SougandhS:
Hi @mateusz-matela , Could you please check this PR when you get time.

SougandhS:
> Hi @mateusz-matela , Could you please check this PR when you get time.

Earlier tokenization of markdown comments/javaDoc weren't working properly. that is also handled in this PR.

SougandhS:
Test failed due to a minor space issue - Will check that

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/formatter/FormatterRegressionTests.java
```diff
@@ -15504,6 +15504,32 @@ public void testBug54627b() throws JavaModelException {
 		"	}\n" +
 		"}");
 }
+
+public void testMarkdownEmptyLinesBtwnDiffTags() throws JavaModelException {
```
mateusz-matela: usually new tests are added at the end of the file, why here?

mateusz-matela: All the other tests have bug/issue number in name and description in comment, but I guess it's not a bad moment to drop that convention and adopt a more modern one.

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/CommentsPreparator.java
```diff
@@ -632,8 +632,10 @@ public boolean visit(TagElement node) {
 			return true;
 
 		int startIndex = tokenStartingAt(node.getStartPosition());
-		this.ctm.get(startIndex + 1).setWrapPolicy(WrapPolicy.DISABLE_WRAP);
-
+		Token nextToken = this.ctm.get(startIndex + 1);
+		if(nextToken!=null) {
```
mateusz-matela: I don't like the change in `Token.get` to retrun null instead of IOOBE. It's safer to assume that all the calls to this method use a valid index and fail fast if they don't. Here you can just compare size instead of checking for null.

Also, what code results in a TagElement that has only one token at the very end of a comment? I tried `/// @return` and it just isn't recognized as a TagElement

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
```diff
@@ -122,12 +122,10 @@ protected boolean token(Token token, int index) {
 
 		bufferWhitespaceBefore(token, index);
 
-		if (token.tokenType == TokenNameCOMMENT_MARKDOWN) {
-			flushBuffer(token.originalStart);
-			this.counter = token.originalEnd + 1;
-			return true; // don't touch markdown format for now.
-		}
-
+		/* [Commenting as of now - For review]
```
mateusz-matela: this is not useful for review, I see the removed code without problems

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
```diff
@@ -150,8 +148,9 @@ protected boolean token(Token token, int index) {
 			for (int i = 0; i < token.getLineBreaksAfter(); i++)
 				bufferLineSeparator(null, i + 1 == token.getLineBreaksAfter());
 			char lastChar = this.source.charAt(this.sourceLimit - 1);
-			if (token.getLineBreaksAfter() == 0 && (lastChar == '\r' || lastChar == '\n'))
+			if (token.getLineBreaksAfter() == 0 && (lastChar == '\r' || lastChar == '\n')) {
```
mateusz-matela: unrelated change - is this helpful?
I don't think coding standards in this project require code blocks for single instructions, but if they do, maybe there should be a separate commit to fix all the places

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/linewrap/CommentWrapExecutor.java
```diff
@@ -70,7 +71,8 @@ public int wrapMultiLineComment(Token commentToken, int startPosition, boolean s
 		this.simulation = simulate;
 		this.wrapDisabled = noWrap;
 		this.potentialWrapToken = this.potentialWrapTokenSubstitute = null;
-		this.newLinesAtBoundries = commentToken.tokenType == TokenNameCOMMENT_JAVADOC
+		this.newLinesAtBoundries = ( commentToken.tokenType == TokenNameCOMMENT_JAVADOC || commentToken.tokenType ==  TokenNameCOMMENT_MARKDOWN
```
mateusz-matela: this is about having `/*` and `*/` in separate lines, doesn't make sense for markdown, better to set to false.
I think this is why in the test case the line `/// @return int` gets separated into two lines

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
```diff
@@ -206,17 +205,20 @@ private void bufferLineSeparator(Token token, boolean emptyLine) {
...
+		}
 		boolean asteriskFound = false;
 		int searchLimit = token != null ? token.originalStart : this.sourceLimit;
 		for (int i = this.counter; i < searchLimit; i++) {
 			char c = this.source.charAt(i);
 			if (c == '*') {
 				this.buffer.append(' ');
 				flushBuffer(i);
-				while (i + 1 < this.sourceLimit && this.source.charAt(i + 1) == '*')
+				while (i + 1 < this.sourceLimit && this.source.charAt(i + 1) == '*') //TODO: Markdown++
```
mateusz-matela: I wouldn't know what this TODO means (same below)

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
```diff
@@ -225,11 +227,32 @@ private void bufferLineSeparator(Token token, boolean emptyLine) {
 				asteriskFound = true;
 				break;
 			}
+			if (c == '/') {
```
mateusz-matela: these ifs should now also check token type, otherwise expecting problems for line comments containing asterisks or block comments containing slashes

SougandhS: Okay, added token type check

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
SougandhS: Forgot to remove that.
Removed 👍

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/linewrap/CommentWrapExecutor.java
SougandhS: Thanks for the clarification 👍 
Updated.

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
SougandhS: Removed it.

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
SougandhS: Had a doubt before removing it, So kept it like that
- Removed it now

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/CommentsPreparator.java
SougandhS: During development I encountered this IOOBE so to prevent it from future edge cases I added.

updated as you suggested

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/formatter/FormatterRegressionTests.java
SougandhS: In JDT Debug most of the tests are named in this convention so followed it here.

SougandhS: Updated it to last

SougandhS:
> The biggest problem is that blank lines are added before every tag, regardless of settings! (at least when I test starting with the Eclipse build-in profile)

I had tested it with custom profile (Didn't encountered the issue), thought behaviour would be same in all - Now I do see the tag issue you encountered when switched to default.
I will try to fix the issue on default profile 👍 

> Another point: remember to reformat the modified lines, I saw quite a few missing spaces and such

Yes, will re-do with default profiler 

Thank you

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
SougandhS: I have reverted the additional `tokenType` check as of now as it was affecting other test cases

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/linewrap/CommentWrapExecutor.java
SougandhS: I solved that issue with 
```
if ((index == 1 || getNext() == null) && this.newLinesAtBoundries && lineBreaksBefore == 0 ) {
	if (!this.simulation && token.tokenType != TokenNameCOMMENT_MARKDOWN) // disabling break for markdowns
		token.breakBefore();
	lineBreaksBefore = 1;
}
```

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
SougandhS: Yet I am facing with additional space and new lines (Even when option is turned off)

Screenshot 2025-07-26 at 10 50 21 AM

SougandhS:
>Spacing issue is still there (In test case)

Test case space issue fixed by a workaround

SougandhS:
Hi @mateusz-matela could you please check this again ?

SougandhS:
tag issue & space issue is now resolved 👍 

https://github.com/user-attachments/assets/c0769b73-5240-4089-9a6d-d693182b93b2

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/formatter/FormatterRegressionTests.java
```diff
@@ -16490,4 +16491,50 @@ void main() {
 		""",
 		CodeFormatter.K_COMPILATION_UNIT);
 }
+	public void testMarkdownSpacingFormat() throws JavaModelException {
+		setComplianceLevel(CompilerOptions.VERSION_23);
+		String expected = """
```
mateusz-matela, line 16504: expected string declaration should be below the input code

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/formatter/FormatterRegressionTests.java
```diff
@@ -16490,4 +16491,50 @@ void main() {
...
+			    	/// @return int
+			    	public int sample(String param1) {
+			    		return 0;
+			    	}
+			    }
+			    """;
+		String expected = """
+		class Mark {
+			/// @param param1
+			/// \n\t/// @return int
```
mateusz-matela: use true newline and tab instead of `\n\t`

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/formatter/FormatterRegressionTests.java
```diff
@@ -15504,6 +15504,7 @@ public void testBug54627b() throws JavaModelException {
 		"	}\n" +
 		"}");
 }
+
```
mateusz-matela: leftover irrelevant change

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
```diff
@@ -206,14 +200,17 @@ private void bufferLineSeparator(Token token, boolean emptyLine) {
...
+			}
 			return; // this is the last token of block comment, asterisk is included
-
+		}
 		boolean asteriskFound = false;
 		int searchLimit = token != null ? token.originalStart : this.sourceLimit;
 		for (int i = this.counter; i < searchLimit; i++) {
 			char c = this.source.charAt(i);
-			if (c == '*') {
+			if (c == '*' ) {
```
mateusz-matela: I suggest `&& token.tokenType != TokenNameCOMMENT_MARKDOWN` - it doesn't break tests

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
```diff
@@ -206,14 +200,17 @@ private void bufferLineSeparator(Token token, boolean emptyLine) {
 			return;
 		if (token != null && token.tokenType == TokenNameNotAToken)
 			return; // this is an unformatted block comment, don't force asterisk
-		if (getNext() == null && !emptyLine)
+		if (getNext() == null && !emptyLine) {
+			if(token.tokenType == TokenNameCOMMENT_MARKDOWN) {
+				this.buffer.append("///"); //$NON-NLS-1$
```
mateusz-matela: this needs to be `"/// "` (tested on a markdown with last line `/// @return`)
edit1: it gets complicated if there can be more than 3 slashes
edit2: it should work when the first if is chaned to `if (getNext() == null && !emptyLine && token.tokenType != TokenNameCOMMENT_MARKDOWN)`, second `if` not needed

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
```diff
@@ -225,11 +222,30 @@ private void bufferLineSeparator(Token token, boolean emptyLine) {
...
 		}
-		if (!asteriskFound)
-			this.buffer.append(" * "); //$NON-NLS-1$
+		if (!asteriskFound) {
+			if(token.tokenType == TokenNameCOMMENT_MARKDOWN) {
+				this.buffer.append("/// "); //$NON-NLS-1$
+			} else {
+				this.buffer.append(" * ");//$NON-NLS-1$
+			}
+
```
mateusz-matela, line 235: usually there are no blank lines before closing brace

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
```diff
@@ -225,11 +222,30 @@ private void bufferLineSeparator(Token token, boolean emptyLine) {
 				asteriskFound = true;
 				break;
 			}
+			if (c == '/' && token.tokenType == TokenNameCOMMENT_MARKDOWN) {
+				while(c=='/') {
+					i++;
+					c = this.source.charAt(i);
+				}
+				this.buffer.append("///"); //$NON-NLS-1$
```
mateusz-matela: what should happen if there are four or more slashes in the input? This will remove the extra slashes. I don't know it it's good, but if it is, then it doesn't work fork the same way for the first line of a markdown comment
edit: it preserves the number of slashes if the code for slashes is identical as the code for asterisks (except for `this.buffer.append(' ');` at the beginning), so maybe it should be the same code, searching for markerChar based on tokenType instead of hardcoded `'*'`

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
```diff
@@ -206,14 +200,17 @@ private void bufferLineSeparator(Token token, boolean emptyLine) {
...
 			return; // this is an unformatted block comment, don't force asterisk
-		if (getNext() == null && !emptyLine)
+		if (getNext() == null && !emptyLine) {
+			if(token.tokenType == TokenNameCOMMENT_MARKDOWN) {
+				this.buffer.append("///"); //$NON-NLS-1$
+			}
 			return; // this is the last token of block comment, asterisk is included
-
+		}
 		boolean asteriskFound = false;
```
mateusz-matela: the term asterisk could be misleading now? in another place it was called markerChar

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
```diff
@@ -206,14 +200,17 @@ private void bufferLineSeparator(Token token, boolean emptyLine) {
 			return;
 		if (token != null && token.tokenType == TokenNameNotAToken)
 			return; // this is an unformatted block comment, don't force asterisk
-		if (getNext() == null && !emptyLine)
+		if (getNext() == null && !emptyLine) {
+			if(token.tokenType == TokenNameCOMMENT_MARKDOWN) {
```
mateusz-matela: so far all the places referencing `token` checked for null, I guess it's safer to keep it that way

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/linewrap/CommentWrapExecutor.java
```diff
@@ -132,8 +133,8 @@ protected boolean token(Token token, int index) {
 		final int positionIfNewLine = getStartingPosition(token, true);
 
 		int lineBreaksBefore = getLineBreaksBefore();
-		if ((index == 1 || getNext() == null) && this.newLinesAtBoundries && lineBreaksBefore == 0) {
-			if (!this.simulation)
+		if ((index == 1 || getNext() == null) && this.newLinesAtBoundries && lineBreaksBefore == 0 ) {
+			if (!this.simulation && token.tokenType != TokenNameCOMMENT_MARKDOWN)
```
mateusz-matela: I don't think tokenType is relevant only if it's not simulation, this condition should be added in the if above

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/CommentsPreparator.java
```diff
@@ -1269,9 +1270,13 @@ private boolean tokenizeMultilineComment(Token commentToken) {
...
 							outputToken.setWrapPolicy(WrapPolicy.DISABLE_WRAP);
-							if (commentToken.tokenType == TokenNameCOMMENT_BLOCK && lineBreaks == 1
-									&& structure.size() > 1) {
-								outputToken.putLineBreaksBefore(cleanBlankLines ? 1 : 2);
+							if (lineBreaks == 1 && structure.size() > 1) {
+								if(commentToken.tokenType == TokenNameCOMMENT_BLOCK) {
+									outputToken.putLineBreaksBefore(cleanBlankLines ? 1 : 2);
+								}
+								if(commentToken.tokenType == TokenNameCOMMENT_MARKDOWN) {
+									outputToken.putLineBreaksBefore(1);
```
mateusz-matela: this could be shortened to `outputToken.breakBefore()`.
edit: is this really needed? when I comment out this line everything seems to work the same, which would mean this whole code section could go back to the original form

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/CommentsPreparator.java
```diff
@@ -1229,7 +1230,7 @@ private boolean tokenizeMultilineComment(Token commentToken) {
 		first.spaceAfter();
 		structure.add(first);
 
-		int lastTokenStart = commentToken.originalEnd - 1;
+		int lastTokenStart = isMarkdown? commentToken.originalEnd+ 1: commentToken.originalEnd - 1;
```
mateusz-matela: for markdown this will be the position after the comment, so `lastTokenStart` is no longer a good name. Maybe `positionLimit`?

SougandhS: `positionLimit` sounds great!

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/CommentsPreparator.java
SougandhS: I have observed an additional space getting generated without this.
shortened to outputToken.breakBefore() as suggested

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/linewrap/CommentWrapExecutor.java
SougandhS: okay, moved it up

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
SougandhS: Null check added 👍

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
SougandhS: refactored it to `markerChar` 👍

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
SougandhS: removed 👍

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
SougandhS: updated

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
SougandhS: I'll try this

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/formatter/FormatterRegressionTests.java
SougandhS: reverted

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/formatter/FormatterRegressionTests.java
SougandhS: the empty line after `/// @param param1` generates `"/// "` (an additional space is added after tag) but when I include this additional space and save the editor some how it removes that added empty space.

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
SougandhS: Sorry I didn't get this. Is it possible for an example ? 
>edit :  what should happen if there are four or more slashes in the input? 

it fails to remove the extra line 

>edit: it preserves the number of slashe...

Didnt understand much

mmatela: one if for both cases:

			if (c == markerChar) {
				if (!isMarkdown)
					this.buffer.append(' ');
				flushBuffer(i);
				while (i + 1 < this.sourceLimit && this.source.charAt(i + 1) == markerChar)
					i++;
				this.counter = i + 1;
				c = this.source.charAt(i + 1);
				if ((c != '\r' && c != '\n') || !emptyLine)
					this.buffer.append(' ');
				markerCharFound = true;
				break;
			}

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/formatter/FormatterRegressionTests.java
SougandhS: done

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
SougandhS: thanks

SougandhS: I'll try this and update

SougandhS:
>Still seeing a few formatting issues - missing spaces and lines exceeding 120 chars limit.

Could please share the test case of this ?

> I've also noticed that the commit's author is `sougandhs`, for most other commits it's the full name

Random github bug I guess..

mateusz-matela:
> > Still seeing a few formatting issues - missing spaces and lines exceeding 120 chars limit.
>
> Could please share the test case of this ?

Oh, I meant you forgot to reformat the code, not a wrong formatter output :)

SougandhS:
I have reverted the current review commit I pushed -> Build was consistently failing. 
I'll try again -> Maybe it was affecting some area

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
SougandhS: > edit2: it should work when the first if is chaned to if (getNext()

I see on chaining it affects other tests

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/formatter/FormatterRegressionTests.java
mateusz-matela: Oh right, it must be the "Remove trailing white spaces on all lines" save action and/or java text block ignoring trailing spaces. It seems that trailing space can also be represented as `\s`, which I think would be more readable.

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/CommentsPreparator.java
```diff
@@ -632,8 +632,9 @@ public boolean visit(TagElement node) {
 			return true;
 
 		int startIndex = tokenStartingAt(node.getStartPosition());
-		this.ctm.get(startIndex + 1).setWrapPolicy(WrapPolicy.DISABLE_WRAP);
-
+		if(this.ctm.size()>startIndex + 1) {
```
mateusz-matela: missing spaces

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/CommentsPreparator.java
```diff
@@ -1229,15 +1230,15 @@ private boolean tokenizeMultilineComment(Token commentToken) {
 		first.spaceAfter();
 		structure.add(first);
 
-		int lastTokenStart = commentToken.originalEnd - 1;
-		while (lastTokenStart - 1 > firstTokenEnd && this.tm.charAt(lastTokenStart - 1) == markerChar)
-			lastTokenStart--;
+		int positionLimit = isMarkdown? commentToken.originalEnd+ 1: commentToken.originalEnd - 1;
```
mateusz-matela: missing spaces

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/CommentsPreparator.java
```diff
@@ -1269,9 +1270,13 @@ private boolean tokenizeMultilineComment(Token commentToken) {
 						}
 						if (this.tm.charAt(tokenStart) == '@') {
 							outputToken.setWrapPolicy(WrapPolicy.DISABLE_WRAP);
-							if (commentToken.tokenType == TokenNameCOMMENT_BLOCK && lineBreaks == 1
-									&& structure.size() > 1) {
-								outputToken.putLineBreaksBefore(cleanBlankLines ? 1 : 2);
+							if (lineBreaks == 1 && structure.size() > 1) {
+								if(commentToken.tokenType == TokenNameCOMMENT_BLOCK) {
```
mateusz-matela: missing space

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/CommentsPreparator.java
```diff
@@ -1269,9 +1270,13 @@ private boolean tokenizeMultilineComment(Token commentToken) {
...
+									outputToken.putLineBreaksBefore(cleanBlankLines ? 1 : 2);
+								}
+								if(commentToken.tokenType == TokenNameCOMMENT_MARKDOWN) {
```
mateusz-matela: missing space

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/CommentsPreparator.java
```diff
@@ -1296,12 +1301,13 @@ private boolean tokenizeMultilineComment(Token commentToken) {
...
 			structure.get(1).clearLineBreaksBefore();
 			last.clearLineBreaksBefore();
-		} else if (this.tm.countLineBreaksBetween(first, last) > 0) {
+		} else if (this.tm.countLineBreaksBetween(first, last) > 0  && commentToken.tokenType != TokenNameCOMMENT_MARKDOWN) {
 			first.breakAfter();
 			last.breakBefore();
 		}
-		last.setAlign(1);
-
+		if(commentToken.tokenType != TokenNameCOMMENT_MARKDOWN) {
```
mateusz-matela: missing space

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
```diff
@@ -205,31 +199,43 @@ private void bufferLineSeparator(Token token, boolean emptyLine) {
 		if (isTextBlock)
 			return;
 		if (token != null && token.tokenType == TokenNameNotAToken)
-			return; // this is an unformatted block comment, don't force asterisk
-		if (getNext() == null && !emptyLine)
-			return; // this is the last token of block comment, asterisk is included
-
-		boolean asteriskFound = false;
+			return; // this is an unformatted block comment, don't force markerChar
```
mateusz-matela: comment says it's a block comment, so it's still asterisk, no need to change to markerChar

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
```diff
@@ -205,31 +199,43 @@ private void bufferLineSeparator(Token token, boolean emptyLine) {
...
+		if (getNext() == null && !emptyLine) {
+			if(token != null && token.tokenType == TokenNameCOMMENT_MARKDOWN) {
+				this.buffer.append("/// "); //$NON-NLS-1$
+			}
+			return; // this is the last token of block comment, markerChar is included
```
mateusz-matela: this comment would better fit in else block of the last if, here it could be misleading... maybe it's not that helpful anyway, I think it can be skipped

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
```diff
@@ -205,31 +199,43 @@ private void bufferLineSeparator(Token token, boolean emptyLine) {
...
 			return;
 		if (token != null && token.tokenType == TokenNameNotAToken)
-			return; // this is an unformatted block comment, don't force asterisk
-		if (getNext() == null && !emptyLine)
-			return; // this is the last token of block comment, asterisk is included
-
-		boolean asteriskFound = false;
+			return; // this is an unformatted block comment, don't force markerChar
+		if (getNext() == null && !emptyLine) {
+			if(token != null && token.tokenType == TokenNameCOMMENT_MARKDOWN) {
```
mateusz-matela: missing space

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
```diff
@@ -205,31 +199,43 @@ private void bufferLineSeparator(Token token, boolean emptyLine) {
...
+				this.buffer.append("/// "); //$NON-NLS-1$
+			}
+			return; // this is the last token of block comment, markerChar is included
+		}
+		boolean markerCharFound = false;
 		int searchLimit = token != null ? token.originalStart : this.sourceLimit;
+		char markerChar = token != null && token.tokenType == TokenNameCOMMENT_MARKDOWN ? '/' : '*';
+		boolean isMarkdown = token != null && token.tokenType == TokenNameCOMMENT_MARKDOWN;
```
mateusz-matela: you can declare `isMarkdwon` first and then use it in markerChar initialization

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
```diff
@@ -205,31 +199,43 @@ private void bufferLineSeparator(Token token, boolean emptyLine) {
...
+				markerCharFound = true;
 				break;
 			}
 			if (!ScannerHelper.isWhitespace(c))
 				break;
 		}
-		if (!asteriskFound)
-			this.buffer.append(" * "); //$NON-NLS-1$
+		if (!markerCharFound) {
+			if(token.tokenType == TokenNameCOMMENT_MARKDOWN) {
```
mateusz-matela: missing space
how about `this.buffer.append(isMarkdown ? "/// " : " * ")`

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/linewrap/CommentWrapExecutor.java
```diff
@@ -132,7 +133,7 @@ protected boolean token(Token token, int index) {
 		final int positionIfNewLine = getStartingPosition(token, true);
 
 		int lineBreaksBefore = getLineBreaksBefore();
-		if ((index == 1 || getNext() == null) && this.newLinesAtBoundries && lineBreaksBefore == 0) {
+		if ((index == 1 || getNext() == null) && this.newLinesAtBoundries && lineBreaksBefore == 0 && token.tokenType != TokenNameCOMMENT_MARKDOWN) {
```
mateusz-matela: 120 char limit exceeded

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/linewrap/CommentWrapExecutor.java
```diff
@@ -211,7 +212,7 @@ private void cleanupIndent(List<Token> structure) {
 		new TokenTraverser() {
 			@Override
 			protected boolean token(Token token, int index) {
-				if (token.tokenType == TokenNameCOMMENT_JAVADOC && token.getInternalStructure() == null) {
+				if ((token.tokenType == TokenNameCOMMENT_JAVADOC || token.tokenType == TokenNameCOMMENT_MARKDOWN) && token.getInternalStructure() == null) {
```
mateusz-matela: 120 char limit exceeded

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/linewrap/WrapPreparator.java
```diff
@@ -1491,7 +1491,7 @@ private void wrapComments() {
 				if (token.tokenType == TokenNameCOMMENT_LINE) {
 					commentWrapper.wrapLineComment(token, startPosition);
 				} else {
-					assert token.tokenType == TokenNameCOMMENT_BLOCK || token.tokenType == TokenNameCOMMENT_JAVADOC;
+					assert token.tokenType == TokenNameCOMMENT_BLOCK || token.tokenType == TokenNameCOMMENT_JAVADOC || token.tokenType == TokenNameCOMMENT_MARKDOWN;
```
mateusz-matela: 120 char limit exceeded

mateusz-matela (changes requested):

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/CommentsPreparator.java
SougandhS: Added

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/CommentsPreparator.java
SougandhS: Added

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/CommentsPreparator.java
SougandhS: Added

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/CommentsPreparator.java
SougandhS: Added

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/CommentsPreparator.java
SougandhS: Added

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
SougandhS: Reverted

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
SougandhS: okay, removing the comment

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
SougandhS: Added

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
SougandhS: okay, changed it

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
SougandhS: `this.buffer.append(isMarkdown ? "/// " : " * ")` is great

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/linewrap/CommentWrapExecutor.java
SougandhS: okay, will add it in a new line

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/linewrap/CommentWrapExecutor.java
SougandhS: refactored

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/linewrap/WrapPreparator.java
SougandhS: refactored

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/CommentsPreparator.java
```diff
@@ -1296,12 +1302,13 @@ private boolean tokenizeMultilineComment(Token commentToken) {
 		if (!newLinesAtBoundries) {
 			structure.get(1).clearLineBreaksBefore();
 			last.clearLineBreaksBefore();
-		} else if (this.tm.countLineBreaksBetween(first, last) > 0) {
+		} else if (this.tm.countLineBreaksBetween(first, last) > 0  && commentToken.tokenType != TokenNameCOMMENT_MARKDOWN) {
```
mateusz-matela: this seems fishy: newLinesAtBoundries will be equal to the _Block comments -> /* and */ on separate lines_, so that setting will affect behavior for markdown comments, probably leading to invalid results.
Looks like the whole section from `Token last = structure....` to `last.setAlign(1);` should be in one big `if (!isMarkdown)` .

Oh, and what about `if (structure.size() == 2)` further down? This used to mean empty comment: `/* */` or `/** */`. For markdown the equivalent would be `///` so `size() == 1` (could you confirm in debug?), `size == 2` means non-empty markdown so it should `setInternalStructure`.

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
```diff
@@ -90,15 +91,15 @@ private List<IRegion> adaptRegions(List<IRegion> givenRegions) {
 
 			if (start > sourceStart) {
 				Token token = this.tm.get(this.tm.findIndex(start, ANY, false));
-				if ((token.tokenType == TokenNameCOMMENT_BLOCK || token.tokenType == TokenNameCOMMENT_JAVADOC)
+				if ((token.tokenType == TokenNameCOMMENT_BLOCK || token.tokenType == TokenNameCOMMENT_JAVADOC || token.tokenType == TokenNameCOMMENT_MARKDOWN)
```
mateusz-matela: line limit exceeded

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
```diff
@@ -90,15 +91,15 @@ private List<IRegion> adaptRegions(List<IRegion> givenRegions) {
...
 						&& start <= token.originalEnd) {
 					start = token.originalStart;
 				}
 			}
 
 			if (end > start && end > sourceStart) {
 				Token token = this.tm.get(this.tm.findIndex(end, ANY, false));
-				if ((token.tokenType == TokenNameCOMMENT_BLOCK || token.tokenType == TokenNameCOMMENT_JAVADOC)
+				if ((token.tokenType == TokenNameCOMMENT_BLOCK || token.tokenType == TokenNameCOMMENT_JAVADOC || token.tokenType == TokenNameCOMMENT_MARKDOWN )
```
mateusz-matela: line limit exceeded

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
mateusz-matela: Oh, `isMarkdown` could be declared even earlier and also used in the `if` above

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/formatter/FormatterRegressionTests.java
```diff
@@ -16490,4 +16490,50 @@ void main() {
 		""",
 		CodeFormatter.K_COMPILATION_UNIT);
 }
+	public void testMarkdownSpacingFormat() throws JavaModelException {
+		setComplianceLevel(CompilerOptions.VERSION_23);
+		String input = """
+			    class Mark {
```
mateusz-matela: these text block lines are indented with 3 tabs followed by 4 spaces. Change to 4 tabs. Same for blocks below

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/formatter/FormatterRegressionTests.java
```diff
@@ -16490,4 +16490,50 @@ void main() {
...
+			    	/// @param param1
+			    	/// @return int
+			    	public int sample(String param1) {
+			    		return 0;
+			    	}
+			    }
+			    """;
+		String expected = """
+		class Mark {
```
mateusz-matela: these text block lines are indented with 2 tabs. Change to 4 tabs.

mateusz-matela (changes requested):
>> I've also noticed that the commit's author is sougandhs, for most other commits it's the full name

> Random github bug I guess..

I'm not sure how exactly you're adding commits, does it not let you set the author field? Maybe you could prepare the final commit in eclipse (egit), I always use that way.

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/CommentsPreparator.java
mateusz-matela: And the same probem with `newLinesAtBoundries` in `CommentWrapExecutor`: it's used in 2 places, you've already secured one of them, but not the other.

SougandhS:
> I'm not sure how exactly you're adding commits, does it not let you set the author field? Maybe you could prepare the final commit in eclipse (egit), I always use that way.

Actually I use egit
image

This is how I always committed :)

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/CommentsPreparator.java
SougandhS: >Looks like the whole section from Token last = structure.... to last.setAlign(1); should be in one big if 

Done

>could you confirm in debug?

Empty markdown comment is currently being removed. No comment arrays are created in `buildCommentsTable(CompilationUnit, int[][])` , it does not reach `CommentsPreparator.tokenizeMultilineComment(Token)`

https://github.com/user-attachments/assets/b7f31115-1dbb-4a3a-9ac4-f58b882e6f2a

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/formatter/FormatterRegressionTests.java
SougandhS: done

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/formatter/FormatterRegressionTests.java
SougandhS: done

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
SougandhS: Done

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
SougandhS: Done.

mateusz-matela:
> Actually I use egit 

So you can edit the Author and Commiter fields when you create a commit, it's not a "random github bug" setting these values.
It's your choice I guess, I'm just pointing out that everyone else puts their full name before the email and it looks more professional in history ;)

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/CommentsPreparator.java
mateusz-matela: > Empty markdown comment is currently being removed

And you want to leave it like that?

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/formatter/FormatterRegressionTests.java
mateusz-matela: It's still mixed - now it's 4 tabs followed by 4 spaces.
You can toggle "Show whitespace characters" in eclispe toolbar to see it

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/formatter/FormatterRegressionTests.java
mateusz-matela: Now it's 5 tabs.

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/linewrap/CommentWrapExecutor.java
```diff
@@ -70,10 +72,11 @@ public int wrapMultiLineComment(Token commentToken, int startPosition, boolean s
 		this.simulation = simulate;
 		this.wrapDisabled = noWrap;
 		this.potentialWrapToken = this.potentialWrapTokenSubstitute = null;
-		this.newLinesAtBoundries = commentToken.tokenType == TokenNameCOMMENT_JAVADOC
-				? this.options.comment_new_lines_at_javadoc_boundaries
-				: this.options.comment_new_lines_at_block_boundaries;
-
+		if(commentToken.tokenType != TokenNameCOMMENT_MARKDOWN) {
```
mateusz-matela: I meant securing the place where newLinesAtBoundries is used, but this will work too, I'm not sure which approach is more intuitive.
With this approach the checking added in line 139 is redundant, it can be reversed.

Also, turn on a save action to format edited line or something. You don't expect me to keep directly pointing out every single missing space, do you?

mateusz-matela (changes requested):

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/linewrap/CommentWrapExecutor.java
mateusz-matela: Actually, wait - we don't create a new instance of CommentWrapExecutor for every token, so we need to explicitly set newLinesAtBoundries to false for markdown, so that previous block comment tokens don't affect it.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/formatter/FormatterRegressionTests.java
SougandhS: Is this correct ?

Screenshot 2025-08-08 at 4 36 34 AM

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/linewrap/CommentWrapExecutor.java
SougandhS: > turn on a save action to format edited line or something

Sure thanks 

>139 is redundant, it can be reversed

Reversed 

>so we need to explicitly set newLinesAtBoundries to false for markdown

Done an approach can you check the commit ?

SougandhS:
> It's your choice I guess, I'm just pointing out that everyone else puts their full name before the email and it looks more professional in history ;)

Thank you for letting me know this, I'll definitely update it.

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/CommentsPreparator.java
SougandhS: This case to is to be handled then

SougandhS: @mateusz-matela : I have done a fix for this in the latest commit -> could you please confirm that ?

SougandhS: https://github.com/user-attachments/assets/1ada8564-a005-48fd-a985-5e5f0b0fb8ad

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/formatter/FormatterRegressionTests.java
```diff
@@ -16490,4 +16490,71 @@ void main() {
...
+				""";
+		String expected = """
+				class Mark {
+					///
+					public int sample(String param1) {
+						return 0;
+					}
+				}
+				""";
+		formatSource(input,expected);
```
mateusz-matela: space

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
```diff
@@ -1726,7 +1726,7 @@ protected TerminalToken getNextToken0() throws InvalidInputException {
 									}
 									if (!lineBeginsWithMarkdown()) {
 										this.currentPosition--;
-										break;
+										//break; ->Commenting for testing
```
mateusz-matela: I don't feel confident reviewing changes outside of the formatter. I'd suggest a separate pull request with the change explained, a compiler specific unit test and a review from a compiler expert.

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/linewrap/CommentWrapExecutor.java
```diff
@@ -211,7 +212,8 @@ private void cleanupIndent(List<Token> structure) {
 		new TokenTraverser() {
 			@Override
 			protected boolean token(Token token, int index) {
-				if (token.tokenType == TokenNameCOMMENT_JAVADOC && token.getInternalStructure() == null) {
+				if ((token.tokenType == TokenNameCOMMENT_JAVADOC || token.tokenType == TokenNameCOMMENT_MARKDOWN)
+					&& token.getInternalStructure() == null) {
```
mateusz-matela: wrong indent

Also, why is this check needed? cleanupIndent is only called with internal tokens of a multiline comment. These comments can also have internal structure, but that's only possible with code snippets in `<pre>` tags, so it doesn't seem to be something relevant now, as code snippets in markdown are not recognized anyway.

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/linewrap/CommentWrapExecutor.java
```diff
@@ -73,7 +75,6 @@ public int wrapMultiLineComment(Token commentToken, int startPosition, boolean s
 		this.newLinesAtBoundries = commentToken.tokenType == TokenNameCOMMENT_JAVADOC
 				? this.options.comment_new_lines_at_javadoc_boundaries
 				: this.options.comment_new_lines_at_block_boundaries;
-
```
mateusz-matela, line 76: irrelevant change: empty line removed with nothing else changing in this method

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/linewrap/CommentWrapExecutor.java
```diff
@@ -309,4 +311,8 @@ private int getLineLimit(int startPosition) {
 			lineLength = pageWidth;
 		return lineLength;
 	}
+
+	public void setNewLinesAtBoundries(boolean value) {
+		this.newLinesAtBoundries = value;
```
mateusz-matela: Could you explain why it's necessary to set this from other classes? I thought it would be enough to initialize it in wrapMultiLineComment like this:

		this.newLinesAtBoundries = commentToken.tokenType != TokenNameCOMMENT_MARKDOWN
				&& commentToken.tokenType == TokenNameCOMMENT_JAVADOC
						? this.options.comment_new_lines_at_javadoc_boundaries
						: this.options.comment_new_lines_at_block_boundaries;

If this approach is faulty, maybe it would be simpler to go back to securing where newLinesAtBoundries is used, not where it is set - restore the checking around line 139 (which was redundant) and also add checking in line 90.

SougandhS:
> Sorry to see you just won't make sure your code is reformatted before pushing changes, I guess I'll have to get used to that...

Apologies, I will fix all formatting issues once every other issue is resolved before merging and make sure I follow this in my next PRs.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
SougandhS: I will discuss this issue with our compiler leads and raise this as a separate issue.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/formatter/FormatterRegressionTests.java
SougandhS: image

Sorry I always enable all code mining features, so it was bit hard to see that, thanks for pointing it out

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/linewrap/CommentWrapExecutor.java
mateusz-matela: Never mind, checking `token.tokenType == TokenNameCOMMENT_MARKDOWN` makes perfect sense, somehow I thought that you added checking `token.getInternalStructure() == null`

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/linewrap/CommentWrapExecutor.java
SougandhS: reverted

SougandhS:
>A quick workaround for now would be to never remove empty lines and never join lines in markdown. A more ambitious solution for the future is for the formatter to parse the markdown elements so that it knows when joining lines is safe (unless the compiler already provides markdown structure of this kind?)

Done the workaround for now

>They could turn off javadoc formatting to prevent this, but it's not very intuitive and they might want to keep the formatting of javadocs on. So we need a separate setting to disable formatting of markdown comments. I'm not saying this must be a part of this pull request, but we have to make sure there won't be a release that introduces markdown formatting but doesn't have a setting to turn it off

I will check this, possibly incorporate this in current PR itself once every other issues are handled

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/linewrap/CommentWrapExecutor.java
SougandhS: the suggested approach was empty markdown at beginning when `comment_new_lines_at_block_boundaries` enabled, tried a straight forward approach now

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TokenManager.java
```diff
@@ -10,6 +10,7 @@
  *
  * Contributors:
  *     Mateusz Matela <mateusz.matela@gmail.com> - [formatter] Formatter does not format Java code correctly, especially when max line width is set - https://bugs.eclipse.org/303519
+ *     IBM Corporation - Markdown support
```
mateusz-matela: Now nothing else changes in this file

mateusz-matela (commented):
Just one trivial thing, otherwise looks good!

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Scanner.java
SougandhS: @mateusz-matela I discussed this issue with Jayaprakash and he is fine with this change, and this will be handled in a separate PR

SougandhS: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4311 Raised

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TokenManager.java
SougandhS: I'll squash the commits and go through the changes for formatting issues 👍

SougandhS:
> > They could turn off javadoc formatting to prevent this, but it's not very intuitive and they might want to keep the formatting of javadocs on. So we need a separate setting to disable formatting of markdown comments. I'm not saying this must be a part of this pull request, but we have to make sure there won't be a release that introduces markdown formatting but doesn't have a setting to turn it off
> 
> I will check this, possibly incorporate this in current PR itself once every other issues are handled

I will raise a separate PR and issue for this since UI related changes are in JDT UI

image

SougandhS:
>A more ambitious solution for the future is for the formatter to parse the markdown elements so that it knows when joining lines is safe (unless the compiler already provides markdown structure of this kind?)

@mateusz-matela This can be configured via `CommentsPreparator.handleHtml(TagElement)` right ? 
- by providing appropriate matcher for markdown patterns

SougandhS:
> @mateusz-matela This can be configured via `CommentsPreparator.handleHtml(TagElement)` right ?

I tried providing a custom pattern matcher for markdown (only for formatting lists in paragraphs lists [ - ] ) it actually solved that issue (Not fully tested)

https://github.com/user-attachments/assets/78f1c2e3-e93e-461a-a98b-1d16156596d1

Just a sample code I tried for markdown lists (Not the final) -
```
private final static Pattern CUSTOM_MARKDOWN_LIST_PATTERN = Pattern.compile("(?<!\\S)- ");
private void handleHtml(TagElement node) {
		String text2 = this.tm.toString(node);
		Matcher matcher2 = CUSTOM_MARKDOWN_LIST_PATTERN.matcher(text2);
		boolean first = true;
		while (matcher2.find()) {
			int startPos = matcher2.start() + node.getStartPosition();
			int endPos = matcher2.end() - 1 + node.getStartPosition();
			int firstTokenIndex = tokenStartingAt(startPos);
			int lastTokenIndex = tokenEndingAt(endPos);
			Token firstPartToken;
			if (first) {
				 firstPartToken = this.ctm.get(firstTokenIndex);
				 first=false;
			} else {
				 firstPartToken = this.ctm.get(firstTokenIndex);
				 firstPartToken.breakBefore();
			}

			firstPartToken.setWrapPolicy(null);
			firstPartToken.spaceBefore();
			firstPartToken.spaceAfter();
			}
		}
}
```

Once this PR is merged I can raise this as a separate issue/Pr for recognising and formatting markdown tags

mateusz-matela (approved):

mateusz-matela:
> Once this PR is merged I can raise this as a separate issue/Pr for recognising and formatting markdown tags

Yep, I think you're on the right track.

SougandhS:
> > Once this PR is merged I can raise this as a separate issue/Pr for recognising and formatting markdown tags
> 
> Yep, I think you're on the right track.

Thanks! I'll merge this once master branch opens 👍

SougandhS:
@mateusz-matela since https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4336/ merged
I made a change enable/disable formatting of markdowns when javadoc is enabled. Please check the new commit when you have time

https://github.com/user-attachments/assets/2a81f420-d8f8-4d01-bec6-8a42b930af56

SougandhS:
> @mateusz-matela since #4336 merged I made a change enable/disable formatting of markdowns when javadoc is enabled. Please check the new commit when you have time
> 
>  MarkdownEnable.mp4

Do we also include "Blank lines btwn tags" and other formatting sub options that are similar to _Classic JavaDoc_ for _Markdown formatter_ in UI or should we keep as it is -> Classic JavaDoc formatting options will be applicable to Markdown comments when _Markdown Formatting_ is enabled ?

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/CommentsPreparator.java
```diff
@@ -1283,25 +1289,26 @@ private boolean tokenizeMultilineComment(Token commentToken) {
...
-		boolean newLinesAtBoundries = commentToken.tokenType == TokenNameCOMMENT_JAVADOC
-				? this.options.comment_new_lines_at_javadoc_boundaries
-				: this.options.comment_new_lines_at_block_boundaries;
-		if (!newLinesAtBoundries) {
-			structure.get(1).clearLineBreaksBefore();
-			last.clearLineBreaksBefore();
-		} else if (this.tm.countLineBreaksBetween(first, last) > 0) {
-			first.breakAfter();
-			last.breakBefore();
+		if(!isMarkdown) {
```
mateusz-matela: missing space

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/linewrap/CommentWrapExecutor.java
```diff
@@ -70,9 +72,14 @@ public int wrapMultiLineComment(Token commentToken, int startPosition, boolean s
 		this.simulation = simulate;
 		this.wrapDisabled = noWrap;
 		this.potentialWrapToken = this.potentialWrapTokenSubstitute = null;
-		this.newLinesAtBoundries = commentToken.tokenType == TokenNameCOMMENT_JAVADOC
-				? this.options.comment_new_lines_at_javadoc_boundaries
-				: this.options.comment_new_lines_at_block_boundaries;
+
+		if(commentToken.tokenType != TokenNameCOMMENT_MARKDOWN) {
```
mateusz-matela: missing space

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
```diff
@@ -122,12 +125,11 @@ protected boolean token(Token token, int index) {
 
 		bufferWhitespaceBefore(token, index);
 
-		if (token.tokenType == TokenNameCOMMENT_MARKDOWN) {
+		if (token.tokenType == TokenNameCOMMENT_MARKDOWN && !this.options.comment_format_markdown_comment) {
```
mateusz-matela: this is not a good place to disable markdown processing (this would perform all the unneeded processing and only ignore the results at the very end), better change `formattingEnabled` in `CommentsPreparator.visit(Javadoc)`

mateusz-matela (changes requested):

mateusz-matela:
> Do we also include "Blank lines btwn tags" and other formatting sub options that are similar to _Classic JavaDoc_ for _Markdown formatter_ in UI or should we keep as it is -> Classic JavaDoc formatting options will be applicable to Markdown comments when _Markdown Formatting_ is enabled ?

IMO it makes more sense to use the same sub options for both classic and markdown. We can change this approach some day if a good argument is presented to separate them.

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/linewrap/CommentWrapExecutor.java
SougandhS: added

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
SougandhS: Updated 👍 

Edit: and tested too

SougandhS:
> IMO it makes more sense to use the same sub options for both classic and markdown. We can change this approach some day if a good argument is presented to separate them.

Should we provide any label's or descriptions for that ?  like adding something like applicable for Markdowns or something ?

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/CommentsPreparator.java
```diff
@@ -597,6 +597,9 @@ public boolean visit(Javadoc node) {
 			// not a proper javadoc, treat as block comment
 			handleWhitespaceAround(commentIndex);
 		}
+		if(node.isMarkdown() && !this.options.comment_format_markdown_comment) {
```
mateusz-matela: missing space...

but why not change below like this instead:
`boolean formattingEnabled = isHeader ? this.options.comment_format_header : node.isMarkdown() ? this.options.comment_format_markdown_comment : this.options.comment_format_javadoc_comment`
this would make the "header comment formatting" logic for markdowns consistent with how it worked for classic
unless invoking `commentToken.setInternalStructure()` will break something?

mateusz-matela:
> Should we provide any label's or descriptions for that ? like adding something like applicable for Markdowns or something ?

Sounds good, but if most items will be applicable to markdowns, then maybe mark with ¹ those applicable only to classic javadoc

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/CommentsPreparator.java
SougandhS: >missing space...

image

I have this option enabled yet it fails somehow, for my other workspaces it is getting formatted without this `Perform the selected actions` enabled-> Any sub option I have to enable other than this or any fix ?

SougandhS: > but why not change below like this instead:

I will check this

SougandhS: >but why not change below like this instead:

This works perfect thanks!

https://github.com/user-attachments/assets/9c864a0b-76d2-4675-ba11-2d06ecb58db4

SougandhS:
> then maybe mark with ¹ those applicable only to classic javadoc

Great!, I'll raise ISSUE/PR for that in UI

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/CommentsPreparator.java
mateusz-matela: > I have this option enabled yet it fails somehow, for my other workspaces it is getting formatted without this `Perform the selected actions` enabled-> Any sub option I have to enable other than this or any fix ?

If this is workspace configuration, it's probably overridden in project configuration.

SougandhS:
> Looks good. (I assume you'll squash the commits into one)

Yep!
Merging this now

SougandhS:
Build failures are unrelated to this PR

SougandhS:
Thank you @mateusz-matela :)

PR #4248, opened 2025-07-25
Fix relevance for completion of receiver of varargs method invocation

### LINKED ISSUE #4247: Erroneous "expected type" when completing receiver of message send
While debugging through `CompletionTests.testBug574982`, I noticed that `java.lang.Object` is being added as an expected type, which affects relevance number calculations.

Here is the code from the test. with `|` inserted where completion is being calculated:

```java
public class ArrayTest {
  public void test() {
    new Runnable() {
      public void run() {
        boolean equals = A|rrays.equals(new Object[0], new Object[0]);
      }
    };
  }
}
```

I think the expected type should either be `boolean` or there should be no expected type.

The root cause is that `CompletionEngine.checkForVarargExpectedTypes` doesn't check to ensure that completion is being performed in the arguments.

datho7561:
## What it does
Fixes #4247 by fixing the bug in `CompletionEngine.checkForVarargExpectedTypes`

## How to test
There's an existing test case that I've updated

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
+++ b/org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
@@ -3887,7 +3887,7 @@ public final class CompletionEngine
 			checkCancel();
 
 			// see if we can find argument type at position in case if we are at a vararg.
-			checkForVarargExpectedTypes(astNodeParent, scope);
+			checkForVarargExpectedTypes(astNodeParent, astNode, scope);
 			findVariablesAndMethods(this.completionToken, scope, singleNameReference, scope, insideTypeAnnotation,
 					singleNameReference.isInsideAnnotationAttribute, true, new ObjectVector());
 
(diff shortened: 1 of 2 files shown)


datho7561:
@jjohnstn when you have time, do mind reviewing this PR?

jjohnstn:
@datho7561 Will do

org.eclipse.jdt.core/codeassist/org/eclipse/jdt/internal/codeassist/CompletionEngine.java
```diff
@@ -3924,7 +3924,9 @@ private void completionOnSingleNameReference(ASTNode astNode, ASTNode astNodePar
 	}
 
 	private void checkForVarargExpectedTypes(ASTNode astNodeParent, Scope scope) {
-		if (astNodeParent instanceof MessageSend m && this.expectedTypesPtr == -1) {
+		if (astNodeParent instanceof MessageSend m
+				&& !(m.receiver instanceof CompletionOnSingleNameReference)
```
jjohnstn: According to the issue description, you say that checkForVarargExpectedTypes isn't checking that the completion is in the arguments.  Shouldn't this test be more specific to ensure we are in the arguments instead of just not CompletionOnSingleNameReference?

datho7561: This method in only called when completing a single name reference, but I made the solution a bit more general

jjohnstn:
@datho7561 Can you check the completion test failures?

datho7561:
I added a null check, which fixed the test failure locally for me.

jjohnstn (approved):
LGTM

Issue #4236, opened 2025-07-22
Regression in generics type inferral


alienisty:
The following fails to compile on Eclipse Java Development Tools 3.20.300.v20250721-2114:
```java
import java.io.Serializable;

public record A(DR<? extends TI<?>> effective) {

  private A(DR<? extends TI<?>> effective,
            Object x) {
    this(effective);
  }
}

class DR<T extends Comparable<? super T> & Serializable> {
}

abstract class TI<M> implements Comparable<TI<M>>, Serializable {
  private static final long serialVersionUID = 1L;
}
```
complaining that `The constructor A(DR<capture#1-of ? extends TI<?>>) is undefined`.

The same class compiles fine using the JDK 17.0.15+6 and Eclipse Java Development Tools 3.20.300.v20250714-2102

srikanth-sankaran:
Thanks for following up @stephan-herrmann

PR #4251, opened 2025-07-26
Regression in generics type inferral

stephan-herrmann:
Capture in more situations:
+ register context for capturing using new class CapturingContext
  + consolidate with previous IC18.maybeCapture()
+ enter/leave from:
  + PTB.capture() - context for glb-computation during initializeBounds
  + Scope.parameterCompatibilityLevel() - context for isCompatibleWith()

Apply normalization like #4041 also during wildcard creation:
  + normalize "? extends cap-of ? extends X" to "? extends X"

Measures to avoid infinite recursion:
+ guard with CapturingContext.isCaptureInProgress
+ don't expand types that already have deeper structure than otherType
  + impl. as typeArgumentDepth(), use in findSuperTypeOriginatingFrom()

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4236

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding.java
@@ -598,6 +598,13 @@ public class CaptureBinding extends TypeVariableBinding {
 	}
 
 	@Override
+	public int typeArgumentDepth() {
+		if (this.wildcard != null)
+			return this.wildcard.typeArgumentDepth();
+		return 1;
+	}
+
+	@Override
 	public String toString() {
 		if (this.wildcard != null) {
 			StringBuilder buffer = new StringBuilder(10);
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CapturingContext.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CapturingContext.java
@@ -1,0 +1,67 @@
+/*******************************************************************************
+ * Copyright (c) 2025 GK Software, and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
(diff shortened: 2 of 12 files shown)


stephan-herrmann:
Some background:

The situation is a regression from #4116 where initialization of capture bounds was made to apply glb. This interacts badly with #3910 which applies capture also during `findSuperTypeOriginatingFrom()`. This method is used from glb-computation as well as from `isTypeBindingContainedBy()`. This led to inconsistent results, depending on whether or not an `InferenceContext18` was available during `maybeCapture()`. When `isTypeBindingContainedBy()` is invoked below `isCompatibleWith()` the result is actually cached, making results depend on who first asks `isCompatibleWith()` and in which context.

So the core problem can be described as inconsistent availability of a context for capturing, in situations where according to JLS capture should indeed be applied.

To move forward I generalized `InferenceContext18.maybeCapture()` into a new class `CapturingContext`. Using this class, more locations invoke `enter()` / `leave()` to manage the context for capturing in more situations. I don't know if the list of participating locations is complete, more invocations of `enter()` / `leave()` may later be added based on future bug reports.

This entailed a series of stack overflows, due to recursive invocation of `maybeCapture()`. An additional mechanism was implemented that detects when repeated capture would create infinite types. This situation is detected by computing the depth of nested type arguments.

Finally, a normalization like #4041 is now applied also during wildcard creation: normalize `? extends cap-of ? extends X` to `? extends X`.

stephan-herrmann:
@srikanth-sankaran FYI:

Introduction of `CapturingContext` with a protocol of `enter()` / `leave()` solves the following architectural dilemma:
* All locations interested in supertypes of a PTB with wildcards need to apply capture (JLS 4.10.2.).
* Interested methods like `isCompatibleWith()` and `findSupertypeOriginatingFrom()` etc. do not have the necessary data (scope & source positions) for invoking `capture()`

Let me know if you have a better idea how to solve this architecturally.
* If we keep this strategy: do we have better / more central locations that should invoke enter/leave? Or is it OK to lazily add more enter/leave calls when needed?
* Would we be better off with computing captures more eagerly (stored in the PTB) while necessary context may still be available (is it?) ?

Do you have a spontaneous answer? If not, just drop a note, in which case I'll go ahead for now, and file a follow-up issue for discussion in a time of leisure :)

srikanth-sankaran:
> file a follow-up issue for discussion

I prefer this. Thanks.

stephan-herrmann:
> > file a follow-up issue for discussion
> 
> I prefer this. Thanks.

See #4252

Issue #4254, opened 2025-07-26
[25] JEP 507 handling of preview flag


stephan-herrmann:
Retesting ecj against https://bugs.openjdk.org/browse/JDK-8348410 I detected a few flaws regarding the preview flag:
* after compiling the test case from JDK-8348410 using `--enable-preview` the program can be run without `--enable-preview`, which is wrong
* several uses of the preview feature don't raise a preview warning.

PR #4255, opened 2025-07-26
[25] JEP 507 handling of preview flag

stephan-herrmann:
+ set preview flag in classfile for case from JDK-8348410
+ raise preview warning in more cases

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4254

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/CaseStatement.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/CaseStatement.java
@@ -187,7 +187,7 @@ private Constant resolvePatternLabel(BlockScope scope, TypeBinding caseType, Typ
 		if (!pattern.isApplicable(selectorType, scope, pattern))
 			return Constant.NotAConstant;
 	} else if (caseType.isValidBinding()) { // already complained if invalid
-		if (Pattern.findPrimitiveConversionRoute(caseType, selectorType, scope) == PrimitiveConversionRoute.NO_CONVERSION_ROUTE) {
+		if (pattern.findPrimitiveConversionRoute(caseType, selectorType, scope) == PrimitiveConversionRoute.NO_CONVERSION_ROUTE) {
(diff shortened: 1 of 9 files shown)


stephan-herrmann:
@mpalat could you please glance over this PR? I was surprised, how incomplete handling of the flag was for this feature, or am I missing any deliberate decision? Hopefully you still remember working on this code :)

I believe our strategy should be that `CompilationResult.previewUsed` should never be assigned without raising a warning, right? (this was not established when I introduced that flag in #3604 :hushed:). With this change we indeed assign this field only from methods of `ProblemReporter`.

stephan-herrmann:
> @mpalat could you please glance over this PR? I was surprised, how incomplete handling of the flag was for this feature, or am I missing any deliberate decision? Hopefully you still remember working on this code :)
> 
> I believe our strategy should be that `CompilationResult.previewUsed` should never be assigned without raising a warning, right? (this was not established when I introduced that flag in #3604 😯). With this change we indeed assign this field only from methods of `ProblemReporter`.

@mpalat did you see my request? Do you need more time? Should I proceed on my own?

mpalat (approved):
+1

> > @mpalat could you please glance over this PR? I was surprised, how incomplete handling of the flag was for this feature, or am I missing any deliberate decision? Hopefully you still remember working on this code :)
> > I believe our strategy should be that `CompilationResult.previewUsed` should never be assigned without raising a warning, right? (this was not established when I introduced that flag in #3604 😯). With this change we indeed assign this field only from methods of `ProblemReporter`.
> 
> @mpalat did you see my request? Do you need more time? Should I proceed on my own?

Sorry @stephan-herrmann  for the delayed reply. Got pulled into something internally although that's should not be an excuse for the delay :( . And to answer your questions, yes, in the light of JDK-8348410, we are not complete in flagging this. and yes, I remember the code. Agree that `CompilationResult.previewUsed` should be assigned only with a warning.

stephan-herrmann:
> @mpalat could you please glance over this PR?

thanks.

FYI: I would have squashed the commits when merging. Any particular reasons why you did not? As a rule of thumb I will indicate in a PR when I believe individual commits are worth preserving. In that case I would also make each commit comment self-contained (i.e., indicate the issue being resolved). By default I would squash, and likely also give a cleaned-up comment for the squashed commit.

Issue #4259, opened 2025-07-28
Add Java 25 constants


trancexpress:
We are starting to test with OpenJDK 25 early access, one immediate problem is having JDT errors when specifying `JavaSE-25` as a `.classpath` JRE. We would like to add constants for Java 25, so that we can continue testing.

See e.g.: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2345

stephan-herrmann:
> We are starting to test with OpenJDK 25 early access, one immediate problem is having JDT errors when specifying `JavaSE-25` as a `.classpath` JRE. We would like to add constants for Java 25, so that we can continue testing.

Why can't you use branch BETA_JAVA25 or https://download.eclipse.org/jdt/updates/4.37-P-builds which has all these constants readily available? You'd help us much more if you would test that, rather than creating conflicts between branches.

trancexpress:
> Why can't you use branch BETA_JAVA25 or https://download.eclipse.org/jdt/updates/4.37-P-builds which has all these constants readily available? You'd help us much more if you would test that, rather than creating conflicts between branches.

I assume it wont be so simple to use another branch in our build process - we test with latest master of SDK bits every week.

We can likely test with the 25 beta branch as a one time thing, when we are through with the basics (no test fails when running with Java 25).

Or, @iloveeclipse , do we put effort in using the beta branch for our builds, whenever that makes sense (such as now, but not before/after several months)?

Otherwise, sorry if we caused merge conflicts.

iloveeclipse:
The point is currently not in testing any new Java 25 feature and compiler implementation, but just enable 4.36 to use Java 25 **runtime** as valid Java runtime for running. That has value for everyone who will use 4.36 with Java 25. Currently it is simply not possible, which is a blocker...

Regarding the merge issues: probably it would be better to merge such changes at the very beginning on master, **before** BETA_JAVA branch is created.

Regarding the BETA_JAVA25 content testing: right now we are interested in sorting out and fixing new **runtime** regressions of Java 25. Mixing that with new compiler implementation would be contraproductive.

Once we are "green" with runtime, we could consider switching the branch, however if we would do so, we would need weekly merges  from master on all branched repos to BETA_JAVA25, otherwise we would miss to test "regular" regressions introduced on master branches (there are enough).

PR #4260, opened 2025-07-28
Add Java 25 constants

trancexpress:
Fixes: #4259

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/classfmt/ClassFileConstants.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/classfmt/ClassFileConstants.java
@@ -141,6 +141,7 @@ public interface ClassFileConstants {
 	int MAJOR_VERSION_22 = 66;
 	int MAJOR_VERSION_23 = 67;
 	int MAJOR_VERSION_24 = 68;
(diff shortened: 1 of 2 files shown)


laeubi:
Should it not be added to the supported versions?

grafik

trancexpress:
> Should it not be added to the supported versions?

I can't say... Java 24 is there so I think yes?

I've set a PR in case its required:

https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4267

@stephan-herrmann is it expected in the list?

iloveeclipse:
> > Should it not be added to the supported versions?
> 
> I can't say... Java 24 is there so I think yes?
> 
> I've set a PR in case its required:
> 
> #4267
> 
> @stephan-herrmann is it expected in the list?

It is not expected yet, because compiler supports it only **after** 4.37 release.

stephan-herrmann:
Are you guys talking about 24 or 25? Maybe you should start adding [26](https://jdk.java.net/26/)?

laeubi:
> It is not expected yet, because compiler supports it only after 4.37 release.

I then really question to add the constant here at all. If the compiler does not support it (or is not allowed to support it) then adding a constant *at compiler code* seems wrong an we would more need a dedicated list of "Available Runtime Java Versions" but outside the batch compiler and as @stephan-herrmann correctly pointed out then adding Java 26 there as well.

Issue #4231, opened 2025-07-20
When returntype is Object and class extends another type one gets error "The type X cannot be resolved. It is indirectly referenced from required type"


laeubi:
## Problem

I today found a case where I think ECJ can avoid emitting `The type X cannot be resolved. It is indirectly referenced from required type `

Take this example project:

https://github.com/eclipse-equinox/equinox/tree/master/bundles/org.eclipse.equinox.jsp.jasper.registry

now remove from the manifest these imports:
https://github.com/eclipse-equinox/equinox/blob/2fd1eb7b348f3a3925b60f0c33dafeb6feae4b9a/bundles/org.eclipse.equinox.jsp.jasper.registry/META-INF/MANIFEST.MF#L13-L14

And you will get this error then

```
Description	Resource	Path	Location	Type
The type javax.servlet.http.HttpServlet cannot be resolved. It is indirectly referenced from required type org.eclipse.equinox.jsp.jasper.JspServlet	JSPFactory.java	/org.eclipse.equinox.jsp.jasper.registry/src/org/eclipse/equinox/jsp/jasper/registry	line 1	Java Problem
```

One would argue that this is unavoidable because `JspServlet` extends `HttpServlet` so the compiler must know the type., but interestingly if you change this line

https://github.com/eclipse-equinox/equinox/blob/2fd1eb7b348f3a3925b60f0c33dafeb6feae4b9a/bundles/org.eclipse.equinox.jsp.jasper.registry/src/org/eclipse/equinox/jsp/jasper/registry/JSPFactory.java#L58

from `public Object create()` to `public JspServlet create()` it works! This shows another issue here, the error is actually shown at the top of the file but the error obviously is related to the return type of the method, so I would have expected it to be shown at line 58!

## Current behavior

As far as I understand the reason for this is that ECJ wants to check that the return type is compatible, so it needs to check `JspServlet`, all its super classes and implemented interfaces and then discovers that it can not see the extended type `HttpServlet`.

## Expected behavior

As every class must extend `Object` there is no need to further check if return types are compatible. It would even valuable to first check if the Type itself already implements any compatible class or interface already, but I do not know if that maybe already happens.

## Impact

Currently users are forced to import unused packages. Also when writing code one sometimes hit this issue and it is very annoying to then need to find out what to do to make the compiler happy, and even PDEs Quick-fix currently can not help here as there is no trace anywhere that `javax.servlet.http` is needed at all.

So summary:
[...]

PR #4285, opened 2025-08-03
Do not ask for unchecked conversion if return type is Object

laeubi:
Currently ReturnStatement.resolve(BlockScope) checks if the type requires an unchecked conversion, it can trigger an errors when the type is not found, but in case of the return type is object what can always be used this check is undesired.

This now first check if the return type is object and then omit the check for unchecked conversion to prevent this situation.

Fix https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4231

This would then help in fixing the problem observed here:
- https://github.com/eclipse-equinox/equinox/pull/1084#issuecomment-3093458618

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ReturnStatement.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ReturnStatement.java
@@ -387,7 +387,7 @@ public void resolve(BlockScope scope) {
 			|| expressionType.isCompatibleWith(methodType, scope)) {
 
 		this.expression.computeConversion(scope, methodType, expressionType);
-		if (expressionType.needsUncheckedConversion(methodType)) {
+		if (!scope.getJavaLangObject().equals(methodType) && expressionType.needsUncheckedConversion(methodType)) {
 		    scope.problemReporter().unsafeTypeConversion(this.expression, expressionType, methodType);
 		}
 		if (this.expression instanceof CastExpression) {


stephan-herrmann (changes requested):
First we need a test case.

Second, "unchecked conversion" warnings are specified in JLS (https://docs.oracle.com/javase/specs/jls/se24/html/jls-5.html#jls-5.1.9), so instead of saying "undesired" we need to check what JLS has to say about that test case (s.above).

Third, we should not opportunistically modify type checking for one particular construct (here: `return`), but if a bug exists fix it globally.

laeubi:
> Third, we should not opportunistically modify type checking for one particular construct (here: `return`), but if a bug exists fix it globally.

I tried to be as local as possible, and I also don't think it is a bug per-se it is more an improvement where ecj can avoid lookup the type (so probably call it performance optimization).

Another reason was I'm not completely sure how to access the scope from `expressionType.needsUncheckedConversion` method, I could try adding it as a parameter but wanted to get some recommendation from the JDT team first.

laeubi:
> Second, "unchecked conversion" warnings are specified in JLS (https://docs.oracle.com/javase/specs/jls/se24/html/jls-5.html#jls-5.1.9), so instead of saying "undesired" we need to check what JLS has to say about that test case (s.above).

If I understand correctly then if `G` is parameterized type then unchecked conversion must be issued when I do something like `G<String> = new G()` for example, so I wonder if the check should only ever be performed if the target type `T` (in this case `Object`) is actually parameterizable?

laeubi:
@stephan-herrmann I now tried this aproach I hope my understanding of the JLS and the JDT code is correct:

- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4286

stephan-herrmann:
> > Second, "unchecked conversion" warnings are specified in JLS (https://docs.oracle.com/javase/specs/jls/se24/html/jls-5.html#jls-5.1.9), so instead of saying "undesired" we need to check what JLS has to say about that test case (s.above).
> 
> If I understand correctly then if `G` is parameterized type ...

replace "parameterized" with "generic"

> ... then unchecked conversion must be issued when I do something like `G<String> = new G()` for example, so I wonder if the check should only ever be performed if the target type `T` (in this case `Object`) is actually parameterizable?

Even stronger: the target type (LHS) must be _parameterized_, i.e., it must have angle brackets :)

So `G g = new G()` does not involve unchecked conversion even if G declares type parameters ("is a generic type").

* "generic" = declares type parameters 
* "parameterized" = instantiates type parameters declared by the generic type with concrete types.

stephan-herrmann:
Closing in favor of #4286

PR #4286, opened 2025-08-03
Do not check for unchecked conversion if target is not parameterized

laeubi:
Currently TypeBinding.needsUncheckedConversion(TypeBinding) try to findSuperTypeOriginatingFrom(..) the target type and then check if compatible is any raw type and the target type is a parameterized type. This can only ever result in a true expression if the target type or any of its enclosing types is a bound parameterized type.

Before lookup a super type of the current type binding, this now first checks if the target actually has any bound parameterized type and prevents the lookup if it will always return false anyways.

Fix https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4231

## Author checklist


laeubi:
I'm still not sure how to properly test this. While it seems quite easy that an error is produced, the checking for its absence is quite hard to assert given the error itself only is producible in special circumstances.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeBinding.java
```diff
@@ -1559,7 +1559,9 @@ public boolean needsUncheckedConversion(TypeBinding targetType) {
 	targetType = targetType.leafComponentType();
 	if (!(targetType instanceof ReferenceBinding))
 		return false;
-
+	if (!hasBoundParameterizedTypeInHierarchy(targetType)) {
```
stephan-herrmann: Looking at JLS I think it suffices to change the previous check to
```java
if (!(targetType instanceof ParameterizedTypeBinding))
    return false;
```

stephan-herrmann:
> I'm still not sure how to properly test this. While it seems quite easy that an error is produced, the checking for its absence is quite hard to assert given the error itself only is producible in special circumstances.

Have you looked at ProblemTypeAndMethodTest? It has tests that quite clearly show how to trigger the error in question (classes hidden from the build path are simulated by deletion after compiling intermediate classes). Just copy one of them and insert your own classes, and you should be able to create a test that fails on HEAD and succeeds with your change.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeBinding.java
laeubi: I have done this but I'm not completely sure here because `isBoundParameterizedType` only return false if it is `Type<?>`, so I'm not completely sure this is covered here as well.

But maybe then `targetType.isBoundParameterizedType()` would be enough as an extra check? But then one would wonder why further down in the method it is even checked again for enclosing types?

laeubi:
> Have you looked at ProblemTypeAndMethodTest? It has tests that quite clearly show how to trigger the error in question (classes hidden from the build path are simulated by deletion after compiling intermediate classes).

Its not always easy to make sense of a 10.000 + LOC test, but the hint with the deletion leads me on the right track and I was able to add a testcase now!

laeubi:
While I found that these cases work in the IDE with this patch they don't work in the testcase

Fails with `java.lang.NoClassDefFoundError: p1/I`:
```
public P test3() {
 return new P(1);
}
```

what seems expected as we have deleted that file

Fails because it looks like `var`	is not detected as a keyword (maybe it uses java 1.8 for the test?)
		
```
public Object test4() {
 var p = new P(1);
 return p;
}
```

So I have omitted these both, but if someone has a hint how to enable these it would of course be good to have them, but in general this PR is ready to be merged as it solves the issue I observed.

laeubi:
@stephan-herrmann would you be fine with merging this and possibly improve this later on (test, other cases to discover...). I would like to see this in action :-)

laeubi:
And it succeeds now!

https://ci.eclipse.org/releng/job/equinox/job/PR-1084/17/

stephan-herrmann:
> While I found that these cases work in the IDE with this patch they don't work in the testcase
> 
> Fails with `java.lang.NoClassDefFoundError: p1/I`:
> 
> ```
> public P test3() {
>  return new P(1);
> }
> ```
> 
> what seems expected as we have deleted that file
> 
> Fails because it looks like `var` is not detected as a keyword (maybe it uses java 1.8 for the test?)

Compiler tests typically run in a loop over suitable compliance settings. The available set of versions is controlled via system property `compliance` (see how this is defined in pom.xml). Additionally, a test that requires a specific language construct has two options to avoid running at inappropriate levels:
* per class:
construct the `suite()` using `buildMinimalComplianceTestSuite(testClass(), X` where X is the minimal compliance
* per test:
let individual tests be skipped at inappropriate levels: `if (this.compliance < ClassFileConstants.JDK_X_) return;`
 
> ```
> public Object test4() {
>  var p = new P(1);
>  return p;
> }
> ```


_If_ the special identifier `var` is detected, we need to infer the type of that local variable. I don't think it's justified to expect that this inference succeeds without inspecting the details of P (incl. supertypes).

laeubi:
> If the special identifier var is detected, we need to infer the type of that local variable. I don't think it's justified to expect that this inference succeeds without inspecting the details of P (incl. supertypes).

I first though that, but

1. It works in the IDE
2. The inferred type is trivially of type `P`, I don't see any further inspection of the type hierarchy is required.
3. Any value is assignable to `var`
4. So we have the same situation that anything can be returned here

But maybe it would be better to look into it when we actually have a case where it makes problems.

stephan-herrmann:
> > If the special identifier var is detected, we need to infer the type of that local variable. I don't think it's justified to expect that this inference succeeds without inspecting the details of P (incl. supertypes).
> 
> I first though that, but
> 
> 1. It works in the IDE
> 2. The inferred type is trivially of type `P`, I don't see any further inspection of the type hierarchy is required.
> 3. Any value is assignable to `var`
> 4. So we have the same situation that anything can be returned here
> 
> But maybe it would be better to look into it when we actually have a case where it makes problems.

My main point is: hardly anything in Java is "trivial", unless you carefully checked JLS.

Indeed, https://docs.oracle.com/javase/specs/jls/se24/html/jls-4.html#jls-4.10.5 has this: "If T does not mention any restricted type variable, then the result is T." So under certain conditions this inference ("projection") is trivial, but we only know _after_ consulting JLS, not before :)

Issue #813, opened 2023-03-08
False positive "The static method X from the type Y should be accessed directly"


laeubi:
Assume the following code: [test.access.staticprotected.zip](https://github.com/eclipse-jdt/eclipse.jdt.core/files/10888888/test.access.staticprotected.zip)

We now change the class like this and get a warning:
![grafik](https://user-images.githubusercontent.com/1331477/223666148-bd73cdf2-e4b1-4040-8eb5-f4e282e9eee9.png)

This is a false-positive because the recommended "direct" access is actually not accessible:
![grafik](https://user-images.githubusercontent.com/1331477/223666570-58f44d70-dca0-42ec-a8ef-6e3ca4891fd0.png)

stephan-herrmann:
While wondering, whether or not that warning should be ditched completely, I found that SonarLint has a similar warning, see, e.g., https://stackoverflow.com/questions/71754314/sonarlint-rule-javas3252-static-base-class-members-should-not-be-accessed-vi

Some reasons to keep this warning can be found in scenarios of code evolution, e.g., when adding a same-named static member to the derived type, which of course does not **override** the inherited member (there is no overriding for statics), but still the client will only be binary compatible, but not source compatible (recompiling will change the result).

Actually, in "normal" use, there is no benefit ever in using a subclass for referencing an inherited static member. So, I believe this pattern will always create some confusion to the reader. Furthermore I've seen cases, where such reference creates a completely unnecessary dependency on the sub class.

On the other hand I'm not sure it's worth adding more complexity to this analysis.

Maybe the following compromise would fit:
* jdt.core: rephrase the warning to: "Static method X from type Y is accessed indirectly via type Z" -- _i.e., refrain from suggesting what "should" happen_
* jdt.ui: perform a visibility check before proposing this particular quick fix, alternatively ditch that quick fix.

laeubi:
@stephan-herrmann the problem is not the warning itself, it is just that in this case the user can't follow the advice without creating invalid code.

So the filter should happen before generating the warning, if the "direct" type is not visible (and with the second screenshot you see that JDT actually knows this in some way already) the warning should be skipped (and then the quick-fix problem is solved as well), but I think strange enough the Quickfix already knows because it only offers to  **remove the static**, but it should actually recommend to use the subtype!

stephan-herrmann:
> @stephan-herrmann the problem is not the warning itself, it is just that in this case the user can't follow the advice without creating invalid code.

[...]

PR #4293, opened 2025-08-07
Do not emit restriction warning on inherited method

laeubi:
Currently when we have a type `A` and it is accessible from package `p1`, but extends a type `B` (that is not accessible in package `p2`) calling any inherited method `A.M` triggers a restriction warning.

This is undesired because as a caller I never access any restricted code here as `A` is inherit all public methods from `B`.

A good example can be found here:
- https://github.com/eclipse-equinox/equinox/pull/1080

in this example the code in `HttpServerManager` has a local variable of type `Server` and calls method `stop()` on this.

```
Server server = servers.remove(pid);
if (server != null) {
	try {
		server.stop();
	} catch (Exception e) {
	}
	File contextWorkDir = new File(workDir, DIR_PREFIX + pid.hashCode());
	deleteDirectory(contextWorkDir);
}
```

this causes the following errors:

```
Description	Resource	Path	Location	Type
Access restriction: The method 'AbstractLifeCycle.stop()' is not API (restriction on required library 'org.eclipse.jetty.util_12.0.23.jar')	HttpServerManager.java	/org.eclipse.equinox.http.jetty/src/org/eclipse/equinox/http/jetty/internal	line 77	Java Problem
```

it does not emit such warnings if the method itself is overridden (an example would be `server.dump(null, "")` that overrides a method in `ContainerLifeCycle` from the same package) because then the method binding equals the actual binding.

To mitigate this, the class `MessageSend` now checks if there is an implicit call and sets `isExplicitUse=false` like it is already done for implicit super constructor calls.

@stephan-herrmann What do you think? Any flaws in my analysis?

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/MessageSend.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/MessageSend.java
@@ -71,6 +71,7 @@ import static org.eclipse.jdt.internal.compiler.ast.ExpressionContext.VANILLA_CO
 
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 import java.util.function.BiConsumer;
 import org.eclipse.jdt.core.compiler.CharOperation;
 import org.eclipse.jdt.internal.compiler.ASTVisitor;
@@ -1002,7 +1003,10 @@ public TypeBinding resolveType(BlockScope scope) {
 		}
 		// abstract private methods cannot occur nor abstract static............
 	}
-	if (isMethodUseDeprecated(this.binding, scope, true, this))
+	TypeBinding declared = this.binding.declaringClass.erasure();
+	TypeBinding actual = this.actualReceiverType.erasure();
+	boolean isExplicitUse = Objects.equals( declared, actual);
+	if (isMethodUseDeprecated(this.binding, scope, isExplicitUse, this))
 		scope.problemReporter().deprecatedMethod(this.binding, this);
 
 	TypeBinding returnType;


stephan-herrmann:
Short answer: I'm not convinced. 

If you look in package p1, there is no such method, right?

If p2 is an internal package, then the owner of that package is allowed to do things like removing `public` from that method, isn't he?

For such reasons, exporting a class that inherits from an internal class is inherently fragile / breaks encapsulation. And if p2 is not on the clients classpath they will get "The type B cannot be resolved. It is indirectly referenced  ..."

laeubi:
> If you look in package p1, there is no such method, right?

Outline tells me something different, also javadoc, code complete,... all tell me `Server` has a public (inherited) method `stop()` :-)

> If p2 is an internal package, then the owner of that package is allowed to do things like removing public from that method, isn't he?

Given this class is extended by legal consumers this will likely break them as the probably calling or overriding the method, so I don't think it is allowed. Also I don't think a restricted package is necessarily internal. If any public code is referencing this in its API it at least (e.g. extend an abstract class with public methods) leaking that (internal) API to the outside.

> For such reasons, exporting a class that inherits from an internal class is inherently fragile / breaks encapsulation

That might be the case, but I don't think that restrictions really make such strong assumption, it just tells you should not directly reference the class.

So now we assume your case of someone really really really wants to remove method `X` from abstract class `A` and I have restricted access to `A` and have legal extenders of `B` and `C`, then I still can do that by:

1. removing `X` from `A`
2. let `B` + `C` implement the method `X` with the old signature

that's why only a direct reference to `A` is problematic but not through public types `B` or `C` ...

stephan-herrmann:
> > If you look in package p1, there is no such method, right?
> 
> Outline tells me something different, also javadoc, code complete,... all tell me `Server` has a public (inherited) method `stop()` :-)

None of them speak of showing the contents of p1. They know about that method because they look into p2, despite the restriction.

Anyway, I have more urgent tasks waiting for me than bending the rules of classpath restrictions.

laeubi:
>  And if p2 is not on the clients classpath they will get "The type B cannot be resolved. It is indirectly referenced ..."

But that's why we _want_ it on the classpath and instead using access rules right?

> None of them speak of showing the contents of p1. They know about that method because they look into p2, despite the restriction.

I must confess I never have heard about a _method_ belong to a _package_ and I would have not even bothered to ask or change anything if not even the compiler seem to emit a call to `p1.A.X()` and not to `p2.B.X()`.
Even API tools seem to complain if I remove any inherited method from a public type regardless of it being internal or not (it just complains if you leak an non-API type from a API type what is of course desired).

And while it makes perfectly sense for a deprecated method to be checked "at the source" I don't get the rationale for an access restriction and even the code seem to indicate with a `TODO` that actually there is something to improve rather than to bend here https://github.com/eclipse-jdt/eclipse.jdt.core/blob/95b6e3e444b0675b954a4cecc54849486112f35e/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ASTNode.java#L529-L534

So I went to the code and checked the callers of the method and found this place, as the exact same is used for implicit constructor calls in `ExplicitConstructorCall` where it decides depending on the context if such warning should be emitted.

Then I looked at the [compiler documentation here](https://help.eclipse.org/latest/topic/org.eclipse.jdt.doc.user/tasks/task-using_batch_compiler.htm) and it tells me

> Each directory or file can specify access rules for **types** between '[' and ']' (e.g. [-X] to forbid access to **type** X, [~X] to discourage access to **type** X, [+p/X:-p/*] to forbid access to all **types** in package p but allow access to p/X).

so it always talks about _types_ not _methods_ of a package. Also please note that there is a _type checking_ also in place with `isTypeUseDeprecated(TypeBinding type, Scope scope)` what is perfectly fine and sensible to me, so if we assume that any _parametertype_ of the method would reference a restricted type that would and should emits a warning.

> Anyway, I have more urgent tasks waiting for me than bending the rules of classpath restrictions.

I really don't want to bend any rules or waste your time, in fact I really appreciate your insights and got the feeling in the past we found some good solutions that are beneficial for the IDE story in general even though the path not always was a direct one. 

But then I don't understand your recommendation [here](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4231#issuecomment-3148606674), as this example shows why PDE currently can not effectively use access restriction rules to mitigate the problem. In OSGi one always "opens" packages not types or methods, that mean if I export Type `T` from package `p` I can access `T` and all its public methods without a problem (assuming no restricted Type `X` is used a parameter or return type). So now for PDE we have the problem that an access is not only restricted by the used type `T` and its package but by _what methods are actually called on an instance of `T`_.

So I really would appreciate any guidance here, because this is one of the problems I really like getting solved as it is highly confusing, requires manual fixing by the user and makes the tooling look brittle compared to other solutions. If it helps I would be fine with even introduce a new restriction type, but as explained above it does not feel right and seems a bit of an overkill here.

stephan-herrmann:
I see one source of confusion in the fact that packages are used as modules, whereas they don't have a clear specification of what are its dependencies and its API - this sounds like an underspecified concept.

_My intuition_ about the API of a package is: all public stuff that you can find in .java files in the package's directory. This gives the following nice guarantee: if a change in any of these files passes compilation of the current artifact, then the change is legal.

_Your intuition_ takes inheritance into consideration, which may have its own advantages.

In case of differing intuitions I prefer to work from specifications. So were can we look for precise answers?

(1) If your example were embedded in a JPMS module which only exports one of its packages, then - much to my surprise - clients of that module are allowed to invoke a method that is declared in a non-exported package. (I haven't checked where in JLS this is defined, but ecj and javac do agree here). Point in favor of considering inheritance.

(2) Next question: Which tool will detect illegal changes in a non-exported package? Assuming that the method in question is not used within the artifact, compilation will not see any problem. Will API tools complain about changes in internal packages, if there is a class in a public package inheriting from the internal one?

Ergo: (1) sets a precedent for selecting your intuition over mine, but (2) needs to be answered regarding API tools, to determine if a change in the ecj implementation can cause bad surprises. The change would allow clients to rely on methods that are _declared_ in a restricted package. If that is so, then who / what would protect that client from incompatible changes? Will API tools put that method under the rules of exported API?

stephan-herrmann:
BTW, did you check the history of those tests in AccessRestrictionsTests that are failing with your change? Perhaps that history would reveal the reasoning behind the current implementation?

laeubi:
> BTW, did you check the history of those tests in AccessRestrictionsTests that are failing with your change? Perhaps that history would reveal the reasoning behind the current implementation?

Yes I need to further investigate and see what cases are not covered correctly by my naive change here. Also there is a difference between _forbidden_ references and _discouraged_ reference and I need to check what applies here.

laeubi:
> If your example were embedded in a JPMS module which only exports one of its packages, then - much to my surprise - clients of that module are allowed to invoke a method that is declared in a non-exported package.

I wanted to verify this as well but as I'm not a JPMS expert I was not fully able to came up with an example, so thanks for doing this experiment. I think this is because of the same reason as it "works" for OSGi, as it is only _classloader visibilty_.

So when we have restricted Type `A` and public type `B` and I can load `B` then I can access all inherited public methods from `A` through a call to the type `B` so even if there is `A.x()` I can always access it like `B.x()` and also with reflection this method will be returned from `B.class.getMethods()` including anything inherited but only **public**. In contrast `B.class.getDeclaredMethods()` will *not* include them (there will likely be a JLS explaining it better) but have e.g. **protected** (and private and so on)  (==aka restricted access). 

So let it be good or bad, from a restriction point of view I never access anything that is restricted that is not restricted by reflection as well (I need access to A.class to get the restricted (non public) method handles). So I would argue, everything that can be legally accessed through reflection without requiring the specific type should be allowed when it comes to restrictions (but JDT itself is sometimes confused as seen [here](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/814) and [here](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/813)).

> Next question: Which tool will detect illegal changes in a non-exported package? 

One thing about the usage of restricted package is, that it not necessarily mean they are _not exported_, but currently _not imported_ so in some cases I can gain access to these packages of course. So currently "the tool" is the compiler (what leads to a problem if one wants to use `javac` for example see [here](https://github.com/eclipse-pde/eclipse.pde/pull/1925#pullrequestreview-3102639905)) so apart from API tools this is of course an important aspect, as otherwise one would fail at runtime (you don't really want API tools for that!).

> Will API tools complain about changes in internal packages, if there is a class in a public package inheriting from the internal one?

Now API tools complicates things a lot (and where we can not fully rely on the compiler anymore):

1. For not imported packages we use _forbidden_ access rules (but of course you can import them and they are on the classpath)
2. We have so called "provisional API" that are packages that are _exported_ (but you should not use them 🙈) so technically they can be _imported_ (but should not 🙈🙈) except for "friends" (🙈🙈🙈) and we use _discouraged_ access rules if not a friend import them
3. Then a type can be marked as `@NoExtend`, `@NoReference`, `@NoImplement`, `@NoInstantiate` ... so actually we would need a way to not restrict _packages_ but a _single type_
4. You can probably already guess, we can have `@NoReference` not only on types but _methods_ or _fields_ so one would need restrictions not only on a type but even on the method level (like `@Deprecated`)

So API tools will need to detect a lot of more case here and some are not discoverable from the raw manifest data.

You can find a list of message one would expect here:

https://github.com/eclipse-pde/eclipse.pde/blob/master/apitools/org.eclipse.pde.api.tools/src/org/eclipse/pde/api/tools/internal/problems/problemmessages.properties

and it complains about as much as is externally visible (except it is marked as `@NoReference`)

stephan-herrmann:
> Now API tools complicates things a lot ...

I can imagine a lot of complexity there, but I wouldn't require _all of that_ to be sorted out, before continuing here. :smile: 

All I'm asking is this: when ecj will (per this issue) legalize access to a method that is _declared_ in a restricted package, this will introduce a maintenance risk, _unless_ API tools will locally detect changes that may break unsuspecting clients. Does API tools detect changes in restricted packages (not exported, or exported x-internal, or such), that may affect clients? For this we don't even need to consider restrictions due to a missing _import_ in the client, I'm only asking about restrictions that are visible at the provider side.

laeubi:
Its quite hard to find a suitable example but:

1. If I extends a "non API type" in a public package API tools warns me 

> Description	Resource	Path	Location	Type
BundleDefaultsScope extends non-API type AbstractScope	BundleDefaultsScope.java	/org.eclipse.equinox.preferences/src/org/eclipse/core/runtime/preferences	line 41	API Usage Problem

so we can say for sure I got warned that I do something nasty, so whatever goes wrong after that fall into the category "you got warned!"

2. I recently fixed a case of such API leak [here](https://github.com/eclipse-equinox/equinox/pull/1066/) and it resulted in some downstream errors here https://github.com/eclipse-equinox/equinox/pull/1066/ so changes in the type hirachy (even though no change in the API) is detected and I remember @heikoklare had has also issues in SWT with API tools being picky when changing even internal details (but I'm not sure what exactly it was about)
3. For API types from other (public) packages it of course always complains anyways.

So I think because API tools forbids that public types extend non public ones, the described problem (I inherit a public method from a non API type and later want to remove that method) should not arise as it is illegal at the first place.

For the case that I extend a public type (and simply do not import the package) there is also no problem as API tools will complain because the base class is part of a public API.

laeubi:
@stephan-herrmann I have now adjusted the first failing `test001()` in the following way:

- if method is accessed through inheritance of public type `X2` it is okay
- if method is accessed through a direct call on restricted type `X1` it is not okay

if we can agree on that being the right thing, I'll look into the other failures (but they look similar).

HeikoKlare:
> I remember HeikoKlare had has also issues in SWT with API tools being picky when changing even internal details (but I'm not sure what exactly it was about)

We had two issues in the recent past of which you probably mean the first:
- When removing overwritten methods already provided by `Object` (`equals`, `hashCode`), API tools complain about the API change even though the API stays the same as those methods will still be there (as the default implementations from `Object`). This was the PR for it: https://github.com/eclipse-platform/eclipse.platform.swt/pull/1880
- When removing a public class of an exported package (thus syntactically being public API) but with a `@noreference` annotation (making it semantically not part of public API), API tools fails because it somehow tries to resolve that removed class (see https://github.com/eclipse-tycho/tycho/issues/5148). But that seems to be very specific to the setup of SWT as I tried to create a test case for it but the scenario worked fine even in a host/fragment setup like SWT.

laeubi:
@HeikoKlare thanks for the reference, did you open already corresponding issues at PDE?
@stephan-herrmann whats you opinion on the matter as described in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4293#issuecomment-3171996499 ? Can we proceed here? It would really be great to get this resolved (in one way or the other) as it is really annoying and often hard to understand for the users.

laeubi:
@stephan-herrmann any chance to get an agreement here? I think it would be good to be merged as early as possible in the release cycle.

laeubi:
@stephan-herrmann I now rebased and adjusted all testcases here, so from my side it would be ready to be submitted for inclusion in M1.

laeubi:
@stephan-herrmann build is finally green, so I think this would be ready to go if we can conclude on this. As this would really improve user experience in the tooling it would be great to make progress here and as both OSGi and JPMS agree here it sounds the right thing to do for me.

laeubi:
@stephan-herrmann can we continue here for maybe get this into M2? Would be highly appreciated!

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/MessageSend.java
```diff
@@ -1002,7 +1003,10 @@ public TypeBinding resolveType(BlockScope scope) {
 		}
 		// abstract private methods cannot occur nor abstract static............
 	}
-	if (isMethodUseDeprecated(this.binding, scope, true, this))
+	TypeBinding declared = this.binding.declaringClass.erasure();
+	TypeBinding actual = this.actualReceiverType.erasure();
+	boolean isExplicitUse = Objects.equals( declared, actual);
```
stephan-herrmann: This should probably use `TypeBinding.equalsEquals()` which also compensates any differences in annotations, which might arise due to `TYPE_USE` annotations. I'm not 100% sure if type annotations will propagate evenly to both sides of the equation, but we made it our rule that all type bindings should be compared using that custom method.

stephan-herrmann:
I was about to be convinced that the change shouldn't hurt, but just upon reviewing test changes this took me back all the way to https://bugs.eclipse.org/bugs/show_bug.cgi?id=76266, which has this description:
> When inheriting a public member from a restricted type, through a legal type, it should be diagnosed as an access violation as well (even though no direct reference to restricted type)

So finally we know that the "undesired" behavior was implemented intentionally exactly that way. And Philipe is an engineer whose judgement I like to follow until proven wrong.

From https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4293#issuecomment-3171912745 :
>  If I extends a "non API type" in a public package API tools warns me

This also gives a hint that we are trying to address a problem which in a perfect world should never arise.

So let's step back one more time:
* I understand that the existing warning can be _surprising_
* Other than this surprise, **does it create any hard problems that are blocking development in any way?**
* If cross-API-boarder inheritance cannot be avoided, isn't the following a reasonable recommendation?
   * If an API class inherits methods from a non-API class, that are meaningful to clients, it may acknowledge this by overriding each such method containing nothing but a super call. With this gesture the API class "owns" the method, and by owning it is entitled to share it with clients.
* JPMS decided differently, but with JPMS the compiler allows several things that will actually blow up at runtime, so being more conservative than JPMS _could_ be an advantage.

stephan-herrmann:
From https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4293#issuecomment-3185016284
> When removing overwritten methods already provided by Object (equals, hashCode), API tools complain ...

At least this will not cause and "restriction" warnings from ecj, as `Object` is not and cannot be restricted. :smile:

stephan-herrmann:
> Its quite hard to find a suitable example but:
> 
> 1. If I extends a "non API type" in a public package API tools warns me
> > Description	Resource	Path	Location	Type
> > BundleDefaultsScope extends non-API type AbstractScope	BundleDefaultsScope.java	/org.eclipse.equinox.preferences/src/org/eclipse/core/runtime/preferences	line 41	API Usage Problem
> 
> so we can say for sure I got warned that I do something nasty, so whatever goes wrong after that fall into the category "you got warned!"

This doesn't actually address the issue I raised. I was asking whether the event of _modifying_ any "semi-API" method (lives in a restricted package but is exposed to the world through a subclass) would cause any tool to shout and tell the developer, that this change may break clients?

The fact that the tool has already shouted before, doesn't help the client who will be broken, if no specific error is raised at the event of the method change.

laeubi:
> So finally we know that the "undesired" behavior was implemented intentionally exactly that way. And Philipe is an engineer whose judgement I like to follow until proven wrong.

I do not want to prove anyone wrong, an maybe it would be worth to add Philipe then here to join the discussions but this was 20 years ago so maybe things have changed in the meanwhile, but what I really miss here is the actual use-case to solve?

So apart from theoretical cases we might want to solve with Access restrictions, what is the practical example where it becomes useful?

1. Java Reflection can not make any use of it and assume it as unrestricted
2. JPMS handle it the same
3. OSGI package import semantic is working the same
4. It is against developers intuition of the topic (and the bug that introduces this does not give any rationale nor is it described in any documentation)

So who we really benefit here? Who will use this feature if PDE + Tycho will need to drop it because it is practically unusable?

> I understand that the existing warning can be surprising

It is not surprising (despite the fact its an error), it is wrong and unintended from the users point of view, there is nothing that ever judge the warning nor is the solution (adding unrelated packages) in any way helpful or required.

> Other than this surprise, does it create any hard problems that are blocking development in any way?

Yes it does, the whole purpose of why we use this feature is to prevent developers to use types (!) that can not be accessed at runtime and will fail. Due to the nature of the problem, that it occurs on top of the class file it is completely disturbing and hard to give good guidance for the tooling so usually people have to intervene manually and are annoyed that we require them to do things that are completely useless.

Even worse, a perfectly working code can suddenly break e.g. because an overridden method is removed or we clean up unused packages: It works in some cases and in other gives strange error then, depending on what methods the user calls.

> If cross-API-boarder inheritance cannot be avoided, isn't the following a reasonable recommendation? 

That's completely unreasonable (beside it bloats the code and the classfile and maybe have runtime performance implications)

1. See very beginning, I don't know any actual case that is treated like this, so I highly doubt I can convince anyone out there writing a library to always override all methods, its often hard enough to convince them to add OSGi metadata at the first place, now asking them to do this because of JDT has an counterintuitive restriction model will simply make them think I'm insane.
2. We often need to to wrap existing libraries, so now having to rewrite the classfiles instead of just the manifest (what creates enough churn on its own) ... 
3. Whenever the "restricted" type gets a new method it becomes immediately visible to the client and most often these are APIs independent of the consumer it needs to be overridden and I need a new library release. So this would tightly couple consumers and providers to specific versions.

> JPMS decided differently, but with JPMS the compiler allows several things that will actually blow up at runtime, so being more conservative than JPMS could be an advantage.

As said nothing blows up at runtime (beside that it makes it harder for the resolver to resolve packages that are never used), we just use usual java classloader visibility.

> The fact that the tool has already shouted before, doesn't help the client who will be broken, if no specific error is raised at the event of the method change.

That's a completely different story and restriction warnings do not solve this because:

1. either I get a warning and then simply import that package --> done
2. I get not warning as I have not called the method, it still might break me depending on what this is used, e.g. public methods might be called in other ways, as part of implemented interfaces, reflection through frameworks, binding to a scripting and so on.

The point is we have tools to restrict "semi API" on methods and this is completely vendor (PDE) specific thing that has no meaning to OSGi or the runtime, so nothing will break if you use such restricted method or fields.

stephan-herrmann:
Wow, the disconnect runs deeper than I thought. In an attempt to debug the discussion I will answer in small increments.

Easiest first:
>>  I understand that the existing warning can be surprising
>
> It is not surprising (despite the fact its an error), ...

If we are speaking of the same thing, then it's each project's choice to configure this diagnostic to Error/Warning/Info/Ignore, in two separate options for "forbidden" and "discouraged".  Since it's not a mandatory error (as those that are defined in JLS) I suggest to use the term warning in a generic way, knowing that warnings can be promoted or demoted in severity by the user.

While I'm in favor of zero-warnings policies, there are means for coping with a small number of exceptions, like `@SuppressWarnings`.

But as the PR headline speaks of "warning", this might be a simple misunderstanding, perhaps when you wrote "error" you were referring to something else?

stephan-herrmann:
Next we should perhaps clarify which situation we are discussing: A client invokes a method that is declared in a class B, superclass of API class A. Possible reasons why this could be a problem:
1. B is defined in a non-API package
2. both packages are API, but the client imports only the package of A, not the package of B

I thought we are speaking of (1) but when you speak of adding unnecessary imports, this sounds more like (2). Again, I might be misunderstanding why you speak of adding more package imports.

laeubi:
> If we are speaking of the same thing, then it's each project's choice to configure this diagnostic to Error/Warning/Info/Ignore, in two separate options for "forbidden" and "discouraged".

And here it starts to getting wrong already, you assume that is a _deliberate choice of the user_ but its not. 
While these would work for a very small scale maybe its completely unmanageable for any larger setup with hundreds of libraries used not talking about the effort to do it "right".

So here a tool is needed (PDE) to translate between OSGi world and JDT world, and it is not a question of choice but correctness, if I do not import a package of a type than it is an error without doubt. But calling a method on a public and imported type is neither an error nor a warning.

So let it be a warning or an error is not that important to me, usually it leads to a compile error.

laeubi:
> I thought we are speaking of (1) but when you speak of adding unnecessary imports, this sounds more like (2)

Both cases do not make a real difference while the first is quite uncommon (see issues with API leaks) but of course it is always possible that `A` (exported) extends `B` (abstract but not exported) where B implements functions defined in `I` (exported) but not implemented by `B` (of course still inherited by `A` and therefore never implemented).

So the whole "API" versus "No API" is a bit misleading, as restrictions do not tell anything about API or not they simply restrict access to packages (at least documentation states that), so it *might* be used to hide "non api" but also anything implicitly hidden (e.g. by class-loader visibility).

stephan-herrmann:
Before we can speak about right or wrong (which is a very strong word), let me ask if their is any warning-free setup of projects / libraries, where this issue is relevant?

If I understand correctly, then (a) we are speaking only about the non-API to API inheritance case and (b) this case will inevitably raise a warning by API tools. 

If this assumption is correct, then I don't even see us speaking about correctness, but about ways for coping with a situation that already violates some rules from the very outset - the library itself is inconsistent. Defining consistent usage of an inconsistent library is a difficult task.

stephan-herrmann:
> So the whole "API" versus "No API" is a bit misleading, as restrictions do not tell anything about API or not they simply restrict access to packages (at least documentation states that), so it _might_ be used to hide "non api" but also anything implicitly hidden (e.g. by class-loader visibility).

You are right in that Access Rules do not specify reasons for restrictions, they are just a technical means to some undefined end. It is that component (or user) that defines Access Rules which does this for a particular purpose (e.g. a classpath container). This is were we enter the realm of use cases, and as you rightly asked which use case we are addressing, it seems important to distinguish different use cases, no?

For me as a user the API / non-API question is something that I can see, so that's why I'm primarily thinking about this. If some classpath container adds Access Rules on elements that have not been imported by the client project, then this is already somewhat hidden. The user has no insights that any non-imported stuff is on the build path at all.

From your latest answer, however, I understand you are not much interested in the API / non-API question. So, sure, we can focus on use cases where everything is API indeed, but some parts of it are simply not directly imported. 

At this point it's of course a pity that JDT cannot know if an Access Rule represents: not accessible because not API, or: not accessible because not imported.

laeubi:
> Before we can speak about right or wrong (which is a very strong word), let me ask if their is any warning-free setup of projects / libraries, where this issue is relevant?

I don't understand? I linked the PR that shows the problem and except of that it is quite warning free, apart from that one encounter it quite easy and the changed testcases also show such case.

> If I understand correctly, then (a) we are speaking only about the non-API to API inheritance case and (b) this case will inevitably raise a warning by API tools.

No it is about accessible Types (imported packages) versus methods inherited by these either directly or indirectly. And no API tools will never complain because it does not care what a user uses (except for no-reference marked things what JDT does not support anyways).

> If this assumption is correct, then I don't even see us speaking about correctness, but about ways for coping with a situation that already violates some rules from the very outset - the library itself is inconsistent. Defining consistent usage of an inconsistent library is a difficult task.

Again none of these situations are violating anything (except the unspoken JDT contract of how access restricted packages work when it comes to inherited methods and I haven't yet seen any compelling usecase why it should be the case).

1. I import package `a.b.c` and it contains a public type `P` inside `example.jar`
2. PDE "opens" package `a.b.c` but restricts all other packages from `example.jar` as I do not import them
3. No I can access all public methods of `P` (either inherited or not) du to the usual java rules and everything should be fine
4. I can even run the code without any problems (e.g. classloader not finding a class)
5. - **BUT** - if method is inherited from `X` in package `x.y.z` of `example.jar` JDT complains I access a restricted type `X` (what I'm doing not --> all public methods of a super-class are accessible through the class `P` itself and i nowhere reference `X` in any way)

stephan-herrmann:
Can we tone down the discussion? Claims about concepts being "wrong", "unusable" etc. don't help a lot as long as misunderstandings are involved. When speaking of the intuition of users, please consider that I am a user, too, and I might have intuitions that differ from your's.

Did we agree that the concept of Access Rules (per its documentation) does not specify the global purpose these rules should be used for? Why should access to some things be restricted, and not others? So one possible way forward could be to put on the table which problems we'd like to see solved by Access Rules. Hint: I'm trying to think of Access Rules as a concept that's agnostic to any technology beyond the intrinsic rules of Java. At this level everything is possible if cleanly specified. There is not right or wrong, as long as there is no specification.

laeubi:
> Did we agree that the concept of Access Rules (per its documentation) does not specify the global purpose these rules should be used for?

I'm not sure do you mean this one? https://help.eclipse.org/latest/topic/org.eclipse.jdt.doc.user/concepts/concept-access-rules.htm

Sure it does not define a purpose, but also there it talks about Types(!) not Method and it does not say anything about inheritance.

> Why should access to some things be restricted, and not others? 

Because one thing is accessible via the classloader and not the others, so I also don't really understand, you asked for an example and I tried to give an as simple one that explains the problem we are facing and maps exactly the my usecase, so what is needed here?

> Hint: I'm trying to think of Access Rules as a concept that's agnostic to any technology beyond the intrinsic rules of Java.

That's okay, but if the concept does not match any uses-cases it is not very practical, so I just described one very common use-case in PDE and have no idea where else it is used ... if we would know that it might be easier to go forward with this.

stephan-herrmann:
Ideally, I would merge this the very moment we agree on a common understanding of things.

Unfortunately, all my questions have not brought us a lot closer to such common understanding.

I don't want to veto a change that will help downstream projects to improve. But before merging I will lay out where I'm coming from:

### Problem 0: Insufficient specification

 The concepts of access restrictions have been developed long ago (first introduced 2004 (52e6fbe4) and updated in a way that is significant for the current issue in 2006), but neither do we have a specification that speaks of all the details we need, nor are the people from the early days available for clarification. Two decades later, we may fix obvious bugs, but reverting a decision that was intentionally and explicitly done back then is not fixing a bug, it's changing the design. When doing so, we should be extra careful not to break any downstream projects that use access restrictions. Mind you: we have no idea how many downstream projects might be doing so.

----
### Problem 1: Are access restrictions a syntactic concept or a semantic one?

At the syntax level access to org/foo/A means: you may refer to (visible) methods which you find in org/foo/A.java or org/foo/A.class. Neither file contains inherited methods, inheritance is a semantic concept.

Pro syntax:
* javadoc of `IAccessRule` starts: "Describes an access rule to source and class files on a classpath entry."
This speaks of files, nothing semantic.
* access rules are specified as file name patterns

Pro semantics:
* other documentation speaks of "types" (*but says nothing about members*)

In my understanding the syntactic view is more strongly stated, than the semantic one.

NB: Also `java.lang.Class` tells use that there are 2 different views on which methods a class "has": `Class.getDeclaredMethod()` vs. `Class.getMethods()`.

----
### Problem 2: What do downstream projects want to express using access rules?

In my understanding access rules / restrictions are an attempt to allow downstream (classpath containers) to express rules of a specific module system so that JDT (which is agnostic of any such system other than - since 9 - JPMS) can help users of that module system.

Thus we'd want to cover a range of systems like maven, OSGi, perhaps academic systems like ArchJava, or component systems from Corba to EJB, and systems we have never heard of.

Ideally, with access restrictions JDT could inform users about the following risk:
* reliance on a component with which no explicit contract exists.

On the other hand, library developers want assurance that parts of their code that are not bound by an explicit contract can be changed at will without breaking clients.  Here we'd be discussing the shape of the boundary between API and non-API code. The semantic view could potentially create quite some swiss cheese.

Different module systems may serve these purposes to different degrees.

#### Contracts based on packages

Unfortunately, when mapping specifically OSGi "Import-Package" to access rules, we're dealing with a concept that is quite weak when used as a module: A Java package can be understood as a container that exports all its public types, but a package does not declare its dependencies.

Consider 
* package p _version x_
   * class A with method m
* package q _version y_
   *  class B extends p.A
* a client imports package q _version y_
   * class C calls m via B

Now while compiling the client we may incidentally have package p _version x_ in our target platform, but nothing in this setup guarantees that this is also true at runtime. Perhaps m was introduced only in _version x_, but at runtime we have p _version (x-1)_.  Or p is a non-API package and m will be removed in _version (x+1)_. 

Package q, OTOH, may have no desire to link specifically to _version x_ of p. This dependency exists only in the client.

Ergo any transitive dependencies via a package are brittle. And just for completeness: classes are not better in this regard (imports at this level are just a syntactic shorthand).

It happens that the syntactic view is more conservative, introducing fewer risks. 

#### More automation? More specification?

I could even think of a classpath container that considers inheritance in a way that it adds super types of API-types with weaker or no access restriction. We could even introduce a new restriction level, e.g., "implicit allowed", which grants access to inherited methods but does not allow explicitly mentioning their type. Just brainstorming.

----

From looking at our terminology (e.g., you: classloader, I: contract) it's clear we have different perspectives. There is no clearly correct or incorrect solution, because we are building on weak specification of access rules and on the weak concept of packages.

Do you still think PDE needs exactly this design change? Are you willing to take responsibility for potential problems caused downstream?

If yes to both, then you should clearly explain this design change in the release notes of the 4.38 release, in which case I'm ready to merge this.

laeubi:
> Unfortunately, all my questions have not brought us a lot closer to such common understanding.

I'm sorry for that but unsure where there is the miss?

> NB: Also java.lang.Class tells use that there are 2 different views on which methods a class "has": Class.getDeclaredMethod() vs. Class.getMethods().

Yes that's exactly the difference here. I would also be fine with a new restriction flag, but this will obviously heavily complicate and bloat up the implementation and will require to trickle down through all the toolchains.

> Unfortunately, when mapping specifically OSGi "Import-Package" to access rules, we're dealing with a concept that is quite weak when used as a module: A Java package can be understood as a container that exports all its public types, but a package does not declare its dependencies.

This is not completely accurate. You are right that import-package does not declare dependencies, it is a dependency on its own. But an import is always resolved to a capability (that is an Export-Package), and exported packages *have* dependencies via the [use-constrains](https://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#i3127019). Still those are not *transitive* so even if the exporter uses package `x` a bundle that wants to use `x` needs to import it, as these are only resolved at runtime and the resolver needs to make sure both `x` are resolved to the same provider.

> Now while compiling the client we may incidentally have package p version x in our target platform, but nothing in this setup guarantees that this is also true at runtime. 

I think nothing can guarantee that, whatever I do I can run with something completely different. Anyways OSGi specifies the use of a [semantic-versioning scheme](https://docs.osgi.org/whitepaper/semantic-versioning/), where one uses a version range on the package for either a consumer or producer scenario (and that where actually api-tools kick in!) to ensure there is a match at runtime.

> Do you still think PDE needs exactly this design change?

Yes I'm very confident with that.

> Are you willing to take responsibility for potential problems caused downstream?

I'm not sure one can commit to such a broad liability. For what it matters I would look into any regressions introduced or cases where it behaves wrong if the use-case is known and suggest further enhancements.

> Still one detail to improve

I'll push a change soon.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/MessageSend.java
laeubi: Done!

stephan-herrmann:
> > Now while compiling the client we may incidentally have package p version x in our target platform, but nothing in this setup guarantees that this is also true at runtime.
> 
> I think nothing can guarantee that, whatever I do I can run with something completely different. 

This is a lame statement. From my previous comment you might have understood that I'm looking for the set of contracts and corresponding tool functions guaranteeing that any misconfiguration will be detected systematically.

Without a contract between client and p.A the problem will tend to blow up at runtime as NoSuchMethodError, which is the worst of all ways to deal with misconfiguration.

Anyway, in an attempt to create concrete examples to demonstrate my point in code rather than prose, I realized that inheritance is totally not supported in a related scenario: Assume packages p and q from the above example sit in different bundles. Client has a dependency on q containing B, but the way PDE configures those projects, the client is unable to call m(), ecj will respond with
```
The type p.A cannot be resolved. It is indirectly referenced from required type q.B
```
Ergo: as soon as we speak of a bundle boundary instead of, e.g., a package boundary, the client invariably needs to specify its dependency on p (p.A) in order to call m() even if done via q.B.

Doesn't this tell us: "if you want to invoke methods _declared in type A_, you need to have a dependency on A (possibly via  one of its containers)"?

I would not like to introduce an inconsistency whereby inheritance is considered in some dependency situations but not in others.

stephan-herrmann:
And here is the example using two packages within the same bundle: 
[restriction-example.zip](https://github.com/user-attachments/files/23101208/restriction-example.zip)

(PDE has some hiccups, when projects Lib1 or Lib2 are closed / opened, but let's ignore that here).

Currently, the only problem reported is the access restriction in question.

From a workspace with those three projects I can launch an OSGi application containing Lib1 and Client but not Lib2. OSGi finds no problem. But any attempt to run `c.Client.m()` is doomed.

In my book this signals that reporting the access restriction is necessary in order to avoid havoc at runtime. 

In your book, when/how should the problem be detected?

laeubi:
> Without a contract between client and p.A the problem will tend to blow up at runtime as NoSuchMethodError, which is the worst of all ways to deal with misconfiguration.

I don't understand?!? For sure we can assume that there is a consistent classpath at runtime I doubt you can really load `p.A` that extends `p.B` and B is not there (I'm not talking about this would be a split package what is bad anyways and you can not use import package for that).

> Assume packages p and q from the above example sit in different bundles. Client has a dependency on q containing B, but the way PDE configures those projects, the client is unable to call m(), ecj will respond with

Yes this is a bug in PDE and I like to fix that (Tycho is not having that problem by the way) but it unlikely brings anything to replace one error/warning with another, so that's why I started to investigate on this (simpler case) first to ~fix~ make it work then.

> From a workspace with those three projects I can launch an OSGi application containing Lib1 and Client but not Lib2. OSGi finds no problem. But any attempt to run c.Client.m() is doomed.

Your example has some problems:

1. You are missing use-clauses*, that way you say both packages are independent, what they are not but this will just lead to problems when you have installed both and then probably getting class cast exceptions in the worst case.
2. You provide two bundles that claim to provide the same API but they don't (1.0) for q, your Lib2 needs to increment is minor version
3. Your client then needs to use the lower bounds of the package import to reflect i wants to use the 1.1 API

So regardless of access restrictions you lie on OSGi here and therefore there is no way for it to complain and you get runtime errors. That's also the reason you get hickups in PDE in the IDE, for what it looks as if Lib1+Lib2 are a split package that you can't import that way.


 (*) it has to look like this
```
Export-Package: p;version="1.0.0",
 q;version="1.0.0";uses:="p"
```

stephan-herrmann:
> > Without a contract between client and p.A the problem will tend to blow up at runtime as NoSuchMethodError, which is the worst of all ways to deal with misconfiguration.
> 
> I don't understand?!? For sure we can assume that there is a consistent classpath at runtime I doubt you can really load `p.A` that extends `p.B` and B is not there.

c'mon, you should be able to see that I'm not talking about a missing class but about a version of the class that doesn't have the method being called.

 
> > Assume packages p and q from the above example sit in different bundles. Client has a dependency on q containing B, but the way PDE configures those projects, the client is unable to call m(), ecj will respond with
> 
> Yes this is a bug in PDE 

who defines if it's a bug or a feature?

> > From a workspace with those three projects I can launch an OSGi application containing Lib1 and Client but not Lib2. OSGi finds no problem. But any attempt to run c.Client.m() is doomed.
> 
> Your example has some problems:

Sure :)
Most users create code with problems much of their time. It's our job to tell them.

> 1. You are missing use-clauses*, 

which tool will tell me that I should add them?

> 2. You provide two bundles that claim to provide the same API but they don't (1.0) for q, your Lib2 needs to increment is minor version

You mean increment the version of q, right? (The version of Lib has been incremented)

But, I don't see any changes in q, why should the version be incremented? Who tells me?

> 3. Your client then needs to use the lower bounds of the package import to reflect i wants to use the 1.1 API

If all else is reported as errors, this step is fair.

> So regardless of access restrictions you lie on OSGi here 

If I make mistakes, it's the job of tools to point them out.
If we merge this MR, it seems that no tool will warn me any more, which is worse, not better.

So, prerequisite for merging is that you tell me, which tool will catch my bug in a better way than the existing error we have. The same also applies to the multi-bundle situation. Will bundle versions require an update as soon as one of their dependencies exist in a newer version (like when you are forcing version of q to be updated)?

> (*) it has to look like this
> 
> ```
> Export-Package: p;version="1.0.0",
>  q;version="1.0.0";uses:="p"
> ```

Interestingly this doesn't seem to change anything in my example. What effect should I see?

Reminder: when done we should actually generalize the discussion to be applicable also to all other classpath containers.

laeubi:
@stephan-herrmann I now converted your example in a testcase here:

- https://github.com/eclipse-tycho/tycho/pull/5511

the result is
grafik

stephan-herrmann:
> @stephan-herrmann I now converted your example in a testcase here:
> 
> * [bundle-with-inheritance test project eclipse-tycho/tycho#5511](https://github.com/eclipse-tycho/tycho/pull/5511)
> 
> the result is ... 

nice.

Does this mean the discussion of uses constraints was a red herring? Or where will those help?

Leaves us with these questions:
* Will the IDE report the same problem as tycho?
* Will the same problem also be reported when p and q reside in separate bundles?
* How will it affect other classpath containers?

laeubi:
> c'mon, you should be able to see that I'm not talking about a missing class but about a version of the class that doesn't have the method being called.

Then it might load but it is simply invalid, I can create a myriad of cases where something compiles without an error and fails at runtime, so is this a fault in the compiler? I would say no. So is a fault of the tooling? It depends and that's where it comes to "I don't see errors", yes PDE as a tooling is not perfect but we try to improve it.

> If I make mistakes, it's the job of tools to point them out.

OSGi is not a tool it is a framework, if you want errors for your case then you need to setup API tool checks good luck with that :-P 

> If we merge this MR, it seems that no tool will warn me any more, which is worse, not better.

Again, you can easily get the same "problems" without access restrictions now doubt, the access restriction is only a red-herring here! Just import the second package (with 1.0 or no version as often seen), access restriction is gone but the error remains and that's exactly the point I try to make you understand, the *only* difference is that you need to add a new (unwanted) import it does not help nor prevents you from errors!

> Will bundle versions require an update as soon as one of their dependencies exist in a newer version (like when you are forcing version of q to be updated)?

It depends, if they are compatible it is okay, but recommended... as some thing recommendations do not matter an we are all smarter than that we have a tool that checks this and then you get these kind of PRs:

- https://github.com/eclipse-platform/eclipse.platform.ui/pull/2831
- https://github.com/eclipse-equinox/p2/pull/945
- https://github.com/eclipse-pde/eclipse.pde/pull/1764
- ... many more ...

> Interestingly this doesn't seem to change anything in my example. What effect should I see?

You will only see an effect at resolution time, I just added it here for completeness...

stephan-herrmann:
> > c'mon, you should be able to see that I'm not talking about a missing class but about a version of the class that doesn't have the method being called.
> 
> Then it might load but it is simply invalid, I can create a myriad of cases where something compiles without an error and fails at runtime, so is this a fault in the compiler? I would say no. 

Yes, for certain classes of errors, compilers give guarantees (assuming you don't bypass the compiler by reflection, replacing classes after compilation etc):
* Static typing guarantees that code that was compiled in one batch and run together will never throw NoSuchMethodError 
* Compiling code with generics will insert invisible class casts with the guarantee: if no "unchecked" warnings where raised then these invisible cast will never throw CCE
* One by one static analysis addresses additional classes of errors (e.g., NPE).

Hence my assumption is: tools specialized in composing software from modules should promise that misconfiguration can be detected at build time. And like the JVM rechecks much of what the compiler already checked, I expect runtimes like OSGi to raise alarm, so the distinction tool vs. runtime is not so relevant. A good runtime will protect against certain classes of errors, too. 

> So is a fault of the tooling? It depends ...

Like I outlined some error classes detected by compilers, it would be great if other tools, too, make their promises explicit.

>  yes PDE as a tooling is not perfect but we try to improve it.

much appreciated

> > If we merge this MR, it seems that no tool will warn me any more, which is worse, not better.
> 
> Again, you can easily get the same "problems" without access restrictions now doubt, the access restriction is only a red-herring here! Just import the second package (with 1.0 or no version as often seen), access restriction is gone but the error remains and that's exactly the point I try to make you understand, the _only_ difference is that you need to add a new (unwanted) import it does not help nor prevents you from errors!

Here we are finally making progress! I had falsely assumed that an explicit package import would add safety. 
Seeing it silencing the error without fixing it is a relevant piece of information.
 
> > Will bundle versions require an update as soon as one of their dependencies exist in a newer version (like when you are forcing version of q to be updated)?
> 
> It depends, if they are compatible it is okay, but recommended... as some thing recommendations do not matter an we are all smarter than that we have a tool that checks this and then you get these kind of PRs:

Will that bot also kick in wrt methods inherited from another bundle?

> > Interestingly this doesn't seem to change anything in my example. What effect should I see?
> 
> You will only see an effect at resolution time, I just added it here for completeness...

Well even at runtime it didn't help to detect the problem, but let's drop this thread here.

stephan-herrmann:
>  I had falsely assumed that an explicit package import would add safety.
Seeing it silencing the error without fixing it is a relevant piece of information.

To explain more: I had been looking for ways how tools enforce more explicit contracts, in the hope that those contracts help detecting misconfigurations. In that I failed to see, that a component may require X version 1.0, but still be compiled against X version 1.1 with no complaints, thus illegally using version 1.1 features. Whether or not that should be allowed opens another can of worms, which should not be opened here (is that the pde-bot thing?). This implies that problems caused by uncoordinated software evolution cannot easily be detected just looking at one set of components, but are better detected by version comparison.

With this I'm willing to go back to MR pragmatics.

@laeubi  can you provide a message to adopters that will inform them about this design change in the release notes?

laeubi:
> Yes, for certain classes of errors, compilers give guarantees (assuming you don't bypass the compiler by reflection, replacing classes after compilation etc):

But exactly that is happening here, the situation you created is a bit like Compiling with Java 11 JDK with a source compliance of 8 and don't using the target option... The compiler can simply not know that at runtime some calls will fail.

One might argue the compiler can still warn you if it would do an further analysis assuming an implicit target option, but this would be likely a different tool (and there are tools for this!)

> And like the JVM rechecks much of what the compiler already checked, I expect runtimes like OSGi to raise alarm, so the distinction tool vs. runtime is not so relevant. A good runtime will protect against certain classes of errors, too.

The problem is it can't if the meta-data is wrong! Assume I compile a class as compliance 11, then would edit the class file and replace  the version with the one for 8 with an hex editor... no good thing could come from this.

For the same OSGi (runtime) must assume the metadata in manifest is correct what is the whole purpose of it, so it does not need to analyze the classfiles for example you need to state in the Manifest what is the minimal java version you want to run with.

Now it comes to the point that PDE (compile time) historically has leaned to the so-called "Manifest-first" that is a developer usually writes the manifest (I'm working on help the developer to get some errors/warning here https://github.com/eclipse-pde/eclipse.pde/pull/1926 if they likely do things wrong), so we need more tools (e.g API tools) that need more data (e.g the baseline) to provide further assistance / checks at "tool time" as we all tend to make mistakes.

laeubi:
@stephan-herrmann I proposed a N&N entry here

- https://github.com/eclipse-platform/www.eclipse.org-eclipse/pull/428

also squashed, rebased and added a reference to this discussion in the commit message.

stephan-herrmann:
> @stephan-herrmann I proposed a N&N entry here
> 
> * [Add N&N entry for improved access restriction warnings eclipse-platform/www.eclipse.org-eclipse#428](https://github.com/eclipse-platform/www.eclipse.org-eclipse/pull/428)

Two suggestions:
* don't claim that your intuition is the only intuition, the change replaces one consistent interpretation with another one.
* please mention the impact on implementors of `org.eclipse.jdt.core.IClasspathContainer` - they may want to revisit what access restrictions they are generating. They have been relying on the existing logic for decades.
 
 > also squashed, rebased and added a reference to this discussion in the commit message.

thanks.

stephan-herrmann:
> if you want errors for your case then you need to setup API tool checks good luck with that

That wasn't so hard. Result: it doesn't detect the need to increment any version due to method added to a super class. Neither when inheriting from a class of another bundle, nor when inheritance is from a different package of the same bundle.

=> one more tool to be updated in order to consistently support inherited methods as API of the subclass.

laeubi:
Does it detect other problems? It is not that easy to setup often to give good results, that's why I wrote the tycho-baseline check :-)

Also API tools is often not that good at detect package-version increments, I recently added some support but there is maybe still things to fix. In this case it could be that API tools is happy because you increased the bundle version already :-\

laeubi:
> Two suggestions:

Do you like to suggest a change in that PR? I can then just apply it to the entry, I'm open for any changes.

stephan-herrmann:
> Does it detect other problems? It is not that easy to setup often to give good results,

It reported quite a number of other bugs I injected.

> Also API tools is often not that good at detect package-version increments, I recently added some support but there is maybe still things to fix. In this case it could be that API tools is happy because you increased the bundle version already :-\

I played with several variants of versioning, none brought the desired warning relating to package q.

stephan-herrmann:
> > Two suggestions:
> 
> Do you like to suggest a change in that PR? I can then just apply it to the entry, I'm open for any changes.

you can do that, it's your feature :)

laeubi:
> > > Two suggestions:
> > 
> > 
> > Do you like to suggest a change in that PR? I can then just apply it to the entry, I'm open for any changes.
> 
> you can do that, it's your feature :)

I'm quite bad ad writing N&N entries but will try to improve that..

laeubi:
@stephan-herrmann just as a follow up now:

As we are close to the 2026-03 release and as far as I know here where no complains about this change I now was able to finally add a major improvement to PDE (even though the code changes are low):

- https://github.com/eclipse-pde/eclipse.pde/pull/2218

this will now fix a large group of annoying 

> The method X from the type Y refers to the missing type Z

what not always was clumsy and hard to understand but even broke code navigation in some cases even when the error itself was not triggered.

So another step for a more satisfactory plugin-development experience :+1:

Issue #4243, opened 2025-07-25
ECJ eats a class annotation is null analysis is performed


syskin345:
OK this is a fun one. Use Eclipse 2025-09 M1 (4.37.0.v20250708-0906), it was fine in 2025-06 release.

1. Make sure you have a working @NonnullByDefault annotation (make one if needed) and null analysis enabled.
2. Compile the following class (any Java compliance that supports records is fine)
```
@NonnullByDefault
@Main.NeededAnnotation
public final class Main {

   public static void main(String[] args) {
      for (Annotation anno : Main.class.getAnnotations()) {
         System.err.println(anno);
      }
   }

   public record Foo(String needed) {}

   @Retention(RetentionPolicy.RUNTIME)
   @interface NeededAnnotation {}
}
```
3. Observe how, in the IDE, @Main.NeededAnnotation is not coloured like an annotation should be, and also if you run this `main()` you will see the annotation is not found on the class.

Swapping the two annotations makes it work again. Removing that record makes it work again. Making the record empty makes it work again....

srikanth-sankaran:
I'll take a look and pass it on to @stephan-herrmann as needed.

srikanth-sankaran:
Why are the imports missing in the snippet ? Is it really `@NonnullByDefault` or did you mean `@NonNullByDefault` ?

syskin345:
Ooops, yes I used our internal `@NonnullByDefault` which has a different name indeed. I intentionally skipped its import since it wouldn't be the same annotation anyway.

The important bit is that it's the configured annotation used by null analysis.

srikanth-sankaran:
I am having a bit of a bother reproducing. With JDT's own annotation `@NonNullByDefault` this program with null annotation analysis enabled:

```
import java.lang.annotation.Annotation;
import java.lang.annotation.Retention;
import java.lang.annotation.RetentionPolicy;

import org.eclipse.jdt.annotation.NonNullByDefault;

@NonNullByDefault
@Main.NeededAnnotation
public final class Main {

   public static void main(String[] args) {
      for (Annotation anno : Main.class.getAnnotations()) {
         System.err.println(anno);
      }
   }

   public record Foo(String needed) {}

   @Retention(RetentionPolicy.RUNTIME)
   @interface NeededAnnotation {}
}
```

prints `@Main.NeededAnnotation()` at runtime with 4,37 M1, 4.36 and master/head 

What am I supposed to see in the problem scenario ?

[...]

PR #4299, opened 2025-08-09
ECJ eats a class annotation if null analysis is performed

stephan-herrmann:
+ fix confusion tagBits vs. extendedTagTypes
+ don't skip annotation resolution when only some have been resolved
+ don't return incompletely initialized annotations array

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4243


stephan-herrmann:
This mainly fixes some more or less obvious flaws of 3ffaed33, where we introduced selective resolving of only `@NonNullByDefault` (following the example of `@Deprecated`).

One change in ASTNode.resolveAnnotations() is less obvious: we copy annotations on variable declarations with multiple parts to all parts, which we detect by a pre-set `annotation.recipient`. Now with partial resolving of an array of annotations, `annotation.recipient` can be pre-set in more cases, in which case the immediate `return annotations` was not appropriate, so that is being pushed into the cases where it is needed, while the default case continues with assigning individual annotation bindings.

Issue #3907, opened 2025-04-08
Compilation error on full build but not on incremental build due to @deprecated


RoiSoleil:
I get a compilation error on full build but not on incremental build due to @deprecated.

![Image](https://github.com/user-attachments/assets/b79d8551-654a-409f-94b5-3c1c1ddbc70d)

I created a repro project at https://github.com/RoiSoleil/repro-extendwith-compilation-error

Just import the project and do a clean full build.

If i remove the @deprectaed in repro.AbstractLoad<T> no problemn anymore

RoiSoleil:
@srikanth-sankaran Any news ?

srikanth-sankaran:
> @srikanth-sankaran Any news ?

No, I haven't had time to look into this at all. Thanks for the reminder. I'll try and close it for M2

srikanth-sankaran:
@RoiSoleil in the meantime it would help if you are able to find out if this is a recent regression and if so the rough point when the faulty behavior originated.

RoiSoleil:
This is due to org.eclipse.jdt.internal.compiler.problem.ProblemReporter.deprecatedSinceValue(Supplier<AnnotationBinding[]>) calling isCompatibleWith without connect SuperClass and superInterfaces before that. So a false not compatible isCompatibleWith is stored in compatibleCache.

srikanth-sankaran:
> This is due to org.eclipse.jdt.internal.compiler.problem.ProblemReporter.deprecatedSinceValue(Supplier<AnnotationBinding[]>) calling isCompatibleWith without connect SuperClass and superInterfaces before that. So a false not compatible isCompatibleWith is stored in compatibleCache.

Useful finding - thanks, will investigate ...

srikanth-sankaran:
Reduced standalone test case with no external dependencies such as junit:

```
public class LoadExtension extends AbstractLoad<Integer> {
} 
```

```
public final class TestIntegrationExtension implements Extension {

}

@ExtendWith(TestIntegrationExtension.class)           
@interface TestIntegration {  
}

@interface ExtendWith {
	Class<? extends Extension>[] value();
}

interface Extension {
}


/**
 * @deprecated 
 */
@TestIntegration()
@Deprecated
abstract class AbstractLoad<T extends Number> {

}
```

srikanth-sankaran:
Regression test that can be used for debugging: (add to GRT*.java for example)

```
static {
[...]

PR #4301, opened 2025-08-09
Compilation error on full build but not on incremental build due to @Deprecated

stephan-herrmann:
+ when error reporting needs the since value, don't resolve all annots
  - this prevents that membervalue pairs of sibling annotations are prematurely type checked (before hierarchy linked)
+ during initialization of deprecated annotation store it for later use

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3907


stephan-herrmann:
We've always had the option of initializing only `@Deprecated` annotations, while leaving others for later.

Extending this concept also to `@NonNullByDefault` helped avoid premature type checking triggered by the need to check for `@NonNullByDefault` (this one has just been refined in #4299).

In the current bug it was problem reporting that needed to inspect a `@Deprecated` annotation to find its `since` value. This lookup did not leverage the option of selective initialization, thus causing the premature resolution and type checking of the sibling annotation.

PR #4302, opened 2025-08-11
Check for thread interruption in Parser::parse

fedejeanne:
## What it does
Regularly checks for the interrupted state of the current thread and correspondingly aborts execution by setting the interrupted state again and throwing an exception

## How to test
- Start the completion proposal that requires an expensive AST calculation (or add a 100ms delay in the very same `for`-loop I modified in this PR to simulate the expensive calculation)
- Cancel the completion proposals by clicking outside of the proposal pop-up window

**Expected result**
- `AsyncCompletionProposalPopup` signals the cancellation (requires https://github.com/eclipse-platform/eclipse.platform.ui/pull/3157)
- The parsing is aborted and the exception is propagated (this PR)
- No exception is logged (requires https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2425)

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
@@ -11605,6 +11605,11 @@ protected void parse() {
 try {
 	this.scanner.setActiveParser(this);
 	ProcessTerminals : for (;;) {
+		if (Thread.currentThread().isInterrupted()) {
+            Thread.currentThread().interrupt();
+            throw new RuntimeException("[" + Thread.currentThread().getName() + "] Task interrupted", new InterruptedException()); //$NON-NLS-1$ //$NON-NLS-2$
+        }
+
 		int stackLength = this.stack.length;
 		if (++this.stateStackTop >= stackLength) {
 			System.arraycopy(


fedejeanne:
It would be nice if someone could review this one. It's a small change and it enables the parser to be canceled when running in separate threads (_e.g._ when calculating the autocompletions).

fedejeanne:
Maybe someone from @mpalat 's team could review this? :-)

fedejeanne:
Or maybe @mickaelistria would like to review? I know you are not a committer but you know your way around JDT too :)

fedejeanne:
Thank you very much @mpalat and @jjohnstn for reviewing and merging! 🚀

stephan-herrmann:
Wow, throwing `RuntimeException` in "the" central method of jdt.core is a bold step. How can we be sure that all call paths leading into this method diligently handle the undeclared exception?

For a start please have a look at `org.eclipse.jdt.internal.compiler.Compiler.compile(ICompilationUnit[], boolean)`. Here we catch `AbortCompilation` **but not** `RuntimeException`.

The same for `org.eclipse.jdt.internal.core.builder.AbstractImageBuilder.compile(SourceFile[], SourceFile[], boolean)`

And that's just one call path leading into this method.

fedejeanne:
@stephan-herrmann the exception is only thrown if the user requests to interrupt the thread so it shouldn't happen very often. Of course, throwing directly an `InterruptedException` would have been my favorite approach but that's a checked exception so it would have required lots of changes in callers too.

I also briefly considered using any of the exceptions in the package `org.eclipse.jdt.internal.compiler.problem` but I wasn't sure if they fit the use case: on the one side they are runtime exceptions (good ✅ ) but their "JavaDoc" says: _should only be thrown **from within problem handlers**_ (bad ❌ ).

Do you suggest to use any other subclass of `RuntimeException`?

stephan-herrmann:
> @stephan-herrmann the exception is only thrown if the user requests to interrupt the thread so it shouldn't happen very often.

This is not meant as an excuse, is it? :)

> I also briefly considered using any of the exceptions in the package `org.eclipse.jdt.internal.compiler.problem` but I wasn't sure if they fit the use case: on the one side they are runtime exceptions (good ✅ ) but their "JavaDoc" says: _should only be thrown **from within problem handlers**_ (bad ❌ ).

Thanks for considering the documentation. However, looking at methods like `org.eclipse.jdt.internal.compiler.Compiler.reportProgress(String)` I feel that comment is outdated :)

I suggest to inspect some more incoming call paths, to see if AbortCompilation is caught at some level in every scenario.

fedejeanne:
I analyzed the call hierarchy of the method and I found out that `AbortCompilation` is pretty safe, it's being caught in many of the callers.

For reference, here is the complete call hierarchy:

image

---

And here I only left those methods which do **not** catch `AbortCompilation` (it might be that the catch it if I go deeper in the call stack, I just stopped after 4-5 levels):

image

---

Notice that I highlighted the same method in both screenshots (`org.eclipse.jdt.internal.compiler.parser.Parser.parse(ICompilationUnit, CompilationResult)`). It's the only one that doesn't immediately catch `AbortCompilation`


I'll prepare a PR that throws `AbortCompilation`

fedejeanne:
PR ready: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4451

stephan-herrmann:
> And here I only left those methods which do not catch AbortCompilation (it might be that the catch it if I go deeper in the call stack, I just stopped after 4-5 levels):

Thanks!

Some more observations:
* ClasspathMultiDirectory.acceptModuleInfo(ICompilationUnit, Parser) -> method is _unused_, candidate for deletion :sunglasses: 
* many paths originate from internal compiler code only, where it's probably safe to assume that AbortCompilation is caught near the top.
* also where HierarchyResolver, MatchLocator, CompletionEngine, SelectionEngine or CompilationUnitProblemFinder are the original callers, this looks OK since those classes do catch AbortCompilation, too. I cannot say if they do so an all paths, but any inconsistency there would be a bug in that specific class, I'd say.

fedejeanne:
> ClasspathMultiDirectory.acceptModuleInfo(ICompilationUnit, Parser) -> method is _unused_, candidate for deletion 😎

Good catch, I totally missed that. Thank you! --> #4455

PR #4313, opened 2025-08-14
Add import module support to ImportRewrite

### LINKED ISSUE #4298: [Java 24] Add support for import module to ImportRewriter
The current ImportRewriter does not support adding/removing import module statements.

jjohnstn:
- keep a map of modules and exported packages
- when deciding whether an import is needed, check the module package lists
- modify ImportName to support module
- add support for writing out an import module statement to ImportDeclarationWriter
- fixes #4298

## What it does

Adds import module support to ImportRewrite and friends so that Organize Imports can support module imports.

## How to test

To test will require the JDT UI patch as well which will be pushed soon.  A self-contained test will be added once it is working.  The JDT UI patch has a test that is working.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/rewrite/describing/ImportRewrite25Test.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/rewrite/describing/ImportRewrite25Test.java
@@ -13,7 +13,8 @@
  *******************************************************************************/
 package org.eclipse.jdt.core.tests.rewrite.describing;
 
-import java.io.IOException;
+import static org.eclipse.jdt.core.tests.rewrite.describing.StringAsserts.assertEqualStringIgnoreDelim;
+
 import java.lang.reflect.Method;
 import java.util.List;
 import junit.framework.Test;
@@ -170,10 +171,6 @@ public class ImportRewrite25Test extends AbstractJavaModelTests {
 		assertTrue(actualType.isParameterizedType());
 		apply(rewrite);
 		assertEqualStringIgnoreDelim(cu.getSource(), contents);
-	}
-
-	private void assertEqualStringIgnoreDelim(String actual, String expected) throws IOException {
-		StringAsserts.assertEqualStringIgnoreDelim(actual, expected);
 	}
 
 	private ImportRewrite newImportsRewrite(ICompilationUnit cu, String[] order, int normalThreshold, int staticThreshold, boolean restoreExistingImports) throws CoreException, BackingStoreException {
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/rewrite/describing/ImportRewrite25Test.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/rewrite/describing/ImportRewrite25Test.java
@@ -1,0 +1,205 @@
+/*******************************************************************************
+ * Copyright (c) 2014, 2025 IBM Corporation and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *		IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package org.eclipse.jdt.core.tests.rewrite.describing;
+
+import java.io.IOException;
+import java.lang.reflect.Method;
+import java.util.List;
+import junit.framework.Test;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.Path;
+import org.eclipse.jdt.core.ICompilationUnit;
+import org.eclipse.jdt.core.IJavaProject;
+import org.eclipse.jdt.core.IPackageFragmentRoot;
+import org.eclipse.jdt.core.ITypeRoot;
+import org.eclipse.jdt.core.JavaCore;
+import org.eclipse.jdt.core.dom.AST;
+import org.eclipse.jdt.core.dom.ASTParser;
+import org.eclipse.jdt.core.dom.CompilationUnit;
+import org.eclipse.jdt.core.dom.FieldDeclaration;
+import org.eclipse.jdt.core.dom.ITypeBinding;
+import org.eclipse.jdt.core.dom.IVariableBinding;
+import org.eclipse.jdt.core.dom.Type;
+import org.eclipse.jdt.core.dom.TypeDeclaration;
+import org.eclipse.jdt.core.dom.VariableDeclarationFragment;
+import org.eclipse.jdt.core.dom.rewrite.ImportRewrite;
+import org.eclipse.jdt.core.formatter.DefaultCodeFormatterConstants;
+import org.eclipse.jdt.core.tests.model.AbstractJavaModelTests;
+import org.eclipse.jdt.internal.compiler.classfmt.ClassFileConstants;
+import org.eclipse.jdt.internal.compiler.impl.CompilerOptions;
+import org.eclipse.jface.text.BadLocationException;
+import org.eclipse.jface.text.Document;
+import org.eclipse.text.edits.MalformedTreeException;
+import org.eclipse.text.edits.TextEdit;
+import org.osgi.service.prefs.BackingStoreException;
+
+
+@SuppressWarnings("rawtypes")
+public class ImportRewrite25Test extends AbstractJavaModelTests {
+
+
+	private static final Class THIS= ImportRewrite25Test.class;
+	private static final String PROJECT = "ImportRewrite25TestProject";
+
+	protected IPackageFragmentRoot sourceFolder;
+	protected IJavaProject project;
+
+	public ImportRewrite25Test(String name) {
+		super(name);
+	}
+
+	public static Test allTests() {
+		return new Suite(THIS);
+	}
+
+	public static Test suite() {
+		return allTests();
+	}
+
+	@Override
+	protected void setUp() throws Exception {
+		super.setUp();
+
+		IJavaProject proj= createJavaProject(PROJECT, new String[] {"src"}, new String[] {"JCL_25_LIB"}, "bin", "25");
+		proj.setOption(DefaultCodeFormatterConstants.FORMATTER_TAB_CHAR, JavaCore.SPACE);
+		proj.setOption(DefaultCodeFormatterConstants.FORMATTER_TAB_SIZE, "4");
+		proj.setOption(JavaCore.COMPILER_COMPLIANCE, JavaCore.VERSION_25);
+		proj.setOption(JavaCore.COMPILER_PB_ASSERT_IDENTIFIER, JavaCore.ERROR);
+		proj.setOption(JavaCore.COMPILER_SOURCE, JavaCore.VERSION_25);
+		proj.setOption(JavaCore.COMPILER_CODEGEN_TARGET_PLATFORM, JavaCore.VERSION_25);
+		proj.setOption(JavaCore.COMPILER_PB_ENABLE_PREVIEW_FEATURES, JavaCore.ENABLED);
+		proj.setOption(JavaCore.COMPILER_PB_REPORT_PREVIEW_FEATURES, JavaCore.IGNORE);
+		proj.setOption(DefaultCodeFormatterConstants.FORMATTER_NUMBER_OF_EMPTY_LINES_TO_PRESERVE, String.valueOf(99));
+
+		proj.setOption(DefaultCodeFormatterConstants.FORMATTER_BLANK_LINES_BETWEEN_IMPORT_GROUPS, String.valueOf(1));
+
+		setupExternalJCL("jclMin25");
+		JavaCore.setClasspathVariables(
+				new String[] {"CONVERTER_JCL_25_LIB", "CONVERTER_JCL_25_SRC", "CONVERTER_JCL_25_SRCROOT"},
+				new IPath[] {getConverterJCLPath("25"), getConverterJCLSourcePath("25"), getConverterJCLRootSourcePath()},
+				null);
+
+		this.project= proj;
+		this.sourceFolder = getPackageFragmentRoot(PROJECT, "src");
+
+		waitUntilIndexesReady();
+	}
+	protected static int getJLS25() {
+		return AST.JLS25;
+	}
+	@Override
+	protected void tearDown() throws Exception {
+		deleteProject(PROJECT);
+		super.tearDown();
+	}
+
+	// From Java 24 onwards, we will keep the jclMin*jar and convertJclMin*jar one and same
+	// The /JCL/build.xml has been updated to produce only jclMin*.jar
+	private String jclMinName(String compliance) {
+		long jdkLevel = CompilerOptions.versionToJdkLevel(compliance);
+		return (jdkLevel >= ClassFileConstants.JDK24) ? "jclMin" : "converterJclMin";
+	}
+	protected IPath getConverterJCLPath() {
+		return getConverterJCLPath(CompilerOptions.getFirstSupportedJavaVersion()); //$NON-NLS-1$
+	}
+
+	protected IPath getConverterJCLPath(String compliance) {
+		String jarName = jclMinName(compliance);
+		return new Path(getExternalPath() + jarName + compliance + ".jar"); //$NON-NLS-1$
+	}
+
+	protected IPath getConverterJCLSourcePath() {
+		return getConverterJCLSourcePath(CompilerOptions.getFirstSupportedJavaVersion()); //$NON-NLS-1$
+	}
+
+	protected IPath getConverterJCLSourcePath(String compliance) {
+		String jarName = jclMinName(compliance);
+		return new Path(getExternalPath() + jarName + compliance + "src.zip"); //$NON-NLS-1$
+	}
+
+	protected IPath getConverterJCLRootSourcePath() {
+		return new Path(""); //$NON-NLS-1$
+	}
+
+	public void testImportClassInJavaBase_since_25() throws Exception {
+		String contents = """
+			package pack1;
+			import module java.base;
+			public class X{
+				List<String> a;
+			}
+			""";
+		createFolder("/" + PROJECT + "/src/pack1");
+		createFile("/" + PROJECT + "/src/pack1/X.java", contents);
+
+		ASTParser parser = ASTParser.newParser(getJLS25());
+
+		ICompilationUnit cu = getCompilationUnit("/" + PROJECT + "/src/pack1/X.java");
+		parser.setSource(contents.toCharArray());
+		parser.setProject(this.project);
+		parser.setKind(ASTParser.K_COMPILATION_UNIT);
+		parser.setUnitName("X.java");
+		parser.setResolveBindings(true);
+		parser.setStatementsRecovery(true);
+		CompilationUnit astRoot = (CompilationUnit) parser.createAST(null);
+		Method setTypeRoot= CompilationUnit.class.getDeclaredMethod("setTypeRoot", ITypeRoot.class);
+		setTypeRoot.setAccessible(true);
+		setTypeRoot.invoke(astRoot, cu);
+		TypeDeclaration type= (TypeDeclaration) astRoot.types().get(0);
+		FieldDeclaration [] fields =  type.getFields();
+		FieldDeclaration field = fields[0];
+		List<VariableDeclarationFragment> fragments= field.fragments();
+		VariableDeclarationFragment fragment= fragments.get(0);
+		IVariableBinding varBinding= fragment.resolveBinding();
+		ITypeBinding typeBinding= varBinding.getType();
+		ImportRewrite rewrite = newImportsRewrite((ICompilationUnit) astRoot.getJavaElement(), new String[0], 99, 99, true);
+		Type actualType = rewrite.addImport(typeBinding, astRoot.getAST());
+		assertEquals("List<String>", actualType.toString());
+		assertTrue(actualType.isParameterizedType());
+		apply(rewrite);
+		assertEqualStringIgnoreDelim(cu.getSource(), contents);
+	}
+
+	private void assertEqualStringIgnoreDelim(String actual, String expected) throws IOException {
+		StringAsserts.assertEqualStringIgnoreDelim(actual, expected);
+	}
+
+	private ImportRewrite newImportsRewrite(ICompilationUnit cu, String[] order, int normalThreshold, int staticThreshold, boolean restoreExistingImports) throws CoreException, BackingStoreException {
+		ImportRewrite rewrite= ImportRewrite.create(cu, restoreExistingImports);
+		rewrite.setImportOrder(order);
+		rewrite.setOnDemandImportThreshold(normalThreshold);
+		rewrite.setStaticOnDemandImportThreshold(staticThreshold);
+		return rewrite;
+	}
+
+	protected ImportRewrite newImportsRewrite(CompilationUnit cu, String[] order, int normalThreshold, int staticThreshold, boolean restoreExistingImports) {
+		ImportRewrite rewrite= ImportRewrite.create(cu, restoreExistingImports);
+		rewrite.setImportOrder(order);
+		rewrite.setOnDemandImportThreshold(normalThreshold);
+		rewrite.setStaticOnDemandImportThreshold(staticThreshold);
+		return rewrite;
+	}
+
+	private void apply(ImportRewrite rewrite) throws CoreException, MalformedTreeException, BadLocationException {
+		TextEdit edit= rewrite.rewriteImports(null);
+
+		// not the efficient way!
+		ICompilationUnit compilationUnit= rewrite.getCompilationUnit();
+		Document document= new Document(compilationUnit.getSource());
+		edit.apply(document);
+		compilationUnit.getBuffer().setContents(document.get());
+	}
+
+}
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/rewrite/describing/ImportRewrite25Test.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/rewrite/describing/ImportRewrite25Test.java
@@ -173,6 +173,45 @@ public class ImportRewrite25Test extends AbstractJavaModelTests {
 		assertEqualStringIgnoreDelim(cu.getSource(), contents);
 	}
 
+	public void testImportListClassNoImportModule_since_25() throws Exception {
+		String contents = """
+			package pack1;
+			public class X{
+				List<String> a;
+			}
+			""";
+		createFolder("/" + PROJECT + "/src/pack1");
+		createFile("/" + PROJECT + "/src/pack1/X.java", contents);
+
+		ASTParser parser = ASTParser.newParser(getJLS25());
+
+		ICompilationUnit cu = getCompilationUnit("/" + PROJECT + "/src/pack1/X.java");
+		parser.setSource(contents.toCharArray());
+		parser.setProject(this.project);
+		parser.setKind(ASTParser.K_COMPILATION_UNIT);
+		parser.setUnitName("X.java");
+		parser.setResolveBindings(true);
+		parser.setStatementsRecovery(true);
+		CompilationUnit astRoot = (CompilationUnit) parser.createAST(null);
+		Method setTypeRoot= CompilationUnit.class.getDeclaredMethod("setTypeRoot", ITypeRoot.class);
+		setTypeRoot.setAccessible(true);
+		setTypeRoot.invoke(astRoot, cu);
+		ImportRewrite rewrite = newImportsRewrite((ICompilationUnit) astRoot.getJavaElement(), new String[0], 99, 99, true);
+		String actualType = rewrite.addImport("java.util.List<String>");
+		assertEquals("List<String>", actualType);
+		apply(rewrite);
+		String expected = """
+				package pack1;
+
+				import java.util.List;
+
+				public class X{
+					List<String> a;
+				}
+				""";
+		assertEqualStringIgnoreDelim(cu.getSource(), expected);
+	}
+
 	private ImportRewrite newImportsRewrite(ICompilationUnit cu, String[] order, int normalThreshold, int staticThreshold, boolean restoreExistingImports) throws CoreException, BackingStoreException {
 		ImportRewrite rewrite= ImportRewrite.create(cu, restoreExistingImports);
 		rewrite.setImportOrder(order);
--- a/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
+++ b/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
@@ -363,7 +363,7 @@ public final class ImportRewrite {
 						List<ImportDeclaration> astImports= compilationUnit.imports();
 						ImportDeclaration foundModuleImport= null;
 						for (ImportDeclaration astImport : astImports) {
-							if (Flags.isModule(astImport.getFlags()) && astImport.getName().getFullyQualifiedName().equals(currName)) {
+							if (Modifier.isModule(astImport.getModifiers()) && astImport.getName().getFullyQualifiedName().equals(currName)) {
 								foundModuleImport= astImport;
 								break;
 							}
@@ -375,7 +375,11 @@ public final class ImportRewrite {
 							}
 						}
 					}
-					moduleEntries.put(curr.getElementName(), packageNames);
+					String name= curr.getElementName();
+					if (name.endsWith(".*")) { //$NON-NLS-1$
+						name= name.substring(0, name.length() - 2);
+					}
+					moduleEntries.put(name, packageNames);
 				}
 			}
 		}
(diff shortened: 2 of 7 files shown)


jjohnstn:
@stephan-herrmann Here is the BETA_JAVA25 patch.  I will post the JDT UI patch shortly and work on getting the test to work in JDT Core (the JDT UI test for Organize Imports works fine).

stephan-herrmann:
> @stephan-herrmann Here is the BETA_JAVA25 patch. I will post the JDT UI patch shortly and work on getting the test to work in JDT Core (the JDT UI test for Organize Imports works fine).

It would of course be great if you'd add also some tests in jdt.core. Perhaps the given failures provide some hints about where and how to create such test? :smile:

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
```diff
@@ -337,15 +340,64 @@ public static ImportRewrite create(ICompilationUnit cu, boolean restoreExistingI
 			throw new IllegalArgumentException("Compilation unit must not be null"); //$NON-NLS-1$
 		}
 		List existingImport= null;
+		Map<String, List<String>> moduleEntries= null;
```
stephan-herrmann, line 336: unlike jdt.ui, jdt.core "traditionally" uses a space on each side of `=`.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
```diff
@@ -337,15 +340,64 @@ public static ImportRewrite create(ICompilationUnit cu, boolean restoreExistingI
...
+			CompilationUnit compilationUnit= null;
 			for (IImportDeclaration curr : imports) {
-				char prefix= Flags.isStatic(curr.getFlags()) ? STATIC_PREFIX : NORMAL_PREFIX;
-				existingImport.add(prefix + curr.getElementName());
+				char prefix= Flags.isStatic(curr.getFlags()) ? STATIC_PREFIX :
+					Flags.isModule(curr.getFlags()) ? MODULE_PREFIX : NORMAL_PREFIX;
+				String currName= curr.getElementName();
+				if (currName.endsWith(".*")) { //$NON-NLS-1$
+					currName= currName.substring(0, currName.length() - 2);
+				}
```
stephan-herrmann: This addition seems to affect normal imports, too. Is that intended? Is it fixing a bug?

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
```diff
@@ -337,15 +340,64 @@ public static ImportRewrite create(ICompilationUnit cu, boolean restoreExistingI
...
+				if (Flags.isModule(curr.getFlags())) {
+					List<String> packageNames= new ArrayList<>();
+					if (compilationUnit == null) {
+						compilationUnit= convertICUtoCU(cu);
+					}
+					if (compilationUnit != null) {
+						List<ImportDeclaration> astImports= compilationUnit.imports();
+						ImportDeclaration foundModuleImport= null;
+						for (ImportDeclaration astImport : astImports) {
+							if (astImport.getName().getFullyQualifiedName().equals(currName)) {
```
stephan-herrmann: Shouldn't `astImport` first be checked if its a module import? A module might share the name of a package.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
```diff
@@ -337,15 +340,64 @@ public static ImportRewrite create(ICompilationUnit cu, boolean restoreExistingI
...
+				}
 			}
 		}
-		return new ImportRewrite(cu, null, existingImport);
+		return new ImportRewrite(cu, null, existingImport, moduleEntries);
+	}
+
+	private static List<String> getPackageNames(IModuleBinding binding) {
+		List<String> result= new ArrayList<>();
+		IPackageBinding[] packageBindings= binding.getExportedPackages();
```
stephan-herrmann, line 426: [JEP 511](https://openjdk.org/jeps/511) also supports indirect exports (via `requires transitive`).

Also the case of qualified exports (`exports p to m`) should be considered.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
```diff
@@ -917,6 +996,18 @@ private static ITypeBinding normalizeTypeBinding(ITypeBinding binding) {
 		return null;
 	}
 
+
+	/**
+	 * Adds a new module import to the rewriter's record.
+	 * @since 3.43
```
stephan-herrmann: Could you provide a bit more text here?

Also by expecting clients to provide the list of packages, they will have to replicate the logic from `getPackageNames()`. 
Perhaps a binding-based API is more suitable here? (I haven't yet looked at the UI part).

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
```diff
@@ -1369,6 +1464,20 @@ public String[] getCreatedStaticImports() {
 		return this.createdStaticImports;
 	}
 
+	/**
+	 * Returns all new module imports created by the last invocation of {@link #rewriteImports(IProgressMonitor)}
+	 * or <code>null</code> if these methods have not been called yet.
+	 * <p>
+	 * Note that this list doesn't need to be the same as the added static imports ({@link #getAddedStaticImports()}) as
```
stephan-herrmann: Why does this mention `getAddedStaticImports()`?

org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/rewrite/imports/OnDemandComputer.java
```diff
@@ -86,15 +86,17 @@ Collection<OnDemandReduction> identifyPossibleReductions(
 	private Map<ImportName, Collection<ImportName>> mapByContainer(Collection<ImportName> imports) {
 		Map<ImportName, Collection<ImportName>> importsByContainer = new HashMap<>();
 		for (ImportName importName : imports) {
-			ImportName containerOnDemand = importName.getContainerOnDemand();
+			if (!importName.isModule) {
```
stephan-herrmann, line 89: Does this indicate, you don't yet consider reducing overlap between module imports and regular ones?

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
jjohnstn: There is lots of code in ImportRewrite that follows this.  In my particular patch, there are lines that surround my changes that have the UI style.  If you like, I can change all the ones I added.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
jjohnstn: No.  The element name for a module import adds .* which is wrong.  I added this check in various places to counter this.  In this case, I have added a check for prefix being MODULE_PREFIX.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
jjohnstn: Agree.  Fixed.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
jjohnstn: I have added recursive logic to find transitive exports.  I also do a recursive search of all required module names and then when I see an export x to m, I check if m is in the list of recursive module names.  If yes, I add it, if not, I don't.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
jjohnstn: Comment was copied and modified.  Fixed.

org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/rewrite/imports/OnDemandComputer.java
jjohnstn: This is already done in ImportRewrite.  This logic is for calculating on demand containers.  On demand imports disappear if they are handled by an import module and regular logic applies if they are not (e.g. grouping n imports into an on demand import).

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
jjohnstn: Currently, the code in Organize imports keeps a map of module names and the package names they export.  So, the only caller right now will already have the list of packages.  After your comment to extend the logic of getPackageNames(), I made it public and renamed it to getPackageNamesForModule() so Organize imports will use it instead of using the getExportedPackages() call directly from the binding.  In the future, I might see a clean-up wanting to add an import module by name in which case I will add a new API that will use the ASTParser to get a binding and hence the list.  I have updated the docs.

jjohnstn:
> > @stephan-herrmann Here is the BETA_JAVA25 patch. I will post the JDT UI patch shortly and work on getting the test to work in JDT Core (the JDT UI test for Organize Imports works fine).
> 
> It would of course be great if you'd add also some tests in jdt.core. Perhaps the given failures provide some hints about where and how to create such test? 😄

Still working on it.  I have a test based on ImportRewriteTest18 but I keep running into a set up problem whereby it doesn't set up the environment/classpath properly and the test is compiled at a lower level so it is full of errors and doesn't have the bindings set.

I believe the build test failures you are seeing are due to my removing the .* without checking for module first.  With the recent change, the test runs successfully locally.  I have rebased my code and will force push a refresh with the changes as mentioned in the responses.  I will continue work on getting the test running properly.  I might ask for you to look at it if I can't make any in-roads.

Thanks for reviewing.

jjohnstn:
Hi @stephan-herrmann figured out why test wasn't working.  Had a typo in the project set-up.  Anyway, I have added the test and made a few fixes based on it.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
stephan-herrmann: I wasn't quite sure, what your "No" referred to, but the additional check resolves my worries alright :)

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
stephan-herrmann: thanks, detailed comments to follow

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
```diff
@@ -337,15 +340,104 @@ public static ImportRewrite create(ICompilationUnit cu, boolean restoreExistingI
...
+						result.add(packageBinding.getName());
+						break;
+					}
+				}
+			} else {
+				result.add(packageBinding.getName());
 			}
 		}
-		return new ImportRewrite(cu, null, existingImport);
+		IModuleBinding[] requiredBindings= binding.getRequiredModules();
```
stephan-herrmann: This should ask for `requires transitive`, but then, alas, I see that DOM's IModuleBinding lacks a necessary API: See that the compiler's `ModuleBinding`  as separate `getRequires()` and `getRequiresTransitive()`.  The latter one is not surfaced in DOM, so that API would need to be added first.

From a quick glance it looks like `getRequiredModules()` doesn't even contain the transitive ones. This additionally spawned #4350.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
```diff
@@ -337,15 +340,104 @@ public static ImportRewrite create(ICompilationUnit cu, boolean restoreExistingI
...
+		}
+	}
+
+	private static Set<String> getPackageNames(IModuleBinding binding, Set<String> moduleNames) {
+		Set<String> result= new HashSet<>();
+		IPackageBinding[] packageBindings= binding.getExportedPackages();
+		for (IPackageBinding packageBinding : packageBindings) {
+			if (binding.getExportedTo(packageBinding).length > 0) {
+				for (String moduleName : binding.getExportedTo(packageBinding)) {
+					if (moduleNames.contains(moduleName)) {
```
stephan-herrmann: I think this should rather check if the package is exported to "the current module", not those in `moduleNames`.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
```diff
@@ -337,15 +340,104 @@ public static ImportRewrite create(ICompilationUnit cu, boolean restoreExistingI
...
+	 * @since 3.43
+	 */
+	public static List<String> getPackageNamesForModule(IModuleBinding binding) {
+		Set<String> moduleNames= new HashSet<>();
+		populateModuleNames(binding, moduleNames);
+		return new ArrayList<>(getPackageNames(binding, moduleNames));
+	}
+
+	private static void populateModuleNames(IModuleBinding binding, Set<String> moduleNames) {
+		moduleNames.add(binding.getName());
```
stephan-herrmann: It would be good to check the result of `add()` so we can avoid infinite recursion in the case of (illegal) cycles.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
```diff
@@ -337,15 +340,104 @@ public static ImportRewrite create(ICompilationUnit cu, boolean restoreExistingI
...
+			populateModuleNames(requiredModule, moduleNames);
+		}
+	}
+
+	private static Set<String> getPackageNames(IModuleBinding binding, Set<String> moduleNames) {
+		Set<String> result= new HashSet<>();
+		IPackageBinding[] packageBindings= binding.getExportedPackages();
+		for (IPackageBinding packageBinding : packageBindings) {
+			if (binding.getExportedTo(packageBinding).length > 0) {
+				for (String moduleName : binding.getExportedTo(packageBinding)) {
```
stephan-herrmann: The result of `getExportedTo()` is worth storing in a local, to avoid double execution.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
```diff
@@ -917,6 +1036,42 @@ private static ITypeBinding normalizeTypeBinding(ITypeBinding binding) {
...
 
+
+	/**
+	 * Adds a new module import to the rewriter's record.  The import will register all exported
+	 * package names (direct or transitive) so the rewriter can eliminate all other imports covered
+	 * by these package names.
+	 *
+	 * @param name - name of module to import
+	 * @param moduleBinding - binding of module to import
+	 * @return name if import added or null if there are issues
```
stephan-herrmann, line 1057: similar methods return either a simple (when import was successfully added) or a qualified name (otherwise), ready for clients to use in references to this type. Would that make sense also here?

org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/rewrite/imports/OnDemandComputer.java
stephan-herrmann: So for existing functionality this looks good.

Do you have plans to parallel the OnDemandComputer with a ModuleComputer that allows collapsing several regular imports into a new module import? _Would that make sense?_

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/rewrite/describing/ImportRewrite25Test.java
```diff
@@ -0,0 +1,205 @@
...
+		Type actualType = rewrite.addImport(typeBinding, astRoot.getAST());
+		assertEquals("List<String>", actualType.toString());
+		assertTrue(actualType.isParameterizedType());
+		apply(rewrite);
+		assertEqualStringIgnoreDelim(cu.getSource(), contents);
+	}
+
+	private void assertEqualStringIgnoreDelim(String actual, String expected) throws IOException {
+		StringAsserts.assertEqualStringIgnoreDelim(actual, expected);
+	}
```
stephan-herrmann: Speaking of imports :smile: - wouldn't a static import make this method unnecessary?

org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/rewrite/imports/OnDemandComputer.java
jjohnstn: Interesting idea.  Perhaps a future clean-up or a preference to use import module.  I think we can cross that road later.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
jjohnstn: Ok. Thanks.

org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/rewrite/imports/OnDemandComputer.java
stephan-herrmann: Do you want to create a follow-up issue, so we can close and forget about this thread?

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
jjohnstn: Ok.  I had thought it was transitive but looking at 7.7.2 it says solely to the module specified.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
jjohnstn: Ok, makes sense.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
jjohnstn: Ok.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
jjohnstn: Not sure what you mean.  What would a caller want with the module name after import?  It is a different situation with the type.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/rewrite/describing/ImportRewrite25Test.java
jjohnstn: :) Yes.  I copied from ImportRewriteTest18.java and modified.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
stephan-herrmann: Sorry, I didn't look carefully at what name this is about. So scratch this question.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/rewrite/describing/ImportRewrite25Test.java
jjohnstn: Actually, it won't let me do a static import for StringAsserts.*

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
jjohnstn: I've actually changed the code to do a recursive search for module bindings instead of names.  It will need to change to use the new transitive requires when it is available.  I add all the direct requires from the initial module and then just add exports from each module binding in the populated list (no recursion needed).  I will add the check to the new code.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
stephan-herrmann: OK, your observation gives a hint at who created this class: Martin Aeschlimann, who was a key person in JDT/UI :)

So it doesn't make sense to insist on space before `=` in this class.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/rewrite/describing/ImportRewrite25Test.java
stephan-herrmann: > Actually, it won't let me do a static import for StringAsserts.*

What does it say when you try to do so? Who is our imports expert? ;-P

org.eclipse.jdt.core/dom/org/eclipse/jdt/internal/core/dom/rewrite/imports/OnDemandComputer.java
jjohnstn: I have created: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2450

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/rewrite/describing/ImportRewrite25Test.java
jjohnstn: I get an error about StringAsserts not used, but when I deleted the private method that called it and forced the code to use the imported static method, it worked.

jjohnstn:
Hi @stephan-herrmann When the new interface to get transitive requires is in, I'll update this patch to use it and that should address the remaining comments.

stephan-herrmann:
> Hi @stephan-herrmann When the new interface to get transitive requires is in, I'll update this patch to use it and that should address the remaining comments.

OK, we still have a few days ... let's see.

Is https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2436 in a state where I can test the overall effect (once the next set of changes is in)?

jjohnstn:
> > Hi @stephan-herrmann When the new interface to get transitive requires is in, I'll update this patch to use it and that should address the remaining comments.
> 
> OK, we still have a few days ... let's see.
> 
> Is [eclipse-jdt/eclipse.jdt.ui#2436](https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2436) in a state where I can test the overall effect (once the next set of changes is in)?

I just need to replace a call to getRequiredModules() to the new added API.  I'll do it right away.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
stephan-herrmann: The code looks better now, but I'll await your analysis if https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2436#issuecomment-3255547235 is caused right here or elsewhere.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
stephan-herrmann: This still doesn't look right. We need to check if a package is exported to the "current" module, i.e., the module enclosing the current compilation unit, where imports are being analyzed (possibly the unnamed module).

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
jjohnstn: Yes.  The fix for the a.b.B being removed is in ImportRewrite.  I had initially added all required modules to my module list which I get exports from.  This should only be the first module and its transitive requires which I have fixed.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
jjohnstn: Ok.  I'll fix this.

jjohnstn: Done

jjohnstn: Need another fix.  My logic for getting the module name is incorrect.

jjohnstn: Ok, fixed.  I ran your example where d exports to main.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
```diff
@@ -337,15 +341,114 @@ public static ImportRewrite create(ICompilationUnit cu, boolean restoreExistingI
...
+	 * @return a list of package names that are exported by the given binding
+	 *
+	 * @since 3.43
+	 */
+	public static List<String> getPackageNamesForModule(IModuleBinding binding, CompilationUnit cu) {
+		Set<IModuleBinding> modules= new HashSet<>();
+		String currentModuleName= null;
+		try {
+			IModuleDescription modDesc= cu.getJavaElement().getJavaProject().getModuleDescription();
+			currentModuleName= modDesc.getElementName();
```
stephan-herrmann: `null` is a legal return value from `getModuleDescription()`.

If that happens, we're in the unnamed module, and hence we cannot leverage any qualified exports (`to`).

Currently, 3 new tests in JDT/UI throw NullPointerException at this location.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
```diff
@@ -337,15 +341,114 @@ public static ImportRewrite create(ICompilationUnit cu, boolean restoreExistingI
...
+		Set<String> packageList= new HashSet<>();
+		for (IModuleBinding moduleBinding : modules) {
+			packageList.addAll(getPackageNames(moduleBinding, currentModuleName));
+		}
+		return new ArrayList<>(packageList);
+	}
+
+	private static void populateTransitiveModules(IModuleBinding binding, Set<IModuleBinding> transitiveModules) {
+		if (transitiveModules.add(binding)) {
+			IModuleBinding[] requiredTransitiveModules= ((IModuleBindingExtended)binding).getRequiredTransitiveModules();
```
stephan-herrmann: I'm curious about this cast, and what it means for the compatibility story.

Do we know for sure that all instance of `IModuleBinding` are also instances of `IModuleBindingExtended`?
* If not, then the cast must be guarded with `instanceof`.
* If yes, i.e., if both interfaces capture the same set of instances, then why do we need separate interfaces?

I guess the reasoning why the cast is safe lies in the `@noimplement` tag on `IModuleBinding`, which allows us to regard `ModuleBinding` as the only possible implementing class. From this we can follow that all instances also implement `IModuleBindingExtended`. But then, we could also directly add the new method to the existing interface, because this is actually safe for a `@noimplement` interface, or how could that break compatibility?

We do have precedents in DOM where a method was added to a `@noimplement` interface, whereas I cannot see any precedent for adding a new interface for just that same purpose.

Sorry, for raising this only now.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
```diff
@@ -337,15 +341,114 @@ public static ImportRewrite create(ICompilationUnit cu, boolean restoreExistingI
...
+					moduleEntries.put(name, packageNames);
+				}
+			}
+		}
+		return new ImportRewrite(cu, null, existingImport, moduleEntries);
+	}
+
+	/**
+	 * Calculate the list of package names that are exported by a module including
+	 * direct and transitive imports.
```
stephan-herrmann: I find "direct and transitive imports" confusing.  Transitivity applies to "read" edges between modules (from `requires transitive`), not to packages.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
```diff
@@ -337,15 +341,114 @@ public static ImportRewrite create(ICompilationUnit cu, boolean restoreExistingI
...
+	 * @since 3.43
+	 */
+	public static List<String> getPackageNamesForModule(IModuleBinding binding, CompilationUnit cu) {
+		Set<IModuleBinding> modules= new HashSet<>();
+		String currentModuleName= null;
+		try {
+			IModuleDescription modDesc= cu.getJavaElement().getJavaProject().getModuleDescription();
+			currentModuleName= modDesc.getElementName();
+		} catch (JavaModelException e) {
+			// should not happen - ignore
```
stephan-herrmann: "should not happen" is always a weak excuse :).
You might get away though with saying: if we're unable to retrieve a module, then we simply cannot benefit from any qualified exports, so we might as well proceed as if the current module is the unnamed module.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
```diff
@@ -393,13 +392,20 @@ public static ImportRewrite create(ICompilationUnit cu, boolean restoreExistingI
 	 * direct and transitive imports.
 	 *
 	 * @param binding - module binding
-	 * @param currentModuleName - name of the module for this program or null
+	 * @param cu - compilation unit
```
stephan-herrmann: I can see that passing `cu` is convenient, but your lack of explanation tells me, that it is not trivial to explain why a cu is relevant for this method. Which cu should clients pass? 

What's easier for you: to clearly motivate the role of the `cu` parameter, or change the type to something that can be easier explained? IMHO, any subexpression of `cu.getJavaElement().getJavaProject().getModuleDescription()` would equally qualify as the value to be provided by callers.

Note, that this is API and should be done considering also clients outside JDT.

jjohnstn: I have changed the 2nd parameter to be a JavaProject.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
jjohnstn: Changed comment.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
jjohnstn: Got rid of the stuff after "by a module"

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
jjohnstn: I'm happy to add it to the IModuleBinding if @noimplement covers it.

jjohnstn: I have created: https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4376  which is against BETA_JAVA25 branch.  If I should start making patches against master, let me know.

stephan-herrmann: I asked colleagues about a suitable deadline for code freeze https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4384#issuecomment-3270495011

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
stephan-herrmann: I assume you have local changes which you didn't yet push? Is that because you are waiting for a decision in #4376 ?

jjohnstn: Yes.

stephan-herrmann: go ahead :)

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/rewrite/ImportRewrite.java
jjohnstn: Ok, thx for the info.

stephan-herrmann:
I had intended to merge this before today's Y-build, but then got distracted.

This implies the UI-part can only be build successfully, after tomorrow's Y-build. This should still allow us to merge that UI-part in time for Java 25 GA.

@jjohnstn meanwhile I made the final merge master->BETA_JAVA25 in JDT/UI. To be on the safe side for tomorrow, you may want to rebase once more.

PR #4334, opened 2025-08-20
[WIP] Fix range of "forbidden reference" error for super constructor invocation

### LINKED ISSUE #4333: "Forbidden reference" error range for super constructor invocation
Most "forbidden reference" error ranges highlight just the method/constructor name.

However in this case, where `disallowed.pkg.Forbidden` is a forbidden type, the error marker for the invocation of the super constructor covers the entire statement (i.e. `super();`).

```java
import disallowed.pkg.Forbidden;
public class Faulty extends Forbidden {
  public Faulty() {
    super();
  }
}
```

Given this is inconsistent with the other "forbidden reference" ranges, I think this is a bug.

I have a fix locally for this.

datho7561:
## What it does
Fixes #4333

Address the bug by implementing the method to get the end offset of the name.

## How to test
See the reproducer in the linked issue. I will also add unit/integration tests.

## Author checklist


datho7561:
I think the test timeout is legitimate. I ran into it locally. I'll need to diagnose it.

mickaelistria:
I'm not sure the proposal here is better: the issue is not really the identifier (which depending on the args may resolve to different method, which might be restricted or not), but really the reference to the super method, and the arguments to matter in identifying the super method.

datho7561:
Well, it's inconsistent with all the other forbidden reference error ranges on methods, which only select the method name.

I'm guessing one rationale for selecting just the method name is to reduce visual clutter; if there's something unrelated that's wrong with one of the arguments, then there would be overlapping error ranges.

datho7561:
The test freeze is due to a UI prompt asking if `.classpath` file should be made writable during `ClasspathTests.testReadOnly`. ~~I think it's supposed to have thrown an exception instead of showing a UI prompt? It seems unrelated to my change, but it's definitely not present before my change.~~ ~~It's also present without my change.~~ ~~I don't see it on master? I'm confused.~~ It's gone

mickaelistria:
> Well, it's inconsistent with all the other forbidden reference error ranges on methods, which only select the method name.

ok

datho7561:
@jjohnstn another PR from me that'd I'd appreciate a review on

jjohnstn (approved):
Makes it consistent.  LGTM.

Issue #4350, opened 2025-08-26
Possible confusion about disjointness of #requires and #requiresTransitive


stephan-herrmann:
Looking at fields #requires and #requiresTransitive of ModuleBinding (compiler variant) I believe that those arrays are disjoint, but then `getRequiredModules(false)` would fail its purpose when used for `collectAllDependencies()`.

jjohnstn:
Hi @stephan-herrmann I added a new interface that ModuleBinding could implement so that it would be possible to get the requires transitive modules.  I added a new interface to avoid a major version bump.  In testing, it was reporting the transitive module in the regular requires list so I think it works as you want it to.

stephan-herrmann:
The ticket title "_Possible confusion about disjointness of #requires and #requiresTransitive_" doesn't bear up under scrutiny. Those fields are _not_ disjoint, the latter is a subset of the former, see how those are populated in 
* BinaryModuleBinding.catchPartsFrom()
* ModuleDeclaration.resolveModuleDirectives()

This well matches to what clients need.

More likely it was just my own self from last week that was confused. 

Since we already have a PR to match this ticket I will opportunistically adjust title and description to match that PR 😄

stephan-herrmann:
This issue has not been closed simply because github autoclosing doesn't work on the branch, but as of https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4313#discussion_r2327891090 I believe we should not release the new interface but fold the new method back into `IModuleBinding`. @jjohnstn  wdyt?

jjohnstn:
@stephan-herrmann IModuleBinding is a public interface so adding a new method is not allowed unless we do a major version bump.  Does the @noimplement let us off the hook?

stephan-herrmann:
> [@stephan-herrmann](https://github.com/stephan-herrmann) IModuleBinding is a public interface so adding a new method is not allowed unless we do a major version bump. Does the @noimplement let us off the hook?

Yes, I think `@noimplement` gives us license to add methods (unless the new method would overload an existing method). What could get broken by that addition? Only implementing classes that fail to implement that new method would get broken, but such classes are already illegal outside jdt.core due to that tag. Illegal meaning: they are operating on their own risk.

stephan-herrmann:
Marking as closed by https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4376

PR #4365, opened 2025-08-30
Add interface to fetch requires transitive from module binding

jjohnstn:
- add new IModuleBindingExtended interface
- modify ModuleBinding to implement new interface
- add test to ASTConverter9Test
- for #4350 and #4313

## What it does

See commit.

## How to test

See new test.

## Author checklist


org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverter9Test.java
```diff
@@ -1630,5 +1630,108 @@ public void testRequiresDirectiveNameIsMandatory() throws JavaModelException, Co
...
 
+	public void testTransitiveRequires_1() throws Exception { //https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4350
+		try {
+
+			IJavaProject project1 = createJavaProject("ConverterTests9", new String[] {"src"}, new String[] {jcl9lib}, "bin", "9");
+			project1.open(null);
+			addClasspathEntry(project1, JavaCore.newContainerEntry(new Path("org.eclipse.jdt.MODULE_PATH")));
+			String fileContent =
+				"module first {\n" +
+				"    requires transitive second;\n" +
```
stephan-herrmann, line 1641: with only one requires statement the test doesn't prove that `transitive` is actually considered.

Please add at least one non-transitive `requires` statements to demonstrate the difference.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/IModuleBindingExtended.java
```diff
@@ -0,0 +1,30 @@
...
+ *******************************************************************************/
+package org.eclipse.jdt.core.dom;
+
+/**
+ * @since 3.43
+ */
+public interface IModuleBindingExtended {
+
+	/**
+	 * Returns all required transivite modules.
```
stephan-herrmann: typo.

I'm also not sure would should be the best wording. Certainly there is no such concept as a "transitive module".

Speaking of "transitively required modules" is also misleading, since transitivity will only happen at the site of clients of this module.

JLS is quite shy as it only says "_The requires keyword may be followed by the modifier transitive. This causes any module which requires the current module to have an implicitly declared dependence on the module specified by the requires transitive directive._"

Hence I suggest a similarly syntactic wording "modules required with the modifier 'transitive'".

stephan-herrmann:
BTW, should we also extend the API of elements like `RequiresDirective`? I don't see any `resolveBinding()` methods as are customary in DOM.

I guess `org.eclipse.jdt.internal.ui.text.correction.proposals.AddModuleRequiresCorrectionProposalCore.createAddRequiresChange()` could use such addition, once we start to consider `transitive` also there (or in the code that creates that Change).

WDYT?

stephan-herrmann:
Build failure indicates we need https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4366

So please rebase before adding any changes in response to my review.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/IModuleBindingExtended.java
jjohnstn: Ok, will do.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverter9Test.java
jjohnstn: Ok, will do.  I thought that because the required and required transitive lists were different that that was enough.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/IModuleBindingExtended.java
jjohnstn: Done

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverter9Test.java
stephan-herrmann: thanks

jjohnstn:
> BTW, should we also extend the API of elements like `RequiresDirective`? I don't see any `resolveBinding()` methods as are customary in DOM.
> 

Not sure it is needed.  We can get the binding from resolving the module.

> I guess `org.eclipse.jdt.internal.ui.text.correction.proposals.AddModuleRequiresCorrectionProposalCore.createAddRequiresChange()` could use such addition, once we start to consider `transitive` also there (or in the code that creates that Change).
> 
> WDYT?

Yes, we definitely could add a second call that adds a requires transitive.

stephan-herrmann:
> Yes, we definitely could add a second call that adds a requires transitive.

I was thinking more the opposite direction: travel transitive requirements to check if a module/package is already available, to avoid adding a redundant requires.

jjohnstn:
> > Yes, we definitely could add a second call that adds a requires transitive.
> 
> I was thinking more the opposite direction: travel transitive requirements to check if a module/package is already available, to avoid adding a redundant requires.


Ok.  That makes sense.

PR #4376, opened 2025-09-08
Move requires transitive API into IModuleBinding

jjohnstn:
- remove new IModuleBindingExtended
- put new method from IModuleBindingExtended into IModuleBinding
- change ASTConverter9Test appropriately

## What it does

Moves the new requires transitive APIs directly into IModuleBinding instead of a new interface.

## How to test

Run ASTConverter9Test.

## Author checklist


stephan-herrmann:
@jarthana @iloveeclipse may I ask a second opinion from either of you?

Question being: is it legal to add a method to an interface that is marked `@noimplement`, without a major version bump? 

In my understanding `@noimplement` protects us from any clients claiming this to be an incompatible change. Of course a client class implementing `IModuleBinding` would be broken when an abstract method is added, needing to be implemented by that client class. But due to `@noimplement` that client class would be illegal to begin with. 

I'm a missing any other scenarios of incompatibility?

If we can't come to a clear answer, we may have to revert #4365 for now, to give us more time to figure out the best course of action.

stephan-herrmann:
First build failed with unresolvable tycho extensions.

Second build had unrelated test failures.

Keep trying.

stephan-herrmann:
OK, I'll merge to unblock @jjohnstn  :)

If someone has objections, we still have a few days to revert, or fix otherwise.

jjohnstn:
Thanks @stephan-herrmann I have updated the ImportRewrite and OrganizeImports patches.

Issue #1303, opened 2023-08-22
Incomplete Type Index / Bug 109084


fedejeanne:
### Current Behavior

In some cases (with unknown reasons), the type index does not contain all available types, so the type search does not find some types that actually exist.
Rebuilding the index via the preferences does not resolve the issue.

![Image](https://user-images.githubusercontent.com/755472/248801293-e2893b9a-13cd-4958-860f-96718b3121ce.jpg)

The bug has been existing for a long time and does not seem to be fixed by now: https://bugs.eclipse.org/bugs/show_bug.cgi?id=109084 

### Expected Behavior

In the best case, the type index should properly contain the expected types.
At least, the "rebuild index" operation in the preferences should properly rebuild the index.

### Existing Workaround
Delete <workspace>\.metadata\.plugins\org.eclipse.jdt.core\*.index, and restart Eclipse.

iloveeclipse:
Current bug report is not actionable, provided links do not contain additional info that allows to draw any conclusions. 

Please provide steps to reproduce, and/or error logs, and/or project which "miss" some types in the Java search.

fedejeanne:
> Current bug report is not actionable, provided links do not contain additional info that allows to draw any conclusions.
> 
> Please provide steps to reproduce, and/or error logs, and/or project which "miss" some types in the Java search.

I know, this is an old issue that we're also facing and we're also having trouble to reproduce it, but it happens rather often (exactly like other users reported in the original issue: https://bugs.eclipse.org/bugs/show_bug.cgi?id=109084). I'm trying to gather enough information to assess the error and hopefully be able to reproduce accurately. 

It's probably going to be a long journey but I hope I can gather some useful information by logging the exceptions that happen during the indexing (and by letting others also report theirs, see #1304)

iloveeclipse:
@fedejeanne : note, I've moved almost all of JDT core "system out / printStackTrace" code to tracing via https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1504.

This was always a big issue in JDT (not properly use tracing), hope this helps here.

fedejeanne:
Thank you very much @iloveeclipse ! This is much appreciated 😃

jukzi:
This issue is tagged needinfo for months - is it done or not relevant anymore?

fedejeanne:
It's been neglected but the issue (probably) still exists. I'll put it in my personal backlog and will come back to it in the coming weeks.

PR #4388, opened 2025-09-11
Log exceptions when indexing #1303

fedejeanne:
## What it does

Don't rely on the `VERBOSE` configuration to be enabled, bring the errors to the user so they are visible in the _Error Log_ view and the error log files too.

Contributes to
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1303

### DIFF
--- a/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/IndexAllProject.java
+++ b/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/IndexAllProject.java
@@ -13,8 +13,6 @@
  *******************************************************************************/
 package org.eclipse.jdt.internal.core.search.indexing;
 
-import static org.eclipse.jdt.internal.core.JavaModelManager.trace;
-
 import java.io.IOException;
 import java.net.URI;
 import java.util.HashSet;
@@ -36,7 +34,6 @@ import org.eclipse.jdt.internal.compiler.util.SimpleLookupTable;
 import org.eclipse.jdt.internal.core.ClasspathEntry;
 import org.eclipse.jdt.internal.core.JavaProject;
 import org.eclipse.jdt.internal.core.index.Index;
-import org.eclipse.jdt.internal.core.search.processing.JobManager;
 import org.eclipse.jdt.internal.core.util.Util;
 
 public class IndexAllProject extends IndexRequest {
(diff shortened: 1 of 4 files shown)


org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/IndexAllProject.java
```diff
@@ -230,9 +227,7 @@ public boolean visit(IResourceProxy proxy) throws CoreException {
 			// request to save index when all cus have been indexed... also sets state to SAVED_STATE
 			this.manager.request(new SaveIndex(this.containerPath, this.manager));
 		} catch (CoreException | IOException e) {
-			if (JobManager.VERBOSE) {
-				trace("-> failed to index " + this.project + " because of the following exception:", e); //$NON-NLS-1$ //$NON-NLS-2$
-			}
+			Util.log(e, "Failed to index " + this.project + " because of the following exception:"); //$NON-NLS-1$ //$NON-NLS-2$
```
iloveeclipse: Here and below: please change error messages so that they make sense (they were written for tracing on system out, where exception stack followed the error line)
Something like `"Failed to index " + this.project + " : " + e.getMessage()`

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/SaveIndex.java
```diff
@@ -43,9 +42,7 @@ public boolean execute(IProgressMonitor progressMonitor) {
 			monitor.enterWrite(); // ask permission to write
 			this.manager.saveIndex(index);
 		} catch (IOException e) {
-			if (JobManager.VERBOSE) {
-				trace("-> failed to save index " + this.containerPath + " because of the following exception:", e); //$NON-NLS-1$ //$NON-NLS-2$
-			}
+			Util.log(Status.warning("Failed to save index " + this.containerPath + " because of the following exception:", e)); //$NON-NLS-1$ //$NON-NLS-2$
```
iloveeclipse: To be consistent it should be not a warning but error, so no need for a new status.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/IndexAllProject.java
fedejeanne: Done

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/SaveIndex.java
fedejeanne: I actually intended it like that since the difference between an error while _creating_ the index vs an error while _saving_ the index is that even if the index is not _saved_, it is still created and it is used _i.e._ the classes in the index are found.

I created 2 projects and let them fail intentionally:
- `failttoindex` (1) fails while creating the index
- `failttosaveindex` (2) fails while saving the index

As you can see, I can find `MyClass` inside `failttosaveindex` but I can't find the one inside `failtoindex`:

image

I have adapted the text to make this more understandable to the user.

fedejeanne: Is that ok @iloveeclipse ?

fedejeanne:
Thank you!

PR #4395, opened 2025-09-15
Strange "heisenbug" error in Apache Lucene codebase when opened in Eclipse regarding sealed types

### LINKED ISSUE #4331: Strange "heisenbug" error in Apache Lucene codebase when opened in Eclipse regarding sealed types
Apache Lucene internally uses ECJ for additional compilation checks like better error/warnings during its linting phase, so we are sure that our general setup and enabled/disabled warnings fit our project.

But there is one strange thing: If you run "gradlew eclipse" and later open the project in Eclipse IDE, the code compiles fine (most of the time), but under certain circumstances (totally unknown, therefore a Heisenbug) it fails with one fatal error:

> Permitted type LeafReaderContext does not declare org.apache.lucene.index.IndexReaderContext as a direct supertype

But actually this is not correct:

IndexReaderContext.java:
```java
public abstract sealed class IndexReaderContext permits CompositeReaderContext, LeafReaderContext {
```

LeafReaderContext.java:
```java
public final class LeafReaderContext extends IndexReaderContext {
```

CompositeReaderContext:
```java
public final class CompositeReaderContext extends IndexReaderContext {
```

As you see its all perfectly valid code. Interestingly it only complains about `LeafReaderContext`. All three classes are in same package next to each other.

There are also sisters `IndexReader`, `CompositeReader`, `LeafReader` that have same structure, but there the error does not show up.

Language level is Java 24 in main branch of Lucene but error also happens in Java 21 (branch_10x) of project. We can't reproduce this in a smaller subset, to reproduce, checkout main branch of Apache Lucene (https://github.com/apache/lucene), execute `gradlew eclipse` and then open the project in Eclipse IDE:

Image

Interestingly, when compiling with ECJ inside Gradle it does not fail:

```
$ gradlew clean :lucene:core:ecjLintMain

[Incubating] Problems report is available at: file:///C:/Users/Uwe%20Schindler/Projects/lucene/lucene/build/reports/problems/problems-report.html

BUILD SUCCESSFUL in 35s
54 actionable tasks: 42 executed, 12 up-to-date
```

It is a heisenbug because the error sometmes dissappears if you add a bit of whitespace and safe one of the three affected files. It reappears after you clean the workspace in Eclipse and after recompilation it reappears.

This makes running developing and running tests in Ecliupse a problem, because the three classes are one of the central classes so almost everything depends on it, so the compilation failure can't be ignored.

We are not sure if this is a bug in ECJ or the Eclipse IDE itsself, but as the error message comes from EDT's ECJ it must be some problem in the compiler.

srikanth-sankaran:
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4331

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ClassScope.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ClassScope.java
@@ -1217,27 +1217,6 @@ public class ClassScope extends Scope {
 						continue nextPermittedType;
 					}
 
-					if (sourceType.isClass()) {
-						ReferenceBinding superClass = permittedType.superclass();
-						superClass = superClass == null ? null : superClass.actualType();
-						if (!TypeBinding.equalsEquals(sourceType, superClass))
-							problemReporter().sealedClassNotDirectSuperClassOf(permittedType, permittedTypeRef, sourceType);
-					} else if (sourceType.isInterface()) {
-						ReferenceBinding[] superInterfaces = permittedType.superInterfaces();
-						boolean hierarchyOK = false;
-						if (superInterfaces != null) {
-							for (ReferenceBinding superInterface : superInterfaces) {
-								superInterface = superInterface == null ? null : superInterface.actualType();
-								if (TypeBinding.equalsEquals(sourceType, superInterface)) {
-									hierarchyOK = true;
-									break;
-								}
-							}
-							if (!hierarchyOK)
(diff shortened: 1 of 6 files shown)


stephan-herrmann:
what's the relationship of the PR to the merged #4409?

srikanth-sankaran:
> what's the relationship of the PR to the merged #4409?

#4409 is identical to this. It was created because this one got caught up in a Jenkins update process glitch I don't know all the details but the "checks" kept failing mysteriously
(according to @iloveeclipse it had something to do with PR first being created with a certain version of infra software and subsequent commits being pushed after an update to the infra software.)

 My attempts to restart the checks didn't resolve so started over with a fresh PR. 

So this one should be closed as is.

PR #4408, opened 2025-09-18
Trace warning about ZIP files being constantly opened

fedejeanne:
## What it does

Write it to the trace logs by default and to the system output only if `TRACE_TO_STDOUT` is set to true. 

Same as the other `trace` method.


fedejeanne:
Failing checks are unrelated: no changes made to the `org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/DubiousOutcomeTest.java`

iloveeclipse (approved):

iloveeclipse:
Thanks.

PR #4415, opened 2025-09-19
Add support for ambiguous type checking for import modules

### LINKED ISSUE #4414: Allow Organize Imports to determine if type is ambiguous over multiple packages/modules
To properly determine if a type import is needed when it would otherwise be ambiguous across one or more modules, Organize Imports needs to be able to query a PackageBinding for an unqualified type name.  There is a commented out method: findTypeBinding() in IPackageBinding which would solve the issue so uncomment and implement this for PackageBinding.

In concert, add a new API in ImportRewrite to provide the transitive list of package bindings for a module.

jjohnstn:
- add new getPackagesForModule() static method to ImportRewrite and change getPackageNamesForModule() to use it
- modify IPackageBinding to add findType() method that takes a simple name
- add new code to PackageBinding and RecoveredPackageBinding to implement findType()
- add new test to ASTConverter9Test
- fixes #4414

## What it does

See issue.

## How to test

See new test.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/PackageBinding.java
+++ b/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/PackageBinding.java
@@ -267,6 +267,9 @@ class PackageBinding implements IPackageBinding {
 
 	@Override
 	public ITypeBinding findTypeBinding(String nameString) {
+		if (nameString == null || nameString.isEmpty()) {
+			return null;
+		}
 		String[] segments = nameString.split("[.]"); //$NON-NLS-1$
 		int i = 0;
 		org.eclipse.jdt.internal.compiler.lookup.PackageBinding pkgBinding = this.binding;


jjohnstn:
This is needed by: https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2502

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/IPackageBinding.java
```diff
@@ -63,21 +63,22 @@ public interface IPackageBinding extends IBinding {
...
-//	public ITypeBinding findTypeBinding(String name);
+	/**
+	 * Finds and returns the binding for the class or interface with the given
+	 * name declared in this package.
+	 * <p>
+	 * For top-level classes and interfaces, the name here is just the simple
+	 * name of the class or interface. For nested classes and interfaces, the
+	 * name is the VM class name (in other words, a name like
+	 * <code>"Outer$Inner"</code> as used to name the class file; see
+	 * <code>ITypeBinding.getName</code>).
```
stephan-herrmann: nowadays a `{@link }` tag would be more helpful.

But when I check the linked documentation I don't even see any mention of inner classes. Which variant is correct? Inside `TypeBinding.getName()` I even see a call to `getSourceName()`, which conflicts with the above.

As this method promises to handle inner classes, it would be good if that aspect would also be asserted by the new test. That would also give clarity of the format of names :)

stephan-herrmann (changes requested):
mostly unproblematic, only the issue of inner classes needs to be clarified.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/IPackageBinding.java
jjohnstn: In this case, it doesn't seem to be documented, perhaps because trying to get an inner class results in a ProblemReferenceBinding which has the type binding as a closest match.  It is Outer$Outer$Inner for first inner class and then Outer$Outer$Inner$Inner2 for a subsequent inner class.  I have documented it in IPackageBinding.  I have also added logic to look for the ProblemReferenceBinding and the nested code and if the closest match is valid.  I also added logic to the new test case to get the first Inner type and it's inner type.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverter9Test.java
```diff
@@ -1630,6 +1631,100 @@ public void testRequiresDirectiveNameIsMandatory() throws JavaModelException, Co
...
+					"module second {\n" +
+					"    exports pack22 to first;\n" +
+					"}";
+			createFile("/second/src/module-info.java",	secondFile);
+			createFolder("/second/src/pack22");
+			createFile("/second/src/pack22/I22.java",
+					"package pack22;\n" +
+					"public interface I22 {}\n");
+			createFile("/second/src/pack22/Class22.java",
+					"pakage pack22;\n" +
```
stephan-herrmann: The typo on this line is the reason for the strange name pattern (duplicate name of toplevel type) you have observed.

jjohnstn: Thanks.  I could have stared at that forever and not seen it.  I have rewritten findTypeBinding() as suggested to go through each segment starting with packages and then using ReferenceBinding.getMemberType() and have decided to use "." as the separator.  The test case has been adjusted accordingly to not repeat the top class in the specifier.

jjohnstn:
Hi @stephan-herrmann I think everything has been dealt with.  Can this be merged?  I have some other minor changes needed to ImportRewrite (remove module support and fixing the loading of current imports to not add .* for modules).  I have the add module imports clean-up ready and need this patch plus the extra stuff just mentioned.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/PackageBinding.java
```diff
@@ -263,4 +264,42 @@ org.eclipse.jdt.internal.compiler.lookup.PackageBinding getCompilerBinding() {
...
+		while (i < segments.length) {
+			char[] segName = segments[i++].toCharArray();
+			foundBinding = pkgBinding.getTypeOrPackage(segName, this.binding.enclosingModule, false);
+			if (foundBinding == null || !foundBinding.isValidBinding()
+					|| !(foundBinding instanceof org.eclipse.jdt.internal.compiler.lookup.PackageBinding foundPackageBinding)) {
+				break;
+			} else {
+				pkgBinding = foundPackageBinding;
+			}
+		}
```
stephan-herrmann: I don't think this loop is needed to implement the updated contract (thanks btw).
Just call `getTypeOrPackage()` once and bail out if `foundBinding` is not a valid `ReferenceBinding`. Am I missing anything?

jjohnstn: I am trying to find a type but I also want to support member types.  If I have a.b.c.D.E, then I need to find a.b.c.D, then find E from the reference binding.  If I call with just a.b.c.D.E, I'll get an error because member types aren't supported.  So, I try one segment at a time until I cross from package binding to reference binding.  Then I am able to find member types using the ReferenceBinding.  I do understand how it is possible to handle all the cases with just one call.

stephan-herrmann: You have two loops, but only the second loop is needed for your purpose.

jjohnstn: Maybe I don't understand.  The 2nd loop is only after we have found the ReferenceBinding by checking some subset of all the segments provided.  The segment count i continues from the first loop to the 2nd so the 2nd loop cannot proceed until the first loop has stopped finding packages.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/IPackageBinding.java
jjohnstn: Resolving this as documentation has been added.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/PackageBinding.java
stephan-herrmann: Here's how you could potentially convince me:
* Show me a test case that runs the first loop in more than one iteration, and
* Demonstrate that this test case is within the specification which we agreed upon

:)

jjohnstn: I finally see what I was missing.  The code will never be asked to find a sub-package so the logic isn't needed.  I have modified the code to remove the first loop.

stephan-herrmann (approved):
Looks good now

jjohnstn:
> Looks good now

Thanks Stephan for your review and patience :)

Issue #4416, opened 2025-09-19
[Java 25] Bogus error: The blank final field o may not have been initialized


kurt-2:
Code below compiles ok with javac 25, Eclipse 4.37 (+BETA 25) Java compilance set to 24,
but not with Eclipse 4.37 (+BETA 25) Java compilance set to 25:

> The blank final field o may not have been initialized	WrongNotInitialized.java	/J25/src/test	line 13	Java Problem

which points to 'o.toString()'

```
package test;

public class WrongNotInitialized {
  private final Object o;

  public WrongNotInitialized(final Object o) {
    super();
    this.o = o;
  }

  public WrongNotInitialized() {
    this(new Object());
    System.out.println(o.toString());
  }
}
```

princec:
Was just coming here to report same issue (even more simple test case below, showing that type and access modifiers don't appear to be a factor):

```java
class Test {

	final int x;

	Test(int x) {
		this.x = x;
	}

	Test(Test a) {
		this(a.x);
		//this(1);

		System.out.println(x);
	}
}
```

I made an extra discovery - if this(a.x) is not run, but instead this(1) (commented out in the example), then there is no problem with accessing x.

Albex:
Got the same issue using a record:
```java
public record TestRecord(String a)
{
    public TestRecord(int b)
    {
        this(String.valueOf(b));
        System.out.println(a.length());
    }
}
```

iloveeclipse:
@srikanth-sankaran : could you please check? Seem to be a popular bug with our Java 25 compiler...

srikanth-sankaran:
I'll follow up.

trancexpress:
For whatever reason I see this only with `--release 25`, at least in the case where we run into the bug. E.g.:

```
$ cat Test.java


public class Test {

        private final Object r;

        Test() {
                this(Integer.valueOf(1));
                r.hashCode();
        }

        Test(Object r) {
                this.r = r;
        }
}
$ /usr/lib/jvm/java-25/bin/javac Test.java
$ /usr/lib/jvm/java-25/bin/javac --release 25 Test.java
$ /usr/lib/jvm/java-25/bin/java -jar  .../3.44_I20250919-0110/ecj/ecj.jar Test.java
$ /usr/lib/jvm/java-25/bin/java -jar  .../3.44_I20250919-0110/ecj/ecj.jar --release 25 Test.java
----------
[...]

PR #4432, opened 2025-09-24
[Java 25] Bogus error: The blank final field o may not have been initialized

srikanth-sankaran:
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4416

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ConstructorDeclaration.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ConstructorDeclaration.java
@@ -111,6 +111,16 @@ public void analyseCode(ClassScope classScope, InitializationFlowContext initial
 			constructorContext = this.prologueContext;
 			flowInfo = this.prologueInfo.addInitializationsFrom(flowInfo);
 			// skip the part already done during PROLOGUE analysis ...
+			if (this.constructorCall != null) {
+				if (this.constructorCall.accessMode == ExplicitConstructorCall.This) {
+					FieldBinding[] fields = this.binding.declaringClass.fields();
+					for (FieldBinding field : fields) {
+						if (!field.isStatic()) {
+							flowInfo.markAsDefinitelyAssigned(field);
+						}
+					}
+				}
+			}
 		} else {
 			flowInfo.setReachMode(initialReachMode);
 
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/InitializationTests.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/InitializationTests.java
@@ -18,11 +18,15 @@ package org.eclipse.jdt.core.tests.compiler.regression;
 
 import java.util.Map;
 import junit.framework.Test;
+import org.eclipse.jdt.internal.compiler.classfmt.ClassFileConstants;
 import org.eclipse.jdt.internal.compiler.impl.CompilerOptions;
 
 @SuppressWarnings({ "unchecked", "rawtypes" })
 public class InitializationTests extends AbstractRegressionTest {
 
+static {
+//	TESTS_NAMES = new String [] { "testIssue4416" };
+}
 public InitializationTests(String name) {
 		super(name);
 }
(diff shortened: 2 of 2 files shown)


srikanth-sankaran:
This is work in progress. My understanding of this feature of flexible constructor bodies is still rudimentary. 

This PR will fail (at least) two tests  `org.eclipse.jdt.core.tests.compiler.regression.SuperAfterStatementsTest.testGH3748b()`  and `org.eclipse.jdt.core.tests.compiler.regression.InitializationTests.testIssue4416d()` perhaps more.

srikanth-sankaran:
The captured expected output of `org.eclipse.jdt.core.tests.compiler.regression.SuperAfterStatementsTest.testGH3748b()` looks suspect. [EDIT: This is a red herring - the revise patch does not fail this test, there is some curious difference in where the error is reported between javac and ECJ, but that is not super relevant to the issue at hand]

Given:

```
				public class X {
					final int fin1;
					final int fin2;
					{
						fin1 = 0;
						fin2 = 1;
					}
					X() {
						this(fin1 = 10);
						fin2 = 11;
					}
					X(int x) {}
				}
```

we report:

			----------
			1. ERROR in X.java (at line 5)
				fin1 = 0;
				^^^^
			The final field fin1 may already have been assigned
			----------
			3. ERROR in X.java (at line 10)
				fin2 = 11;
				^^^^
			The final field fin2 may already have been assigned
			----------

while javac reports:

X.java:9: error: variable fin1 might already have been assigned
                                                this(fin1 = 10);
                                                    ^
X.java:10: error: variable fin2 might already have been assigned
                                                fin2 = 11;
                                                ^
2 errors

I think javac reports are correct and eclipse has a problem here.

srikanth-sankaran:
The three failing tests are: 

org.eclipse.jdt.core.tests.compiler.regression.InitializationTests.testIssue4416d
org.eclipse.jdt.core.tests.compiler.regression.NullTypeAnnotationTest.testBug445227
org.eclipse.jdt.core.tests.compiler.regression.SuperAfterStatementsTest.testGH3748b

srikanth-sankaran:
> I think javac reports are correct and eclipse has a problem here.

May be not!

12.5 Creation of New Class Instances states:

Just before a reference to the newly created object is returned as the result, the
indicated constructor is processed to initialize the new object using the following
procedure:
1. Assign the arguments for the constructor to newly created parameter variables
for this constructor invocation.
2. If this constructor does not contain an explicit constructor invocation (§8.8.7.1)
then continue from step 5.
3. Execute the BlockStatements, if any, of the prologue of the constructor body.
If execution of any statement completes abruptly, then execution of the
constructor completes abruptly for the same reason, otherwise continue with
the next step.
4. The explicit constructor invocation is either an invocation of another
constructor in the same class (using this) or an invocation of a superclass
constructor (using super). Evaluate the arguments of the constructor
invocation and process the constructor invocation recursively using these
same seven steps. If the constructor invocation completes abruptly, then this
procedure completes abruptly for the same reason. Otherwise, continue from
step 7 if the invocation is of another constructor in the same class, and continue
from step 6 if the invocation is of a superclass constructor.
5. If this constructor is for a class other than Object, then this constructor
contains an implicit invocation of a superclass constructor with no arguments.
In this case, process the implicit constructor invocation recursively using these
same seven steps. If that constructor invocation completes abruptly, then this
procedure completes abruptly for the same reason, otherwise continue with the
next step.
6. Execute the instance initializers and instance variable initializers for this class,
assigning the values of instance variable initializers to the corresponding
instance variables, in the left-to-right order in which they appear textually in
the source code for the class. If execution of any of these initializers results
in an exception, then no further initializers are processed and this procedure
completes abruptly with that same exception, otherwise continue with the next
step.
7. Execute the BlockStatements, if any, of the epilogue of this constructor. If
execution of any statement completes abruptly, then this procedure completes
abruptly for the same reason. Otherwise, this procedure completes normally.

Upon carefully reading the above, it would appear ECJ's flagging of the error in line 5 is more appropriate than javac's flagging the error at line 9.

srikanth-sankaran:
OK, I have a better understanding of the defect. The crux of the problem seems to be that we analyze the epilog of a `this(...)` chaining constructor without the initializations performed by the chained constructor. Revised PR will follow shortly.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ConstructorDeclaration.java
```diff
@@ -111,6 +111,16 @@ public void analyseCode(ClassScope classScope, InitializationFlowContext initial
...
+			if (this.constructorCall != null) {
+				if (this.constructorCall.accessMode == ExplicitConstructorCall.This) {
+					FieldBinding[] fields = this.binding.declaringClass.fields();
+					for (FieldBinding field : fields) {
+						if (!field.isStatic()) {
+							flowInfo.markAsDefinitelyAssigned(field);
+						}
+					}
+				}
+			}
```
srikanth-sankaran, line 123: May be we could extract this block into a method - it features in two places (at least)

srikanth-sankaran:
@stephan-herrmann - Thanks in advance for your review.

@iloveeclipse @trancexpress - you are welcome to try this already to see if it addresses the problem @trancexpress reported

iloveeclipse:
@stephan-herrmann : I would be very grateful if you could spend some time today or tomorrow with the goal to get the fix into the nightly build tomorrow. We are trying to move to Java 25 and this is a blocker for us, as compiler shokes now on a trivial non-Java 25 code.

iloveeclipse:
> you are welcome to try this already

I usually run our "default" internal SDK build & update from master every friday morning, so let's hope the PR will be reviewed & merged before.

stephan-herrmann:
@srikanth-sankaran  semantically this fix is spot-on. 

Only your comment regarding code duplication led me to some experiments resulting in the slight variant #4435 - where instead of duplication I simply move one block out the the big else block. While preparing that PR I also updated code comments a bit.

Feel free to merge either PR.

stephan-herrmann:
> where instead of duplication I simply move one block out the the big else block.

this describes the [first commit](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4435/commits/eeda0a557b2a3e621a9897974280c0dc0c1874ea), before I found and [fixed](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4435/commits/d09d1bdbe4431ad3ebe3a27d036856cd1bce00b3) another flavor of this bug.

srikanth-sankaran:
> @srikanth-sankaran semantically this fix is spot-on.
> 
> Only your comment regarding code duplication led me to some experiments resulting in the slight variant #4435 - where instead of duplication I simply move one block out the the big else block. While preparing that PR I also updated code comments a bit.
> 
> Feel free to merge either PR.

Thanks!

Looks like the other PR has some failing tests ATM. So I will go ahead and merge this one - which addresses failures in building _old code that **doesn't use** flexible constructor bodies_ in source/compliance level 25 and the other PR which includes additional fixes for code that _actually has a prologue_ (and so uses Java 25 features) can come in on its own. 

This will unblock right away at least 4 users who have reported problems in https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4416

stephan-herrmann:
> Looks like the other PR has some failing tests ATM.

All are like

```
org.eclipse.jdt.core.tests.compiler.regression.InitializationTests.testIssue4416_withPrologue - 1.8
...
The Java feature 'Flexible Constructor Bodies' is only available with source level 25 and above
```

PR #4435, opened 2025-09-24
[Java 25] Bogus error: The blank final field o may not have been initialized

stephan-herrmann:
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4416

### DIFF
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/InitializationTests.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/InitializationTests.java
@@ -514,6 +514,8 @@ public void testIssue4416() {
 	});
 }
 public void testIssue4416_withPrologue() {
+	if (this.complianceLevel < ClassFileConstants.JDK25)
+		return; // uses flexible constructor bodies
 	runConformTest(new String[] {
 			"WrongNotInitialized.java",
 			"""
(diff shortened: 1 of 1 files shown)


stephan-herrmann:
I found one more problem ... when we add a prologue before any of the `this()` calls in new tests, we get `The blank final field o may not have been initialized`.

stephan-herrmann:
The code is not perfectly clear about the boundaries between PROLOGUE and REST and this peculiar block representing the effects of an explicit `this()` call.

In `PROLOGUE` mode we should stop processing at either of these points:
* when `constructorCall` is set, stop right after analysing that call, don't look at any `statements`
* otherwise assume that `lateConstructorCall` is contained in statements, and stop right after analysing this call

This block marking fields as assigned is relevant only for `ALL` and `REST` modes. In particular those inits must not leak via `prologueInfo` into TypeDeclaration's `nonStaticFieldInfo` and thus back into `REST`  analysis of all constructors.

According to the two points where `PROLOGUE` processing should stop, there are also two points where fields should be marked during `ALL` or `REST` analysis:
* when `constructorCall` is set, right before processing `statements`
* otherwise right after  finding `lateConstructorCall` within `statements` (the point where `REST` will actually start analysing - `ALL` is not applicable in this case)

srikanth-sankaran (approved):
Looks good.

iloveeclipse:
Thank you @srikanth-sankaran & @stephan-herrmann for the fast fix!

We will verify it tomorrow in our application tests. Should anything be wrong with the next SDK build I will report.

PR #4441, opened 2025-09-26
Log exceptions when indexing #1303

fedejeanne:
Similar to #4388 (but this time for the indexing of JARs/JRTs), do not rely on the `VERBOSE` flag, let the exceptions come to the error log and to the UI instead for better debugging.

Notice that I logged some exceptions as _warnings_ since the operation will be retried _i.e._ it's recoverable.

Contributes to https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1303

### DIFF
--- a/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/IndexManager.java
+++ b/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/IndexManager.java
@@ -896,9 +896,7 @@ public synchronized Index recreateIndex(IPath containerPath) {
 		return index;
 	} catch (IOException e) {
 		// The file could not be created. Possible reason: the project has been deleted.
-		if (VERBOSE) {
-			trace("-> failed to recreate index for path: "+containerPathString, e); //$NON-NLS-1$
-		}
+		Util.log(Status.warning("Failed to recreate index for path: " + containerPathString)); //$NON-NLS-1$
 		return null;
 	}
 }
@@ -1100,9 +1098,7 @@ public synchronized boolean resetIndex(IPath containerPath) {
 		return true;
 	} catch (IOException e) {
 		// The file could not be created. Possible reason: the project has been deleted.
-		if (VERBOSE) {
-			trace("-> failed to reset index for path: "+containerPathString, e); //$NON-NLS-1$
-		}
+		Util.log(Status.warning("Failed to reset index for path: " + containerPathString)); //$NON-NLS-1$
 		return false;
 	}
 }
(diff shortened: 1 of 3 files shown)


org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/IndexManager.java
```diff
@@ -896,9 +896,7 @@ public synchronized Index recreateIndex(IPath containerPath) {
 		return index;
 	} catch (IOException e) {
 		// The file could not be created. Possible reason: the project has been deleted.
-		if (VERBOSE) {
-			trace("-> failed to recreate index for path: "+containerPathString, e); //$NON-NLS-1$
-		}
+		Util.log(Status.warning("Failed to recreate index for path: " + containerPathString)); //$NON-NLS-1$
```
fedejeanne, line 899: Note: if you check the call hierarchy of the method, you'll notice that this ends up returning `false` in these jobs:

image

Which means it is a recoverable error, ergo I just posted a _warning_.

fedejeanne:
Would it be possible to review and merge this one too, please?

@trancexpress since Andrey said he is going to be busy in the next few weeks, maybe you could take a peek? This PR is very similar to the one Andrey already merged: #4388 .

trancexpress:
@fedejeanne what is the user expected to do with the log entries?

Generally the indexer runs in parallel with user operations, its possible it runs into errors due to e.g. deleted files. What does the user do with the log in that case?

fedejeanne:
@trancexpress if errors occurs then:
1. It is an indication that some process may be blocking a file -> That process might have to be terminated or at least the user might try and file a report saying that Eclipse keeps getting blocked by other processes and it affects the indexing
2. The user will know that functionality might be affected e.g. classes might not be found, class hierarchies might be inconsistent, search results too, etc.

This would give the user a heads-up and more insight as to why some things might be wrong in the views (hierarchies, searches, etc).

The goal of the PR is to bring the errors to the surface without the need to activate the `VERBOSE` mode. These kind of errors shouldn't happen very often (so the user should not be flooded by them) but if they do, the user should know it, know what to expect and maybe even take action.

fedejeanne:
@trancexpress after some consideration, I decided to first try and see if these errors really don't happen that often and if they are actionable. As you pointed out, there is not much to do when an error occurs since the indexes should try to rebuild themselves anyway and the user would only be annoyed if these are only informational messages.

I rolled out the tracing preference internally by adding these 2 lines in the `.options` file next to `eclipse.exe`.

```
org.eclipse.jdt.core/debug=true
org.eclipse.jdt.core/debug/indexmanager=true
```
Our centralized logs should start picking up these messages.

This solves the issue for me at the moment so there is no need to change the code.

Thank you for your hint, @trancexpress !

PR #26, opened 2022-04-19
Bug 575436 - rework JavaModelManager.getZipFile()

jukzi:
### High level overview
on the client side:

```
	JavaModelManager manager = JavaModelManager.getJavaModelManager();
	try {
		zip = manager.getZipFile(jarPath);
		...
	} finally {
		manager.closeZipFile(zip);
	}
```
to be replaced by

```
	JavaModelManager manager = JavaModelManager.getJavaModelManager();
	try (ThreadLocalZipFile zip = manager.getZipFile(jarPath)) {
	...
	}
```
which allows resource checking and prevents left open Zip files.
... and actually at least one possible **resource leak** was discovered and **fixed**.

### Details
Moved code from JavaModelManager to new seperate Classes:
*ThreadLocalZipFiles: general caching for ZipFiles
*ZipState: java model handling of invalid ZipFiles

"throwIoExceptionsInGetZipFile" was not moved but deleted to avoid
testcode in production code. Therefore
testTransitionFromInvalidToValidJar() was adapted to produce a real
invalid jar.

ThreadLocalZipFiles contains now AutoCloseable Helper for
try-with-resource resource management.

Calls to the refactored code (in "jdt.core") where refactored to use new
the new try-with-resource pattern.

There is a little difference between the new behavior of old
JavaModelManager.getZipFile() during a JavaModelManager.cacheZipFiles()
block() and new ZipState.createZipFile() in a
ThreadLocalZipFiles.createZipHolder() block:
When blocks are nested they now return the same instance on any nesting
level while previously new instance of ZipFile had been created on every
nesting level.

There is a little difference between the new behavior of old
JavaModelManager.flushZipFiles and new ZipState.close():
Previously flushZipFiles did close all zipFiles in the innermost block;
Now ZipState.close() does only close a zipFile if it was not allocated
in a outer block.

I.e. we have now well defined try-with-resource block: During execution
of any block there is only a single instance of ZipFile for each file.

To use this new behavior builder.ClasspathJar (and its subclasses)
needed to be updated to not hold a local instance of ZipFile (which
could have been leaked).

--
JavaModelManager.getLocalFile(IPath) did return a wrong (relative) path
for deleted workspace resource. Did not matter much as the file did not
exist anyway but it destroyed JavaSearchBugsTests.testBug261722() when
caching filename instead of IPath.

--
In AddJarFileToIndex there is a big block indent that git-diff doesnt
show as such.

Change-Id: Idd94ae90b375db3e475432e710ae7babb99cc5ac

PR #4445, opened 2025-09-27
Enhance Indexer performance

### LINKED ISSUE #4444: Enhance Indexer performance
I have found two issues related to indexer performance:

1. AddJarFileToIndex doesn't use zip cache. See 
- https://github.com/eclipse-jdt/eclipse.jdt.core/blob/0dbe2967235238fb5375924d17a13bcd9f9f6483/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/AddJarFileToIndex.java#L151
- https://github.com/eclipse-jdt/eclipse.jdt.core/blob/0dbe2967235238fb5375924d17a13bcd9f9f6483/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/AddJarFileToIndex.java#L158

2. SourceIndexer creates a new JavaSearchNameEnvironment for each source file https://github.com/eclipse-jdt/eclipse.jdt.core/blob/0dbe2967235238fb5375924d17a13bcd9f9f6483/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/SourceIndexer.java#L186
We would need to create one JavaSearchNameEnvironment for a project.

I have created a PR and tested it using the [spring-boot](https://github.com/spring-projects/spring-boot) project. The Rebuild Index action takes 10-20% less when using the zip and name environment cache.

I have tested the Rebuild Index action using the following patch:
```
diff --git a/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/processing/JobManager.java b/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/processing/JobManager.java
index 1f28bb2fa1..f46f1b1465 100644
--- a/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/processing/JobManager.java
+++ b/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/processing/JobManager.java
@@ -482,6 +482,7 @@ public abstract class JobManager {
 				}
 			}
 			this.progressJob = null;
+			long start = 0;
 			while (getProcessingThread() != null) {
 				try {
 					IJob job;
@@ -513,6 +514,7 @@ public abstract class JobManager {
 						if (cacheZipFiles) {
 							JavaModelManager.getJavaModelManager().flushZipFiles(this);
 							cacheZipFiles = false;
+							System.out.println("RebuildIndex Took: " + (System.currentTimeMillis() - start ));
 						}
 						// just woke up, delay before processing any new jobs, allow some time for the active thread to finish
 						synchronized (this.idleMonitor) {
@@ -535,10 +537,17 @@ public abstract class JobManager {
 							this.progressJob = pJob;
 						}
 						if (!cacheZipFiles) {
+							start = System.currentTimeMillis();
 							JavaModelManager.getJavaModelManager().cacheZipFiles(this);
 							cacheZipFiles = true;
 						}
+						long s1 = System.currentTimeMillis();
+						// System.out.println("Job: " + job + " Start ");
 						job.execute(null); // may enqueue a new job
+						long took = System.currentTimeMillis() - s1;
+						if (took > 2000) {
+							System.out.println("Job: " + job + " Took: " + took);
+						}
 					} finally {
 						this.executing = false;
 						if (VERBOSE) {

```

The related issue - https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2501 @jukzi  The PR reduces the number of source file reads.

snjeza:
Fixes #4444

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java
@@ -131,10 +131,12 @@ public class JavaModelManager implements ISaveParticipant, IContentTypeChangeLis
 	 */
 	static class ZipCache {
 		private final Map<Object, ZipFile> map;
+		private final Map<Object, Long> lastModifiers;
 		Object owner;
 
 		ZipCache(Object owner) {
 			this.map = new HashMap<>();
+			this.lastModifiers = new HashMap<>();
 			this.owner = owner;
 		}
 
@@ -159,7 +161,12 @@ public class JavaModelManager implements ISaveParticipant, IContentTypeChangeLis
 			return this.map.get(path);
 		}
 
-		public void setCache(IPath path, ZipFile zipFile) {
+		public long lastModified(IPath path) {
+			Long result = this.lastModifiers.get(path);
+			return result == null ? 0 : result;
+		}
+
+		public void setCache(IPath path, ZipFile zipFile, long lastModified) {
 			try (ZipFile old = this.map.put(path, zipFile)) {
 				if (old != null) {
 					if (JavaModelManager.ZIP_ACCESS_VERBOSE) {
@@ -168,10 +175,26 @@ public class JavaModelManager implements ISaveParticipant, IContentTypeChangeLis
 								+ "].setCache()] leaked ZipFile on " + old.getName() + " for path: " + path); //$NON-NLS-1$ //$NON-NLS-2$
 					}
 				}
+				this.lastModifiers.put(path, lastModified);
 			} catch (IOException e) {
 				if (VERBOSE) {
 					trace("", e); //$NON-NLS-1$
 				}
+			}
+		}
+
+		public void clearCache(IPath path) {
+			ZipFile zipFile = this.map.remove(path);
+			if (zipFile != null) {
+				if (JavaModelManager.ZIP_ACCESS_VERBOSE) {
+					Thread currentThread = Thread.currentThread();
+					trace("(" + currentThread + ") [ZipCache[" + this.owner //$NON-NLS-1$//$NON-NLS-2$
+							+ "].clearCache()] ZipFile on " + zipFile.getName() + " for path: " + path); //$NON-NLS-1$ //$NON-NLS-2$
+				}
+			}
+			Long lastModifier = this.lastModifiers.get(path);
+			if (lastModifier != null) {
+				this.lastModifiers.remove(lastModifier);
 			}
 		}
 	}
@@ -2954,10 +2977,23 @@ public class JavaModelManager implements ISaveParticipant, IContentTypeChangeLis
 		}
 		ZipCache zipCache;
 		ZipFile zipFile;
-		if ((zipCache = this.zipFiles.get()) != null && (zipFile = zipCache.getCache(path)) != null) {
-			return zipFile;
-		}
 		File localFile = getLocalFile(path);
+		if ((zipCache = this.zipFiles.get()) != null && (zipFile = zipCache.getCache(path)) != null) {
+			if (zipCache.lastModified(path) == localFile.lastModified()) {
+				return zipFile;
+			} else {
+				try {
+					zipFile.close();
+					zipCache.clearCache(path);
+				} catch (IOException e) {
+					int code = -1;
+					if(e instanceof FileNotFoundException || e instanceof NoSuchFileException) {
+						code = IJavaModelStatusConstants.ELEMENT_DOES_NOT_EXIST;
+					}
+					throw new JavaModelException(new JavaModelStatus(code, e));
+				}
+			}
+		}
 		try {
 			if (ZIP_ACCESS_WARNING) {
 				traceZipAccessWarning(path);
@@ -2970,7 +3006,7 @@ public class JavaModelManager implements ISaveParticipant, IContentTypeChangeLis
 			}
 			zipFile = new ZipFile(localFile);
 			if (zipCache != null) {
-				zipCache.setCache(path, zipFile);
+				zipCache.setCache(path, zipFile, localFile.lastModified());
 			}
 			addInvalidArchive(path, ArchiveValidity.VALID); // remember its valid
 			return zipFile;
--- a/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/AddJarFileToIndex.java
+++ b/org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/AddJarFileToIndex.java
@@ -22,6 +22,7 @@ import java.nio.charset.Charset;
 import java.nio.file.NoSuchFileException;
 import java.util.Enumeration;
 import java.util.zip.ZipEntry;
+import java.util.zip.ZipException;
 import java.util.zip.ZipFile;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.runtime.CoreException;
@@ -120,7 +121,6 @@ class AddJarFileToIndex extends BinaryContainer {
 				return true; // index got deleted since acquired
 			}
 			index.separator = JAR_SEPARATOR;
-			@SuppressWarnings("resource")
 			ZipFile zip = null;
 			try {
 				// this path will be a relative path to the workspace in case the zipfile in the workspace otherwise it will be a path in the
@@ -148,15 +148,15 @@ class AddJarFileToIndex extends BinaryContainer {
 					}
 					if (JavaModelManager.ZIP_ACCESS_VERBOSE)
 						trace("(" + Thread.currentThread() + ") [AddJarFileToIndex.execute()] Creating ZipFile on " + this.containerPath); //$NON-NLS-1$	//$NON-NLS-2$
-					zip = new ZipFile(file);
 					zipFilePath = (Path) this.resource.getFullPath().makeRelative();
+					zip = close(zipFilePath, file);
 					// absolute path relative to the workspace
 				} else {
 					if (JavaModelManager.ZIP_ACCESS_VERBOSE)
 						trace("(" + Thread.currentThread() + ") [AddJarFileToIndex.execute()] Creating ZipFile on " + this.containerPath); //$NON-NLS-1$	//$NON-NLS-2$
 					// external file -> it is ok to use toFile()
-					zip = new ZipFile(this.containerPath.toFile());
 					zipFilePath = (Path) this.containerPath;
+					zip = close(this.containerPath, this.containerPath.toFile());
 				}
 
 				if (this.isCancelled) {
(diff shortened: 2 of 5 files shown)


rgrunber:
The zipfile part seems similar to some of the work done in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/26/ .

stephan-herrmann:
We get a new warning: "At least one of the problems in category 'resource' is not analysed ... " which is indeed a bit obscure here, but what it tries to convey: `@SuppressWarnings("resource")` is no longer necessary (but ecj isn't 100% sure about that).

I believe the issue arises with the proposed change, because flow analysis is less sure about `zip` because we no longer get a fresh `ZipFile` on each path.

Before the change the annotation suppresses this warning:
> Resource 'zip' should be managed by try-with-resource

To sum up: removing the annotation should be fine and resolve the warning

snjeza:
> > To sum up: removing the annotation should be fine and resolve the warning

@stephan-herrmann I have removed it.

stephan-herrmann:
> > > To sum up: removing the annotation should be fine and resolve the warning
> 
> @stephan-herrmann I have removed it.

Thanks.

Unfortunately we now have a failure in org.eclipse.jdt.core.tests.model.SearchTests.testConcurrentJob.

This _might_ be an intermittent problem, but since the changes relate to indexing, a failure in a search test could hint at a real problem. Could your changes possibly introduce some kind of race condition?

snjeza:
> Unfortunately we now have a failure in org.eclipse.jdt.core.tests.model.SearchTests.testConcurrentJob.

I can reproduce the issue. Trying to fix it.

snjeza:
> Unfortunately we now have a failure in org.eclipse.jdt.core.tests.model.SearchTests.testConcurrentJob.
This might be an intermittent problem, but since the changes relate to indexing, a failure in a search test could hint at a real problem. Could your changes possibly introduce some kind of race condition?

The test failed because of a bug in JavaModelManager.getZipFile(IPath). If a zip file has been changed, JavaModelManager won't register that  change.
@stephan-herrmann I have updated the PR.

fbricon:
@stephan-herrmann @jjohnstn could we please get another review for this PR? thanks!

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/SourceIndexer.java
```diff
@@ -87,13 +89,21 @@ public class SourceIndexer extends AbstractIndexer implements ITypeRequestor, Su
 	private CompilationUnit compilationUnit;
 	private CompilationUnitDeclaration cud;
 	private org.eclipse.jdt.core.dom.ASTNode dom;
+	private static Map<JavaProject, INameEnvironment> environments = new WeakHashMap<>();
```
iloveeclipse: The static map is not nice, but if this the only way, it must be synchronized.
- Please wrap it via `Collections.synchronizedMap()`
- Please move the init code to the constructor below.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/SourceIndexer.java
```diff
@@ -183,7 +193,11 @@ public void resolveDocument() {
 				this.basicParser.scanner.taskTags = null;
 				this.cud = this.basicParser.parse(this.compilationUnit, new CompilationResult(this.compilationUnit, 0, 0, this.options.maxProblemsPerUnit));
 				// Use a non model name environment to avoid locks, monitors and such.
-				INameEnvironment nameEnvironment = new JavaSearchNameEnvironment(javaProject, JavaModelManager.getJavaModelManager().getWorkingCopies(DefaultWorkingCopyOwner.PRIMARY, true/*add primary WCs*/));
+				INameEnvironment nameEnvironment = environments.get(javaProject);
+				if (nameEnvironment == null) {
+					nameEnvironment = new JavaSearchNameEnvironment(javaProject, JavaModelManager.getJavaModelManager().getWorkingCopies(DefaultWorkingCopyOwner.PRIMARY, true/*add primary WCs*/));
```
iloveeclipse: The "weak" part of the map will not work because the value has strong reference to the key.
Quote javadoc

> care should be taken to ensure that value objects do not strongly refer to their
own keys, either directly or indirectly, since that will prevent the keys
from being discarded.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/SourceIndexer.java
```diff
@@ -87,13 +89,21 @@ public class SourceIndexer extends AbstractIndexer implements ITypeRequestor, Su
...
+	private static Map<JavaProject, INameEnvironment> environments = new WeakHashMap<>();
 	private static final boolean DEBUG = false;
 
 	public SourceIndexer(SearchDocument document) {
 		super(document);
 		this.requestor = new SourceIndexerRequestor(this);
 	}
 
+	public static void cleanup() {
+		for (INameEnvironment environment: environments.values()) {
```
iloveeclipse: This code will throw if cleanup() will run in parallel with the code that updates the map. The for loop should catch exceptions.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/AddJarFileToIndex.java
```diff
@@ -148,15 +147,27 @@ public boolean execute(IProgressMonitor progressMonitor) {
 					}
 					if (JavaModelManager.ZIP_ACCESS_VERBOSE)
 						trace("(" + Thread.currentThread() + ") [AddJarFileToIndex.execute()] Creating ZipFile on " + this.containerPath); //$NON-NLS-1$	//$NON-NLS-2$
-					zip = new ZipFile(file);
 					zipFilePath = (Path) this.resource.getFullPath().makeRelative();
+					try {
```
iloveeclipse: The two new blocks of code are identical, please extract to a method.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java
```diff
@@ -159,7 +161,11 @@ public ZipFile getCache(IPath path) {
 			return this.map.get(path);
 		}
 
-		public void setCache(IPath path, ZipFile zipFile) {
+		public long lastModified(IPath path) {
+			return this.lastModifiers.get(path);
```
iloveeclipse: this will throw NPE if there is no value for given key

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java
```diff
@@ -2954,10 +2961,22 @@ public ZipFile getZipFile(IPath path, boolean checkInvalidArchiveCache) throws C
...
 		ZipCache zipCache;
 		ZipFile zipFile;
+		File localFile = getLocalFile(path);
 		if ((zipCache = this.zipFiles.get()) != null && (zipFile = zipCache.getCache(path)) != null) {
-			return zipFile;
+			if (zipCache.lastModified(path) == localFile.lastModified()) {
+				return zipFile;
+			} else {
+				try {
+					zipFile.close();
```
iloveeclipse, line 2986: I'm not sure this is correct, as per design the owner of the zip file is supposed to close it. Here the close() may close the file even if it is still used by a client. So a simply "touch" on a file, followed by the call here may break some code now.

iloveeclipse (changes requested):
Please before doing any changes, rebase on latest master.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/SourceIndexer.java
snjeza: Fixed.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/SourceIndexer.java
snjeza: Fixed.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/AddJarFileToIndex.java
snjeza: Fixed.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java
snjeza: Fixed.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/SourceIndexer.java
snjeza: Fixed.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java
snjeza: The zip file is reopened at 
 https://github.com/eclipse-jdt/eclipse.jdt.core/blob/7e87ad5a5ed1189e552bebf0d7ab1d81a0ed3b6b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java#L2971
Please see also https://github.com/eclipse-jdt/eclipse.jdt.core/blob/7e87ad5a5ed1189e552bebf0d7ab1d81a0ed3b6b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/SearchTests.java#L545

iloveeclipse:
I've just rebased on master.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaModelManager.java
```diff
@@ -2954,10 +2962,22 @@ public ZipFile getZipFile(IPath path, boolean checkInvalidArchiveCache) throws C
...
+				} catch (IOException e) {
+					int code = -1;
+					if(e instanceof FileNotFoundException || e instanceof NoSuchFileException) {
+						code = IJavaModelStatusConstants.ELEMENT_DOES_NOT_EXIST;
+					}
+					throw new JavaModelException(new JavaModelStatus(code, e));
+				}
+			}
 		}
```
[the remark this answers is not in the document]
snjeza: Fixed.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/matching/ClasspathSourceDirectory.java
```diff
@@ -118,8 +120,12 @@ public NameEnvironmentAnswer findClass(String sourceFileWithoutExtension, String
...
 		if (file != null) {
-			return new NameEnvironmentAnswer(new ResourceCompilationUnit(file,
-					this.module == null ? null : this.module.name()), null /* no access restriction */);
+			ResourceCompilationUnit cu = this.cuCache.get(file);
+			if (cu == null) {
+				cu = new ResourceCompilationUnit(file, this.module == null ? null : this.module.name());
+				this.cuCache.put(file, cu);
+			}
+			return new NameEnvironmentAnswer(cu, null /* no access restriction */);
 		}
```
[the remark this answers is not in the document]
snjeza: Fixed.

Issue #4446, opened 2025-09-28
Annotation processing with Intersection Type method leads to compiler error


kamilkrzywanski:
When processing a method with an intersection-type type parameter, for example:

` public <U extends Number & Runnable> void doSomething(U input) { }
`
a StackOverflowException is thrown. I believe the root cause is that the fix from PR #2702 is not being triggered in getUpperBound(...) when the number of superinterfaces is exactly one — which causes Number (the first bound) to be lost, leaving only the interface (Runnable) considered. This in turn leads to incorrect recursion or infinite processing.
<img width="1046" height="888" alt="Image" src="https://github.com/user-attachments/assets/07a46a60-f1f5-4f6c-b251-9e0a218d2950" />

previous issue: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/565

Reproducer: https://github.com/kamilkrzywanski/ECJIntersectionBug

iloveeclipse:
Thanks. Do you plan to push a PR with the fix ?

kamilkrzywanski:
@iloveeclipse I’m not sure yet. The last fix took me a few months — I’m still not very familiar with the ECJ project. The previous issue was a blocker for me but probably easier to handle since I managed to find a fix. Now it seems to be broken while traversing the tree. If I can find some free time, I’ll give it a try, but I can’t promise anything.

kamilkrzywanski:
@iloveeclipse update - it fails because of we have one superInterface and one superClass 
when i chaned method to 
`public <U extends Comparable<U> & Runnable> void doSomething(U input) { }`
it's work fine

ecj source:

```
ReferenceBinding superRefType = (ReferenceBinding) superType;
if (!superType.isInterface()) {
	typeVariable.setSuperClass(superRefType);
} else {
	typeVariable.setSuperInterfaces(new ReferenceBinding[] {superRefType});
}
```

kamilkrzywanski:
@iloveeclipse I've created a quick-fix for this issue, but it's not even tested and will check it tomorrow. 
I've created a PR, I'd like to check do this change may broke sth.  If it's fine tomorrow I'll write tests

kamilkrzywanski:
@iloveeclipse Test attached

iloveeclipse:
> [@iloveeclipse](https://github.com/iloveeclipse) Test attached

Thanks, really appreciated!
@jarthana : do you have time for a review?

PR #4447, opened 2025-09-28
issue #4446 - fix for including superclasses in intersection type in case of type variable

kamilkrzywanski:
issue #4446 - fix for including superclasses in intersection type in case of type variable

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/apt/model/TypeVariableImpl.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/apt/model/TypeVariableImpl.java
@@ -22,6 +22,7 @@ import javax.lang.model.type.TypeVisitor;
 import org.eclipse.jdt.internal.compiler.apt.dispatch.BaseProcessingEnvImpl;
 import org.eclipse.jdt.internal.compiler.lookup.ReferenceBinding;
 import org.eclipse.jdt.internal.compiler.lookup.TypeBinding;
+import org.eclipse.jdt.internal.compiler.lookup.TypeConstants;
 import org.eclipse.jdt.internal.compiler.lookup.TypeVariableBinding;
 
 /**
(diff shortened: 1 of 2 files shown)


iloveeclipse:
@kamilkrzywanski : please undo last merge, and **rebase** your patches on latest master state instead.
Merges are not allowed in platform.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/apt/model/TypeVariableImpl.java
```diff
@@ -65,8 +67,15 @@ public TypeMirror getUpperBound() {
 			// only one bound that is an interface
 			return this._env.getFactory().newTypeMirror(typeVariableBinding.upperBound());
 		}
-        if (superInterfaces.length > 1) {
-            return new IntersectionTypeImpl(this._env, typeVariableBinding);
+
+        boolean superClassIsObject = TypeBinding.equalsEquals(
```
jarthana, line 71: Since we are only comparing with Object, we can simplify this by doing a comparison of superclass.id with TypeIds.T_JavaLangObject.

org.eclipse.jdt.apt.pluggable.tests/resources/targets/issue4446/A.java
```diff
@@ -0,0 +1,9 @@
+package targets.issue4446;
+
+import targets.issue4446.Annotation4446;
```
jarthana, line 3: This is causing the test to fail. The import can be removed.

org.eclipse.jdt.apt.pluggable.tests/resources/targets/issue4446/A.java
```diff
@@ -0,0 +1,9 @@
...
+import java.io.Serializable;
+
+@Annotation4446
+public class Example<T extends Number & Runnable & Serializable> {
```
jarthana, line 7: This is causing the test to fail. Please rename this to Example.java

iloveeclipse:
@jarthana : could you please re-review https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4450 ?

**This** PR is already closed.

org.eclipse.jdt.apt.pluggable.tests/resources/targets/issue4446/A.java
kamilkrzywanski: Fixed in another PR, this one is closed
https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4450/

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/apt/model/TypeVariableImpl.java
kamilkrzywanski: Fixed in another PR, this one is closed
https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4450/

PR #4452, opened 2025-09-30
[Java 25] Duplicate assignment to final field possible with multiple constructors using this() delegation

### LINKED ISSUE #4449: [Java 25] Duplicate assignment to final field possible with multiple constructors using this() delegation
The following code should NOT compile and fail with "The final field i may already have been assigned":

```java
public class TestDuplicateAssignment {

  final int i;
  
  // if you comment out this constructor it reports correct error message
  public TestDuplicateAssignment(Integer o) {
    this(o.intValue());
  }
  
  public TestDuplicateAssignment() {
    this.i = 1;
    this(2);
    System.out.print(this.i);
  }
  
  public TestDuplicateAssignment(int i) {
    this.i = i; // <-- the failure must be here
  }

}
```

This happens with the Eclipse IDE, but also ECJ version I20250926-1800.

Interestingly, if you comment out the *first* constrcutor it fails due to the duplicate assignment. The fix from https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4416 does not help, it fails in the same way before and after.

It fails consistently with OpenJDKs javac:

```
> Compilation failed; see the compiler output below.
  Hinweis: Wiederholen Sie die Kompilierung mit -Xlint:deprecation, um Details zu erhalten.
  Hinweis: Einige Eingabedateien verwenden oder ├╝berschreiben eine veraltete API.
  C:\...\TestDuplicateAssignment.java: Fehler: Variable i wurde möglicherweise bereits zugewiesen
      this(2);
          ^
  1 Fehler
```

@srikanth-sankaran @stephan-herrmann

stephan-herrmann:
+ introduce (Unconditional)DualFlowInfo to co-maintain two flow infos
  1. field initialization info
  2. ctor prologue infos + field initialization info
Constructed and deconstructed in TypeDeclaration.internalAnalyseCode
Transparent for all other code locations.
+ further fine tuning:
  - if all ctors have a prologue then their combined infos are binding for all initializers
  - otherwise that DualFlowInfo is created
+ additionally complain about final fields initialized before this()
  - actually before analysing its arguments
  - assignments in arguments are already reported during their analysis

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4449

### DIFF
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/SuperAfterStatementsTest.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/SuperAfterStatementsTest.java
@@ -3331,4 +3331,161 @@ public class SuperAfterStatementsTest extends AbstractRegressionTest9 {
 		----------
 		""");
 	}
+	public void testGH4449a() {
+		runNegativeTest(new String[] {
+			"X.java",
+			"""
+			public class X {
+
+			  final int i;
+
+			  public X() {
+			    this.i = 1;
+			    this(2);
+			    System.out.print(this.i);
+			  }
+
+			  public X(int i) {
+			    this.i = i;
+			  }
+
+			}
+			"""
+		},
+		"""
+		----------
+		1. ERROR in X.java (at line 7)
+			this(2);
+			^^^^^^^^
+		The final field i may already have been assigned
+		----------
+		""");
+	}
+	public void testGH4449b() {
+		runNegativeTest(new String[] {
+			"X.java",
+			"""
+			public class X {
+
+			  final int i;
+			  public X(Integer o) { this(o.intValue()); }
+			  public X() {
+			    this.i = 1;
+			    this(2);
+			    System.out.print(this.i);
+			  }
+
+			  public X(int i) {
+			    this.i = i;
+			  }
+
+			}
+			"""
+		},
+		"""
+		----------
+		1. ERROR in X.java (at line 7)
+			this(2);
+			^^^^^^^^
+		The final field i may already have been assigned
+		----------
+		""");
+	}
+	public void testGH4449c() {
+		runNegativeTest(new String[] {
+			"X.java",
+			"""
+			public class X {
+
+			  final int i;
+			  public X(Integer o) { this.i = o.intValue(); }
+			  public X() {
+			    this.i = 1;
+			    this(2);
+			    System.out.print(this.i);
+			  }
+
+			  public X(int i) {
+			    this.i = i;
+			  }
+
+			}
+			"""
+		},
+		"""
+		----------
+		1. ERROR in X.java (at line 7)
+			this(2);
+			^^^^^^^^
+		The final field i may already have been assigned
+		----------
+		""");
+	}
+	public void testGH4449d() {
+		runNegativeTest(new String[] {
+			"X.java",
+			"""
+			class X {
+				final Integer f1 = 1;
+				final Integer f2;
+				final Integer f3;
+				final Integer f4;
+				{
+					// f1 DA
+					// f2 DU
+					// f3 PUA
+					// f4 PUA
+					Integer l1;
+					f2 = 2 * f1;
+					if (cond())
+						l1 = f3;
+					else
+						f3 = 3;
+					// f1 DA
+					// f2 DA
+					// f3 PUA
+					// f4 PUA
+				}
+
+				X() {
+					f3 = 31;
+					f4 = 4;
+					// after prologue:
+					// f1 DU
+					// f2 DU
+					// f3 DA
+					// f4 DA
+					super();
+				}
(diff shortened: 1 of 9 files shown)


stephan-herrmann:
Now for some words what's going on here, since the change is quite large, and still known to be incomplete:

Following [the observation](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4449#issuecomment-3351776115) that our flow analysis infra structure is lacking for the task at hand, I invented a new mechanism of piggy-back dual analysis, that produces two results during one AST traversal.
1. for analysing initializers (field initializers and initializer blocks) we need to start with a merge of all ctor prologue infos. This is needed because factually initializers run *after* any such prologue
2. next we may need a result of analysing all initializers *without* such prologue information as the starting info for analysing contructors that don't (directly or indirectly via `this()` call) leverage any prologue.

Both results are now produced behind the facade of a `DualFlowInfo` which externally behaves like a regular `FlowInfo`. Only, when done we can magically extract the (2) results :)

Creating class `DualFlowInfo` by subclassing `ConditionalFlowInfo` was surprisingly easy. Those already do all the dispatching to two internal flow infos, so only few methods had to be adapted. Many callers, however, explicitly expect to get an `UnconditionalFlowInfo` for further processing. This required the creation of a second new class `UnconditionalDualFlowInfo`, which is already larger then `DualFlowInfo`, and still missing a few overrides that are probably necessary for full operation. Though large, I still believe this class to be appropriate because most overrides could be done quite mechanically.

This already passed all of compiler regression tests. I nevertheless plan to reduce the impact by more smarts of avoiding the creation of a `DualFlowInfo` where not strictly necessary, notably, when both constituents are essentially equal. Perhaps this optimization could then be turned off using a hidden flag, for more stress testing of the new classes :)

stephan-herrmann:
> I nevertheless plan to reduce the impact by more smarts of avoiding the creation of a `DualFlowInfo` where not strictly necessary, notably, when both constituents are essentially equal.

I first determined that out of 13523 compiler regression tests 28 tests created a `DualFlowInfo` with equals constituents. When I saw in one of these tests, that both parts where actually "empty", i.e., no relevant inits recorded, I added a check preventing further processing of prologueInfo if empty, in that block where `DualFlowInfo` are created. Result: all of those 28 tests ceased creating `DualFlowInfo` (all of these had non-trivial arguments in their `constructorCall`, which however don't produce any inits).

> Perhaps this optimization could then be turned off using a hidden flag, for more stress testing of the new classes :)

Given the above I abandoned that idea.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/SuperAfterStatementsTest.java
```diff
@@ -3331,4 +3331,161 @@ public CN(final List<Class<?>> xtypes) {
...
+		"""
+		----------
+		1. ERROR in X.java (at line 7)
+			this(2);
+			^^^^^^^^
+		The final field i may already have been assigned
+		----------
+		""");
+	}
+	public void testGH4449c() {
```
uschindler, line 3394: Isn't this test a duplicate of previous?

uschindler (commented):
Duplicate test?

uschindler:
Thanks for the great fast fix! 👍

srikanth-sankaran:
> Following [the observation](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4449#issuecomment-3351776115) that our flow analysis infra structure is lacking for the task at hand, I invented a new mechanism of piggy-back dual analysis, that produces two results during one AST traversal.
> 
> 1. for analysing initializers (field initializers and initializer blocks) we need to start with a merge of all ctor prologue infos. This is needed because factually initializers run _after_ any such prologue
> 2. next we may need a result of analysing all initializers _without_ such prologue information as the starting info for analysing contructors that don't (directly or indirectly via `this()` call) leverage any prologue.
> 
> Both results are now produced behind the facade of a `DualFlowInfo` which externally behaves like a regular `FlowInfo`. Only, when done we can magically extract the (2) results :)

See https://github.com/eclipse-jdt/eclipse.jdt.core/pull/5382#issuecomment-5662955782 also - ATM we don't know whether #5382 will be the future.

2. above is problematic I feel. Constructors that chain to alternate constructors via `this(...)` are best NOT presented with ANY information from field analysis. Such constructors should not influence field analysis, and field analysis should not influence any part of such constructors - otherwise the model control flow deviates from actual control flow that can result in complications.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/SuperAfterStatementsTest.java
srikanth-sankaran: Nope. Previous test  chains to alternate constructor with `this(...)` while this doesn't.

Issue #4458, opened 2025-10-01
ArrayIndexOutOfBoundsException in PolyTypeBinding


basilevs:
# Test steps 
- Create following class in JDT
- Compile or open it

```java
package oob_reproducer4;

import java.util.function.Function;
import java.lang.Exception;
import java.util.stream.Stream;

public class Reproducer {
	public interface ThrowingRunnable {
		void run() throws Exception;
	}
	
	public static Runnable encode(ThrowingRunnable throwing) {
		return () -> {
			try {
				throwing.run();
			} catch (Exception e) {
				throw new RuntimeException(e);
			} 
		};
	}
	
	public interface ThrowingFunction<T, R> {
		R apply(T t) throws Exception;
	}
	
	public static <T,R> Function<T, R> encode(ThrowingFunction<T,R> throwing) {
		return t -> {
			try {
				return throwing.apply(t);
			} catch (Exception e) {
				throw new RuntimeException(e);
			} 
		};
	}
	
	public void artifactRefs(String... suites)
			throws InterruptedException {
		Stream<Exception> elements = Stream.empty();

		elements 
			.map(encode(  Exception::getMessage)) // a call to overloaded method
			.forEach(context -> {
				();  // Syntax error
		});
	}

}

```

The code looks synthetic because of minimization effort.

# Expected results
Some syntax errors are reported in Problem view

# Actual result
Java compiler and editor fail with exception:

```
null
org.eclipse.core.resources
Error
Wed Oct 01 23:26:15 GMT+04:00 2025
Problems occurred when invoking code from plug-in: "org.eclipse.core.resources".

java.lang.RuntimeException: Internal Error compiling /oob_reproducer/src/org/eclipse/rcptt/cloud/client/Q7ArtifactLoader.java
	at org.eclipse.jdt.internal.compiler.ProcessTaskManager.processing(ProcessTaskManager.java:139)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:545)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:328)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1090)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:614)
	at java.base/java.lang.Thread.run(Thread.java:1474)
[...]

PR #4462, opened 2025-10-06
ArrayIndexOutOfBoundsException in PolyTypeBinding

srikanth-sankaran:
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4458
* Back out the fixes for https://bugs.eclipse.org/bugs/show_bug.cgi?id=500503 & https://bugs.eclipse.org/bugs/show_bug.cgi?id=559677 and refix them comprehensively along side current fix


srikanth-sankaran:
Problem arises from the carte-blanche accorded and alluded to by https://bugs.eclipse.org/bugs/show_bug.cgi?id=559677#c5 - that comment and behavior tries to avoid secondary errors by claiming compatibility even when unknown, but that is troublesome. 

Two SAM types that differ in SAM signature end up in `org.eclipse.jdt.internal.compiler.ast.ReferenceExpression.sIsMoreSpecific` 
which has this code and comment:

```
		TypeBinding[] sParams = sSam.parameters;
		TypeBinding[] tParams = tSam.parameters;
		// Both must have the same number of parameters if we got this far
		for (int i = 0; i < sParams.length; i++) {
			if (TypeBinding.notEquals(sParams[i], tParams[i]))
				return false;
		}
```

The comment `Both must have the same number of parameters if we got this far` is just not true because we allowed incompatible SAM types as compatible

srikanth-sankaran:
This PR also backs out the fix provided by https://bugs.eclipse.org/bugs/show_bug.cgi?id=500503 and refixes it alternately.

A reference expression in AST form can be "serialized" into text form and can be deserialized into AST form by reparsing. This is used in conjunction with the fix for https://bugs.eclipse.org/bugs/show_bug.cgi?id=571467 which stashes away a textual copy of the functional expression to avoid having to use the recovery scanner. Normal scanner can always be used. 

Please note that a `LambdaExpression` AST node may not likewise be serialized and deserialized successfully due to how `org.eclipse.jdt.internal.compiler.ast.LambdaExpression.printExpression(int, StringBuilder, boolean)` may inject extraneous text into the toString output. 

(If needed we can solve the problem so a lambda can be cleanly serialized and deserialized but I am leaning on the observation in comment https://bugs.eclipse.org/bugs/show_bug.cgi?id=500503#c16 to say we don't need it right away)

srikanth-sankaran:
@stephan-herrmann - this undoes and refixes two prior fixes - Could you glance through the changes please ? I think this leaves us overall on firmer grounds - we can go create copies of reference expressions more robustly without having to worry about the identity of the scanner involved and whether there were syntax errors encountered.

stephan-herrmann:
This change as been wrongly pushed to https://github.com/eclipse-jdt/eclipse.jdt.core/tree/aioob-in-polytype here in the main repo. So when the change is merged, we may need help from someone with super powers to remove that branch.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ReferenceExpression.java
```diff
@@ -122,10 +114,7 @@ public void initialize(CompilationResult result, Expression expression, TypeRefe
 	 */
 	private ReferenceExpression copy() {
 		final Parser parser = new Parser(this.enclosingScope.problemReporter(), false);
-		char [] source = new char [this.sourceEnd+1];
-		System.arraycopy(this.text, 0, source, this.sourceStart, this.sourceEnd - this.sourceStart + 1);
-		parser.scanner = this.scanner;
-		ReferenceExpression copy =  (ReferenceExpression) parser.parseReferenceExpression(source, this.sourceStart, this.sourceEnd - this.sourceStart + 1,
+		ReferenceExpression copy =  (ReferenceExpression) parser.parseReferenceExpression(this.text, 0, this.text.length,
```
stephan-herrmann, line 117: Will this create AST claiming to start at source position 0?

Have we established that we never use a `copy` for error reporting? Otherwise this will be a problem, right?

Similar for code select?

srikanth-sankaran:
> This change as been wrongly pushed to https://github.com/eclipse-jdt/eclipse.jdt.core/tree/aioob-in-polytype here in the main repo. So when the change is merged, we may need help from someone with super powers to remove that branch.

Grrrr. My bad - sorry about that. I recently rebooted and started over on my set up and have a glitchy configuration. I think there could be a 2-3 more such branches, I will enumerate them in https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4472 so @iloveeclipse can help with the removal.

Thanks for catching it!

srikanth-sankaran:
> _If_ we decide that rescanning the original text is fubar, then yes, this approach looks promising, but this creates an obligation that each and every reference expression renders correctly to string.

I think this is a workable approach - but I am punting on this question for now by opting for an alternate fix (see PR https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4541 - which makes `org.eclipse.jdt.internal.compiler.ast.ReferenceExpression.contextHasSyntaxError()` operate on a fine grained fashion.)

Closing this PR and shifting action to https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4541

Issue #4412, opened 2025-09-19
Internal Compile error problem since 2025-09


ChrisLewold:
I upgraded two different eclipse installations from 2025-06 to 2025-09. Suddenly I get an internal compile error. 
I was able to create a very small sample class (see below for both java code and the stacktrace as comment). 
Project is simple to set up - all you need is a dependency to jackson-annotations.

1. The file cannot be compiled
2. The file cannot be opened using eclipse java editor. 

[EclipseInternalCompileErrorSample.java](https://github.com/user-attachments/files/22427207/EclipseInternalCompileErrorSample.java)

[pom.xml](https://github.com/user-attachments/files/22427216/pom.xml)

Source file attached.

srikanth-sankaran:
I'll take a look redirecting as needed.

ChrisLewold:
Just adding a screenshot ...

<img width="1573" height="942" alt="Image" src="https://github.com/user-attachments/assets/9451cc66-0cab-4a85-9e44-2c38bf40bce8" />

srikanth-sankaran:
Shorter test case without any dependencies:

```
import java.util.Collection;
import java.util.stream.Collectors;

@interface A {
    
    Id use();
    
    public enum Id {
        CLASS;
    }    
}

public record R<V>(String name, @A(use = A.Id.CLASS) V value) {
    
    @Override
    public String toString() {
        return String.format("[%s] %s",
                             name,
                             (value instanceof Collection<?> coll
                                     ? coll.stream().map(Object::toString).collect(Collectors.joining(", ", "[ ", " ]"))
                                     : value.toString()));
    }
}

```

srikanth-sankaran:
4.36 compiles it fine, 4.37 fails.

PR #3868, opened 2025-03-30
"Type mismatch" regression since 4.29 with generics and null-analysis

stephan-herrmann:
+ when looking for @NNBD only do a selective search for annotations
  (the current bug happens because this search triggers resolving
   of unrelated annotations with member value pairs that aren't ready)
+ implement specific resolving similar to resolveDeprecatedAnnotations()

Includes two preparatory commits to restructure the code ahead of the actual fix.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3461

PR #1447, opened 2023-09-28
Field "cannot be resolved or is not a field" when referenced from parent class within annotation

stephan-herrmann:
fixes #1431

The provided example code demonstrates once again that our phases within `LookupEnvironment.completeTypeBindings()` were too coarse grained, making it impossible to find an order of steps that works in all possible scenarios.

## What it does

_Semantically_, I simply split the step `BUILD_FIELDS_AND_METHODS` to extract `CHECK_PARAMETERIZED_TYPES` into its own phase. Now what previously I called `CONNECT_TYPE_HIERARCHY2` can be squeeze between those phases, to fix the root cause of this current bug: annotation members where resolved before another class's fields where built. Hence the annotation couldn't resolve its value. With `CONNECT_TYPE_HIERARCHY2` running after `BUILD_FIELDS_AND_METHODS` this problem was resolved, but for bound checks of null annotated parameterized types, `CHECK_PARAMETERIZED_TYPES` had to run still later, in order to "see" the annotations resolved during `CONNECT_TYPE_HIERARCHY2`.

_Technically_ I made a proposal how to modernize just this slice of code, see https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1376#issuecomment-1720253488 -- I moved all code pertaining to sequencing the individual steps of `LE.completeTypeBindings()` into a new `enum CompleteTypeBindingSteps`, thus eliminating some boiler plate and redundant code.

While I was at it, I renamed `CONNECT_TYPE_HIERARCHY2` to `INTEGRATE_ANNOTATIONS_IN_HIERARCHY` (the method accordingly) and renamed `CONNECT_TYPE_HIERARCHY1`  back to `CONNECT_TYPE_HIERARCHY` :)

## How to test

JUnit included.

PR #4468, opened 2025-10-07
Internal Compile error problem since 2025-09

srikanth-sankaran:
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4412

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/SourceTypeBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/SourceTypeBinding.java
@@ -937,6 +937,7 @@ private VariableBinding resolveTypeFor(VariableBinding variable) {
 		if (variable instanceof FieldBinding field)
 			initializationScope.initializedField = field;
 		AbstractVariableDeclaration variableDeclaration = variable instanceof FieldBinding field ? field.sourceField() : ((RecordComponentBinding) variable).sourceRecordComponent();
+		ASTNode.resolveNullDefaultAnnotations(initializationScope, variableDeclaration.annotations, variable);
 		TypeBinding variableType =
 			variableDeclaration.getKind() == AbstractVariableDeclaration.ENUM_CONSTANT
 				? initializationScope.environment().convertToRawType(this, false /*do not force conversion of enclosing types*/) // enum constant is implicitly of declaring enum type


srikanth-sankaran:
This PR fixes the problem under consideration and if would pass muster would be a welcome change that avoids reentrancy that is painful to deal with. I don't see why the deprecation bits should be processed there - which results in ClassScope.buildFields() being entered even before record components are collated fully. 

I have seen such reentrancy due to annotation resolution and this would be a good problem to solve

stephan-herrmann:
> I have seen such reentrancy due to annotation resolution and this would be a good problem to solve

Right, there have been many instances of problems in this category and much has been done against it already, e.g.:
* Introduce more fine-grained phases via `CompleteTypeBindingsSteps`
  *  #1077
   * #1447
* Look for certain annotations only, viz.  `ASTNode.resolveNullDefaultAnnotations()` in analogy to `ASTNode.resolveDeprecatedAnnotations()`
   *  #3868

And those are only the more fundamental fixes alongside a bunch of point fixes.

I'm looking into the current instance now.

stephan-herrmann:
One of the regressions is in NullTypeAnnotationTest.testBug484926().

Here resolving the field `f` should produce the unannotated type `AtomicReference<String>`, although the class is annotated `@NonNullByDefault`, which would make it `AtomicReference<@NonNull String>`. To avoid this mistake, STB.resolveTypeFor() must see the effect of `@NonNullByDefault({})` right on the field (which cancels the outer default).

If we can't block the reentrance right where it happens (could we?), then this puts the current bug into the bucket of issues where annotation resolving should be made more precise: resolve only what's needed right here.

The compiler is directly interested in only a few annotation types:
* `@Deprecated`
* null annotations
* ownership annotations

Can these annotations produce the re-entrance observed in this bug? Technically, yes:
```java
public class GH4468 {
	public record R<V>(String name, @Deprecated(forRemoval=remove) V value) {
	
		static final boolean remove = false;
	    @Override
	    public String toString() {
	    	return value.toString();
	    }
	}
}
```

Does such style make any sense for the relevant annotations? No.

stephan-herrmann:
Are we comfortable with the assumption that `@NonNullByDefault` is not parameterized via a local constant, but only with direct references to constants from `DefaultLocation`? If so then the following line will fix the regressions:
```java
		ASTNode.resolveNullDefaultAnnotations(initializationScope, variableDeclaration.annotations, variable);
```
It would have to be placed before `variableDeclaration.type.resolveType(..)`, to compensate the proposed change wrt null annotations.

I'll still check the other side of the equation: could we easily reduce the bad effect of `getAnnotationTagBits()` in its former location?

stephan-herrmann:
> ownership annotations

`@Owning` and `@NotOwning` are not involved in this issue, as they don't carry any attributes that could trigger unwanted resolving.

stephan-herrmann:
Here's a fix that's opportunistic in a different way:
```java
	private boolean lookForField(ReferenceBinding currentType, char[] fieldName) {
		if ((currentType.tagBits & TagBits.AreFieldsComplete) != 0)
			return true;
		if (currentType.original() instanceof SourceTypeBinding stb && stb.scope != null) {
			FieldDeclaration[] fields = stb.scope.referenceContext.fields;
			if (fields != null) {
				for (FieldDeclaration fieldDeclaration : fields)
					if (CharOperation.equals(fieldName, fieldDeclaration.name))
						return true;
			}
			RecordComponent[] comps = stb.scope.referenceContext.recordComponents;
			if (comps != null) {
				for (RecordComponent compDeclaration : comps)
					if (CharOperation.equals(fieldName, compDeclaration.name))
						return true;
			}
			return false;
		}
		return true;
	}
```

Use this method in `Scope.findField()` to skip processing the currentType if (a) this type hasn't yet completed fields and (b) doesn't have a field declaration by the sought name :)

```diff
diff --git a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
index 3880c39..d18919a 100644
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
@@ -1395,17 +1395,20 @@
 		if (!currentType.canBeSeenBy(this))
 			return new ProblemFieldBinding(currentType, fieldName, ProblemReasons.ReceiverTypeNotVisible);
 
-		currentType.initializeForStaticImports();
-		FieldBinding field = currentType.getField(fieldName, needResolve);
-		if (field != null) {
-			if (invisibleFieldsOk) {
-				return field;
-			}
-			if (invocationSite == null
-				? field.canBeSeenBy(getCurrentPackage())
-				: field.canBeSeenBy(currentType, invocationSite, this))
+		FieldBinding field = null;
+		if (lookForField(currentType, fieldName)) {
+			currentType.initializeForStaticImports();
+			field = currentType.getField(fieldName, needResolve);
+			if (field != null) {
+				if (invisibleFieldsOk) {
 					return field;
-			return new ProblemFieldBinding(field /* closest match*/, field.declaringClass, fieldName, ProblemReasons.NotVisible);
+				}
+				if (invocationSite == null
+					? field.canBeSeenBy(getCurrentPackage())
+					: field.canBeSeenBy(currentType, invocationSite, this))
+						return field;
+				return new ProblemFieldBinding(field /* closest match*/, field.declaringClass, fieldName, ProblemReasons.NotVisible);
+			}
 		}
 		// collect all superinterfaces of receiverType until the field is found in a supertype
 		ReferenceBinding[] interfacesToVisit = null;
```

This approach, however, is a bit wasteful, I'd say:
* running compiler regression tests at 25 the linear search within FieldDeclarations is triggered 7715 times (in 13524 tests), so it's not a rare situation
* it's also "intellectually wasteful": it introduces a slightly odd shortcut, and we can't be quite sure of the completeness of this fix

Hence my vote goes for https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4468#issuecomment-3377541466

srikanth-sankaran:
> Are we comfortable with the assumption that `@NonNullByDefault` is not parameterized via a local constant, but only with direct references to constants from `DefaultLocation`?

Yes, I think so. I think we don't have to worry about such extremely corner case. This reentrancy issue keeps biting us over and over - I remember battling it a dozen years ago.

> If so then the following line will fix the regressions:
> 
> ```java
> 		ASTNode.resolveNullDefaultAnnotations(initializationScope, variableDeclaration.annotations, variable);
> ```
> 
> It would have to be placed before `variableDeclaration.type.resolveType(..)`, to compensate the proposed change wrt null annotations.

Thanks, I will give this a shot.

> 
> I'll still check the other side of the equation: could we easily reduce the bad effect of `getAnnotationTagBits()` in its former location?

For me, just seeing the modifier bits that are meant to be set by it and seeing those modifiers having no reason to be set _that eagerly_ settles the argument

srikanth-sankaran:
> This approach, however, is a bit wasteful, I'd say:
> 
> * running compiler regression tests at 25 the linear search within FieldDeclarations is triggered 7715 times (in 13524 tests), so it's not a rare situation
> * it's also "intellectually wasteful": it introduces a slightly odd shortcut, and we can't be quite sure of the completeness of this fix
> 
> Hence my vote goes for [#4468 (comment)](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4468#issuecomment-3377541466)

Yeah. Mine too.

mpalat:
Taking a look at the failure on @srikanth-sankaran's request, these failures look disconnected to the fix and more of a network/timeout issues. I think you can go ahead and merge. thanks.

PR #4501, opened 2025-10-13
Activate disabled tests

srikanth-sankaran:
Activate `org.eclipse.jdt.core.tests.dom.ASTConverter18Test.testBug425601_001()` and `org.eclipse.jdt.core.tests.dom.ASTConverter18Test.testBug425601_002` fixing problems in project set up for 1.8 compliance

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverter17Test.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverter17Test.java
@@ -772,7 +772,7 @@ public class ASTConverter17Test extends ConverterTestSetup {
 	/*
 	 * https://bugs.eclipse.org/bugs/show_bug.cgi?id=399791
 	 */
-	public void _2551_test0021() throws JavaModelException {
+	public void test0021() throws JavaModelException {
 		String contents =
 				"public interface X {\n" +
 				"	static void foo(){}\n" +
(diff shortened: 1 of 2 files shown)


srikanth-sankaran:
As `org.eclipse.jdt.core.tests.dom.ConverterTestSetup.setUpJavaProject(String, String)` disables raw type warnings for 1.8 compliance, we don't need the SuppressWarning annotation in `org.eclipse.jdt.core.tests.dom.ASTConverter18Test.testBug425601_001()` and `org.eclipse.jdt.core.tests.dom.ASTConverter18Test.testBug425601_002()`

Having the suppress warnings annotation in place would result in the following diagnostic (correct) but unexpected by the test

```
1. INFO in /Converter18/src/testBug425601_001/Outer.java (at line 14)\r\n
	@SuppressWarnings("rawtypes")    Outer.Middle m3; \r\n
	                  ^^^^^^^^^^\n
At least one of the problems in category 'rawtypes' is not analysed due to a compiler option being ignored\n

```

( `org.eclipse.jdt.core.tests.dom.ConverterTestSetup.setUpJavaProject(String, String)`  should really disable raw type warnings for tests that use to run in 1.4- compliance mode earlier but are bumped up to 1.8 mode but it does what it does and the present PR leaves it alone and fixes ripples arising out of it)

srikanth-sankaran:
org.eclipse.jdt.core.tests.dom.ASTConverter17Test.test0021() and org.eclipse.jdt.core.tests.dom.ASTConverter17Test.test0022() are straightforward - They are expecting that AST would be flagged malformed if the code contains interface static methods, default methods and type annotations - simply not true in minimum 1.8 mode.

PR #4502, opened 2025-10-13
Active a couple of more disabled tests

srikanth-sankaran:
Adjust expectation that at 1.8, interface static methods, default methods and type annotations present in source would result in AST being flagged MALFORMED and reenabled disabled tests


srikanth-sankaran:
Subsumed into https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4501

PR #4509, opened 2025-10-14
Investigate tests disabled during transition to minimal 1.8 target

stephan-herrmann:
+ improve protocol for initialization null annotations & their package
+ re-enable NullAnnotationBatchCompilerTest.test315_warn_options()

See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2758

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ExtendedTagBits.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ExtendedTagBits.java
@@ -39,4 +39,6 @@ public interface ExtendedTagBits {
 	static boolean areAllAnnotationsResolved(long extendedTagBits) {
 		return (extendedTagBits & AllAnnotationsResolved) == AllAnnotationsResolved;
 	}
+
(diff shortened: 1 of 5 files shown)


stephan-herrmann:
The protocol for initialization of null annotations and their package had gone wild:
* fields like `LE.nullableAnnotationPackage` had served a dual purpose
   * remember a detected PackageBinding to answer `LE.isNullnessAnnotationPackage()`
   * reset to null in `PackageBinding.checkIfAnalysisAnnotationType()` to signal that the matching type has been found and no further checking is needed
   * this makes the protocol vulnerable to bad ordering of events
* when switching from 1.5 to 1.8, `LE.initializeUsesNullTypeAnnotation()` triggered resolving of null annotations at an unhappy point in time
   * resolving triggered `PackageBinding.checkIfAnalysisAnnotationType()` which reset package bindings to null
   * later `SourceTypeBinding.evaluateNullAnnotations()` was no longer able to detect that we _are_ in the null annotation package.

Fixed by replacing those contentious fields in LE with setting `ExtendedTagBits.IsNullAnnotationPackage` in the PackageBinding.  At that point I realized that the generalizations from #1716 made `PackageBinding.checkIfAnalysisAnnotationType()` essentially unnecessary. We no longer need to worry about unnecessarily performing this check, since it was essentially reduced to just one map access.

PR #4510, opened 2025-10-14
Fix and enable the tests in JavaSearchBugsTests #2758

jarthana:
## What it does
Fixes some of the failing tests mentioned in #2758

## Author checklist


jarthana:
This was somewhat a tough nut to crack. The test `JavaSearchBugsTests.testBug573486_showReferences_inMethodsAndFields_whenNoSource()` was only passing because of a quirk in the search/compiler. Two of the elements in the testcase were referencing the type java.util.List that wasn't present in the jclMin1.5.jar. Because they were missing from the jar, they somehow made it to the search result as `POTENTIAL_MATCH`. But I still don't know how they don't make it when run with jclMin1.8.jar that includes the `java.util.List`. Anyway, the PR alters the testcase to reference `java.util.ArrayList` (which is absent in the jclMin1.8). Status quo restored.

Just for the records, the code that returns an INACCURATE_MATCH without investigating further is here: `org.eclipse.jdt.internal.core.search.matching.PatternLocator.resolveLevelForType(char[], TypeBinding)`

PR #4517, opened 2025-10-14
Re-enable org.eclipse.jdt.core.tests.dom.ASTConverter17Test.test0018()

stephan-herrmann:
Repair BindingResolver.isResolvedTypeInferredFromExpectedType() in 1.8+

Remove never-assigned PGMB.inferredReturnType, use instead wasInferred
+ minor adjustment in CastExpression.isIndirectlyUsed()

see https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2758

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverter15JLS4Test.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverter15JLS4Test.java
@@ -8236,7 +8236,7 @@ public class ASTConverter15JLS4Test extends ConverterTestSetup {
 		expression = fragment.getInitializer();
 		assertEquals("Not a super method invocation", ASTNode.SUPER_METHOD_INVOCATION, expression.getNodeType());
 		methodInvocation = (SuperMethodInvocation) expression;
-		assertFalse("Wrong value", methodInvocation.isResolvedTypeInferredFromExpectedType());
+		assertTrue("Wrong value", methodInvocation.isResolvedTypeInferredFromExpectedType());
 	}
 
 	//https://bugs.eclipse.org/bugs/show_bug.cgi?id=174436
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverter15JLS8Test.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverter15JLS8Test.java
@@ -8233,7 +8233,7 @@ public class ASTConverter15JLS8Test extends ConverterTestSetup {
 		expression = fragment.getInitializer();
 		assertEquals("Not a super method invocation", ASTNode.SUPER_METHOD_INVOCATION, expression.getNodeType());
 		methodInvocation = (SuperMethodInvocation) expression;
-		assertFalse("Wrong value", methodInvocation.isResolvedTypeInferredFromExpectedType());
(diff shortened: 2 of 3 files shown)


stephan-herrmann:
As mentioned in https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2758#issuecomment-3402550549 test `ASTConverter17Test.test0018()` apparently was the only test needing the flag `PGMB.inferredReturnType`, which, however, was never assigned at compliance 1.8+.

This flag surfaced in public API  `org.eclipse.jdt.core.dom.ClassInstanceCreation.isResolvedTypeInferredFromExpectedType()`  based on implementation in `DefaultBindingResolver.isResolvedTypeInferredFromExpectedType()`. JDT/UI uses this API in one quick fix and one refactoring. Without in-depth investigation this looks like certain modifications are interested in knowing if a target type was needed to correctly infer an invocation, so that in absence of a target type explicit type arguments will be generated.

Now there can be several levels of involvement of type inference, and the given `test0018()` had explicit expectations in which cases the flag should be set, looking at these examples:
```java
		X<Object> a = new X<Object>("hello");
		X<String> b = new X<>("hello");
		X<Object> c = new X<>(null);
		X<String> d = new X<>(null);
```
the test expected the flag to be set only for cases `c` and `d`. While `a` and `d` are obvious, I can only guess about the other cases (now that the old implementation is gone):
* in case `b` the argument `"hello"` suffices for successful inference, so this works as a standalone expression.
* in case `c` the default result of `j.l.Object` suffices, yet this counts as "inferred"

To re-implement the lost strategy I could think of two approaches:
1. between the two phases of inference (applicability inference vs. invocation type inference) we could use the intermediate BoundSet and instantiate a parameterized method with this result, then compare the return type of this intermediate instantiation with the return type of the final inference result. If types differ then the target type provably made a difference for the outcome
2. we could also just compare the return type of the final inferred method with the return type of its `original()`. If those differ, then inference had some impact.

Approach (1) looks a bit heavy-weight and penalizes _every_ invocation of type inference.

For approach (2) I actually found a precedent in `DefaultBindingResolver.isResolvedTypeInferredFromExpectedType(MethodInvocation)`.  This approach looks simpler and kicks in only when needed. It more eagerly flags "inferred" than the old implementation but I guess this change is safe, it could only cause refactoring to generate explicit type arguments in a few cases where those are not needed, which is better than generating code that doesn't compiler.

After deciding for (2) I could actually remove the original flag `PGMB.inferredReturnType`, relying on the related flag `PGMB.wasInferred` plus comparing original and inferred types when needed.

I checked that JDT/UI's `QuickFixTestSuite` finds no problem, which isn't much of a guarantee given that those tests never barked about the missing functionality at compliance 1.8+.

stephan-herrmann:
I did some debugging of how the Inline local variable refactoring deals with cases a-d above. This led me to https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1573 where the exact problem of our flag no longer being reliable at 1.8+ was actually diagnosed, but then an elaborate solution was implemented in JDT/UI - without reliance on the flag from JDT/Core.

stephan-herrmann:
Finally my search led me back home to https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2776

A little more than one year ago, and already I couldn't remember having been at that very place before. :flushed: 

So, the precedent mentioned above was created exactly in that #2776, but we simply forgot to cover also allocation expressions. Only by completing this task, `PGMB.inferredReturnType` can finally be retired now.

OTOH, in light of https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1573 it's OK to do a simple classification in JDT/Core. JDT/UI no longer relies on it being accurate. If there are any external clients, they will at least get some help from the reinstated functionality.

Issue #4497, opened 2025-10-13
The Markdown parser parses the comment incorrectly


subyssurendran666:
Please consider the below  snippet

```
/// {@inheritDoc}
/// In addition, this methods calls [#wait()].
///
/// @param i the index
public class Spec03Tags extends Super {}
```
After the `{@inheritDoc}` TagElement, the parser still treating the `}` of the TagElement as a TextElement.

This problem causes regression in the [JDT.UI](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2533).

iloveeclipse:
I I understand it right, this is regression from https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4203
@subyssurendran666 : do you plan to work on this?
If not, should https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4203 be reverted?

It is not OK to keep tests failed for a long time.
See https://en.wikipedia.org/wiki/Broken_windows_theory

subyssurendran666:
I’m currently working on this issue and will try to fix it ASAP.

subyssurendran666:
This issue was inadvertently fixed along with #4626 . Therefore, I’m closing this issue.

PR #4519, opened 2025-10-15
The Markdown parser parses the comment incorrectly

subyssurendran666:
Fixed the issues in AbstractCommentParser related to the inheritDoc inconsistency

Fix: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4497

## How to test
**Scenario 1**
```
/// {@inheritDoc}
/// In addition, this methods calls [#wait()].
///
/// @param i the index
```

**Scenario 2**
```
/// {@inheritDoc}}
/// In addition, this methods calls [#wait()].
/// @param i the index
```

**Scenario 3**
```
/// {@inheritDoc}
/// {@link
///SomeClass}
```

**Scenario 4**
```
/// {@inheritDoc} }
/// {@inheritDoc}}
/// Duplicate inheritDoc
```

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/AbstractCommentParser.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/AbstractCommentParser.java
@@ -405,8 +405,12 @@ public abstract class AbstractCommentParser implements JavadocTagConstants {
 								// Reset textStart only if there's content after the tag
 								if (this.index < this.javadocEnd) {
 									char next = this.source[this.index];
-									if (!(this.markdown && (next == '\r' || next == '\n'))) {
+									if (!this.markdown || !(next == '\r' || next == '\n') || (previousChar == '@' && nextCharacter == '}')) {
 										this.textStart = this.index;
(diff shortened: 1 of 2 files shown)


jarthana:
@subyssurendran666 while this patch does fix the problem raised, I think we should take a good look at the implementation of markdown from beginning, esp. for the inline tags. May be leverage IMarkdownCommentHelper more. As for this patch, this doesn't work for cases like below:

	/// {@literal abc
	///}
	/// In addition, this methods calls [#wait()].

And we have more issues like https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4531 to deal with too.
I suggest we revert the fix that caused https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2533 and do a revamp when we have more time. WDYT?

subyssurendran666:
> I suggest we revert the fix that caused https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2533 and do a revamp when we have more time. WDYT?

Okay Jay, I will revamp it way you suggested.

subyssurendran666:
Closing this PR in lights of https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4519#issuecomment-3430723318

Issue #4503, opened 2025-10-13
[Mockito] Compilation error "Unhandled exception type Throwable"


yuriypalych:
Hi all,

Our project contains numerous **JUnit** tests based on the **Mockito** framework. The project has maintained successful **maven/javac** compilation for years. However, **Eclipse** compiler generates "Unhandled exception type Throwable" compilation errors in many places.

My synthetic test to reproduce the issue:
````
import org.mockito.ArgumentMatchers;
import org.mockito.Mock;
import org.mockito.Mockito;
import org.springframework.retry.support.RetryTemplate;
import org.springframework.web.client.RestClientException;

public class MyTest {

    @Mock()
    private RetryTemplate mockRetryTemplate;

    public void test() {
        Mockito.when(mockRetryTemplate.execute(ArgumentMatchers.any()))    // "Unhandled exception type Throwable" is here
            .thenThrow(new RestClientException("Mocked exception"));
    }
}
````
Do you have any idea why this code cannot be compiled in Eclipse, but javac compiles it?
BR, Yuriy.

srikanth-sankaran:
Can you please attach a project with dependencies and such ? TIA

yuriypalych:
Please find attached Eclipse project with 1 java class only to reproduce the issue: [CompileTest.zip](https://github.com/user-attachments/files/22901578/CompileTest.zip)
I use Win11 and Eclipse 2025-06 M1 (4.36.0 M1), Build id: 20250410-0644
Thank you in advance.

srikanth-sankaran:
My development environment is plain eclipse SDK - If I import projects from that zip file, I don't get a reproducer. What are the steps I am supposed to follow ? (I don't use maven etc)

yuriypalych:
Hi Srikanth
Please find attached [CompileTest-v2.zip](https://github.com/user-attachments/files/22902619/CompileTest-v2.zip) project without maven and pom.xml, but with two required JARs (mockito-core-5.11.0.jar and spring-retry-2.0.11.jar). Hope it will help to reproduce the issue on plain Eclipse SDK.
Thank you.

srikanth-sankaran:
> Hope it will help to reproduce the issue on plain Eclipse SDK.

Thanks! Yes, reproduced. Will check what is going on there

srikanth-sankaran:
Could you also point me to the source zips for the dependencies jars ? That will help me extract a case outside of Mockito and spring. TIA

[...]

PR #4520, opened 2025-10-15
Minimal test case for Stephan for https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4503

srikanth-sankaran:
Minimal test case for @stephan-herrmann 

One test shows differing behavior wrt javac while another slightly modified (eliminating nested inference) case shows matching behavior. Perhaps that will offer a clue to what is going on ?


stephan-herrmann:
> One test shows differing behavior wrt javac while another slightly modified (eliminating nested inference) case shows matching behavior. Perhaps that will offer a clue to what is going on ?

Indeed it does! Thanks a lot.

PR #4524, opened 2025-10-16
Reenabled disabled test with a suitable fix

srikanth-sankaran:
Reenabled `org.eclipse.jdt.core.tests.model.CompletionTests.testBug402574()`


srikanth-sankaran:
The infra is flaky - I am integrating this as it is just a test only change and all tests have passed.

Issue #2575, opened 2024-06-16
"Call hierarchy" does not return all calling methods


ricetrac:
Hi, 

My environment on Windows 10:

eclipse.buildId=4.32.0.20240606-1231
java.version=21.0.2
java.vendor=Eclipse Adoptium
BootLoader constants: OS=win32, ARCH=x86_64, WS=win32, NL=fr_FR
Framework arguments:  -data C:/eclipse/workspace -product org.eclipse.epp.package.jee.product
Command-line arguments:  -os win32 -ws win32 -arch x86_64 -data C:/eclipse/workspace -product org.eclipse.epp.package.jee.product

The "call hierarchy" function on a method does not always return all results, since Eclipse version 4.31. I waited for version 4.32 to see if the bug was fixed, but it wasn't.
To get all the results, you have to go through the “References -> Project” function.
The result that is missing via "Call Hierarchy" is indicated as "potential match" by the "References" function, but it is not a potential match, it is a real call to the method.

I reproduced the problem:
- A "methodeappel" method which expects a String as a parameter
- 2 methods which call it by constructing the String via another method "CFileUtils.InputStreamtoString()":
      - the first passes the InputStream of the "request" => it does not appear in the "call hierarchy"
      - the second by passing a FileInputStream => it appears in the "call hierarchy"
      - both appear in "References", but the first with "potential match"

Attached a screenshot showing the result:
![bug eclipse callhierarchy](https://github.com/eclipse-jdt/eclipse.jdt.core/assets/1933261/7b4112d4-3a05-4f58-a4eb-80bb89733bb1)

Thanks

Julien

iloveeclipse:
Could you please create a small self containing project that shows the problem and attach it here? Or just all required source files?

fedejeanne:
I also experience the same problem, maybe even the exact same root cause. See https://github.com/vi-eclipse/Eclipse-JDT/issues/9

I created a PR that solves **my** issue --> https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4341

@ricetrac could you please test that PR and see if it solves your issue too?

PR #4341, opened 2025-08-21
Fix missing matches in the call hierarchy

fedejeanne:
## What it does
Integrate the annotations in the hierarchy using their own set of navigated types. The cycle-detection algorithm assumed that the mere presence of a type in the set of visited types was enough to detect a cycle in the hierarchy of classes/interfaces, but integrating the annotations can navigate down the hierarchy (i.e. to subclasses) leading to false positives in the algorithm.

Fixes https://github.com/vi-eclipse/Eclipse-JDT/issues/9 (internal issue)
Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2575 (does it, @ricetrac ?)

## How to test
I wasn't able to create a self-contained reproducer, sorry for that. 

There are some details about (part of) our complex set of classes involved in [this comment](https://github.com/vi-eclipse/Eclipse-JDT/issues/9#issuecomment-3204675292), but I must be missing some piece of the puzzle because I tried exactly that and I can't reproduce the error. 

I can still reproduce it in my real WS though. I can provide details and also share the analysis that I made in this PR if necessary.

## Author checklist

- [x] I have thoroughly tested my changes **in my own Workspace**
- [x] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [x] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

PR #4526, opened 2025-10-16
Fix missing matches in the call hierarchy

stephan-herrmann:
Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2575

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchBugsTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchBugsTests.java
@@ -15161,9 +15161,10 @@ public void testBug521240_001() throws CoreException {
 }
 
 /*
- * See https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4341
+ * See problem report https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2575
+ * See also analysis in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4341
  */
-public void testBugGH4341() throws CoreException, IOException {
+public void testGH2575() throws CoreException, IOException {
 	String testProjectName = "BugGH4341";
 	try {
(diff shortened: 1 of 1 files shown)


org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchBugsTests.java
```diff
@@ -15159,6 +15159,87 @@ public void testBug521240_001() throws CoreException {
 			"src/pack1/X.java void pack1.X.foo(Y) [foo] EXACT_MATCH"
 	);
 }
+
+/*
+ * See https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4341
+ */
+public void testBugGH4341() throws CoreException, IOException {
+	String testProjectName = "BugGH4341";
```
fedejeanne, line 15168: NIT: I named the test based on my PR (I wasn't sure it actually fixed #2575 ), but since you uncovered the bigger issue, it safe to rename the test method and point to the issue ;-)

```suggestion
/*
 * See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2575
 */
public void testBugGH2575() throws CoreException, IOException {
	String testProjectName = "BugGH2575";
```

fedejeanne (commented):
Thank you for the analysis and the fix!

I have just one nit-picky comment.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaSearchBugsTests.java
stephan-herrmann: thanks for paying attention to detail :)

I actually think both references are useful since #4341 has a lot of analysis which shouldn't go to waste :)

OTOH, my first navigation typically goes to blame annotations, and then use the link in the commit comment.

fedejeanne: Makes sense, #4341 is also valuable 👍

Issue #4399, opened 2025-09-15
Multirelease support: imports in editor cannot be resolved


tomaswolf:
Thanks for providing multirelease support in JDT. (m2e should also get updates...) The new feature works for compiling, but in an open Java editor, references to imports or classes from `java.lang` cannot be resolved.

Eclipse version:

Version: 2025-09 (4.37.0)
Build id: 20250905-1456

org.eclipse.jdt.core 3.43.0.v20250820-1501
org.eclipse.jdt.ui 3.35.100.v20250905-1111
org.eclipse.m2e.jdt 2.4.101.20250727-0653

To reproduce:

1. Import the "mrproject" maven project from [mrproject.zip](https://github.com/user-attachments/files/22347267/mrproject.zip) into the workspace.
2. Reset the project's default compliance to 1.8 (Package explorer context menu&rarr;Properties&rarr;Java Compiler; uncheck "Use compliance from execution environment" and set "Compiler compliance level" to 1.8, then apply.) Importing the maven project unfortunately resets that to 15.
3. Now the project should show no errors in the package explorer.
4. Open src/main/java15/com/example/MultiReleaseClass.java.
5. The editor cannot resolve `java.lang.System`:

<img width="664" height="333" alt="Image" src="https://github.com/user-attachments/assets/226fe237-3f6d-4d48-858a-0ba895eea8b1" />

tomaswolf:
Related to #4275.

laeubi:
@tomaswolf sorry I have missed this issue, I'm currently looking into this.

laeubi:
@tomaswolf I testes this now and interestingly it only happens **with Java 1.8** ... if I select e.g. Java 9 it works. So something seems special here ...

laeubi:
@tomaswolf should be fixed with:
- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4529

laeubi:
> Importing the maven project unfortunately resets that to 15.

This is something independent from MR support, but of course much more visible as one uses multiple executions of compiler mojo.

I created a PR here to fix that:
- https://github.com/eclipse-m2e/m2e-core/pull/2050

laeubi:
@tomaswolf M3 will be published soon, can you the test with the milestone repo and M2E Snapshots if you see any more issues?

josephtarbit:
[...]

PR #4529, opened 2025-10-16
Take release level into account when check for java9

laeubi:
Currently in SearchableEnvironment we only check if the project level is larger java9plus but when constructing a release specific SearchableEnvironment this needs to take the release option into account instead.

Fix https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4399

@stephan-herrmann can you review / merge this?

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ReconcilerMultiReleaseTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ReconcilerMultiReleaseTests.java
@@ -45,9 +45,8 @@ public class ReconcilerMultiReleaseTests extends ModifyingResourceTests {
 	@Override
 	public void setUpSuite() throws Exception {
 		super.setUpSuite();
-
 		IJavaProject mrproject = createJavaProject("ReconcilerMR", new String[] { "src", "src9", "src21" },
-				new String[] { "JCL18_LIB" }, "bin");
+				new String[] { "JCL_21_LIB" }, "bin", "21", true);
 		createFolder("/ReconcilerMR/src/p");
 		createFolder("/ReconcilerMR/src9/p");
 		createFolder("/ReconcilerMR/src21/p");
@@ -112,6 +111,9 @@ public class ReconcilerMultiReleaseTests extends ModifyingResourceTests {
 		mrproject.setOption(JavaCore.COMPILER_RELEASE, JavaCore.ENABLED);
 		mrproject.setOption(JavaCore.COMPILER_PB_UNUSED_LOCAL, JavaCore.IGNORE);
 		mrproject.setOption(JavaCore.COMPILER_PB_INVALID_JAVADOC, JavaCore.WARNING);
+		mrproject.setOption(JavaCore.COMPILER_COMPLIANCE, "1.8");
+		mrproject.setOption(JavaCore.COMPILER_SOURCE, "1.8");
+		mrproject.setOption(JavaCore.COMPILER_CODEGEN_TARGET_PLATFORM, "1.8");
 	}
 
 	@Override
@@ -136,6 +138,8 @@ public class ReconcilerMultiReleaseTests extends ModifyingResourceTests {
 					void m(B b) {
 						// Calling method m only available from Java 21+
 						b.m();
+						// Calling a JDK class should work
+						System.out.println("Running");
 					}
 				}
 			""";
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/SearchableEnvironment.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/SearchableEnvironment.java
@@ -115,7 +115,7 @@ public class SearchableEnvironment
 			|| !JavaCore.IGNORE.equals(project.getOption(JavaCore.COMPILER_PB_DISCOURAGED_REFERENCE, true));
 		this.workingCopies = workingCopies;
 		this.nameLookup = project.newNameLookup(workingCopies, excludeTestCode);
-		boolean java9plus = JavaCore.callReadOnly(() -> CompilerOptions
+		boolean java9plus = release >=JavaProject.FIRST_MULTI_RELEASE || JavaCore.callReadOnly(() -> CompilerOptions
 				.versionToJdkLevel(project.getOption(JavaCore.COMPILER_COMPLIANCE, true)) >= ClassFileConstants.JDK9);
 		if (java9plus) {
 			this.knownModuleLocations = new HashMap<>();


stephan-herrmann:
Could we please have a unit test? ReconcilerMultiReleaseTests seems to have all the infrastructure needed for this, right?

laeubi:
@stephan-herrmann sorry I forget about this but now was hit by the problem as well... I adjusted the test and hope it is ready for a submission now!

laeubi:
Build is green, so would be ready for submission!

stephan-herrmann:
> @stephan-herrmann sorry I forget about this but now was hit by the problem as well... I adjusted the test and hope it is ready for a submission now!

Sorry, I'm still more than busy with the fallout from https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4293

laeubi:
> > @stephan-herrmann sorry I forget about this but now was hit by the problem as well... I adjusted the test and hope it is ready for a submission now!
> 
> Sorry, I'm still more than busy with the fallout from #4293

I understand that but given only one line changes and covered by test you might consider it somewhere in between. If I can help anything let me know but it feels you are the expert and probably can handle it better than me :-)

stephan-herrmann:
FYI, the build problem is caused by https://github.com/eclipse-pde/eclipse.pde/issues/2096

laeubi:
> FYI, the build problem is caused by [eclipse-pde/eclipse.pde#2096](https://github.com/eclipse-pde/eclipse.pde/issues/2096)

I take a look into this, and hope it is resolved soon (at worst we need to revert the changes in API tools) if this is resolved do you think we can have this for M3?

laeubi:
@stephan-herrmann kindly request your review, build is green again!

stephan-herrmann:
> @stephan-herrmann kindly request your review, build is green again!

looking through my list of review requests I didn't find this PR :) 

Anyway, I found it again, looking into it now.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ReconcilerMultiReleaseTests.java
```diff
@@ -45,9 +46,16 @@ public IProblemRequestor getProblemRequestor(ICompilationUnit unit) {
 	@Override
 	public void setUpSuite() throws Exception {
 		super.setUpSuite();
-
+		//Because we use 1.8 compliance on the project, tje JCL21 wil not be setup automatically so we need to to it explicitly.
```
stephan-herrmann: 2 typos on this line

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/SearchableEnvironment.java
```diff
@@ -115,7 +115,7 @@ public SearchableEnvironment(JavaProject project, org.eclipse.jdt.core.ICompilat
 			|| !JavaCore.IGNORE.equals(project.getOption(JavaCore.COMPILER_PB_DISCOURAGED_REFERENCE, true));
 		this.workingCopies = workingCopies;
 		this.nameLookup = project.newNameLookup(workingCopies, excludeTestCode);
-		boolean java9plus = JavaCore.callReadOnly(() -> CompilerOptions
+		boolean java9plus = release >=JavaProject.FIRST_MULTI_RELEASE || JavaCore.callReadOnly(() -> CompilerOptions
```
stephan-herrmann, line 118: This change might be correct, but I am unable to verify this, since neither the test, nor the example from https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4399 needs this change. IOW, I am unable to reproduce the bug that is being fixed here. Project and editor are error free even in stock SDK 4.37.

Note: when I import the project attached to #4399 into Eclipse without this PR, the only warning I see is regarding the unknown m2e nature, as I don't have m2e installed. I did follow the steps in #4399.

stephan-herrmann (changes requested):
Please provide a test case that shows the bug (i.e., fails without the fix).

laeubi:
@stephan-herrmann I have simplified the test project find it here  now it works without the m2e part 
[mrproject_plain.zip](https://github.com/user-attachments/files/23476703/mrproject_plain.zip).

If I import that project, then go to the java15 folder and open MultiReleaseClass I get

grafik

Regarding the test it seems that I do something wrong previously used `JCL_8_LIB` and did not failed... I'll investigate on this but I'm quite sure this change is need I just seem to fail write correct reconcile tests :-\

laeubi:
I played around with the test and it seems to reproduce the issue but sometimes when switching between branches the test is not updated and give strange results... I have now put the changes on a new PR that hopefully will show the testfailure:

- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4612

laeubi:
@stephan-herrmann this one now shows the failure on master:

https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-4612/1/testReport/junit/org.eclipse.jdt.core.tests.model/ReconcilerMultiReleaseTests/testNoErrorsAfterEdit/

I have rebased my change here and now we should see no failure anymore.

stephan-herrmann:
> I have rebased my change here and now we should see no failure anymore.

The test shows a different error, which is only the result of using jclMin, which doesn't have `System.getProperty()`. And since it's a different error, it's not fixed in this PR.

laeubi:
> > I have rebased my change here and now we should see no failure anymore.
> 
> The test shows a different error, which is only the result of using jclMin, which doesn't have `System.getProperty()`. And since it's a different error, it's not fixed in this PR.

Then I need some hint from the JDT experts here, I can see the error clearly in the IDE (see screenshot above) but have no clue how to get the same setup in the tests, whatever I try I either get no errors at all or even completely different ones like java.lang.Object not found...

stephan-herrmann:
> If I import that project, then go to the java15 folder and open MultiReleaseClass I get ...

Now I see it, and here's the difference between the example project and the unit test: the latter fails to set "module=true" on the JCL_21_LIB classpath entry. Simply because the project compliance is 1.8. It seems you need to modify one more classpath entry after `createJavaProject()`

laeubi:
> It seems you need to modify one more classpath entry after createJavaProject()

I now created it as java 21 project, then set compliance manually to java 1.8... this still gives more errors than in the IDE but at least fails without and succeeds with the patch... the module=true is automatically set somewhere in the magic of the test helper classes.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/SearchableEnvironment.java
stephan-herrmann: Meanwhile I could reproduce the bug, and also the test succeeds to reproduce (without the fix). => Resolving.

stephan-herrmann (approved):
Updated test reproduces the problem, which is effectively fixed by this PR. We're good.

laeubi:
@stephan-herrmann thanks for taking care of this. As M3 is soon I likely will need to delay other work to the next release, but this bug was really a blocker so good it is solved now!

Issue #4288, opened 2025-08-06
Formatter gets confused between records and annotations


alienisty:
Given the following records:
```java
public record X(String a,
                @SuppressWarnings({ "a", "b" }) String b) {
  
}

record Y(String c){
}
```
and this formatter configuration

[code-formatter.xml.txt](https://github.com/user-attachments/files/21610236/code-formatter.xml.txt)

Then, if we format the above code we get the following:
```java
public record X(String a,
                @SuppressWarnings( {  // <-- the space should not be inserted
                  "a", "b" }) String b){ // <-- the space should be inserted

}

record Y(String c) { // <-- this works just fine
}
```
In record `X`, a space has been inserted between the opening bracket of the annotation and the opening brace, where the configuration says it should not. The space is removed, instead, between the closing bracket of the record field declaration and the opening brace of its body, where the configuration says it should insert a space.
In record `Y`, where there are no annotations, the formatter works as expected.

Eclipse Java Development Tools	3.20.300.v20250805-1800

PR #4542, opened 2025-10-21
Fix Incorrect Spacing for Left Braces in Record Declarations

SougandhS:
This commit fixes formatter recognizing wrong left braces when record contains left braces in param annotations.

Fixes : https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4288

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/META-INF/MANIFEST.MF
+++ b/org.eclipse.jdt.core.tests.model/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jdt.core.tests.model;singleton:=true
-Bundle-Version: 3.13.600.qualifier
+Bundle-Version: 3.13.700.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Export-Package: org.eclipse.jdt.core.tests,
--- a/org.eclipse.jdt.core.tests.model/pom.xml
+++ b/org.eclipse.jdt.core.tests.model/pom.xml
@@ -19,7 +19,7 @@
     <relativePath>../tests-pom/</relativePath>
   </parent>
   <artifactId>org.eclipse.jdt.core.tests.model</artifactId>
-  <version>3.13.600-SNAPSHOT</version>
+  <version>3.13.700-SNAPSHOT</version>
   <packaging>eclipse-test-plugin</packaging>
 
   <properties>


SougandhS:
Hi @mateusz-matela, could you please check this too ?

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/SpacePreparator.java
```diff
@@ -1083,6 +1083,15 @@ private void handleCommas(List<? extends ASTNode> nodes, boolean spaceBefore, bo
 
 	private void handleToken(ASTNode node, TerminalToken tokenType, boolean spaceBefore, boolean spaceAfter) {
 		if (spaceBefore || spaceAfter) {
+			if (tokenType == TokenNameLBRACE && node.getParent() instanceof RecordDeclaration record
```
mateusz-matela: This is a very general and short helper method, logic specific to `RecordDeclaration` should go to the corresponding `visit` method.

There it calls `handleToken(node.getName(), ...)`. Instead of assuming `node.getName()`, it should find `nodeBeforeLBrace` in similar way to `nodeBeforeLParen` in `handleRecordComponents` and then call `handleTokenAfter`.

mateusz-matela:
I've given just the issue description and the SpacePreparator code to Gemini AI and it solved it in one shot 🤯 
It actually went a little too careful IMO and we had an argument about it - the code it marks as "Simplified, incorrect logic" is the one I would use, but AI tries to keep its ground, you can judge who's right ;)

https://aistudio.google.com/app/prompts?state=%7B%22ids%22:%5B%221eRoK5eaGtnovX--k46eG3UJeqS2QfQ-o%22%5D,%22action%22:%22open%22,%22userId%22:%22102087613826719514078%22,%22resourceKeys%22:%7B%7D%7D&usp=sharing

SougandhS:
Thanks, I will check both the solutions you suggested

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/SpacePreparator.java
SougandhS: I tried this approach and it worked.

SougandhS: looks this affected `FormatterRegressionTests.testIssue264b()` & `FormatterRegressionTests.testIssue264a()`

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/SpacePreparator.java
```diff
@@ -178,17 +178,21 @@ public boolean visit(RecordDeclaration node) {
...
 					this.options.insert_space_after_comma_in_superinterfaces);
 		}
 
+		List<? extends ASTNode> components = node.recordComponents();
 		handleRecordComponents(node.recordComponents(), node.getName(), typeParameters);
+		ASTNode nodeBeforeLParen = typeParameters.isEmpty() ? node.getName()
+				: typeParameters.get(typeParameters.size() - 1);
+		ASTNode nodeBeforeRParen = components.isEmpty() ? nodeBeforeLParen
+				: components.get(components.size() - 1);
+		handleTokenAfter(nodeBeforeRParen, TokenNameLBRACE,
```
mateusz-matela: I think this will work, but at first glance it looks fishy to use `nodeBeforeRParen` rather than `nodeBeforeLBrace` to look for `TokenNameLBRACE`.

Why this approach? It seems to be a strange mixture of the original AI solution and my shorter one as it considers `typeParameters` but not `superInterfaces`.

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/SpacePreparator.java
```diff
@@ -178,17 +178,21 @@ public boolean visit(RecordDeclaration node) {
...
-		handleToken(node.getName(), TokenNameLBRACE,
-				this.options.insert_space_before_opening_brace_in_record_declaration, false);
 		List<Type> superInterfaces = node.superInterfaceTypes();
 		if (!superInterfaces.isEmpty()) {
 			handleTokenBefore(superInterfaces.get(0), TokenNameimplements, true, true);
 			handleCommas(superInterfaces, this.options.insert_space_before_comma_in_superinterfaces,
 					this.options.insert_space_after_comma_in_superinterfaces);
 		}
 
+		List<? extends ASTNode> components = node.recordComponents();
```
mateusz-matela, line 188: presumably this declaration is so early so it can be used in the line below?

SougandhS: Handled

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/SpacePreparator.java
SougandhS: > it looks fishy to use nodeBeforeRParen rather than nodeBeforeLBrace to look for TokenNameLBRACE.

I have reused the `nodeBeforeLBrace` itself

> but not superInterfaces

It also works with superInterfaces, I have added an additional test for it

mateusz-matela: > I have reused the `nodeBeforeLBrace` itself

What do you mean by that? There's no `nodeBeforeLBrace` variable nowhere in the code.
What I meant was that the purpose of this code is to find and handle the left brace, so it's quite unintuitive that it starts by looking for left paren and right paren. Now after your last change it's even more confusing, because it only uses `nodeBeforeLParen` but can assign a component node, which is definitely after the left paren.

> It also works with superInterfaces, I have added an additional test for it

OK, but it would also work if it didn't consider typeParameters, but it does. So this approach is kind of inconsistent.

stephan-herrmann:
FYI, the build problem is caused by https://github.com/eclipse-pde/eclipse.pde/issues/2096

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/SpacePreparator.java
SougandhS: I have refactored bit more, handling the record param in its syntax flow

mateusz-matela: OK, so you went in the direction suggested by AI.
To be frank, the original AI solution looks better. It knows that if there are `superInterfaces` then it doesn't need to check for `components` or `typeParameters` and so on, so it save cycles, but also connected if-else are more concise in code

SougandhS: I have updated with else if flow, similar to what original AI suggested..

SougandhS:
Thanks!, I'll merge this soon 👍

PR #4549, opened 2025-10-23
ECJ accepts ambiguous method call between concrete parameter and bounded type variable while javac rejects as ambiguous

### LINKED ISSUE #4357: ECJ accepts ambiguous method call between concrete parameter and bounded type variable while javac rejects as ambiguous
## ECJ Version:
```bash
Eclipse Compiler for Java(TM) v20250814-1944, 3.43.0, Copyright IBM Corp 2000, 2020. All rights reserved.
```

## Javac Version:
```bash
javac 21.0.8
```

## Description:
When compiling a method invocation that could apply to both a concrete-parameter method (`specializedMethod(String)`) and a generic method with a bounded type variable (`<U extends T> specializedMethod(U)`), ECJ and javac exhibit divergent behavior:

- javac rejects the code with a compile-time error, reporting that the method reference is ambiguous.
- ECJ  accepts the code without error or warning, silently selecting one of the overloaded methods.

## Test program:
```java
public class Test {
        TemplateClass<String> instance = new TemplateClass<>();
        instance.specializedMethod("test");
    }
}

class TemplateClass<T> {

    public void specializedMethod(String value) {
    }

    public <U extends T> void specializedMethod(U value) {
    }
}
```

Using javac it will have an error:
```bash
javac Test.java
```
```bash
Test.java:4: error: reference to specializedMethod is ambiguous
        instance.specializedMethod("test");
                ^
  both method specializedMethod(String) in TemplateClass and method <U>specializedMethod(U) in TemplateClass match
  where U,T are type-variables:
    U extends T declared in method <U>specializedMethod(U)
    T extends Object declared in class TemplateClass
1 error
```

However, ecj accepts it with no output:
```bash
java -jar ecj-4.37M3.jar -21 Test.java
# No output
```

srikanth-sankaran:
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4357

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
@@ -1286,53 +1286,6 @@ public abstract class Scope {
 	}
 
 	// Internal use only
-	public MethodBinding findExactMethod(ReferenceBinding receiverType, char[] selector, TypeBinding[] argumentTypes, InvocationSite invocationSite) {
-		CompilationUnitScope unitScope = compilationUnitScope();
-		unitScope.recordTypeReferences(argumentTypes);
-		MethodBinding exactMethod = receiverType.getExactMethod(selector, argumentTypes, unitScope);
-		if (exactMethod != null && exactMethod.typeVariables == Binding.NO_TYPE_VARIABLES && !exactMethod.isBridge()) {
-			// in >= 1.5 mode, ensure the exactMatch did not match raw types
-			for (int i = argumentTypes.length; --i >= 0;) {
-				// workaround for bug 464229: The type * cannot be resolved. It is indirectly referenced from required .class files
-				TypeBinding t = argumentTypes[i].leafComponentType();
-				if (! (t instanceof ReferenceBinding))
-					continue;
-				ReferenceBinding r = (ReferenceBinding)t;
-				if (r.isHierarchyConnected()) {
-					if (isSubtypeOfRawType(r))
-						return null;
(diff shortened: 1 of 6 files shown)


srikanth-sankaran:
I don't see any justification in JLS `15.12.2 Compile-Time Step 2: Determine Method Signature` or in `15.12.2.1 Identify Potentially Applicable Methods` for `Scope.findExactMethod(ReferenceBinding receiverType, char[] selector, TypeBinding[] argumentTypes, InvocationSite invocationSite)` - This PR is experiments by pulling the rugs from under this step.

srikanth-sankaran:
It is heading in directions that make me a bit uncomfortable. 

As can be seen from the modified test files:

    - we seem to be reading lot more class files via super class traversal in `org.eclipse.jdt.internal.compiler.lookup.Scope.findMethod0(ReferenceBinding, char[], TypeBinding[], InvocationSite, boolean)` than we just pick the `exact` method (even though that doesn't have JLS blessing)
    - we could trigger missing type referenced from a required class file more often

So we need to take a call whether this is a bug we want to stay with. As early as JDK6 we have a difference in behavior

srikanth-sankaran:
The failure in `org.eclipse.jdt.core.tests.builder.MultiProjectTests.test104_missing_required_binaries` itself is resolvable. But for reasons cited in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4549#issuecomment-3442568749 this issue may have performance side effects that have a serious impact and so I am inclined to let the sleeping dog lie.

@stephan-herrmann @jarthana - FYI. If there is a strong counter opinion let me know - current behavior has been there nearly forever.

jarthana:
> I am inclined to let the sleeping dog lie.

OK for me.

Issue #4553, opened 2025-10-25
members of a deprecated type are not implicitly deprecated


HannesWell:
Currently the `eclipse.platform.releng.aggregator` master build (and I expect also the next I-build) fails with the following compilation error:
```
[ERROR] Failed to execute goal org.eclipse.tycho:tycho-compiler-plugin:5.0.1-SNAPSHOT:compile (default-compile) on project org.eclipse.jdt.debug: Compilation failure: Compilation failure: 
[ERROR] /home/jenkins/agent/workspace/atform.releng.aggregator_PR-3432/eclipse.jdt.debug/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIStackFrame.java:[1574] 
[ERROR] 	@SuppressWarnings("deprecation")
[ERROR] 	                  ^^^^^^^^^^^^^
[ERROR] Unnecessary @SuppressWarnings("deprecation")
```

As the code didn't change for quite some time, I assume it's actually a bug in the compiler, that the deprecated method called within it, is not detected:
https://github.com/eclipse-jdt/eclipse.jdt.debug/blob/d0cd55117a2260b7cdd3f71210befbde9786b055/org.eclipse.jdt.debug/model/org/eclipse/jdt/internal/debug/core/model/JDIStackFrame.java#L1574-L1578

With the latest I-build from this afternoon, `I20251025-0900`, it was still working.

laeubi:
Maybe a side-effect of
- https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4293

but deprecation should not be touched here?!?

laeubi:
If I remove that in the IDE updated to latest ibuild I still get an error that the suppress is required.

iloveeclipse:
> Maybe a side-effect of
> 
> * [Do not emit restriction warning on inherited method #4293](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4293)
> 
> but deprecation should not be touched here?!?

This is a regression from 545483d5e69ae02a4ecbb98b1b813054abb0d0d8 / https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4293.

Considering how much discussion was done to merge https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4293 I don't expect the fix to be fast, so ideally we simply revert 545483d5e69ae02a4ecbb98b1b813054abb0d0d8 to unblock SDK build.

[...]

PR #4556, opened 2025-10-26
Add test cases for issue #4553 - inherited interface deprecated method

laeubi:
See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4553


stephan-herrmann:
I don't think this PR is needed anymore?

Issue #4531, opened 2025-10-16
Markdown Javadoc links to URLs not properly interpreted


garretwilson:
I'm using Eclipse EE 2025-09 with Java 25. I'm trying to use the new Javadoc Markdown format from [JEP 467](https://openjdk.org/jeps/467).

You'll notice here in GitHub I just added a Markdown URL link like this:

```markdown
[JEP 467](https://openjdk.org/jeps/467)
```

Unfortunately this form of link is broken in Eclipse, and doesn't show up correctly on the "Javadoc" tab.

Consider this code:

```java
/// Do nothing.
/// @see [Example Site](https://example.com/)
public void doNothing() {
}
```

In the "Javadoc" page, this shows up as:

```
Example(https://example.com/)
```

"Example" is underlined, but the link does not work. (I suspect the link is not correct because it is being interpreted as a Markdown reference link, not a Markdown URL link.) Moreover "(https://example.com/)" should not appear.

I realize that [JEP 467](https://openjdk.org/jeps/467) focuses on linking to other Java elements in the project rather than on URL links. Nevertheless this specification says:

> You can use all other forms of Markdown links, including links to URLs, but links to other program elements are likely to be the most common.

Thus URL links should be supported as well. (And I think in good documentation they will actually be quite common, linking to authoritative specifications of implementing code, for example.)

garretwilson:
Related to #4244, if you decide to do Markdown line splitting when formatting, be careful about splitting URL links. See [break long lines in markdown code](https://stackoverflow.com/questions/19985235/break-long-lines-in-markdown-code) if you want to get into the nitty-gritty. Frankly it doesn't matter so much to me whether you split links or not. You could even leave the links fully intact and consider the entire URL link to be a token. It doesn't matter so much. What matters more is splitting the line itself to stay inside the max width specified for formatting.

I imagine you can just naively split on whitespace and get away with it, which would potentially split a label inside the brackets but not the URL inside the parentheses, and that may be just fine. Alternatively you could break inside the opening parenthesis probably, as explained in the link I gave in the previous paragraph.

SougandhS:
Currently the formatter doesn't break the links on formatting

https://github.com/user-attachments/assets/10d50fd9-da4e-4769-9118-5626e0b8304a


And yes it is not rendering like in github (looks like a javaDoc specific issue)

[...]

PR #4561, opened 2025-10-27
Markdown Javadoc links to URLs not properly interpreted 4531

subyssurendran666:
Markdown links `[description](url)` syntax now interpreting properly

Fix: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4531

## How to test
Scenario 1
---------
`/// @see [Ex Si](ex.com)`

Scenario 2
----------
`/// @see [Ex Si](http://ex.com)`. --> `http/https/ http://www` similar cases

Scenario 3
----------
`/// @see [Ex Si][http://ex.com]`
## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/AbstractCommentParser.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/AbstractCommentParser.java
@@ -1398,6 +1398,40 @@ public abstract class AbstractCommentParser implements JavadocTagConstants {
 		return parseReference(false);
 	}
 
+	// Parses a complete URL reference starting from current position
+	protected boolean parseURLReference(int pos, boolean advanceEndPos) throws InvalidInputException {
+		char[]fullURL = null;
+		int firstTokenStartPos;
+		StringBuilder urlBuilder = new StringBuilder();
+		char c;
+		firstTokenStartPos = pos;
+		while (pos < this.source.length) {
+			c = this.source[pos];
+			if (c == '[') // invalid syntax for url
+				return false;
+			if (c == '(' || c == ' ') {
+				pos++;
+				continue;
+			}
+			if (c == '\n' || c == '\r' || c == ')') {
+	            break;
+	        }
+	        urlBuilder.append(c);
+	        pos++;
+		}
+		if (advanceEndPos)
+			this.index = pos;
+		fullURL = urlBuilder.toString().toCharArray();
+
+		this.identifierPtr = 0;
+		this.identifierStack[this.identifierPtr] =  fullURL;
+		this.identifierPositionStack[this.identifierPtr] = (((long) firstTokenStartPos) << 32) + (pos - 1);
+		this.identifierLengthStack[this.identifierLengthPtr] = 1;
+		Object typeRef = createTypeReference(TerminalToken.TokenNameInvalid, true);
+		pushSeeRef(typeRef);
+		return true;
+	}
+
 	/*
 	 * Parse a reference in @see tag
 	 */
@@ -3617,6 +3651,7 @@ public abstract class AbstractCommentParser implements JavadocTagConstants {
 		// Whitespace or inline tag closing brace
 		char ch = peekChar();
 		switch (ch) {
+			case ')':
 			case ']':
 				// TODO: Check if we need to exclude escaped ]
 				if (this.markdown)
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverterMarkdownTest.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverterMarkdownTest.java
@@ -1931,4 +1931,250 @@ public class ASTConverterMarkdownTest extends ConverterTestSetup {
 			assertEquals("Incorrect TextElement value", "Where is my **bold text**???", textElement.getText());
 		}
 	}
+	public void testMarkdownURLs4531_01() throws JavaModelException {
+		String source = """
+				/// @see [Ex Si](ex.com)
+				public class Markdown() {}
+				""";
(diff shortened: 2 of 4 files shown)


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/AbstractCommentParser.java
```diff
@@ -1393,6 +1393,61 @@ protected boolean parseReference() throws InvalidInputException {
...
+		StringBuilder urlBuilder = new StringBuilder();
+		char c;
+		if (this.source[this.index] == '(') {
+			firstTokenStartPos = this.index;
+			while (pos < this.source.length) {
+				c =  readChar();
+				if (c == '(' || c == '[' || c == ' ')
+					continue;
+
+		        if (c == '\t' || c == '\n' || c == '\r' || c == '*' || c == ')' || c == '[' || c == ']' ) {
```
jarthana: Not sure if we should break for anything other than \r and \n.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/AbstractCommentParser.java
```diff
@@ -1393,6 +1393,61 @@ protected boolean parseReference() throws InvalidInputException {
...
+		            break;
+		        }
+		        urlBuilder.append(c);
+			}
+			pos = this.index - 1;
+			if (validateMarkdownSyntax(this.source[firstTokenStartPos], this.source[pos]))
```
jarthana: Not sure if this validation is necessary. We should simply focus on the success case, which is encountering ')' before a \r or \n. If that happens, we should create the URL else, simply return false.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/AbstractCommentParser.java
```diff
@@ -1393,6 +1393,61 @@ protected boolean parseReference() throws InvalidInputException {
...
+				fullURL = urlBuilder.toString().toCharArray();
+		} else {
```
jarthana: We should try and get this wrapped into the IF part, if possible.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverterMarkdownTest.java
```diff
@@ -1914,4 +1914,249 @@ public class Table {}
...
+				/// @see [Ex Si)[http://ex.com)
+				public class Markdown() {}
+				""";
+		this.workingCopies = new ICompilationUnit[1];
+		this.workingCopies[0] = getWorkingCopy("/Converter_25/src/markdown/gh3761/Markdown.java", source, null);
+		if (this.docCommentSupport.equals(JavaCore.ENABLED)) {
+			CompilationUnit compilUnit = (CompilationUnit) runConversion(this.workingCopies[0], true);
+			TypeDeclaration typedeclaration =  (TypeDeclaration) compilUnit.types().get(0);
+			Javadoc javadoc = typedeclaration.getJavadoc();
+			assertEquals("Tags count does not match", 0, javadoc.tags().size());
```
jarthana: I agree with ignoring the link. But that doesn't mean the Javadoc is ignored too. We should retain the Javadoc even if it's malformed.

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/DocCommentParser.java
```diff
@@ -733,21 +733,25 @@ protected boolean parseMarkdownLinks(int previousPosition) throws InvalidInputEx
 						readChar();
 					}
 					break;
+				case ')':
```
jarthana, line 736: Treating ) same as ] is asking for trouble. We might end up processing [) instead of rejecting it

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/AbstractCommentParser.java
subyssurendran666: You are right, we do not need most of the conditions. However we may need `c == ')'`, since it calculates the end of the url. `c == ']'` does not required, since the java library references are already addressed in the parseReference method.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/AbstractCommentParser.java
subyssurendran666: I will remove `validateMarkdownSyntax()` from the validation. But I need to add a simple check inside the while loop to prevent invalid syntax from being parsed incorrectly.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverterMarkdownTest.java
subyssurendran666: Done

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/AbstractCommentParser.java
subyssurendran666: The index position decrement is already inside an if condition. Could you please guide me on which condition you would like this change to be applied?

org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/DocCommentParser.java
subyssurendran666: Done

PR #1460, opened 2023-10-02
Propagation deprecation status from enclosing context into LE

srikanth-sankaran:
## Author checklist

- [ ] I have thoroughly tested my changes
- [x] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [x] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

PR #4564, opened 2025-10-27
members of a deprecated type are not implicitly deprecated

stephan-herrmann:
Remove everything relating to AccDeprecatedImplicitly

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4553

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
@@ -3890,9 +3890,6 @@ public abstract class Scope {
 						MethodBinding context = ((AbstractMethodDeclaration) referenceContext).binding;
 						if (context != null && context.isViewedAsDeprecated())
 							return true;
-					} else if (referenceContext instanceof LambdaExpression) {
-						if (methodScope.parent.isInsideDeprecatedCode())
-							return true;
 					} else if (referenceContext instanceof ModuleDeclaration) {
 						ModuleBinding context = ((ModuleDeclaration) referenceContext).binding;
 						return context != null && context.isDeprecated();
@@ -3928,6 +3925,8 @@ public abstract class Scope {
 					}
 				}
 		}
+		if (this.parent != null)
+			return this.parent.isInsideDeprecatedCode();
 		return false;
 	}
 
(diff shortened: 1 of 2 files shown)


stephan-herrmann:
@iloveeclipse don't you like it when old code gets removed? :sunglasses: 

As this is a bit of a fundamental change, should we ask @srikanth-sankaran  to review this?

@srikanth-sankaran in case you wonder about the removal in LambdaExpression: I saw your work from #1460 but figured with `AccDeprecatedImplicitly` gone, it is simpler to improve the search in `Scope.isInsideDeprecatedCode()` rather than to propagate `AccDeprecated`  to a synthetic binding.

stephan-herrmann:
One test failure points to https://bugs.eclipse.org/214948 where JDT was on the brink of recognizing that `@Deprecated` on a package should have no effect at all, but then after an interesting discussion shied back, despite the observed difference to javac. Hihi, here @srikanth-sankaran spoke as a "newcomer"  :smile: 

From those times there are more interesting links: 
* https://bugs.java.com/bugdatabase/view_bug?bug_id=6492694
   * Evaluation: "One should be able to tag a package `@deprecated`"
   * backed by reference to https://docs.oracle.com/javase/1.5.0/docs/tooldocs/windows/javadoc.html#@deprecated

Anyway, howadays https://docs.oracle.com/javase/specs/jls/se24/html/jls-9.html#jls-9.6.4.6 explicitly states:
>  No deprecation warning or removal warning is produced when:
> [...]
> * the name of a package is used (referenced by a qualified type name, or an import declaration, or an exports or opens directive), even if the declaration of the package is annotated with `@Deprecated`.

So, yes, it is possible to annotate a package (package-info.java) with `@Deprecated` but JLS gives no meaning to this.

srikanth-sankaran:
> Hihi, here @srikanth-sankaran spoke as a "newcomer" 😄

17 years later, I am still a newcomer! 

It is scary sometimes to think that you could be held against your words almost 2 decades (and more) later 😄

stephan-herrmann:
> > Hihi, here @srikanth-sankaran spoke as a "newcomer" 😄
> 
> 17 years later, I am still a newcomer!

Now, that's news for me!

> It is scary sometimes to think that you could be held against your words almost 2 decades (and more) later 😄

No blame implied :)

I'm just always curious when a bug has been hiding in plain view for so long: how did the bug manage to remain undetected: we all have our preconceptions. Grouping is a strong concept so we tend to assume that the reputation of the family propagates to all its members.

In fact, I missed a good opportunity to find out when I learned that `@Deprecated` on a **module** doesn't propagate to its content, during Java 9 endgame. I'll look for any bug / ticket from that context.

iloveeclipse:
Note, if merging this PR we would need https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/800 to be merged too, otherwise SDK build will fail again.

Issue #4568, opened 2025-10-28
Option to inform about members of a deprecated type not being deprecated


stephan-herrmann:
From https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4553:

> I also wonder if all the deprecated types that not flagged as "for delete" should also mark all the methods in the type to be deprecated now? Before today the common assumption was that it is enough to deprecated the type, not all methods.

I agree to this proposal with two constraints:
* this diagnostic should not claim to be part of the official "deprecation" warnings
* we need an option to tune this diagnostic to the users' needs, I suggest "info" as the default level.

It would be good to deliver this with M3 so users have a good chance to react to the changes of #4553  / #4564

stephan-herrmann:
Also to note, that the new diagnostic should _not_ be suppressable with `@SuppressWarnings("deprecation")` (for the reason given above). That's why a configurable option is crucial here.

stephan-herrmann:
Question here is, _which_ members should be affected:
* public: YES
* private: NO
* protected: probably
* package private: I suggest: no

OTOH, all kinds of members should be covered: fields, methods, types.

stephan-herrmann:
https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4569 outlines what such an info could look like.

Except for Deprecated15Test, tests have not yet been updated.

While working on this I wondered:
* if users follow the recommendation and deprecate all members of deprecated types, clients will then "normally" see two deprecation warnings: one on the type and one on the member
* perhaps the info should specifically focus on the situation where a deprecated type is subclassed with a non-deprecated type.
  * I hold that this is the main leak, where members of deprecated types leak into the non-deprecated world.
  * perhaps the info should be designed similar to missing implementation of inherited abstract methods:
"type X inherits the following methods from deprecated type Y without overriding them: m1(), m2()."
  * the suggested override can then decide about deprecation or not, to give proper recommendations to clients

Two exceptions sticks out: in Java a subclass cannot override an inherited member type, nor a field. So perhaps we need a mix as to suggest:
* non-private member types and fields of deprecated types should be marked as deprecated, too
* subtypes of deprecated types should override all accessible methods

@iloveeclipse at your convenience please weigh in which kind of info you consider useful.

iloveeclipse:
> * if users follow the recommendation and deprecate all members of deprecated types, clients will then "normally" see two deprecation warnings: one on the type and one on the member

[...]

PR #4569, opened 2025-10-28
Option to inform about members of a deprecated type not being deprecated

stephan-herrmann:
Quick draft for https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4568


stephan-herrmann:
Oops, drafts are no longer built and tested?

So let's remove the draft label, while expecting tests to fail :)

PR #4603, opened 2025-11-06
Improve statement recovery in AST generated by ASTParser when there's a lambda

### LINKED ISSUE #4530: Stream chained method call doesn't show up in the AST when there are compiler problems after it
For the following code:

```
package tests;

import java.util.List;

public class TestPostCompletion {
	
	public static void main(String[] args, List<String> x) {
		int a = 4;
		Stream.iterate(0,i -> ++i).limit(10);
		"String".var
	}
}
```

which is attempting to do a Post completion for "String", the AST viewer is only showing one statement for the block under main and it is the "int a = 4;".  It is not attempting any recovery for the "String".var and not showing the ExpressionStatement that precedes it.  There shouldn't be any confusion because it is a complete statement with semi-colon.   When the Post completion tries to find the best location, it finds the declaration of a and tries to add another in the middle (invalid code).  

If the Stream statement is removed, the block now shows 2 statements with the "String".var being recovered and Post completion works because it can find the "String" node to turn into a statement declaration.

datho7561:
## What it does
Fixes #4530.

Recover statements in method bodies with syntax errors when parsing using `ASTParser`.

Also recover expressions as expression statements.

## How to test
I've added unit tests.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
@@ -1115,6 +1115,13 @@ public RecoveredElement buildInitialRecoveryState(){
 	}
 
 	if (this.statementRecoveryActivated) {
+		if (this.expressionPtr >= 0 && this.expressionStack[0] instanceof FieldReference fr) {
+			var snr = new SingleNameReference(new char[0], fr.sourceEnd);
+			snr.sourceStart = fr.sourceEnd;
+			var assignment = new Assignment(fr, snr, fr.sourceEnd);
+			assignment.statementEnd = assignment.sourceEnd;
+			element = element.add(assignment, 0);
+		}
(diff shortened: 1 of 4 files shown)


datho7561:
~~This recovers statements after the lambda statement but does not recover `"String".var` into `"String".var = $missing$;`, so the postfix completion doesn't work as expected.~~

It should recover `"String".var` into `"String".var = $missing$;` now.

datho7561:
I'm trying to remember if the error in Jenkins secretly means I've caused too many test errors for Jenkins to process or perhaps one of the test failures has a really long result string, but I'm not really sure. I usually see this error when Jenkins isn't functioning correctly.

datho7561:
Tests are failing (locally) because using this aggressive statement recovery gets rid of some of the compiler errors somehow (eg. missing `;` errors are not reported). I'll need to continue to investigate.

stephan-herrmann:
Please request reviews from @srikanth-sankaran  and myself when you have a fix without regressions. TIA.

_(This is any area we're planning to improve soonish, and hence would like to weigh in, before committing to any new strategy)._

Issue #4585, opened 2025-11-03
[25] ECJ reports a bogus "The final field t may already have been assigned" error


jarthana:
Given the code:

```
public class X {
	class TestClass<T extends Object> {
		T t;
		TestClass() {}

		TestClass(T v) {
			switch (0) {
				case 2:
					t = v;
					break;
			}
			this(); // Error here
		}
	}
}
```

Now,  I don't know why the T t is considered to be final. But the question is, can the possible assignment inside one case make the flow analysis report this error?

srikanth-sankaran:
I'll take a look tapping @stephan-herrmann as needed for help

srikanth-sankaran:
Appears to be a regression from PR https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4452 - PR https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4611 under test

PR #4611, opened 2025-11-11
[25] ECJ reports a bogus "The final field t may already have been assigned" error

srikanth-sankaran:
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4585


srikanth-sankaran:
Appears to be a regression from PR #4452

PR #4624, opened 2025-11-13
Almost complete revert of Bug 452788

stephan-herrmann:
see https://github.com/eclipse-jdt/eclipse.jdt.core/commit/bac0b688c9760eba84723fa01c9ef8fce3bf0175

Discovery from discussion in #4565: this change was not strictly backed by JLS ("reading between the lines"), and as of today it doesn't seem to be needed any more.


stephan-herrmann:
In the context of #4565 I instrumented the code to detect when during the loop over constraints from set C we add a new bound that connects ivars from different "components". Interestingly with the revert tested here, the detected cases in GenericsRegressionTest_1_8 went down from 13 to 11. So there is _some_ connection between the code under investigation here and the undesired condition.

PR #4625, opened 2025-11-13
Remove left-overs from Java 1.7 type inference

stephan-herrmann:
The following are no longer needed since compiler support for 1.7 was dropped:
* class `InferenceContext`
* methods `collectSubstitutes()` which held the only remaining references to `InferenceContext`


srikanth-sankaran:
Very welcome, I have been meaning to do this - as part of preparation of my getting reacquainted with generics and type inference. We should also get rid of WildcardBinding doubling as old intersection type via additional bounds and perhaps rename IntersectionTypeBinding18 to just IntersectionTypeBinding

Issue #4002, opened 2025-05-14
Generic record pattern not working in EJC


schosin:
I've run into an issue with the EJC in a Java 21 project. The below example does not compile within Eclipse (4.35.0), but does with javac (JDK 21, JDK 24).

The problem is line 33 and happens when Wildcard implements ComponentType with anything but `ComponentType<T>`.

```java
import java.util.List;

public class Example {

    private static boolean matches(ComponentType<?> type, ComponentType.RegularComponentType<?> eventType) {
        return switch (type) {
            case ComponentType.RegularComponentType<?> regular -> switch (regular) {
                case ComponentType.ClassType(var clazz) -> switch (eventType) {
                    case ComponentType.ClassType(var eventClazz) -> clazz == eventClazz;
                };
            };
            /* Workaround: 
            case ComponentType.Wildcard<?> wildcard -> switch (eventType) {
                case ComponentType.ClassType(var eventClazz) -> wildcard.bound().isAssignableFrom(eventClazz);
            }; 
            /**/
            case ComponentType.Wildcard(var bound) -> switch (eventType) {
                case ComponentType.ClassType(var eventClazz) -> bound.isAssignableFrom(eventClazz);
            };
            /**/
        };
    }

    sealed interface ComponentType<T> {

        sealed interface RegularComponentType<T> extends ComponentType<T> {
        }

        record ClassType<T>(Class<T> clazz) implements RegularComponentType<T> {
        }

        // "implements ComponentType<T>" works, javac (JDK 21 & JDK 24) accepts both
        record Wildcard<T>(Class<T> bound) implements ComponentType<List<? extends T>> {
        }

    }

}
```

srikanth-sankaran:
@mpalat  - thanks for taking a look. Co-tagged @stephan-herrmann as owner for any consulting and code review requirements. Thanks gentlemen.

stephan-herrmann:
@mpalat  do you still plan to work on this?

mpalat:
> [@mpalat](https://github.com/mpalat) do you still plan to work on this?

I would love to, but maybe able to do so only after Oct 5 (in another 10 days or so). However, if you feel it needs attention earlier, please feel free to reassign.

stephan-herrmann:
Inference fails when trying to reduce `⟨capture#1-of ? = List<? extends T#0>⟩`

[...]

PR #4627, opened 2025-11-13
Generic record pattern not working in EJC

stephan-herrmann:
+ remove unmotivated check parameterizedWithWildcard()
+ consistently use the captured type Tc
+ extract wildcard from the expected capture

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4002

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
@@ -2014,13 +2014,9 @@ public class InferenceContext18 {
 		return true;
 	}
 	private TypeBinding deriveTPrime(RecordPattern recordPattern, TypeBinding candidateT, InferenceVariable[] alphas, TypeBinding typeBinding) {
-		ParameterizedTypeBinding parameterizedType = null;
-		TypeBinding tPrime = null;
-		if (candidateT.isParameterizedType()) {
-			parameterizedType = InferenceContext18.parameterizedWithWildcard(candidateT);
-		}
-		if (parameterizedType != null && parameterizedType.arguments != null) {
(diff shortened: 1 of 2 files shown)


stephan-herrmann:
Fixes relate to this from JLS 18.5.5 (3):
> A type T' is derived from T, as follows:
>
> If T is a parameterized type, let Tc be the result of capture conversion ([§5.1.10](https://docs.oracle.com/javase/specs/jls/se25/html/jls-5.html#jls-5.1.10)) applied to T, and let Z1, ..., Zk (k ≥ 0) be the type variables produced by capture that are type arguments in Tc. (This includes type variables produced by the capture conversion in this step, and type variables produced by capture conversion elsewhere.) Let β1, ..., βk (k ≥ 0) be inference variables, and let θ be the substitution [Z1:=β1, ..., Zk:=βk]. T' is Tc θ.
>
> Additional bounds for β1, ..., βk are incorporated into B0 to form a bound set B1, as follows:
>
> * If βi (1 ≤ i ≤ k) replaced a type variable with ...

1. Spec asks "is a parameterized type" where our implementation used `InferenceContext18.parameterizedWithWildcard()` which returns null if no wildcard is found. This may happen when the type is already captured. The spec is explicit in handling also captures found before the explicit capture operation.
2. Spec continues to work with `Tc` (the captured type), where part of our implementation was working with `T` before capture.
3. Spec speaks of "... replaced a type variable with ...". Here "type variable" refers to a capture binding, so we should extract the wildcard binding from it for inspecting details / bounds.

stephan-herrmann:
PrimitiveInPatternsTest.testIssue2936_001() revealed another problem:

Method `IC18.getTPrimeArgumentsAndCreateBounds()` was substituting type arguments with type variables of the generic type, which had no backing by JLS. Instead the substitution should keep all original type arguments except those wildcards that are to be replaced with betas. To fix this, some signatures had to be widened to accept `TypeBinding[]` rather than `TypeVariable[]`.

PR #4644, opened 2025-11-20
Add support for JSpecify annotations

coehlrich:
## What it does
Adds support for JSpecify annotations

## Author checklist

Currently I have only started implementing support for the null mark annotations (and the `kotlin.Metadata` check is a part of the spec in the [Null-marked Scope](https://jspecify.dev/docs/spec/#null-marked-scope) section)

There are 4 [Nullness Operators](https://jspecify.dev/docs/spec/#nullness-operator):
- `UNION_NULL`: from the `Nullable` annotation and includes null
- `MINUS_NULL`: from the `NonNull` annotation and excludes null
- `UNSPECIFIED`: for unannotated code outside of null marked locations and is unspecified
- `NO_CHANGE`: used internally including for everything recognized inside null marked locations that aren't explicitly annotated where for regular types is the same as `MINUS_NULL` and for type variables uses the provided type argument

`detectStandardAnnotation` currently only returns a long and so can't currently return anything defined in `ExtendedTagBits`. How should I change `detectStandardAnnotation` to allow using `ExtendedTagBits`?

Currently there are multiple places where annotations are added to existing types internally. How should they be handled since `NullMarked` applies `NO_CHANGE` instead of `MINUS_NULL`?

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Annotation.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/Annotation.java
@@ -423,16 +423,6 @@ public abstract class Annotation extends Expression {
 				tagBits |= TagBits.AnnotationNonNull;
 			} else if (annotationType.hasNullBit(TypeIds.BitNonNullByDefaultAnnotation)) {
 				tagBits |= determineNonNullByDefaultTagBits(annotationType, valueAttribute);
-			} else if (annotationType.hasNullBit(TypeIds.BitJSpecifyNonNullAnnotation)) {
-				extendedTagBits |= ExtendedTagBits.IsJSpecifyNonNull;
-			} else if (annotationType.hasNullBit(TypeIds.BitJSpecifyNullableAnnotation)) {
-				extendedTagBits |= ExtendedTagBits.IsJSpecifyNullable;
-			} else if (annotationType.hasNullBit(TypeIds.BitJSpecifyNullMarkedAnnotation)) {
-				extendedTagBits |= ExtendedTagBits.IsJSpecifyNullMarked;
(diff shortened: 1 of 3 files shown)


stephan-herrmann:
Originally I was optimistic regarding jspecify filling a relevant gap, but it's been quite a while since. At some point I lost interest as I sensed more focus on making things different rather than unifying prior art. I haven't checked their status for a while (last time I locked it lacked specification at the level of detail that would be necessary for implementing compliant checking).

With the prospect of null-restricted types in standard Java I'm not convinced that now is a good point in time to invest much into supporting jspecify.

As you focus on null marked, can you describe the differences between that one and ecj's `@NonNullByDefault`?

coehlrich:
The main difference is:
ECJ's `NonNullByDefault` applies the `NonNull` annotation to everything in context that doesn't have a null annotation.

While JSpecify's `NonNull` annotation applies the nullness operator `MINUS_NULL` (which should be atleast roughly equivalent to ECJ's `NonNull`), the `NullMarked` annotation applies the operator `NO_CHANGE` which for regular types is equivalent to `MINUS_NULL` but for type variables it basically copies the nullness operator from the type argument.

stephan-herrmann:
> ECJ's `NonNullByDefault` applies the `NonNull` annotation to everything in context that doesn't have a null annotation.

There is more to it:
* the meaning of "everything" is fine tuned by the annotation detail of type DefaultLocation
   * note that omitting details selects the subset `{PARAMETER, RETURN_TYPE, FIELD, TYPE_BOUND, TYPE_ARGUMENT, RECORD_COMPONENT}` (excluding `ARRAY_CONTENT` and `TYPE_PARAMETER`).
* DefaultLocation.TYPE_ARGUMENT specifically excludes wildcards and type variables

Perhaps this allows users to apply `NonNullByDefault` in the exact same way as `NullMarked`? Or what exact difference remains?

Issue #4632, opened 2025-11-14
The type ... from the descriptor computed for the target context is not visible here


Tiller:
In both Eclipse 4.37 and 4.38 M2 nightly (I haven't tested on previous versions) I have the following error:
> The type HttpSecurity from the descriptor computed for the target context is not visible here.  	ProfileSecurityConfiguration.java	/test/src/main/java/test	line 14	Java Problem

From :
```
public class ProfileSecurityConfiguration {
    FrontEndSecurityCustomizer profileFrontEndSecurityCustomizer() {
        return auth -> auth.requestMatchers("test").permitAll();
    }
}
```

However, the full form works fine:
```
public class ProfileSecurityConfiguration {
    FrontEndSecurityCustomizer profileFrontEndSecurityCustomizer() {
        return new FrontEndSecurityCustomizer() {
            @Override
            public void customize(AuthorizeHttpRequestsConfigurer<HttpSecurity>.AuthorizationManagerRequestMatcherRegistry auth) {
                auth
                    .requestMatchers("test")
                    .permitAll();
            }
        };
    }
}
```

Minimal reproduction code:
https://github.com/Tiller/jdt-annot-issue/tree/lambda-issue

iloveeclipse:
> Minimal reproduction code:
> https://github.com/Tiller/jdt-annot-issue/tree/lambda-issue

May I ask for steps to create error-free Eclipse project?  I assume you example requires gradle plugins installed in the IDE, I don't have them and don't plan to install them.

tiller:
If you run `./gradlew eclipse` then import the project in eclipse it should work without the eclipse plugin. Is that enought?
Edit: eclipse plugin was not enabled in build.gradle, I udpated the PR with the plugin

iloveeclipse:
> If you run ./gradlew eclipse then import the project in eclipse 

I've tried that already, `./gradlew eclipse` says there is no such task:
```
./gradlew eclipse

[Incubating] Problems report is available at: file:///data/git/jdt-annot-issue/build/reports/problems/problems-report.html

FAILURE: Build failed with an exception.

* What went wrong:
Task 'eclipse' not found in root project 'jdt-annot-issue'.
```

tiller:
Yes sorry I did not enable the eclipse plugin, I pushed the modification

iloveeclipse:
[...]

PR #4647, opened 2025-11-21
Ensure URB's are resolved before performing visibility checks.

srikanth-sankaran:
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4632

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/META-INF/MANIFEST.MF
+++ b/org.eclipse.jdt.core.compiler.batch/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Main-Class: org.eclipse.jdt.internal.compiler.batch.Main
 Bundle-ManifestVersion: 2
 Bundle-Name: Eclipse Compiler for Java(TM)
 Bundle-SymbolicName: org.eclipse.jdt.core.compiler.batch
-Bundle-Version: 3.44.0.qualifier
+Bundle-Version: 3.44.100.qualifier
 Bundle-ClassPath: .
 Bundle-Vendor: Eclipse.org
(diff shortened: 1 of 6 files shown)


srikanth-sankaran:
This problem goes back to the original implementation of lambda expressions. While checking whether the computed SAM descriptor is "kosher", we must ensure all types mentioned in the SAM (return types, enclosing types, arguments, thrown exceptions etc.,) are visible at the lambda invocation site. While doing these checks we may end up with an `UnresolvedReferenceBinding` in one of the involved types. If that happens, we must resolve the `UnresolvedReferenceBinding` (a place holder proxy for on demand resolution into a sharp type) to compute the real type. Which is what the proposed fix does.

@stephan-herrmann - since you are fascinated by bugs that have managed to hide for a long time, you may want to study this one 😆 

@iloveeclipse @jarthana - Let me know if you want this for 4.38 - On the one hand, this bug has been there from day 1 and so perhaps could wait for 4.39. On the other hand, the fix is simple so we could consider this fix for 4.38 RC1/RC2 - I guess it is too late for RC1.

(PR still under test - I don't expect  any failures - famous last words 😆)

srikanth-sankaran:
> @jarthana - Let me know if you want this for 4.38

Jay and I discussed this offline and decided we will release this as soon as the tree opens for 4.39. I looked through bugzilla and github issues tracker for similar defect reports and didn't find any. This seems one off case and given it is not a regression, it is prudent to wait for 4.38 to ship.

iloveeclipse:
> This seems one off case and given it is not a regression, it is prudent to wait for 4.38 to ship.

I fully agree with that

Issue #1045, opened 2023-05-10
Investigate if ECJ's parser be simplified by incorporating overlooked last minute spec revisions 


srikanth-sankaran:
Per observations from @stephan-herrmann here: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/125#issuecomment-1489366453
we may have inadvertently overlooked last minute changes to Java 8 spec that may have allowed us to simplify the Parser/Scanner in ECJ.

See 

https://bugs.eclipse.org/bugs/show_bug.cgi?id=561934#c4
https://bugs.eclipse.org/bugs/show_bug.cgi?id=561934#c10

trancexpress:
Thanks for looking into this @srikanth-sankaran !

stephan-herrmann:
In an attempt to clarify what can/cannot be done I simplified the Java grammar to a minimum that still shows a remaining conflict:

[mini2.txt](https://github.com/eclipse-jdt/eclipse.jdt.core/files/13697841/mini2.txt)
Rename this file to `mini.g` and pass it to `jikespg`, it will show the dreaded
```
*** Shift/reduce conflict on "MULTIPLY" with rule 2

This grammar is not LALR(1).
```

In all the rest of the Java grammar, productions are chained in a way that anything on the RHS always has a reduced set of possibilities, ensuring that the right-most non-terminal will never branch back to the current LHS.

In the reduced version `Cast ::= LParent Identifier RParent Lambda` violates that rule, because `Lambda` may end with a generic `Expr`.

As a result this expression is ambiguous:
```
(Type) () -> a * b
```

The parser cannot decide which way to interpret (denoted by resulting AST):
* Mul ( Cast ( Lambda (Identifier)), Identifier )
* Cast ( Lambda ( Mul (Identifier, Identifier) ) )

IOW, this mini grammar (and transitively the Java grammar) cannot decide if the trailing `*b` binds to the inner expression `a` or to the entire expression so far `(Type) () -> a`.

I couldn't yet find documentation of all of those options of jikespg, if something there allows to define precedence to the "inner" rule?

Semantically, multiplying a cast with some factor _normally_ doesn't make sense, but when casting to a numerical boxing type this may still do something.

Casting a lambda, however, will never produce a numerical type, as none of those are functional interfaces, not sure if this can be turned into something useful at the syntax level ...

stephan-herrmann:
The following change would fix the mini grammar, to be LALR(1):
```
[...]

Issue #903, opened 2023-03-23
ElidedSemicolonAndRightBrace in compile error message


trancexpress:
Found during experiments for #859.

To reproduce, compile snippet:

```
public class X {
	public interface S<T> {
	    T s(T t);
	}
    public void f() {
    	S<?> s = v -> v ) ; 
    	{} // empty block to force scope handling for the lambda line
    }
}
```

Observe compile error from `ecj` (4.27):

```
----------
1. ERROR in .../X.java (at line 6)
        S<?> s = v -> v ) ;
                        ^
Syntax error on token ")", ElidedSemicolonAndRightBrace expected
----------
1 problem (1 error)
```

Compile error from `javac` (OpenJDK 17):

```
X.java:6: error: ';' expected
        S<?> s = v -> v ) ;
                       ^
1 error
```

This seems to not be expected (unsurprisingly), see `ProblemReporter.replaceIfSynthetic(String)`:

```
private String replaceIfSynthetic(String token) {
	/* Java 8 grammar changes use some synthetic tokens to make the grammar LALR(1). These tokens should not be exposed in messages
	   as it would make no sense to the programmer whatsoever. Replace such artificial tokens with some "suitable"  alternative. At
	   the moment, there are two synthetic tokens that need such massaging viz : "BeginLambda" and "BeginTypeArguments". There is a
	   third synthetic token "ElidedSemicolonAndRightBrace" that we don't expect to show up in messages since it is manufactured by
	   the parser automatically.
	*/
```

mpalat:
@trancexpress Am planning to work with a couple of newcomers (already identified) for this - hence self-assigned for now.

trancexpress:
> @trancexpress Am planning to work with a couple of newcomers (already identified) for this - hence self-assigned for now.

Sounds great! If possible, can you look into #859, #367, and #125  as well? They seems to be very related to this bug. Possibly all can be fixed with the same change. See also the comment from Stephan here: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/125#issuecomment-1489366453

There seems to be some complicated workaround for lambdas, that is maybe not necessary anymore. To my understanding, this workaround is the cause for the unexpected error message here.

srikanth-sankaran:
I don't think this qualifies as a good-first-issue. Nothing to do with DiagnoseParse qualifies as a "good" issue or a "first" issue let alone a "good first" issue (guffaw)

[...]

PR #1750, opened 2023-12-15
Very early and partial grammar changes for lambda rewiring

srikanth-sankaran:
## What it does
Attempts to rewire LambdaExpressions from being Primary expressions to top level expression
Grammar changes are not fully there - so expecting several failures. 

## Author checklist

- [ ] I have thoroughly tested my changes
- [x] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [x] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

PR #4667, opened 2025-12-01
Rewire lambda expression in the proper place in the grammar hierarchy in alignment with JLS

srikanth-sankaran:
- Rewire lambda expression in the proper place in the grammar hierarchy in alignment with JLS (Lambdas are not primary expressions, but are top level expressions)
- Eliminate the synthetic token `ElidedSemicolonAndRightBrace` and the very tricky parser code that injects this token on the fly as this is no longer needed. 
- Earlier, as the `DiagnoseParser` was not privy to of these tricks, a trip through the diagnose parser due to syntax errors could produce poor quality diagnostics. These improve with the elimination of such tricks (mostly)

* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1045
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/903

### DIFF
--- a/org.eclipse.jdt.core.tests.performance/META-INF/MANIFEST.MF
+++ b/org.eclipse.jdt.core.tests.performance/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.jdt.core.tests.performance
-Bundle-Version: 3.12.750.qualifier
+Bundle-Version: 3.12.850.qualifier
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
 Export-Package: org.eclipse.jdt.core.tests.performance,
--- a/org.eclipse.jdt.core.tests.performance/pom.xml
+++ b/org.eclipse.jdt.core.tests.performance/pom.xml
@@ -19,7 +19,7 @@
     <relativePath>../tests-pom/</relativePath>
   </parent>
   <artifactId>org.eclipse.jdt.core.tests.performance</artifactId>
-  <version>3.12.750-SNAPSHOT</version>
+  <version>3.12.850-SNAPSHOT</version>
   <packaging>eclipse-test-plugin</packaging>
   <properties>
   	<testSuite>${project.artifactId}</testSuite>


srikanth-sankaran:
The PR as of now has this grammar change - this aligns perfectly with JLS and is LALR(1) (because the various synthetic harbinger terminal tokens that continue to be injected)

```
diff --git a/org.eclipse.jdt.core.compiler.batch/grammar/java.g b/org.eclipse.jdt.core.compiler.batch/grammar/java.g
index 309a224..5435370 100644
--- a/org.eclipse.jdt.core.compiler.batch/grammar/java.g
+++ b/org.eclipse.jdt.core.compiler.batch/grammar/java.g
@@ -1751,7 +1751,6 @@
 --                   Start of rules for JSR 335
 -----------------------------------------------------------------------
 
-PrimaryNoNewArray -> LambdaExpression
 PrimaryNoNewArray -> ReferenceExpression
 /:$readableName Expression:/
 
@@ -2083,6 +2082,17 @@
 /.$putCase consumeCastExpressionLL1WithBounds(); $break ./
 CastExpression ::= PushLPAREN Name Dims AdditionalBoundsListOpt PushRPAREN InsideCastExpression UnaryExpressionNotPlusMinus
 /.$putCase consumeCastExpressionWithNameArray(); $break ./
+--------
+CastExpression ::= PushLPAREN Name OnlyTypeArgumentsForCastExpression Dimsopt AdditionalBoundsListOpt PushRPAREN InsideCastExpression LambdaExpression
+/.$putCase consumeCastExpressionWithGenericsArray(); $break ./
+CastExpression ::= PushLPAREN Name OnlyTypeArgumentsForCastExpression '.' ClassOrInterfaceType Dimsopt AdditionalBoundsListOpt PushRPAREN InsideCastExpressionWithQualifiedGenerics LambdaExpression
+/.$putCase consumeCastExpressionWithQualifiedGenericsArray(); $break ./
+CastExpression ::= PushLPAREN Name PushRPAREN InsideCastExpressionLL1 LambdaExpression
+/.$putCase consumeCastExpressionLL1(); $break ./
+CastExpression ::=  BeginIntersectionCast PushLPAREN CastNameAndBounds PushRPAREN InsideCastExpressionLL1WithBounds LambdaExpression
+/.$putCase consumeCastExpressionLL1WithBounds(); $break ./
+CastExpression ::= PushLPAREN Name Dims AdditionalBoundsListOpt PushRPAREN InsideCastExpression LambdaExpression
+/.$putCase consumeCastExpressionWithNameArray(); $break ./
 /:$readableName CastExpression:/
 
 AdditionalBoundsListOpt ::= $empty
@@ -2188,6 +2198,8 @@
 ConditionalExpression -> ConditionalOrExpression
 ConditionalExpression ::= ConditionalOrExpression '?' Expression ':' ConditionalExpression
 /.$putCase consumeConditionalExpression(OperatorIds.QUESTIONCOLON) ; $break ./
+ConditionalExpression ::= ConditionalOrExpression '?' Expression ':' LambdaExpression
+/.$putCase consumeConditionalExpression(OperatorIds.QUESTIONCOLON) ; $break ./
 /:$readableName Expression:/
 
 AssignmentExpression -> ConditionalExpression
@@ -2195,7 +2207,7 @@
 /:$readableName Expression:/
 /:$recovery_template Identifier:/
 
-Assignment ::= PostfixExpression AssignmentOperator AssignmentExpression
+Assignment ::= PostfixExpression AssignmentOperator Expression
 /.$putCase consumeAssignment(); $break ./
 /:$readableName Assignment:/
 
@@ -2238,6 +2250,8 @@
 -- has been reduced.
 Expression ::= AssignmentExpression
 /.$putCase consumeExpression(); $break ./
+Expression ::= LambdaExpression
+/.$putCase consumeExpression(); $break ./
 /:$readableName Expression:/
 /:$recovery_template Identifier:/
 
@@ -2843,6 +2857,7 @@
 /:$readableName Expression:/
 
 Expression_NotName -> AssignmentExpression_NotName
+Expression_NotName -> LambdaExpression
 /:$readableName Expression:/
 -----------------------------------------------
 -- 1.5 features : end of generics
@@ -3119,3 +3134,4 @@
 
 $end
 -- need a carriage return after the $end
+
```

srikanth-sankaran:
I am locally getting 9 distinct failures in `org.eclipse.jdt.core.tests.compiler.regression.LambdaExpressionsTest` and 10 distinct failures in `org.eclipse.jdt.core.tests.compiler.regression.NegativeLambdaExpressionsTest` and wanted to get a flavor for what is in store in the entire test suite run - but the build itself fails - so will run the tests locally and address the issues first

srikanth-sankaran:
OK, build errors gone, there now seem to be about 28 distinct failures. Will resume analysis tomorrow.

srikanth-sankaran:
An earlier very partial attempt at this can be seen at this at the abandoned PR https://github.com/eclipse-jdt/eclipse.jdt.core/pull/1750

srikanth-sankaran:
@laeubi @iloveeclipse - for this kind of a report : https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4667/checks?check_run_id=56905329431 where would I access the differing class files ? It is not immediately obvious to me - Thanks for any pointers

When the class file comparison fails on an I-build, we get an archive of differing files. Can I get similar data for CI builds ?

iloveeclipse:
You have to click on the "Details" of jenkins/pr-head entry above, and there on the Jenkins build number:

https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-4667/9/

srikanth-sankaran:
> You have to click on the "Details" of jenkins/pr-head entry above, and there on the Jenkins build number:
> 
> https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-4667/9/

Thanks - that is useful.

For posterity, this is what I had to do:

From the main PR page, I clicked the failing check `Compiler`. This took me to https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4667/checks?check_run_id=56905329431 
From that page I clicked `View more details on Jenkins - Eclipse JDT` to be taken to https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-4667/9/eclipse/
From this page, if I click on `Status` on the top left, I get to see the comparator failing files.

srikanth-sankaran:
OK, those comparator differences are expected and don't account for the 477 tests failures seen in CI. These tests are not failing locally, so something mysterious is happening here

srikanth-sankaran:
> OK, those comparator differences are expected and don't account for the 477 tests failures seen in CI. These tests are not failing locally, so something mysterious is happening here

Not only are these tests not failing locally - they don't fail in an inner eclipse instance launched from a JDT-Core development workspace and into which all jdt source and tests projects are imported as source projects. 

Plot thickens 🤔

srikanth-sankaran:
Here is one interesting observation - all failing tests are from the package `org.eclipse.jdt.core.tests.compiler.parser` and these tests fail only in 1.8 mode and seemingly pass in all other compliance modes -

iloveeclipse:
Just a guess
- something local is not committed
- local build does not build (or clean) same bits like remote
- execution of tests in Eclipse uses different arguments

You could run build and tests locally by maven to verify that.

srikanth-sankaran:
> Just a guess
> 
> * something local is not committed
> * local build does not build (or clean) same bits like remote
> * execution of tests in Eclipse uses different arguments
> 
> You could run build and tests locally by maven to verify that.

I would/could readily rule out the first two, the maven run suggestion I will try after resorting to some easier (for me) experiments such as rolling back the last change - these failures started suddenly (earlier there were failures but along expected lines and locally reproducible). I don't see how the last change itself could be connected but it would be a worthwhile exercise to roll back to a cleaner state and move forward.

(note to self - I need to stop amending previous commits 😧)

srikanth-sankaran:
Backing out the recent change still produces the same 477 AIOOBs - I am 100% certain these failures were NOT there before the last one line change to the grammar file - I suspect some flakiness in github CI infra

To check all is well, I created a dummy PR https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4669 - that fails with: 

15:15:13  org.eclipse.jdt.core.tests.compiler.regression.LocalStaticsTest.testBug564557AnnotInterface_001 - 21
Creating placeholder flownodes because failed loading originals.
ERROR: Cannot resume build because FlowNode 34 for FlowHead 1 could not be loaded. This is expected to happen when using the PERFORMANCE_OPTIMIZED durability setting and Jenkins is not shut down cleanly. Consider investigating to understand if Jenkins was not shut down cleanly or switching to the MAX_SURVIVABILITY durability setting which should prevent this issue in most cases.
[GitHub Checks] GitHub check (name: Jenkins, status: completed) has been published.

mpalat:
Updating with some related info on an offline conversation from Srikant:
- [Artifact here](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-4667/15/artifact/)
- [log here](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-4667/15/artifact/org.eclipse.jdt.core.tests.model/target/work/data/.metadata/.log/*view*/)
- Launcher [Command-line link](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-4667/15/pipeline-overview/?start-byte=42469027&selected-node=34#log-34-1476)

Additional (maybe useless) info - @srikanth-sankaran : I did a parser generation of the "new" java.g from the Pulled PR. PR looks good - no omissions/delta after the generation as well - this was just to double check. Tests are passing on my machine.

Note: Just did a [rebuild](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-4667/16/) as well in case it turns out to be a timeout issue

srikanth-sankaran:
> You could run build and tests locally by maven to verify that.

This calls for a bit more set up and skills than I have readily at my disposal :)

@mpalat @jarthana and I did some brainstorming and decided to take the equivalent route of downloading the built bits from github CI infra itself.  From https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-4667/15/artifact/repository/target/, download the zipped bits, explode them and into a clean workspace import these as `binary projects with linked content` under `import plugins and fragments`.

This establishes the JDT projects in a debuggable workspace exactly incorporating the bits from jenkins.

Now I imported the test projects from the target platform and ran the failing tests and voila! I was able to reproduce the problem.

Now here is the thing that surprised me totally. When the build reports a class file comparator failure, I always assumed that it is flagging a unchanged source file that has resulted in a changed class file **that needed to be validated for the quality gate check to pass and for the PR to be integrated.** I had paid scant attention to the part of the diagnostic about the concerned artifact being changed and **replaced with the baseline version**!!! That step in the present instance replaced valid new changed class file (change due to externally defined inlined constant) with a version with the stale value of the constant. 

So when class file differences are reported in the checks steps, the required protocol seem to be:

- ignore test results (for the time being)
- validate class file change, if warranted force qualifier update, trigger a rebuild and wait for **fresh** test result

as opposed to:

 - validate class file change
 - Use test results as a further validation that the new code generation is good ---- This is WRONG assumption because it is _not_ testing the new code generation but baseline version of class file for source files unchanged in the PR.

That the logs contain a very clear message about the artifact being replaced with baseline version is undeniably true, but _the assumption that a class file can change only when the corresponding source has changed is a very brittle one_!

srikanth-sankaran:
The tests will take a while to finish but it is good news that I see some of the tests were failing earlier with `AIOOB` now passing

srikanth-sankaran:
> The tests will take a while to finish

Hoorah! All green - this basically refactors the lambda grammar to align closer with JLS. But would/could such alignment result in some simplifications (such as eliminating synthetic terminal injection or vanguard parser detours) - that is to be taken up next week!

srikanth-sankaran:
Thanks @iloveeclipse @mpalat @jarthana for your tips and inputs in reproducing the failures seen in Jenkins. Appreciated!

srikanth-sankaran:
> Now here is the thing that surprised me totally. When the build reports a class file comparator failure, I always assumed that it is flagging a unchanged source file that has resulted in a changed class file **that needed to be validated for the quality gate check to pass and for the PR to be integrated.** I had paid scant attention to the part of the diagnostic about the concerned artifact being changed and **replaced with the baseline version**!!! That step in the present instance replaced valid new changed class file (change due to externally defined inlined constant) with a version with the stale value of the constant.
> 
> So when class file differences are reported in the checks steps, the required protocol seem to be:
> 
> * ignore test results (for the time being)
> * validate class file change, if warranted force qualifier update, trigger a rebuild and wait for **fresh** test result
> 
> as opposed to:
> 
> * validate class file change
> * Use test results as a further validation that the new code generation is good ---- This is WRONG assumption because it is _not_ testing the new code generation but baseline version of class file for source files unchanged in the PR.
> 
> That the logs contain a very clear message about the artifact being replaced with baseline version is undeniably true, but _the assumption that a class file can change only when the corresponding source has changed is a very brittle one_!

@stephan-herrmann - I checked with @mpalat and @jarthana about this - they confess to being as surprised as I am. What is your reaction ?

stephan-herrmann:
> @stephan-herrmann - I checked with @mpalat and @jarthana about this - they confess to being as surprised as I am. What is your reaction ?

Already in the past I have been confused about the various baseline-related checks performed by the build. Your latest observation adds to that confusion :)

My conclusion: we need to enhance our understanding of what the build does and why.
* Looking at the build overview like https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-4667/16/ is not enough, it doesn't show the crucial "has been replaced" message!
* In this case it's the execution of goal [tycho-p2-plugin:p2-metadata-default](https://tycho.eclipseprojects.io/doc/main/tycho-p2-plugin/p2-metadata-default-mojo.html) that breaks the build
  * It's the parameter [`<baselineReplace>`](https://tycho.eclipseprojects.io/doc/main/tycho-p2-plugin/p2-metadata-default-mojo.html#baselineReplace) that seems to trigger the unwanted replacement.
      * IIUC the intention is exclusively to optimize the final p2 metadata produced by the build.
      * From that description it's a long way to understanding why it could possibly affect tests!!!!
      * I recall being confused by this replacement specifically in builds for BETA_ branches, though I have no exact data here.
      * Perhaps we can disable this behavior for CI builds?
* The above shouldn't be confused with other goals like [tycho-p2-extras:compare-version-with-baselines](https://tycho.eclipseprojects.io/doc/main/tycho-extras/tycho-p2-extras-plugin/compare-version-with-baselines-mojo.html)

srikanth-sankaran:
This next revision of the PR gets rid of the synthetic token `ElidedSemicolonAndRightBrace` and also eliminates the need for various sleights of hand originally implemented to parse expression bodied lambda expressions.

Passes `org.eclipse.jdt.core.tests.compiler.regression.LambdaExpressionsTest` and `org.eclipse.jdt.core.tests.compiler.regression.NegativeLambdaExpressionsTest` (except for one fake failure which needs test tweak)

Let us see what else breaks.

Yet to rigorously verify grammar is equivalent.

srikanth-sankaran:
The PR is green with jenkins succeeding in 59m - that looks suspicious!

stephan-herrmann:
> The PR is green with jenkins succeeding in 59m - that looks suspicious!

why do yo say suspicious? Do you suspect a performance problem? Looking at https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/view/change-requests/ I see recent successful builds between one and two hours. 59m would be at the fast end of the spectrum, not an outlier in any sense.

srikanth-sankaran:
> why do yo say suspicious? Do you suspect a performance problem? Looking at https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/view/change-requests/ I see recent successful builds between one and two hours. 59m would be at the fast end of the spectrum, not an outlier in any sense.

At that point, the PR was still in early larval stage with expected test failures - It couldn't have legitimately passed all green!

srikanth-sankaran:
@stephan-herrmann - I think this PR is ready for review and integration. While more simplification/improvements may be possible, this is a logical unit of work from a review/integration pov. Further simplifications may happen on their own tickets.

You may choose to ignore the various comments on this PR apart from the top level description which is brought upto date and the "Notes for the reviewer" I am providing my annotating the various code hunks in the files changed.

org.eclipse.jdt.core.compiler.batch/grammar/java.g
```diff
@@ -2085,6 +2078,22 @@ CastExpression ::= PushLPAREN Name Dims AdditionalBoundsListOpt PushRPAREN Insid
...
+/.$putCase consumeCastExpressionWithQualifiedGenericsArray(); $break ./
+CastedLambdaExpression ::= PushLPAREN Name PushRPAREN InsideCastExpressionLL1 NakedOrCastedLambdaExpression
+/.$putCase consumeCastExpressionLL1(); $break ./
+CastedLambdaExpression ::=  BeginIntersectionCast PushLPAREN CastNameAndBounds PushRPAREN InsideCastExpressionLL1WithBounds NakedOrCastedLambdaExpression
+/.$putCase consumeCastExpressionLL1WithBounds(); $break ./
+CastedLambdaExpression ::= PushLPAREN Name Dims AdditionalBoundsListOpt PushRPAREN InsideCastExpression NakedOrCastedLambdaExpression
+/.$putCase consumeCastExpressionWithNameArray(); $break ./
+/:$readableName CastedLambdaExpression:/
+
 AdditionalBoundsListOpt ::= $empty
```
srikanth-sankaran, line 2097: JLS grammar reads:

```
CastExpression:
( PrimitiveType ) UnaryExpression
( ReferenceType {AdditionalBound} ) UnaryExpressionNotPlusMinus
( ReferenceType {AdditionalBound} ) LambdaExpression
```

Essentially, where a `LambdaExpression` ceases to be a primary, we need to wire it expressly into `CastExpression` productions so casts are applicable to them

org.eclipse.jdt.core.compiler.batch/grammar/java.g
```diff
@@ -2188,14 +2197,16 @@ ConditionalOrExpression ::= ConditionalOrExpression '||' ConditionalAndExpressio
 ConditionalExpression -> ConditionalOrExpression
 ConditionalExpression ::= ConditionalOrExpression '?' Expression ':' ConditionalExpression
 /.$putCase consumeConditionalExpression(OperatorIds.QUESTIONCOLON) ; $break ./
+ConditionalExpression ::= ConditionalOrExpression '?' Expression ':' NakedOrCastedLambdaExpression
+/.$putCase consumeConditionalExpression(OperatorIds.QUESTIONCOLON) ; $break ./
 /:$readableName Expression:/
```
srikanth-sankaran, line 2202: JLS rules read:

```
ConditionalExpression:
ConditionalOrExpression
ConditionalOrExpression ? Expression : ConditionalExpression
ConditionalOrExpression ? Expression : LambdaExpression
```

We will need a similar update to `ConditionalExpression`'s doppleganger `ConditionalExpression_NotName`

org.eclipse.jdt.core.compiler.batch/grammar/java.g
```diff
@@ -2188,14 +2197,16 @@ ConditionalOrExpression ::= ConditionalOrExpression '||' ConditionalAndExpressio
...
 
 AssignmentExpression -> ConditionalExpression
 AssignmentExpression -> Assignment
 /:$readableName Expression:/
 /:$recovery_template Identifier:/
 
-Assignment ::= PostfixExpression AssignmentOperator AssignmentExpression
+Assignment ::= PostfixExpression AssignmentOperator Expression
 /.$putCase consumeAssignment(); $break ./
```
srikanth-sankaran, line 2210: JLS rule:

```
Assignment:
LeftHandSide AssignmentOperator Expression
```

In https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1045#issuecomment-3717405795 you write: `Need to check if we break anything regarding: ... assignment expressions` - I believe this change ensures that a lambda or a casted lambda can participate in assignment even after ceasing to being a primary

org.eclipse.jdt.core.compiler.batch/grammar/java.g
```diff
@@ -2234,10 +2245,11 @@ AssignmentOperator ::= '|='
...
--- For handling lambda expressions, we need to know when a full Expression
--- has been reduced.
-Expression ::= AssignmentExpression
-/.$putCase consumeExpression(); $break ./
+NakedOrCastedLambdaExpression -> LambdaExpression
+NakedOrCastedLambdaExpression -> CastedLambdaExpression
+
+Expression -> AssignmentExpression
+Expression -> NakedOrCastedLambdaExpression
 /:$readableName Expression:/
```
srikanth-sankaran, line 2253: Hook both `LambdaExpression` (with JLS blessings) and `CastedLambdaExpression` (without JLS's express blessings) into top level `Expression` via a level of indirection through `NakedOrCastedLambdaExpression` - I will separately write a proof of correctness for (with caveats) hooking `CastedLambdaExpression` directly to top level expression instead of  to `UnaryExpressionNotPlusMinus` or `UnaryExpressionNotPlusMinus_NotName` as normally happens with all other `CastExpression`s

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
```diff
@@ -8366,33 +8385,10 @@ protected void consumeBracketedTypeElidedUnderscoreLambdaParameter() {
...
-		this.currentToken = TokenNameElidedSemicolonAndRightBrace; // conjure a rabbit out of the hat ...
-		Expression exp = this.expressionStack[this.expressionPtr--];
-		this.expressionLengthPtr--;
-		pushOnAstStack(exp);
-	}
+protected void consumeLambdaBodyExpression() {
+	Expression exp = this.expressionStack[this.expressionPtr--];
+	this.expressionLengthPtr--;
+	pushOnAstStack(exp);
 }
```
srikanth-sankaran, line 8392: `consumeExpression()` would transfer lambdas from expression stack to AST stack - this still needs to happen - the new method accomplishes that

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
```diff
@@ -11572,11 +11567,11 @@ We do this by claiming to the automaton that the next token seen is the (suitabl
...
-		token = token == TokenNameLPAREN ? TokenNameBeginLambda : TokenNameBeginTypeArguments;
-	}
-
-	return automatonWillShift(token, this.unstackedAct);
+	return switch (token) {
+		case TokenNameLPAREN -> automatonWillShift(TokenNameBeginLambda, this.unstackedAct) || automatonWillShift(TokenNameBeginIntersectionCast, this.unstackedAct);
+		case TokenNameLESS -> automatonWillShift(TokenNameBeginTypeArguments, this.unstackedAct);
+		default -> automatonWillShift(token, this.unstackedAct);
+	};
 }
```
srikanth-sankaran, line 11576: This code change is not strictly required, but makes things explicit. Earlier, `org.eclipse.jdt.internal.compiler.parser.Scanner.maybeAtLambdaOrCast()` would depend on the possible locations for both a lambda parameter list and a cast expression identical. This has not changed, but the code prepares for a possible future change.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
```diff
@@ -11618,7 +11613,12 @@ protected void parse() {
 				stackLength);
 		}
 		this.stack[this.stateStackTop] = act;
+		int oldAct = act;
 		this.unstackedAct = act = actFromTokenOrSynthetic(act);
+		if (DEBUG_AUTOMATON) {
+			if (act > NUM_RULES && act <= ERROR_ACTION)
+				System.out.println("Shift on " + this.currentToken + " from " + original_state(oldAct) + " to " + original_state(act));  //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
+		}
```
srikanth-sankaran, line 11622: These various changes to code under `if (DEBUG)` and if (DEBUG_AUTOMATON)` produce much better debug traces ... Not integral to the fix itself.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/diagnose/DiagnoseParser.java
```diff
@@ -2303,10 +2299,6 @@ private void reportPrimaryError(int msgCode, int nameIndex, int token, int scope
 	            }
 
 				if (scopeNameIndex != 0) {
-					if (insertedToken == TerminalToken.TokenNameElidedSemicolonAndRightBrace.tokenNumber()) {
-						reportMisplacedConstruct(errorStart, errorEnd, false);
-						break;
-					}
 					if(this.reportProblem) problemReporter().parseErrorInsertToComplete(
```
srikanth-sankaran, line 2302: This hack to keep internal/artificial tokens is not relevant anymore

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/parser/ParserTest.java
```diff
@@ -810,12 +810,12 @@ public void testBug485477() {
...
 		"	Object o = T.super; // error: \'.\' expected\n" +
 		"	             ^^^^^\n" +
-		"Syntax error, insert \". Identifier\" to complete Expression\n" +
+		"Syntax error, insert \". Identifier\" to complete VariableInitializer\n" +
 		"----------\n" +
 		"2. ERROR in T.java (at line 7)\n" +
 		"	Object o2 = T.super;\n" +
 		"	              ^^^^^\n" +
-		"Syntax error, insert \". Identifier\" to complete Expression\n" +
+		"Syntax error, insert \". Identifier\" to complete VariableInitializer\n" +
```
srikanth-sankaran, line 818: Grammar changes induced changes in diagnostics produced by `DiagnoseParser`

In eyeballing through them, I see significantly cleaner messages in many instances, equivalent grade messages in a few others and very rarely degraded but still correct messages.

This one is equivalent grade message

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/ExpressionContextTests.java
```diff
@@ -613,8 +608,8 @@ public void test021() {
 			"----------\n" +
 			"1. ERROR in X.java (at line 6)\n" +
 			"	return \"Hello\" + () -> {};\n" +
-			"	                 ^^^^^\n" +
-			"The target type of this expression must be a functional interface\n" +
+			"	               ^\n" +
+			"Syntax error on token \"+\", invalid AssignmentOperator\n" +
 			"----------\n" +
```
srikanth-sankaran, line 613: Lambdas are not primaries and so can't participate in arithmetic (at syntax level itself)

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/LambdaExpressionsTest.java
```diff
@@ -8989,6 +8989,174 @@ public static void main(String[] args) {
 			},
 			"class X$A");
 }
+public void testLambdaAsAssignmentRHS() {
+	this.runConformTest(
```
srikanth-sankaran, line 8993: These tests are to verify the concern expressed in https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1045#issuecomment-3717405795 viz `Need to check if we break anything regarding: ... ternaries`

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/NegativeLambdaExpressionsTest.java
```diff
@@ -7862,18 +7862,8 @@ public void test405134a() {
...
-			"	^\n" +
-			"Syntax error, insert \";\" to complete FieldDeclaration\n" +
-			"----------\n" +
-			"3. ERROR in X.java (at line 12)\n" +
-			"	}\n" +
-			"	^\n" +
-			"Syntax error, insert \"}\" to complete InterfaceBody\n" +
+			"	                                                               ^^^^^^^^^^\n" +
+			"Syntax error, insert \";\" to complete ReturnStatement\n" +
 			"----------\n",
```
srikanth-sankaran, line 7867: Improved error messages in these cases

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/NegativeLambdaExpressionsTest.java
```diff
@@ -10391,8 +10387,228 @@ public void test553601() {
 		"----------\n" +
 		"1. ERROR in TestOptional3.java (at line 10)\n" +
 		"	.orElseThrow(() -> new Exception()))))))))))))))))))))))))))))))))))))))))))))));\n" +
+		"	                                 ^\n" +
```
srikanth-sankaran, line 10390: One instance where diagnostics degrade while still being right

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/SwitchExpressionsYieldTest.java
```diff
@@ -5769,7 +5769,7 @@ public void testBug565844_07() {
 				"1. ERROR in X.java (at line 5)\n" +
 				"	case AAABBB -> 1;\n" +
 				"	                ^\n" +
-				"Syntax error on token \";\", [ expected\n" +
+				"Syntax error on token \";\", , expected\n" +
 				"----------\n" +
```
srikanth-sankaran, line 5773: Neither makes sense ? :)

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/SwitchExpressionsYieldTest.java
```diff
@@ -6141,20 +6136,9 @@ public void testBug567112_001() {
...
-				"	}\n" +
-				"	^\n" +
-				"Syntax error, insert \")\" to complete Expression\n" +
-				"----------\n" +
-				"3. ERROR in X.java (at line 8)\n" +
-				"	}\n" +
-				"	^\n" +
+				"	                                                ^\n" +
 				"Syntax error, insert \";\" to complete BlockStatements\n" +
-				"----------\n"
```
srikanth-sankaran, line 6156: Better diagnostics

srikanth-sankaran:
Here is an analysis of the effects of hooking `CastedLambdaExpression` to top level `Expression` (via `NakedOrCastedLambdaExpression`) instead of hooking it to `UnaryExpressionNotPlusMinus` or `UnaryExpressionNotPlusMinus_NotName` as is done for "normal" `CastExpression` from which the productions for  `CastedLambdaExpression` are cloned to implement the JLS rule: 
`CastExpression: ( ReferenceType {AdditionalBound} ) LambdaExpression`

By ceasing to be `UnaryExpressionNotPlusMinus`, a `CastedLambdaExpression`  will fail to reduce (transitively) to the following non-terminals:

- `UnaryExpression` (and so cannot be operated upon by the unary operators {+, -, ~, !, ++, --} or be cast to a primitive type)
- `MultiplicativeExpression` (and so cannot be operated on by the binary ops {*,/, %})
- `AdditionExpression` (and so cannot be operated on by the binary ops {+, -})
- `ShiftExpression` (and so cannot be operated on by the shift operators {<<. >>, >>>}
- `RelationalExpression` (and so cannot be operated on by the relational operators {<, >, <=, >=})
- `InstanceofExpression` (more on this below separately)
- `EqualityExpression`  (more on this below separately)
- `AndExpression` (and so cannot be operated on by the bitwise &)
- `ExclusiveOrExpression` (and so cannot be operated on by the bitwise ^)
- `InclusiveOrExpression` (and so cannot be operated on by the bitwise |)
- `ConditionalAndExpression` (and so cannot be operated on by the bitwise &&)
- `ConditionalOrExpression`  (and so cannot be operated on by the bitwise ||)
- `ConditionalExpression` - This means a `CastedLambdaExpression` cannot participate in the **existing** ternary ?: rules. In order to facilitate their participation, we have cloned the concerned productions - so we are good.
- `AssignmentExpression` = This means a `CastedLambdaExpression` cannot participate in the **existing** rule for `Assignment`  namely `Assignment ::= PostfixExpression AssignmentOperator AssignmentExpression` but we have amended it to be `Assignment ::= PostfixExpression AssignmentOperator Expression` - so we are good.
- Expression` - but we have directly hooked it to top level - so we are good.

Basically, this means a lambda expression in the new grammar cannot participate **at the syntax level** in any operations involving the above operators - This is a non-issue as the language does not define any operations on lambdas using these operators anyways. So no valid program could be affected.

(Same chain of reasoning holds for the mirror productions XYZ_NotName)

So in essence correctness/equivalence of grammar boils down to point raised in comment https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1045#issuecomment-1859318327 by @stephan-herrmann which is quoted here for convenience:

```
A casted lambda expression can actually be used as an operand in certain expressions.

Need to check if we break anything regarding:

instanceof
!= and == comparisons
ternaries
assignment expressions
```

I have already established that for ternaries and assignment expressions we do the right thing. 

For instanceof here is the situation: For expression bodied lambdas, an instanceof check must first wrap the casted lambda into a parenthesis pair () in which case instanceof would still work. But the code fragment from https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1045#issuecomment-3713769639 reproduced here:

```
import java.util.function.Supplier;

public class X {
		boolean b = (Supplier<String>) () -> { return "Hello"; } instanceof Supplier<String>;
	}
}
```


Likewise the following won't compile with the new grammar:

```
import java.util.function.Supplier;

public class X {
		if ((Supplier<String>) () -> { return "Hello"; } != null)
		    System.out.print("yes");
		if ((Supplier<String>) () -> { return "Hello"; } == null)
		    System.out.print("yes");
	}
}
```

When the casted lambda is parenthesized each would compile.

I think instanceof checking a casted lambda (a naked lambda cannot be checked thus) or equality checking of casted lambda (a naked lambda cannot be checked thus) is a historic accidental quirk. As the lambda would be garbage collected before it can be used and the computed results are compile time constants, I say don't bother.

srikanth-sankaran:
@stephan-herrmann - Please review. 

Credit where credit is due: This is essentially your idea, I have simply validated/double checked much of your analysis in https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1045 and provided an implementation. Anything that takes away things that happen behind the back of DiagnoseParser is very welcome. Thanks for driving this!

Perhaps we can take a look at the other synthetic tokens one by one to see if we are able to get rid of anymore ? But on different tickets ...

srikanth-sankaran:
Rebased and ready to go, hint, hint! ;-)

org.eclipse.jdt.core.compiler.batch/grammar/java.g
stephan-herrmann: For the fun of it I created some "interesting" nonsense which ecj (with or without this change) has no problems compiling:
```java

public class C {
	Function<String,Integer> f;
	Function<String,Integer> get() {
		if (f != null)
			return f;
		return f = _ -> 1;
	}
	C() {
		if ((f = (Function<String,Integer>) _ -> null) == get())
			System.out.println("equal");
		else
			System.out.println("not equal");
	}
		new C();
	}
}
```
:tada:

The parser still needs parens around the assignment expression inside the if condition, but this is not specific to lambdas.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
stephan-herrmann: The `default` branch makes it look as if we need to run the automaton for each and every token. Only by looking at callers of this method, I can see that `default` here is an alias for `case TokenNameAT`, isn't it?

What do you think about this even more explicit variant?
```java
	return switch (token) {
		case TokenNameLPAREN -> automatonWillShift(TokenNameBeginLambda, this.unstackedAct) || automatonWillShift(TokenNameBeginIntersectionCast, this.unstackedAct);
		case TokenNameLESS -> automatonWillShift(TokenNameBeginTypeArguments, this.unstackedAct);
		case TokenNameAT -> automatonWillShift(token, this.unstackedAct);
		default -> throw new IllegalArgumentException("unexpected token "+token); //$NON-NLS-1$
	};
```

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
```diff
@@ -11721,29 +11716,24 @@ protected void parse() {
...
-				System.out.println(name[non_terminal_index[lhs[act]]]);
-			}
-
 			this.stateStackTop -= (rhs[act] - 1);
 			this.unstackedAct = ntAction(this.stack[this.stateStackTop], lhs[act]);
 			consumeRule(act);
+			if (DEBUG_AUTOMATON) {
+				System.out.print("Reduce by rule " + act + " "); //$NON-NLS-1$ //$NON-NLS-2$
+				if (!DEBUG)
+					System.out.print(name[non_terminal_index[lhs[act]]]);
```
stephan-herrmann, line 11726: I have always been puzzled about the difference between `DEBUG` and `DEBUG_AUTOMATON`, but now I can no longer hold back this question.

Why should this second print _not_ happen when `DEBUG` is set?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/diagnose/DiagnoseParser.java
stephan-herrmann: Wow, I never would've dared to touch this bit, but seeing the original author from 2012 removing stuff which he calls a hack is a reason to celebrate!  :fireworks: 

In trying to appear smart I still have a question: are you sure that none of the remaining synthetic tokens need this?

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/NegativeLambdaExpressionsTest.java
```diff
@@ -8528,7 +8518,7 @@ public void test426965() {
 		"");
 }
 // https://bugs.eclipse.org/bugs/show_bug.cgi?id=427207, - [1.8][bytecode] Runtime type problem: Instruction type does not match stack map
-public void test427207() {
+public void test427207() { // here
```
stephan-herrmann: what's "here"? :grinning:

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/NegativeLambdaExpressionsTest.java
stephan-herrmann: silly bug calls for silly errors :smile:

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/SwitchExpressionsYieldTest.java
stephan-herrmann: don't tell anybody! :)

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/SwitchExpressionsYieldTest.java
stephan-herrmann: :+1:

org.eclipse.jdt.core.compiler.batch/grammar/java.g
stephan-herrmann: I see you have an elaborate set of new tests in this area already. Resolving.

stephan-herrmann:
> ... I have simply validated/double checked ...

This is a grave misuse of the word "simply". Thanks for all your diligence! All remaining questions I could think of you can find in comments to comments. Nothing severe. Please proceed as you see fit.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
srikanth-sankaran: `if (DEBUG)` is true, then consumeRule() is going to print the rule anyway, so there would be duplicate output. 

That said, I think we may as well merge the two control variables. A quick search shows `DEBUG` is used primarily only in `consumeRule` with the other two uses in `parse` being general enough to make sense for `DEBUG_AUTOMATON`

There is another problem with if (DEBUG) outputs: these truncate the RHS of the rule if it is long. For example,

```
    case 94 : if (DEBUG) { System.out.println("InternalCompilationUnit ::= PackageDeclaration"); }  //$NON-NLS-1$
		    consumeInternalCompilationUnit(); 			break;

    case 95 : if (DEBUG) { System.out.println("InternalCompilationUnit ::= PackageDeclaration..."); }  //$NON-NLS-1$
		    consumeInternalCompilationUnit(); 			break;

    case 96 : if (DEBUG) { System.out.println("InternalCompilationUnit ::= PackageDeclaration..."); }  //$NON-NLS-1$
		    consumeInternalCompilationUnitWithTypes(); 			break;

    case 97 : if (DEBUG) { System.out.println("InternalCompilationUnit ::= PackageDeclaration..."); }  //$NON-NLS-1$
		    consumeInternalCompilationUnitWithTypes(); 			break;

```

Based on the output, it would not be possible to say which rule was actually reduced as they all share a common prefix.

I have raised https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4741 for follow up

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
srikanth-sankaran: Will opt for this hunk instead, thanks!

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/NegativeLambdaExpressionsTest.java
srikanth-sankaran: What's where ? 🤣

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/diagnose/DiagnoseParser.java
srikanth-sankaran: Easy part: Removal of this code is really removal of dead code as the token `TokenNameElidedSemicolonAndRightBrace` ceases to exist and so any blocks that are conditionally executed on 
`token == TokenNameElidedSemicolonAndRightBrace` are dead and can be pruned along with any variable assignments that are used only in such blocks.

More involved part: I think for historic reasons (which is an euphemism for `I don't recall` 😆), surfacing of the synthetics is worked around in multiple places and we are covered I would say. This PR itself does not change anything about how other synthetics are handled. So I would say we are good.

See `org.eclipse.jdt.internal.compiler.problem.ProblemReporter.replaceIfSynthetic(String)` and
`org.eclipse.jdt.internal.compiler.problem.ProblemReporter.syntaxError(int, int, int, TerminalToken, char[], String, String)`

iloveeclipse:
@srikanth-sankaran : should I trigger SDK build to check whether there is any side effect on the build?

srikanth-sankaran:
> @srikanth-sankaran : should I trigger SDK build to check whether there is any side effect on the build?

Yes, please. 

We absolutely don't expect any class file changes - this PR just changes how lambda's are _parsed_ - the AST should still be identical and so _all downstream pipeline stages including the code generator should be completely oblivious to the change_. 

But given this is significant change, it helps to build the sdk as an additional validation of the parser changes - Thanks!

iloveeclipse:
Started https://ci.eclipse.org/releng/job/Builds/job/I-build-4.39/58/

iloveeclipse:
Build is there (and is "stable"): https://download.eclipse.org/eclipse/downloads/drops4/I20260114-0210/
Let see whether there will be any related test failures in JDT UI/PDE.

iloveeclipse:
> Build is there (and is "stable"): https://download.eclipse.org/eclipse/downloads/drops4/I20260114-0210/ Let see whether there will be any related test failures in JDT UI/PDE.

Update: this build has no new regressions (which would be visible via tests).

srikanth-sankaran:
> 
> Update: this build has no new regressions (which would be visible via tests).

Excellent, thanks!

Issue #4672, opened 2025-12-03
When attempting to build a type hierarchy with cyclic hierarchy, the method never returns


datho7561:
When trying to build a type hierarchy of a cyclic type hierarchy, the method never returns.

eg. with these classes

```java
class ClassA extends ClassC {}
class ClassB extends ClassA {}
class ClassC extends ClassB {}
```

Get the `IType` of `ClassA` then call `ITypeHierarchy typeHierarchy = type.newSupertypeHierarchy(monitor);`

**Expected**: The cycle is detected and an exception is thrown (or maybe the incomplete hierarchy with the cycle broken at some point is returned)

**Actual**: The method call never exits

iloveeclipse:
Is this a regression? Stack?

datho7561:
I think this is a regression, yes. We had a test that created a type hierarchy for cyclical tests in one of our projects that uses the JDT API. Based on the expected test results, it seems JDT API used to create as much as the hierarchy as it could before it hit a cycle, then return that.

iloveeclipse:
Could you bisect that?

datho7561:
I think I can do that, it should just be a matter of writing a new test case for it and running against different releases/commits.

iloveeclipse:
Great. If you would add similar regression test for JDT core, would be also good.

datho7561:
4.37 is good, working through the bisect now...

datho7561:
b4bfe4ef880662c08b98686d38a68b43a0c3132c is what broke this

iloveeclipse:
> [b4bfe4e](https://github.com/eclipse-jdt/eclipse.jdt.core/commit/b4bfe4ef880662c08b98686d38a68b43a0c3132c) is what broke this

Great! If you could add test, it would be even better.

@stephan-herrmann , @fedejeanne : could you please check?

fedejeanne:
I can confirm that b4bfe4ef880662c08b98686d38a68b43a0c3132c introduced this behavior but I don't have a fix for it yet. I'm looking into it today but I'll be out-of-office for 1 month starting next Monday _i.e._ it would be nice if @stephan-herrmann can look into it too.

fedejeanne:
I am debugging the test (#4673) and the infinite loop happens in `org.eclipse.jdt.internal.compiler.lookup.TypeBinding.findSuperTypeOriginatingFrom(int, boolean)`. The `do-while`-loop here:

[...]

PR #4673, opened 2025-12-03
Test case and fix for #4672

datho7561:
## What it does
Adds a (failing) test case for #4672

## How to test
Through CI

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/RunVariousSealedTypeTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/RunVariousSealedTypeTests.java
@@ -63,7 +63,7 @@ public class RunVariousSealedTypeTests extends TestCase {
 				FormatterRegressionTests.class,
 				CompletionTests16_1.class,
 				CompletionTests17.class,
-				Java21ElementTests.class,
+				Java25ElementTests.class,
 				JavaSearchBugs15Tests.class,
 				JavaSearchBugs17Tests.class,
 				JavaSearchBugs19Tests.class,
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/AllJavaModelTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/AllJavaModelTests.java
@@ -243,7 +243,7 @@ private static Class[] getAllTestClasses() {
 		JavaElement8Tests.class,
 
 		Java9ElementTests.class,
-		Java21ElementTests.class,
+		Java25ElementTests.class,
 
 		NullAnnotationModelTests9.class,
 
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/Java25ElementTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/Java25ElementTests.java
@@ -20,7 +20,7 @@ import org.eclipse.jdt.core.IJavaProject;
 import org.eclipse.jdt.core.JavaCore;
 import org.eclipse.jdt.core.tests.util.AbstractCompilerTest;
 
-public class Java21ElementTests extends AbstractJavaModelTests {
+public class Java25ElementTests extends AbstractJavaModelTests {
 
 	private IJavaProject project;
 
@@ -34,24 +34,24 @@ public class Java21ElementTests extends AbstractJavaModelTests {
 		this.project.setOption(JavaCore.COMPILER_SOURCE, JavaCore.VERSION_25);
 		this.project.setOption(JavaCore.COMPILER_CODEGEN_TARGET_PLATFORM, JavaCore.VERSION_25);
 	}
-	public Java21ElementTests(String name) {
+	public Java25ElementTests(String name) {
 		super(name);
 	}
 
 	public static Test suite() {
-		return buildModelTestSuite(AbstractCompilerTest.F_25, Java21ElementTests.class);
+		return buildModelTestSuite(AbstractCompilerTest.F_25, Java25ElementTests.class);
 	}
 
 	@Override
 	protected void setUp() throws Exception {
 		super.setUp();
-		this.project = createJavaProject("Java21Elements", new String[] { "src" },
-				new String[] { "JCL_21_LIB" }, "bin", "21");
+		this.project = createJavaProject("Java25Elements", new String[] { "src" },
+				new String[] { "JCL_25_LIB" }, "bin", "25");
 	}
 
 	@Override
 	protected void tearDown() throws Exception {
-		deleteProject("Java21Elements");
+		deleteProject("Java25Elements");
 		super.tearDown();
 	}
 
@@ -63,9 +63,9 @@ public class Java21ElementTests extends AbstractJavaModelTests {
 					}
 				}
 					""";
-		createFile("/Java21Elements/src/Test.java", fileContent);
+		createFile("/Java25Elements/src/Test.java", fileContent);
 
-		ICompilationUnit unit = getCompilationUnit("/Java21Elements/src/Test.java");
+		ICompilationUnit unit = getCompilationUnit("/Java25Elements/src/Test.java");
 		assertTrue(unit.getTypes()[0].getMethods()[0].isMainMethodCandidate());
 	}
 
@@ -77,9 +77,9 @@ public class Java21ElementTests extends AbstractJavaModelTests {
 					}
 				}
 					""";
-		createFile("/Java21Elements/src/Test.java", fileContent);
+		createFile("/Java25Elements/src/Test.java", fileContent);
 
-		ICompilationUnit unit = getCompilationUnit("/Java21Elements/src/Test.java");
+		ICompilationUnit unit = getCompilationUnit("/Java25Elements/src/Test.java");
 		assertTrue(unit.getTypes()[0].getMethods()[0].isMainMethodCandidate());
 	}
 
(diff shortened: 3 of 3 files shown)


org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CircularDependencyModelTests.java
```diff
@@ -0,0 +1,99 @@
...
+import org.eclipse.jdt.core.IPackageFragment;
+import org.eclipse.jdt.core.IPackageFragmentRoot;
+import org.eclipse.jdt.core.IType;
+import org.eclipse.jdt.core.ITypeHierarchy;
+import org.eclipse.jdt.core.JavaModelException;
+
+/**
+ * Model tests related to circular dependencies.
+ */
+public class CircularDependencyModelTests extends AbstractJavaModelTests {
```
fedejeanne: You could add your test into `org.eclipse.jdt.core.tests.model.TypeHierarchyTests`, no need for a new test class.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CircularDependencyModelTests.java
```diff
@@ -0,0 +1,99 @@
...
+
+	public static Test suite() {
+		return buildModelTestSuite(CircularDependencyModelTests.class);
+	}
+
+	@Override
+	protected void setUp() throws Exception {
+		super.setUp();
+		createJavaProject(PROJECT_NAME, new String[] { "src" },
+				new String[] { "JCL21_LIB" }, "bin", "25");
```
fedejeanne: Earlier Java versions should be tested too. @stephan-herrmann wasn't there a test class that ran several tests with different Java versions? I can't find it anymore.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CircularDependencyModelTests.java
```diff
@@ -0,0 +1,99 @@
...
+			try {
+				return type.newSupertypeHierarchy(new NullProgressMonitor());
+			} catch (JavaModelException e) {
+				throw new RuntimeException(e);
+			}
+		});
+
+		try {
+			// throws if the operation takes more than 5 seconds
+			hierarchyFuture.get(5, TimeUnit.SECONDS);
```
fedejeanne: No need to wait for so long.

```suggestion
			hierarchyFuture.get(200, TimeUnit.MILLISECONDS);
```

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CircularDependencyModelTests.java
stephan-herrmann: Actually this line causes error `The type java.lang.Object cannot be resolved. It is indirectly referenced from required .class files` because no classpath variable "JCL21_LIB" has been initialized.

It looks like the line was copied from Java21ElementTests, which shows the same bug, actually two bugs:
* compliance and classpath should use the same version
* initialized variables have one more underscore, like "JCL_25_LIB"

Interestingly, this error has no impact on the actual bug being witnessed by the test.

fedejeanne: Oh, thank you for the info. I didn't actually experience that error while running the test locally. 

What do you recommend to use in this case? Does the test need to explicitly test for several Java versions or is it enough to test for 1 version (which one)?

datho7561: I'll update the library and compliance to be the same version.

> initialized variables have one more underscore, like "JCL_25_LIB"

Most other places I see references to JCL*_LIB it only has one underscore

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CircularDependencyModelTests.java
datho7561: I've moved the test into this class

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CircularDependencyModelTests.java
stephan-herrmann: > Oh, thank you for the info. I didn't actually experience that error while running the test locally.

Just put a breakpoint in `org.eclipse.jdt.internal.compiler.problem.ProblemHandler.handle(int, String[], int, String[], int, int, int, ReferenceContext, CompilationResult)` (I constantly have one breakpoint right after `case ProblemSeverities.Error`), to see the error (still).

It has no obvious effect on the test, but adds unnecessary noise.

stephan-herrmann: > Most other places I see references to JCL*_LIB it only has one underscore

"JCL21_LIB"  has exactly one precedent in Java21ElementTests, which is bogus, too. I get the same result actually when searching for "JCL2?_LIB".

"JCL_21_LIB" has 8 matches, "JCL_2?_LIB" gives 43 matches.

Aside from statistics, the relevant strings can be found in `org.eclipse.jdt.core.tests.model.AbstractJavaModelTests.setUpJCLClasspathVariables(String, boolean)`.

stephan-herrmann: > What do you recommend to use in this case? Does the test need to explicitly test for several Java versions or is it enough to test for 1 version (which one)?

I don't see significant impact of the Java version. Testing at one version should be enough.

datho7561: Okay, I'll address this for the test case I added.

Do you think it's important to fix the other occurances (in a separate PR?)

datho7561: Oh I see, the pre-Java 17 variables have one underscore and after that they have two.

stephan-herrmann:
from https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4672#issuecomment-3612495969 : 

> Some options:
> 
> 1. Could `LE.askForType()` know about the "purpose" of why a type is being requested?      
>     * If that "purpose" is supertype lookup, we should not stash and restore
>     * All other purposes continue to use stash and restore

This option looked simplest to me. 

By integrating the new flag also in GlobalStateMemento we support arbitrary interleaving between two modes, supertype lookup, other type lookup:
* while ClassScope.connectHierarchy is active do not clear `typesBeingConnected`, but still use stash / restore for the sake of the new flag.
* yet, in the middle of supertype lookup, any resolving during `accept()` should not blindly assume we're still in hierarchy lookup, achieved by temporarily _resetting_ the new flag.

I put the new flag into the ROOT_ONLY discipline, meaning: we don't maintain separate flags per `module`.

Perhaps even `typesBeingConnected` should always be referenced via `root`, but then a cycle across module boundaries would first require an illegal cycle among modules.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CircularDependencyModelTests.java
stephan-herrmann: > Okay, I'll address this for the test case I added.
> 
> Do you think it's important to fix the other occurances (in a separate PR?)

Not a big deal, so I would actually appreciate this little cleanup even as part of the current PR.
I'll make a small note about it in the final commit comment when merging.

As we are in collaboration mode here, please make any changes now as separate commits, rather than history re-writing, thanks.

stephan-herrmann: > Oh I see, the pre-Java 17 variables have one underscore and after that they have two.

ups, "nice" way to confuse ourselves... at least it's consistently inconsistent ;-P

datho7561:
Looks like that fix worked; I'll push the change to the variable in the test now.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CircularDependencyModelTests.java
stephan-herrmann: > > Okay, I'll address this for the test case I added.
> > Do you think it's important to fix the other occurances (in a separate PR?)
> 
> Not a big deal, so I would actually appreciate this little cleanup even as part of the current PR. I'll make a small note about it in the final commit comment when merging.
> 
> As we are in collaboration mode here, please make any changes now as separate commits, rather than history re-writing, thanks.

You didn't see these, @datho7561 ?

datho7561: oops no my bad

datho7561: I've added the fix for Java21ElementTests, sorry about rewriting history on the PR earlier.

stephan-herrmann: > I've added the fix for Java21ElementTests, 

thanks

> sorry about rewriting history on the PR earlier.

It's OK, people have different habits. 

To explain my request: I personally prefer to see all incremental changes as a fix evolves. It helps my style of reviewing. And unfortunately, github is inferior (compared to gerrit, gitlab) in showing these differences when force-push is involved.

stephan-herrmann: Unfortunately there's more confusion regarding Java21ElementTests: 
* Apparently it was introduced when Java 21 introduced instance main methods as a preview.
* Meanwhile that feature has matured (no more preview) but it requires source level 25

@datho7561  Do you mind further cleaning up in this direction?
* rename the class to Java25ElementTests
   * check if it is still included at the correct location in the suite 
* configure it to use "JCL_25_LIB" and "25"

stephan-herrmann:
Thanks @datho7561  & @fedejeanne  for great collaboration!

Issue #4609, opened 2025-11-10
Compiler fails to generate tagElements for Markdown fences


SougandhS:
As observed in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4418#issuecomment-3498160191 and suggested in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4418#issuecomment-3513635375

Consider given exmaple

```
/// The following is some sample code which illustrates source formatting within markdown comments:
/// ``` public class Example { final int a = 1; final boolean b = true;} ```
/// Descriptions of parameters and return values are best appended at end of the markdown comment.
/// @param first  The first parameter. For an optimum result, this should be an odd number between 0 and 100.
/// @param second The second parameter.
/// @throws Exception when the foo operation cannot be performed for one reason or another.
/// @return The result of the foo operation, usually an even number within 0 and 1000.
class Mark62 {
}
```

org.eclipse.jdt.core.dom.Javadoc.tags() will generate

<img width="684" height="160" alt="Image" src="https://github.com/user-attachments/assets/21cb4bee-8e1c-48f9-b217-d3da49dcaa68" />

entire markdown comment will be treated as one single tagElement

but now if we remove the snippet/fences (```) in the above markdown

```
/// The following is some sample code which illustrates source formatting within markdown comments:
/// @param first  The first parameter. For an optimum result, this should be an odd number between 0 and 100.
/// @param second The second parameter.
/// @throws Exception when the foo operation cannot be performed for one reason or another.
/// @return The result of the foo operation, usually an even number within 0 and 1000.
class Mark62 {
}
```
org.eclipse.jdt.core.dom.Javadoc.tags() will generate proper tagElements

<img width="657" height="272" alt="Image" src="https://github.com/user-attachments/assets/faaccd4f-c873-47fb-9228-644fda96f46d" />

subyssurendran666:
<img width="2852" height="634" alt="Image" src="https://github.com/user-attachments/assets/b45c533b-3a80-4a5e-9ca4-f79ddb97a146" />
There are also issues with the use of triple backticks. In the above sample code, the `@param`, `@throws`, and `@return` tags should be parsed as TagElements, but they are currently being recognised as TextElements.

subyssurendran666:
<img width="2402" height="594" alt="Image" src="https://github.com/user-attachments/assets/2bfa1eb6-6323-4e97-9f4c-81cc56c5b239" />

[...]

PR #4678, opened 2025-12-04
Compiler fails to generate tagElements for Markdown fences

subyssurendran666:
Markdown parser fails to generate TagElements after single-line triple backtick fence

Fix: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4609

## How to test
```
/// The following is some sample code which illustrates source formatting within markdown comments:
/// ``` public class Example { final int a = 1; final boolean b = true;} ```
/// Descriptions of parameters and return values are best appended at end of the markdown comment.
/// @param first  The first parameter. For an optimum result, this should be an odd number between 0 and 100.
/// @param second The second parameter.
/// @throws Exception when the foo operation cannot be performed for one reason or another.
/// @return The result of the foo operation, usually an even number within 0 and 1000.
public class Markdown{}
```

Hover over the Markdown portion of the example and examine the AST view to verify that TagElements are properly generated once the triple backtick fence has been parsed.
## Author checklist


subyssurendran666:
image

image


Consider the two images above (both showing faulty single and triple backticks). The remaining TagElements should not be treated as part of the backtick elements. This behavior should be fixed along with this issue.

PR #4681, opened 2025-12-07
Ensure URB's are resolved before performing visibility checks

stephan-herrmann:
+ try to avoid root cause of URB surfacing in unexpected locations

Alternative fix for
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4632

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/FunctionalExpression.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/FunctionalExpression.java
@@ -239,8 +239,6 @@ public abstract class FunctionalExpression extends Expression {
 		}
 
 		private void checkVisibility(ReferenceBinding referenceBinding) {
-			if (referenceBinding.isUnresolvedType())
-				referenceBinding = (ReferenceBinding) BinaryTypeBinding.resolveType(referenceBinding, this.scope.environment(), false);
 			if (!referenceBinding.canBeSeenBy(this.scope)) {
 				this.visible = false;
 				if (this.shouldChatter)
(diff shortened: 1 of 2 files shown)


stephan-herrmann:
From https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4647#issuecomment-3561955872
> @stephan-herrmann - since you are fascinated by bugs that have managed to hide for a long time, you may want to study this one 😆

You know me well :)

Sorry, @srikanth-sankaran,  for distracting you from more interesting task for just a few minutes:

Asking myself how this bug had managed to hide for such a long time, my first question was: How do we know which access to any TypeBinding needs to check for URB? Certainly we don't want to do this EVERYWHERE :smile:

Seeing a PTB involved, I came up with the theory, that every access to details of this type using any of the provided access methods should be safe.

In the test case introduced via #4647 the offending URB was found in the following location:
* parameter of sam
   * its enclosing type
      * its type argument
 
This led me to method PTB.resolve() which seems to be responsible for ensuring that a PTB is fully resolved, right?
* This method diligently descends into type arguments to resolve any URBs left in any of them
* But a PTB depends on one more thing: its enclosing type

Et voilà: in the new test we can observe exactly this: a PTB failing to resolve its enclosing type, and adding just this little detail seems to cover for the original problem from #4632 

Q:   Do you agree that fixing the issue in PTB is preferable over the previous fix?

Note: if we agree on PTB as the actual bug location, then the bug is much older than even lambdas :sunglasses:

srikanth-sankaran:
I'll formally review this Tuesday but accept my congratulations. Your approach is certainly more methodical than mine and so is a much more appropriate fix. Thanks for doing this.

stephan-herrmann:
> This fix looks better than my original attempt. Thanks!

thank you.

Issue #4704, opened 2025-12-21
Refactor rename doesnt work with lambda subtype in different source 


enrdenti:
In Eclipse 2025-12 (Java+DSL edition), using the Refactor -> Rename command (or SHIFT+ALT+R) in a Java source file leads to an inconsistent behaviour: instead of applying the required rename, the older name is maintained after hitting ENTER, as if something failed behind the scenes. This is quite a problem, since renaming something is one of the most common refactoring actions.

Sometimes, applying the same command from another file works: I noticed that trying to do the rename in the class where the method is _defined_ leads the command to fail, but doing the same from _another_ class, where the method is _used_, tends to work.

This bug does not exist in Eclipse 2025-09, so this is a regression issue.

trancexpress:
Please provide a minimal reproducer.

I downloaded: https://www.eclipse.org/downloads/packages/release/2025-12/r/eclipse-ide-java-and-dsl-developers

```
Eclipse DSL Tools (includes Incubating components)
Version: 2025-12 (4.38.0)
Build id: 20251204-0850
```

Created a new Java project in a new workspace and did a refactor rename in a minimal snippet. No issue.

enrdenti:
Hello, in order to reproduce the issue, a specific situation needs to be set up.

1) First, other than the standard 'src' source folder, another 'test' source folder should be defined, where test cases will be put.

2) For each package in the source folder, a corresponding test package is defined in the 'test' source folder: the naming convention is that for source package "a.b.c", the corresponding test package be named 'a.b.c.test' - see attached image

3) Then, if you try to rename a method in a class, say the Campo class, sometimes it works, sometimes it doesn't, depending on the method you want to rename: for instance, renaming cella() in "casella()" and vice-versa tends not to work (but sometimes it does, especially if you rename something else in between), while renaming other methods mostly work. 
I also tried removing the corresponding CampoTest class from the test folder and, surprise!, that seems to fix the issue, at least at a first glance. Another way to partially fix the issues seems is reducing the class complexity, removing some methods from the class - but obviously, this is a nonsense. So, the feeling is that the renaming command gets confused by something, in some situations, that make it fail underneath.

Needless to say, in Eclipse 2025-09 everything works fine in each situation. Does this help?

<img width="219" height="202" alt="Image" src="https://github.com/user-attachments/assets/434841d3-446c-426b-b334-1c49f0cd007b" />

trancexpress:
[...]

PR #4708, opened 2025-12-22
Add tests for refactor rename with subtype in lambda

trancexpress:
See: #4704


trancexpress:
The test passes. Though note that if the subtype implements the renamed method, in Eclipse at least the rename doesn't work - if initiated from the subtype in the lamda. Likely we need a bug and test for that case too...

Issue #4715, opened 2025-12-26
[Eclipse 2025-12 regression]: Type inference issue in Collectors.toMap method within Collectors.partitioningBy method


dinoophari:
Following code is not compiling with latest Eclipse 2025-12 (4.38.0), but is perfectly fine in previous releases.
Build id: 20251204-0850

I am using **JDK 21** (IBM Semeru JDK) for the below code.

```java
import java.util.Map;
import java.util.Map.Entry;
import java.util.Set;
import java.util.stream.Collectors;

public class TestMain {
	public static void main(String[] args) {
		final Map<String, Set<String>> pathToPIDMap = Map.ofEntries(
				Map.entry("i1", Set.of("PID1")),
				Map.entry("i1,i2", Set.of("PID2"))
				);
		
		final Map<Boolean, Map<String, Set<String>>> partitiionedPathToPIDSet = pathToPIDMap.entrySet().stream() // Error 3
			.collect(Collectors.partitioningBy(entry -> {
				final Set<String> pidSet = entry.getValue();
				final boolean isSinglePIDSet = pidSet.size() == 1;
				return isSinglePIDSet;
			}, Collectors.mapping(entry -> {
				final String path = entry.getKey();
				final Set<String> pidSet = entry.getValue();
				
				return Map.entry(path, pidSet);
			}, Collectors.toMap(Entry::getKey, Entry::getValue)))); //Errors 1 and 2
	}
}
```

Eclipse complaining 3 errors here:
1. The type `Map.Entry` does not define `getKey(Object)` that is applicable here
2. The type `Map.Entry` does not define `getValue(Object)` that is applicable here
3. Type mismatch: cannot convert from `Map<Boolean,Map<Object,Object>>` to `Map<Boolean,Map<String,Set<String>>>`

It seems the Eclipse compiler is not able to infer the types properly. Compiling with `javac` is working.

PFA a zip file containing the Eclipse project.

[Eclipse_2025-12_testPrj.zip](https://github.com/user-attachments/files/24346230/Eclipse_2025-12_testPrj.zip)

Issue #4635, opened 2025-11-14
Type mismatch with generic wildcard


tiller:
With Eclipse 4.37 & 4.38M2 nightly, the following code:

```
public class Test {
    static class Obj<T> {
    }

    public static void main(String[] args) {
        String arg = "";

        Stream<Obj<?>> stream11 = Stream.of(newObj());
        Stream<Obj<?>> stream21 = Stream.of(newObj2(arg));

        Stream<Obj<?>> stream12 = Stream.of(newObj(), newObj());
        Stream<Obj<?>> stream22 = Stream.of(newObj2(arg), newObj2(arg));

        //Stream<Obj<?>> stream11f = Stream.of(newObj()).filter(Objects::nonNull); // javac & eclipse KO
        //Stream<Obj<?>> stream21f = Stream.of(newObj2(arg)).filter(Objects::nonNull); // javac & eclipse KO

        Stream<Obj<?>> stream12f = Stream.of(newObj(), newObj()).filter(Objects::nonNull);
        Stream<Obj<?>> stream22f = Stream.of(newObj2(arg), newObj2(arg)).filter(Objects::nonNull); // javac OK, eclipse KO
    }

    public static Obj<?> newObj() {
        return new Obj<>();
    }

    public static <T> Obj<?> newObj2(T arg) {
        return new Obj<>();
    }
}
```

Gives an inconsistent result for `stream22f` where it works on javac but not on eclipse :

Type mismatch: cannot convert from Stream<Test.Obj<Object>> to Stream<Test.Obj<?>>	Test.java	/test/src/main/java/test	line 25	Java Problem

(that's my last issue, thanks guys for your work)

srikanth-sankaran:
@stephan-herrmann - this one seems to be up your alley.

stephan-herrmann:
> Type mismatch: cannot convert from Stream<Test.Obj> to Stream<Test.Obj<?>> Test.java /test/src/main/java/test line 25 Java Problem

I was about to complain about the raw type `Test.Obj` but then found the error message to be
`Type mismatch: cannot convert from Stream<Test.Obj<Object>> to Stream<Test.Obj<?>>`

stephan-herrmann:
Thanks, this report finally made me revisit some code design I made in 2014, which on second look is totally bogus.

[...]

PR #4716, opened 2025-12-28
[Eclipse 2025-12 regression]: Type inference issue in Collectors.toMap method within Collectors.partitioningBy method

stephan-herrmann:
Behavior change is due to the fix #4520.

Fixed by heuristically including / excluding capture bounds during copy

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4715


stephan-herrmann:
This touches an area that may require a significant revamp as of #4635 (which involves fixing a blatant bug in ecj while at the same time compensating for insufficient specification).

For the current bug it would suffice to simply remove one line from #4520, but (incomplete) experiments in #4635 indicate that selectively including captured bounds on some call paths while excluding them in other situations gives better results.

PR #4719, opened 2026-01-02
Record type inference fails when 'var' and <> mixed

### LINKED ISSUE #1361: [21] Record type inference fails when 'var' and <> mixed
The following code is rejected by the ECJ:

```
record Record<T>(T a) {}
public class X {
	static void test(Object m) {
		if (m instanceof Record(var x)) {
			var wrapped = new Record<>(x);
			Record<Object> r = wrapped; // Type mismatch: cannot convert from Record<capture#2-of ?> to Record<Object>
		}
	}
}
```
Of course, if I use explicit types in either of the 'var', the code compiles. Javac is able to infer the type correctly.

mpalat:
Fixes #1361

## What it does
- The type inference in record pattern was more restrictive in implementation compared to what was mentioned in spec.

In [JLS Section 18.5.5](https://docs.oracle.com/javase/specs/jls/se25/html/jls-18.html#jls-18.5.5), item 5 says "Unlike normal resolution, in this case resolution skips the step that attempts to produce an instantiation for an inference variable from its proper lower bounds or proper upper bounds; instead, any new instantiations are created by skipping directly to the step that introduces fresh type variables." 

However, the entire processing was missed via the [!recordTypeInference](https://github.com/eclipse-jdt/eclipse.jdt.core/blob/df7f21e771a9130efeb7c64eb6af5088a122674a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java#L1152C21-L1152C51) condition which implied that the " α1 = T1, ..., αn = Tn" were consequently missed being incorporated with the current bound set as per [this line in JLS Section 18.4 ](https://docs.oracle.com/javase/specs/jls/se25/html/jls-18.html#jls-18.4:~:text=The%20bounds%20%CE%B11%20%3D%20T1%2C%20...%2C%20%CE%B1n%20%3D%20Tn%20are%20incorporated%20with%20the%20current%20bound%20set.)

The fix addresses this part.

Fixes issue #1361

## How to test
The reviewer can use the code mentioned in issue #1361 to test the fix.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
@@ -1149,48 +1149,52 @@ public class InferenceContext18 {
 				final int numVars = variableSet.size();
 				if (numVars > 0) {
 					final InferenceVariable[] variables = variableSet.toArray(new InferenceVariable[numVars]);
-					variables: if (!isRecordPatternTypeInference && !tmpBoundSet.hasCaptureBound(variableSet)) {
+					variables: if (!tmpBoundSet.hasCaptureBound(variableSet)) {
 						// try to instantiate this set of variables in a fresh copy of the bound set:
 						BoundSet prevBoundSet = tmpBoundSet;
 						tmpBoundSet = tmpBoundSet.copy();
 						for (int j = 0; j < variables.length; j++) {
 							InferenceVariable variable = variables[j];
-							// try lower bounds:
-							TypeBinding[] lowerBounds = tmpBoundSet.lowerBounds(variable, true/*onlyProper*/);
-							if (lowerBounds != Binding.NO_TYPES) {
-								TypeBinding lub = this.scope.lowerUpperBound(lowerBounds);
-								if (lub == TypeBinding.VOID || lub == null)
-									return null;
-								tmpBoundSet.addBound(new TypeBound(variable, lub, ReductionResult.SAME), this.environment);
+							if (isRecordPatternTypeInference) {
+								tmpBoundSet.addBound(new TypeBound(variable, this.object, ReductionResult.SAME), this.environment);
 							} else {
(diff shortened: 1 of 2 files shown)


mpalat:
@stephan-herrmann : requesting for a review both for the reasoning and the implementation (thanks!) - one clarification here (assuming that you find the logic above correct), is whether a similar skip is required at the Capture conversion part in the [second attempt](https://docs.oracle.com/javase/specs/jls/se25/html/jls-18.html#jls-18.4:~:text=If%20the%20bound%20set%20contains%20a%20bound%20of%20the%20form%20G%3C...%2C%20%CE%B1i%2C%20...%3E%20%3D%20capture(G%3C...%3E)%20for%20some%20i%20(1%20%E2%89%A4%20i%20%E2%89%A4%20n)).

stephan-herrmann:
I would need some help to follow the reasoning.

> "Unlike normal resolution, in this case resolution skips the step that attempts to produce an instantiation for an inference variable from its proper lower bounds or proper upper bounds; instead, any new instantiations are created by skipping directly to the step that introduces fresh type variables"

At a quick look this seems to correspond directly to the block being skipped based on `recordTypeInference`. Inside that block each inference variable is treated by one of three branches for (a) proper lower bounds, (b) inThrows, (c) proper upper bounds. Each branch has its specific definition for `Ti`, which is then used in 
>  The bounds α1 = T1, ..., αn = Tn are incorporated with the current bound set. 

Hence we directly skip to the "second attempt" for which no skipping is defined in the spec. All this looks fine to me in master.

Am I missing anything?

mpalat:
@stephan-herrmann just  letting you know that I am still doing a relook at the spec, reading up some theory and trying to "decrypt" it again to check the issue -

mpalat:
@stephan-herrmann - right. we need to get to the fresh capture variables.  So am closing this PR. 

This seems to be a bug with javac itself.

stephan-herrmann:
> This seems to be a bug with javac itself.

To close the loop also here: can you provide a program where javac accepts the mix of `var` and `<>`?

Issue #1600, opened 2023-11-14
Problem with calling super in overridden default method


Frettman:
This problem is probably easiest to describe using the following standalone code. It's cut down as much as possible from a real world example. Basically I'm trying to extend a class and override a method whose implementation has been provided as default method in an interface somewhere in the hierarchy. Trying to call the overridden method via `super` fails with `Cannot directly invoke the abstract method setReadOnly(boolean) for the type HasValue`.

The same code compiles fine with OpenJDK 17.

```java
public class TestField
  extends
    CustomField
{
  @Override
  public void setReadOnly(boolean readOnly)
  {
    // Eclipse shows this compile error: Cannot directly invoke the abstract method setReadOnly(boolean) for the type HasValue
    super.setReadOnly(readOnly);
  }
}

abstract class CustomField
  extends
    AbstractField
  implements
    HasValue
{
}

abstract class AbstractField
  implements
    HasValueAndElement
{
}

interface HasValue
{
  void setReadOnly(boolean readOnly);
}

interface HasValueAndElement
  extends
    HasValue
{
  @Override
  default void setReadOnly(boolean readOnly)
  {
  }
}
```

iloveeclipse:
Is this a recent regression? In which version do you see this bug?

Frettman:
This happened after upgrading the version of a framework. I'm using a fairly recent Eclipse version (4.29), but I just tried and it's the same on 2021-09 (4.21.0). So it's not a recent regression as far as JDT is concerned.

iloveeclipse:
OK, thanks. I see it also still on latest 4.30 master.
Here the same example just reformatted for clarity

```
public class TestField extends CustomField {
	@Override
	public void setReadOnly(boolean readOnly) {
		// Eclipse shows this compile error: Cannot directly invoke the abstract
		// method setReadOnly(boolean) for the type HasValue
		super.setReadOnly(readOnly);
	}
}

abstract class CustomField extends AbstractField implements HasValue {}

abstract class AbstractField implements HasValueAndElement {}

interface HasValue {
	void setReadOnly(boolean readOnly);
}

interface HasValueAndElement extends HasValue {
	@Override
	default void setReadOnly(boolean readOnly) {
	}
}
```

roeltje25:
I have this issue as well. in 2025-06.  In exactly the same framework (Vaadin) and the same code. with javac I can compile it just without issues, in IntelliJ: also no issues. only in eclipse...

[...]

PR #4727, opened 2026-01-07
Problem with calling super in overridden default method

stephan-herrmann:
among equivalent methods prefer concrete over default over abstract

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1600

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
@@ -4673,6 +4673,13 @@ public abstract class Scope {
 		if (receiverType != null)
 			receiverType = receiverType instanceof CaptureBinding ? receiverType : (ReferenceBinding) receiverType.erasure();
 
+		// within the boundaries of "chosen arbitrarily among the subset of the maximally specific methods that are preferred"
+		// put concrete methods first, default methods second:
+		Arrays.sort(moreSpecific, (m1, m2) -> {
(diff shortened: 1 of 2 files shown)


stephan-herrmann:
I'm having some issues mapping the last long section in `Scope.mostSpecificMethodBinding()` to [JLS 15.12.2.5](https://docs.oracle.com/javase/specs/jls/se25/html/jls-15.html#jls-15.12.2.5). This code section basically still has the structure from 2006 / 1b0238cbdda2702f17c4f0b30e4c8493818f8888 . Probably one day we should rewrite this in a style closer to current JLS.

I observed, however,  that a `return current;` may happen even before all methods have been touched by the outer loop (over `i`). Since the order of methods in `moreSpecific` seemed random to me, I added some opportunistic sorting, so we prefer concrete methods over default methods over abstract methods.

jarthana:
The following code, that we were rejecting before is now accepted after this change:

```
interface SuperIntf {
	public default int foo() {
		return -1;
	}
}
interface TestIntf extends SuperIntf {
	public int foo();
}
abstract class SuperClass implements TestIntf {
}
class TestedClass extends SuperClass {
	public int foo() {
		return 0;
	}
	class InnerClass {
		int foo() {
			return 1;
		}
		int bar() {
			return TestedClass.super.foo();
		}
	}
}
```

@stephan-herrmann Please let me know if you prefer a separate issue.

srikanth-sankaran:
> Please let me know if you prefer a separate issue.

Note: #1600 is reopened so perhaps that could include the new scenario you cite

srikanth-sankaran:
> The following code, that we were rejecting before is now accepted after this change:

Thanks @jarthana - I have added this as a regression test to https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4775

PR #4682, opened 2025-12-07
Trouble annotating explicit receiver with ECJ (works fine with javac)

stephan-herrmann:
+ band aid for type bindings with unexpectedly different ids.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4494

Issue #4494, opened 2025-10-10
Trouble annotating explicit receiver with ECJ (works fine with javac)


lukehutch:
Example:

```java
class TypeAnnotationTest {
    @Target({ ElementType.FIELD, ElementType.TYPE_USE, ElementType.TYPE_PARAMETER })
    private static @interface A {
    }

    @Target({ ElementType.TYPE_USE, ElementType.TYPE_PARAMETER })
    private static @interface B {
    }

    @Target({ ElementType.TYPE_USE, ElementType.TYPE_PARAMETER })
    private static @interface C {
    }

    @Target({ ElementType.TYPE_USE, ElementType.TYPE_PARAMETER })
    private static @interface F {
    }

    static class U {
    }

    interface V {
    }

    static class P<@A T extends @B U & @C V> {
        public void explicitReceiver(@F P<T> this) {
        }
    }
}
```

`@F P` is underlined in red, with the error message:

```
The declared type of the explicit 'this' parameter is expected to be TypeAnnotationTest.P<T>
```

(This works fine, with no error, with the JDK)

Interestingly, this gives no error (removing `@A` from `T`):

```
    static class P<T extends @B U & @C V> {
        public void explicitReceiver(@F P<T> this) {
        }
    }
```

and this also works fine:

```
    static class P {
        public void explicitReceiver(@F P this) {
        }
    }
```

Eclipse version:
* Version: 2025-06 (4.36.0)
* Build id: 20250605-1316

srikanth-sankaran:
Reproduced, investigating ...

srikanth-sankaran:
Please don't elide imports and such while reporting issues in future. TIA.

Full test case:

```
import java.lang.annotation.ElementType;
import java.lang.annotation.Target;

class TypeAnnotationTest {
    @Target({ ElementType.FIELD, ElementType.TYPE_USE, ElementType.TYPE_PARAMETER })
    private static @interface A {
    }

    @Target({ ElementType.TYPE_USE, ElementType.TYPE_PARAMETER })
    private static @interface B {
    }

    @Target({ ElementType.TYPE_USE, ElementType.TYPE_PARAMETER })
    private static @interface C {
    }

    @Target({ ElementType.TYPE_USE, ElementType.TYPE_PARAMETER })
    private static @interface F {
    }

    static class U {
    }

    interface V {
    }

    static class P<@A T extends @B U & @C V> {
        public void explicitReceiver(@F P<T> this) {
[...]

Issue #3623, opened 2025-01-27
NPE in compiler when using external NP annotations for Set.of(E...)


sithmein:
The following call to the varargs variant of `java.util.Set.of()` leads to an internal compiler error:
```
import java.util.Set;

public class Test {
	public static void main(String[] args) {
		var b = Set.of("1", "2", "3", "4", "5", "6", "7", "8", "9", "10", "11");
	}
}
```
With up to 10 arguments one of the dedicated `Set.of()` methods is being used which don't cause a problem.
![Image](https://github.com/user-attachments/assets/5c51259e-71ad-4f00-8238-a6d44012828a)

iloveeclipse:
Can't reproduce on I20250126-1800.
Please check with https://download.eclipse.org/eclipse/downloads/drops4/I20250126-1800/

iloveeclipse:
@sithmein : please simply attach zipped project with the .settings, I guess some option breaks compiler.

sithmein:
The problem seems to be caused by external nullness annotations. Attached are two projects, one with a single external annotation (the one for `Set.of(E...)` and a project that makes use of the external annotation.

[compiler-bug-external-null-annotations.zip](https://github.com/user-attachments/files/18559225/compiler-bug-external-null-annotations.zip)

stephan-herrmann:
Two findings so far:
* the 11-args invocation is affected because that's where `Set` starts to use varargs rather than an overload for the exact number of explicit arguments
* the problem occurs when external annotations are configured, but annotation based null analysis itself is disabled

@sithmein can you confirm that the problem disappears if you do either of the following?
* enable annotation based null analysis in project Test (& configure the annotation library)
* or: remove the `annotationpath` configuration in `.classpath`

PR #4740, opened 2026-01-13
Alternate management of TypeParameters with declaration site annotations

srikanth-sankaran:
- Substantially revert PR
https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3724 retaining test
and infrastructure changes and some valid peripheral cleanups
- Substantially revert PR
https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2173 retaining
tests
- Substantially revert commit 37b18fe3dec30aacdbf97606c4be5a6236865064
made for https://bugs.eclipse.org/456487 retaining tests
- Substantially revert commit fa4debd5be46821e4afe7fa2082f7a976e89007c
made for https://bugs.eclipse.org/440143
- Borrow test from PR
https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4682/ skipping code
changes
- Additional test to make sure that PR
https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4554 does not
inadvertently mask problems (See
https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4740#issuecomment-3746960540)
- Removes workaround from previous bug fixes for https://bugs.eclipse.org/418041
- Removes workaround from previous bug fixes for  https://bugs.eclipse.org/429958
- Removes workaround from previous bug fixes for https://bugs.eclipse.org/501031

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/apt/dispatch/AnnotationDiscoveryVisitor.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/apt/dispatch/AnnotationDiscoveryVisitor.java
@@ -155,7 +155,7 @@ public class AnnotationDiscoveryVisitor extends ASTVisitor {
 			// when we get here, it is guaranteed that class type parameters are connected, but method type parameters may not be.
 			MethodBinding methodBinding = (MethodBinding) binding.declaringElement;
 			((SourceTypeBinding) methodBinding.declaringClass).resolveTypesFor(methodBinding);
-			this.resolveAnnotations(scope, annotations, binding);
+			this.resolveAnnotations(scope, annotations, typeParameter.binding /* treat as volatile as Scope.preprocessTypeVariables() called by resolveTypesFor() may modify!!! */);
 		}
 		return false;
 	}
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeParameter.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeParameter.java
@@ -135,6 +135,10 @@ public class TypeParameter extends AbstractVariableDeclaration {
 				this.binding.setTypeAnnotations(annotationBindings, isAnnotationBasedNullAnalysisEnabled);
 				scope.referenceCompilationUnit().compilationResult.hasAnnotations = true;
 			}
+			if (this.binding.getTypeAnnotations() == Binding.AWAITED_ANNOTATIONS) { // In <T extends @Nullable Number>; T itself doesn't carry annotations, some bound does
+				this.binding.setTypeAnnotations(Binding.NO_ANNOTATIONS, isAnnotationBasedNullAnalysisEnabled); // not awaited anymore
+				scope.referenceCompilationUnit().compilationResult.hasAnnotations = true;
+			}
 			if (isAnnotationBasedNullAnalysisEnabled) {
 				if (this.binding != null && this.binding.isValidBinding()) {
 					if (scope.hasDefaultNullnessFor(Binding.DefaultLocationTypeParameter, this.sourceStart())) {
@@ -142,22 +146,12 @@ public class TypeParameter extends AbstractVariableDeclaration {
 							if ((this.binding.tagBits & TagBits.AnnotationNonNull) != 0)
 								scope.problemReporter().nullAnnotationIsRedundant(this);
 						} else { // no explicit type annos, add the default:
-							TypeVariableBinding previousBinding = this.binding;
-							this.binding = (TypeVariableBinding) environment.createNonNullAnnotatedType(this.binding);
-
-							if (scope instanceof MethodScope) {
-								/*
-								 * for method type parameters, references to the bindings have already been copied into
-								 * MethodBinding.typeVariables - update them.
-								 */
-								MethodScope methodScope = (MethodScope) scope;
-								if (methodScope.referenceContext instanceof AbstractMethodDeclaration) {
-									MethodBinding methodBinding = ((AbstractMethodDeclaration) methodScope.referenceContext).binding;
-									if (methodBinding != null) {
-										methodBinding.updateTypeVariableBinding(previousBinding, this.binding);
-									}
-								}
-							}
+							annotationBindings = this.binding.getTypeAnnotations();
+							AnnotationBinding [] newAnnotations = new AnnotationBinding[annotationBindings == null ? 1 : annotationBindings.length + 1];
+							if (annotationBindings != null)
+								System.arraycopy(annotationBindings, 0, newAnnotations, 0, annotationBindings.length);
+							newAnnotations[newAnnotations.length - 1] = environment.getNonNullAnnotation();
+							this.binding.setTypeAnnotations(newAnnotations, true);
 						}
 					}
 					this.binding.evaluateNullAnnotations(scope, this);
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/SourceTypeBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/SourceTypeBinding.java
@@ -2018,7 +2018,10 @@ private MethodBinding resolveTypesWithSuspendedTempErrorHandlingPolicy(MethodBin
 
 	final long sourceLevel = this.scope.compilerOptions().sourceLevel;
 	ReferenceBinding object = this.scope.getJavaLangObject();
+	AbstractMethodDeclaration methodDecl = method.sourceMethod();
+	TypeParameter[] typeParameters = methodDecl != null ? methodDecl.typeParameters() : null;
 	TypeVariableBinding[] tvb = method.typeVariables;
+	this.scope.preprocessTypeVariables(tvb, typeParameters);
 	for (int i = 0; i < tvb.length; i++)
 		tvb[i].superclass = object;		// avoid null (see https://bugs.eclipse.org/426048)
 
@@ -2027,11 +2030,11 @@ private MethodBinding resolveTypesWithSuspendedTempErrorHandlingPolicy(MethodBin
 	if (hasRestrictedAccess())
 		method.modifiers |= ExtraCompilerModifiers.AccRestrictedAccess;
 
-	AbstractMethodDeclaration methodDecl = method.sourceMethod();
+
 	if (methodDecl == null) return null; // method could not be resolved in previous iteration
 
 
-	TypeParameter[] typeParameters = methodDecl.typeParameters();
+
 	if (typeParameters != null) {
 		methodDecl.scope.connectTypeVariables(typeParameters, true);
 		// Perform deferred bound checks for type variables (only done after type variable hierarchy is connected)
@@ -2364,24 +2367,6 @@ public void evaluateNullAnnotations() {
 		this.scope.problemReporter().missingNonNullByDefaultAnnotation(this.scope.referenceContext);
 		if (!isInDefaultPkg)
 			pkg.setDefaultNullness(NULL_UNSPECIFIED_BY_DEFAULT);
-	}
-	maybeMarkTypeParametersNonNull();
-}
-
-private void maybeMarkTypeParametersNonNull() {
-	if (this.typeVariables != null && this.typeVariables.length > 0) {
-		// when creating type variables we didn't yet have the defaultNullness, fill it in now:
-		if (this.scope == null || !this.scope.hasDefaultNullnessFor(DefaultLocationTypeParameter, this.sourceStart()))
-			return;
-		AnnotationBinding[] annots = new AnnotationBinding[]{ this.environment.getNonNullAnnotation() };
-		for (int i = 0; i < this.typeVariables.length; i++) {
-			TypeVariableBinding tvb = this.typeVariables[i];
-			TypeParameter typeParameter = this.scope.referenceContext.typeParameters[i];
-			if (typeParameter.annotations != null && (tvb.extendedTagBits & ExtendedTagBits.AnnotationResolved) == 0)
-				continue; // not yet ready
-			if ((tvb.tagBits & TagBits.AnnotationNullMASK) == 0)
-				this.typeVariables[i] = (TypeVariableBinding) this.environment.createAnnotatedType(tvb, annots);
-		}
 	}
 }
 
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeVariableBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeVariableBinding.java
@@ -871,11 +871,6 @@ public class TypeVariableBinding extends ReferenceBinding {
 		return this;
 	}
 
-	@Override
-	public void setTypeAnnotations(AnnotationBinding[] annotations, boolean evalNullAnnotations) {
-		this.environment.getUnannotatedType(this); // exposes original TVB/capture to type system for id stamping purposes.
-		super.setTypeAnnotations(annotations, evalNullAnnotations);
-	}
 	/**
      * @see org.eclipse.jdt.internal.compiler.lookup.ReferenceBinding#shortReadableName()
      */
@@ -1085,6 +1080,8 @@ public class TypeVariableBinding extends ReferenceBinding {
 			for (int i = 0, length = annotatedTypes == null ? 0 : annotatedTypes.length; i < length; i++) {
 				TypeVariableBinding annotatedType = (TypeVariableBinding) annotatedTypes[i];
 				annotatedType.firstBound = firstBound;
+				if (firstBound != null)
+					annotatedType.tagBits |= firstBound.tagBits & (TagBits.HasNullTypeAnnotation|TagBits.HasMissingType);
 			}
 		}
 		if (firstBound != null)
@@ -1098,11 +1095,15 @@ public class TypeVariableBinding extends ReferenceBinding {
 		if (!isPrototype())
 			return this.prototype.setSuperClass(superclass);
 		this.superclass = superclass;
+		if (superclass != null)
+			this.tagBits |= superclass.tagBits & (TagBits.HasNullTypeAnnotation|TagBits.HasMissingType);
 		if ((this.tagBits & TagBits.HasAnnotatedVariants) != 0) {
 			TypeBinding [] annotatedTypes = getDerivedTypesForDeferredInitialization();
 			for (int i = 0, length = annotatedTypes == null ? 0 : annotatedTypes.length; i < length; i++) {
 				TypeVariableBinding annotatedType = (TypeVariableBinding) annotatedTypes[i];
 				annotatedType.superclass = superclass;
+				if (superclass != null)
+					annotatedType.tagBits |= superclass.tagBits & (TagBits.HasNullTypeAnnotation|TagBits.HasMissingType);
 			}
 		}
 		return superclass;
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/AnnotatableTypeSystem.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/AnnotatableTypeSystem.java
@@ -235,12 +235,16 @@ public class AnnotatableTypeSystem extends TypeSystem {
 				ArrayBinding arrayBinding = (ArrayBinding) type;
 				annotatedType = getArrayType(arrayBinding.leafComponentType, arrayBinding.dimensions, flattenedAnnotations(annotations));
 				break;
+			case Binding.TYPE_PARAMETER:
+				// simplified version given that type parameters have only one level of annotations.
+				// this also spares the below algorithm the need to specifically handle Binding.AWAITED_ANNOTATIONS
+				annotatedType = getAnnotatedType(type, null, annotations[0]);
+				break;
 			case Binding.BASE_TYPE:
 			case Binding.TYPE:
 			case Binding.GENERIC_TYPE:
 			case Binding.PARAMETERIZED_TYPE:
 			case Binding.RAW_TYPE:
-			case Binding.TYPE_PARAMETER:
 			case Binding.WILDCARD_TYPE:
 			case Binding.INTERSECTION_TYPE:
 			case Binding.INTERSECTION_TYPE18:
@@ -372,6 +376,8 @@ public class AnnotatableTypeSystem extends TypeSystem {
 		if (baseType != null && baseType.hasTypeAnnotations())
 			return true;
 		if (someType != null && someType.hasTypeAnnotations())
+			return true;
+		if (annotations == Binding.AWAITED_ANNOTATIONS)
 			return true;
 		for (int i = 0, length = annotations == null ? 0 : annotations.length; i < length; i++)
 			if (annotations [i] != null)
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Binding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Binding.java
@@ -67,6 +67,7 @@ public abstract class Binding {
 	public static final ReferenceBinding[] NO_MEMBER_TYPES = new ReferenceBinding[0];
 	public static final TypeVariableBinding[] NO_TYPE_VARIABLES = new TypeVariableBinding[0];
 	public static final AnnotationBinding[] NO_ANNOTATIONS = new AnnotationBinding[0];
+	public static final AnnotationBinding[] AWAITED_ANNOTATIONS = new AnnotationBinding[0];
 	public static final ElementValuePair[] NO_ELEMENT_VALUE_PAIRS = new ElementValuePair[0];
 	public static final char[][] NO_PARAMETER_NAMES = new char[0][];
 	public static final RecordComponentBinding[] NO_COMPONENTS = new RecordComponentBinding[0];
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ClassScope.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ClassScope.java
@@ -1390,6 +1390,7 @@ public class ClassScope extends Scope {
 		boolean previousFlag = environment().enterSuperTypeLookup(sourceType);
 		if ((sourceType.tagBits & TagBits.BeginHierarchyCheck) == 0) {
 			sourceType.tagBits |= TagBits.BeginHierarchyCheck;
+			preprocessTypeVariables(this.referenceContext.binding.typeVariables, this.referenceContext.typeParameters);
 			boolean noProblems = connectSuperclass();
 			noProblems &= connectSuperInterfaces();
 			if ((sourceType.typeBits & (TypeIds.BitAutoCloseable|TypeIds.BitCloseable)) != 0) {
@@ -1433,6 +1434,7 @@ public class ClassScope extends Scope {
 		boolean previousFlag = environment().enterSuperTypeLookup(sourceType);
 		try {
 			sourceType.tagBits |= TagBits.BeginHierarchyCheck;
+			preprocessTypeVariables(this.referenceContext.binding.typeVariables, this.referenceContext.typeParameters);
 			boolean noProblems = connectSuperclass();
 			noProblems &= connectSuperInterfaces();
 			environment().typesBeingConnected.remove(sourceType);
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/MethodBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/MethodBinding.java
@@ -1420,17 +1420,6 @@ public boolean doesParameterLengthMatch(int suggestedParameterLength) {
 	int len = this.parameters.length;
 	return len <= suggestedParameterLength || (isVarargs() && len == suggestedParameterLength + 1);
 }
-public void updateTypeVariableBinding(TypeVariableBinding previousBinding, TypeVariableBinding updatedBinding) {
-	TypeVariableBinding[] bindings = this.typeVariables;
-	if (bindings != null) {
-		for (int i = 0; i < bindings.length; i++) {
-			if (bindings[i] == previousBinding) { //$IDENTITY-COMPARISON$
-				bindings[i] = updatedBinding;
-			}
-		}
-	}
-}
-
 /**
  * Identifies whether the method has Polymorphic signature based on <a href=https://docs.oracle.com/javase/specs/jls/se11/html/jls-15.html#jls-15.12.3>jls-15.12.3</a><br/>
  *
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/Scope.java
@@ -854,16 +854,32 @@ public abstract class Scope {
 		return null; // incompatible
 	}
 
+	/* If the type variables will eventually come to be type annotated (express declaration site type annotations or applicable null defaults),
+	 * proactively create an "annotatable" variant now. When integrating annotations into hierarchy, we can then apply annotations `in place'.
+	 * Doing this now, ahead of connecting super types ensures that all references to the type variables in various AST nodes, will see the
+	 * annotatable variant being created here and once that gets annotated, all AST nodes into which the type variable binding was fanned out
+	 * by resolution will automatically see the update.
+	 */
+	void preprocessTypeVariables(TypeVariableBinding [] typeVariables, TypeParameter [] typeParameters) {
+		if (environment().typeSystem instanceof AnnotatableTypeSystem) {
+			int length;
+			if (typeVariables != null && typeParameters != null && (length = typeVariables.length) == typeParameters.length) {
+				for (int i = 0; i < length; i++) {
+					TypeParameter typeParameter = typeParameters[i];
+					TypeVariableBinding typeVariable = typeVariables[i];
+					if (typeParameter.binding != typeVariable) //$IDENTITY-COMPARISON$
+						throw new AssertionError("Unexpected type variable"); //$NON-NLS-1$
+					if ((typeParameter.bits & ASTNode.HasTypeAnnotations) != 0 || hasDefaultNullnessFor(Binding.DefaultLocationTypeParameter, typeParameter.sourceStart()))
+						typeParameters[i].binding = typeVariables[i] = (TypeVariableBinding) environment().createAnnotatedType(typeVariable, new AnnotationBinding [][] { Binding.AWAITED_ANNOTATIONS });
+				}
+			}
+		}
+	}
+
 	/**
 	 * Connect type variable supertypes, and returns true if no problem was detected
 	 */
 	protected boolean connectTypeVariables(TypeParameter[] typeParameters, boolean checkForErasedCandidateCollisions) {
-		/* https://bugs.eclipse.org/bugs/show_bug.cgi?id=305259 - We used to not bother with connecting
-		   type variables if source level is < 1.5. This creates problems in the reconciler if a 1.4
-		   project references the generified API of a 1.5 project. The "current" project's source
-		   level cannot decide this question for some other project. Now, if we see type parameters
-		   at all, we assume that the concerned java element has some legitimate business with them.
-		 */
 		if (typeParameters == null || typeParameters.length == 0) return true;
 		Map<TypeBinding, Object> invocations = new HashMap<>(2);
 		boolean noProblems = true;
@@ -1025,7 +1041,6 @@ public abstract class Scope {
 	}
 
 	public TypeVariableBinding[] createTypeVariables(TypeParameter[] typeParameters, Binding declaringElement) {
-		// https://bugs.eclipse.org/bugs/show_bug.cgi?id=324850, If they exist at all, process type parameters irrespective of source level.
 		if (typeParameters == null || typeParameters.length == 0)
 			return Binding.NO_TYPE_VARIABLES;
 
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeBinding.java
@@ -45,6 +45,7 @@ import org.eclipse.jdt.core.compiler.CharOperation;
 import org.eclipse.jdt.internal.compiler.ast.Wildcard;
 import org.eclipse.jdt.internal.compiler.impl.CompilerOptions;
 import org.eclipse.jdt.internal.compiler.tool.EclipseCompiler;
+import org.eclipse.jdt.internal.compiler.util.Util;
 
 /*
  * Not all fields defined by this type (& its subclasses) are initialized when it is created.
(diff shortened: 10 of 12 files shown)


srikanth-sankaran:
Hmm, I am surprised by a green test run from github for commit 1 - I expected at least the following tests to fail:

org.eclipse.jdt.core.tests.model.ExternalAnnotations9Test.testGH3623_withNullAnnoatations()
org.eclipse.jdt.core.tests.model.ExternalAnnotations9Test.testGH3623_nullAnnotationsNotEnabled()
org.eclipse.jdt.core.tests.compiler.regression.NullTypeAnnotationTest.testGH2158()

srikanth-sankaran:
> Hmm, I am surprised by a green test run from github for commit 1 - I expected at least the following tests to fail:
> 
> org.eclipse.jdt.core.tests.model.ExternalAnnotations9Test.testGH3623_withNullAnnoatations() org.eclipse.jdt.core.tests.model.ExternalAnnotations9Test.testGH3623_nullAnnotationsNotEnabled() org.eclipse.jdt.core.tests.compiler.regression.NullTypeAnnotationTest.testGH2158()

It gets curiouser and curiouser!

I have pushed the regression test (alone, without code changes) from https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4682 here and I expect it to pass also - as it passes locally. This is completely counterintuitive!!!

So it would seem backing out the fixes for #2158 and #3623 solves #2158 and #3623 and https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4494 - no explanations yet. I ran into this scenario while simply preparing the grounds for understanding status quo ante  #2158 and #3623 

@stephan-herrmann - FYI. You can hold off on tracking this until I call out.

stephan-herrmann:
I have some news for you here: if additionally you'd backout #4554 (4.38 M3), then you'd see org.eclipse.jdt.core.tests.compiler.regression.NullTypeAnnotationTest.testGH2158() failing again.

Method `Scope.areSignificantlyDifferent(TypeBinding, TypeBinding)` is intended to filter out differences caused only be "equivalent" but different **wildcard** bindings. Apparently that method now also masks a problem of unexpectedly different ids. I can indeed observe two different PTBs with different ids, while all their ingredients (type, arguments ...) have the same ids pairwise. I.e., `!areSignificantlyDifferent()` and also `TypeBinding.notEquals()` and even with no wildcards in the picture.

HTH

srikanth-sankaran:
> I have some news for you here: if additionally you'd backout #4554 (4.38 M3), then you'd see org.eclipse.jdt.core.tests.compiler.regression.NullTypeAnnotationTest.testGH2158() failing again.

This saves me lots of time in digging - thanks.

srikanth-sankaran:
This PR at this point backs out various fixes made in the past - Some tests are expected to fail. Alternate fix will follow in separate commit

srikanth-sankaran:
@stephan-herrmann - The last commit contains larval version of evolving solution - you are welcome to peek at it - but I am aware lot more loose ends needing addressing so wait for a review request before spending too much time on it ...

Passes NTAT and `org.eclipse.jdt.core.tests.compiler.regression.TestAll` locally at levels 1.8, 16 and 25.

srikanth-sankaran:
> @stephan-herrmann - The last commit contains larval version of evolving solution - you are welcome to peek at it - but I am aware lot more loose ends needing addressing so wait for a review request before spending too much time on it ...
> 
> Passes NTAT and `org.eclipse.jdt.core.tests.compiler.regression.TestAll` locally at levels 1.8, 16 and 25.

Tests are green, but the CI build and test fails a quality gate criterion due to an additional warning resulting in the build. Pushed a commit to address that - expecting all green now

srikanth-sankaran:
Next order business is to move the creation of `annotatable` type variables from `org.eclipse.jdt.internal.compiler.lookup.Scope.createTypeVariables(TypeParameter[], Binding)` into `org.eclipse.jdt.internal.compiler.lookup.ClassScope.connectTypeHierarchy()` and `org.eclipse.jdt.internal.compiler.lookup.ClassScope.connectTypeHierarchyWithoutMembers()` for classes and into `org.eclipse.jdt.internal.compiler.lookup.SourceTypeBinding.resolveTypesFor(MethodBinding)` for generic methods. 

Reason for this is in `org.eclipse.jdt.internal.compiler.lookup.Scope.createTypeVariables(TypeParameter[], Binding)` imports are not yet processed and so `hasDefaultNullnessFor()` cannot be queried to create an annotatable `TypeVariableBinding` (that can then be annotated in place, instead of asking the TypeSystem for an annotated type)

Painfully, `org.eclipse.jdt.internal.compiler.lookup.ClassScope.hasDefaultNullnessFor(int, int)` has the side effect of evaluating null annotations instead of being a pure yes/no answering method. This will require some doing to work around ...

srikanth-sankaran:
> Painfully, `org.eclipse.jdt.internal.compiler.lookup.ClassScope.hasDefaultNullnessFor(int, int)` has the side effect of evaluating null annotations instead of being a pure yes/no answering method. This will require some doing to work around ...

This itself turned out to be easy - we just needed to see if type variables are connected ...

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceVariable.java
```diff
@@ -85,7 +85,7 @@ public static InferenceVariable get(TypeBinding typeParameter, int rank, Invocat
 	InvocationSite site;
 	TypeBinding typeParameter;
 	long nullHints; // one of TagBits.{AnnotationNonNull,AnnotationNullable} may steer inference into inferring nullness as well; set both bits to request avoidance.
-	private InferenceVariable prototype;
+	//private InferenceVariable prototype;
```
stephan-herrmann: quick question: is this change part of a behavior preserving pull-up refactoring, or does that change also fix a semantic problem?

srikanth-sankaran: No, this is not to fix any semantic problem. As I have had to introduce a `prototype` state variable in `TypeVariableBinding` for the current solution under consideration, I am thinking the same state variable would suffice for the purposes of `InferenceVariable` and `CaptureBinding18` - I have not yet studied the uses of this state in these classes to convince myself that this is doable - notwithstanding the tests being green, all existing uses will have to be studied.

srikanth-sankaran:
A much more evolved solution has been pushed - this fails two tests when running `org.eclipse.jdt.core.tests.compiler.regression.TestAll` locally with -Dcompliance=25 the failing tests being:

`org.eclipse.jdt.core.tests.compiler.regression.NullTypeAnnotationTest.testBug501564()` and
`org.eclipse.jdt.core.tests.compiler.regression.NullTypeAnnotationTest.testBug501564interface()`

srikanth-sankaran:
Basically we introduce a "preprocessing" step, wherein type variables that will eventually get decorated with type annotations from declaration site (explicitly or magically via `@NonNullByDefault`) will get replaced with a `TypeVariableBinding` instance that can be annotated **in place** - this instance is a `derived` binding from the original unannotated type variable which stays unannotated even as the clone gets decorated. 

In essence, the original type variables created in `org.eclipse.jdt.internal.compiler.lookup.Scope.createTypeVariables(TypeParameter[], Binding)` serve as the unannotated primordial prototype from which annotated variants get cloned. 

It is of particular importance to note that TypeParameter's (declaration site) annotations are decorated in place while type argument's (use site) annotations may result in a fresh binding.

srikanth-sankaran:
@stephan-herrmann - I think this PR is failing two tests viz

`org.eclipse.jdt.core.tests.compiler.regression.NullTypeAnnotationTest.testBug501564()` and
`org.eclipse.jdt.core.tests.compiler.regression.NullTypeAnnotationTest.testBug501564interface()`

I would need you to weigh in on this. I think the test expectations need to be adjusted. Prior to the PR, given this class

```
@NonNullByDefault(DefaultLocation.TYPE_PARAMETER) 
static class Bar<E3 extends Generic<E3>> { }
```

after `org.eclipse.jdt.internal.compiler.ast.TypeParameter.resolveAnnotations(Scope)` `E3`'s binding reads: 

`@NonNull E3 extends Generic<E3 extends Generic<E3>>`

while after the PR at the same point it reads: 

`@NonNull E3 extends Generic<@NonNull E3>`

See that shallow application of NonNull annotation earlier vs deeper application now (actually happens because of in place annotation application on the same TVB that have been fanned out to decorate different AST nodes earlier by lookup)

I think the new behavior is the correct one and this would call for the failing tests to be adjusted. Could you push a commit after verifying this claim ? TIA.

If you apply this diff on top of this PR, you can see interesting behavior change on these two tests. One of them starts passing and other shows one error less.

```
diff --git a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeParameter.java b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeParameter.java
index 8736f06..8e55b0c 100644
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeParameter.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeParameter.java
@@ -148,6 +148,15 @@
 								System.arraycopy(annotationBindings, 0, newAnnotations, 0, annotationBindings.length);
 							newAnnotations[newAnnotations.length - 1] = environment.getNonNullAnnotation();
 							this.binding.setTypeAnnotations(newAnnotations, isAnnotationBasedNullAnalysisEnabled);
+							if (CharOperation.equals(this.name, "E3".toCharArray())) { //$NON-NLS-1$
+								// Prior to the PR 4740, E3's binding reads: `@NonNull E3 extends Generic<E3 extends Generic<E3>>` while after the PR it reads: `@NonNull E3 extends Generic<@NonNull E3>`
+								// This is material in the test failures for NTAT.testBug501564*
+								if (this.binding.superclass instanceof ParameterizedTypeBinding)
+									this.binding.superclass = (ReferenceBinding) scope.environment().getUnannotatedType(this.binding.superclass);
+								else if (this.binding.superInterfaces != null && this.binding.superInterfaces.length == 1)
+									this.binding.superInterfaces[0] = (ReferenceBinding) scope.environment().getUnannotatedType(this.binding.superInterfaces[0]);
+								this.binding.firstBound = scope.environment().getUnannotatedType(this.binding.firstBound);
+							}
 						}
 					}
 					this.binding.evaluateNullAnnotations(scope, this);
```

This confirms the deeper vs shallow annotation application on self referential type variables being the material factor in these couple of tests "failing"

srikanth-sankaran:
Notes to reviewer: Here is the recommended order of review.

- Glance through the first 6 commits which reverse past fixes in this space, verifying structural goodness of the reverts.

For the core fix embodied by the 7th commit:

- Scope.preprocessTypeVariables(): Create _annotatable_ variants ahead of time. See comment in code.
- ClassScope: Invokes Scope.preprocessTypeVariables()
- SourceTypeBinding: Invokes Scope.preprocessTypeVariables(); Looses a method no longer needed
- `org.eclipse.jdt.internal.compiler.ast.TypeParameter.resolveAnnotations(Scope)` : **In place** application of annotations on the TVB that has been prepared for annotatability by the previous step. This ensures that for recursive/self referential TVBs all nodes into which the TVB has been fanned out during resolve see the change. (see that earlier code does in place updation in the first part and creates a fresh variant in the second part)
- AnnotatableTypeSystem: Changes here are to cope with "awaited" type annotations.: to create an annotated type even though its annotations are not available yet, recognize types as being type-annotated if they are awaiting annotations etc.
- TypeBinding: Show awaited annotations in debug toString output; Accept awaited annotations in set annotations.
- CaptureBinding18 & InferenceVariable : I have only cursorily verified that the pull up of the prototype doesn't have any unintended effects. **Please review this part carefully!** 
- Glance through the other files

srikanth-sankaran:
@stephan-herrmann - Please review! 

CaptureBinding18 & InferenceVariable changes and the pending test changes to fix the failures I am leaning on you extra to verify. TIA

stephan-herrmann:
> I think the new behavior is the correct one and this would call for the failing tests to be adjusted. Could you push a commit after verifying this claim ? TIA.

You are right. I updated/extended the two failing tests in [21cf797](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4740/commits/21cf797f7327ba75c5f57c721ad42b83b751b40c)

stephan-herrmann:
> * `org.eclipse.jdt.internal.compiler.ast.TypeParameter.resolveAnnotations(Scope)` : **In place** application of annotations on the TVB that has been prepared for annotatability by the previous step. This ensures that for recursive/self referential TVBs all nodes into which the TVB has been fanned out during resolve see the change. (see that earlier code does in place updation in the first part and creates a fresh variant in the second part)

Quick check on the last part: you are saying that this line in existing code
https://github.com/eclipse-jdt/eclipse.jdt.core/blob/71f2d35f5e5a4344e087b2f1255e4f6e9eddff6c/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeParameter.java#L146
was breaking the self-referential type, right?

srikanth-sankaran:
> 
> Quick check on the last part: you are saying that this line in existing code
> 
> https://github.com/eclipse-jdt/eclipse.jdt.core/blob/71f2d35f5e5a4344e087b2f1255e4f6e9eddff6c/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeParameter.java#L146
> 
> was breaking the self-referential type, right?

Yes. Also line 135 and line 146 are taking different routes.

stephan-herrmann:
Before I do a final walkthrough do you have a comment on the relation of your changes with the compiler's phase model, in particular with TypeDeclaration.updateSupertypesWithAnnotations(Map<ReferenceBinding, ReferenceBinding>).  Is there any interaction, or is it orthogonal? More specifically:
* do these changes affect the order of processing in any way?
* could these changes make any late updating unnecessary?

My first intuition was, that those concerns do interact, but after spending a few minutes on it I'm less sure.

stephan-herrmann:
> > Quick check on the last part: you are saying that this line in existing code
> > https://github.com/eclipse-jdt/eclipse.jdt.core/blob/71f2d35f5e5a4344e087b2f1255e4f6e9eddff6c/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeParameter.java#L146
> > 
> > was breaking the self-referential type, right?
> 
> Yes. Also line 135 and line 146 are taking different routes.

I read this as:
* line 135 does inline updating: :+1: 
* line 146 was replacing with a new binding: :-1: 

Right?

srikanth-sankaran:
> Before I do a final walkthrough do you have a comment on the relation of your changes with the compiler's phase model, in particular with TypeDeclaration.updateSupertypesWithAnnotations(Map<ReferenceBinding, ReferenceBinding>). Is there any interaction, or is it orthogonal? 

No, there is no interaction except the obvious preprocessing of type variables for annotatability by a downstream phase.


>More specifically:
> 
> do these changes affect the order of processing in any way?
> could these changes make any late updating unnecessary?

I verified the the code in TypeDeclaration.updateSupertypesWithAnnotations(Map<ReferenceBinding, ReferenceBinding>) is very much essential although very early on I felt these would get significantly simplified. In the end the final form of the fix doesn't touch these methods that implement the integrate annotations into hierarchy phase.

srikanth-sankaran:
> I read this as:
> 
> line 135 does inline updating: 👍
> line 146 was replacing with a new binding: 👎
> Right?

Not quite a straightforward thumbs up or down situation. 

In the old scheme of things both have drawbacks. 

Directly annotating the TVB in place is trouble because the prior unannotated version has been fanned out to various nodes where it is intended to stay unannotated. 


OTOH, Spawning a clone comes with its own challenges wrt to shallow annotation v deep as we saw in those failing tests. 

The present approach is safe because we recognize bindings that will _eventually_ be annotated and prepare them for annotatability by creating a variant of the prototype (a pure unannotated version.)

When we subsequently annotate in place, all nodes which should see the annotation will automatically reflect them while locations which should see the unannotated version will not be impacted.

stephan-herrmann:
During debugging I found a bug in `TypeVariableBinding.toString()`, to be fixed like this:
```diff
- 		if (this.hasTypeAnnotations())
-			return annotatedDebugName();
+		if (this.hasTypeAnnotations()) 
+			return '<'+annotatedDebugName()+'>';
```

Without this an STB `Foo<@NonNull M>` would be rendered as `Foo<NonNull >`, because `STB.toString()` expects a TVB name to be enclosed in `<>` which it then cuts off in 
```
			buffer.append(varChars, 1, varChars.length - 2);
```

@srikanth-sankaran do you mind if I push this fix to the current PR? The relation being: without the fix the effect of properly attaching type annotations to type variables is harder to see in the debugger :)

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/apt/dispatch/AnnotationDiscoveryVisitor.java
```diff
@@ -155,7 +155,7 @@ public boolean visit(TypeParameter typeParameter, BlockScope scope) {
 			// when we get here, it is guaranteed that class type parameters are connected, but method type parameters may not be.
 			MethodBinding methodBinding = (MethodBinding) binding.declaringElement;
 			((SourceTypeBinding) methodBinding.declaringClass).resolveTypesFor(methodBinding);
-			this.resolveAnnotations(scope, annotations, binding);
+			this.resolveAnnotations(scope, annotations, typeParameter.binding /* treat as volatile as Scope.preprocessTypeVariables() called by resolveTypesFor() may modify!!! */);
```
stephan-herrmann, line 158: I could not find a test where `binding != typeParameter.binding`.
Are you able to produce a test case that needs to re-fetch `typeParameter.binding`?
Or is this just caution based on principle?

srikanth-sankaran: Yes, Java8ElementsTests fails two tests without this change. Caught by an earlier CI run.

srikanth-sankaran:
> do you mind if I push this fix to the current PR?

Not at all. I'll push a commit in a few hours.

srikanth-sankaran:
> Not at all. I'll push a commit in a few hours.

Done

srikanth-sankaran:
The 118 failures are due to https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4821 - It is unclear why the PR https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4821 did not result in these failures! Some sort of a process glitch ?? Looking at that page, I see only 6 checks succeeded instead of the usual 13 or so. 

@mpalat is planning to revert that fix I believe, I will rebase and push once that happens.

mpalat:
> @mpalat is planning to revert that fix I believe, I will rebase and push once that happens.

@srikanth-sankaran - I have merged the revert #4822.

srikanth-sankaran:
> @srikanth-sankaran - I have merged the revert #4822.

Thanks! I have rebased the current PR on top of it.

srikanth-sankaran:
@stephan-herrmann - one more asymmetry I wanted you to comment on is 

`org.eclipse.jdt.internal.compiler.lookup.TypeVariableBinding.setFirstBound(TypeBinding)` doing this:

`if (firstBound != null)
			this.tagBits |= firstBound.tagBits & (TagBits.HasNullTypeAnnotation|TagBits.HasMissingType);`

but not `org.eclipse.jdt.internal.compiler.lookup.TypeVariableBinding.setSuperClass(ReferenceBinding)` (on master) 

In the current PR, I have made an half-hearted effort at fixing this. i.e fixed for `org.eclipse.jdt.internal.compiler.lookup.TypeVariableBinding.setSuperClass(ReferenceBinding)`  but not for `org.eclipse.jdt.internal.compiler.lookup.TypeVariableBinding.setSuperInterfaces(ReferenceBinding[])` 

Could you check what is the right thing here ? TIA

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceVariable.java
stephan-herrmann: I have isolated that "refactoring" into a separate patch:
[TypeVariableBinding-prototype-protocol.patch](https://github.com/user-attachments/files/25150925/TypeVariableBinding-prototype-protocol.patch)

Looking at it this way I don't see any risks in this part.

@srikanth-sankaran do you have plans how to finally release this PR? I think it would make sense to retain some separation of reverts and new solution. Perhaps this small refactoring would also make for a separate, intermediate commit in the final history. With this the new solution would hopefully remain easier to grasp for posterity? WDYT?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/apt/dispatch/AnnotationDiscoveryVisitor.java
stephan-herrmann: Thanks, debugging that test helped understand this change.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceVariable.java
srikanth-sankaran: I am fine with the suggestion.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeParameter.java
```diff
@@ -142,22 +142,12 @@ public void resolveAnnotations(Scope scope) {
...
-										methodBinding.updateTypeVariableBinding(previousBinding, this.binding);
-									}
-								}
-							}
+							annotationBindings = this.binding.getTypeAnnotations();
+							AnnotationBinding [] newAnnotations = new AnnotationBinding[annotationBindings == null ? 1 : annotationBindings.length + 1];
+							if (annotationBindings != null)
+								System.arraycopy(annotationBindings, 0, newAnnotations, 0, annotationBindings.length);
+							newAnnotations[newAnnotations.length - 1] = environment.getNonNullAnnotation();
+							this.binding.setTypeAnnotations(newAnnotations, true);
```
stephan-herrmann, line 154: Thinking aloud during review, I see two changes in this block
* switch from replacing `this.binding` to in-place updating with `setTypeAnnotations()`
* revert the workaround from https://bugs.eclipse.org/501031
   * see also removal of `MethodBinding.updateTypeVariableBinding()`

✔️  looks good so far.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceVariable.java
srikanth-sankaran: I don't have my laptop for next 36 hours or so. If you wish to push commits to reorganize the PR by isolating the prototype pull up related changes and have the cycles do so. TIA.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/SourceTypeBinding.java
```diff
@@ -2366,24 +2368,6 @@ public void evaluateNullAnnotations() {
...
-		AnnotationBinding[] annots = new AnnotationBinding[]{ this.environment.getNonNullAnnotation() };
-		for (int i = 0; i < this.typeVariables.length; i++) {
-			TypeVariableBinding tvb = this.typeVariables[i];
-			TypeParameter typeParameter = this.scope.referenceContext.typeParameters[i];
-			if (typeParameter.annotations != null && (tvb.extendedTagBits & ExtendedTagBits.AnnotationResolved) == 0)
-				continue; // not yet ready
-			if ((tvb.tagBits & TagBits.AnnotationNullMASK) == 0)
-				this.typeVariables[i] = (TypeVariableBinding) this.environment.createAnnotatedType(tvb, annots);
-		}
-	}
```
stephan-herrmann, line 2386: thinking aloud:
* this removes a kludge from https://bugs.eclipse.org/429958 :

> Marking nonnullness of type variables happens after the fact, because when type variables are created we don't yet have the necessary information. Instead, a new method maybeMarkTypeParametersNonNull() is hooked into STB.evaluateNullAnnotations().

Fortunately this issue ("we don't yet have the necessary information") is now being resolved properly :)

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeVariableBinding.java
```diff
@@ -103,11 +99,15 @@ protected TypeVariableBinding(char[] sourceName, LookupEnvironment environment)
...
 		this.typeBits = TypeIds.BitUninitialized;
+		this.prototype = this;
 		// don't yet compute the ID!
 	}
 
 	public TypeVariableBinding(TypeVariableBinding prototype) {
 		super(prototype);
+		this.prototype = prototype.prototype;
+		this.prototype.tagBits |= TagBits.HasAnnotatedVariants;
+		this.tagBits &= ~TagBits.HasAnnotatedVariants;
```
stephan-herrmann, line 110: These are subtle changes, for which I'd appreciate a little explanation:
* pulling up from the bottom of this method looks like a (useful) cosmetic change, right?
* changing from `prototype.tagBits` to `this.prototype.tagBits` looks like a bug fix? Or should that be seen as part of the refactoring "pull up prototype"?
* while not new in this change I'm a bit puzzled about the justification for setting these tagbits in the first place: call hierarchy shows we are in the context of some `clone()` method. Is that a sufficient justification for marking the type with `HasAnnotatedVariants`?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeVariableBinding.java
```diff
@@ -864,18 +871,6 @@ ReferenceBinding resolve() {
...
 
-	@Override
-	public void setTypeAnnotations(AnnotationBinding[] annotations, boolean evalNullAnnotations) {
-		if (getClass() == TypeVariableBinding.class && this.environment.globalOptions.storeAnnotations) {
-			// TVB only: if the declaration itself carries type annotations,
-			// make sure TypeSystem will still have an unannotated variant at position 0, to answer getUnannotated()
-			// (in this case the unannotated type is never explicit in source code, that's why we need this charade).
-			this.environment.typeSystem.forceRegisterAsDerived(this);
-		} else {
-			this.environment.getUnannotatedType(this); // exposes original TVB/capture to type system for id stamping purposes.
```
stephan-herrmann, line 875: Thinking aloud:
* In addition to the workarounds explicitly reverted, this line had survived from https://bugs.eclipse.org/418041 and finally is no longer needed. It's this removal which finally made the entire method override obsolete.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeVariableBinding.java
srikanth-sankaran: > These are subtle changes, for which I'd appreciate a little explanation:
> * pulling up from the bottom of this method looks like a (useful) cosmetic change, right?

Yes. 

> * changing from `prototype.tagBits` to `this.prototype.tagBits` looks like a bug fix? Or should that be seen as part of the refactoring "pull up prototype"?


These changes mimic changes in other classes that define a prototype. STB, URB, BTB etc. 

Earlier on TVB didn't have a prototype field, this PR introduces it. We really want to say the prototype has annotated variants and so updating the bits on this.prototype is the right thing to do. 

I think it is not technically necessary to clear that bit on an annotated variant ..,


> * while not new in this change I'm a bit puzzled about the justification for setting these tagbits in the first place: call hierarchy shows we are in the context of some `clone()` method. Is that a sufficient justification for marking the type with `HasAnnotatedVariants`? 

The very reason clone() exists/was created was to allow for annotated variants. It is not meant to be a general purpose copy creation utility method, though it could serve that purpose too

stephan-herrmann: thanks

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeVariableBinding.java
```diff
@@ -1077,12 +1072,16 @@ private TypeReference findBound(TypeBinding bound, TypeParameter parameter) {
...
+		if (!isPrototype())
+			return this.prototype.setFirstBound(firstBound);
 		this.firstBound = firstBound;
 		if ((this.tagBits & TagBits.HasAnnotatedVariants) != 0) {
 			TypeBinding [] annotatedTypes = getDerivedTypesForDeferredInitialization();
 			for (int i = 0, length = annotatedTypes == null ? 0 : annotatedTypes.length; i < length; i++) {
 				TypeVariableBinding annotatedType = (TypeVariableBinding) annotatedTypes[i];
 				annotatedType.firstBound = firstBound;
+				if (firstBound != null)
+					annotatedType.tagBits |= firstBound.tagBits & (TagBits.HasNullTypeAnnotation|TagBits.HasMissingType);
```
stephan-herrmann, line 1084: Looking for a test needing this addition, I made a strange observation:

Test is NullTypeAnnotationTest.testBug466562() where this change toggles the exact error message being reported.

Set a breakpoint in `ProblemReporter.nullityMismatchIsNull(Expression, TypeBinding)`

When it hits, I see `requiredType` as `<@<Awaited ...> T1 extends @Nullable Number>`

Shouldn't all references to `<Awaited ...>` be replaced with real stuff by the time we proceed to `analyseCode()`?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceVariable.java
stephan-herrmann: done

stephan-herrmann:
I pushed a variant prepared for merging without squashing:
* the series of revert commits
* one more preparation commit (pull-up refactoring plus tiny fix in toString())
* the squashed new solution (where I only [sneaked in](https://github.com/eclipse-jdt/eclipse.jdt.core/compare/a6436f0311ce4e59855e1c991c336d03bbc672f8..079d9fdb49b6df5ea82f555c80ad7d44a013358e) one more code comment :smile:)

I confirmed that the revert commits are cleanly done.

For the final solution commit I inspected every line of code and found this change solid and well comprehensible! 

At this point my main open question is https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4740/changes#r2777610810

srikanth-sankaran:
> At this point my main open question is https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4740/changes#r2777610810

Is this link correct ? I am not able to click on it and figure out the open question. 

(Is that about '@Awaited' still being seen in analyseCode? Yes that looks odd I'll debug that when I get to my laptop Monday morning)

stephan-herrmann:
> > At this point my main open question is https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4740/changes#r2777610810
> 
> Is this link correct ? I am not able to click on it and figure out the open question.

While github indeed has the habit to invalidate links the worked a minute ago, it still works for me ...

> (Is that about '@awaited' still being seen in analyseCode?

You guessed right.

> Yes that looks odd I'll debug that when I get to my laptop Monday morning)

thanks!

stephan-herrmann:
> @stephan-herrmann - one more asymmetry I wanted you to comment on is
> 
> `org.eclipse.jdt.internal.compiler.lookup.TypeVariableBinding.setFirstBound(TypeBinding)` doing this:
> 
> `if (firstBound != null) this.tagBits |= firstBound.tagBits & (TagBits.HasNullTypeAnnotation|TagBits.HasMissingType);`
> 
> but not `org.eclipse.jdt.internal.compiler.lookup.TypeVariableBinding.setSuperClass(ReferenceBinding)` (on master)
> 
> In the current PR, I have made an half-hearted effort at fixing this. i.e fixed for `org.eclipse.jdt.internal.compiler.lookup.TypeVariableBinding.setSuperClass(ReferenceBinding)` but not for `org.eclipse.jdt.internal.compiler.lookup.TypeVariableBinding.setSuperInterfaces(ReferenceBinding[])`
> 
> Could you check what is the right thing here ? TIA

I'll come back here, once we have clarity on `<Awaited ...>` appearing still during `analyseCode()`, as that one was related to exactly these code locations.

stephan-herrmann:
Jenkins writes:
> All nodes of label ‘[ubuntu-latest](https://ci.eclipse.org/jdt/label/ubuntu-latest/)’ are offline

stephan-herrmann:
> Jenkins writes:
> 
> > All nodes of label ‘[ubuntu-latest](https://ci.eclipse.org/jdt/label/ubuntu-latest/)’ are offline

See https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/7125

srikanth-sankaran:
> While github indeed has the habit to invalidate links the worked a minute ago, it still works for me ...

I think it is a problem that shows up only on github mobile app. Now that I am back on my laptop the link works fine.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeVariableBinding.java
srikanth-sankaran: Problem arises from there being two empty annotations array viz `org.eclipse.jdt.internal.compiler.lookup.Binding.NO_ANNOTATIONS` and `org.eclipse.jdt.internal.compiler.lookup.Binding.AWAITED_ANNOTATIONS` and we fail to recognize that they are semantically not the same 😆 

Fix is to recognize that not all empties are created equal and force replace one empty with another when called for.

srikanth-sankaran: Fix as diff: 

```
diff --git a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeParameter.java b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeParameter.java
index dcc36cc..ea30f3f 100644
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeParameter.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeParameter.java
@@ -135,6 +135,10 @@
 				this.binding.setTypeAnnotations(annotationBindings, isAnnotationBasedNullAnalysisEnabled);
 				scope.referenceCompilationUnit().compilationResult.hasAnnotations = true;
 			}
+			if (this.binding.getTypeAnnotations() == Binding.AWAITED_ANNOTATIONS) { // In <T extends @Nullable Number>; T itself doesn't carry annotations, some bound does
+				this.binding.setTypeAnnotations(Binding.NO_ANNOTATIONS, isAnnotationBasedNullAnalysisEnabled); // not awaited anymore
+				scope.referenceCompilationUnit().compilationResult.hasAnnotations = true;
+			}
 			if (isAnnotationBasedNullAnalysisEnabled) {
 				if (this.binding != null && this.binding.isValidBinding()) {
 					if (scope.hasDefaultNullnessFor(Binding.DefaultLocationTypeParameter, this.sourceStart())) {
diff --git a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeBinding.java b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeBinding.java
index 398ef21..d5e019c 100644
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/TypeBinding.java
@@ -45,6 +45,7 @@
 import org.eclipse.jdt.internal.compiler.ast.Wildcard;
 import org.eclipse.jdt.internal.compiler.impl.CompilerOptions;
 import org.eclipse.jdt.internal.compiler.tool.EclipseCompiler;
+import org.eclipse.jdt.internal.compiler.util.Util;
 
 /*
  * Not all fields defined by this type (& its subclasses) are initialized when it is created.
@@ -1648,7 +1649,7 @@
 
 public void setTypeAnnotations(AnnotationBinding[] annotations, boolean evalNullAnnotations) {
 	this.tagBits |= TagBits.HasTypeAnnotations;
-	if (annotations == null || (annotations.length == 0 && annotations != Binding.AWAITED_ANNOTATIONS))
+	if (annotations == null || (annotations.length == 0 && Util.effectivelyEqual(annotations, this.typeAnnotations)))
 		return;
 	this.typeAnnotations = annotations;
 	if (evalNullAnnotations) {

```

srikanth-sankaran:
> I confirmed that the revert commits are cleanly done.
> 
> For the final solution commit I inspected every line of code and found this change solid and well comprehensible!
> 
> At this point my main open question is https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4740/changes#r2777610810

Thanks! With the last point addressed trivially I will go ahead and integrate this and raise separate ticket for tangential follow up work

srikanth-sankaran:
The follow up actions are: 

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4823
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4824
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2210 (ready to be closed ? I believe so)

srikanth-sankaran:
> The follow up actions are:
> 
> #4823 #4824 #2210 (ready to be closed ? I believe so)

See also https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4825 (under test, but passes TestAll at `-Dcompliance=25` locally)

srikanth-sankaran:
> See also #4825 (under test, but passes TestAll at `-Dcompliance=25` locally)

Is all green

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeParameter.java
```diff
@@ -135,29 +135,23 @@ public void resolveAnnotations(Scope scope) {
 				this.binding.setTypeAnnotations(annotationBindings, isAnnotationBasedNullAnalysisEnabled);
 				scope.referenceCompilationUnit().compilationResult.hasAnnotations = true;
 			}
+			if (this.binding.getTypeAnnotations() == Binding.AWAITED_ANNOTATIONS) { // In <T extends @Nullable Number>; T itself doesn't carry annotations, some bound does
+				this.binding.setTypeAnnotations(Binding.NO_ANNOTATIONS, isAnnotationBasedNullAnalysisEnabled); // not awaited anymore
```
stephan-herrmann, line 139: A late finding: this `setTypeAnnotations(Binding.NO_ANNOTATIONS, ...)` will set `TagBits.HasTypeAnnotations` which looks wrong.

@srikanth-sankaran  do you have a follow-up issue where we could sneak in just a tiny check for `NO_ANNOTATIONS` inside `setTypeAnnotations()`?

srikanth-sankaran: > A late finding: this `setTypeAnnotations(Binding.NO_ANNOTATIONS, ...)` will set `TagBits.HasTypeAnnotations` which looks wrong.

Are you sure ??

I think the concerned TVB *already* has that bit set and *correctly* so. (otherwise it would never have the `AWAITED_ANNOTATIONS` place holder in the no null default scenario) 

The ASTNode `TypeParameter` gets its `HasTypeAnnotations` bit set even if `TypeParameter` itself does not have type annotations, but its supertype or bounds have them.

See 

`org.eclipse.jdt.internal.compiler.parser.Parser.consumeTypeParameter1WithExtends()`
`org.eclipse.jdt.internal.compiler.parser.Parser.consumeTypeParameter1WithExtendsAndBounds()`
`org.eclipse.jdt.internal.compiler.parser.Parser.consumeTypeParameterHeader()` 

etc.

I think what is true for `ASTNode` is also the case for `TypeBinding` including `TypeVariableBinding`

Let me know if you see any specific scenario that indicates otherwise.

stephan-herrmann: Thanks for explanations!

Indeed I missed, that (a) the bit was already set when we get there, and that this happens (b) even if type annotations appear only on details, not the type parameter itself.

As for the need to have `HasTypeAnnotations` in `TypeVariableBinding.tagBits` in such a case: I couldn't readily find any productive code that really needs it (tests run green also when `preprocessTypeVariables()` checks `typeParameter.annotations` rather than `.bits`), but indeed `TypeVariableBinding.debugName()` needs the flag set :smile: 

So while there is a tiny bit of unnecessary churn on handling these things when type annotations exist only on details, I don't (no longer) see a semantic problem. Thanks.

Issue #4745, opened 2026-01-14
No build triggered if output folder was deleted externally


iloveeclipse:
1) Create plain Java project with default `src` & `bin` directories, have some source classes, let Eclipse build everything.
2) Shutdown Eclipse, remove `bin` output folder completely from file system
3) Start Eclipse again
4) Expected: Eclipse detects missing output folder, creates it and triggers a build, so the `bin` directory has same classes as before shutdown
5) Observed: `bin` directory is recreated but it is **empty**. The project itself has no compilation errors **but** all dependent projects show compilation errors because (of course) nothing is compiled. 

The use case is coming from a real world scenario where users do a `git clean -fdx & git pull`, getting latest state of the code & seeing compilation errors on next Eclipse startup even if no compilation errors exist on the branch.

**Notes**

The `bin` output folder is recreated here:
```
	at org.eclipse.core.internal.resources.Folder.create(Folder.java:87)
	at org.eclipse.jdt.internal.core.builder.NameEnvironment.createOutputFolder(NameEnvironment.java:556)
	at org.eclipse.jdt.internal.core.builder.NameEnvironment.computeClasspathLocations(NameEnvironment.java:198)
	at org.eclipse.jdt.internal.core.builder.NameEnvironment.<init>(NameEnvironment.java:82)
	at org.eclipse.jdt.internal.core.builder.JavaBuilder.initializeBuilder(JavaBuilder.java:692)
	at org.eclipse.jdt.internal.core.builder.JavaBuilder.build(JavaBuilder.java:211)
	at org.eclipse.core.internal.events.BuildManager$2.run(BuildManager.java:1109)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:47)
	at org.eclipse.core.internal.events.BuildManager.basicBuild(BuildManager.java:299)
	at org.eclipse.core.internal.events.BuildManager.basicBuild(BuildManager.java:355)
	at org.eclipse.core.internal.events.BuildManager$1.run(BuildManager.java:449)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:47)
	at org.eclipse.core.internal.events.BuildManager.basicBuild(BuildManager.java:452)
	at org.eclipse.core.internal.events.BuildManager.basicBuildLoop(BuildManager.java:564)
	at org.eclipse.core.internal.events.BuildManager.basicBuildLoop(BuildManager.java:512)
[...]

PR #4746, opened 2026-01-14
Perform a full build if output folder was deleted externally

iloveeclipse:
If the output folder is deleted externally, Java builder doesn't rebuild anything on next auto / incremental build request.

The change checks the missing output folder on the next build request and converts build kind to a FULL_BUILD if at least one output folder in the project is missing.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4745

### DIFF
--- a/org.eclipse.jdt.core.tests.builder/src/org/eclipse/jdt/core/tests/builder/BuildpathTests.java
+++ b/org.eclipse.jdt.core.tests.builder/src/org/eclipse/jdt/core/tests/builder/BuildpathTests.java
@@ -1018,24 +1018,52 @@ public void test0100() throws JavaModelException {
 }
 
 // https://bugs.eclipse.org/bugs/show_bug.cgi?id=143025
+// https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4745
 public void testMissingOutputFolder() throws JavaModelException {
 	IPath projectPath = env.addProject("P"); //$NON-NLS-1$
+	IJavaProject p = env.getJavaProject(projectPath);
 	env.removePackageFragmentRoot(projectPath, ""); //$NON-NLS-1$
-	env.addPackageFragmentRoot(projectPath, "src"); //$NON-NLS-1$
+	IPath root = env.addPackageFragmentRoot(projectPath, "src"); //$NON-NLS-1$
 	IPath bin = env.setOutputFolder(projectPath, "bin"); //$NON-NLS-1$
-
+	env.addExternalJars(projectPath, Util.getJavaClassLibs());
+	env.addClass(root, "q", "Y", """
+			package q;
+			public class Y {}
+			""");
 	fullBuild();
 	expectingNoProblems();
+	IPath expectedClassFilePath = bin.append("q").append("Y.class");
+	expectingPresenceOf(new IPath[]{ expectedClassFilePath });
 
 	env.removeFolder(bin);
+	expectingNoPresenceOf(new IPath[]{ expectedClassFilePath });
 
 	incrementalBuild();
 	expectingNoProblems();
 	expectingPresenceOf(bin); // check that bin folder was recreated and is marked as derived
-	if (!env.getProject(projectPath).getFolder("bin").isDerived())
+	if (!env.getProject(projectPath).getFolder("bin").isDerived()) {
 		fail("output folder is not derived");
+	}
+	expectingNoPresenceOf(new IPath[]{ expectedClassFilePath });
+
+	// now enable the (disabled by default) preference and try again
+	p.setOption(JavaCore.CORE_JAVA_BUILD_RECREATE_MODIFIED_CLASS_FILES_IN_OUTPUT_FOLDER, JavaCore.ENABLED);
+
+	env.removeFolder(bin);
+	expectingNoPresenceOf(new IPath[]{ expectedClassFilePath });
+
+	incrementalBuild();
+	expectingNoProblems();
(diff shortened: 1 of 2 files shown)


iloveeclipse:
While looking at the test results I've stumbled over old opened bug https://bugs.eclipse.org/bugs/show_bug.cgi?id=551526 which should be fixed now (hopefully).

stephan-herrmann:
> While looking at the test results I've stumbled over old opened bug https://bugs.eclipse.org/bugs/show_bug.cgi?id=551526 which should be fixed now (hopefully).

Internal Server Error

Do you have an excerpt from that bug to quote here?

iloveeclipse:
> Do you have an excerpt from that bug to quote here?

The fun is, the bug was reported by me internally / Simeon on eclipse.org 5 years ago because of very similar issue we saw in our old codebase. Steps to reproduce were very similar, with the exception that the setup of the one original project was very awkward and it wasn't fully clear what of many "special" configuration options were responsible. At that time Simeon found & fixed some of the build delta bugs (see `org.eclipse.jdt.core.tests.builder.Bug564905Test` & 03c0989a5bc07e20793e344c6ad788f76b8de370) and we weren't able to reproduce the problem anymore.

However that was a single, very "special" project with no dependencies and not ~20 "usual" interdependent projects as we observe now. Looking on the current issue I've seen that most of trouble we observe is caused by the fact that if the output folders were fully deleted externally (git clean -fdx), incremental build only recreates the folders but doesn't do anything else useful afterwards, which results in hundreds of compilation problems in dependent code complaining about missing types.

iloveeclipse:
@stephan-herrmann : if you want review, you can, as the build is green now. I plan to submit this for M2.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/JavaBuilder.java
```diff
@@ -878,6 +885,55 @@ private void recordNewState(State state) {
...
+			IPath outputLocation = entry.getOutputLocation();
+			if (outputLocation != null) {
+				IContainer outputContainer;
+				if(outputLocation.segmentCount() == 1) {
+					outputContainer = project.getProject();
+				} else {
+					outputContainer = root.getFolder(outputLocation);
+				}
+				if (!existsOnDisk(outputContainer)) {
+					return FULL_BUILD;
```
trancexpress, line 916: Shouldn't this just return a boolean? And the method can be called `hasMissingOutputFolders` or similar.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/JavaBuilder.java
```diff
@@ -878,6 +885,55 @@ private void recordNewState(State state) {
...
 
+/**
+ * Checks whether all output folders for the given project exist on disk.
+ *
+ * @param project
+ *                    The Java project to check
+ * @return If any output folder is missing, {@link IncrementalProjectBuilder#FULL_BUILD} is returned and original
+ *         <code>buildKind</code> argument otherwise.
+ */
+protected int checkOutputFolders(IJavaProject project, int buildKind) {
```
trancexpress, line 896: Do you expect someone to override this? From my POV it can be `private static`?

org.eclipse.jdt.core.tests.builder/src/org/eclipse/jdt/core/tests/builder/BuildpathTests.java
```diff
@@ -1018,24 +1018,52 @@ public void test0100() throws JavaModelException {
...
+	incrementalBuild();
+	expectingNoProblems();
+	expectingPresenceOf(bin); // check that bin folder was recreated and is marked as derived
+	if (!env.getProject(projectPath).getFolder("bin").isDerived()) {
+		fail("output folder is not derived");
+	}
+	expectingNoPresenceOf(new IPath[]{ expectedClassFilePath });
+
+	// now enable the (disabled by default) preference and try again
+	p.setOption(JavaCore.CORE_JAVA_BUILD_RECREATE_MODIFIED_CLASS_FILES_IN_OUTPUT_FOLDER, JavaCore.ENABLED);
```
trancexpress, line 1050: IMO this belongs to a new test.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/JavaBuilder.java
iloveeclipse: Yes :-)
If the someone would like to trigger or not to trigger full build for some reason without touching the preference

org.eclipse.jdt.core.tests.builder/src/org/eclipse/jdt/core/tests/builder/BuildpathTests.java
iloveeclipse: I found it very much the same test. Bug https://bugs.eclipse.org/bugs/show_bug.cgi?id=143025 was about recreating output folder too, just they didn't had chance to observe the bug we observed.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/builder/JavaBuilder.java
iloveeclipse: Similar to the reason this is protected method, I wanted to allow the build flags to be modified, not only yes/no.

Issue #4476, opened 2025-10-08
JEP Draft: Primitive Types in Patterns, instanceof, and switch (Fourth Preview)


mpalat:
Ref: https://openjdk.org/jeps/8359136

stephan-herrmann:
Changes are being drafted in https://bugs.openjdk.org/browse/JDK-8369153 as announced also in https://mail.openjdk.org/pipermail/amber-spec-experts/2025-October/004300.html

In the same context a refactoring of JLS chap. 5 was proposed in https://mail.openjdk.org/pipermail/amber-spec-experts/2025-October/004302.html ,  document available at https://cr.openjdk.org/~abimpoudis/posts/amber-design-chapter5.html

mpalat:
> document available at https://cr.openjdk.org/~abimpoudis/posts/amber-design-chapter5.html

@stephan-herrmann: in the above, sub-section 2 is missing under section dealing with [Part 2 - Enhancements to Language ](https://cr.openjdk.org/~abimpoudis/posts/amber-design-chapter5.html#type-conversions-static-planning-and-runtime-execution) - Not sure whether it's a typo or whether an entire sub-section is omitted by mistake. Should we raise it in amber or maybe a direct mail to the author is more appropriate?

stephan-herrmann:
I'm late to find that this paragraph is new in JLS 5.7.2:
> A value-based unconditionally exact conversion from a type S to a type T, for a value s, is one which the input value is a constant expression of type S ([15.29](https://docs.oracle.com/javase/specs/jls/se25/html/jls-15.html#jls-15.29)), the testing conversion for the value s is exact and the testing conversion is one of the following:
> * a narrowing primitive conversion
> * a widening and narrowing primitive conversion
> * a widening primitive conversion that is not exact

Application of this follows in a new item of dominance:
> A case label with a case constant c is dominated if one of the following holds:
> * ...
> * c is a constant expression of a primitive type S, and there is a preceding case label in the switch block with an unguarded case pattern p of type T, where the conversion from S to T for the value c is a value-based unconditionally exact conversion ([5.7.2](https://cr.openjdk.org/~abimpoudis/instanceof/jep530-20251112/specs/primitive-types-in-patterns-instanceof-switch-jls.html#jls-5.7.2)).

[...]

PR #4748, opened 2026-01-15
Updates for JEP 530

stephan-herrmann:
+ WIP Implement check for value-based unconditionally exact conversion
  + apply it in dominance checks
+ implement: type pattern unconditional on the switch type is exhaustive
+ several tests from the new spec

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4476


stephan-herrmann:
I made a full scan of textual spec differences between 
* https://cr.openjdk.org/~abimpoudis/instanceof/jep507-20250429/specs/primitive-types-in-patterns-instanceof-switch-jls.html
* https://cr.openjdk.org/~abimpoudis/instanceof/jep530-20251112/specs/primitive-types-in-patterns-instanceof-switch-jls.html

The PR is still incomplete in that method `Constant.isExactTestingConversion(BaseTypeBinding)` needs to cover more combinations and also tests need to cover all these combinations.

Issue #4767, opened 2026-01-19
Unnecessary error marker for **nested** record's constructor that uses varargs


hjohn:
Similar to issue #342 

When a record with a varargs parameter is nested inside another class, calling its constructor from another class generates an error in the editor, but no marker in the problem view.

Given two classes:

```java
package test;

public class Outer {
  public record NestedRecord(String key, Object... args) {}
}
```

```java
package test;

public class Two {
  public static void main(String[] args) {
    new Outer.NestedRecord("bla");   // <---- error here in editor, but no marker in problem view
  }
}
```

# Expected Behavior

No error in the editor.

# Work-arounds

- Pass an empty `Object[]` as 2nd parameter
- Don't nest the record

srikanth-sankaran:
Note: The PR https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4771 backs out the fix for https://github.com/eclipse-jdt/eclipse.jdt.core/issues/342 and comprehensively fixes the problem for both top level types and nested types. Basically the problem is that the varargs'ness of the record component is not persisted in the model (and cannot be reconstructed as a result)

PR #4771, opened 2026-01-20
Reconciler error with nested record's constructor that uses varargs

srikanth-sankaran:
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4767


srikanth-sankaran:
Note: This PR backs out the fix for https://github.com/eclipse-jdt/eclipse.jdt.core/issues/342 and comprehensively fixes the problems for both top level types and nested types. Basically the problem is that the varargs'ness of the record component is not persisted in the model (and cannot be reconstructed as a result)

PR #4776, opened 2026-01-22
Problem with calling super in overridden default method

stephan-herrmann:
Improve the fix from #4727:
- treat default methods like abstract (1 location)
- treat default methods like concrete (1 location)

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1600


stephan-herrmann:
@srikanth-sankaran this change occurs in direct neighborhood of your 039347d7e4ee8e1deb95513f2bd485f15185146e.

You (2014):
```diff
					if (!original.isAbstract()) {
-						if (original2.isAbstract())
+						if (original2.isAbstract() || original2.isDefaultMethod()) 
```

I (2026):
```diff
-					if (!original.isAbstract()) {
+					if (!original.isAbstract() && !original.isDefaultMethod()) {
						if (original2.isAbstract() || original2.isDefaultMethod()) 
```

 It actually makes your change more symmetric, in that we treat default methods like abstract methods. Do you see any problem with that?

Unfortunately, the sorting from #4727 needs a different strategy, here we need to prioritize implemented methods (i.e., default methods are like concrete).

If you agree, then we could actually fix #1600 with no known (!) regression - in a minimal way, i.e.

I still believe that `msmb()` should not see some of the method it sees - but that might be something for 4.40.

PR #4807, opened 2026-01-29
Refactor AbortCompilation to allow non compiler code see the root cause

iloveeclipse:
After 0dbe2967235238fb5375924d17a13bcd9f9f6483 &
15c8bfb5a86e8978a858bbd8233068e2ad8e0e7f JDT UI may report errors where none should be reported because it is supposed to be "silent cancellation".

This change uses standard Java "cause" in the `AbortCompilation` and removes two different public fields used to keep exceptions before.

An extra change will be needed in at least one place in the UI, where `org.eclipse.jdt.core.manipulation.CoreASTProvider.createAST(ITypeRoot, IProgressMonitor)` should be able to inspect the provided "cause" of the `RuntimeException` and handle it in same way as
`OperationCanceledException`.

Note: `AbortCompilation` itself is not visible outside compiler code, so this type can't be used in UI for any checks.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2769

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/Compiler.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/Compiler.java
@@ -626,11 +626,16 @@ public class Compiler implements ITypeRequestor, ProblemSeverities {
 		CompilationUnitDeclaration unit) {
 
 		/* special treatment for SilentAbort: silently cancelling the compilation process */
+		Throwable cause = abortException.getCause();
 		if (abortException.isSilent) {
-			if (abortException.silentException == null) {
+			if (cause == null) {
 				return;
 			}
-			throw abortException.silentException;
+			if(cause instanceof RuntimeException) {
+				throw (RuntimeException) cause;
+			} else {
+				throw new RuntimeException(cause);
+			}
 		}
 
 		/* uncomment following line to see where the abort came from */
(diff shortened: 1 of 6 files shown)


iloveeclipse:
@stephan-herrmann : since you were involved in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4302 change, could you please check this one? 

The UI part needed for the fix is https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2770.

stephan-herrmann:
> @stephan-herrmann : since you were involved in #4302 change, could you please check this one?
> 
> The UI part needed for the fix is [eclipse-jdt/eclipse.jdt.ui#2770](https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2770).

Connecting the dots for posterity:
* I [commented](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2769#issuecomment-3818115171) on the issue https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2769
* @iloveeclipse  then created https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4808
   * I gave a basic [lgtm](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4808#issuecomment-3819504479) with just one follow-up question

In the end, this current PR will probably be closed without merging.

iloveeclipse:
Let close this PR for now, however I highly dislike code that exposes public instance fields...

https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2769 should be fixed by https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4808.

PR #4451, opened 2025-09-30
Do not throw RuntimeException in Parser::parse #4302

fedejeanne:
See https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4302#issuecomment-3348205396

If this PR gets merged then this one can also be merged:
- https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2524

PR #4808, opened 2026-01-29
Don't report AbortCompilation exceptions from parser

iloveeclipse:
Regression from
https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4451 & https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4302.

If Java Editor interrupts hover computation, nothing should be reported, as it is expected. After changes above two errors are reported into the log (with same stack).

Fixes https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2769

### DIFF
--- a/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolver.java
+++ b/org.eclipse.jdt.core/dom/org/eclipse/jdt/core/dom/CompilationUnitResolver.java
@@ -1326,6 +1326,14 @@ class CompilationUnitResolver extends Compiler {
 			return unit;
 		} catch (AbortCompilation e) {
 			this.handleInternalException(e, unit);
+			if(e.isSilent) {
+				if (e.silentException == null) {
+					if (this.monitor != null) {
+						this.monitor.setCanceled(true);
+					}
+					throw new OperationCanceledException();
+				}
+			}
 			return unit == null ? this.unitsToProcess[0] : unit;
 		} catch (Error | RuntimeException e) {
 			this.handleInternalException(e, unit, null);


stephan-herrmann:
This looks well aligned with existing stuff!

By debugging tests like org.eclipse.jdt.core.tests.dom.ASTConverterTestAST3_2.test0521(), I could observe that the very location of your fix may already throw OperationCanceledException, which is wrapped in AbortCompilation in this call stack:
```
AbortCompilation.<init>(boolean, RuntimeException) line: 57	
CancelableNameEnvironment.checkCanceled() line: 40	
CancelableNameEnvironment.findType(char[], char[][]) line: 52	
LookupEnvironment.askForType(PackageBinding, char[], ModuleBinding) line: 386	
PlainPackageBinding(PackageBinding).getTypeOrPackage(char[], ModuleBinding, boolean) line: 277	
CompilationUnitScope.getDefaultImports() line: 761	
CompilationUnitScope.checkAndSetImports() line: 234	
LookupEnvironment$CompleteTypeBindingsSteps.perform(CompilationUnitScope) line: 220	
LookupEnvironment.completeTypeBindings() line: 612	
CompilationUnitResolver(Compiler).internalBeginToCompile(ICompilationUnit[], int) line: 787	
CompilationUnitResolver(Compiler).beginToCompile(ICompilationUnit[]) line: 277	
CompilationUnitResolver.resolve(CompilationUnitDeclaration, ICompilationUnit, NodeSearcher, boolean, boolean, boolean) line: 1240	
CompilationUnitResolver.resolve(ICompilationUnit, IJavaProject, List, NodeSearcher, Map, WorkingCopyOwner, int, IProgressMonitor) line: 830	
CompilationUnitResolver.toCompilationUnit(ICompilationUnit, boolean, IJavaProject, List<Classpath>, NodeSearcher, int, Map<String,String>, WorkingCopyOwner, WorkingCopyOwner, int, IProgressMonitor) line: 1405	
CompilationUnitResolver$ECJCompilationUnitResolver.toCompilationUnit(ICompilationUnit, boolean, IJavaProject, List<Classpath>, int, int, Map<String,String>, WorkingCopyOwner, WorkingCopyOwner, int, IProgressMonitor) line: 109	
ASTParser.internalCreateASTCached(IProgressMonitor) line: 1410	
ASTParser.lambda$1(IProgressMonitor) line: 1289	
Lambda.call() line: not available	
JavaModelManager.cacheZipFiles(JavaCallable<T,E>) line: 5709	
ASTParser.internalCreateAST(IProgressMonitor) line: 1289	
ASTParser.createAST(IProgressMonitor) line: 933	
ASTConverterTestAST3_2.test0521() line: 3491	
```
So adding another scenario to the same policy looks safe.

And in the scenario from https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2769 this OperationCanceledException will be explicitly caught from the try-catch surrounding:
> 	at org.eclipse.jdt.core.manipulation.CoreASTProvider$1.run(CoreASTProvider.java:294)

While looking left and right at other silent AbortCompilation, I wondered if this guy should be improved, too: org.eclipse.jdt.internal.compiler.ProcessTaskManager.removeNextUnits() throws AbortCompilation like this:
```java
			} catch (InterruptedException interrupt) {
				throw new AbortCompilation(true/* silent */, new RuntimeException(interrupt));
			}
```
This looks similar to the Parser use case, doesn't it? Could we think of a situation where this `new RuntimeException(interrupt)` could be useful? Or should we change this to `null` like in the Parser case? I see you already had a discussion concerning this very line in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2948#discussion_r1763140931 :smile: 

I'll leave it to you, if you want to add a change for ProcessTaskManager. Otherwise the proposed change already looks good to me.

iloveeclipse:
> While looking left and right at other silent AbortCompilation, I wondered if this guy should be improved, too: org.eclipse.jdt.internal.compiler.ProcessTaskManager.removeNextUnits() throws AbortCompilation like this:
> 
> ```java
> 			} catch (InterruptedException interrupt) {
> 				throw new AbortCompilation(true/* silent */, new RuntimeException(interrupt));
> 			}
> ```
> 
> This looks similar to the Parser use case, doesn't it? Could we think of a situation where this `new RuntimeException(interrupt)` could be useful? Or should we change this to `null` like in the Parser case? I see you already had a discussion concerning this very line in [#2948 (comment)](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/2948#discussion_r1763140931) 😄

This opens another can of worms.
If I see it right, the AbortCompilation from that code is re-thrown here:

https://github.com/eclipse-jdt/eclipse.jdt.core/blob/b52cd52c0720bad9e39c50505c0c8f44d01e0f9f/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/Compiler.java#L491-L495

immediately handled here in same method:

https://github.com/eclipse-jdt/eclipse.jdt.core/blob/b52cd52c0720bad9e39c50505c0c8f44d01e0f9f/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/Compiler.java#L527-L529

and the wrapped `RuntimeException` is then re-thrown here:

https://github.com/eclipse-jdt/eclipse.jdt.core/blob/b52cd52c0720bad9e39c50505c0c8f44d01e0f9f/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/Compiler.java#L624-L633

... and then it looks like no one handles this exception properly, we have lot of callers here:

image

iloveeclipse:
Follow up: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4809

Issue #4810, opened 2026-01-31
ECJ vs Javac: Regression in Generic Type Inference


flori-schwa:
Hello, the following Code is accepted by javac (Version `21.0.8`) and ECJ Version `3.44.0.v20251118-1623` but rejected by ECJ Version `3.45.0.v20260122-1615` with the Message `Type MyCallable<Object,Exception> inferred for MyCallable<>, is not valid for an anonymous class with '<>'` at the anonymous class creation (`new MyCallable<>()`)

```java
package repro;

import java.util.concurrent.Callable;

public class Repro {

  Object test() {
    try {
      return myMethod(new MyCallable<>() {
        @Override
        public Object call() {
          return new Object();
        }
      });
    } catch (Exception e) {
      e.printStackTrace();
      return null;
    }
  }

  <T, E extends Exception> T myMethod(MyCallable<T, E> callable) throws E {
    return callable.call();
  }

}

interface MyCallable<T, E extends Exception> extends Callable<T> {
  @Override
  T call() throws E;
}
```

<img width="1643" height="591" alt="Image" src="https://github.com/user-attachments/assets/a4c54ef2-4c36-4473-b50b-6b780bfb5667" />

<img width="1546" height="597" alt="Image" src="https://github.com/user-attachments/assets/cd025c06-b565-4275-a699-d1500b4fdd27" />

<img width="1019" height="166" alt="Image" src="https://github.com/user-attachments/assets/e03dea86-aff0-412d-8662-b830e667edc0" />

srikanth-sankaran:
Reproduced on master, Stephan, thanks for taking a look

fedejeanne:
If it helps, the range is smaller: the issue was introduced in **M2**.

This is taken from _Help > About > Features_:

- **Not** reproducible in M**1**: `Eclipse.org	Eclipse Java Development Tools	3.20.500.v20260103-1800	org.eclipse.jdt`
- Reproducible in: `Eclipse.org	Eclipse Java Development Tools	3.20.500.v20260118-1800	org.eclipse.jdt`

stephan-herrmann:
The first hurdle in understanding the problem was the fact, that the error message did not reveal at all which error the compiler was (wrongly) seeing:
* `Type MyCallable<Object,Exception> inferred for MyCallable<>, is not valid for an anonymous class with '<>'`

[...]

PR #4814, opened 2026-02-02
ECJ vs Javac: Regression in Generic Type Inference

stephan-herrmann:
Pragmatically handle equivalent exception ivars:
+ if one resolves to RuntimeException by special rule immediately incorporate to be able to propagate to equivalent ivars

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4810

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
@@ -1174,12 +1174,24 @@ public class InferenceContext18 {
 				final int numVars = variableSet.size();
 				if (numVars > 0) {
 					final InferenceVariable[] variables = variableSet.toArray(new InferenceVariable[numVars]);
+					// NON-JLS: prioritize ivars in 'inThrows' as those may pull in new information by extra rule below (=RuntimeException)
+					BoundSet tSet = tmpBoundSet;
+					Arrays.sort(variables, (v1, v2) -> {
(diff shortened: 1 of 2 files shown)


stephan-herrmann:
Inference failed in its first attempt of 18.4. Resolution. This happened with type bounds like this:
```
E#1 <: Exception
E#1 = E#2
E#2 <: Exception
throws E#1
```
Now the following rule applied to `E#1` but not `E#2`:
>  if the bound set contains throws αi, and each proper upper bound of αi is a supertype of RuntimeException, then Ti = RuntimeException. 

So `E#1 = RuntimeException` but for `E#2` we only had `Exception` as the upper bound, causing incorporation to create the contradiction `RuntimeException = Exception`.

Then during the second attempt, both ivars are resolved to not-denotable fresh type variables (JLS: Y, ecj: Z#`n`). This not-denotable type is not allowed for the anonymous diamond.

The tentative solution goes in three steps:
1. when the above rule conjures up new information out of the hat, we immediately incorporate that type bound to ensure this information will be propagated to the equivalent ivar.
2. before resolving an individual ivar first check if it already has an instantiation. This can now happen if such instantiation was produced by the above incorporation.
3. to ensure the additional info doesn't come late, sort variables as to prioritize those that have an in-throws bound.

None of this is specified in JLS, but I hold that all these steps are OK by common sense:
1. JLS says: "_The bounds α1 = T1, ..., αn = Tn are incorporated with the current bound set_" - without being specific whether or not this happens in bulk or individually.
2. if an ivar already has an instantiation there is no use in deriving an inferior solution from lower or upper bounds
3. JLS does not speak about the order in which  `α1, ..., αn` are resolved, under the silent assumption that order should not matter. If a specific order allows us to infer a better result, why not do so?

Issue #4797, opened 2026-01-28
Type Hierarchy has duplicated entries (classes) and one of them is incomplete


fedejeanne:
Sometimes when I get the _Type Hierarchy_ of a given class, I get duplicated entries.

In the following screenshot you can see duplicated entries of `AbstractCustomDispatch`.

<img width="579" height="421" alt="Image" src="https://github.com/user-attachments/assets/65d49cd4-790e-450d-99ee-07786122f410" />

I copied the _fully qualified name_ of both classes and noticed that the entry on the top is inside a fictional package that has an additional `src.` right at the beginning. The _real_ package starts with `aq`.

## Additional info

1. `AbstractCustomDispatch` and **all of its subclasses** are ours _i.e._ they are present in the workspace as source code.
2. The subclasses are in different plugins

<img width="811" height="313" alt="Image" src="https://github.com/user-attachments/assets/1a54d05a-7b7a-42ee-8ae5-7c79de1c586f" />

---

<img width="805" height="770" alt="Image" src="https://github.com/user-attachments/assets/c9e9799f-ef7e-4330-a88f-4e4b2ffbf2a0" />

---

<img width="806" height="391" alt="Image" src="https://github.com/user-attachments/assets/9698deea-86de-49e9-a705-ca7cc701e295" />

---

3. The methods  of the `AbstractCustomDispatch` that is found under `src.aq...` (and the methods of its subclasses) are missing 👎  

<img width="622" height="395" alt="Image" src="https://github.com/user-attachments/assets/2096b558-2cef-43b3-8da7-e0c16324649d" />

4. The methods  of the `AbstractCustomDispatch` that is found under `aq...` (and the methods of its subclasses) are there 👍 

<img width="622" height="395" alt="Image" src="https://github.com/user-attachments/assets/94ecce7d-57b1-4d65-8235-82deef597c6e" />

I am not able to consistently reproduce the error but I just happen to run into it today and I started debugging.

## Current findings

My assumption is that the type `AbstractCustomDispatch` is still not completely discovered by the time its first 3 subclasses are being populated and it is completely known (and it's known to be in the workspace) by the time the rest of its subclasses are being computed.

I set a breakpoint in here

[...]

PR #4817, opened 2026-02-03
Print warning if adding parent folder of library to the classpath #4797

fedejeanne:
## What it does

Print a warning in the trace log if the Class-Path entry in the manifest file of a third-party library includes a parent folder. This might not be forbidden but it's a dangerous practice.

Contributes to: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4797

## How to test
- Launch the _Runtime Workspace_, making sure that _tracing_ is enabled like this:
 
image

- Import the 2 projects in this ZIP: [projects.zip](https://github.com/user-attachments/files/25047329/runtime-workspace.zip)
- Restart the application

### Expected
This should appear in the Console output:

```
| main | 2026-02-03 14:26:33.814 | org.eclipse.jdt.core | /debug | org.eclipse.jdt.internal.core.ClasspathEntry | resolvedChainedLibraries | 993 | WARN: Classpath entry ../ in manifest of jar file: \a\lib\badlib.jar references a parent folder. This may cause issues like: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4797 |
```

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/ClasspathEntry.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/ClasspathEntry.java
@@ -986,6 +986,14 @@ public class ClasspathEntry implements IClasspathEntry {
 						}
 						continue;
 					}
+					if (JavaModelManager.CP_RESOLVE_VERBOSE_FAILURE && calledJar.isPrefixOf(jarPath)) {
+						// Adding parent folders is a hack that some JARs use to spare some effort,
+						// but it may cause issues like having invalid duplicated classes in the classpath.
+						// See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4797
+						trace("WARN: Classpath entry " + calledFileName + " in manifest of jar file: " //$NON-NLS-1$ //$NON-NLS-2$
+								+ jarPath.toOSString()
+								+ " references a parent folder. This may cause issues like duplicated classes in the classpath, wrong hierarchies, etc. For more details, see https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4797."); //$NON-NLS-1$
+					}
 					resolvedChainedLibraries(calledJar, visited, result);
 					result.add(calledJar);
 				}


fedejeanne:
@iloveeclipse I know you said you are busy in the next few weeks but maybe you could still make some time to take a look at this PR? It is pretty small and it merely adds a warning when some 3rd-party library does something risky like adding a parent folder to the classpath (which, as you said yourself in https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4797#issuecomment-3842098885, is a PITA). That way we could at least identify such bad libraries and handle them better.

Thank you in advance!

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/ClasspathEntry.java
```diff
@@ -986,6 +986,14 @@ private static void resolvedChainedLibraries(IPath jarPath, HashSet visited, Arr
 						}
 						continue;
 					}
+					if (JavaModelManager.CP_RESOLVE_VERBOSE_FAILURE && calledJar.isPrefixOf(jarPath)) {
+						// Adding parent folders is a hack that some JARs use to spare some effort,
+						// but it may cause issues like having invalid duplicated classes in the classpath.
+						// See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4797
+						trace("WARN: Classpath entry " + calledFileName + " in manifest of jar file: " //$NON-NLS-1$ //$NON-NLS-2$
+								+ jarPath.toOSString()
+								+ " references a parent folder. This may cause issues like: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4797"); //$NON-NLS-1$
```
trancexpress: Please give a short example of the issue. IMO its fine to add issues in comments, in output I wouldn't.

fedejeanne: You mean I should change the output to something like:

```
[...] This may cause issues like duplicated classes in the classpath, wrong hierarchies, etc.
```

?

Actually, I find references to issues very helpful and I would like to keep it, if it's ok to you. How about a middle-ground like:

```
[...] This may cause issues like duplicated classes in the classpath, wrong hierarchies, etc. For more details, see https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4797.
```

?

fedejeanne: @trancexpress I changed it to the 2nd suggestion, which includes the link to the issue too. I hope that's ok.

fedejeanne:
Hi. Would it be possible to merge this for M1? We had yet another issue with 3rd-party libraries adding its parent folder to the classpath and it would be nice to be able to catch these errors sooner.

Thank you in advance!

PR #4819, opened 2026-02-04
Add generic type arguments to map

fedejeanne:
Just a nit.

The map is `protected` but the class is inside an `internal` package so it should be fine.


iloveeclipse:
Could you please change the commit message, amend the commit & force push?

fedejeanne:
Of course and thank you for the hint! I took your proposal for the commit-message :-)

Sorry for the late reply, I was working on the folding issue for JDT.

PR #4836, opened 2026-02-13
NPE on hyperlink request

srikanth-sankaran:
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4217


srikanth-sankaran:
@jarthana - Could you please review this for 4.39 RC1 ? TIA.

Basically, the problem is that since the lambda features as an argument of an unknown method, the second resolution of the lambda (after the target type is determined post overload resolution) does not happen. The lambda body is never traversed for the purpose of resolution and scope of the body block is null.

Fix is to recognize this condition and not descend into the lambda at all - since unresolved AST nodes are full of null references.
This won't somehow magically make selection work under such broken situations, but we won't NPE.

srikanth-sankaran:
Thanks Jay!

PR #4837, opened 2026-02-13
ECJ fails to complain about static member being selected from a parameterized class

### LINKED ISSUE #4702: ecj ignores error "cannot select a static class from a parameterized type"
Here's a minimal demo:

```java
public class OuterStaticNestedDemo<E> {
    class Outer {
        static class StaticNested {}
    }

    void qualifiedNew(Outer outer) {
        new Outer.StaticNested();
    }
}
```

When I compile with javac, it raises an error (using version 25 here, but it's the same for versions 11, 17 and 21) :

```
javac -g -source 25 -target 25 OuterStaticNestedDemo.java
OuterStaticNestedDemo.java:8: error: cannot select a static class from a parameterized type
        new Outer.StaticNested();
                 ^
1 error
```

When I compile with ecj, it compiles :

`java -jar ecj-3.44.0.jar -g -source 25 -target 25 OuterStaticNestedDemo.java`

Could you please fix to raise the same error as javac ?

srikanth-sankaran:
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4702

### DIFF
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericsRegressionTest_9.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericsRegressionTest_9.java
@@ -2034,6 +2034,113 @@ public void testGH4810() {
 	});
 }
 
+// https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4702
+// ecj ignores error "cannot select a static class from a parameterized type"
+public void testIssue4702() {
+	if (this.complianceLevel < ClassFileConstants.JDK16)
+		return;
+	runNegativeTest(new String[] {
+		"OuterStaticNestedDemo.java",
+		"""
+		public class OuterStaticNestedDemo<E> {
+		    class Outer {
+		        static class StaticNested {}
+		    }
+
+		    void qualifiedNew(Outer outer) {
+		        new OuterStaticNestedDemo<String>.Outer.StaticNested();
+		    }
+		}
+		"""
+	},
+		"----------\n" +
+		"1. ERROR in OuterStaticNestedDemo.java (at line 7)\n" +
+		"	new OuterStaticNestedDemo<String>.Outer.StaticNested();\n" +
+		"	    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n" +
+		"The member type OuterStaticNestedDemo.Outer.StaticNested cannot be qualified with a parameterized type, since it is static. Remove arguments from qualifying type OuterStaticNestedDemo<String>.Outer\n" +
+		"----------\n");
+}
+// https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4702
+// ecj ignores error "cannot select a static class from a parameterized type"
+public void testIssue4702_1() {
+	if (this.complianceLevel < ClassFileConstants.JDK16)
+		return;
+	runNegativeTest(new String[] {
+		"OuterStaticNestedDemo.java",
+		"""
+		public class OuterStaticNestedDemo<E> {
+		    class Outer {
+		        static class StaticNested {}
+		    }
+
+		    void qualifiedNew(Outer outer) {
+		        new Outer.StaticNested();
+		    }
+		}
+		"""
+	},
+		"----------\n" +
+		"1. ERROR in OuterStaticNestedDemo.java (at line 7)\n" +
+		"	new Outer.StaticNested();\n" +
+		"	    ^^^^^^^^^^^^^^^^^^\n" +
+		"The member type OuterStaticNestedDemo.Outer.StaticNested cannot be qualified with a parameterized type, since it is static. Remove arguments from qualifying type OuterStaticNestedDemo<E>.Outer\n" +
+		"----------\n");
+}
+// https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4702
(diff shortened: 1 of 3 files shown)


srikanth-sankaran:
@stephan-herrmann - Can you please review this for 4.39 RC1 ? TIA

I have organized it as two commits, the first one is merely a rename refactoring.

stephan-herrmann:
Pondering about this issue teleported me back into https://bugs.eclipse.org/bugs/show_bug.cgi?id=460491  -  refreshing memory of that old story took a bit longer than expected ...

I'll complete my review tomorrow.

srikanth-sankaran:
> I'll complete my review tomorrow.

Thanks.

Fix is two fold: 

- To be a parameterized type, a type _in and off itself_ need not have type arguments, if its enclosing instance is parameterized it becomes parameterized.
- Additionally, in QTR we fail to recognize a qualifying class as being a parameterized type under similar circumstances.

stephan-herrmann:
> To be a parameterized type, a type in and off itself need not have type arguments, if its enclosing instance is parameterized it becomes parameterized.

This part looks good. I see this area as dual to  https://bugs.eclipse.org/460491 : while the old bug ensured that _implicit_ references to static members of parameterized types do not resolve as parameterized (since no type variables are in scope), this current fix makes sure you cannot _explicitly_ request such an illegal type via a PQTR. :heavy_check_mark:

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericsRegressionTest_9.java
```diff
@@ -2034,6 +2034,113 @@ interface MyCallable<U, F extends Exception> extends Callable<U> {
...
+		        new Outer.StaticNested();
+		    }
+		}
+		"""
+	},
+		"----------\n" +
+		"1. ERROR in OuterStaticNestedDemo.java (at line 7)\n" +
+		"	new Outer.StaticNested();\n" +
+		"	    ^^^^^^^^^^^^^^^^^^\n" +
+		"The member type OuterStaticNestedDemo.Outer.StaticNested cannot be qualified with a parameterized type, since it is static. Remove arguments from qualifying type OuterStaticNestedDemo<E>.Outer\n" +
```
stephan-herrmann, line 2086: Here I don't understand what's going on. The programmer did not specify any type arguments, yet the compiler requires arguments to be removed? How could that code be corrected?

I realize that javac raises a similar error, but if there is any chance that this code is legal per JLS then rejecting this now sounds like a change too disruptive for RC1.

@srikanth-sankaran  have you checked this against JLS?

stephan-herrmann (commented):
After doubts regarding testIssue4702_1() I was about to suggest to only merge the changes in PQTR, but then I tried this:
```java
		public class Demo<E> {
		    class Outer {
		        static class StaticNested<X> {}
		    }
		    void qualifiedNew() {
		        new Outer.StaticNested<String>();
		    }
		}
```

accepted in 4.38, rejected with the change under review:
```
----------
1. ERROR in /tmp/Demo.java (at line 6)
        new Outer.StaticNested<String>();
            ^^^^^^^^^^^^^^^^^^
The member type Demo.Outer.StaticNested<X> cannot be qualified with a parameterized type, since it is static. Remove arguments from qualifying type Demo<E>.Outer
----------
1 problem (1 error)
```
Again, how to you remove an absent `<E>`?

----
I do like the move to make code blocks look more similar between PQTR and QTR.  If we'd invert the enclosing `if` in PQTR to align as `if (qualifyingType == null)` then we'd see even more clearly, that then-blocks are still different between both types. Is that difference intended? How motivated?

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericsRegressionTest_9.java
stephan-herrmann: For the general discussion I moved back to the issue, see
* https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4702#issuecomment-3914560955
* https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4702#issuecomment-3915363558

:)

srikanth-sankaran:
To address the questions raised during review I'll defer to this to 4.40 M1. 

Since this is not a case of valid code being rejected but of invalid code not being rejected it need not be rushed.

srikanth-sankaran:
> Again, how to you remove an absent `<E>`?

By saying `new Demo.Outer.StaticNested<String>();` ??

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericsRegressionTest_9.java
srikanth-sankaran: > The programmer did not specify any type arguments, yet the compiler requires arguments to be removed? How could that code be corrected?
> 
> I realize that javac raises a similar error

By saying `new OuterStaticNestedDemo.Outer.StaticNested();` ? 

Javac as well as the patched ECJ accept this.

srikanth-sankaran:
> If we'd invert the enclosing `if` in PQTR to align as `if (qualifyingType == null)` then we'd see even more clearly, that then-blocks are still different between both types. Is that difference intended? How motivated?

Given it is RC1, I tried to minimize the delta - so didn't look into it ...

PR #4839, opened 2026-02-13
Avoid duplicated read of file attributes in ExternalFoldersManager

HannesWell:
## What it does
Avoid duplicated read of file attributes in `ExternalFoldersManager` in case the tested path is a directory.
`Files.isRegularFile()` and `Files.isDirectory()` both read the attributes of the specified file/path and check if the corresponding flags for file or directory are set.
Actually it is sufficient to read the attributes once and only check the read attributes for both cases.

I haven't run any benchmarks, but since this method is called frequently (at least that's my impression) such simple optimization could already help a little bit.
In the worst case, it's not notably faster.

## Author checklist


iloveeclipse:
Patch is OK for RC1.

PR #4842, opened 2026-02-16
CompleteTypeBindingsSteps.INTEGRATE_ANNOTATIONS_IN_HIERARCHY violates TypeBinding's protocol by direct updation of fields in various places

### LINKED ISSUE #4823: CompleteTypeBindingsSteps.INTEGRATE_ANNOTATIONS_IN_HIERARCHY violates TypeBinding's protocol by direct updation of fields in various places
1. See `org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.updateSupertypesWithAnnotations(Map<ReferenceBinding, ReferenceBinding>)`:

Updates `this.binding.superclass` thusly: `this.binding.superclass = updateWithAnnotations(this.superclass, this.binding.superclass, outerUpdates, updates);`

`org.eclipse.jdt.internal.compiler.lookup.SourceTypeBinding.superclass` carries a comment: `// MUST NOT be modified directly, use setter !`

The problem with direct update is that annotated variants would not co-evolve!

To see why this is a problem, consider this code:

```
import java.lang.annotation.Target;

@Target(ElementType.TYPE_USE)
@interface TA {
	
}

class Y extends @TA X {}

public class X extends @TA Object {
	
} 
```

After X has had `updateSupertypesWithAnnotations` called on it, if you inspect Y's supertype `@TA X`'s superclass you will see `Object` not `@TA Object` 

If the update was done via a setter such as:

```
                 ReferenceBinding updatedSuperclass = updateWithAnnotations(this.superclass, this.binding.superclass, outerUpdates, updates);
		if (this.binding.superclass != updatedSuperclass) //$IDENTITY-COMPARISON$
			this.binding.setSuperClass(updatedSuperclass);
```

all annotated variants would co-evolve.

2. Same method updates super interfaces but this is not problematic as the array holding super interfaces is shared by annotated variants

3. `org.eclipse.jdt.internal.compiler.ast.TypeParameter.updateWithAnnotations(ClassScope)` exhibits similar problems, directly updating `firstBound` and `superclass` fields which should not be set directly

srikanth-sankaran:
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4823

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeDeclaration.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeDeclaration.java
@@ -1881,7 +1881,7 @@ public void updateSupertypesWithAnnotations(Map<ReferenceBinding,ReferenceBindin
 		}
 	}
 	if (this.superclass != null) {
-		this.binding.superclass = updateWithAnnotations(this.superclass, this.binding.superclass, outerUpdates, updates);
+		this.binding.setSuperClass(updateWithAnnotations(this.superclass, this.binding.superclass, outerUpdates, updates)); // don't assign directly to superclass!
 	}
 	if (this.superInterfaces != null) {
 		ReferenceBinding[] superIfcBindings = this.binding.superInterfaces;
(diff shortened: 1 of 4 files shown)


srikanth-sankaran:
In this PR I have inspected and sanitized the assignments to all 8 fields of `org.eclipse.jdt.internal.compiler.lookup.SourceTypeBinding` that are marked with the comment ` // MUST NOT be modified directly, use setter !`

The 3 fields of `TypeVariableBinding` that carry the same comment are better scrutinized by @stephan-herrmann 

So this PR should be seen as work in progress and taken to completion.

I don't think we need to bend over backwards writing tests for these changes. To future proof this code against regression by a future change, I have simply opted to call out in the code that direct assignment should not be done.

I think that plenty future proofing :)

iloveeclipse:
> sanitized the assignments to all 8 fields of `org.eclipse.jdt.internal.compiler.lookup.SourceTypeBinding` that are marked with the comment ` // MUST NOT be modified directly, use setter !`

Can we made the fields protected or private in 4.40? It is not API.
For now, I would also add **javadoc** style comments to them too saying that one should never modify them directly.

srikanth-sankaran:
> Can we made the fields protected or private in 4.40?

Thing is they should not be assigned directly by the declaring class itself, because important side effects are encoded inside the corresponding set method. (Unless the developer is absolutely sure those side effects are not desirable in some situation, that is). This intent is not capturable directly under the access protection model of the language. (although it would be step in the right direction - yes so may be we should/could)

Issue #258, opened 2022-07-26
compiler.batch.FileSystem.getClasspath() may return ClasspathJar for non existing classpath entry


iloveeclipse:
See https://github.com/eclipse-pde/eclipse.pde/pull/236#issuecomment-1195506020
```
java.nio.file.NoSuchFileException: eclipse.pde/ui/org.eclipse.pde.core/target/ant_tasks/pde-ant.jar-classes
	at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:106)
	at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
	at java.base/sun.nio.fs.UnixFileAttributeViews$Basic.readAttributes(UnixFileAttributeViews.java:55)
	at java.base/sun.nio.fs.UnixFileSystemProvider.readAttributes(UnixFileSystemProvider.java:149)
	at java.base/sun.nio.fs.LinuxFileSystemProvider.readAttributes(LinuxFileSystemProvider.java:99)
	at java.base/java.nio.file.Files.readAttributes(Files.java:1851)
	at java.base/java.util.zip.ZipFile$Source.get(ZipFile.java:1264)
	at java.base/java.util.zip.ZipFile$CleanableResource.<init>(ZipFile.java:709)
	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:243)
	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:172)
	at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:186)
	at org.eclipse.jdt.internal.compiler.batch.ClasspathJar.initialize(ClasspathJar.java:201)
	at org.eclipse.jdt.internal.compiler.batch.ClasspathMultiReleaseJar.initialize(ClasspathMultiReleaseJar.java:38)
	at org.eclipse.jdt.internal.compiler.batch.FileSystem.<init>(FileSystem.java:235)
	at org.eclipse.jdt.internal.compiler.batch.Main.getLibraryAccess(Main.java:3495)
	at org.eclipse.jdt.internal.compiler.batch.Main.performCompilation(Main.java:4741)
	at org.eclipse.jdt.internal.compiler.batch.Main.compile(Main.java:1802)
```

[...]

PR #4843, opened 2026-02-16
[WIP] Allow IDE enable/disable escaping classpath entries in jars

iloveeclipse:
Since b74a7b82d9221b86893a88ea5453e92534ffe51b JDT Core (but not ecj!) supported Class-Path attribute for external jars and added all entries from such jar to the current project classpath.

This change introduces new JavaCore preference that allows JDT ignore "escaping" classpath entries read from external jars (entries, containing ".." path segment).

By default, JDT disables now old behavior and disallows such paths.

See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/258 See https://bugs.eclipse.org/bugs/show_bug.cgi?id=198572

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
@@ -2716,6 +2716,34 @@ public final class JavaCore extends Plugin {
 	public static final String CORE_MAIN_ONLY_PROJECT_HAS_TEST_ONLY_DEPENDENCY = PLUGIN_ID + ".classpath.mainOnlyProjectHasTestOnlyDependency";  //$NON-NLS-1$
 
 	/**
+	 * Core option ID: Enable escaping classpath entries in jar manifests (like ../bad.jar).
+	 * <p>
+	 * When enabled, the classpath entries in jar manifests can escape current jar directory tree (like
+	 * ../lib/some.jar). When disabled, all classpath entries in manifest must be inside current jar directory tree.
+	 * </p>
+	 * <p>
+	 * This option should be disabled to be consistent with command line compiler, but it is enabled by default for
+	 * backward compatibility reasons (in the IDE escaping classpath entries was possible since Eclipse 3.5, see
+	 * https://bugs.eclipse.org/bugs/show_bug.cgi?id=198572).
+	 * </p>
+	 * <p>
+	 * For performance reasons, any presence of the &quot;..&quot; segment in a classpath entry in a jar manifest will
+	 * cause the compiler to assume that the entry is escaping the current jar directory tree.
+	 * </p>
+	 * <dl>
+	 * <dt>Option id:</dt>
+	 * <dd><code>"org.eclipse.jdt.core.classpath.enableEscapingCpEntriesInJarManifest"</code></dd>
+	 * <dt>Possible values:</dt>
+	 * <dd><code>{ "enabled", "disabled" }</code></dd>
+	 * <dt>Default:</dt>
+	 * <dd><code>"disabled"</code></dd>
+	 * </dl>
+	 *
+	 * @since 3.45
+	 */
+	public static final String CORE_ENABLE_ESACAPING_CP_ENTRIES_IN_JAR_MANIFEST = PLUGIN_ID + ".classpath.enableEscapingCpEntriesInJarManifest";  //$NON-NLS-1$
+
+	/**
 	 * Compiler option ID: Enabling support for preview language features.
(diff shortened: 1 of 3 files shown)


iloveeclipse:
### DO NOT MERGE, it is a draft PR.

The only reason it is not "draft" is this: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4844 - I want see test results.

org.eclipse.jdt.core/model/org/eclipse/jdt/core/JavaCore.java
```diff
@@ -2715,6 +2715,34 @@ public final class JavaCore extends Plugin {
...
+	 * For performance reasons, any presence of the &quot;..&quot; segment in a classpath entry in a jar manifest will
+	 * cause the compiler to assume that the entry is escaping the current jar directory tree.
+	 * </p>
+	 * <dl>
+	 * <dt>Option id:</dt>
+	 * <dd><code>"org.eclipse.jdt.core.classpath.enableEscapingCpEntriesInJarManifest"</code></dd>
+	 * <dt>Possible values:</dt>
+	 * <dd><code>{ "enabled", "disabled" }</code></dd>
+	 * <dt>Default:</dt>
+	 * <dd><code>"disabled"</code></dd>
```
fedejeanne, line 2739: The JavaDoc is confusing:

> This option should be disabled to be consistent with command line compiler, **but it is enabled by default** for backward compatibility reasons ...

I see that it is `DISABLED` in `JavaCorePreferenceInitializer`.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/ClasspathEntry.java
```diff
@@ -978,6 +978,16 @@ private static void resolvedChainedLibraries(IPath jarPath, HashSet visited, Arr
 						trace("Invalid Class-Path entry " + calledFileName + " in manifest of jar file: " + jarPath.toOSString()); //$NON-NLS-1$ //$NON-NLS-2$
 					}
 				} else {
+					String escapePref = JavaCore.getOptions()
+							.get(JavaCore.CORE_ENABLE_ESACAPING_CP_ENTRIES_IN_JAR_MANIFEST);
+					if (JavaCore.DISABLED.equals(escapePref) && calledFileName.indexOf(DOT_DOT) != -1
+							&& hasDotDot(Path.fromPortableString(calledFileName))) {
+						if (JavaModelManager.CP_RESOLVE_VERBOSE_FAILURE) {
+							trace("Invalid (escaping jar directory) Class-Path entry " + calledFileName //$NON-NLS-1$
```
fedejeanne, line 986: NIT: to be able to search the log files more easily (a search for `Invalid Class-Path` would yield all problems)


```suggestion
							trace("Invalid Class-Path entry (escaping jar directory)" + calledFileName //$NON-NLS-1$
```

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/ClasspathEntry.java
```diff
@@ -978,6 +978,16 @@ private static void resolvedChainedLibraries(IPath jarPath, HashSet visited, Arr
 						trace("Invalid Class-Path entry " + calledFileName + " in manifest of jar file: " + jarPath.toOSString()); //$NON-NLS-1$ //$NON-NLS-2$
 					}
 				} else {
+					String escapePref = JavaCore.getOptions()
+							.get(JavaCore.CORE_ENABLE_ESACAPING_CP_ENTRIES_IN_JAR_MANIFEST);
+					if (JavaCore.DISABLED.equals(escapePref) && calledFileName.indexOf(DOT_DOT) != -1
+							&& hasDotDot(Path.fromPortableString(calledFileName))) {
```
fedejeanne, line 984: Question: the JavaDoc of `Path.fromPortableString(String)` mentions:

1.
> The path string must have been produced by a previous call to <code>IPath.toPortableString</code>.

... and also 

2.
> Instead of calling this method it is recommended to call <code>IPath.fromPortableString(String)</code> instead

w.r.t. **1)** I couldn't find any calls to `toPortableString` in the generation, but I assume it is safe to use this method anyway?

w.r.t. to **2)** The code already calls `IPath.fromPortableString(...)` 

Image

so I would simply skip the indirection and call it directly:


```suggestion
							&& hasDotDot(IPath.fromPortableString(calledFileName))) {
```

fedejeanne (commented):
Thank you very much for looking into this, @iloveeclipse !

I just have some minor comments.

Also, how do I test this?

PR #4847, opened 2026-02-17
Type Inference issue in spring-boot project

### LINKED ISSUE #4846: Type Inference issue in spring-boot project
The following code is a simplified class from the spring-boot project - https://github.com/spring-projects/spring-boot/blob/main/core/spring-boot/src/main/java/org/springframework/boot/logging/structured/ContextPairs.java
JDT incorrectly displays four errors on https://github.com/spring-projects/spring-boot/blob/25ed44368a47012cf9d753e97a8578409d999671/core/spring-boot/src/main/java/org/springframework/boot/logging/structured/ContextPairs.java#L141

```
import java.util.Map;

public class Pairs<T> {

	public <V> void addMapEntries(Function<T, Map<String, V>> extractor) {
		add(extractor.andThen(Map::entrySet), Map.Entry::getKey, Map.Entry::getValue);
		// add(extractor.andThen((Map<String, V> map) -> map.entrySet()), Map.Entry::getKey, Map.Entry::getValue);
	}
 
	public <E> void add(Function<T, Iterable<E>> elementsExtractor, PairExtractor<E> pairExtractor) {
		add(elementsExtractor, pairExtractor::getName, pairExtractor::getValue);
	} 

	public <E, V> void add(Function<T, Iterable<E>> elementsExtractor, Function<E, String> nameExtractor,
			Function<E, V> valueExtractor) {
	}
	
	interface PairExtractor<E> {
		<N> N getName(E element);
		<V> V getValue(E element);
		static <T> PairExtractor<T> of(Function<T, ?> nameExtractor, Function<T, ?> valueExtractor) {
			return new PairExtractor<>() {

				@Override
				@SuppressWarnings("unchecked")
				public <N> N getName(T instance) {
					return (N) nameExtractor.apply(instance);
				}

				@Override
				@SuppressWarnings("unchecked")
				public <V> V getValue(T instance) {
					return (V) valueExtractor.apply(instance);
				}

			};
		}
	}

}
```

The issue can’t be reproduced in Eclipse 4.38. Looks like that it has been introduced with https://github.com/eclipse-jdt/eclipse.jdt.core/commit/22dc3e1acc5b23b878945d6d5c5e200935c83edd

snjeza:
Fixes #4846 

~~I have changed `GenericsRegressionTest_1_8.testBug427479b` and `NullTypeAnnotationTest.test443467()`. In these tests, JDT with the PR shows two errors like javac does.~~

cc @stephan-herrmann @rgrunber

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
@@ -1529,12 +1529,6 @@ public class InferenceContext18 {
 	Map<InferenceVariable,Set<InferenceVariable>> collectDependencies(BoundSet bounds) {
 		// Implements the definition of dependencies from JLS §18.4:
 		Map<InferenceVariable,Set<InferenceVariable>> dependsOn = new LinkedHashMap<>();
-		// "An inference variable α depends on the resolution of itself."
-		for (InferenceVariable iv : this.inferenceVariables) {
-			Set<InferenceVariable> selfSet = new LinkedHashSet<>();
-			selfSet.add(iv);
-			dependsOn.put(iv, selfSet);
-		}
 		for (TypeBound typeBound : bounds.flatten()) {
 			// "Given a bound of one of the following forms:" (ecj may represent some using :> rather than <:)
 			// α = T
@@ -1616,6 +1610,15 @@ public class InferenceContext18 {
 				}
 			}
 		} while (hasChange);
+		// "An inference variable α depends on the resolution of itself."
+		// https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4846
+		for (InferenceVariable iv : this.inferenceVariables) {
+			if (!dependsOn.containsKey(iv)) {
+				Set<InferenceVariable> selfSet = new LinkedHashSet<>();
+				selfSet.add(iv);
+				dependsOn.put(iv, selfSet);
+			}
+		}
 		return dependsOn;
 	}
 
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
@@ -1529,6 +1529,12 @@ public class InferenceContext18 {
 	Map<InferenceVariable,Set<InferenceVariable>> collectDependencies(BoundSet bounds) {
 		// Implements the definition of dependencies from JLS §18.4:
 		Map<InferenceVariable,Set<InferenceVariable>> dependsOn = new LinkedHashMap<>();
+		// "An inference variable α depends on the resolution of itself."
+		for (InferenceVariable iv : this.inferenceVariables) {
+			Set<InferenceVariable> selfSet = new LinkedHashSet<>();
+			selfSet.add(iv);
+			dependsOn.put(iv, selfSet);
+		}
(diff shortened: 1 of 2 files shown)


stephan-herrmann:
Congratulations, @snjeza for digging deep into type inference and finding the location where things go wrong.

I will look into it during the next release cycle.

While test and the suggested fix are certainly helpful, I do have my doubts that we are allowed to simply ignore a constraint when it creates a contradiction with existing bounds. Let's see.

snjeza:
@stephan-herrmann I have updated the PR. Could you, please, review it again?

stephan-herrmann:
As @snjeza correctly identified, the problem relates to how we handle dependencies between inference variables.

After some detours in #4846 I came to identify method `IC18.canInfluenceAnyOf()` as the culprit. This method claims in a comment:
> // can influence == lives in the same component

Firstly, this was not correctly implemented: we only consulted the first component containing `in`. The fix is to search _all_ compontens containing `in`. What a tiny fix for a complex problem :)

Next, I was questioning the above claim, so I re-checked how JLS defines whether one ivar can influence another one, see [18.5.2.2](https://docs.oracle.com/javase/specs/jls/se26/html/jls-18.html#jls-18.5.2.2):
> An inference variable α can influence an inference variable β if α depends on the resolution of β ([§18.4](https://docs.oracle.com/javase/specs/jls/se26/html/jls-18.html#jls-18.4)), or vice versa; or if there exists a third inference variable γ such that α can influence γ and γ can influence β.

How does this related to our "components"?

In `IC18.collectDependencies()` we compute a map, that maps each ivar to all ivars on which it depends. Such dependencies include transitive dependencies as defined in 18.4. The above "components" are simply obtained as the `values()` of that map.

I was tempted to re-implement `canInfluenceAnyOf()` by direct queries to the original dependencies _Map_, which caused exactly one regression in all our tests. This pointed out some important properties of the existing implementation:
* by checking if two ivars exist in the same component we get the "_or vice versa_" part from the spec for free, as the set "components" has no directionality.
* additionally we also get the next part for free: "_or if there exists a third inference variable γ such that α can influence γ and γ can influence β_":
   * consider a component `[ R#0, S#1, T#2 ]` as the set of variables on which `R#0` depends.
   * ask the question whether `S#1` can influence `T#2`
   * answer: yes it can, because 
      * `S#1` can influence `R#0` (direct dependency)
      * `R#0` can influence `T#2` (inverse dependency).
   
Ergo: the tiny fix in this PR is all that is needed to correct `canInfluenceAnyOf()`.

stephan-herrmann:
@srikanth-sankaran @iloveeclipse @jarthana I'd like to propose this fix for RC1. It is a tiny change fixing a complex problem. And with the reasoning of the previous comment I'm quite confident that it is also correct :smile: 

I also compiled a full SDK workspace with these changes and found no errors, except for resolve errors in org.eclipse.swt.snippets.Snippet382, which look like real errors to me.

trancexpress:
@iloveeclipse is on vacation until early September. I'm not sure Andrey will answer.

stephan-herrmann:
> @iloveeclipse is on vacation until early September. I'm not sure Andrey will answer.

Thanks for letting me know. 
Let's see who else might answer to my proposal :)

stephan-herrmann:
@mpalat As you asked about RC1 candidates, perhaps you may want to weigh in on this issue, where I'd like to get the +1 of a project lead.

mpalat:
> @mpalat As you asked about RC1 candidates, perhaps you may want to weigh in on this issue, where I'd like to get the +1 of a project lead.

+1 as project lead. @stephan-herrmann  please merge for RC1. Thx!

srikanth-sankaran:
> Firstly, this was not correctly implemented: we only consulted the first component containing `in`. The fix is to search _all_ compontens containing `in`. What a tiny fix for a complex problem :)

Interesting. Good show.

Issue #4826, opened 2026-02-09
Warning for incompatible class files version in required binaries


eric-milles:
Full warning:
```
Incompatible .class files version in required binaries. Project 'Abc' is targeting a 17 runtime, but is compiled against '.../com.oracle.coherence.ce/coherence/jars/coherence-25.03.2.jar' (from the container 'Ivy') which requires a 21 runtime
```

I have a Java 17 project -- `JavaSE-17` in `.classpath` and `org.eclipse.jdt.core.prefs` contians:
```
org.eclipse.jdt.core.compiler.codegen.targetPlatform=17
org.eclipse.jdt.core.compiler.compliance=17
org.eclipse.jdt.core.compiler.release=enabled
org.eclipse.jdt.core.compiler.source=17
```

The `coherence-25.03.2.jar` is a multi-release jar that has the following attributes in its manifest:
```
Build-Jdk-Spec: 21
Multi-Release: true
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=17))"
```

I think `org.eclipse.jdt.internal.core.util.Util.getJdkLevel(Object)` is getting the java class version from one of the multi-release classes.  Just a guess.

This is in Eclipse 4.37 (2025-09)

possibly related to #4398

iloveeclipse:
> This is in Eclipse 4.37 (2025-09)

Please test with 4.38.

eric-milles:
Still an issue in 4.39 M2.

iloveeclipse:
@eric-milles : could you please provide self contained example project(s) required to reproduce this issue?

eric-milles:
Another example is the `logback-core` jar which has one multi-release class:
<img width="231" height="121" alt="Image" src="https://github.com/user-attachments/assets/79a87661-99c2-4c83-a7a2-e7cd28f5c7a1" />
<img width="446" height="319" alt="Image" src="https://github.com/user-attachments/assets/f62cb101-e063-4fc0-9679-1fc9f44b0aab" />

iloveeclipse:
@eric-milles : thanks. If you want that something changes here, it would be very helpful to get a **self contained** example (one or many projects).

eric-milles:
<img width="1431" height="155" alt="Image" src="https://github.com/user-attachments/assets/accf1606-ed7c-49a7-993d-ed96e45b4456" />

[...]

PR #4848, opened 2026-02-17
Fix warning for incompatible class files version using multi release jars

iloveeclipse:
The jar top level directory of multi release jars is expected to contain classes compiled for lowest supported Java version. Classes under META-INF/versions/N can be compiled against higher Java version.

The `org.eclipse.jdt.internal.core.util.Util.getJdkLevel(Object)` is supposed to return **lowest** required Java version, so it must ignore any jar entries below META-INF/versions/ to avoid higher version to be returned.

* Use static imports in `org.eclipse.jdt.internal.core.util.Util`
* Introduce `Util.METAINF_VERSIONS` & use it in all places
* Added multi-release jar binary for tests, targeting Java 8 and containing extra versions for all major LTS releases since 8 (11-25).
* Added a working multi-release jar source project used to generate the multi-release jar for tests
* Note: Eclipse can't compile added `multi_release_jar_example` project yet, it does not support properly different JLS levels per source folder, so the jar from it is built on command line.
* Added test case for #4826.

See https://docs.oracle.com/en/java/javase/25/docs/specs/jar/jar.html

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4826

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/util/Util.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/util/Util.java
@@ -910,7 +910,7 @@ public class Util {
 	}
 
 	/**
-	 * Get the jdk level of this root.
+	 * Get the lowest jdk compliance level required by given library.
 	 * The value can be:
 	 * <ul>
 	 * <li>{@code major<<16 + minor} : see predefined constants on ClassFileConstants </li>
@@ -946,7 +946,9 @@ public class Util {
 							for (Enumeration<? extends ZipEntry> e= jar.entries(); e.hasMoreElements();) {
 								ZipEntry member= e.nextElement();
 								String entryName= member.getName();
-								if (isClassFileName(entryName)) {
+								// Ignore class files in META-INF/versions as they are provided for higher JLS and not
+								// relevant to determine the lowest jdk level supported by the multi-release jar
+								if (isClassFileName(entryName) && !entryName.startsWith(METAINF_VERSIONS)) {
 									reader = ClassFileReader.read(jar, entryName);
 									break;
 								}
--- a/org.eclipse.jdt.core.tests.builder/src/org/eclipse/jdt/core/tests/builder/Bug571363Test.java
+++ b/org.eclipse.jdt.core.tests.builder/src/org/eclipse/jdt/core/tests/builder/Bug571363Test.java
@@ -14,12 +14,8 @@
 package org.eclipse.jdt.core.tests.builder;
 
 import java.io.File;
-import java.io.IOException;
-import java.net.URL;
 import junit.framework.Test;
-import org.eclipse.core.runtime.FileLocator;
 import org.eclipse.core.runtime.IPath;
-import org.eclipse.core.runtime.Platform;
 import org.eclipse.jdt.core.JavaCore;
 import org.eclipse.jdt.core.JavaModelException;
 import org.eclipse.jdt.core.tests.util.Util;
@@ -32,19 +28,6 @@ public class Bug571363Test extends BuilderTests {
 
 	public static Test suite() {
 		return buildTestSuite(Bug571363Test.class);
-	}
-
-	/**
-	 * Returns the OS path to the directory that contains this plugin.
-	 */
-	protected String getCompilerTestsPluginDirectoryPath() {
-		try {
-			URL platformURL = Platform.getBundle("org.eclipse.jdt.core.tests.builder").getEntry("/");
-			return new File(FileLocator.toFileURL(platformURL).getFile()).getAbsolutePath();
-		} catch (IOException e) {
-			e.printStackTrace();
-		}
-		return null;
 	}
 
 	public void _testBug571363() throws JavaModelException, Exception {
--- a/org.eclipse.jdt.core.tests.builder/src/org/eclipse/jdt/core/tests/builder/BuilderTests.java
+++ b/org.eclipse.jdt.core.tests.builder/src/org/eclipse/jdt/core/tests/builder/BuilderTests.java
@@ -17,10 +17,12 @@ import static org.eclipse.jdt.core.tests.util.AbstractCompilerTest.F_12;
 import static org.eclipse.jdt.core.tests.util.AbstractCompilerTest.F_9;
 import static org.eclipse.jdt.core.tests.util.AbstractCompilerTest.getPossibleComplianceLevels;
 
+import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
+import java.net.URL;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
@@ -30,7 +32,9 @@ import java.util.List;
 import junit.framework.Test;
 import junit.framework.TestSuite;
 import org.eclipse.core.resources.IProject;
+import org.eclipse.core.runtime.FileLocator;
 import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.Platform;
 import org.eclipse.jdt.core.IClasspathEntry;
 import org.eclipse.jdt.core.JavaCore;
 import org.eclipse.jdt.core.JavaModelException;
@@ -636,6 +640,16 @@ public class BuilderTests extends TestCase {
 			jarPath = env.addInternalJar(projectPath, jarName, jarContent);
 		}
 		return jarPath;
+	}
+
+
+	/**
+	 * Returns the OS path to the directory that contains this plugin.
+	 * @throws IOException
+	 */
+	protected static String getCompilerTestsPluginDirectoryPath() throws IOException {
+		URL platformURL = Platform.getBundle("org.eclipse.jdt.core.tests.builder").getEntry("/");
+		return new File(FileLocator.toFileURL(platformURL).getFile()).getAbsolutePath();
 	}
 
 	protected static void expectCompileProblem(IPath project, String expectedProblemMessage) {
--- a/org.eclipse.jdt.core.tests.builder/src/org/eclipse/jdt/core/tests/builder/BuildpathTests.java
+++ b/org.eclipse.jdt.core.tests.builder/src/org/eclipse/jdt/core/tests/builder/BuildpathTests.java
@@ -757,7 +757,7 @@ public void testMissingLibrary4() throws JavaModelException {
 	env.removeProject(projectPath);
 }
 // https://bugs.eclipse.org/bugs/show_bug.cgi?id=172345
-public void testIncompatibleJdkLEvelOnProject() throws JavaModelException {
+public void testIncompatibleJdkLevelOnProject() throws JavaModelException {
 
 	// Create project
 	IPath projectPath = env.addProject("Project");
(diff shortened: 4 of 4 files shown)


org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/util/Util.java
```diff
@@ -934,7 +946,9 @@ public static long getJdkLevel(Object targetLibrary) {
 							for (Enumeration<? extends ZipEntry> e= jar.entries(); e.hasMoreElements();) {
 								ZipEntry member= e.nextElement();
 								String entryName= member.getName();
-								if (org.eclipse.jdt.internal.compiler.util.Util.isClassFileName(entryName)) {
+								// Ignore class files in META-INF/versions as they are provided for higher JLS and not
+								// relevant to determine the lowest jdk level supported by the multi-release jar
+								if (isClassFileName(entryName) && !entryName.startsWith(METAINF_VERSIONS)) {
```
iloveeclipse, line 951: This one line is the actual fix. The rest is cosmetics and tests.

iloveeclipse:
@srikanth-sankaran or @stephan-herrmann : considering the one line change, WDYT about merging this for RC1?

stephan-herrmann:
* The preliminary refactorings are clean :heavy_check_mark: 
* The actual change makes sense :heavy_check_mark: 
   * the claim to return the **lowest** contained JDK level is based on the assumption that the jar file exactly follows the specification of MR jars, as it doesn't really compare levels of contained class files.
* Is the **semantic change** safe? Luckily `Util.getJdkLevel(Object)` has only 2 callers in production code. 
   * `ClasspathEntry.validateLibraryEntry(..)` is the exact code location that wants this change :heavy_check_mark: 
   * in `JarPackageFragmentRoot.computeChildren(..)` this method is called from a context that is actually prepared to handle MR jars. Now the resulting `classLevel` is only used for validating a package name inside the jar, so we might in theory be using the wrong version of validation rules for anything in versioned folders of the jar. But as name restrictions monotonically increase with version numbers, using the lowest version could only mean we validate that folder/package with more lenient rules than advised. So there is a microscopic risk that we fail to detect an invalid MR jar.  That's way better then the original problem from #4826 
* I am disappointed, though, by the way the example jar is created:
   * compilation should actually work after #3900, I see you already file #4849 :heavy_check_mark: 
   * I didn't succeed to re-create the MR jar, which means I wouldn't be able to make adjustments to that test.
      * I can't run .bat files on my box :) -- perhaps s.o. with a windows box may verify the .bat script?
      * The given `jar` commandline didn't produce the expected structure inside that jar.
I admit that I didn't consult any documentation for `jar cfm ...` at this point.

Summary: no strong objections :)

iloveeclipse:
> I didn't succeed to re-create the MR jar, which means I wouldn't be able to make adjustments to that test.
> * I can't run .bat files on my box :) -- perhaps s.o. with a windows box may verify the .bat script?
> * The given `jar` commandline didn't produce the expected structure inside that jar.

I would provide Linux build script once I get access to the Linux box again. Usually I'm working on Linux, but right now I've lost access to my workstation for few days.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/util/Util.java
```diff
@@ -132,6 +136,14 @@ public static interface BindingsToNodesMap {
 
 	private static char[][] JAVA_DERIVED_EXTENSIONS;
 
+	/**
+	 * The jar entry path under which JDK expects compiler to place class files for multi-release JARs. See
+	 * https://docs.oracle.com/javase/9/docs/specs/jar/jar.html#multi-release-jar-files.
+	 * <p>
+	 * The value is "META-INF/versions/".
+	 */
+	public static String METAINF_VERSIONS = org.eclipse.jdt.internal.compiler.util.Util.METAINF_VERSIONS;
```
laeubi, line 145: The other constants regarding MR Jar support are in JavaProject:

https://github.com/eclipse-jdt/eclipse.jdt.core/blob/90dce92f3d0b89f81e6387dec9a4e3ea110e8fb3/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/JavaProject.java#L115-L123

so maybe better place this there as well.

org.eclipse.jdt.core.tests.builder/workspace/multi-release-jar-example/.classpath
```diff
@@ -0,0 +1,26 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<classpath>
+	<classpathentry kind="con" path="org.eclipse.jdt.launching.JRE_CONTAINER/org.eclipse.jdt.internal.debug.ui.launcher.StandardVMType/JavaSE-1.8"/>
```
laeubi, line 3: Must be `JavaSE-25` for MR projects that want to use `release=25`

org.eclipse.jdt.core.tests.builder/workspace/multi-release-jar-example/.classpath
```diff
@@ -0,0 +1,26 @@
...
+	<classpathentry kind="src" path="src/java8"/>
+	<classpathentry kind="src" output="bin11" path="src/java11">
```
laeubi, line 5: You should not define an won output folder for that, even though it might work it is at least to verbose.

org.eclipse.jdt.core.tests.builder/workspace/multi-release-jar-example/.classpath
```diff
@@ -0,0 +1,26 @@
...
+		<attributes>
+			<attribute name="release" value="21"/>
+		</attributes>
+	</classpathentry>
+	<classpathentry kind="src" output="bin25" path="src/java25">
+		<attributes>
+			<attribute name="release" value="25"/>
+		</attributes>
+	</classpathentry>
+	<classpathentry kind="output" path="bin8"/>
```
laeubi, line 25: You can use the usual bin folder.

MR classes will automatically be written into `bin/META-INF/versions/<version>` then.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/util/Util.java
iloveeclipse: Yes, it wasn't easy to find good place. I was looking for most commonly used place and found `Util` being very popular - in all code that changed the import was already there. Also it matches nice the "batch" part.

org.eclipse.jdt.core.tests.builder/workspace/multi-release-jar-example/.classpath
iloveeclipse: > You should not define an won output folder for that,

Sorry, what is a "won output folder" and what is meant by "that"?

laeubi:
```suggestion
	<classpathentry kind="src" path="src/java11">
```

> > You should not define an won output folder for that,
> 
> Sorry, what is a "won output folder" and what is meant by "that"?

won output folder -> own output folder

org.eclipse.jdt.core.tests.builder/workspace/multi-release-jar-example/.classpath
iloveeclipse: This doesn't work as intended, because UI doesn't allow to select target 8 for the main folder and uses project default 25 JLS. This then makes no sense as "java8" code is compiled with Java 25 JLS and not compatible to Java 8 anymore. Same for every other folder which will be incompatible to the expected JLS. Additionally to that, with 25 JLS used for project, I get `Target release for src/java11 (11) is lower or equal to the project's default release (25).` for every extra source folder configured with different (lower) target as project default.

If I set project to use Java 25 JRE **but** set project JLS to 8, then there are still compilation errors as before (just no such errors on java11 because there was no new syntax used, only new JRE API).

Note: the main idea of MR jars is to allow main code be compatible with lowest possible JLS (==8) and only few extras be adopted for newer JLS. See https://docs.oracle.com/en/java/javase/25/docs/specs/jar/jar.html.

So the Java will look down the road for "extra" versions, falling back to most compatible version in the jar root. Therefore natural user expectation is to setup the main code to the lowest possible JLS version, not the opposite.

> Any versioned directory whose value of N is less than 9 is ignored ...
> If a multi-release JAR file is deployed on the class path or module path (as an automatic module or an explicit [multi-release module](https://docs.oracle.com/en/java/javase/25/docs/specs/jar/jar.html#modular-multi-release-jar-files)) of major version N of a Java platform release runtime, then a class loader loading classes from that JAR file will first search for class files under the Nth versioned directory, then prior versioned directories in descending order (if present), down to a lower major version bound of 9, and finally under the top-level directory.

Anyway, let discuss that on https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4849.

org.eclipse.jdt.core.tests.builder/workspace/multi-release-jar-example/.classpath
iloveeclipse: If the system allows me to do that, I prefer to have separated folders, especially if I want different JLS used for compilation. If the system already uses some specific output folder, it should stick on that and do not allow user to change it. As before, let move the discussion how to get compilable MR example on https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4849.

iloveeclipse:
> Summary: no strong objections :)

Thanks for looking into that. I understand this as OK to merge.

org.eclipse.jdt.core.tests.builder/workspace/multi-release-jar-example/.classpath
laeubi: > This doesn't work as intended, because UI doesn't allow to select target 8 for the main folder and uses project default 25 JLS.

Change project compliance levels do not work (for obvious reasons) but you need a higher JDK, as with your batch file you will need to use a java 25 `javac`...

> If I set project to use Java 25 JRE but set project JLS to 8, then there are still compilation errors as before 

That might be then a different error, but is the right approach here.

> Note: the main idea of MR jars is to allow main code be compatible with lowest possible JLS (==8) and only few extras be adopted for newer JLS
> Therefore natural user expectation is to setup the main code to the lowest possible JLS version, not the opposite.

As I have initially implemented the feature (and using it in other places) we can assume I understand the concept and I never claimed you should use a different JLS setting anywhere. This is all done automatically "down the road"...

Now merging code that is using a wrong project setup certainly does not help in adopting / supporting the feature better.

org.eclipse.jdt.core.tests.builder/workspace/multi-release-jar-example/.classpath
laeubi: > If the system allows me to do that, I prefer to have separated folders

This is supported but fells a bit superflous

> especially if I want different JLS used for compilation

That's completely independent from output folders...

>  If the system already uses some specific output folder, it should stick on that and do not allow user to change it

I don't see why it should forbid this I can just tell you that it is usually not needed and brings you no benefit and instead let it look like this feature is more complicated that it should be.

laeubi:
> > Summary: no strong objections :)
> 
> Thanks for looking into that. I understand this as OK to merge.

Not that my opinion count on here, but it feels like a bad example (much to complicated settings), requiring a special batch script (only working on windows), using `javac` instead of `ecj` and pollutes repository with binary content (jar produced on a possibly untrusted system) ... 

So I would have appreciate to address the recommendation made by others (me) if asking for feedback or looking into problems.

iloveeclipse:
> Not that my opinion count on here, but it feels like a bad example (much to complicated settings), 

The life is complicated, so our settings too, but I don't think it is "much to complicated", for me it is usual to have multiple source folders with different output folders.

> requiring a special batch script (only working on windows)

As you might have seen in our conversation with Stephan above, I plan to provide Linux shell script as soon as I can get my Linux access again.
 
> using `javac` instead of `ecj` 

Nothing bad with that, we often use javac compiler because it is easier and more compact to write simple test code with that, and it is always available on the system and also produces some "default" input data we have to deal with in JDT anyway.

> and pollutes repository with binary content (jar produced on a possibly untrusted system)

Sure, binary files are not nice, therefore there is the source project provided along that allows to update that file if needed. There are plenty of binary files used in JDT tests for various reasons, one of them is that it is not easy to make sure that something generated "on the fly" is really providing "original" reproducer, so you have to have yet another validation test for the generated binary. Surely possible too, but at some point in time it is a question of ROI.

Issue #4849, opened 2026-02-17
Support compilation for different JLS levels in multi-release projects


iloveeclipse:
In the PR https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4848 I've needed an example of a multi-release jar and I've tried to create a simple example project so that I could play with it.

Unfortunately I've found that we still lack basic MR support for sources that need different JLS levels to be properly compiled in such a project, because the sources use more advances Java features as the base project JLS.

Here is what I see: (one can get the sources for the project from [this commit](https://github.com/eclipse-jdt/eclipse.jdt.core/commit/b87a2bfc4e595863e5ca89b118e5ffd2df17fe3b))

<img width="1513" height="960" alt="Image" src="https://github.com/user-attachments/assets/ebd9e1d6-c3a0-47fa-bfb3-a976154309e2" />

Expected is that there should be no compilation errors if the classpath entries provide "right" Java release targets.

stephan-herrmann:
@laeubi do you plan to continue work on MR projects?

Here's a first hint: We'd need to ensure that also parsers always work with the intended release settings (`options` as well as flags `parsingJava*Plus`). Note that a `Compiler` holds on to an initialized `Parser` which can then be used for parsing source files from different source folders. For those scenarios a temporary configuration change must be implemented for re-used parsers.

In other contexts, like SourceElementParser I don't yet see any connection to MR configuration, see, e.g., `org.eclipse.jdt.internal.core.CompilationUnit.buildStructure(OpenableElementInfo, IProgressMonitor, Map<IJavaElement, IElementInfo>, IResource)` where it uses `getOptions(true)` - which is still unaware of per-sourcefolder options!

This will likely be a sizable project, but without a complete fix, the existing label "Incubating feature"  certainly has to remain.

laeubi:
I could try to look into this but the project setup looks wrong, it uses a Java 1.8 project JVM what can't work, you need to use (obviously) a JVM with the highest level you want to support (e.g. 25 in this case) there should even be an error marker (maybe cut of here) mentioning this, so I would fix this first.

We already have an example of a MR in PDE since a while now here https://github.com/eclipse-pde/eclipse.pde/tree/master/ui/org.eclipse.pde.junit.runtime if one needs a reference.

At best one would provide a minimal reproducer project for inspection or enhances the existing MultiRelease* tests with an example as its easier to grasp they key points.

iloveeclipse:
[...]

PR #4856, opened 2026-02-19
Check live options when determine what java level is used for parsing

laeubi:
Currently the Parser determines what java version is parsed once in the
constructor and store them into (modifiable) fields. This causes issues
when the options are modified as everything else works but this
particular fields are not updated.

Instead of having a way to update those fields and as they do not cache
any complex computations, this fields are now replaced with appropriate
getters like it is already done in some cases.

Fix https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4849

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
@@ -915,17 +915,6 @@ public JavadocParser javadocParser;
 // used for recovery
 protected int lastJavadocEnd;
 public org.eclipse.jdt.internal.compiler.ReadManager readManager;
-final protected boolean parsingJava8Plus = true;
-protected boolean parsingJava9Plus;
-protected boolean parsingJava10Plus;
-protected boolean parsingJava14Plus;
-protected boolean parsingJava15Plus;
-protected boolean parsingJava17Plus;
-protected boolean parsingJava18Plus;
-protected boolean parsingJava21Plus;
-protected boolean parsingJava22Plus;
-protected boolean previewEnabled;
-protected boolean parsingJava11Plus;
 protected int unstackedAct = ERROR_ACTION;
 private boolean haltOnSyntaxError = false;
 private boolean tolerateDefaultClassMethods = false;
@@ -944,16 +933,6 @@ public Parser(ProblemReporter problemReporter, boolean optimizeStringLiterals) {
 	this.options = problemReporter.options;
 	this.optimizeStringLiterals = optimizeStringLiterals;
 	initializeScanner();
-	this.parsingJava9Plus = this.options.sourceLevel >= ClassFileConstants.JDK9;
-	this.parsingJava10Plus = this.options.sourceLevel >= ClassFileConstants.JDK10;
-	this.parsingJava11Plus = this.options.sourceLevel >= ClassFileConstants.JDK11;
-	this.parsingJava14Plus = this.options.sourceLevel >= ClassFileConstants.JDK14;
-	this.parsingJava15Plus = this.options.sourceLevel >= ClassFileConstants.JDK15;
-	this.parsingJava17Plus = this.options.sourceLevel >= ClassFileConstants.JDK17;
-	this.parsingJava18Plus = this.options.sourceLevel >= ClassFileConstants.JDK18;
-	this.parsingJava21Plus = this.options.sourceLevel >= ClassFileConstants.JDK21;
-	this.parsingJava22Plus = this.options.sourceLevel >= ClassFileConstants.JDK22;
-	this.previewEnabled = this.options.sourceLevel == ClassFileConstants.getLatestJDKLevel() && this.options.enablePreviewFeatures;
 	this.astLengthStack = new int[50];
 	this.expressionLengthStack = new int[30];
 	this.typeAnnotationLengthStack = new int[30];
@@ -1154,7 +1133,7 @@ protected void checkAndSetModifiers(int flag){
 	of a list of several modifiers. The startPosition
 	is zeroed when a copy of modifiers-buffer is push
 	onto the this.astStack. */
-	if (flag == ClassFileConstants.AccStrictfp && this.parsingJava17Plus) {
+	if (flag == ClassFileConstants.AccStrictfp && isParsingJava17Plus()) {
 		problemReporter().StrictfpNotRequired(this.scanner.startPosition, this.scanner.currentPosition - 1);
 	}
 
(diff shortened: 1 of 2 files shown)


laeubi:
@stephan-herrmann can you review? This now passes the enhanced test-case with check for new language levels.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
```diff
@@ -13309,13 +13288,47 @@ public boolean automatonWillShift(TerminalToken token) {
...
-	return (this.parsingJava9Plus && this.compilationUnit != null && this.compilationUnit.isModuleInfo());
+	return (isParsingJava9Plus() && this.compilationUnit != null && this.compilationUnit.isModuleInfo());
 }
+
+protected boolean isParsingJava8Plus() {
+	return true;
+}
+protected boolean isParsingJava9Plus() {
+	return this.options!=null && this.options.sourceLevel >= ClassFileConstants.JDK9;
+}
```
srikanth-sankaran, line 13301: A space before and after `!=` is the coding style norm in JDT/Core

But looking at assignments to `this.options` and its usage we can readily conclude that it can never be `null` and so get rid of these null checks

laeubi: > But looking at assignments to `this.options` and its usage we can readily conclude that it can never be `null` and so get rid of these null checks

The test fail because there is a default constructor that left this field `null`...

srikanth-sankaran: > The test fail because there is a default constructor that left this field `null`...

Ah, so we can readily conclude `this.options` can never be null and that conclusion would be bogus 😆 

Ok, please ignore the suggestion to remove the null checks, but coding style change suggested would be good. 

Thanks for correcting me.

laeubi: Done, I also only know because I first tried without the `null` check :-D

iloveeclipse: Just curious: why can't we initialize this with default `CompilerOptions`?

laeubi: > Just curious: why can't we initialize this with default `CompilerOptions`?

I have no clue, the comment says the constructor is for subclasses and for special cases... as I don't want to change current behavior I choose this defensive approach, but of course it could be refactored later on when one think it fits better.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
```diff
@@ -13309,13 +13288,57 @@ public boolean automatonWillShift(TerminalToken token) {
...
+	return isParsingJava14Plus();
 }
 @Override
 public boolean isParsingModuleDeclaration() {
 	// It can be a null in case of a Vanguard parser, which means no module to be dealt with.
-	return (this.parsingJava9Plus && this.compilationUnit != null && this.compilationUnit.isModuleInfo());
+	return (isParsingJava9Plus() && this.compilationUnit != null && this.compilationUnit.isModuleInfo());
 }
+
+protected boolean isParsingJava8Plus() {
```
iloveeclipse: this method can be removed, we don't support Java below 8 anymore.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
```diff
@@ -13309,13 +13288,57 @@ public boolean automatonWillShift(TerminalToken token) {
...
+
+protected boolean isParsingJava9Plus() {
+	return this.options != null && this.options.sourceLevel >= ClassFileConstants.JDK9;
+}
+
+protected boolean isParsingJava10Plus() {
+	return this.options != null && this.options.sourceLevel >= ClassFileConstants.JDK10;
+}
+
+protected boolean isParsingJava11Plus() {
```
iloveeclipse: this method can be removed too (unused)

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
```diff
@@ -13309,13 +13288,57 @@ public boolean automatonWillShift(TerminalToken token) {
...
+	return this.options != null && this.options.sourceLevel >= ClassFileConstants.JDK11;
+}
+
+protected boolean isParsingJava14Plus() {
```
iloveeclipse, line 13307: This method can be removed, all callers too, but in a later PR

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
```diff
@@ -13309,13 +13288,57 @@ public boolean automatonWillShift(TerminalToken token) {
...
+	return this.options != null && this.options.sourceLevel >= ClassFileConstants.JDK14;
+}
+
+protected boolean isParsingJava15Plus() {
```
iloveeclipse: this method can be removed, it is unused

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
```diff
@@ -13309,13 +13288,57 @@ public boolean automatonWillShift(TerminalToken token) {
...
+	return this.options != null && this.options.sourceLevel >= ClassFileConstants.JDK15;
+}
+
+protected boolean isParsingJava17Plus() {
+	return this.options != null && this.options.sourceLevel >= ClassFileConstants.JDK17;
 }
 
+protected boolean isParsingJava18Plus() {
```
iloveeclipse: this method can be removed, it is unused

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
```diff
@@ -13309,13 +13288,57 @@ public boolean automatonWillShift(TerminalToken token) {
...
+	return this.options != null && this.options.sourceLevel >= ClassFileConstants.JDK18;
+}
+
+protected boolean isParsingJava21Plus() {
```
iloveeclipse: this method can be removed, it is unused

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
```diff
@@ -13309,13 +13288,57 @@ public boolean automatonWillShift(TerminalToken token) {
...
+	return this.options != null && this.options.sourceLevel >= ClassFileConstants.JDK21;
+}
+
+protected boolean isParsingJava22Plus() {
```
iloveeclipse: this method can be removed, it is unused

iloveeclipse:
@laeubi : could you please squash three last commits into one, so that PR has two commits - one for tests and one for the fix? Thanks.

laeubi:
Squashed and rebased now.

iloveeclipse:
OK all tests green, let's merge for RC1.

Issue #4421, opened 2025-09-23
[build][test] ModuleBuilderTests fails with Assertion error


subyssurendran666:
In https://download.eclipse.org/eclipse/downloads/drops4/I20250918-0350/testresults/html/org.eclipse.jdt.core.tests.model_ep438I-unit-macosx-aarch64-java21_macosx.cocoa.aarch64_21.html

org.eclipse.jdt.core.tests.model.**ModuleBuilderTests**. **testConvertToModule**

```
**incorrect result. ----------- Expected ------------ java.desktop\n java.rmi\n java.sql ------------ but was ------------ java.datatransfer\n java.desktop\n java.net.http\n java.rmi\n java.sql ---------------------- ---------- expected:<java.d[eskto]p\n java.rmi\n java....> but was:<java.d[atatransfer\n java.desktop\n java.net.htt]p\n java.rmi\n java....>

junit.framework.ComparisonFailure: incorrect result.
----------- Expected ------------
java.desktop\n
java.rmi\n
java.sql
------------ but was ------------
java.datatransfer\n
java.desktop\n
java.net.http\n
java.rmi\n
java.sql
---------------------- ----------
expected:<java.d[eskto]p\n
java.rmi\n
java....> but was:<java.d[atatransfer\n
java.desktop\n
java.net.htt]p\n
java.rmi\n
java....>
at org.eclipse.jdt.core.tests.junit.extension.TestCase.assertStringEquals(TestCase.java:265)
at org.eclipse.jdt.core.tests.junit.extension.TestCase.assertEquals(TestCase.java:240)
at org.eclipse.jdt.core.tests.model.AbstractJavaModelTests.assertStringsEqual(AbstractJavaModelTests.java:1524)
at org.eclipse.jdt.core.tests.model.ModuleBuilderTests.testConvertToModule(ModuleBuilderTests.java:715)
at java.base/java.lang.reflect.Method.invoke(Method.java:580)
at junit.framework.TestCase.runTest(TestCase.java:177)
at org.eclipse.jdt.core.tests.junit.extension.TestCase.runTest(TestCase.java:970)
at junit.framework.TestCase.runBare(TestCase.java:142)
at junit.framework.TestResult$1.protect(TestResult.java:122)
at junit.framework.TestResult.runProtected(TestResult.java:142)
at junit.framework.TestResult.run(TestResult.java:125)
at junit.framework.TestCase.run(TestCase.java:130)
[...]

PR #4860, opened 2026-02-19
Set default VM for JavaSE-9 execution environment during module tests

trancexpress:
See: #4421

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ModuleBuilderTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ModuleBuilderTests.java
@@ -34,6 +34,7 @@ import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.Path;
+import org.eclipse.core.runtime.Status;
 import org.eclipse.jdt.core.*;
 import org.eclipse.jdt.core.dom.AST;
 import org.eclipse.jdt.core.dom.CompilationUnit;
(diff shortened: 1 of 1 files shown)


org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ModuleBuilderTests.java
```diff
@@ -691,6 +699,7 @@ public void _test019() throws CoreException, IOException {
 	}
 	public void testConvertToModule() throws CoreException, IOException {
 		Hashtable<String, String> javaCoreOptions = JavaCore.getOptions();
+		IVMInstall vm = prepareExecutionEnvironment(JAVASE_9);
```
trancexpress, line 702: The test sets JLS 10, but the test projects `.classpath` is using `JavaSE-9` (I have no idea how that makes sense, it of course works when the backing JRE is 10+...).

So we set the 9 execution environment. Initially I thought we'll need `JavaSE-10` here and `JavaSE-9` in the other test, so that is how I wrote the code. I guess it doesn't hurt to be able to call it for different versions.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/ModuleBuilderTests.java
```diff
@@ -9136,4 +9148,25 @@ protected void sortMarkers(IMarker[] markers) {
 		Arrays.sort(markers, Comparator.comparingInt((IMarker a) -> a.getAttribute(IMarker.CHAR_START, 0))
 									   .thenComparing((IMarker a) -> a.getAttribute(IMarker.MESSAGE, "")));
 	}
+
+	private IVMInstall prepareExecutionEnvironment(String environmentId) {
```
iloveeclipse, line 9158: Similar to https://github.com/eclipse-jdt/eclipse.jdt.debug/pull/866#discussion_r2828883773, I miss explanation what the code does and why. We've spent a day reverse engineering the other test failure, so we should not waste time for other maintainers to do the same later on trying to understand what we did here :-) .

Issue #4861, opened 2026-02-19
Eclipse 4.39M3 - Compilation fails due to java.lang.NullPointerException: Cannot read field "compatibleCache" because "this" is null


niks:
I've tested the following developer build of Eclipse:
```
    Eclipse SDK
    Version: 2026-03 (4.39)
    Build id: I20260212-1800
```

The compilation of our product fails with internal compiler issues. The error log contains the errors below. It is not clear to me how I can create a minimal example for this issue. 
We are currently stuck on the September release of Eclipse as the Dezember release also fails to build our product due to #4699. Therefore, I hope that the issue can be fixed before the 2026-03 release.

"Error in JDT Core during AST creation"

>    java.lang.NullPointerException: Cannot read field "compatibleCache" because "this" is null
    at org.eclipse.jdt.internal.compiler.lookup.ReferenceBinding.isCompatibleWith(ReferenceBinding.java:1491)
    at org.eclipse.jdt.internal.compiler.lookup.TypeBinding.isCompatibleWith(TypeBinding.java:730)
    at org.eclipse.jdt.internal.compiler.lookup.TypeBinding.isTypeArgumentContainedBy(TypeBinding.java:1397)
    at org.eclipse.jdt.internal.compiler.lookup.ParameterizedTypeBinding.isEquivalentTo(ParameterizedTypeBinding.java:925)
    at org.eclipse.jdt.internal.compiler.lookup.ReferenceBinding.isSuperclassOf(ReferenceBinding.java:1770)
    at org.eclipse.jdt.internal.compiler.lookup.ReferenceBinding.isCompatibleWith0(ReferenceBinding.java:1569)
    at org.eclipse.jdt.internal.compiler.lookup.ReferenceBinding.isCompatibleWith(ReferenceBinding.java:1490)
    at org.eclipse.jdt.internal.compiler.lookup.TypeBinding.isCompatibleWith(TypeBinding.java:730)
    at org.eclipse.jdt.internal.compiler.lookup.TypeBinding.isTypeArgumentContainedBy(TypeBinding.java:1397)
    at org.eclipse.jdt.internal.compiler.lookup.ParameterizedTypeBinding.isEquivalentTo(ParameterizedTypeBinding.java:925)
    at org.eclipse.jdt.internal.compiler.lookup.ReferenceBinding.implementsInterface(ReferenceBinding.java:1379)
[...]

PR #4862, opened 2026-02-19
Eclipse 4.39M3 - Compilation fails due to java.lang.NullPointerException: Cannot read field "compatibleCache" because "this" is null

stephan-herrmann:
Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4861

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding.java
@@ -307,7 +307,9 @@ public class CaptureBinding extends TypeVariableBinding {
 		if(scope.environment().usesNullTypeAnnotations()) {
 			evaluateNullAnnotations(scope, null);
 		}
-		this.compatibleCache = null; // any checks during initialization may have produced bogus results, get rid of them now
+		if (this.compatibleCache != null) {
+			this.compatibleCache.clear(); // any checks during initialization may have produced bogus results, get rid of them now
+		}
 	}
 
 	@Override


stephan-herrmann:
Analysis:
> java.lang.NullPointerException: Cannot read field "compatibleCache" because "this" is null
at org.eclipse.jdt.internal.compiler.lookup.ReferenceBinding.isCompatibleWith(ReferenceBinding.java:1491)

So NPE on `compatibleCache` happens here:

https://github.com/eclipse-jdt/eclipse.jdt.core/blob/ca03567dddeb2883b58221c1e0394932de6649bb/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java#L1491

previous two lines are
```java
	this.compatibleCache.put(otherType, Boolean.FALSE); // protect from recursive call
	if (isCompatibleWith0(otherType, captureScope)) {
```
The first of these lines sees `compatibleCache` as non-null, so something must have re-assigned the variable to `null` during `isCompatibleWith0()`.

The code base contains exactly 3 assignments to this field.
* initialization with a fresh hashmap just above this snippet.
* copy-assignment in the copy ctor of `ReferenceBinding`, since this is a ctor (with now dangerous control flow) it cannot possibly overwrite an existing value.
* assignment with `null` from `CaptureBinding.initializeBounds()` as added in #4813 

Hence, "forgetting" a cached value must not reset to null, but clear the map (if any), even though `null` generally is a legal value, just not during execution of `isCompatibleWith0()`.

I don't yet have a regression test, and while the sequence of events is quite clear, its not clear how the same capture binding could be initialized while already being the receiver of `isCompatible0()` somewhere deeper on the stack.

srikanth-sankaran:
> * copy-assignment in the copy ctor of `ReferenceBinding`, since this is a ctor (with now dangerous control flow) it cannot possibly overwrite an existing value.

Could you explain what you mean by _with now dangerous control flow_ ??

srikanth-sankaran:
> Analysis:

Sounds reasonable. 

> Hence, "forgetting" a cached value must not reset to null, but clear the map (if any), even though `null` generally is a legal value, just not during execution of `isCompatibleWith0()`.

This looks correct too.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding.java
```diff
@@ -307,7 +307,9 @@ public void initializeBounds(Scope scope, ParameterizedTypeBinding capturedParam
 		if(scope.environment().usesNullTypeAnnotations()) {
 			evaluateNullAnnotations(scope, null);
 		}
-		this.compatibleCache = null; // any checks during initialization may have produced bogus results, get rid of them now
+		if (this.compatibleCache != null) {
+			this.compatibleCache.clear(); // any checks during initialization may have produced bogus results, get rid of them now
+		}
 	}
```
srikanth-sankaran, line 313: Looks good to me.

srikanth-sankaran: > Looks good to me.

Even if by an extremely rare scenario we are actually looking at VM optimizer bug, the proposed change still looks good.

stephan-herrmann:
> > * copy-assignment in the copy ctor of `ReferenceBinding`, since this is a ctor (with now dangerous control flow) it cannot possibly overwrite an existing value.
> 
> Could you explain what you mean by _with now dangerous control flow_ ??

sorry for the typo, should read "with no dangerous control flow"
(as dangerous I'd judge, e.g., invoking an instance method which could be overridden to do unexpected stuff, or anything with substantial fan out).

stephan-herrmann:
Trying to reverse engineer a failing test case, let's see how the exception could possibly be triggered.

The culprit `null` is assigned in `CaptureBinding.initializeBounds()`.

That method has exactly one caller: `ParameterizedTypeBinding.capture(Scope, int, int)`.

For the exception to be raised in `ReferenceBinding.isCompatibleWith()` method `capture()` would have to be invoked during `isCompatibleWith0()`, and, the receiver of `isCompatibleWith()` must be the same instance as the receiver of `initializeBounds()`, in order to affect that very field `compatibleCache`.

Inside `capture()` method `initializeBounds()` is invoked on only such elements `capturedArguments[i]` where `freshCaptures[i]`, which should "normally" imply that this capture binding has been created just now, i.e., never exposed to any outside code. Under this "normal" condition the fresh capture could never be the receiver of any `isCompatibleWith()` call down the call stack.

Theory thus suggests, that the capture isn't really fresh, i.e., `this.environment.createCapturedWildcard()` probably answered a pre-existing capture binding. This smells like inference computing the same capture more than once, i.e., inference performing several attempts with similar inputs.

That theory also suggest, that it would be prudent to add an early exit to `CaptureBinding.initializeBounds()` to avoid repeated initialized in the first place (that would be an additional safety measure on top of the current fix).

As for the exact call path in the unhappy case, it seems like `isCompatibleWith0()` itself is calling `capture()` (as it happens in one rarely used branch, triggered in just 7 cases of our regression suite). Surely, the various checks in `isCompatibleWith0()` could theoretically bounce int `capture()` via some odd indirection, but I could not find any likely candidate, based on the assumption that `isCompatibleWith0()` works only on resolved type bindings, never branching back into any resolution logic (where capture could of course happen).

I'll continue by studying how exactly these two rare conditions can be forced:
* `createCapturedWildcard()` answering a pre-existing capture binding
* `isCompatibleWith0()` invoking `capture()`

If we can trigger both conditions in the same test case, we might have a chance of creating a regression test.

stephan-herrmann:
Borrowing from existing tests that trigger one of the two necessary conditions I crafted this program, which indeed triggers both conditions:
```java
			interface SelfBound<S extends SelfBound<S, T>, T> { }
			public interface X <A extends SelfBound<?,A>> {
				default void foo2(A arg) {
					SelfBound<? extends SelfBound<?, A>, ?> var1 = m1(m2(String::toUpperCase));
				}
				<B> A m1(SelfBound<? extends SelfBound<?, B>, ?> s);
				<R> SelfBound<? extends SelfBound<?,R>, R> m2(Function<String, ?> f);
			}
```

Only now I realize the this trail doesn't seem to lead to a regression test because I wanted `foo.isCompatibleWith0()` to invoke `bar.capture()` and then `foo.initializeBounds()` (for the same `foo`!!!), but `isCompatibleWith0()` can only invoke `capture()` if `foo` is a TypeVariableBinding, but for `initializeBounds()` we need `foo` to be a WildcardBinding.

Thus I'm running out of ideas, how the NPE can be triggered.

stephan-herrmann:
> As for the exact call path in the unhappy case, it seems like `isCompatibleWith0()` itself is calling `capture()`.

This possibility has been ruled out (see prev. comment: receiver needs to be TypeVariableBinding for that branch).

So we're left with:

> Surely, the various checks in `isCompatibleWith0()` could theoretically bounce int `capture()` via some odd indirection, but I could not find any likely candidate, based on the assumption that `isCompatibleWith0()` works only on resolved type bindings, never branching back into any resolution logic (where capture could of course happen).

Indeed I found an interesting call stack in GenericsRegressionTest.testGH4236():
```
ParameterizedTypeBinding.capture(Scope, int, int) line: 191	
CapturingContext.maybeCapture(ReferenceBinding) line: 59	
ParameterizedTypeBinding(ReferenceBinding).isSuperclassOf(ReferenceBinding) line: 1782	
CaptureBinding(ReferenceBinding).isCompatibleWith0(TypeBinding, Scope) line: 1582	
CaptureBinding(ReferenceBinding).isCompatibleWith(TypeBinding, Scope) line: 1492	
CaptureBinding(TypeBinding).isCompatibleWith(TypeBinding) line: 730	
CaptureBinding(TypeBinding).isTypeArgumentContainedBy(TypeBinding) line: 1397	
ParameterizedTypeBinding.isEquivalentTo(TypeBinding) line: 929	
ParameterizedTypeBinding(ReferenceBinding).isCompatibleWith0(TypeBinding, Scope) line: 1521	
ParameterizedTypeBinding(ReferenceBinding).isCompatibleWith(TypeBinding, Scope) line: 1492	
MethodScope(Scope).parameterCompatibilityLevel(TypeBinding, TypeBinding, LookupEnvironment, MethodBinding) line: 5038	
MethodScope(Scope).parameterCompatibilityLevel(MethodBinding, TypeBinding[], InvocationSite) line: 4994	
MethodScope(Scope).computeCompatibleMethod(MethodBinding, TypeBinding[], InvocationSite) line: 835	
MethodScope(Scope).getConstructor0(ReferenceBinding, TypeBinding[], InvocationSite) line: 2475	
MethodScope(Scope).getConstructor(ReferenceBinding, TypeBinding[], InvocationSite) line: 2438	
ExplicitConstructorCall(Statement).findConstructorBinding(BlockScope, Invocation, ReferenceBinding, TypeBinding[]) line: 575	
ExplicitConstructorCall.resolve(BlockScope) line: 486	
ConstructorDeclaration.resolveStatements() line: 796	
ConstructorDeclaration(AbstractMethodDeclaration).resolve(ClassScope) line: 628	
ConstructorDeclaration.resolve(ClassScope) line: 766	
TypeDeclaration.resolve() line: 1459	
TypeDeclaration.resolve(CompilationUnitScope) line: 1610	
CompilationUnitDeclaration.resolve() line: 661	
```

We still need to observe that `capture()` call reusing existing capture bindings. In particular it must be re-using the very capture on which `isCompatibleWith0()` is being executed. 

It this point I start to believe that that NPE is possible, but I'm not able to create a test that demonstrates it.

iloveeclipse:
@stephan-herrmann , @srikanth-sankaran : RC2 is this Wednesday. Do we have a conclusion here? Should it be merged or not?

srikanth-sankaran:
IMO, we can merge the fix. A regression test would be ideal but given sufficient effort has been expended without full success, I wouldn't like the absence of a test to be a blocker for fixing a recently introduced regression that can be clearly tracked to a specific line of change.

iloveeclipse:
@stephan-herrmann, @srikanth-sankaran : could it be, there is some multi-threaded issue with the code? 
Could it be that NPE is caused by entering the code from different threads? In such case, the fix would be not complete, as now we would get issues with the invalid state of the map.

stephan-herrmann:
> @stephan-herrmann, @srikanth-sankaran : could it be, there is some multi-threaded issue with the code? Could it be that NPE is caused by entering the code from different threads? In such case, the fix would be not complete, as now we would get issues with the invalid state of the map.

Generally, the compiler is single-threaded by design (separate instances of `Compiler` etc. per thread). I recall only one [concurrency bug](https://bugs.eclipse.org/bugs/show_bug.cgi?id=372319), which I "skillfully" injected  into the compiler :flushed:  (Srikanth surely remembers!!!!). That one involved modifying a global constant ... which is not in sight here.

iloveeclipse:
> That one involved modifying a global constant ...

8654e35d625c39f9786ab81699fccb6057b0bc72

> which is not in sight here.

OK, was just an idea.
So let's merge this.

stephan-herrmann:
>  So let's merge this.

thanks!

Issue #4865, opened 2026-02-21
compiler error when accessing final field in static initializer (in anonymous inner class)


martinlippert:
JDT creates an error for:

```
public class Foo {
	
	private final String greeting;
	
	public Foo() {
		this.greeting = "Hello";
		
		Comparator<String> comp = new Comparator<String>() {
			
			{
				System.out.println(greeting);
			}

			@Override
			public int compare(String o1, String o2) {
				return o1.compareTo(o2);
			}
		};
	}

}
```

The access to `greeting` inside of the static initializr of the anonymous inner class triggers a compiler error.

This happens in this version:
- `org.eclipse.jdt.core_3.45.0.v20260211-1013`
- `org.eclipse.jdt.core.compiler.batch_3.45.0.v20260212-1559`

`javac` compiles the code just fine.
JDT from Eclipse 2025-12 compiles the code as well just fine and does not report an error.

Looks like a regression in the latest 2026-03 milestone builds.
Hope this can be fixed for the release... :-)

HannesWell:
> Hope this can be fixed for the release... :-)

Since we are already in RC2 phase and RC2 build is planned for Wednesday, the runway for this is close to zero.

Still, @jarthana or @stephan-herrmann are you aware of this?

I encountered this as well and I could only solve this by refactoring the code not not use static initializers (which, at least SonarLint, recommends anyways).

iloveeclipse:
@coehlrich : this is regression from 78ef5fc5afbff5f37603bc38060908c8930d9989.
Could you please check?

jarthana:
> > Hope this can be fixed for the release... :-)
> 
> Since we are already in RC2 phase and RC2 build is planned for Wednesday, the runway for this is close to zero.
> 
> Still, [@jarthana](https://github.com/jarthana) or [@stephan-herrmann](https://github.com/stephan-herrmann) are you aware of this?

I can take a look at it tomorrow. But still our best bet is @coehlrich or @stephan-herrmann coming up with a simple fix.

stephan-herrmann:
Reproduced, I'm on it.

stephan-herrmann:
>  [@stephan-herrmann](https://github.com/stephan-herrmann) are you aware of this?

No, I wasn't. Thanks for the ping.

jarthana:
I have verified the fix in the latest nightly build.

martinlippert:
[...]

PR #4871, opened 2026-02-23
compiler error when accessing final field in static initializer (in anonymous inner class) #4865

stephan-herrmann:
revert to coehlrich's first version of PR #4701

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4865

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeDeclaration.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeDeclaration.java
@@ -264,7 +264,7 @@ public FlowInfo analyseCode(BlockScope currentScope, FlowContext flowContext, Fl
 		}
 		manageEnclosingInstanceAccessIfNecessary(currentScope, flowInfo);
 		updateMaxFieldCount(); // propagate down the max field count
-		internalAnalyseCode(flowContext, flowInfo.unconditionalFieldLessCopy()); // discards info about fields of enclosing classes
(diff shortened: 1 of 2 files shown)


stephan-herrmann:
In #4701 I had urged @coehlrich  to extend his initial fix, as to draw a clear line between outside and inside of a class, because inside the class fields of an outer class could be confused with local fields of the some id (the root cause of #4700).

In one particular branch of `TypeDeclaration.internalAnalyseCode()` this proved to be wrong. Hence this PR goes back to what we had in https://github.com/eclipse-jdt/eclipse.jdt.core/commit/962c1d9eb5a0e3deae0707cee807883155373204

To avoid the bug #4865 any code that executes during the anonymous instantiation must be analyzed in the context of the enclosing constructor.

stephan-herrmann:
For future consideration:

Debugging the new test case I found it quite confusing how, e.g., scopes are navigated in `needBlankFinalFieldInitializationCheck()` to determine responsibility and hence avoid the confusion of field ids mentioned above. If ever we want to revisit this, then a good starting point may still be the analysis in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4701#discussion_r2648719025

HannesWell:
Thanks for the prompt fix.
We probably have another I-build run soon(-ish) after the the build failure from tomorrow is fixed.
Do you want me to await the submission of this one?

jarthana:
> Thanks for the prompt fix. We probably have another I-build run soon(-ish) after the the build failure from tomorrow is fixed. Do you want me to await the submission of this one?

Yes, that'll be nice as we wait for another review from @srikanth-sankaran.

iloveeclipse:
Thanks all, merging.

Issue #4870, opened 2026-02-23
Performance degradation in jdt core 3.41


robserm:
I observe a significant performance degradation, 3.36 vs 3.44, when parsing a single file (createAST method)  in standalone mode. This is probably a regression. From my measurement, it looks like 3.44 is about **10 times** slower than 3.36 for the same task. 3.36 required about 28s to parse 1190 files, and 3.44 required about 325s for the same task.

See the results below:
JDT 3.36
Measurement (5 rounds)...
Round 1: 28807.6 ms (24.208 ms/file)
Round 2: 28377.9 ms (23.847 ms/file)
Round 3: 27850.4 ms (23.404 ms/file)
Round 4: 28333.4 ms (23.810 ms/file)
Round 5: 27773.0 ms (23.339 ms/file)

=== Results ===
Rounds: 5
Files per round: 1190
Avg round time: **28228.5 ms**
Min round time: 27773.0 ms
Max round time: 28807.6 ms
Avg time/file: 23.721 ms

JDT 3.44:

Measurement (5 rounds)...
Round 1: 325603.8 ms (273.617 ms/file)
Round 2: 326659.5 ms (274.504 ms/file)
Round 3: 325608.6 ms (273.621 ms/file)
Round 4: 323995.5 ms (272.265 ms/file)
Round 5: 327399.0 ms (275.125 ms/file)

=== Results ===
Rounds: 5
Files per round: 1190
Avg round time: **325853.2 ms**
Min round time: 323995.5 ms
Max round time: 327399.0 ms
Avg time/file: 273.826 ms

Please find attached the source file used to perform the measurements. The only difference between the runs was the library version, which was changed from 3.66 to 3.44.
The code requires a path to a directory containing the Java source files, along with a classpath.txt file that includes the necessary classpath entries.
If you require any additional information or would like me to conduct further measurements with different configurations, I would be happy to assist.

[JdtParserBenchmark.java](https://github.com/user-attachments/files/25493308/JdtParserBenchmark.java)

iloveeclipse:
Thnaks for the report.
Two things:
1) Can you please test with latest build https://download.eclipse.org/eclipse/downloads/drops4/I20260221-1800/ ?
2) If the problem is still there, please could you bisect to the exact revision with the regression, because the range 3.36 vs 3.44 is pretty big.

robserm:
sure here are the exact versions:
[...]

PR #4886, opened 2026-03-02
Don't call Charset.forName() with null argument

iloveeclipse:
The Charset.forName() will throw `IllegalArgumentException` on a null argument and after that we will return `Charset.defaultCharset()` anyway. So let avoid the exception throwing, for performance and for better profiling of more important exceptions.

See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4870


iloveeclipse:
Test failure in `ClasspathTests.testDotDotLibraryEntry8` is unrelated.
```
/P[CPE_SOURCE][K_SOURCE][isExported:false]\n
../external.jar[CPE_LIBRARY][K_BINARY][isExported:false]
/P[CPE_SOURCE][K_SOURCE][isExported:false]\n
JCL18_LIB[CPE_VARIABLE][K_SOURCE][isExported:false]
```

Issue #4903, opened 2026-03-06
ClasspathTests doesn't join auto-refresh job


trancexpress:
If I set a breakpoint at the start of `RefreshJob.runInWorkspace(IProgressMonitor)` and debug `ClasspathTests`, I see about 90 failures (out of 250 tests) that look like this:

```
junit.framework.AssertionFailedError: Project 'P' should be present in JavaModel, but we found only: []
	at junit.framework.Assert.fail(Assert.java:57)
	at junit.framework.Assert.assertTrue(Assert.java:22)
	at junit.framework.TestCase.assertTrue(TestCase.java:192)
	at org.eclipse.jdt.core.tests.model.AbstractJavaModelTests.createJavaProject(AbstractJavaModelTests.java:2379)
	at org.eclipse.jdt.core.tests.model.AbstractJavaModelTests.createJavaProject(AbstractJavaModelTests.java:2004)
	at org.eclipse.jdt.core.tests.model.AbstractJavaModelTests.createJavaProject(AbstractJavaModelTests.java:1969)
	at org.eclipse.jdt.core.tests.model.AbstractJavaModelTests.createJavaProject(AbstractJavaModelTests.java:1765)
	at org.eclipse.jdt.core.tests.model.ClasspathTests.testClasspathValidation16(ClasspathTests.java:1669)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at junit.framework.TestCase.runTest(TestCase.java:177)
	at org.eclipse.jdt.core.tests.junit.extension.TestCase.runTest(TestCase.java:970)
	at junit.framework.TestCase.runBare(TestCase.java:142)
	at junit.framework.TestResult$1.protect(TestResult.java:122)
	at junit.framework.TestResult.runProtected(TestResult.java:142)
	at junit.framework.TestResult.run(TestResult.java:125)
	at junit.framework.TestCase.run(TestCase.java:130)
	at junit.framework.TestSuite.runTest(TestSuite.java:241)
	at org.eclipse.jdt.core.tests.model.SuiteOfTestCases$Suite.runTest(SuiteOfTestCases.java:114)
	at junit.framework.TestSuite.run(TestSuite.java:236)
[...]

PR #4904, opened 2026-03-06
Wait for auto-refresh at ClasspathTests.tearDown()

trancexpress:
Fixes: #4903


trancexpress:
This does seem to make tests noticeably slower...

trancexpress:
I wonder what will happen if we disable auto-refreshing altogether, for JDT core tests...

Seems like we can do this with this change in `AbstractJavaModelTests.setUpSuite()`:

```
	@Override
		super.setUpSuite();
		IEclipsePreferences resourcesPreferences = InstanceScope.INSTANCE.getNode(ResourcesPlugin.PI_RESOURCES);
		resourcesPreferences.putBoolean(ResourcesPlugin.PREF_AUTO_REFRESH, false);
		resourcesPreferences.putBoolean(ResourcesPlugin.PREF_LIGHTWEIGHT_AUTO_REFRESH, false);
		resourcesPreferences.flush();
```

I'm not sure its a good idea though... Both are enabled per default and we likely should test with default settings... If the tests are having trouble with refreshes, likely we should just wait. Even if that costs extra time...

Issue #4915, opened 2026-03-10
Enable verbose mode for JavaSearchBugsTests2


trancexpress:
See: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/421

Its not clear what the problem is, the test seems to properly wait on the indexer. We can enable verbose mode, maybe that helps to narrow down the cause. The extra infos look like this:

```
17:53:59.784 SETUP JavaSearchBugsTests2.testBug483650_0001
# !SESSION 2026-03-10 17:53:59.789
# version: 1.1
# verbose: true
# The following option strings are specified for this debug session:
# 	null
| main | 2026-03-10 17:53:59.789 | org.eclipse.jdt.core | /debug | org.eclipse.jdt.internal.core.search.processing.JobManager | performConcurrentJob | 265 | STARTING  concurrent job - searching ClassDeclarationPattern: pkg<*>, enclosing<*>, type<!@$#!@>, exact match, case sensitive, generic full match, fine grain: none |
| main | 2026-03-10 17:53:59.797 | org.eclipse.jdt.core | /debug | org.eclipse.jdt.internal.core.search.IndexSelector | getIndexLocations | 324 | -> selected 0 indexes out of total indexes 0 after qualify filtering - org.eclipse.jdt.internal.core.search.IndexSelector@69a5c6be |
| main | 2026-03-10 17:53:59.799 | org.eclipse.jdt.core | /debug | org.eclipse.jdt.internal.core.search.indexing.IndexManager | hasJavaLikeNamesChanged | 634 | No Java like names found and there is atleast one non-default javaLikeName |
| main | 2026-03-10 17:53:59.800 | org.eclipse.jdt.core | /debug | org.eclipse.jdt.internal.core.search.IndexSelector | getIndexLocations | 324 | -> selected 0 indexes out of total indexes 0 after qualify filtering - org.eclipse.jdt.internal.core.search.IndexSelector@69a5c6be |
| main | 2026-03-10 17:53:59.802 | org.eclipse.jdt.core | /debug | org.eclipse.jdt.internal.core.search.PatternSearchJob | execute | 125 | -> execution time: 1ms - searching ClassDeclarationPattern: pkg<*>, enclosing<*>, type<!@$#!@>, exact match, case sensitive, generic full match, fine grain: none |
[...]

Issue #2876, opened 2024-08-30
Failure in org.eclipse.jdt.core.tests.model.JavaSearchBugsTests. testBug339891


mpalat:
ref: https://download.eclipse.org/eclipse/downloads/drops4/I20240828-1820/testresults/html/org.eclipse.jdt.core.tests.model_ep433I-unit-macM1-java17_macosx.cocoa.aarch64_17.html
Unexpected search results. ----------- Expected ------------ Test.java void Test.foo(Ref) [foo()] EXACT_MATCH ------------ but was ------------ ---------------------- ---------- expected:<[Test.java void Test.foo(Ref) [foo()] EXACT_MATCH]> but was:<[]>

junit.framework.ComparisonFailure: Unexpected search results.
----------- Expected ------------
Test.java void Test.foo(Ref) [foo()] EXACT_MATCH
------------ but was ------------

---------------------- ----------
expected:<[Test.java void Test.foo(Ref) [foo()] EXACT_MATCH]> but was:<[]>
at org.eclipse.jdt.core.tests.junit.extension.TestCase.assertStringEquals(TestCase.java:267)
at org.eclipse.jdt.core.tests.junit.extension.TestCase.assertEquals(TestCase.java:242)
at org.eclipse.jdt.core.tests.model.AbstractJavaSearchTests.assertSearchResults(AbstractJavaSearchTests.java:948)
at org.eclipse.jdt.core.tests.model.AbstractJavaSearchTests.assertSearchResults(AbstractJavaSearchTests.java:906)
at org.eclipse.jdt.core.tests.model.AbstractJavaSearchTests.assertSearchResults(AbstractJavaSearchTests.java:903)
at org.eclipse.jdt.core.tests.model.JavaSearchBugsTests.testBug339891(JavaSearchBugsTests.java:12939)
at java.base/java.lang.reflect.Method.invoke(Method.java:568)
at junit.framework.TestCase.runTest(TestCase.java:177)
at org.eclipse.jdt.core.tests.junit.extension.TestCase.runTest(TestCase.java:972)
at junit.framework.TestCase.runBare(TestCase.java:142)
at junit.framework.TestResult$1.protect(TestResult.java:122)
at junit.framework.TestResult.runProtected(TestResult.java:142)
at junit.framework.TestResult.run(TestResult.java:125)
at junit.framework.TestCase.run(TestCase.java:130)
at junit.framework.TestSuite.runTest(TestSuite.java:241)
[...]

PR #4916, opened 2026-03-10
Enable verbose mode for JavaSearchBugsTests and JavaSearchBugsTests2

trancexpress:
Fixes: #4915


trancexpress:
See also https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2876, seems like we can use verbose output also for `JavaSearchBugsTests`.

PR #4934, opened 2026-03-14
Cache PackageFragment.getElementName() to avoid repeated String creation

### LINKED ISSUE #4930: `PackageFragment.getElementName()` creates a new String on every call
During analysis of two heapdumps shared [here](https://github.com/eclipse-pde/eclipse.pde/pull/2253#issuecomment-4054961583) the following issue was discovered as a hotspot that can benefit from optimization:

### Performance Data

| Metric | WITH transitive | WITHOUT transitive | Ratio |
|--------|----------------:|-------------------:|------:|
| getElementName() self time (µs) | 3,463,714 | 1,325,883 | **2.6×** |
| toStringInfo() self time (µs) | 1,555,866 | 504,270 | **3.1×** |
| Util.concatWith() (µs) | related | related | — |
| StringBuilder.ensureCapacity (µs) | 9,432,131 | 4,138,082 | **2.3×** |

### Description

`PackageFragment.getElementName()` concatenates the `names` array with a dot separator on **every invocation**:

```java
public String getElementName() {
    if (this.names.length == 0)
        return DEFAULT_PACKAGE_NAME;
    return Util.concatWith(this.names, '.'); // creates new StringBuilder + String each time
}
```

Where `Util.concatWith()` creates a new `StringBuilder` and iterates:

```java
public static final String concatWith(String[] array, char separator) {
    StringBuilder buffer = new StringBuilder();
    for (int i = 0, length = array.length; i < length; i++) {
        buffer.append(array[i]);
        if (i < length - 1) buffer.append(separator);
    }
    return buffer.toString();
}
```

The `PackageFragment.names` array is immutable after construction. With transitive dependencies, there are far more `PackageFragment` objects, and `getElementName()` is called repeatedly during classpath traversal, dependency resolution, and builder operations. The per-thread data shows one worker thread accumulated significant time in `toStringInfo()` → `getElementName()` → string building.

### Suggested Fix

Cache the concatenated element name as a field in `PackageFragment`:

```java
private volatile String cachedElementName;

public String getElementName() {
    if (this.names.length == 0)
        return DEFAULT_PACKAGE_NAME;
    String name = this.cachedElementName;
    if (name == null) {
        name = Util.concatWith(this.names, '.');
        this.cachedElementName = name;
    }
    return name;
}
```

laeubi:
PackageFragment.getElementName() was creating a new String via Util.concatWith() on every invocation, despite the names array being immutable after construction. This adds a volatile cached field using the racy single-check idiom (safe for immutable String values).

Fix https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4930

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/PackageFragment.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/PackageFragment.java
@@ -56,6 +56,8 @@ public class PackageFragment extends Openable implements IPackageFragment, Suffi
 
 	private final boolean isValidPackageName;
 
+	private volatile String cachedElementName;
+
 protected PackageFragment(PackageFragmentRoot root, String[] names) {
 	super(root);
 	this.names = names;
@@ -322,7 +324,12 @@ public ICompilationUnit[] getCompilationUnits(WorkingCopyOwner owner) {
 public String getElementName() {
 	if (this.names.length == 0)
 		return DEFAULT_PACKAGE_NAME;
-	return Util.concatWith(this.names, '.');
+	String name = this.cachedElementName;
+	if (name == null) {
+		name = Util.concatWith(this.names, '.');
+		this.cachedElementName = name;
+	}
+	return name;
 }
 /**
  * @see IJavaElement


fedejeanne (commented):
Given that `names` is public and there is the possibility that outside callers modify it **after** it has been provided to the constructor of this class, I think this approach (calculating the value in the getter) is the safest choice.

If the field would have been `private` and a defensive copy of the parameter would have been used instead then one could have simply calculated the `(cached)elementName` in the constructor.

laeubi:
Yep the data-model is a bit "wide open" at jdt so *theoretically* it could change (but hard to prove it does not).

laeubi:
@stephan-herrmann @iloveeclipse can someone review and probably merge?

laeubi:
@jarthana can you take a look here as well?

jarthana (approved):

PR #4939, opened 2026-03-17
Cache isTest/isWithoutTestCode in ClasspathEntry constructor

### LINKED ISSUE #4927: `IClasspathEntry.isTest()` iterates extra attributes array on every call and can become a perfromance hot-spot
During analysis of two heapdums shared [here](https://github.com/eclipse-pde/eclipse.pde/pull/2253#issuecomment-4054961583) the following issue was discovered as a hotspot that can benefit from optimization:

### Performance Data

| Metric | WITH transitive | WITHOUT transitive | Ratio |
|--------|----------------:|-------------------:|------:|
| Self time (µs) | 2,233,733 | 366,222 | **6.1×** |

### Description

`IClasspathEntry.isTest()` is a `default` interface method that iterates over the entire `getExtraAttributes()` array on every invocation:

```java
default public boolean isTest() {
    for (IClasspathAttribute attribute : getExtraAttributes()) {
        if (IClasspathAttribute.TEST.equals(attribute.getName()) && "true".equals(attribute.getValue()))
            return true;
    }
    return false;
}
```

This method is called on **every classpath entry** during classpath expansion and package fragment root computation. With transitive dependencies, the number of classpath entries grows, and `isTest()` is called on each one — often multiple times during the same build phase. The same pattern exists for `isWithoutTestCode()`.

### Suggested Fix

Override `isTest()` in `ClasspathEntry` (the concrete implementation) to cache the result. The extra attributes are immutable after construction, so the result can be computed once and stored:

```java
// In ClasspathEntry.java
private int cachedIsTest = -1; // -1 = not computed, 0 = false, 1 = true

@Override
public boolean isTest() {
    if (cachedIsTest == -1) {
        cachedIsTest = IClasspathEntry.super.isTest() ? 1 : 0;
    }
    return cachedIsTest == 1;
}
```

Alternatively, compute `isTest` and `isWithoutTestCode` in the `ClasspathEntry` constructor and store as boolean fields, since these are queried on almost every entry.

laeubi:
Override `isTest()` and `isWithoutTestCode()` in `ClasspathEntry` to avoid iterating over the extra attributes array on every call. Instead, compute and cache the boolean values in the constructor, since extra attributes are immutable after construction.

Fixes eclipse-jdt/eclipse.jdt.core#4927

### DIFF
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/ClasspathEntry.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/ClasspathEntry.java
@@ -244,6 +244,9 @@ public class ClasspathEntry implements IClasspathEntry {
 	 */
 	public IClasspathAttribute[] extraAttributes;
 
+	private final boolean isTest;
+	private final boolean isWithoutTestCode;
+
 	public ClasspathEntry(
 			int contentKind,
 			int entryKind,
@@ -325,6 +328,18 @@ public class ClasspathEntry implements IClasspathEntry {
 		this.combineAccessRules = combineAccessRules;
 		this.extraAttributes = extraAttributes.length > 0 ? extraAttributes : NO_EXTRA_ATTRIBUTES;
 
+		String test = null;
+		String withoutTestCode = null;
+		for (IClasspathAttribute attribute : this.extraAttributes) {
+			if (IClasspathAttribute.TEST.equals(attribute.getName())) {
+				test = attribute.getValue();
+			} else if (IClasspathAttribute.WITHOUT_TEST_CODE.equals(attribute.getName()) ) {
+				withoutTestCode = attribute.getValue();
+			}
+		}
+		this.isTest = Boolean.valueOf(test);
+		this.isWithoutTestCode = Boolean.valueOf(withoutTestCode);
+
 	    if (inclusionPatterns != INCLUDE_ALL && inclusionPatterns.length > 0) {
 			this.fullInclusionPatternChars = UNINIT_PATTERNS;
 	    }
@@ -335,6 +350,16 @@ public class ClasspathEntry implements IClasspathEntry {
 		this.sourceAttachmentRootPath = sourceAttachmentRootPath;
 		this.specificOutputLocation = specificOutputLocation;
 		this.isExported = isExported;
+	}
+
+	@Override
+	public boolean isTest() {
+		return this.isTest;
+	}
+
+	@Override
+	public boolean isWithoutTestCode() {
+		return this.isWithoutTestCode;
 	}
 
 	@Override
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/core/IClasspathEntry.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/core/IClasspathEntry.java
@@ -475,7 +475,7 @@ public interface IClasspathEntry {
 	IClasspathEntry getResolvedEntry();
 
 	/**
-	 * This is a convience method, that returns <code>true</code> if the extra attributes contain an attribute whose name
+	 * This is a convenience method, that returns <code>true</code> if the extra attributes contain an attribute whose name
 	 * is {@link IClasspathAttribute#TEST} and whose value is 'true'.
 	 *
 	 * @see #getExtraAttributes()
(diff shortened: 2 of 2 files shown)


org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/ClasspathEntry.java
```diff
@@ -325,6 +328,18 @@ public ClasspathEntry(
...
+		String withoutTestCode = null;
+		for (IClasspathAttribute attribute : this.extraAttributes) {
+			if (IClasspathAttribute.TEST.equals(attribute.getName())) {
+				test = attribute.getValue();
+			} else if (IClasspathAttribute.WITHOUT_TEST_CODE.equals(attribute.getName()) ) {
+				withoutTestCode = attribute.getValue();
+			}
+		}
+		this.isTest = Boolean.valueOf(test);
+		this.isWithoutTestCode = Boolean.valueOf(withoutTestCode);
```
fedejeanne, line 341: I would extract it to a method for better readability and to avoid duplicated code:

```suggestion
		this.test = hasAttribute(IClasspathAttribute.TEST);
		this.withoutTestCode = hasAttribute(IClasspathAttribute.WITHOUT_TEST_CODE);
```

---

```java
private boolean hasAttribute(String attributeName) {
	for (IClasspathAttribute attribute : this.extraAttributes) {
		if (attributeName.equals(attribute.getName())) { 
			return Boolean.valueOf(attribute.getValue());
		}
	}
	return false;
}

@Override
public boolean isTest() {
	return this.test;
}

@Override
public boolean isWithoutTestCode() {
	return this.withoutTestCode;
}
```

fedejeanne (commented):
Thank you for providing a PR! I have a minor improvement proposal.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/ClasspathEntry.java
laeubi: > I would extract it to a method for better readability and to avoid duplicated code

But this would iterate the array two times, so even not so nice it is more performant.

fedejeanne: It is the same amount of instructions since 1 iteration with 2 `if`s is the same as 2 iterations with 1 `if` each. Same `O(n)`

laeubi: I think we are splitting hairs here an O-notation is only valid for large sample sets not for small numbers like here... so before we talk about cache coherence, branch prediction or whatever while actually usually one has usually no attributes or only one or two a *method call* is usually more expensive than a local loop in all cases (you can look at the bytecode if you like) and count instructions ... in any case nothing will really likely make a difference and I don't mind any way just wanted to get this concern of the table.

Just for the sake of completeness:

Assume you have an array with 5 items and has one loop you need:

- one assignment to a local variable
- have 5 increments
- have 5 comparisons for the loop condition 
- have 10 if checks

-> 21 instructions roughly

Using a method involves:

- putting 2 arguments on the stack
- one assignment to a local variable
- have 5 increments
- have 5 comparisons for the loop condition 
- have 5 if checks
- return the value

-> 19 *2 = 38 instructions roughly so almost 100% increase

fedejeanne: If we're talking about small numbers then I favor readability and maintainability over speed. Code duplication is known to cause problems, hence: avoid it.

laeubi: I fail to understand why my proposal should be less readable or duplicating any code (except for one `Boolean.valueOf(...)`).

Second I though the whole complains where about **perfromance** so I choosed an implementation that is both easy to understand and allows the JIT for easy optimization (e.g. completely local and side-effect free).

In any case I'm not a JDT maintainer, the change is trivial enough and I would leave it to whoever cares to merge this to maybe came up with a ~better~ different solution.

fedejeanne: Agreed, this is the kind of discussions we shouldn't put excessive effort in. Let's let the maintainer make the call :-)

I consider this discussion to be settled (but I can't click the "resolve" button, I don't see it).

jarthana (approved):
@laeubi This point has nothing to do your patch, but can you fix the typo(spelling) of "convience" of the two APIs in question in IClasspathAttribute?

laeubi:
> @laeubi This point has nothing to do your patch, but can you fix the typo(spelling) of "convience" of the two APIs in question in IClasspathAttribute?

I'll take a look, but yes it should be possible ;-)

laeubi:
@jarthana I hope it is as desired now.

Issue #4955, opened 2026-03-18
ECJ generates wrong method descriptor for calls to native varargs methods


kaluchi:
## Summary

ECJ incorrectly treats `native` varargs methods as signature-polymorphic, generating method descriptors with actual argument types instead of `Object[]`. This causes `NoSuchMethodError` at runtime.

Non-native varargs methods with the same signature are compiled correctly.

## Minimal reproduction

```java
public class NativeVarargs {

    // native varargs — ECJ generates wrong descriptor for callers
    public static native void debug(Object... args);

    // non-native varargs — ECJ generates correct descriptor
    public static void debugRegular(Object... args) {
    }

    public static void main(String[] args) {
        // These two calls should produce identical bytecode descriptors:
        debug("hello");           // BUG: ECJ generates debug(String)V
        debugRegular("hello");    // OK:  ECJ generates debugRegular([Object)V

        debug("hello", 42);      // BUG: ECJ generates debug(String,int)V
        debugRegular("hello", 42); // OK: ECJ generates debugRegular([Object)V
    }
}
```

## Actual bytecode (javap -c)

```
2: invokestatic  #23  // Method debug:(Ljava/lang/String;)V          ← WRONG
14: invokestatic  #26  // Method debugRegular:([Ljava/lang/Object;)V  ← correct

21: invokestatic  #28  // Method debug:(Ljava/lang/String;I)V         ← WRONG
41: invokestatic  #26  // Method debugRegular:([Ljava/lang/Object;)V  ← correct
```

## Expected bytecode

Both `debug` and `debugRegular` calls should use descriptor `([Ljava/lang/Object;)V`.

## Runtime result

```
Exception in thread "main" java.lang.NoSuchMethodError:
  'void NativeVarargs.debug(java.lang.String)'
```

## Root cause analysis

The issue appears to originate in `MethodBinding.hasPolymorphicSignature()` ([source](https://github.com/eclipse-jdt/eclipse.jdt.core/blob/master/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/MethodBinding.java#L1436-L1455)):

```java
public boolean hasPolymorphicSignature(Scope scope) {
    if ((this.tagBits & TagBits.AnnotationPolymorphicSignature) != 0) {
        return true;
    }
    if (this.isNative() && this.isVarargs() && this.parameters.length == 1) {
[...]

PR #4956, opened 2026-03-18
Fix native varargs methods incorrectly treated as signature-polymorphic #4955

kaluchi:
Regression from #147: `hasPolymorphicSignature()` used `id`-based comparison to check if a method's declaring class is `MethodHandle` or `VarHandle`. Since neither type has a pre-assigned `TypeId`, both default to `id = 0` — same as any user-defined class. The comparison `0 == 0` caused false positives.

**Effect:** Any `static native void foo(Object... args)` method on any class was treated as signature-polymorphic. ECJ generated `invokestatic` with direct argument types (e.g. `debug(String)`) instead of the varargs descriptor (`debug(Object[])`), causing `NoSuchMethodError` at runtime.

**Fix:** Replace `declaringClassLocal.id == scope.getJavaLangInvokeMethodHandle().id` with `TypeBinding.equalsEquals(declaringClassLocal, scope.getJavaLangInvokeMethodHandle())`.

**Regression test:** `BatchCompilerTest2.testGH4955()` — compiles a class with `static native void debug(Object... args)`, verifies the disassembled bytecode uses `Object[]` descriptor.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4955

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/MethodBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/MethodBinding.java
@@ -1437,21 +1437,10 @@ public boolean hasPolymorphicSignature(Scope scope) {
 	if ((this.tagBits & TagBits.AnnotationPolymorphicSignature) != 0) {
 		return true;
 	}
-	if (this.isNative()	&& this.isVarargs() && this.parameters.length == 1) {
-		/*
-		 *  here type will be arrayType we will come here only if the method is of type
-		 *  varargs(represented by arraytype) and with only one parameter.
-		 */
-		if (this.parameters[0].leafComponentType().id == TypeIds.T_JavaLangObject) {
-			ReferenceBinding declaringClassLocal = this.declaringClass;
-			if ((declaringClassLocal != null) && (declaringClassLocal.id == scope.getJavaLangInvokeMethodHandle().id
-					|| declaringClassLocal.id == scope.getJavaLangInvokeVarHandle().id)) {
-				return true;
-			}
-		}
-	}
-
-	return false;
+	return this.isNative() && this.isVarargs() && this.parameters.length == 1 &&
+			this.parameters[0].leafComponentType().id == TypeIds.T_JavaLangObject &&
+				(TypeBinding.equalsEquals(this.declaringClass, scope.getJavaLangInvokeMethodHandle())
+						|| TypeBinding.equalsEquals(this.declaringClass, scope.getJavaLangInvokeVarHandle()));
 }
 public boolean isClosingMethod() {
 	boolean isCloseMethod = CharOperation.equals(this.selector, TypeConstants.CLOSE) && this.parameters == NO_PARAMETERS;  // close()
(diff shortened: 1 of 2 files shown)


srikanth-sankaran:
> Regression from #147: `hasPolymorphicSignature()` used `id`-based comparison to check if a method's declaring class is `MethodHandle` or `VarHandle`. Since neither type has a pre-assigned `TypeId`, both default to `id = 0` — same as any user-defined class. The comparison `0 == 0` caused false positives.

These assertions seem faulty. Types start with a default value of `org.eclipse.jdt.internal.compiler.lookup.TypeIds.NoId` and not with a value of `0` and get stamped with a _well known type id_ or a highwater mark id when _some derivation happens_.

In the provided regression test (without the fix) I see `MethodHandle` has an id of 156 and actually fails the == comparison which then ends up succeeding with the `VarHandle` comparison as both the declaring class and `VarHandle` end up having `NoId`.

Certainly the original code is suspect - I don't know why a direct comparison of `id` should be attempted at all. Type comparisons can/should be done with `==/!=` for identity comparisons and `equalsEquals/notEquals` for annotation agnostic comparisons.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/MethodBinding.java
```diff
@@ -1444,8 +1444,8 @@ public boolean hasPolymorphicSignature(Scope scope) {
 		 */
 		if (this.parameters[0].leafComponentType().id == TypeIds.T_JavaLangObject) {
 			ReferenceBinding declaringClassLocal = this.declaringClass;
-			if ((declaringClassLocal != null) && (declaringClassLocal.id == scope.getJavaLangInvokeMethodHandle().id
-					|| declaringClassLocal.id == scope.getJavaLangInvokeVarHandle().id)) {
+			if ((declaringClassLocal != null) && (TypeBinding.equalsEquals(declaringClassLocal, scope.getJavaLangInvokeMethodHandle())
+					|| TypeBinding.equalsEquals(declaringClassLocal, scope.getJavaLangInvokeVarHandle()))) {
```
srikanth-sankaran: The null check is noise and can be elided.

Please remove the original comment: It is just noise with no value. 

```
/*
		 *  here type will be arrayType we will come here only if the method is of type
		 *  varargs(represented by arraytype) and with only one parameter.
		 */
```

How about the method simply reading:

```
public boolean hasPolymorphicSignature(Scope scope) {
	if ((this.tagBits & TagBits.AnnotationPolymorphicSignature) != 0) {
		return true;
	}
	return this.isNative() && this.isVarargs() && this.parameters.length == 1 &&
				this.parameters[0].leafComponentType().id == TypeIds.T_JavaLangObject &&
					(TypeBinding.equalsEquals(this.declaringClass, scope.getJavaLangInvokeMethodHandle())
							|| TypeBinding.equalsEquals(this.declaringClass, scope.getJavaLangInvokeVarHandle()));
}
```

This looks much cleaner to me. I didn't run the tests

srikanth-sankaran:
Thanks for your efforts here @kaluchi !

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/MethodBinding.java
srikanth-sankaran: The set check for the bit `AnnotationPolymorphicSignature` could also be pushed into the return statement, but that perhaps degrades readability.

kaluchi:
@srikanth-sankaran Thanks for the quick feedback! Applied the suggested cleanup.

srikanth-sankaran:
@jarthana - Could you look through too and take it forward ? TIA.

PR #4961, opened 2026-03-20
Fix unchecked conversion not failed compilation

### LINKED ISSUE #4957: Eclipse 2026-03 compiles fine but Java does not: unchecked conversion
I just discovered this in a broken build today. 

```
Eclipse IDE for Eclipse Committers (includes Incubating components)

Version: 2026-03 (4.39.0)
Build id: 20260305-0817

```
## Reproducer

Compiles fine in Eclipse 2026-03:

```java
import java.util.Comparator;
import java.util.Map;

public class Snippet {

        Comparator<Map<String, Object>> eventComparator = Comparator
            .<Map<String, Object>, Comparable>comparing(e -> (Comparable) e.get("event_date"),
                Comparator.nullsLast(Comparator.naturalOrder()))
            .thenComparing(e -> (String) e.get("event_type"));
    }

}

```

**but Fails in Java:**

```
java --version
openjdk 21.0.7 2025-04-15 LTS
OpenJDK Runtime Environment Temurin-21.0.7+6 (build 21.0.7+6-LTS)
OpenJDK 64-Bit Server VM Temurin-21.0.7+6 (build 21.0.7+6-LTS, mixed mode, sharing)

java -Duser.language=en Snippet.java

Snippet.java:12: warning: [unchecked] unchecked method invocation: method nullsLast in interface Comparator is applied to given types
                Comparator.nullsLast(Comparator.naturalOrder()))
                                    ^
  required: Comparator<? super T>
  found:    Comparator<Comparable>
  where T is a type-variable:
    T extends Object declared in method <T>nullsLast(Comparator<? super T>)
Snippet.java:11: warning: [unchecked] unchecked method invocation: method comparing in interface Comparator is applied to given types
            .<Map<String, Object>, Comparable>comparing(e -> (Comparable) e.get("event_date"),
                                                       ^
  required: Function<? super T,? extends U>,Comparator<? super U>
  found:    Function<Map<String,Object>,Comparable>,Comparator
  where T,U are type-variables:
    T extends Object declared in method <T,U>comparing(Function<? super T,? extends U>,Comparator<? super U>)
    U extends Object declared in method <T,U>comparing(Function<? super T,? extends U>,Comparator<? super U>)
Snippet.java:12: warning: [unchecked] unchecked conversion
                Comparator.nullsLast(Comparator.naturalOrder()))
                                    ^
  required: Comparator<? super U>
  found:    Comparator
  where U,T are type-variables:
    U extends Object declared in method <T,U>comparing(Function<? super T,? extends U>,Comparator<? super U>)
    T extends Object declared in method <T,U>comparing(Function<? super T,? extends U>,Comparator<? super U>)
Snippet.java:13: warning: [unchecked] unchecked call to <U>thenComparing(Function<? super T,? extends U>) as a member of the raw type Comparator
            .thenComparing(e -> (String) e.get("event_type"));
                          ^
  where U,T are type-variables:
    U extends Comparable<? super U> declared in method <U>thenComparing(Function<? super T,? extends U>)
    T extends Object declared in interface Comparator
Snippet.java:13: error: cannot find symbol
            .thenComparing(e -> (String) e.get("event_type"));
                                          ^
  symbol:   method get(String)
  location: variable e of type Object
Snippet.java:13: warning: [unchecked] unchecked conversion
            .thenComparing(e -> (String) e.get("event_type"));
                          ^
  required: Comparator<Map<String,Object>>
  found:    Comparator
1 error
5 warnings
error: compilation failed
```

## Fix

If I rewrite it to something like this, then it compiles in Eclipse and also in Java:

```java
import java.util.Comparator;
import java.util.Map;

public class Snippet {

        Comparator<Object> nullableComparableObjectComparator = Comparator.nullsLast((a,
            b) -> ((Comparable<Object>) a).compareTo(b));

        Comparator<Map<String, Object>> eventComparator = Comparator
            .comparing((Map<String, Object> e) -> e.get("event_date"),
                nullableComparableObjectComparator)
            .thenComparing((Map<String, Object> e) -> (String) e.get("event_type"),
                Comparator.nullsLast(String::compareTo));
    }

}
```

```
java -Duser.language=en Snippet.java
Snippet.java:11: warning: [unchecked] unchecked cast
            b) -> ((Comparable<Object>) a).compareTo(b));
                                        ^
  required: Comparable<Object>
  found:    Object
1 warning
```

## Settings

JRE: points to a `jdk-21.0.7+6/Contents/Home`

Image

chrisrueger:
Closes #4957

## What it does

- Detect when any explicit type argument is a raw type and then fail compilation (to achieve same compilation error as javac)
- This seems to be related to https://docs.oracle.com/javase/specs/jls/se21/html/jls-15.html#jls-15.12.2.6

> If unchecked conversion was necessary for the method to be applicable, then the invocation type's parameter types are obtained by applying the substitution [P1:=T1, ..., Pp:=Tp] to the parameter types of the method's type, and the invocation type's return type and thrown types are given by the erasure of the return type and thrown types of the method's type.

- Adds regression test testGH4957 reproducing the Comparator.comparing + raw Comparable scenario  With this fix testGH4957 passes.

## How to test

- run `org.eclipse.jdt.core.tests.compiler.regression.GenericsRegressionTest_9.testGH4957()`  without the fix in `org.eclipse.jdt.internal.compiler.lookup.ParameterizedGenericMethodBinding.computeCompatibleMethod(MethodBinding, TypeBinding[], Scope, InvocationSite)`

or just use e.g. Eclipse 2025-03 (or even 2026-06 (4.40) Build id: I20260319-1800 master) and paste the the following class:

```
import java.util.Comparator;
import java.util.Map;

public class Snippet {

        Comparator<Map<String, Object>> eventComparator = Comparator
            .<Map<String, Object>, Comparable>comparing(e -> (Comparable) e.get("event_date"),
                Comparator.nullsLast(Comparator.naturalOrder()))
            .thenComparing(e -> (String) e.get("event_type"));
    }

}

```

Expected Result: Compilation should fail in Eclipse
Actual result:  Compiliation succeeds. 

But in javacc compilation fails. 

With this PR compilation fails also in Eclipse. 

But note: Eclipse seems to error with a different message than javac. Not sure this is a deeper problem. But at least it fails at the same place. 

ECJ fails with:
```
"5. ERROR in Snippet.java (at line 11)\n" +
			"	.thenComparing(e -> (String) e.get(\"event_type\"));\n" +
			"	                               ^^^\n" +
			"The method get(String) is undefined for the type Object\n" +
```

while javac failed with:

```
Snippet.java:13: error: cannot find symbol
            .thenComparing(e -> (String) e.get("event_type"));
                                          ^
  symbol:   method get(String)
  location: variable e of type Object
Snippet.java:13: warning: [unchecked] unchecked conversion
            .thenComparing(e -> (String) e.get("event_type"));
                          ^
  required: Comparator<Map<String,Object>>
  found:    Comparator
1 error
5 warnings
error: compilation failed
```

## Disclaimer: 
I used AI to help me understand the issue and spec and find the places in the code. I have manually tested and debugged via latest JDT Dev environment via Oomph and ran my test in Eclipse.
This is my first step in the compiler space. Feel free to reject it or just use it as a base for a better fix if needed. 

## Author checklist

- [x] I have ~~thoroughly~~ tested my changes as good as I can

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ParameterizedGenericMethodBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ParameterizedGenericMethodBinding.java
@@ -26,6 +26,7 @@
  *								Bug 434483 - [1.8][compiler][inference] Type inference not picked up with method reference
  *								Bug 446442 - [1.8] merge null annotations from super methods
  *								Bug 457079 - Regression: type inference
+ *     Christoph Rueger         GH4957     - ECJ compiles fine but Java does not: unchecked conversion
  *******************************************************************************/
 package org.eclipse.jdt.internal.compiler.lookup;
 
@@ -75,8 +76,17 @@ public class ParameterizedGenericMethodBinding extends ParameterizedMethodBindin
 			        // incompatible due to wrong arity
 			        return new ProblemMethodBinding(originalMethod, originalMethod.selector, substitutes, ProblemReasons.TypeParameterArityMismatch);
 				}
-				methodSubstitute = environment.createParameterizedGenericMethod(originalMethod, substitutes);
-				break computeSubstitutes;
+				// JLS 15.12.2.6: Check if any explicit type argument is raw.
+				boolean usesUncheckedConversion = false;
+				for (int i = 0; i < typeVariables.length; i++) {
+					if (substitutes[i].isRawType()) {
+						usesUncheckedConversion = true;
+						break;
+					}
+				}
+				methodSubstitute = environment.createParameterizedGenericMethod(originalMethod, substitutes,
+						usesUncheckedConversion, false, null);
+			    break computeSubstitutes;
 			}
 			// perform type argument inference (15.12.2.7)
 			// initializes the map of substitutes (var --> type[][]{ equal, extends, super}
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericsRegressionTest_9.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericsRegressionTest_9.java
@@ -2295,6 +2295,68 @@ public void testGH4731() {
 	runner.javacTestOptions = JavacHasABug.JavacBug8016207;
 	runner.runNegativeTest();
 }
+/**
+ * https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4957Eclipse
+ * 2026-03 compiles fine but Java does not: unchecked conversion with Comparator.comparing + raw Comparable
+ */
+public void testGH4957() {
+	this.runNegativeTest(
+			new String[] {
+				"Snippet.java",
+				"import java.util.Comparator;\n"
+				+ "import java.util.Map;\n"
+				+ "\n"
+				+ "public class Snippet {\n"
+				+ "\n"
+				+ "    public static void main(String[] args) {\n"
+				+ "\n"
+				+ "        Comparator<Map<String, Object>> eventComparator = Comparator\n"
+				+ "            .<Map<String, Object>, Comparable>comparing(e -> (Comparable) e.get(\"event_date\"),\n"
+				+ "                Comparator.nullsLast(Comparator.naturalOrder()))\n"
+				+ "            .thenComparing(e -> (String) e.get(\"event_type\"));\n"
+				+ "    }\n"
+				+ "\n"
+				+ "}",
+			},
+			// related to https://docs.oracle.com/javase/specs/jls/se21/html/jls-15.html#jls-15.12.2.6
+			// Eclipse was incorrectly accepting this without warnings/errors;
+		    // javac reports unchecked conversion because raw Comparable is used
+		    // and the resulting Comparator is a raw Comparator (not Comparator<Map<String,Object>>),
+		    // causing thenComparing to operate on raw type Object.
+			"----------\n" +
+			"1. WARNING in Snippet.java (at line 8)\n" +
+			"	Comparator<Map<String, Object>> eventComparator = Comparator\n" +
+			"            .<Map<String, Object>, Comparable>comparing(e -> (Comparable) e.get(\"event_date\"),\n" +
+			"                Comparator.nullsLast(Comparator.naturalOrder()))\n" +
+			"            .thenComparing(e -> (String) e.get(\"event_type\"));\n" +
+			"	                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n" +
+			"Type safety: The method thenComparing(Function) belongs to the raw type Comparator. References to generic type Comparator<T> should be parameterized\n" +
+			"----------\n" +
+			"2. WARNING in Snippet.java (at line 8)\n" +
+			"	Comparator<Map<String, Object>> eventComparator = Comparator\n" +
+			"            .<Map<String, Object>, Comparable>comparing(e -> (Comparable) e.get(\"event_date\"),\n" +
+			"                Comparator.nullsLast(Comparator.naturalOrder()))\n" +
+			"            .thenComparing(e -> (String) e.get(\"event_type\"));\n" +
+			"	                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n" +
+			"Type safety: The expression of type Comparator needs unchecked conversion to conform to Comparator<Map<String,Object>>\n" +
+			"----------\n" +
+			"3. WARNING in Snippet.java (at line 9)\n" +
+			"	.<Map<String, Object>, Comparable>comparing(e -> (Comparable) e.get(\"event_date\"),\n" +
+			"	                       ^^^^^^^^^^\n" +
+			"Comparable is a raw type. References to generic type Comparable<T> should be parameterized\n" +
+			"----------\n" +
+			"4. WARNING in Snippet.java (at line 9)\n" +
+			"	.<Map<String, Object>, Comparable>comparing(e -> (Comparable) e.get(\"event_date\"),\n" +
+			"	                                                  ^^^^^^^^^^\n" +
+			"Comparable is a raw type. References to generic type Comparable<T> should be parameterized\n" +
+			"----------\n" +
+			"5. ERROR in Snippet.java (at line 11)\n" +
+			"	.thenComparing(e -> (String) e.get(\"event_type\"));\n" +
+			"	                               ^^^\n" +
+			"The method get(String) is undefined for the type Object\n" +
+			"----------\n");
+}
+
 public static Class<GenericsRegressionTest_9> testClass() {
 	return GenericsRegressionTest_9.class;
 }


venkateshkona3456-oss:
Hello,
I would like to work on this issue. Is it available?

Thank you!

stephan-herrmann:
Thanks for submitting.

A few preliminary remarks:

> Detect when any explicit type argument is a raw type and then fail compilation (to achieve same compilation error as javac)

Of course causality is much more indirect: detecting raw types in that position should not (and does not) directly cause compilation to fail. Did you understand _how_ the change triggers the compile error in this particular case? Perhaps adding another similar test which passes helps (now and for posterity) to understand the actual effect of the change?

> This seems to be related to https://docs.oracle.com/javase/specs/jls/se21/html/jls-15.html#jls-15.12.2.6

It would help to be more specific, stating which part of that section is not correctly implemented  in ecj.

Actually my own first reaction was: isn't all this already handled in type inference? (no, it is not).

> Expected Result: Compilation should fail in Eclipse
> Actual result: Compiliation succeeds.

In many cases of type checking it is not obvious, which way a compiler should behave. Comparing ecj to  javac gives a _hint_ but not a proof. Ideally, when a compiler accepts a program wrongly, this can be proven by crafting a program that is accepted by the compiler but throws an unexpected exception at run time. Do you think this kind of proof might be possible for this example? Is anything obviously wrong about the byte code that ecj generates?

> But note: Eclipse seems to error with a different message than javac.

The exact error messages are not standardized.  If you look closer, do both error messages tell different stories or the same story just with different words?

chrisrueger:
Thanks @stephan-herrmann I need to wrap my around this again. Will take some time, but do my best.

chrisrueger:
I did had a look at it and tried to understand your comment.

> > This seems to be related to https://docs.oracle.com/javase/specs/jls/se21/html/jls-15.html#jls-15.12.2.6
> 
> It would help to be more specific, stating which part of that section is not correctly implemented in ecj.
> 
> Actually my own first reaction was: isn't all this already handled in type inference? (no, it is not).

I think I meant the following section of the spec:

> If unchecked conversion was necessary for the method to be applicable, then the invocation type's parameter types are obtained by applying the substitution [P1:=T1, ..., Pp:=Tp] to the parameter types of the method's type, and the invocation type's return type and thrown types are given by the erasure of the return type and thrown types of the method's type.

> Did you understand how the change triggers the compile error in this particular case? 

maybe....maybe not...
I need to be honest here. As said, this is my first step this compiler space and I using AI to help me understand it. 

My interpretation is this:
After the change:
- `Comparator.<Map<String, Object>, Comparable>comparing(e -> (Comparable) e.get("event_date")`
- by JLS 15.12.2.6 the invocation return type is erased: meaning result type becomes raw `Comparator`, not `Comparator<Map<String,Object>>`
- `.thenComparing(...)` is invoked on raw Comparator
- lambda parameter `e` is treated as Object
- and `e.get(String)` does not exist on object
- Boom, error.

Before the change the `Comparator<Map<String,Object>> was **not** erased and carried into `.thenComparing(...)` so in ECJ thenComparing(...) thought: " hey...`e` is Map<String,Object>>` on which I can call `e.get(String)`)

But javac seems to be stricter.
Something like that. 

> Ideally, when a compiler accepts a program wrongly, this can be proven by crafting a program that is accepted by the compiler but throws an unexpected exception at run time.

I tried to construct a runtime-failing example. 
The snippet below is accepted by ECJ and fails with ClassCastException at runtime, but still **fails** compilation in javac.

```java
import java.util.Comparator;
import java.util.Map;

public class Snippet2 {

    	Comparator raw = Comparator
			        .<Map<String, Object>, Comparable>comparing(
			            e -> (Comparable) e.get("event_date"),
			            Comparator.nullsLast(Comparator.naturalOrder())).thenComparing(
                e -> (String)  e.get("event_type"));

            System.out.println(raw.compare("not a map", "also not a map"));
    }
}
```


```
Exception in thread "main" java.lang.ClassCastException: class java.lang.String cannot be cast to class java.util.Map (java.lang.String and java.util.Map are in module java.base of loader 'bootstrap')
	at java.base/java.util.Comparator.lambda$comparing$ea9a8b3a$1(Comparator.java:440)
	at java.base/java.util.Comparator.lambda$thenComparing$36697e65$1(Comparator.java:220)
	at org.eclipse.jdt.core.tests.compiler.regression.Snippet2.main(Snippet2.java:18)
```

Javac gives:

```
java -Duser.language=en /Users/christophrueger/jdt-master/git/eclipse.jdt.core/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/Snippet2.java

            e -> (String)  e.get("event_type"));
                            ^
  symbol:   method get(String)
  location: variable e of type Object
/Users/christophrueger/jdt-master/git/eclipse.jdt.core/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/Snippet2.java:15: warning: [unchecked] unchecked call to compare(T,T) as a member of the raw type Comparator
        System.out.println(raw.compare("not a map", "also not a map"));
                                      ^
  where T is a type-variable:
    T extends Object declared in interface Comparator
1 error
5 warnings
error: compilation failed
```

I am not sure this is what you were looking for.

This is not a good proof of the original bug in isolation, because the runtime failure also depends on explicit raw usage at the call site. But it does show that the unchecked/raw behavior in this area can lead to concrete runtime problems.

What do you think is a valid next step. Would you like to take over and just use my beginner thoughts as input?

stephan-herrmann:
Thanks for your response @chrisrueger 

Let me put the cards on the table: 

What I've seen so far looks very good. Still we need to build trust that the fix really fixes the core problem, and does so without causing grief elsewhere.

Several ways how this trust can be established:
1. You explain in clear words, where the current implementation is wrong, and what exactly is wrong about it. Then argue why the fix resolves exactly that wrong (and nothing else).
2. Or: Failing (2) I would have to repeat much of the background analysis to come to that explanation on my own.
3. No, I will not trust AI. I.e., no matter what "smart" things AI "suggested", some human in the loop must own the reasoning.

Now, let's see where we stand with your response:

> I think I meant the following section of the spec:
>> If unchecked conversion was necessary for the method to be applicable, then the invocation type's parameter types are obtained by applying the substitution [P1:=T1, ..., Pp:=Tp] to the parameter types of the method's type, and the invocation type's return type and thrown types are given by the erasure of the return type and thrown types of the method's type.

The weak part here is "I think I meant", the rest is good.

> As said, this is my first step this compiler space

You chose no simple code area, and hence no need to be shy if you are not 100% certain. Let's work it out together.

> My interpretation is this:
> After the change: ...

All that follows looks good. I'll still verify it in the debugger, but so far no complaints.

> I tried to construct a runtime-failing example.

This was a tremendous challenge, actually (I had briefly tried and failed).

> The snippet below is accepted by ECJ and fails with ClassCastException at runtime, but still fails compilation in javac.

The effect looks pretty similar to what I was looking for, but here the ClassCastException happens "before" the location where compilers disagree. See:
```java
import java.util.Comparator;
import java.util.Map;

public class Snippet3 {

    	Comparator raw = Comparator
			        .<Map<String, Integer>, Comparable>comparing(
			            e ->  e.get("event_date"));

            System.out.println(raw.compare("not a map", "also not a map"));
    }
}
```
This is accepted by both compilers and still throws the exact CCE you also observed. Who is to blame? Right, any use of raw types can produce CCE in location where the user didn't even write a cast.

So, the example did not succeed to distinguish between correct and incorrect compilation, which only means, we should probably focus on evidence by reasoning, rather then evidence by a witness.

> What do you think is a valid next step?

I'll take a closer look at the change both in the debugger and in relation to JLS. From there I'll either come back with more questions / requests, or I may find all to be in good shape and merge it right away. Stay tuned.

stephan-herrmann:
Erasing the return type under specific conditions is good, but in JLS the decision whether or not to apply erasure is more complex than what is implemented in this PR:

The PR asks if any explicit type argument is a raw type.

In JLS this involves the following chain of reasoning:
* 15.12.2.6: "If unchecked conversion was necessary for the method to be applicable"
* 15.12.2.2 (applicability): "ei is compatible in a strict invocation context with Fi[R1:=U1, ..., Rp:=Up] ([§5.3](https://docs.oracle.com/javase/specs/jls/se26/html/jls-5.html#jls-5.3))." (or _... loose invocation context ..._)
* 5.3: "If, after the conversions listed for an invocation context have been applied, the resulting type is a raw type ([§4.8](https://docs.oracle.com/javase/specs/jls/se26/html/jls-4.html#jls-4.8)), an unchecked conversion ([§5.1.9](https://docs.oracle.com/javase/specs/jls/se26/html/jls-5.html#jls-5.1.9)) may then be applied."

Let's apply it to the example:
* "the method" = `Comparator<T> comparing(Function<? super T,? extends U>, Comparator<? super U>)`
* arguments ei:
   * e1 = `e -> (Comparable) e.get("event_date")`
   * e2 = `Comparator.nullsLast(Comparator.naturalOrder())`
* substituted parameters Fi
   * F1 = `Function<? super Map<String,Object>,? extends Comparable#RAW>`
   * F2 = `Comparator<? super Comparable#RAW>`

The problem here is: the types of `e1` and `e2` are still unknown (`PolyTypeBinding`) when `computeCompatibleMethod()` is operating for `comparing()`. The type of the lambda is computed later in this call stack:
```
ASTNode.resolvePolyExpressionArguments0(Invocation, MethodBinding, TypeBinding[], BlockScope) line: 747	
ASTNode.resolvePolyExpressionArguments(Invocation, MethodBinding, TypeBinding[], BlockScope) line: 720	
MessageSend.findMethodBinding(BlockScope) line: 1132	
MessageSend.resolveType(BlockScope) line: 869	
```
Here we have a resolved binding for `comparing()` and then the parameter type `Function<? super Map<String,Object>,? extends Comparable#RAW>` serves as the target type for resolving the lambda. Resolving the lambda against that target type does not detect any problem. The mentioned target type is accepted as the lambda's resolved type. No unchecked conversion involved.

With these considerations I don't see how 15.12.2.2 plus 5.3 would detect any unchecked conversion affecting resolution of `comparing`. Yes, we have several raw type warnings, but no unchecked warning.

Also note, that javac issues an unchecked warning only against the chained `thenComparing()`, which is resolved only _after_ `comparing()` is completely dealt with, so this warning cannot influence resolution of `comparing()`.

stephan-herrmann:
Given the above analysis (with no indication that unchecked conversion is indeed involved)  I'm stepping down from reviewing the PR. If someone finds a complete and convincing reasoning showing that javac is right and ecj wrong, I might have another look.

Stepping down happens in the light of raw types being a legacy concept since Java 1.5. Also, javac is known to have bugs in this area. With these factors combined I don't give priority to fixing this.

Many thanks anyway, @chrisrueger for your first deep dive into compiler land. Maybe next time you find an issue that does not involve raw types :)

chrisrueger:
Thanks @stephan-herrmann for explaining in such detail and for taking the time to look at it... learned a lot. 

> Maybe next time you find an issue that does not involve raw types :)

I do my best 🤣

Issue #4964, opened 2026-03-21
Provide NameLookup.Answer check if type has forbidden access


trancexpress:
See: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2830

Looks like providing the flag `checkRestrictions=true` means `NameLookup.Answer` has the access rule, but the rule is not checked. We should add an option to check it.

PR #4965, opened 2026-03-21
Provide NameLookup.Answer check if type has forbidden access

trancexpress:
Calling `NameLookup.findType()` with flag `checkRestrictions=true` results in `NameLookup.Answer` containing the respective `AccessRestriction` in the resulting `NameLookup.Answer`.

This change adds a method `Answer.isForbidden()`,
which can be used to determine if the type in the `Answer` is forbidden according to the access rules of the types classpath entry.

Callers can then use this method to only use the found type, if its not forbidden by classpath access rules.

Such checks are useful in the context of:

https://github.com/eclipse-pde/eclipse.pde/issues/2244

The PDE classpath container contains transitive dependencies, with forbidden pattern `**/*`.

Fixes: #4964

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/NameLookupTests2.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/NameLookupTests2.java
@@ -425,5 +425,41 @@ public void testTransitionFromInvalidToValidJar() throws CoreException, IOExcept
 		deleteProject("P");
 	}
 }
+/**
+ * Test for: {@link org.eclipse.jdt.internal.core.NameLookup.Answer#isNonAccessible()}
+ */
+public void testNonAccessibleAccessAnswer() throws Exception {
+	try {
+		JavaProject project = (JavaProject)createJavaProject("P");
+		addLibrary(
+				project, "lib.jar", "libsrc.zip",
+				new String[] {
+					"p/A.java",
+					"""
+					package p;
+					public class A {}
+					""",
+				},
+				null,
+				null,
+				new String[] {"**/*"}, // all types in the jar should be inaccessible for this test
+				CompilerOptions.getFirstSupportedJavaVersion(),
+				null);
+		waitForAutoBuild();
+
+		NameLookup.Answer answer = getNameLookup(project).findType(
+				"p.A",
+				false /* no partial matches */,
+				NameLookup.ACCEPT_ALL,
+				false /* no secondary types */,
+				true /* wait for indexer */,
+				true /* check restrictions */,
+				null);
+		assertNotNull("Expected to find type", answer);
+		assertTrue("Expected type to be non-accessible", answer.isNonAccessible());
+	} finally {
+		deleteProject("P");
+	}
+}
 }
 
--- a/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/NameLookup.java
+++ b/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/NameLookup.java
@@ -33,6 +33,7 @@ import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.jdt.core.*;
 import org.eclipse.jdt.core.compiler.CharOperation;
+import org.eclipse.jdt.core.compiler.IProblem;
 import org.eclipse.jdt.internal.compiler.ast.ASTNode;
 import org.eclipse.jdt.internal.compiler.ast.TypeDeclaration;
 import org.eclipse.jdt.internal.compiler.env.AccessRestriction;
@@ -99,6 +100,14 @@ public class NameLookup implements SuffixConstants {
 			if (this.restriction == null) return true;
 			return otherAnswer.restriction != null
 				&& this.restriction.getProblemId() < otherAnswer.restriction.getProblemId();
+		}
+		/**
+		 * Returns whether the type in this answer is forbidden/non-accessible
+		 * according to access rules.
+		 * See: {@link IProblem#ForbiddenReference}, {@link org.eclipse.jdt.core.IAccessRule#K_NON_ACCESSIBLE}
+		 */
+		public boolean isNonAccessible() {
+			return this.restriction != null && this.restriction.getProblemId() == IProblem.ForbiddenReference;
 		}
 		@Override
 		public String toString() {


org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/NameLookup.java
```diff
@@ -100,6 +101,22 @@ public boolean isBetter(Answer otherAnswer) {
 			return otherAnswer.restriction != null
 				&& this.restriction.getProblemId() < otherAnswer.restriction.getProblemId();
 		}
+		/**
+		 * Returns whether the type in this answer is forbidden
+		 * according to access rules.
+		 * See: {@link IProblem#ForbiddenReference}
+		 */
+		public boolean isForbidden() {
+			if (this.restriction != null && this.restriction.getProblemId() == IProblem.ForbiddenReference) {
```
trancexpress: It might not hurt to add this to some utility class, I didn't find a central place to do this (the code is however taken from other places where the forbidden pattern is tested). Maybe we can add it to `AccessRestriction` even, something like `isForbidden(IType)`?

trancexpress: Seems that `AccessRestriction` can't know `IType`... I've adjusted what I could.

trancexpress:
@iloveeclipse anything against merging this? Running platform/JDT tests from Eclipse is still more difficult than it should be.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/NameLookup.java
```diff
@@ -100,6 +101,21 @@ public boolean isBetter(Answer otherAnswer) {
 			return otherAnswer.restriction != null
 				&& this.restriction.getProblemId() < otherAnswer.restriction.getProblemId();
 		}
+		/**
+		 * Returns whether the type in this answer is forbidden
+		 * according to access rules.
+		 * See: {@link IProblem#ForbiddenReference}
+		 */
+		public boolean isForbidden() {
```
iloveeclipse: I don't see any code in JDT that uses this in this PR and no other PR is linked. 
How this new method can fix https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4964?

trancexpress: See: https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2882

Screenshot_20260324_142224

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/NameLookup.java
```diff
@@ -100,6 +101,21 @@ public boolean isBetter(Answer otherAnswer) {
...
+			if (this.restriction != null) {
+				char[] packageName = this.type.getPackageFragment().getElementName().toCharArray();
+				char[][] packageChars = CharOperation.splitOn('.', packageName);
+				char[] fileWithoutExtension = this.type.getElementName().toCharArray();
+				char[] typeFilePath = CharOperation.concatWith(packageChars, fileWithoutExtension, '/');
+				return this.restriction.isForbidden(typeFilePath);
+			}
+			return false;
```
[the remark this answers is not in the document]
trancexpress: I'll check `$` and `.`, I'm not sure there.

A simple `restriction != null && restriction.getProblemId() == IProblem.ForbiddenReference` won't work since the pattern can be anything, it doesn't need to be the `**/*` from the PDE container.

Or actually, let me check when the restriction is added...

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/env/AccessRestriction.java
```diff
@@ -36,4 +39,12 @@ public int getProblemId() {
 	public boolean ignoreIfBetter() {
 		return this.accessRule.ignoreIfBetter();
 	}
+
+	public boolean isForbidden(char[] typeFilePath) {
+		if (this.accessRule.getProblemId() == IProblem.ForbiddenReference) {
+			char[] pattern = this.accessRule.pattern;
+			return CharOperation.pathMatch(pattern, typeFilePath, true/*case sensitive*/, '/');
+		}
+		return false;
```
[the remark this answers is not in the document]
trancexpress: I don't see how that is useful.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/NameLookup.java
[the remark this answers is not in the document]
trancexpress: Seems to be correct, the restriction is set only if the type is matched by the pattern.

org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/NameLookup.java
```diff
@@ -100,6 +101,14 @@ public boolean isBetter(Answer otherAnswer) {
 			return otherAnswer.restriction != null
 				&& this.restriction.getProblemId() < otherAnswer.restriction.getProblemId();
 		}
+		/**
+		 * Returns whether the type in this answer is forbidden
+		 * according to access rules.
+		 * See: {@link IProblem#ForbiddenReference}
+		 */
+		public boolean isForbidden() {
```
iloveeclipse: Two things: I've checked how the code uses `org.eclipse.jdt.core.compiler.IProblem.ForbiddenReference` and in most cases we don't use internal compiler "forbidden" language but translate internal compiler problems into tree different variants of:
- `IAccessRule.K_NON_ACCESSIBLE`
- `IAccessRule.K_DISCOURAGED`
- `IAccessRule.K_ACCESSIBLE`

See 

https://github.com/eclipse-jdt/eclipse.jdt.core/blob/fb1844c934409986d2ad332fce39f563cc210846/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/ClasspathAccessRule.java#L59-L68

and similar wording also used to store & read attributes:

like in `org.eclipse.jdt.internal.core.ClasspathEntry.encodeAccessRule(AccessRule, XMLWriter, boolean, boolean)`:

https://github.com/eclipse-jdt/eclipse.jdt.core/blob/fb1844c934409986d2ad332fce39f563cc210846/org.eclipse.jdt.core/model/org/eclipse/jdt/internal/core/ClasspathEntry.java#L691-L700

Also this class already "mirrors" `IAccessRule.ignoreIfBetter()` method. 

So I would propose to either change the code to provide `int getKind()` and do what `ClasspathAccessRule.getKind()` does or rename the `isForbidden()` method to `isNonAccessible()`.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/NameLookupTests2.java
```diff
@@ -425,5 +425,41 @@ public void testTransitionFromInvalidToValidJar() throws CoreException, IOExcept
...
+				new String[] {
+					"p/A.java",
+					"""
+					package p;
+					public class A {}
+					""",
+				},
+				null,
+				null,
+				new String[] {"**/*"},
```
iloveeclipse: would be nice to add comment on the right side of `new String[] {"**/*"}` that this is forbidden

Issue #4974, opened 2026-03-29
Use debug tracing instead of standard out in BufferManager


trancexpress:
Setting `org.eclipse.jdt.core/debug/buffermanager` results in `System.out` prints, which provides less information than debug traces - thread name, method name, timestamp and so on are missing. We should use `JavaModelManager.trace()` instead.

PR #4975, opened 2026-03-29
Use debug tracing instead of standard out in BufferManager

trancexpress:
This change adjusts `BufferManager` to use `JavaModelManager.trace()` when outputting debug infos due to `BufferManager.VERBOSE=true`.

Fixes: #4974


trancexpress:
Current build fails with:

```
Caused by: java.io.IOException: Server returned HTTP code: 502 for URL https://download.eclipse.org/eclipse/updates/4.40-I-builds/I20260326-1800/plugins/org.eclipse.jdt.compiler.apt.tests_1.3.1200.v20260318-0415.jar
```

PR #4990, opened 2026-04-06
Different behaviour between ECJ and javac for generic type inference

### LINKED ISSUE #3351: Different behaviour between ECJ and javac for generic type inference
Issue seems similar to #366 but this time it's the other way around: `ecj` compiles while `javac` fails.

```java
import java.util.List;
public class PassThroughGenerics {
    private class MyComp implements Comparable<MyComp> {
        @Override public int compareTo(MyComp other) { return 0; }
    }
    
    static <E extends Comparable<E>> List<E> sort(List<E> list) {
        return list;
    }
    
    static <T> List<T> genericList() {
        return null;
    }
    
        List<MyComp> sorted = sort(genericList());
        System.out.println(sorted);
    }
}
```
So `sort` should pass its generics through to `genericList` and `ecj` is capable of inferring `T` to `MyComp` while `javac` fails compilation with
> `PassThroughGenerics.java:16: error: method sort in class PassThroughGenerics cannot be applied to given types;`
> `        List<MyComp> sorted = sort(genericList());`
> `                              ^`
> `  required: List<E#1>`
> `  found:    List<Object>`
> `  reason: inferred type does not conform to equality constraint(s)`
> `    inferred: T`
> `    equality constraints(s): E#2`
> `  where E#1,T,E#2 are type-variables:`
> `    E#1 extends Comparable<E#1> declared in method <E#1>sort(List<E#1>)`
> `    T extends Comparable<E#2>`
> `    E#2 extends Comparable<E#2>`

**Versions**
`javac`
openjdk version "24-ea" 2025-03-18
OpenJDK Runtime Environment (build 24-ea+17-1938)
OpenJDK 64-Bit Server VM (build 24-ea+17-1938, mixed mode, sharing)
Tested in OpenJDK 7 - JDK 24, IBM Semeru 11, 17, 21; fails in every version

`ecj`
Eclipse Compiler for Java(TM) v20240215-1558, 3.37.0
Eclipse Compiler for Java(TM) v20240405-1402, 3.38.0

snjeza:
Fixes #3351 
Related to https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4838#issuecomment-3917324020

I have created a simpler class to reproduce the issue.
```
import java.util.List;
import java.util.Collections;
public class Test {
    static <E extends Comparable<E>> List<E> sort(List<E> list) {
        return list;
    }
        sort(Collections.emptyList());
        System.out.println("hello");
    }
}
```

The IDEA editor also doesn’t report any error in these cases.
If the class is tried to be run, IDEA will report an error when the `javac` compiler is used.

The InferenceContext18.resolve() method will add the `V = java.lang.Object` bound if it cannot resolve some inference variable.
The PR tries to postpone or skip this decision unless there is a bound that:

- isn’t a variable bound 
- doesn’t include raw types
- doesn’t include a lambda expression

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/InferenceContext18.java
@@ -1232,7 +1232,25 @@ public class InferenceContext18 {
 											}
 										}
 									}
-									tmpBoundSet.addBound(new TypeBound(variable, glb, ReductionResult.SAME), this.environment);
+									boolean addBound = true;
+									if (toResolveSet.contains(variable) && !variable.typeParameter.isRawType()) {
+										boolean add = false;
+										if (this.currentInvocation instanceof MessageSend messageSend) {
+											add = shouldAddBound(messageSend);
+										}
+										if (add && glb.equals(this.object)
+												&& !variable.typeParameter.isWildcard()) {
+											for (TypeBound bound : tmpBoundSet.flatten()) {
+												if (bound.left.equals(variable) && !bound.right.isTypeVariable()) {
+													addBound = false;
+													break;
+												}
+											}
+										}
+									}
+									if (addBound) {
+										tmpBoundSet.addBound(new TypeBound(variable, glb, ReductionResult.SAME), this.environment);
+									}
 								}
 							}
 							toResolveSet.remove(variable);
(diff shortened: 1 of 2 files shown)


stephan-herrmann:
Some background analysis:

Adding a same-bound `Object` as fallback has been introduced for https://bugs.eclipse.org/bugs/show_bug.cgi?id=426652

In that bug the fallback was added to prevent that inference would get stuck, i.e., simply not make any progress for an ivar without suitable upper nor lower bounds.

OTOH, not having any bound in some cases seems to be a bug introduced in 550a751ce1bc873d333fa63fec3308d2d2a04ec5 (no bugzilla reference). @srikanth-sankaran  fyi.

But the test case discussed here would not benefit from resolving the above wrinkles. 

---- 

When I look the simplicity of the spec implemented here:
> a candidate instantiation Ti is defined for each αi: 
> * [...]
> * Otherwise, where αi has proper upper bounds U1, ..., Uk, Ti = glb(U1, ..., Uk) ([§5.1.10](https://docs.oracle.com/javase/specs/jls/se26/html/jls-5.html#jls-5.1.10)).

... and compare this with the complexity of the proposed fix, I'm afraid we might be looking in the wrong place.

stephan-herrmann:
I just tried this example against work in progress in #5004  and this is what I get
```
1. ERROR in /tmp/Test.java (at line 8)
        sort(Collections.emptyList());
        ^^^^
The method sort(java.util.List<E extends java.lang.Comparable<E>>) in the type Test is not applicable for the arguments (java.util.List<E extends java.lang.Comparable<E>>)
```

Given that this matches to what javac answers, perhaps we don't need a specific fix for this test case :) 
I will retest when #5004 is nearing completion.

stephan-herrmann:
> I just tried this example against work in progress in #5004 and this is what I get
> 
> ```
> 1. ERROR in /tmp/Test.java (at line 8)
>         sort(Collections.emptyList());
>         ^^^^
> The method sort(java.util.List<E extends java.lang.Comparable<E>>) in the type Test is not applicable for the arguments (java.util.List<E extends java.lang.Comparable<E>>)
> ```
> 
> Given that this matches to what javac answers, perhaps we don't need a specific fix for this test case :) I will retest when #5004 is nearing completion.

Same result with #5129 (with plans to improve the error message via #5078 ).

With this we are done.

Thanks anyway, @snjeza for your efforts.

PR #4996, opened 2026-04-13
Add textBlock break logic to Formatter

inuyasha82:
## What it does
Implement issue: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2731

## How to test
See pr: https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2936 for instructions

## Author checklist


jjohnstn (changes requested):
Some problems exist.  When I format the following code with the new option set in the UI (via your other patch):

```
package tests

public class TestTextBlockFormatter {

	public static void foo() {
		final String x = """
				abc
				def
				ghi""";
		System.out.println("<" + x + ">");
	}

	public static void foo2() {
		final String str3 = "asd" + """
				subd3
				ospl
				test""" + "osd\n";
		System.out.println("<" + str3 + ">");
	}

		TestTextBlockFormatter.foo();
		TestTextBlockFormatter.foo2();
	}
}
```

I sometimes get:

```
package tests;

public class TestTextBlockFormatter {

	public static void foo() {
		final String x =
		"""
				abc
				def
				ghi\
				"""
		;
		System.out.println("<" + x + ">");
	}

	public static void foo2() {
		final String str3 = "asd" +
				"""
						subd3
						ospl
						test\
				"""
				+ "osd\n";
		System.out.println("<" + str3 + ">");
	}

		TestTextBlockFormatter.foo();
		TestTextBlockFormatter.foo2();
	}
}
```

This is wrong for foo2() and matches your test.  The closing triple quote where it is means that all the text is indented by 2 tabs when you print it.  The foo() case is correct and puts the tripe quote where it should go (matching the start of each text line).

The weird part is that sometimes I get the following without changing any preferences after undoing and re-running format:

```
package tests;

public class TestTextBlockFormatter {

	public static void foo() {
		final String x = """
				abc
				def
				ghi\
				""";
		System.out.println("<" + x + ">");
	}

	public static void foo2() {
		final String str3 = "asd" + """
				subd3
				ospl
						test\
				"""
				+ "osd\n";
		System.out.println("<" + str3 + ">");
	}

		TestTextBlockFormatter.foo();
		TestTextBlockFormatter.foo2();
	}
}
```

As you can see, the foo2() is wrong on the 3rd text block line that got modified.  The triple end-quote is correct but it didn't add the newline for the opening triple quotes.

inuyasha82:
> Some problems exist. When I format the following code with the new option set in the UI (via your other patch):
> 
> ```
> package tests
> 
> public class TestTextBlockFormatter {
> 
> 	public static void foo() {
> 		final String x = """
> 				abc
> 				def
> 				ghi""";
> 		System.out.println("<" + x + ">");
> 	}
> 
> 	public static void foo2() {
> 		final String str3 = "asd" + """
> 				subd3
> 				ospl
> 				test""" + "osd\n";
> 		System.out.println("<" + str3 + ">");
> 	}
> 
> 	public static void main(String[] args) {
> 		TestTextBlockFormatter.foo();
> 		TestTextBlockFormatter.foo2();
> 	}
> }
> ```
> 
> I sometimes get:
> 
> ```
> package tests;
> 
> public class TestTextBlockFormatter {
> 
> 	public static void foo() {
> 		final String x =
> 		"""
> 				abc
> 				def
> 				ghi\
> 				"""
> 		;
> 		System.out.println("<" + x + ">");
> 	}
> 
> 	public static void foo2() {
> 		final String str3 = "asd" +
> 				"""
> 						subd3
> 						ospl
> 						test\
> 				"""
> 				+ "osd\n";
> 		System.out.println("<" + str3 + ">");
> 	}
> 
> 	public static void main(String[] args) {
> 		TestTextBlockFormatter.foo();
> 		TestTextBlockFormatter.foo2();
> 	}
> }
> ```
> 
> This is wrong for foo2() and matches your test. The closing triple quote where it is means that all the text is indented by 2 tabs when you print it. The foo() case is correct and puts the tripe quote where it should go (matching the start of each text line).

For this one I need to do some investigation, but I think from what I see that it happens when there is no String Concatenation, so is not going to add any indent, to the new block. I need to check how to handle this case.

> 
> The weird part is that sometimes I get the following without changing any preferences after undoing and re-running format:
> 
> ```
> package tests;
> 
> public class TestTextBlockFormatter {
> 
> 	public static void foo() {
> 		final String x = """
> 				abc
> 				def
> 				ghi\
> 				""";
> 		System.out.println("<" + x + ">");
> 	}
> 
> 	public static void foo2() {
> 		final String str3 = "asd" + """
> 				subd3
> 				ospl
> 						test\
> 				"""
> 				+ "osd\n";
> 		System.out.println("<" + str3 + ">");
> 	}
> 
> 	public static void main(String[] args) {
> 		TestTextBlockFormatter.foo();
> 		TestTextBlockFormatter.foo2();
> 	}
> }
> ```
> 
> As you can see, the foo2() is wrong on the 3rd text block line that got modified. The triple end-quote is correct but it didn't add the newline for the opening triple quotes.

This one I think i know when it happens, at least this is what I get in this case, when we test it in the child eclipse, do a ctrl+shift+f it format correctly, then when undoing it with ctrl+z it undo the format, but it left a character selecected between the end of the text block and the "+" symbol.  If this empty space is selected, this is causing the strange format.  I tested it on an  eclipse without my patch, and I still see this weir  behaviour (I tried on TextEditsBuilder.java file) formatted the code, and then ctrl+z to undo, and it leaves a character selected (i tihink it depends on where the cursor is), and if we press format again, it add something like a one and half line. 

Maybe is this a bug of the undo?

inuyasha82:
Below an explanation of the workflow of the formatter code (it is also useful for whoever need to update any other formatting code): 

The formatting is applied after the various token have been created, and for the `Textblock` the format is creating several internalstructure tokens.
The closing quote is part of the last token, when is not on a newline of it's own.

The format code is calle before the `TextEditsBuilder` traversal of the tokens.

The traversal of the token, as part of it's code is calling this token method, that as first thing does the `bufferWhiteSpaceBefore` (for each token), updating an intenrnal field this.buffer.
At the end of the `bufferWhiteSpaceBefore` we have the exact indentation for the token. Once the buffer is applied to the current token, it is cleared, and then it is recomputed for the next token
After the  `bufferWhiteSpaceBefore` is completed, it calls the `flushBuffer` method that apply the buffer to the current portion of code analyzed.
The problem is the closing quotes, that are part of the last token when not on a newline. What happens is that it applies indentation buffer  to the current token correctly, but let's say  for example  that the contents of the last token is  `ads"""`  The `flushBuffer` method after  applying the buffer to the token creates an edit for adding the content of the buffer that will be in the format of: `\n\t\t\t\t\t\t`   at the beginning of the token.
And at the end of this code the buffer is cleared and indentation recomputed and applied on the next token. So the closing quotes aren't updated because they are not on a new token.

At this point i put my changes while flushing the buffer it always checks:

* if the current token is not `null` (apparently the `flushbuffer` can be called even with current token null (as some tests were failing because of it being `null`),
*the current token string ends with the sequence: `"""` 
* I have to check if the token is not a `TokenTextBlock`, because the traversal is done for both full tokens and internal tokens. And I don't want to change the full `TextBlock` token
* Is not a `TokenNameTextBlock` type


Now if the checks above are passing (and of course the option for newline on text block is enabled), i get the starting position of the substring with the closing quotes. I check if it is greater than 0, because if it is 0, it means that these aer the opening quotes.

So if everything above pass, this means I need to create an extra edit at the position of the split, adding the buffer, before it is being cleared.

Now why i do it in two places? Because there are two possibilities:

* First: The position of the current token already has the correct indentation, so the sourceMatch variable is true, and it skip any buffer updates, so I want to catch this case, since we still need to update the closing quotes.
* Second: We need to apply the indentation, so sourceMatch is false (then it enters into the while loop), it applies all the formats required, and then we need to again care about the closing quotes.

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/LineBreaksPreparator.java
```diff
@@ -614,20 +620,30 @@ public boolean visit(TextBlock node) {
 			}
 		}
 		WrapPolicy wrapPolicy = new WrapPolicy(WrapMode.DISABLED, 0, -1, 0, 0, 1, false, false);
-		for (i = 1; i < lines.size(); i++) {
-			Token t = lines.get(i);
-			Token line = new Token(t, t.originalStart + incidentalWhitespace, t.originalEnd, TokenNameTextBlock);
-			line.setWrapPolicy(wrapPolicy);
-			lines.set(i, line);
+		if(!this.options.put_text_block_quotes_on_new_line || ((TokenTextBlock)block).hasReplace()) {
+			for (i = 1; i < lines.size(); i++) {
```
jjohnstn: There appears to be a scenario where you won't create the internal lines (the preference is on but there is no replacement needed).  Why wouldn't you do the old behavior there?

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/LineBreaksPreparator.java
```diff
@@ -659,8 +675,7 @@ private void handleBracedCode(ASTNode node, ASTNode nodeBeforeOpenBrace, String
 
 	private void handleBracedCode(ASTNode node, ASTNode nodeBeforeOpenBrace, String bracePosition, boolean indentBody,
 			int blankLinesAfterOpeningBrace, int blankLinesBeforeClosingBrace) {
-		int openBraceIndex = nodeBeforeOpenBrace == null
-				? this.tm.firstIndexIn(node, TokenNameLBRACE)
+		int openBraceIndex = nodeBeforeOpenBrace == null ? this.tm.firstIndexIn(node, TokenNameLBRACE)
```
jjohnstn: You need to avoid formatting old code other than lines you have changed.  When one runs a Team Revision changes on the source, it will see your format change and this stomps over the original PR plus logic that made the line change.  I might suggest you do a compare of the sources and move the old code back in the Comparison Editor.

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/LineBreaksPreparator.java
```diff
@@ -614,20 +620,30 @@ public boolean visit(TextBlock node) {
 			}
 		}
 		WrapPolicy wrapPolicy = new WrapPolicy(WrapMode.DISABLED, 0, -1, 0, 0, 1, false, false);
-		for (i = 1; i < lines.size(); i++) {
-			Token t = lines.get(i);
-			Token line = new Token(t, t.originalStart + incidentalWhitespace, t.originalEnd, TokenNameTextBlock);
-			line.setWrapPolicy(wrapPolicy);
-			lines.set(i, line);
+		if(!this.options.put_text_block_quotes_on_new_line || ((TokenTextBlock)block).hasReplace()) {
```
jjohnstn: code format is inconsistent.  I would suggest you set a save action to format the lines changed and this will solve your problems.  In this particular case you have no space between if and (.  In other cases you have blanks after ( but not before ).  The formatting defaults for the project are set and the formatter will obey them.

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
```diff
@@ -92,17 +92,15 @@ private List<IRegion> adaptRegions(List<IRegion> givenRegions) {
 			if (start > sourceStart) {
 				Token token = this.tm.get(this.tm.findIndex(start, ANY, false));
 				if ((token.tokenType == TokenNameCOMMENT_BLOCK || token.tokenType == TokenNameCOMMENT_JAVADOC
-						|| token.tokenType == TokenNameCOMMENT_MARKDOWN)
-						&& start <= token.originalEnd) {
+						|| token.tokenType == TokenNameCOMMENT_MARKDOWN) && start <= token.originalEnd) {
```
jjohnstn: more unnecessary formatting changes.

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
```diff
@@ -389,10 +389,30 @@ private void flushBuffer(int currentPosition) {
...
 			this.counter = regionEnd;
 		}
+		if (sourceMatch && this.currentRegion < this.regions.size()) {
+			checkClosingQuotes(currentPosition, this.regions.get(this.currentRegion));
+		}
 		this.buffer.setLength(0);
 		this.counter = currentPosition;
 	}
 
+	private void checkClosingQuotes(int position, IRegion region) {
```
jjohnstn: Rename to checkTextBlockClosingQuotes() for clarity

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
```diff
@@ -389,10 +389,30 @@ private void flushBuffer(int currentPosition) {
...
+		Token token = this.getCurrent();
+		if (token == null) return;
+		String closingQuotes = "\"\"\""; //$NON-NLS-1$
+		if (this.options.put_text_block_quotes_on_new_line && !(token instanceof TokenTextBlock)
```
jjohnstn, line 413: This could use a comment to define the actual scenario occurring.

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
```diff
@@ -389,10 +389,30 @@ private void flushBuffer(int currentPosition) {
...
+				&& token.tokenType == TokenNameTextBlock
+				&& position < token.originalEnd
+				&& this.source.substring(position, token.originalEnd + 1).endsWith(closingQuotes)) {
+			String stringToCheck = this.source.substring(position, token.originalEnd + 1);
+			int splitPlace = stringToCheck.indexOf(closingQuotes);
+			if (splitPlace > 0) {
+				this.edits.add(getReplaceEdit(position + splitPlace, position + splitPlace,
+						'\\' + this.buffer.toString(), region));
```
jjohnstn, line 422: where does the newline after the \ get added?  Again, a comment may clarify this.

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TokenTextBlock.java
```diff
@@ -0,0 +1,33 @@
...
+
+	public TokenTextBlock(int sourceStart, int sourceEnd, TerminalToken tokenType) {
+		super(sourceStart, sourceEnd, tokenType);
+	}
+
+	public boolean hasReplace() {
+		return this.hasReplace;
+	}
+
+	public void addNewlineReplace() {
```
jjohnstn: Rename to setNewlineReplace() for clarity as it makes it clear you are setting a boolean instead of actually making the replace.

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/Token.java
```diff
@@ -150,6 +150,9 @@ public static Token fromCurrent(Scanner scanner, TerminalToken currentToken) {
 			token.breakAfter();
 			return token;
 		} else {
+			if (add_newline_textblock && currentToken == TerminalToken.TokenNameTextBlock) {
```
jjohnstn: I think you should move this logic to the call to fromCurrent().  If the option is on and the returned token has a tokenType field that is TerminalToken.TokenNameTextBlock, then overwrite the token with your new token.  This prevents this static method from changing it's API in case someone has overridden the formatter.

jjohnstn (changes requested):
Works much better.  I can confirm that highlighting the space and doing a format causes the misformat.  I think you need to figure this out as you have noted the undo highlights the space.  When something is selected and you do a format, it should only format the selected region so you should probably recognize this somewhere and do nothing since your text block is not involved (perhaps the preparator needs to do this..check the other preparators or look for a check on the selected region).

I also noticed that the first quotes for my simple case without any concatenation doesn't indent the opening quotes but does so if concatenation occurs before the text block.  It would be good to understand why this is so.

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
inuyasha82: It is already present in this.buffer, i'll add a comment.

inuyasha82:
> Works much better. I can confirm that highlighting the space and doing a format causes the misformat. I think you need to figure this out as you have noted the undo highlights the space. When something is selected and you do a format, it should only format the selected region so you should probably recognize this somewhere and do nothing since your text block is not involved (perhaps the preparator needs to do this..check the other preparators or look for a check on the selected region).

I think I fixed it. Now it applies the format only around the selection (so if I select something on the line of the textblock it will format that line. 

Apparently while formatting there is no real way to know what is the selected text, the only thing I know is that the region is the selected line with the selected character, but I can't know what is the selected part apparently.  

I would say any further improvement to the selected case could be treated in a separate issue, since It requires more investigation. 
> 
> I also noticed that the first quotes for my simple case without any concatenation doesn't indent the opening quotes but does so if concatenation occurs before the text block. It would be good to understand why this is so.

You said that the foo1 example was ok (and the opening quotes were not indented in that example), so I didn't checked it. Anyway the format is decided again by the bufferAlign and bufferIndent. So when we create the 
newline this is what the TextEditBlock computes.

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/LineBreaksPreparator.java
inuyasha82: Ok so the if is no longer needed, since the change has been moved inside the TextBlockEditor, latest push ir reflecting this change.

jjohnstn:
I made some changes.  I made the hasReplace() method meaningful and this helped to simplify logic.  The TokenTextBlock only gets created if the text block is entirely within a selection region.  A change to the indent call fixes the problem with selecting a single space on the end line of a text block.  Now, it moves the following code to the next line but doesn't touch the text block.  I also removed format-only changes so once the build passes it can be merged.

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/Token.java
jjohnstn: Done

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TokenTextBlock.java
jjohnstn: Done

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/LineBreaksPreparator.java
jjohnstn: Formatting changes undone.

jjohnstn (approved):
Looks good.

PR #5002, opened 2026-04-16
Stop using Color constructors with Display

akurtakov:
Simplifies the code and reduces chances for problems. Contributes to eclipse-platform/eclipse.platform.swt#3233 .

## Author checklist

- [x] I have thoroughly tested my changes
- [x] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [x] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

PR #5007, opened 2026-04-17
Fix test broken by previous PR

fedejeanne:
See https://github.com/eclipse-jdt/eclipse.jdt.core/pull/5002#issuecomment-4267122352

cc @akurtakov


akurtakov:
Merge this one as it seems as the right thing to do and the test failure looks unrelated.

PR #5014, opened 2026-04-20
[Formatter] Remove empty spaceBefore handling for lambdas

### LINKED ISSUE #4998: Lambda expressions are incorrectly formatted based on method opening parenthesis
Component: Eclipse Formatter

UI Setting: Whitespace / Declarations / Methods / Before opening parenthesis

XML Setting:

```xml
<setting id="org.eclipse.jdt.core.formatter.insert_space_before_opening_paren_in_method_declaration" value="insert"/>
```

Actual behavior:
```java
@Override
public Stream<TestTemplateInvocationContext> provideTestTemplateInvocationContexts (ExtensionContext context) {
    Selector selector =
        findSelector(context).orElseThrow( () -> new IllegalStateException("Missing @Selector annotation"));
    // ...
}
```

Expected  behavior:
```java
@Override
public Stream<TestTemplateInvocationContext> provideTestTemplateInvocationContexts (ExtensionContext context) {
    Selector selector =
        findSelector(context).orElseThrow(() -> new IllegalStateException("Missing @Selector annotation"));
    // ...
}
```

Speculation: AST possibly treats lambda as an anonymous method declaration

Version: 2025-09 (4.37.0)

Build id: 20250905-1456

SougandhS:
This change introduces a dedicated formatter option to control spacing before lambda empty parameter parentheses. Lambda formatting is now handled independently, ensuring method declaration settings no longer affect lambda expressions.

Lambda

Screenshot 2026-04-20 at 4 02 16 PM

Fixes : https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4998

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/SpacePreparator.java
+++ b/org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/SpacePreparator.java
@@ -596,12 +596,8 @@ public class SpacePreparator extends ASTVisitor {
 				this.options.insert_space_after_lambda_arrow);
 		List<VariableDeclaration> parameters = node.parameters();
 		if (node.hasParentheses()) {
-			if (handleEmptyParens(node, this.options.insert_space_between_empty_parens_in_method_declaration)) {
-				handleToken(node, TokenNameLPAREN,
-						this.options.insert_space_before_opening_paren_in_method_declaration, false);
-			} else {
-				handleToken(node, TokenNameLPAREN,
-						this.options.insert_space_before_opening_paren_in_method_declaration,
+			if (!handleEmptyParens(node, this.options.insert_space_between_empty_parens_in_method_declaration)) {
+				handleToken(node, TokenNameLPAREN, false,
 						this.options.insert_space_after_opening_paren_in_method_declaration);
 
 				handleTokenBefore(node.getBody(), TokenNameRPAREN,


SougandhS:
Hi @mateusz-matela 
Could you please check this PR when you get time ?

mateusz-matela:
I don't think it's a good approach add a setting like that. A lambda with empty parenthesis is just a kind of expression, it's always in a context where there's already a setting for space (after equals in assignment, after opening parenthesis or after a comma in method call parameters). I see no reason to have a setting that will override that in this particular case (or to handle other cases in similar way, like lambdas with non-empty parenthesis, conditional expressions, etc).

I think the behavior is just a mistake in the SpacePreparator. It calls `handleEmptyParens(....insert_space_between_empty_parens_in_method_declaration)`  which makes sense (it could be an improvement to have a separate setting for a space between empty params in lambda, but it probably won't be needed in practice), so it also felt natural to call `handleToken(....insert_space_before_opening_paren_in_method_declaration,);` but there evidently was not enough thought and testing.

I think we should just skip the second call.

mateusz-matela:
Oh, and in the next line where it uses `insert_space_before_opening_paren_in_method_declaration` for non-empty parens it should also assume `false` by the same logic.
As for `insert_space_after_opening_paren_in_method_declaration` and `insert_space_before_closing_paren_in_method_declaration` I think they can stay, they won't conflict with other configs in the same way (only possible conflicts are the same that already occur in actual method declarations).

SougandhS:
@mateusz-matela so its better to keep this always false for lambda ?
```
if (handleEmptyParens(node, this.options.insert_space_between_empty_parens_in_method_declaration)) {
      handleToken(node, TokenNameLPAREN,false, false);
```

mateusz-matela:
Well, yes, and when called it with both parameters false, `handleToken` will do nothing, so it can be just skipped.

SougandhS:
> Well, yes, and when called it with both parameters false, `handleToken` will do nothing, so it can be just skipped.

Added the change 👍

mateusz-matela (approved):

mateusz-matela:
Oops, I missed that there's no unit test... Oh well, I guess it's fine just this one time ;)

Issue #5035, opened 2026-04-28
ASTConveter javadoc pending tests for the snippet tag


subyssurendran666:
[Snippet tag](https://openjdk.org/jeps/413) was introduced in java18 and some of the scenarios covered in the ASTConverterJavadocTest_18. The remaining scenarios are following.

1. Empty Body
-------------

  ```
{@snippet :
  }
```

2. Single Element
------------------

  ```
{@snippet :
 int a = 10; 
}
```

3. Multi-line plain code
----------------------

  ```
{@snippet :
    int a = 1;
    int b = 2;
  }
```

4. Blank line inside body
-----------------------

  ```
{@snippet :
  
  int a = 1;
  
  int b = 2;
  }
```

5. Multiple attribute
-------------------

  ```
{@snippet id="x1" lang=java region="main" :
    int a;
  }
```

6. Single quote attributes
-------------------------

  ```
{@snippet id='demo' lang='java' :
    int a;
  }
```

7. Nested blocks
----------------

  ```
{@snippet :
    if (a) {
       while (b) {
          foo();
       }
    }
  }
```

8. {@code} inside snippet (here code tag will treat as plain text)
---------------------------------------------------------------

  ```
{@snippet :
   {@code int a = 0;}
  }
```

9. Replace tag
--------------

  ```
{@snippet :
  System.out.println("abc"); // @replace substring="abc" replacement="xyz"
  }
```

10. Link tag
-----------
```
{@snippet : 
  System.out.println("abc"); //@link substring="System" target="System"
 }
```

RagiGeorge:
@subyssurendran666 
Can I please work on it

PR #5041, opened 2026-04-30
Add missing ASTConverter Javadoc tests for @snippet tag (Java 18)

RagiGeorge:
# why
https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5035
ASTConveter javadoc pending tests for the snippet tag 

# What 
This change adds a set of new ASTConverter tests covering previously missing scenarios for the @snippet Javadoc tag introduced in Java 18 (JEP 413).
The tests focus on validating AST‑level parsing behavior (DOM construction) and intentionally avoid doclet/rendering semantics.

The following scenarios are now covered in ASTConverterJavadocTest_18:
• Empty snippet body
• Single‑element snippet
• Multi‑line plain snippet
• Blank lines inside snippet body
• Multiple attributes (id, lang, region)
• Single‑quoted attributes
• Nested code blocks
• @replace markup inside snippet
• @link markup inside snippet

###  Note on {@code} Inside @snippet
A dedicated test for the case:
{@snippet :
 {@code int a = 0;}
}
was intentionally not added.
While JEP 413 specifies that {@code} inside a snippet is treated as plain text, the current ASTConverterJavadocTest_18 infrastructure counts raw @tag tokens during source scanning and expects a matching number of AST TagElements.

tested:
Screenshot 2026-04-30 at 2 53 16 PM

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverterJavadocTest_18.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverterJavadocTest_18.java
@@ -1057,4 +1057,315 @@ public class ASTConverterJavadocTest_18 extends ConverterTestSetup {
 		assertEquals("Tag element fragment should be TextElement", true, tagElem.fragments().get(0) instanceof TextElement);
 		assertEquals("Fourth fragment should be TextElement", true, fragments.get(3) instanceof TextElement);
 	}
+
+	public void testSnippetMultilineOnlyJavadoc5() throws JavaModelException {
+		this.workingCopies = new ICompilationUnit[1];
+		this.workingCopies[0] = getWorkingCopy("/Converter_15_1/src/javadoc/X.java",
+				"""
+					package javadoc;
+					public class X {
+						/**
+						 * {@snippet :
+						 * 	int a = 1;
+						 * 	int b = 2;
+						 * }
+						 */
+						 public static void foo(Object object) {}
+					}
+				"""
+			);
+		CompilationUnit compilUnit = verifyComments(this.workingCopies[0]);
+		List unitComments = compilUnit.getCommentList();
+		assertEquals("Wrong number of comments", 1, unitComments.size());
+		Javadoc javadoc = (Javadoc) unitComments.get(0);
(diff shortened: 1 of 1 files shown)


org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverterJavadocTest_18.java
```diff
@@ -1057,4 +1057,516 @@ public void testSnippetMultiLineTagsJavadoc4() throws JavaModelException {
...
+
+		Object validity = snippetTag.getProperty(TagProperty.TAG_PROPERTY_SNIPPET_IS_VALID);
+		assertTrue(
+			"Snippet with nested blocks should be valid",
+			validity instanceof Boolean && ((Boolean) validity).booleanValue()
+		);
+
+		List fragments = snippetTag.fragments();
+		assertFalse("Snippet should have fragments", fragments.isEmpty());
+		long textElementCount = fragments.stream().filter(TextElement.class::isInstance).count();
```
subyssurendran666: you can avoid textElementCount variable and replace it with fragments.size() since the fragments variable is a List.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/ASTConverterJavadocTest_18.java
```diff
@@ -1057,4 +1057,516 @@ public void testSnippetMultiLineTagsJavadoc4() throws JavaModelException {
...
+		assertTrue( "Snippet should contain multiple text elements",textElementCount >= 3);
+		StringBuilder content = new StringBuilder();
+		for (Object fragment : fragments) {
+			if (fragment instanceof TextElement textElement) {
+				content.append(textElement.getText());
+			}
+		}
+
+		String text = content.toString();
+		assertTrue("Snippet should contain if block", text.contains("if (a)"));
```
subyssurendran666: For loop is more expensive here. Instead, you can use `((TextElement)fragments.get(0)).getText();`.

subyssurendran666 (changes requested):

subyssurendran666 (approved):
Looks good.

PR #5054, opened 2026-05-06
"var" type inferrence bug with generic wrapping

### LINKED ISSUE #2857: Bug 549025 - [1.8][spec] equivalence of "?" vs. "? extends Object"
From https://bugs.eclipse.org/549025

> We had discussed the significance of "extends Object" in a wildcard before.
Only today I found https://bugs.openjdk.java.net/browse/JDK-6480391 which had been resolved for 8.
> 
> Note that it says "? extends Object" is an unbounded wildcard ...

and

> The following sentence can indeed by found in JLS §4.5.1 since version 8:
> * The wildcard ? extends Object is equivalent to the unbounded wildcard ?.

stephan-herrmann:
+ Ignore type bound of "? extends Object"
+ Preserve null info from erased bound '? extends @nonnull Object'
+ Test adjustments

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5028
Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2857


stephan-herrmann:
@srikanth-sankaran fyi I organized the changes as 3 commits for easier reviewing, in case you find the time for it.

commit 1 brings the main change in TypeSystem.getWildcard(), where we simply drop "extends Object". In this commit I adjusted tests only where we expected error messages with "? extends Object", which in many cases is no longer shown. Only for one error message I restore that information from the AST.

commit 2 resolves all remaining failures in GenericTypeTest, where changes were indeed significant, but AFAICS all changes are going towards alignment with javac.  This is the part that might deserve special scrutiny, but running this test class with `-Drun.javac=enabled` indeed shows better results than before! Note, that in this test mode you can use the console output to see all error messages from javac, too.

commit 3 finally restores nullness information from `? extends @NonNull Object`, which otherwise got dropped when dropping the bound altogether.

srikanth-sankaran:
> @srikanth-sankaran fyi I organized the changes as 3 commits for easier reviewing

I'll wait for the test failures to be cleaned up and then review - OK ? Thanks

stephan-herrmann:
Test failures revealed that curiously our jclMinXY.jar declare `Object.getClass()` as returning `Class<? extends Object>` which is neither useful nor does it reflect the situation in real JDK. Anyway, via find+replace test expectations no longer insist on the useless bound (commit 4).

stephan-herrmann:
> > @srikanth-sankaran fyi I organized the changes as 3 commits for easier reviewing
> 
> I'll wait for the test failures to be cleaned up and then review - OK ? Thanks

OK, test failures have been resolved. I only need to remove a now-obsolete `@SuppressWarnings("unchecked")` in our code. Will have a look why that one changed.

@jarthana since this is a fundamental change it would be great to have this change stress tested. If possible with imposing stress on you :) I'd like to merge this for M3.

stephan-herrmann:
> I only need to remove a now-obsolete `@SuppressWarnings("unchecked")` in our code. Will have a look why that one changed.

Previously, we suppressed this warning:
`Type safety: The expression of type List needs unchecked conversion to conform to List<? extends Object>`

Inside `TypeBinding.needsUncheckedConversion(TypeBinding)` we call `isBoundParameterizedType()` which says: "Returns true if parameterized type AND not of the form List\<?\>". By erasing the useless `extends Object` we fall into that exception exactly.

A second warning was 
`Type safety: Unchecked invocation unmodifiableList(List) of the generic method unmodifiableList(List<? extends T>) of type Collections`

but that one is tightly coupled to the former warning via code `INVOCATION_ARGUMENT_UNCHECKED`.

Interestingly, javac is reporting the second warning but not the first:
```
warning: [unchecked] unchecked method invocation: method unmodifiableList in class Collections is applied to given types
                                        return Collections.unmodifiableList(list);
                                                                           ^
  required: List<? extends T>
  found:    List
  where T is a type-variable:
    T extends Object declared in method <T>unmodifiableList(List<? extends T>)
```

I doubt the usefulness of that warning, though, since no part of the code should make any assumptions based on the inferred `List<?>`, no assumptions which could blow up at runtime.

jarthana:
> @jarthana since this is a fundamental change it would be great to have this change stress tested. If possible with imposing stress on you :) I'd like to merge this for M3.

Looks good.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/WildcardBinding.java
```diff
@@ -881,4 +884,10 @@ TypeBinding propagateNonConflictingNullAnnotations(TypeBinding type) {
 			return type;
 		return this.environment.createAnnotatedType(type, annots);
 	}
+
+	public boolean hasNullTagBits(long nullTagBits) {
+		if (nullTagBits == this.nullTagBitsFromErasedObjectBound)
+			return true;
+		return (this.tagBits & TagBits.AnnotationNullMASK) == nullTagBits;
```
srikanth-sankaran, line 891: The way I read this, `List<@NonNull ?>` is the same as `List<? extends @NonNull Object>` - is that right ? if so do we need the new attribute `nullTagBitsFromErasedObjectBound` at all ??

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/classfmt/ExternalAnnotationProvider.java
```diff
@@ -167,6 +169,23 @@ void initialize(InputStream input) throws IOException {
 		}
 	}
 
+	private String normalizeWildcards(String signature, boolean isAnnotated) {
+		signature = signature.replaceAll("\\+Ljava/lang/Object;", "*"); //$NON-NLS-1$ //$NON-NLS-2$
+		if (isAnnotated) {
+			// '1' on either the wildcard or its Object bound or both => treat like '@NonNull ?'
+			final String nonNullWildcard = "*1"; //$NON-NLS-1$
```
srikanth-sankaran, line 176: Related to the review comment in WildcardBinding.java, the way this reads is that `List<@NonNull ?>` is the same as `List<? extends @NonNull Object>`  if this is true, do we need the new attribute `org.eclipse.jdt.internal.compiler.lookup.WildcardBinding.nullTagBitsFromErasedObjectBound` ??

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/WildcardBinding.java
```diff
@@ -161,46 +162,46 @@ public long determineNullBitsFromDeclaration(Scope scope, Wildcard wildcard) {
...
-								TypeBinding newBound = this.bound.withoutToplevelNullAnnotation();
-								this.bound = newBound;
-								wildcard.bound.resolvedType = newBound;
-							} else {
-								scope.problemReporter().contradictoryNullAnnotationsOnBounds(annotation, nullTagBits);
-							}
+		long boundNullTagBits = this.bound != null && this.bound.isValidBinding()
+				? this.bound.tagBits & TagBits.AnnotationNullMASK
+				: this.nullTagBitsFromErasedObjectBound;
+		if (boundNullTagBits != 0L) {
```
srikanth-sankaran, line 168: A comment here that bound cannot be jlo as such would have been erased could be useful here.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/problem/ProblemReporter.java
```diff
@@ -10090,10 +10091,36 @@ public void redundantSpecificationOfTypeArguments(ASTNode location, TypeBinding[
...
+		} else {
+			problemArguments = typesAsString(argumentTypes, false);
+			messageArguments = typesAsString(argumentTypes, true);
+		}
 		this.handle(
 			IProblem.RedundantSpecificationOfTypeArguments,
-			new String[] {typesAsString(argumentTypes, false)},
-			new String[] {typesAsString(argumentTypes, true)},
+			new String[] {problemArguments},
+			new String[] {messageArguments},
```
srikanth-sankaran, line 10123: Is this a material/relevant change ? A search for  `Redundant specification of type arguments` does not bring up any matches in the patch. Or is it that in order to preserve the old diagnostics unchanged, you have implemented this ??

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/AbstractRegressionTest.java
```diff
@@ -945,9 +945,6 @@ Excuse excuseFor(JavacCompiler compiler) {
 				new EclipseHasABug(MismatchType.JavacErrorsEclipseWarnings),
 			EclipseBug421922 = // https://bugs.eclipse.org/bugs/show_bug.cgi?id=421922
 						new EclipseHasABug(MismatchType.EclipseErrorsJavacNone),
-			EclipseBug428061 = // https://bugs.eclipse.org/bugs/show_bug.cgi?id=428061
-								new EclipseHasABug(MismatchType.JavacErrorsEclipseNone |
-										MismatchType.JavacErrorsEclipseWarnings),
```
srikanth-sankaran, line 950: Is this deleted because it is unused ? Previous one looks unused too ?

org.eclipse.jdt.core.tests.model/workspace/Converter15/src/test0067/X.java
```diff
@@ -1,5 +1,5 @@
 package test0067;
 
 public class X {
-	private void foo(Class<? extends Object> class1) {}
+	private void foo(Class<? extends Number> class1) {}
```
srikanth-sankaran, line 4: What motivates this change ??

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/BatchASTCreationTests.java
```diff
@@ -941,12 +941,12 @@ public void test039() throws CoreException {
 				"public class X<T> {\n" +
 				"  void foo() {\n" +
 				"  }\n" +
-				"  void bar(X<? extends Object> x) {\n" +
+				"  void bar(X<? extends Exception> x) {\n" +
```
srikanth-sankaran, line 944: What motivates this pattern of changes ??

srikanth-sankaran (commented):
I have made one pass over the changes. Code changes received close attention while test changes were only glanced through. I will closely scrutinize the test changes tomorrow.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/problem/ProblemReporter.java
stephan-herrmann: >  is it that in order to preserve the old diagnostics unchanged, you have implemented this ?

That's it exactly. Without this change we would see the following change of error messages in situations like GenericsRegressionTest_1_7.test0052()
```diff
	X<Class<? extends Object>> x7 = new X<Class<? extends Object>>();
	                                    ^
- Redundant specification of type arguments <Class<? extends Object>>
+ Redundant specification of type arguments <Class<?>>
```

Where user specified `? extends Object` it didn't look right to say, you shouldn't say `?`.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/AbstractRegressionTest.java
stephan-herrmann: Yes, I deleted `EclipseBug428061` when I remove the last reference to it.

As for `EclipseBug421922` et al, it looks like b90ae111 was a bit over-eager, introducing several excuses without ever using them. _Or_ under-ambitious in not introducing the tests that would need the excuse. Take your pick.

org.eclipse.jdt.core.tests.model/workspace/Converter15/src/test0067/X.java
stephan-herrmann: White box tests using this class (ASTConverter15Test.test0067() et al.) want to see an upper bound on the wildcard binding. With `extends Object` this would have been erased.

org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/dom/BatchASTCreationTests.java
stephan-herrmann: Similar to the change in `.../Converter15/src/test0067/X.java`. This time the white-box-ness stems from comparing binding keys.

IIRC I had to pick a different bound this time, because `Number` is missing from jclMin.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/WildcardBinding.java
stephan-herrmann: In theory, there are two options for moving the annotation from the bound to the wildcard: (a) copy the annotations or (b) operate directly on tagBits. I tried the latter.

So if I skip `nullTagBitsFromErasedObjectBound` and directly merge those bits into `WildcardBinding.tagBits` the following tests would fail:

**NullTypeAnnotationTest.testBug485988Contradictory()**

Here we would no longer detect the conflict in `A<@Nullable ? extends @NonNull Object>`, which should raise: `This nullness annotation conflicts with a '@Nullable' annotation which is effective on the same type parameter` against the occurrence of `@NonNull`.

I could have implemented a new conflict detection based only on `tagBits`, but at that point I would no longer know where those bits originate from. This didn't sound compelling to me.

**NullTypeAnnotationTest.testNonNullBoundedWildcard()**

Here we would get the error even on the good list `list2`. This has to do with how we initialize a CaptureBinding from the wildcard:
```java
		if (wildcard.hasTypeAnnotations()) {
[...]
			super.setTypeAnnotations(wildcard.getTypeAnnotations(), wildcard.environment.globalOptions.isAnnotationBasedNullAnalysisEnabled);
			this.tagBits |= wildcard.tagBits & (TagBits.HasNullTypeAnnotation|TagBits.HasMissingType);
			this.tagBits |= wildcard.nullTagBitsFromErasedObjectBound;
		} else {
[...]
			this.tagBits |= wildcard.tagBits & (TagBits.AnnotationNullMASK|TagBits.HasNullTypeAnnotation|TagBits.HasMissingType);
		}
```
The first branch evaluates null annotations during `super.setTypeAnnotations()` (no annoations are present in this test), and avoids copying null related tagbits (in contrast to the else branch). Blindly copying all null-tagBits from the wildcard to the capture opened another can of worms elsewhere (the nature of which I don't recall).

In the end the simplest solution was to avoid any direct tweaking of `WildcardBinding.tagBits` and pull the rabbit out off the hat only during capture.


With more time we could also try to move / copy / merge annotation bindings instead of tweaking tagBits, but the current solution was the smallest fix I could think of in the given time.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/classfmt/ExternalAnnotationProvider.java
stephan-herrmann: Are you commenting here essentially to emphasize that moving the annotation should be all that is needed, rather than adding on more field to WildcardBinding?

Or should I comment more, what this normalization is about? (It's essentially so we can still interpret external annotations that where generated before this change, where `+Ljava/lang/Object;` may still appear in signatures instead of just `*`).

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/WildcardBinding.java
stephan-herrmann: It's true that jlO cannot occur here, but I don't see why that is relevant for understanding this code.

* If we have a valid bound: take the tag bits from it
* Otherwise, if we have stashed bits from an erased jlO bound, take those.

Should I perhaps expand the comment of the field declaration?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/WildcardBinding.java
stephan-herrmann: > I could have implemented a new conflict detection based only on tagBits, but at that point I would no longer know where those bits originate from. This didn't sound compelling to me.

As one force in this design space, AnnotatableTypeSystem is not in the position to raise any errors right here, because we have no location to report against (neither AST nor Scope). That's why we must carefully feed the required information in those locations that already do the error detection/reporting.

srikanth-sankaran:
> A second warning was `Type safety: Unchecked invocation unmodifiableList(List) of the generic method unmodifiableList(List<? extends T>) of type Collections`
> 
> but that one is tightly coupled to the former warning via code `INVOCATION_ARGUMENT_UNCHECKED`.
> 
> Interestingly, javac is reporting the second warning but not the first:

This worries me somewhat - I haven't fully investigated it. My worry is about any ripple effects around not detecting an unchecked conversion. As I recall there are various clauses in the JLS that stipulate different behavior based on whether an unchecked cast was involved. Don't have a ready reference though. I'll trust your intuition here.

If raw types were not in the picture, the warning would go away even with javac ? That would mitigate my worries a whole lot.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericsRegressionTest_9.java
```diff
@@ -2104,6 +2104,69 @@ static <U extends Ifc<U>> U someMapper(U u) {
...
+			       ^^^^
+		List is a raw type. References to generic type List<E> should be parameterized
+		----------
+		2. WARNING in X.java (at line 3)
+			public List getOriginalList(List list) {
+			                            ^^^^
+		List is a raw type. References to generic type List<E> should be parameterized
+		----------
+		""");
+}
```
srikanth-sankaran, line 2169: As commented in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/5054#issuecomment-4428869535 the difference in warning behavior worries me a bit - I will go with your analysis and intuition here.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericTypeTest.java
```diff
@@ -6623,11 +6607,6 @@ public void test0223() {
 			"	MX<Class> mx2 = new MX<Class>();\n" +
 			"	                       ^^^^^\n" +
 			"Class is a raw type. References to generic type Class<T> should be parameterized\n" +
-			"----------\n" +
-			"5. WARNING in X.java (at line 18)\n" +
-			"	mx1.foo(mx2.get());\n" +
-			"	        ^^^^^^^^^\n" +
-			"Type safety: The expression of type Class needs unchecked conversion to conform to Class<? extends Object>\n" +
```
srikanth-sankaran, line 6630: This is what you allude to in the top half of https://github.com/eclipse-jdt/eclipse.jdt.core/pull/5054#issuecomment-4399216464 I believe. Looks good.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericTypeTest.java
```diff
@@ -7238,12 +7217,7 @@ public void test0242() {
...
-			"Type safety: Unchecked cast from Class to Class<? extends Object>\n" +
-			"----------\n" +
-			"2. WARNING in X.java (at line 8)\n" +
-			"	final Class<? extends Object> clazz = (Class<? extends Object>) classes.get(\"test\");\n" +
-			"	                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n" +
-			"Unnecessary cast from Class to Class<? extends Object>\n" +
+			"Unnecessary cast from Class to Class<?>\n" +
 			"----------\n" +
 			"3. WARNING in X.java (at line 9)\n" +
 			"	final Class<? extends String> clazz2 = (Class<? extends String>) classes.get(\"test\");\n" +
```
srikanth-sankaran, line 7223: Diagnostics skip from 1 to 3 ... I have seen this before that such jumps don't cause test failures. If you have other changes, fix this one too along with it.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericTypeTest.java
```diff
@@ -22132,20 +22106,17 @@ public void test0702() {
...
 		"	      ^^^\n" +
-		"Type mismatch: cannot convert from X<X<? extends Object>> to X<X<? extends Cloneable>>\n" +
+		"Type mismatch: cannot convert from X<X<?>> to X<X<? extends Cloneable>>\n" +
+/* this is reported by javac despite equivalence of ? with ? extends Object, see case // 4
 		"----------\n" +
 		"3. ERROR in X.java (at line 13)\n" +
 		"	lhs = rhs; // 3\n" +
 		"	      ^^^\n" +
 		"Type mismatch: cannot convert from X<X<? extends Object>> to X<X<? extends Runnable>>\n" +
+*/
```
srikanth-sankaran, line 22119: Do you plan to raise this with the OpenJDK team ?

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericTypeTest.java
```diff
@@ -22132,20 +22106,17 @@ public void test0702() {
...
 		"----------\n" +
 		"4. ERROR in X.java (at line 19)\n" +
```
srikanth-sankaran, line 22121: Nit: Diagnostics number is not sequential

srikanth-sankaran (approved):
Overall looks good. 

1. A couple of minor comments about diagnostics not being sequential. Since these seem tolerated, you may ignore them unless you are also considering other changes.
2. A question about follow up with javac folk on behavior. 
3. Please double check once again to satisfy yourself that what is documented in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/5054#issuecomment-4399216464 is all well (I have no definite problem to report here)

srikanth-sankaran:
I ran a query to pull up all open github tickets that contain the string "extends Object" somewhere in the body.

[Here ](https://github.com/eclipse-jdt/eclipse.jdt.core/issues?q=is%3Aissue%20state%3Aopen%20%22extends%20Object%22) is the result. At leisure you may want to see if any of them change behavior due to the fix here, hopefully in the right direction.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericTypeTest.java
stephan-herrmann: I found it kind of cute,  that this is tolerated, allowing us to minimize the diff. :)
OTOH, I agree that this looks confusing now.

stephan-herrmann:
> > A second warning was `Type safety: Unchecked invocation unmodifiableList(List) of the generic method unmodifiableList(List<? extends T>) of type Collections`
> > but that one is tightly coupled to the former warning via code `INVOCATION_ARGUMENT_UNCHECKED`.
> > Interestingly, javac is reporting the second warning but not the first:
> 
> This worries me somewhat - I haven't fully investigated it. My worry is about any ripple effects around not detecting an unchecked conversion.

Please note that javac does not report unchecked **conversion** but unchecked **invocation**.  I would expect this warning to be specified somewhere in [JLS 15.12](https://docs.oracle.com/javase/specs/jls/se26/html/jls-15.html#jls-15.12) but so far I found nothing in this direction.

While I was deep into arguing that there is no reason for any warning I found something worth further investigation -> #5062.

Back to the issue at hand, however, we have
* a declaration `<T> List<T> unmodifiableList(List<? extends T>`
* inference instantiating `T` as `Object`
* a resulting signature of `List<Object> unmodifiableList(List<? extends Object>`
* which is equivalent to `List<Object> unmodifiableList(List<?>)`
* we provide an argument of type `List#RAW` to a parameter `List<?>`
   * due to "All of the type arguments of T are unbounded wildcards" this is exempted from unchecked warnings

I see no reason for any warning, pending the result of discussing #5062.

> As I recall there are various clauses in the JLS that stipulate different behavior based on whether an unchecked cast was involved. Don't have a ready reference though. I'll trust your intuition here.
> 
> If raw types were not in the picture, the warning would go away even with javac ? That would mitigate my worries a whole lot.

Right, replacing both instances of `List` with `List<?>` silences javac.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericTypeTest.java
stephan-herrmann: Yea, I probably should. Reduced test:
```java
public abstract class X<T extends Runnable> implements Runnable {
	void foo1(X<X<? extends Runnable>> lhs, X<X<? extends Object>> rhs) {
		lhs = rhs;
	}
	void foo2(X<X<? extends Runnable>> lhs, X<X<?>> rhs) {
		lhs = rhs;
	}
}
```
Should in any way report an even number or errors, but:
* ecj: 0
* javac: 1

stephan-herrmann:
> I ran a query to pull up all open github tickets that contain the string "extends Object" somewhere in the body.
> 
> [Here ](https://github.com/eclipse-jdt/eclipse.jdt.core/issues?q=is%3Aissue%20state%3Aopen%20%22extends%20Object%22) is the result. At leisure you may want to see if any of them change behavior due to the fix here, hopefully in the right direction.

Thanks, this is very helpful.

For some anecdotal reference I found in an email as early as Feb 2014:
> Does it make a difference whether we have S<?> or S<? extends Object>?
> I don't have a lot of answers here, maybe the JLS has some? 

:-/

stephan-herrmann:
> Overall looks good.

Thanks for your diligence!

>  1. A couple of minor comments about diagnostics not being sequential. Since these seem tolerated, you may ignore them unless you are also considering other changes.

Since no further changes are planned I'll leave it at this. I noted, though, that in the future I will prefer contiguous numbering over minimal test changes.

>  2. A question about follow up with javac folk on behavior.

Noted, will do. Perhaps after scanning your list of "extends Object" issues.
 
>  3. Please double check once again to satisfy yourself that what is documented in [#5054 (comment)](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/5054#issuecomment-4399216464) is all well (I have no definite problem to report here)

For the case at hand I argue that no harm is done, while a more general question has been out sourced to #5062 

Based on the approval, the only thing left to do is coordination regarding https://github.com/eclipse-platform/eclipse.platform.ui/issues/3992

stephan-herrmann:
> [Here ](https://github.com/eclipse-jdt/eclipse.jdt.core/issues?q=is%3Aissue%20state%3Aopen%20%22extends%20Object%22) is the result. At leisure you may want to see if any of them change behavior due to the fix here, hopefully in the right direction.

No changes found in any of those examples. Actually, there was hardly any mention of `? extends Object`, but I also found no way to further restrict the search.

iloveeclipse:
@stephan-herrmann : FYI: one jdt ui test needs expectation update:
https://download.eclipse.org/eclipse/downloads/drops4/I20260512-0925/testresults/html/org.eclipse.jdt.ui.tests.refactoring_ep440I-unit-linux-x86_64-java21_linux.gtk.x86_64_21.html

PR #5070, opened 2026-05-17
[Generics] Switching the order of extended interfaces alters the result of @NonNull annotation analysis

### LINKED ISSUE #4297: [Generics] Switching the order of extended interfaces alters the result of @NonNull annotation analysis
I'm not sure if this is a bug or if it works as expected but I'm reporting it anyway. Any infos are appreciated.

## Description

If I invert the order of the extended classes inside a generic, the result of the annotation analysis is affected.

## To reproduce
- Activate the **Null analysis** in the [preferences](https://github.com/user-attachments/assets/b1a3b8dd-e541-4cca-aa2c-b61cafe7be49)
- Add this snippet (observe the usage of `@NonNull`)

```java

public class GenericsOrder {
	public <T extends a & b> void test(T my) { // Change this line later
		my.method(null);
	}
}

interface a { int method(@NonNull String a); }

interface b { int method(String a); }
```

If you change the marked line from `<T extends a & b>` to `<T extends b & a>` then the compilation error goes away.

### Version
**I20250805-1800**

### Other details
I only tried with `@NonNull` but it might be that this also happens for other annotations?

stephan-herrmann:
+ report unchecked warning for unannotated 'this' argument
+ recover the actual contract that introduces an unmet constraint

Partly fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4297

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/.settings/.api_filters
+++ b/org.eclipse.jdt.core.compiler.batch/.settings/.api_filters
@@ -1,13 +1,5 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <component id="org.eclipse.jdt.core.compiler.batch" version="2">
-    <resource path="META-INF/MANIFEST.MF">
-        <filter id="926941240">
-            <message_arguments>
-                <message_argument value="3.46.0"/>
-                <message_argument value="3.45.0"/>
-            </message_arguments>
-        </filter>
-    </resource>
     <resource path="src/org/eclipse/jdt/core/compiler/CategorizedProblem.java" type="org.eclipse.jdt.core.compiler.CategorizedProblem">
         <filter comment="BETA_JAVA22 temporary issue" id="576725006">
             <message_arguments>
--- a/org.eclipse.jdt.core.compiler.batch/META-INF/MANIFEST.MF
+++ b/org.eclipse.jdt.core.compiler.batch/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Main-Class: org.eclipse.jdt.internal.compiler.batch.Main
 Bundle-ManifestVersion: 2
 Bundle-Name: Eclipse Compiler for Java(TM)
 Bundle-SymbolicName: org.eclipse.jdt.core.compiler.batch
-Bundle-Version: 3.46.0.qualifier
+Bundle-Version: 3.46.100.qualifier
 Bundle-ClassPath: .
 Bundle-Vendor: Eclipse.org
 Automatic-Module-Name: org.eclipse.jdt.core.compiler.batch
--- a/org.eclipse.jdt.core.compiler.batch/pom.xml
+++ b/org.eclipse.jdt.core.compiler.batch/pom.xml
@@ -17,7 +17,7 @@
     <version>4.41.0-SNAPSHOT</version>
   </parent>
   <artifactId>org.eclipse.jdt.core.compiler.batch</artifactId>
-  <version>3.46.0-SNAPSHOT</version>
+  <version>3.46.100-SNAPSHOT</version>
   <packaging>eclipse-plugin</packaging>
 
   <properties>
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ReferenceExpression.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ReferenceExpression.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2025 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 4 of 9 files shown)


stephan-herrmann:
This PR fixes some but not all problems observed in https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4297

Let's classify use cases:

The common situation consists of two independent interfaces that declare methods which are equivalent in terms of JLS, but have different annotations (which JLS doesn't care about). The following considerations hold if those annotations are null annotations evaluated by ecj; 3rd party annotations are out off scope here:

* Already in released versions we force any *class* explicitly implementing both interfaces to provide  an implementation that is conform to all inherited contracts. For any parameters with different null annotations, the class will be forced to declare a `@Nullable` parameter (or unannotated for the price of an unchecked warning). With this the implementation will accept all values that are legal by any of the contracts. :heavy_check_mark: 
* ecj doesn't check this for an interface extending both interfaces (see https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4297#issuecomment-3169398088), but still any class implementing that combined interface will be checked by the above rule. :heavy_check_mark: 
* Similarly, an intersection type `a & b` isn't immediately checked for conflicts (see the original https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4297#issue-3303630786), ~~but the above two items should "normally" ensure that no implementation of such an intersection type exists (modulo an implementation with unchecked warnings)~~ _this part was wrong, back to https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4297#issuecomment-4472041501._ but any implementation of the intersection type is checked  ( :heavy_check_mark: )
* There is a bug highlighted by @fedejeanne's witness from https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4297#issuecomment-3173921724 so we need to specifically check method references / lambdas as implementations for a type with null conflicts. This is what the current PR addresses. :x: 

We need to further distinguish the position of the ambiguous annotation:

This variant produces a warning already in 4.39:
```java
import org.eclipse.jdt.annotation.*;

public class GenericsOrder {
    public <T extends b & a> void test(@NonNull T my) {
        my.method(null);
    }

    public <T extends I> void test(@NonNull T my) {
        my.method(null); // The call goes through here
    }

        new GenericsOrder().test(GenericsOrder::mImpl); // problem against regular param 's'
    }

    public static int mImpl(@NonNull String s) { return 0; }
}

interface a { int method(@NonNull String a); }

interface b { int method(String a); }

interface I extends b, a { }
```
ecj complains:
```
1. WARNING in /tmp/GenericsOrder.java (at line 13)
        new GenericsOrder().test(GenericsOrder::mImpl); // problem against regular param 's'
                                 ^^^^^^^^^^^^^^^^^^^^
Null type safety: parameter 1 provided via method descriptor b.method(String) needs unchecked conversion to conform to '@NonNull String'
```
This part is independent of interface order. :heavy_check_mark: 

There still is order dependence inside both `test()` methods, complaining about the `null` argument in one case, but not with swapped order. This, however, can be seen as a secondary problem, as the declared parameter type is inconsistent, and there is no clean way to invoke either of the `test()` methods.

The original witness from https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4297#issuecomment-3173921724, however, adds one more trick to the mix: here we are passing a **reference to an instance method**, where the first parameter of the function type will be implicitly mapped to a `this` receiver of the target method. For this situation released  versions of ecj only checked if the function type declared a `@Nullable` parameter, which would trigger NPE against the method receiver. We failed to complain if the receiver is mapped to an unannotated parameter.

This is the main fix in this PR: warn when an unannotated parameter is mapped to the receiver of a reference to an instance method.

That witness will now raise 
```
1. WARNING in /tmp/GenericsOrder.java (at line 13)
        new GenericsOrder().test(String::length);
                                 ^^^^^^^^^^^^^^
Null type safety: parameter 'this' provided via method descriptor b.method(String) needs unchecked conversion to conform to '@NonNull String'
```

Swapping interface order will still change warnings against the bodies of `test()` methods, but it is no longer possible to make a "clean" **call** to `test()`, so, as argued above, the implementation of `test()` is of reduced interest.

As a side issue, the PR also fixes the exact message in this warning (and its sibling error in case of `@Nullable` instead of unannotated): Previously, we could mention the wrong contract `a.method(String)`, instead of `b.method(String)` due to the underlying arbitrary selection. This is fixed by searching (again) for the contract that has the exact parameter type we're complaining about.

With this I believe the issue is resolved for **method reference** as argument to an inconsistent functional interface. :+1: 

When using the **lambda** form of the same example, ecj has no extra knowledge about a `this` parameter that definitely must be nonnull.  Hence we keep silent about dereferencing the lambda argument, just as we always do for unannotated variables with no flow information. If `@Nullable` is added for `b.method()` then the expected error is raised. Anyway, this case is not order dependent :smile:

stephan-herrmann:
I was about to propose this fix for RC1, but at a closer look we are fixing an extreme corner case, depending on all of these conditions:
* an interface (explicit or implicit as an intersection type) inherits several equivalent abstract methods, which differ only in annotations
* those annotations are null annotations interpreted by ecj
* the method is invoked with a reference to an instance method
* the implicit `this` parameter will be mapped to a parameter of the function type, for which a `@NonNull` and an unannotated version exist, but no inherited contract declares it as `@Nullable`.
* interfaces are listed in that order, which will prefer the unannotated parameter over the `@NonNull` one

In this very specific case a relevant problem is not reported, thus hiding the fact that no conflict-free implementation of that functional interface is possible.

The bigger problem is that JLS is silent about all this, and thus for 3rd party annotations, for which ecj has no insider knowledge, order dependence cannot be avoided without a fix in JLS.

In that light I guess this fix can wait till after 4.40 GA.

PR #5064, opened 2026-05-13
Fix Signature to check context before converting $ to dot

jjohnstn:
- add a check in Signature.appendClassTypeSignature to see if a $ is preceded or followed by a dot before assuming it is an inner class and not part of a package name
- add new test to SignatureTests
- for https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5056

## What it does

See referenced issue.

## How to test

See referenced issue.

## Author checklist

- [x] I have thoroughly tested my changes
- [x] The change is following the [coding conventions](https://wiki.eclipse.org/Platform/How_to_Contribute#Coding_Conventions)
- [x] I have signed the [Eclipse Contributor Agreement (ECA)](https://www.eclipse.org/legal/ECA.php)

PR #5073, opened 2026-05-19
Tweak new signature logic that checks context of dollar

jjohnstn:
- modify Signature.appendClassTypeSignature() logic for the
  last segment to use the old logic as it appears there is other
  code that is not prepared for the dollar being part of the class
  name and expects the dollar being converted to dot

## What it does

See commit message.

## How to test

See #5064 comments.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/SignatureTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/SignatureTests.java
@@ -1600,9 +1600,17 @@ public void testCreateIntersectionTypeSignature2() {
 }
 
 public void testDollarPackage() {
-	String signature = Signature.getTypeErasure("Ljava.util.$foo$.Optional");
+	String signature = new String(Signature.toCharArray(new String("Ljava.util.$foo$.Optional;").toCharArray()));
 	assertEquals(
-			"Ljava.util.$foo$.Optional",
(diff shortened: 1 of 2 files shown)


iloveeclipse:
> * it appears there is other
>   code that is not prepared for the dollar being part of the class
>   name and expects the dollar being converted to dot

Could you please give more insights? Which code exactly, in JDT or Xtext?

iloveeclipse:
Is it possible to add test for this?

jjohnstn:
> > * it appears there is other
> >   code that is not prepared for the dollar being part of the class
> >   name and expects the dollar being converted to dot
> 
> Could you please give more insights? Which code exactly, in JDT or Xtext?

In Javadoc of Signature.toCharArray():  

Note: This method assumes that a type signature containing a <code>'$'</code>
 * is an inner type signature. While this is correct in most cases, someone could
 * define a non-inner type name containing a <code>'$'</code>. Handling this
 * correctly in all cases would have required resolving the signature, which
 * generally not feasible.

I will add a test to SignatureTests.

jjohnstn:
@iloveeclipse Just to confirm that the two XTest startswithdollar tests run succesfully with this change.

cdietrich:
can confirm this would make our test green again / restore the old behaviour

iloveeclipse (approved):

iloveeclipse:
Thanks Jeff for the fast fix & Christian for verification. Looks good, let's merge.

Issue #4731, opened 2026-01-09
Type error due to wrong binding of type variables in generic methods with wildcard argument type "T foo(X<? extends T> param)"


haumacher:
A regression in 4.39 M1 causes reporting a type error a situation where a parametric method is expected to "absorb" a wildcard type from its parameter and return the base type as result (the code was working at least in 4.36).

The following compiles without errors (the method `concat()` reduces  `Iterable<? extends T>` from its parameter to the base type `T`):

```
public class TestWildcard {

	private Collection<? extends Collection<? extends Runnable>> _parts;

	public Iterator<Runnable> iterator() {
		Iterable<Runnable> x = concat(_parts);
		return x.iterator();
	}

	public static <T> Iterable<T> concat(Iterable<? extends Iterable<? extends T>> entries) {
		return null;
	}

}
```

But when inlining `x` (directly using the result of `concat()` in as result of `iterator()`), to:

```
import java.util.Collection;
import java.util.Iterator;

public class TestWildcard {

	private Collection<? extends Collection<? extends Runnable>> _parts;

	public Iterator<Runnable> iterator() {
		return TestWildcard.concat(_parts).iterator();
	}

	public static <T> Iterable<T> concat(Iterable<? extends Iterable<? extends T>> entries) {
		return null;
	}

}
```

The error `Type mismatch: cannot convert from Iterator<capture#2-of ? extends Runnable> to Iterator<Runnable>` is reported. However, for me that seem to be the exact same situation which I would expect to be accepted, if the first example is considered well-typed. 

At first glance, this looks very similar to the problem reported at https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4730 but I'm not sure. However, I verified the examples above which were derived from a real-world situation in our code base (https://github.com/top-logic/tl-engine/blob/2ebc93e7e92aa15ace8df3329473a9a2c46d0782/com.top_logic.element/src/main/java/com/top_logic/element/model/PersistentScope.java#L201)

stephan-herrmann:
Behavior changed, indeed, error is first reported by 4.37.

```
1. ERROR in /tmp/TestWildcard.java (at line 9)
        return TestWildcard.concat(_parts).iterator();
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Type mismatch: cannot convert from Iterator<capture#2-of ? extends Runnable> to Iterator<Runnable>
```

[...]

PR #5101, opened 2026-06-02
Type error due to wrong binding of type variables in generic methods with wildcard argument type "T foo(X<? extends T> param)"

stephan-herrmann:
+ includes more precise computation of capturing supertype
  - no observable difference for any test including the new one here

Test for https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4731

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ParameterizedTypeBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ParameterizedTypeBinding.java
@@ -1455,11 +1455,10 @@ public class ParameterizedTypeBinding extends ReferenceBinding implements Substi
     		}
 	    }
 	    if (CapturingContext.isActive()) {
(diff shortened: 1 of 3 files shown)


stephan-herrmann:
> * includes more precise computation of capturing supertype
>    * no observable difference for any test including the new one here

Little explanation for this part:

[JLS 4.10.2](https://docs.oracle.com/javase/specs/jls/se26/html/jls-4.html#jls-4.10.2) says:
> Given a generic class or interface C with type parameters F1,...,Fn (n > 0), the direct supertypes of the parameterized type C<R1,...,Rn> where at least one of the Ri (1 ≤ i ≤ n) is a wildcard type argument, are the direct supertypes of the parameterized type C<X1,...,Xn> which is the result of applying capture conversion to C<R1,...,Rn> ([§5.1.10](https://docs.oracle.com/javase/specs/jls/se26/html/jls-5.html#jls-5.1.10)).

Previously we individually captured all type arguments of supertypes of `C<R1,...,Rn>`. With the change here we reverse the order: first capture (to yield `C<X1,...,Xn>`) then navigate to supertypes.

This is included here, because I happened to see this discrepancy while debugging the new test case.

PR #5128, opened 2026-06-09
Make readable names of CaptureBinding18 more explicit

stephan-herrmann:
+ print as special form of capture
  + naming update from JLS where they are now called "Y"
+ show all upper and lower bounds


stephan-herrmann:
@srikanth-sankaran  while comparing tracing from javac vs ecj I noticed that CaptureBinding18 was not visible as such. Of course, these beasts don't have a canonical syntax because they are purely internal concepts. Still they may appear in error messages, and simply hiding their capture nature can give quite a wrong impression of what the compiler is complaining about.

Long story short: do you want to glance through these changes? Test adjustments might give a good idea of the effect. Is this a good way to communicate with users?

For good measure I also migrated nomenclature from `Z#0-...` to `Y#0-...` to follow a renaming in JLS (many years ago).

PR #5137, opened 2026-06-11
Fix newline bug reported and add new test

inuyasha82:
## What it does
It fix the bug reported here: [eclipse.jdt.ui:3003](org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/formatter/FormatterTextBlockTests.java)

## How to test

1.  Make sure that the formatter option: New Lines/"After and before  opening and closing a text block " is enabled.
2.  Create a String like: 

```java
String str3 = """
			subd3
			ospl

			test""";
```

Make sure to keep the newline in that position.
3. Run the formatter 

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
+++ b/org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
@@ -418,11 +418,20 @@ public class TextEditsBuilder extends TokenTraverser {
 			int splitPlace = stringToCheck.indexOf(closingQuotes);
 			if (splitPlace > 0) {
 				// There is no new line because it is already present in the buffer field.
+				String newBuffer = getClosingQuotesBuffer(this.buffer.toString());
 				this.edits.add(getReplaceEdit(position + splitPlace, position + splitPlace,
-						'\\' + this.buffer.toString(), region));
+						'\\' + newBuffer, region));
 			}
 		}
 		return;
+	}
+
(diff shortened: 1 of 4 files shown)


org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
```diff
@@ -418,13 +418,22 @@ private void checkTextBlockClosingQuotes(int position, IRegion region) {
...
 			}
 		}
 		return;
 	}
 
+	private String getClosingQuotesBuffer(String curBuffer) {
+		// FOr the closing quote buffer, we want to ensure that there is only a single newline.
+		// But this can't be the case if the line before the last contains an extra newline (one or more).
+		// So we need to strip any extra newline.
+		StringBuilder newBuffer = new StringBuilder (this.options.line_separator + curBuffer.replace(this.options.line_separator, "")); //$NON-NLS-1$
```
jjohnstn, line 433: I don't think you need to use line_separator as the JVM handles this for you.  I believe you just need to use newline.

jjohnstn (changes requested):

org.eclipse.jdt.core/formatter/org/eclipse/jdt/internal/formatter/TextEditsBuilder.java
jjohnstn: Actually, I am wrong.  The resultant String will be the same whether on Windows or Linux so this is fine.

jjohnstn (approved):
Looks fine.

PR #5148, opened 2026-06-18
Detect duplicate case labels across promoted constant kinds

jarthana:
Duplicate case labels were not reported when two case constants had different compile-time constant kinds but the same value after binary numeric promotion to int.

The duplicate check in SwitchStatement#duplicateConstant relied on Constant#equals(), whose first guard is getClass() != obj.getClass(). Constants of byte/short/char/int that share a value but differ in kind (e.g. ByteConstant vs IntConstant) therefore compared unequal.

Per JLS 5.6 / 14.11.1, case constants of byte, short, char and int are compared after binary numeric promotion to int. Introduce Constant#compareAfterPromoting() implementing this rule once, in the base class, so detection is symmetric regardless of label order, and use it at the call site instead of equals(). long/float/double/boolean switches still require an exactly-typed case constant, so they are left unaffected.

## Author checklist


srikanth-sankaran:
@mpalat  - could you review this please ? TIA

mpalat (approved):
+1
Changes look good. @jarthana once the tests pass I leave it to you to decide on addressing the Co-Pilot's suggestions.

PR #5157, opened 2026-06-20
Multi-Release JAR is not recognized as such

### LINKED ISSUE #2495: Multi-Release JAR is not recognized as such
Recently, I update com.google.code.gson:gson to version 2.11.0 from version 2.10.1.

The project use module-info.java and use "requires com.google.gson;"

gson-2.11.0.jar is a Multi-Release JAR and the module declaration is at META-INF/versions/9/module-info.class

I open [an issue ](https://github.com/google/gson/issues/2679) and it got the full details.
Also included in that issue a small project to demonstrate the problem.(https://github.com/google/gson/files/15412480/ModuleTest.zip)

**When using pure maven build "mvn clean test" it is compiled successfully.
Also I can run it using  "mvn compile exec:java -Dexec.mainClass="main.Main"" .**

I use: 
M2E Maven Integration for Eclipse Core 2.6.0.20240220-1109
Eclipse 2024-03 (4.31)
Java openjdk version "22.0.1" 2024-04-16

Thanks in advance for any response.

stephan-herrmann:
+ overhaul jar scanning strategy:
  - check the MANIFEST for "Multi-Release: true"
  - scan all jar entries searching for versions, since dirs are optional
  - store versions not paths for easier comparison
  - JarPackageFragmentRoot collects versions only via class overrides
+ add more challenges to the test:
  - no directory entries
  - multiple versions of a/B.class with different methods
  - enable jar as "Multi-Release: true" in MANIFEST.MF

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2495


stephan-herrmann:
Fixing several issues of how JarPackageFragmentRoot and ClasspathMultiReleaseJar (builder) interpret multi-release jars:

(1) Always first check for "Multi-Release: true" in the manifest, otherwise ignore any version folders.

(2) Don't rely on directory entries in the jar. My worries about performance of jar scanning were unnecessary: we already perform such scans in various places.
* for JarPackageFragmentRoot the essential thing to capture is the mapping from class names to versions, no additional scan needed.
* for JarPackageFragmentRoot a similar scan happens during `readPackageNames()`. If we can assert that this method will always be invoked before `supportedVersions` is accessed then we only need to ensure that that method correctly implements the MR semantics and `scanMultiReleaseVersions()` can be removed altogether.

In JarPackageFragmentRoot the way we collect `overridden` was actually bogus: it didn't respect the precedence of higher versions.

stephan-herrmann:
I out-sourced a related observation to #5160

Issue #4979, opened 2026-03-30
ECJ doesn't report expected error about domination


jarthana:
Here's the testcase:

```
public class X {
	public int foo(Character c) {
		int result = 0;
		switch (c) {
			case Character c1 -> {
				result = c1;
				break;
			}
			case 0 -> {  // Same goes for case (int) 0
				result = 0;
				break;
			}
		}
		return result;
	}
}
```

Javac reports this -` error: this case label is dominated by a preceding case label` but we allow this.

srikanth-sankaran:
I'll investigate.

iloveeclipse:
Javac 25 doesn't report any warning or error for me?

srikanth-sankaran:
> Javac reports this -` error: this case label is dominated by a preceding case label` but we allow this.

Can you double check ? I don't get any error from javac 26 either

jarthana:
@srikanth-sankaran Can you please try Javac with preview enabled?

srikanth-sankaran:
> [@srikanth-sankaran](https://github.com/srikanth-sankaran) Can you please try Javac with preview enabled?

Wow, you know that crossed my mind although I didn't try it - being 100% certain they won't change the meaning of programs with new features. So with the primitive features enabled the meaning of the program changes in a constricting way ? (legal to illegal)

srikanth-sankaran:
FTR, This code compiles with JDK25 with preview enabled.

~/jdk/jdk-25/jdk-25.0.1/bin/javac --enable-preview -source 25  -g X.java

srikanth-sankaran:
@mpalat @stephan-herrmann - this being a preview feature I think you want to work on this ?

mpalat:
yes @srikanth-sankaran . Will work on this once I am back

mpalat:
The above pr fixes the test case given, but some variants need to be looked more closely in terms of dominance rules. Looking into the same - hence keeping this open

mpalat:
In the context in https://cr.openjdk.org/~abimpoudis/instanceof/jep532-20260504/specs/primitive-types-in-patterns-instanceof-switch-jls.html.
```java
public class X {
        public int foo1(Short c) {
                int result = 0;
                switch (c) {
                  case Short c1 -> {
                        result = c1;
                        break;
                  }
                  case (byte) 0 -> { // what is the reason for this case to be dominated - as per spec?
                        result = 0;
                        break;
                  }
                }
                return result;
        }
}
[...]

PR #5166, opened 2026-06-25
Fixing the Short_byte_dominance issue of issue #4979

mpalat:
## What it does
Fixes the dominance issue as mentioned in https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4979#issuecomment-4715810255
## How to test
PrimitivesInPatternsTest.testDominanceIssue4979_002() is the junit test case to test

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/PrimitiveInPatternsTest.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/PrimitiveInPatternsTest.java
@@ -30,7 +30,7 @@ public class PrimitiveInPatternsTest extends AbstractRegressionTest9 {
 	static {
 //		TESTS_NUMBERS = new int [] { 1 };
 //		TESTS_RANGE = new int[] { 1, -1 };
-//		TESTS_NAMES = new String[] { "testDominanceIssue4979_001" };
+//		TESTS_NAMES = new String[] { "testDominanceIssue4979_002" };
 	}
 	private String extraLibPath;
 	public static Class<?> testClass() {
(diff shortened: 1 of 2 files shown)


stephan-herrmann:
As per [this post](https://mail.openjdk.org/archives/list/amber-spec-experts@openjdk.org/message/3FESIKU55QW7X3LCGHWBIJOZTWL6XKRP/) the spec will say:
> Any label is dominated if there is a preceding unguarded case label in the switch block with a case pattern p where p is unconditional for the type of the selector expression

This rule doesn't seem to be specific to primitives in any way. So why does this PR restrict the change to primitives?

I understand why this PR is merged to master, but the resulting situation is a bit confusing: master and javac 26 (and earlier?) conform to a rule that will appear in JLS probably in 28. :shrug: 

What's more: #4979 targets BETA_JAVA27_M3 and is still open. Please coordinate :)

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/PrimitiveInPatternsTest.java
```diff
@@ -7636,16 +7636,72 @@ public int foo(Character c) {
...
+					}
+				}
+				"""
+			},
+				"----------\n" +
+				"1. ERROR in X.java (at line 15)\n" +
+				"	Zork();\n" +
+				"	^^^^\n" +
+				"The method Zork() is undefined for the type X\n" +
+				"----------\n"
```
stephan-herrmann, line 7703: Such unrelated errors spoil comparison in run.javac mode: both compiler find the trivial error and so comparison succeeds. It would be more interesting to test if both compilers succeed in the core part of the test.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/PrimitiveInPatternsTest.java
```diff
@@ -7636,16 +7636,72 @@ public int foo(Character c) {
...
+				"""
+				@SuppressWarnings("preview")
+				public class X {
+					void foo() {
+						int j = 1;
+						switch(j) {
+							case byte b ->
+								System.out.println("A byte");
+							case 260 ->						// not dominated
+								System.out.println("An int that can be represented as a byte exactly");
```
stephan-herrmann, line 7687: The printout string is inconsistent with the case constant `260`.

Perhaps we should have two test cases out of this: one with a case constant that is dominated, and one that passes compilation?

stephan-herrmann:
> This rule doesn't seem to be specific to primitives in any way. So why does this PR restrict the change to primitives?

There are two parts to this: Any changes are indeed happening under the preview flag, only. So asking `PRIMITIVES_IN_PATTERNS.isSupported()` is good.  It's the last part `&& resolvedType1.isPrimitiveType()` which still looks unmotivated to me. OTOH, I found no example where it makes a difference...

stephan-herrmann (commented):
Although being late here, I dropped two more comments concerning the new tests.

mpalat:
@stephan-herrmann sorry could not respond earlier - will take a look and will address these comments. Thanks for taking a deeper look at this.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/PrimitiveInPatternsTest.java
mpalat: resolved via #5180

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/PrimitiveInPatternsTest.java
mpalat: resolved via #5180

mpalat:
> @stephan-herrmann sorry could not respond earlier - will take a look and will address these comments. Thanks for taking a deeper look at this.

resolved via #5180

PR #5168, opened 2026-06-29
Fix completions to offer fields and types after method returning type

jjohnstn:
- add new CompletionOnExpressionOfType class to represent completing after an expression that returns a type (in this case a method or constructor call)
- add new completionOnExpressionOfType() method to CompletionEngine and call it when node is new CompletionOnExpressionOfType
- modify CompletionEngine.findFields() and findLocalMethods() to add a dot operator to fields and methods found when we are dealing with a CompilationOnExpressionOfType node
- add new checkMethodResult() method to CompletionParser and call it in completionIdentifier() just before default checkNameCompletion()
- add new pushCompletionOnMethodCallResultOnExpressionStack() which pushes a CompletionOnExpresionOnType instance on expression stack
- add new tests to CompletionTests
- needed for: https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/832

## What it does

See issue.

## How to test

See issue or new tests.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/CompletionTests.java
@@ -26252,9 +26252,7 @@ public void testIssue832a() throws Exception {
 	int cursorLocation = str.lastIndexOf(completeBehind) + completeBehind.length();
 	this.workingCopies[0].codeComplete(cursorLocation, requestor, this.wcOwner);
 	assertResults(
-			"clone[METHOD_REF]{.clone(), Ljava.lang.Object;, ()Ljava.lang.Object;, clone, null, 52}\n" +
 			"equals[METHOD_REF]{.equals(), Ljava.lang.Object;, (Ljava.lang.Object;)Z, equals, (obj), 52}\n" +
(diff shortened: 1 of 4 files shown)


stephan-herrmann:
FYI: The huge portion of white space changes in CompletionParser is costing a lot of work downstream.

In my current case it's a fork of jdt.core that currently has near 100% merge conflicts on this file.

For the future: @jjohnstn if you think something like fully formatting a source file is needed, before changes can be made to it, I'd appreciate VERY MUCH, if such changes are isolated into a commit of their own, describing how the change was performed.

Edit: At first I thought it's only indentation but apparently the enter file has been reformatted. If you ask me, this format change isn't even an improvement. To the contrary, much content looked better readable to me before the reformatting, in particular code comments made at the right of a long line of code make no sense when squeezed into a column of very few characters.  I'm not good at reading long text vertically ;-P  

Surely git blame will have issues with this file in the future from now on.

jjohnstn:
Hi @stephan-herrmann I didn't mean to reformat the file.  It was an accident.  I realize that causes you grief and try to avoid it.  I must have had format on save set.

stephan-herrmann:
> Hi @stephan-herrmann I didn't mean to reformat the file. It was an accident. I realize that causes you grief and try to avoid it. I must have had format on save set.

Thanks!

PR #5173, opened 2026-07-01
Pattern variable rejected as not in scope after a labeled block exited via break <label> #5172

### LINKED ISSUE #5172: Pattern variable rejected as not in scope after a labeled block exited via break <label>
ECJ reports a spurious compile error for a pattern variable that is declared in a while condition, when the while loop is inside a labeled statement and the loop body exits only via a break to that enclosing label. javac compiles the same code without error.

In the following code, the pattern variable s is definitely matched on normal completion of the loop, so it must be in scope afterwards.

```
public class X {
	public static void foo(Object obj) {
		Object obj1 = obj;
		Object obj2 = null;
		X: {
			while (!(obj1 instanceof String s)) {
				break X;
			}
			obj2 = s;
		}
		System.out.println(obj2);
	}
	public static void main(String[] argv) {
		new X().foo("a");
	}
}
```

jarthana:
## What it does
Fixes #5172

## Author checklist


srikanth-sankaran:
I offer to review after all tests are green.

jarthana:
Srikanth, there are two changes as part of this PR.
1. The change in LabeledStatement ensures that the variable bindings in scope of a statement (for e.g., while loop) are fetched even when it's wrapped in a labeled statement. This is guarded by a check that tests that the break label is not the same label as that of the labeled statement.
2. The second change in Statement, visit(...) - Earlier, breaksOut returned true if a null label was passed. But I think that should only be allowed when for a non-labeled break statement.

Note: The comments and some of the code were added by the bot :)

srikanth-sankaran (approved):
Please proceed.

PR #5178, opened 2026-07-03
Qualifier change to fix split package signing problem

merks:
https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/3927


akurtakov:
As this is rebuild to fix signatures and the urge to get M1 build ASAP I'm merging this one without waiting for verification.

PR #5189, opened 2026-07-11
ecj spams the console with irrelevant zip exceptions

### LINKED ISSUE #1578: ecj spams the console with irrelevant zip exceptions

stephan-herrmann:
+ improve error message
+ remove printing of stack traces
+ make err printout testable

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1578

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathLocation.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathLocation.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2017 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -39,6 +39,7 @@ public abstract class ClasspathLocation implements FileSystem.Classpath,
 	char[] normalizedPath;
 	public final AccessRuleSet accessRuleSet;
 	IModule module;
+	private PathKind kind = PathKind.CP;
 
 	public final String destinationPath;
 		// destination path for compilation units that are reached through this
@@ -55,6 +56,16 @@ public abstract class ClasspathLocation implements FileSystem.Classpath,
 			String destinationPath) {
 		this.accessRuleSet = accessRuleSet;
 		this.destinationPath = destinationPath;
+	}
+
+	@Override
+	public PathKind getPathKind() {
+		return this.kind;
+	}
+
+	@Override
+	public void setPathKind(PathKind kind) {
+		this.kind = kind;
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


stephan-herrmann:
Implements the following idea from #1578:
>> @jukzi:
>> do not log stacktrace for ZipException - only log filename
>
> @iloveeclipse:
> I would be fine with that, assuming we produce error messages distinct enough to identify the exact place in compiler code that produces it.

I removed `printStackTrace()` only in one location, and at the same time improved the error messages as to tell the user, how exactly (by which command line option) the offending file entered compilation.

In order to make this testable, I also redirect the printout via `Main.err` (since plain `System.err.println()` almost never is a good idea to begin with).

This comes with the risk, that some build tools will only now begin to "see" these messages, and ecj cannot know, how those tools will interpret this (perhaps this will cause them to mark a compilation as failed).

PR #5192, opened 2026-07-11
Reject non-exhaustive nested sealed record patterns in switch #5182

### LINKED ISSUE #5182: ECJ accepts a non-exhaustive switch over nested sealed record patterns that `javac` rejects
## Summary

When a `switch` expression matches on a sealed interface, and one case uses a **record deconstruction pattern that narrows a nested component to only one of that component's own sealed permitted subtypes** (without covering the others, and without a `default`), ECJ compiles it without error or warning. `javac` rejects the same code with "the switch expression does not cover all possible input values". The ECJ-compiled class throws `java.lang.MatchException` at runtime when given an input that hits the uncovered case.

## Environment

- ECJ: `org.eclipse.jdt.core.compiler.batch_3.45.0.v20260224-0835` (bundled with Eclipse, JDT Core 3.45.0)
- Compared against: `javac` 25.0.1 (Oracle OpenJDK), both with default options and with `-source 21 -target 21`
- Both compilers invoked directly (standalone `ecj.jar`/batch compiler vs. `javac`), not through any build tool

## Reproducer

```java
public class ExhaustivenessGap {

    sealed interface Position permits GlobalPosition, StartPosition {}
    record GlobalPosition(long index) implements Position {}
    record StartPosition() implements Position {}

    sealed interface SourcingStrategy permits Absolute, Snapshot {}
    record Absolute(Position position) implements SourcingStrategy {}
    record Snapshot(Position maximumPosition) implements SourcingStrategy {}

    // This switch narrows Absolute's component to GlobalPosition only - it does NOT
    // cover Absolute(StartPosition). There is no default branch. javac correctly rejects
    // this as non-exhaustive. ECJ accepts it without error or warning.
    static long toIndex(SourcingStrategy strategy) {
        return switch (strategy) {
            case Absolute(GlobalPosition p) -> p.index();
            case Snapshot(Position p) -> p instanceof GlobalPosition g ? g.index() : -1;
        };
    }

        System.out.println("Absolute(GlobalPosition): " + toIndex(new Absolute(new GlobalPosition(42))));
        System.out.println("Snapshot(StartPosition): " + toIndex(new Snapshot(new StartPosition())));

        System.out.println("Absolute(StartPosition): about to throw MatchException...");
        System.out.println(toIndex(new Absolute(new StartPosition())));
    }
}
```

## Steps to reproduce

```
java -jar ecj.jar -21 -d out ExhaustivenessGap.java
java -cp out ExhaustivenessGap
```

## Expected result

Compilation should fail, matching `javac`:

```
ExhaustivenessGap.java:17: error: the switch expression does not cover all possible input values
        return switch (strategy) {
               ^
1 error
```

## Actual result

ECJ compiles with exit code 0, no errors, no warnings. Running the class produces:

```
Absolute(GlobalPosition): 42
Snapshot(StartPosition): -1
Absolute(StartPosition): about to throw MatchException...
Exception in thread "main" java.lang.MatchException
	at ExhaustivenessGap.toIndex(ExhaustivenessGap.java:17)
	at ExhaustivenessGap.main(ExhaustivenessGap.java:28)
```

## Analysis

The switch's outer selector type (`SourcingStrategy`) is sealed to `Absolute` and `Snapshot`, and both are covered by exactly one case each - so far, correctly exhaustive at that level. However, `Absolute`'s own record component is typed as `Position`, itself sealed to `GlobalPosition` and `StartPosition`. The `case Absolute(GlobalPosition p)` arm only matches when `Absolute`'s `position` component happens to be a `GlobalPosition` - it does not match `Absolute(StartPosition)`. Since no other case (and no `default`) covers that remaining combination, the switch is not actually exhaustive over all reachable runtime values of `SourcingStrategy`, and per JLS pattern-matching exhaustiveness rules, this should be a compile error. `javac` enforces this correctly (confirmed with both types declared in the same compilation unit and with the sealed types precompiled into a separate JAR dependency - same result either way). ECJ does not perform this recursive exhaustiveness check for nested record patterns and silently accepts the code, deferring the failure to a runtime `MatchException`.

## Impact

Code relying on IDE feedback from Eclipse (backed by ECJ) will show no compile error or warning for this pattern, while a `javac`-based build (e.g. via Maven/Gradle) will correctly reject it - or, if compiled exclusively via ECJ/Eclipse-based tooling, will silently ship code with a latent `MatchException` risk for any input the narrowed nested pattern doesn't cover.

kamilkrzywanski:
## Summary

- Fixes incorrect exhaustiveness for switches on sealed types when a case uses a **record pattern that only covers some nested sealed subtypes** of a component (e.g. `case Absolute(GlobalPosition p)` without covering `Absolute(StartPosition)`).
- ECJ previously treated the permitted record type as covered by type name alone, accepted the switch, and deferred failure to a runtime `MatchException`. `javac` correctly rejects the same code.
- For permitted record types, reuse existing record-pattern coverage (`RNode` / `CoverageCheckerVisitor`) so nested components are checked (JLS 14.11.1.1). Type patterns such as `case Absolute a` still fully cover the record.
- Adds regression tests in `SwitchPatternTest`.

Fixes #5182

## Test plan

- [x] Reproduced with ECJ (accepted) vs `javac` (rejected) using the issue sample
- [x] After fix, ECJ reports non-exhaustive switch expression for the sample
- [x] Exhaustive nested cases and type-pattern coverage still compile
- [x] `SwitchPatternTest` — 334 tests, 0 failures

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/SwitchStatement.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/SwitchStatement.java
@@ -148,7 +148,9 @@ public class SwitchStatement extends Expression {
 		}
 
 		void addPattern(Pattern p) {
-			if (p instanceof RecordPattern rp && TypeBinding.equalsEquals(this.type, rp.type.resolvedType) && this.firstComponent != null)
+			if (p instanceof RecordPattern rp && rp.type.resolvedType != null
+					&& TypeBinding.equalsEquals(this.type.erasure(), rp.type.resolvedType.erasure())
+					&& this.firstComponent != null)
 				this.firstComponent.addPattern(rp, 0);
 		}
 
@@ -269,7 +271,7 @@ public class SwitchStatement extends Expression {
 				}
 			}
 			if (node.type instanceof ReferenceBinding ref && ref.isSealed()) {
-				this.covers &= caseElementsCoverSealedType(ref, availableTypes);
+				this.covers &= caseElementsCoverSealedType(ref, availableTypes, false);
 				return this.covers;
 			}
 			return this.covers = false; // no need to visit further.
@@ -413,7 +415,8 @@ public class SwitchStatement extends Expression {
 			return;
 		}
 
-		if (JavaFeature.PATTERN_MATCHING_IN_SWITCH.isSupported(compilerOptions) && selectorType.isSealed() && caseElementsCoverSealedType((ReferenceBinding) selectorType, this.caseLabelElementTypes))
+		if (JavaFeature.PATTERN_MATCHING_IN_SWITCH.isSupported(compilerOptions) && selectorType.isSealed()
+				&& caseElementsCoverSealedType((ReferenceBinding) selectorType, this.caseLabelElementTypes, true))
 			this.switchBits |= SwitchStatement.Exhaustive;
 		else if (selectorType.isRecordWithComponents() && this.containsRecordPatterns && caseElementsCoverRecordType(upperScope, compilerOptions, (ReferenceBinding) selectorType))
 			this.switchBits |= SwitchStatement.Exhaustive;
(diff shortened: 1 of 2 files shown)


srikanth-sankaran:
Manoj - this is your code simplified by me - could you be the first reviewer please ? TIA

srikanth-sankaran:
@mpalat - I gave you a wrong PR number. Here is the PR where I simplified your code retaining the original design entirely. 
https://github.com/eclipse-jdt/eclipse.jdt.core/pull/3172

Let me know if I inadvertently broke something!

mpalat:
> #3172


Thanks @srikanth-sankaran. Taking a look

mpalat (approved):
+1
Code looks good. Thanks @kamilkrzywanski !

PR #5196, opened 2026-07-15
Let insert_new_line_before_closing_brace_in_array_initializer support NEXT_LINE_ON_WRAP

### LINKED ISSUE #5195: Formatter: let insert_new_line_before_closing_brace_in_array_initializer support a wrap-conditional value
`org.eclipse.jdt.core.formatter.brace_position_for_array_initializer` already supports `NEXT_LINE_ON_WRAP`, which places the **opening** brace on its own line only if the array initializer's content actually wraps across multiple lines, and keeps it inline otherwise. There's no equivalent for the **closing** brace: `insert_new_line_before_closing_brace_in_array_initializer` only accepts `INSERT` / `DO_NOT_INSERT`, so it's all-or-nothing regardless of whether the content wrapped.

This means today you can get:
```java
int[] shortArr = { 1, 2, 3 };

int[] longArr = {
    1111111111,
    2222222222,
    3333333333 };   // closing brace stuck to the last element
```
but not a closing brace that follows the same "only break if wrapped" rule the opening brace already supports, which reads more naturally:
```java
int[] longArr = {
    1111111111,
    2222222222,
    3333333333
};
```

**Proposed fix:** extend the accepted values of `insert_new_line_before_closing_brace_in_array_initializer` to include a value mirroring `NEXT_LINE_ON_WRAP`. The underlying mechanism for this already exists and looks reusable as-is: `Token.setNextLineOnWrap()` / `isNextLineOnWrap()`, currently only invoked for the opening brace in `LineBreaksPreparator.visit(ArrayInitializer)`. Calling `closeBraceToken.setNextLineOnWrap()` there (guarded by the new option value) instead of the current unconditional `closeBraceToken.breakBefore()` appears to plug into the same deferred wrap-resolution pass (`WrapExecutor.WrapsApplier`) that already resolves the opening brace's flag — happy to prototype this and open a PR with formatter regression tests if that approach sounds right to maintainers.

netomi:
Fixes #5195

## What
`brace_position_for_array_initializer` already supports `NEXT_LINE_ON_WRAP` for the opening brace: it moves to its own line only if the array initializer's content wraps across multiple lines, staying inline otherwise. The closing brace had no equivalent — `insert_new_line_before_closing_brace_in_array_initializer` only accepted `INSERT`/`DO_NOT_INSERT`, so it was all-or-nothing regardless of whether the content actually wrapped.

This PR adds `NEXT_LINE_ON_WRAP` as a third accepted value for `insert_new_line_before_closing_brace_in_array_initializer`, reusing the deferred `Token.setNextLineOnWrap()` / `isNextLineOnWrap()` mechanism already used for the opening brace, resolved later by `WrapExecutor` once wrapping is decided.

Before (with `brace_position_for_array_initializer=NEXT_LINE_ON_WRAP`, closing brace unconditional):
```java
int[] shortArr = { 1, 2, 3 };

int[] longArr = {
    1111111111,
    2222222222,
    3333333333 };   // stuck to the last element regardless of wrapping
```
After (with the new value on the closing-brace option too):
```java
int[] shortArr = { 1, 2, 3 };   // unwrapped: unchanged

int[] longArr = {
    1111111111,
    2222222222,
    3333333333
};                               // wrapped: closing brace gets its own line
```

## How
- `DefaultCodeFormatterOptions` gains a new `insert_new_line_before_closing_brace_in_array_initializer_on_wrap` boolean field, kept separate from the existing `insert_new_line_before_closing_brace_in_array_initializer` field so none of the existing call sites that set the old field directly (in `FormatterBugsTests`/`FormatterRegressionTests`) need to change.
- `LineBreaksPreparator.visit(ArrayInitializer)` calls `closeBraceToken.setNextLineOnWrap()` instead of the unconditional `breakBefore()` when the new flag is set.
- `DefaultCodeFormatterConstants` javadoc updated to document the new value.

## Testing
Added `testIssue5195_shortArrayStaysInline` and `testIssue5195_wrappedArrayGetsOwnLine` to `FormatterBugsTests`, covering both the unwrapped (no-op) and wrapped (brace gets its own line) cases. I also verified backward compatibility of the existing `INSERT`/`DO_NOT_INSERT` values is unaffected — both still produce identical output to before this change.

I wasn't able to run the full Tycho-based test suite in my environment (this repo depends on a sibling `eclipse-platform-parent` checkout for its build), so I instead verified the formatter behavior directly: compiled the three modified classes against the released 3.43.0 artifacts (`org.eclipse.jdt.core`, `ecj`, `org.eclipse.jface.text`, `org.eclipse.text`), loaded them ahead of the stock jar via `ToolFactory.createCodeFormatter`, and confirmed the output for both the wrapped and unwrapped cases matches what's shown above, plus regression-checked the untouched `INSERT`/`DO_NOT_INSERT` paths produce byte-identical output to the unpatched formatter. Happy to have a maintainer confirm against the full test suite, and to adjust naming/approach based on review feedback.

---
*This PR was drafted with assistance from Claude (Anthropic).*

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/formatter/FormatterBugsTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/formatter/FormatterBugsTests.java
@@ -13408,4 +13408,40 @@ public void testIssue4122() {
 		""";
 	formatSource(source);
 }
+
+/**
+ * https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5195
+ * insert_new_line_before_closing_brace_in_array_initializer should support NEXT_LINE_ON_WRAP,
+ * mirroring brace_position_for_array_initializer: closing brace goes on its own line only if
+ * the array initializer's content itself wrapped across multiple lines.
+ */
+public void testIssue5195_shortArrayStaysInline() {
+	this.formatterPrefs.insert_new_line_before_closing_brace_in_array_initializer_on_wrap = true;
+	String source =
+		"""
+		public class A {
+			int[] shortArr = { 1, 2, 3 };
+		}
+		""";
+	formatSource(source);
+}
+public void testIssue5195_wrappedArrayGetsOwnLine() {
+	this.formatterPrefs.page_width = 40;
+	this.formatterPrefs.insert_new_line_before_closing_brace_in_array_initializer_on_wrap = true;
+	String source =
+		"""
+		public class A {
+			int[] longArr = { 1111111111, 2222222222, 3333333333, 4444444444 };
+		}
+		""";
+	formatSource(source,
+		"""
+		public class A {
+			int[] longArr = { 1111111111,
+					2222222222, 3333333333,
+					4444444444
+			};
+		}
+		""");
+}
 }
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/formatter/FormatterBugsTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/formatter/FormatterBugsTests.java
@@ -13444,4 +13444,78 @@ public void testIssue5195_wrappedArrayGetsOwnLine() {
 		}
 		""");
 }
+
+/**
+ * https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5195
+ * With NEXT_LINE_ON_WRAP, an existing new line before the closing brace is removed when the
+ * initializer's elements are not wrapped - here because wrapping is disabled altogether.
+ */
+public void testIssue5195_noWrappingRemovesNewLineBeforeClosingBrace() {
+	this.formatterPrefs.insert_new_line_before_closing_brace_in_array_initializer_on_wrap = true;
+	this.formatterPrefs.alignment_for_expressions_in_array_initializer = Alignment.M_NO_ALIGNMENT;
+	String source =
+		"""
+		public class A {
+			int[] arr = { 1111111111, 2222222222,
+					3333333333
+			};
+		}
+		""";
+	formatSource(source,
+		"""
+		public class A {
+			int[] arr = { 1111111111, 2222222222, 3333333333 };
+		}
+		""");
+}
+
+/**
+ * https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5195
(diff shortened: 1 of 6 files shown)


jjohnstn (changes requested):
Works well.  Thanks for the UI to test it.  

Just a minor comment to add some more tests.  Have a test where the array init is wrapped and has a newline before closing brace, then specify either no wrapping or wrap if necessary and verify with the new setting that it puts the whole initializer on one line with no newline before closing brace.  A second test could verify that forced wrapping (always wrap initializer) results in the desired behavior.

netomi:
Thanks for the review! Added three tests as suggested in `FormatterBugsTests`:

- `testIssue5195_noWrappingRemovesNewLineBeforeClosingBrace` — already-wrapped initializer with a new line before the closing brace, formatted with wrapping disabled (`M_NO_ALIGNMENT`) → collapses onto one line, new line before the closing brace removed.
- `testIssue5195_wrapWhereNecessaryRemovesNewLineBeforeClosingBrace` — same input with "wrap where necessary" (`M_COMPACT_SPLIT`) where the elements fit on one line → same result.
- `testIssue5195_forcedWrappingGetsOwnLine` — with forced wrapping (`M_ONE_PER_LINE_SPLIT | M_FORCE`) the elements always wrap, so the closing brace gets its own line even for an initializer that would otherwise fit on a single line.

I still can't run the Tycho suite locally, so I verified with the standalone harness again (patched formatter loaded ahead of the stock jar), driving the exact literals extracted from the test methods themselves, including the repeat-format check `formatSource` performs. I also ran a `DO_NOT_INSERT` / `INSERT` / `NEXT_LINE_ON_WRAP` × alignment-mode matrix to confirm the new value behaves as `DO_NOT_INSERT` when unwrapped and as `INSERT` when wrapped, and that the forced-wrap test genuinely discriminates (`3 };` with `DO_NOT_INSERT` vs `3` + `};` on its own line with the new value).

One thing worth flagging: the two collapse tests produce output identical to plain `DO_NOT_INSERT`. That is the intended semantics, but it does mean they would also pass without this change — they still add value in verifying the formatter actively *removes* an existing new line rather than merely not adding one. The discriminating test for the new behaviour is `testIssue5195_forcedWrappingGetsOwnLine` (alongside the existing `testIssue5195_wrappedArrayGetsOwnLine`).

---
*This comment was drafted with assistance from Claude (Anthropic).*

jjohnstn (approved):
Looks good.  We'll have to wait until the repo opens up after 4.41 RC2 (RC1 is already past and RC2 next week, is only for important fixes).

PR #5205, opened 2026-07-17
Avoid ClassCastException for array capture erasures #5204

### LINKED ISSUE #5204: CaptureBinding18 casts an array erasure to ReferenceBinding

## Steps to reproduce

Save this Java source as `T8062977.java`:

```java
import java.util.List;

class T8062977 {
    <T extends B, B> T m(Class<B> cb) { return null; }

    void test1(Class<Iterable<?>> cb) {
        List<Integer>[] r1 = m(cb);
        List<Integer> r2 = m(cb);
    }

    void test2(Class<Iterable<?>[]> cb) {
        List<Integer>[] r1 = m(cb);
        List<Integer> r2 = m(cb);
    }

    void test3(Class<Iterable<?>[][]> cb) {
        List<Integer>[][] r1 = m(cb);
        List<Integer>[] r2 = m(cb);
        List<Integer> r3 = m(cb);
    }
}
```

This is a negative compiler test from [T8062977.java](https://github.com/openjdk/jdk/blob/be40b6bcdab37368ea3c769e575b36e290d0c6a1/test/langtools/tools/javac/generics/inference/8062977/T8062977.java) at `openjdk/jdk@be40b6bcdab37368ea3c769e575b36e290d0c6a1`. OpenJDK marks it with `@compile/fail` and records the expected source diagnostics in [T8062977.out](https://github.com/openjdk/jdk/blob/be40b6bcdab37368ea3c769e575b36e290d0c6a1/test/langtools/tools/javac/generics/inference/8062977/T8062977.out). The file is syntactically valid Java, but several assignments are deliberately incompatible.

Build the batch compiler from the tested JDT checkout, then run:

```text
java -jar ecj.jar -25 T8062977.java
```

I expected: ECJ rejects the source with ordinary type-mismatch diagnostics and a nonzero exit status. Successful compilation is not expected.

But got: ECJ throws `ClassCastException` while processing the invalid assignments instead of completing normal diagnostic reporting.

During type inference, ECJ creates an internal capture type to represent an unknown wildcard. Because this source is invalid, one provisional upper bound has an array as its erased type.

Erasure removes generic details before ECJ compares or combines bounds. For a normal intersection such as `A & B`, [`CaptureBinding18.erasure()`](https://github.com/eclipse-jdt/eclipse.jdt.core/blob/792a9156c119d28063904c85726e486ec2fb2206/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding18.java) starts with the leftmost erased type and combines later class or interface bindings. The method cast every erased bound to `ReferenceBinding`, but an `ArrayBinding` is not a `ReferenceBinding`.

Invalid source is routine compiler input, and type inference still has to represent intermediate bounds while it determines which assignments are invalid. An invalid assignment may produce a provisional non-reference bound, but it must not make the compiler's own Java code throw an exception.

Reference-only intersections must keep their existing behavior. If a provisional invalid bound erases to a non-reference type, the method must return the leftmost erasure without making that cast. The compilation must then finish through the ordinary source-diagnostic path.

```text
java.lang.ClassCastException: class ArrayBinding cannot be cast to class
ReferenceBinding
    at CaptureBinding18.erasure(CaptureBinding18.java:101)
```

- Existing-issue search: no issue matching `T8062977`, `CaptureBinding18`, and `ArrayBinding` was found

## Community

martinfrancois:
## What it does

Fixes #5204.

The OpenJDK regression test [`T8062977.java`](https://github.com/openjdk/jdk/blob/be40b6bcdab37368ea3c769e575b36e290d0c6a1/test/langtools/tools/javac/generics/inference/8062977/T8062977.java) contains both valid and invalid assignments. ECJ previously rejected two valid array assignments and threw `ClassCastException` while checking two invalid assignments.

The failure started during generic method inference. ECJ kept only one upper bound whose type was not a `ReferenceBinding`. This lost an additional array bound and made inference fall back to an invalid capture. `CaptureBinding18.erasure()` later tried to cast the array binding to `ReferenceBinding`.

This change:

- keeps every proper upper bound so greatest-lower-bound reduction can select a compatible array type;
- checks array component types when it looks for common generic supertypes;
- rejects an intersection when its array type is incompatible with another bound;
- updates a capture only after all new upper bounds pass validation.

The valid assignments now compile. The invalid assignments produce normal type-mismatch diagnostics without an internal compiler exception. There is no public API change.

## How to test

- Run `GenericsRegressionTest_1_8`: 5,652 tests pass with no failures or errors.
- Run `GenericsRegressionTest`, `JEP286Test`, and `LambdaExpressionsTest`: 10,152 tests pass with no failures or errors.
- The new conforming test covers compatible one-dimensional and two-dimensional array upper bounds.
- The new negative tests cover an incompatible scalar target and an unresolved capture requested as an array type. Both produce ordinary type-mismatch diagnostics without an internal compiler exception.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/BoundSet.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/BoundSet.java
@@ -111,27 +111,18 @@ class BoundSet {
 		// pre: this.subBounds != null
 		public TypeBinding[] upperBounds(boolean onlyProper, InferenceVariable variable) {
 			TypeBinding[] rights = new TypeBinding[this.subBounds.size()];
-			TypeBinding simpleUpper = null;
 			Iterator<TypeBound> it = this.subBounds.iterator();
 			long nullHints = variable.nullHints;
 			int i = 0;
 			while(it.hasNext()) {
 				TypeBinding right=it.next().right;
 				if (!onlyProper || right.isProperType(true)) {
-					if (right instanceof ReferenceBinding) {
-						rights[i++] = right;
-						nullHints |= right.tagBits & TagBits.AnnotationNullMASK;
-					} else {
-						if (simpleUpper != null)
-							return Binding.NO_TYPES; // shouldn't
-						simpleUpper = right;
-					}
+					rights[i++] = right;
+					nullHints |= right.tagBits & TagBits.AnnotationNullMASK;
 				}
 			}
 			if (i == 0)
-				return simpleUpper != null ? new TypeBinding[] { simpleUpper } : Binding.NO_TYPES;
-			if (i == 1 && simpleUpper != null)
-				return new TypeBinding[] { simpleUpper }; // no nullHints since not a reference type
+				return Binding.NO_TYPES;
 			if (i < rights.length)
 				System.arraycopy(rights, 0, rights=new TypeBinding[i], 0, i);
 			useNullHints(nullHints, rights, variable.environment);
@@ -1287,6 +1278,12 @@ class BoundSet {
 	private void allSuperPairsWithCommonGenericTypeRecursive(TypeBinding s, TypeBinding t, List<Pair<TypeBinding>> result, HashSet<Integer> visited) {
 		if (s == null || s.id == TypeIds.T_JavaLangObject || t == null || t.id == TypeIds.T_JavaLangObject)
 			return;
+		if (s.isArrayType() && t.isArrayType()) {
+			// Common generic supertypes of array types are found through their component types.
+			allSuperPairsWithCommonGenericTypeRecursive(
+					((ArrayBinding) s).elementsType(), ((ArrayBinding) t).elementsType(), result, visited);
+			return;
+		}
 		if (!visited.add(s.id))
 			return;
 
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/BoundSet.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/BoundSet.java
@@ -1279,9 +1279,13 @@ class BoundSet {
 		if (s == null || s.id == TypeIds.T_JavaLangObject || t == null || t.id == TypeIds.T_JavaLangObject)
 			return;
 		if (s.isArrayType() && t.isArrayType()) {
+			ArrayBinding sArray = (ArrayBinding) s;
+			ArrayBinding tArray = (ArrayBinding) t;
+			if (sArray.dimensions() != tArray.dimensions())
+				return;
 			// Common generic supertypes of array types are found through their component types.
 			allSuperPairsWithCommonGenericTypeRecursive(
-					((ArrayBinding) s).elementsType(), ((ArrayBinding) t).elementsType(), result, visited);
+					sArray.elementsType(), tArray.elementsType(), result, visited);
 			return;
 		}
 		if (!visited.add(s.id))
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding18.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding18.java
@@ -40,22 +40,25 @@ public class CaptureBinding18 extends CaptureBinding {
 	}
 
 	public boolean setUpperBounds(TypeBinding[] upperBounds, ReferenceBinding javaLangObject) {
-		this.upperBounds = upperBounds;
-		if (upperBounds.length > 0)
-			this.firstBound = upperBounds[0];
 		int numReferenceInterfaces = 0;
 		if (!isConsistentIntersection(upperBounds, InferenceContext18.SIMULATE_BUG_JDK_8026527))
 			return false;
+		ReferenceBinding newSuperclass = this.superclass;
 		for (TypeBinding aBound : upperBounds) {
 			if (aBound instanceof ReferenceBinding) {
-				if (this.superclass == null && aBound.isClass())
-					this.superclass = (ReferenceBinding) aBound;
+				if (newSuperclass == null && aBound.isClass())
+					newSuperclass = (ReferenceBinding) aBound;
 				else if (aBound.isInterface())
 					numReferenceInterfaces++;
 			} else if (TypeBinding.equalsEquals(aBound.leafComponentType(), this)) {
 				return false; // cycle detected
 			}
 		}
+		// Publish the new state only after all bounds have passed validation.
+		this.upperBounds = upperBounds;
+		if (upperBounds.length > 0)
+			this.firstBound = upperBounds[0];
+		this.superclass = newSuperclass;
 		this.superInterfaces = new ReferenceBinding[numReferenceInterfaces];
 		int idx = 0;
 		for (TypeBinding aBound : upperBounds) {
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding18.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding18.java
@@ -40,25 +40,22 @@ public class CaptureBinding18 extends CaptureBinding {
 	}
 
 	public boolean setUpperBounds(TypeBinding[] upperBounds, ReferenceBinding javaLangObject) {
+		this.upperBounds = upperBounds;
+		if (upperBounds.length > 0)
+			this.firstBound = upperBounds[0];
 		int numReferenceInterfaces = 0;
 		if (!isConsistentIntersection(upperBounds, InferenceContext18.SIMULATE_BUG_JDK_8026527))
 			return false;
-		ReferenceBinding newSuperclass = this.superclass;
 		for (TypeBinding aBound : upperBounds) {
 			if (aBound instanceof ReferenceBinding) {
-				if (newSuperclass == null && aBound.isClass())
-					newSuperclass = (ReferenceBinding) aBound;
+				if (this.superclass == null && aBound.isClass())
+					this.superclass = (ReferenceBinding) aBound;
 				else if (aBound.isInterface())
 					numReferenceInterfaces++;
 			} else if (TypeBinding.equalsEquals(aBound.leafComponentType(), this)) {
 				return false; // cycle detected
 			}
 		}
-		// Publish the new state only after all bounds have passed validation.
-		this.upperBounds = upperBounds;
-		if (upperBounds.length > 0)
-			this.firstBound = upperBounds[0];
-		this.superclass = newSuperclass;
 		this.superInterfaces = new ReferenceBinding[numReferenceInterfaces];
 		int idx = 0;
 		for (TypeBinding aBound : upperBounds) {
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
@@ -2535,6 +2535,16 @@ public static boolean isConsistentIntersection(TypeBinding[] intersectingTypes, 
 		else
 			return false;
 	}
+	if (mostSpecific.isArrayType()) {
+		// Apart from array supertypes, an array subtype implements only Cloneable and Serializable.
+		for (TypeBinding intersectingType : intersectingTypes) {
+			if (intersectingType.isTypeVariable() || intersectingType.isWildcard()
+					|| !intersectingType.isProperType(true))
+				return false;
+			if (!mostSpecific.isCompatibleWith(intersectingType))
+				return false;
+		}
+	}
 	return true;
 }
 public ModuleBinding module() {
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
@@ -2541,7 +2541,7 @@ public static boolean isConsistentIntersection(TypeBinding[] intersectingTypes, 
 			if (intersectingType.isTypeVariable() || intersectingType.isWildcard()
 					|| !intersectingType.isProperType(true))
 				return false;
-			if (!mostSpecific.isCompatibleWith(intersectingType))
+			if (!mostSpecific.isSubtypeOf(intersectingType, simulatingBugJDK8026527))
 				return false;
 		}
 	}
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericsRegressionTest_1_8.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericsRegressionTest_1_8.java
@@ -10964,4 +10964,85 @@ public void testBug508834_comment0() {
 			"The method bar(One<Inner<?>>) in the type Bug is not applicable for the arguments (One<Inner<X>>)\n" +
 			"----------\n");
 	}
+	public void testIssue5204CompatibleArrayUpperBound() {
+		runConformTest(
+			new String[] {
+				"ArrayBoundInference.java",
+				"""
+				import java.util.concurrent.atomic.AtomicReference;
+				import java.util.function.Function;
+				import java.util.function.UnaryOperator;
+
+				class ArrayBoundInference {
+					static <Bound, Selection extends Bound> Selection read(AtomicReference<Bound> reference) {
+						throw new AssertionError();
+					}
+
+					UnaryOperator<String>[] compatible(AtomicReference<Function<?, ?>[]> reference) {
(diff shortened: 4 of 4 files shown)


stephan-herrmann:
Thanks for the PR and the detailed description!

A few questions, still:

> This is a compiler robustness correction for invalid input. Invalid source is routine compiler input and must not crash the compiler.

Can you tell whether the CCE _only_ happened when processing invalid input?

Reason for asking: when we know that a type is not well-formed, we should not let compilation proceed with a valid type binding. 

> The review loop kept the issue scoped to preventing the internal compiler exception. An attempted broader greatest-lower-bound reduction was removed ...

That's a good direction of questions to ask: 
* is glb defined for an array plus a class / interface type?
* is the resulting intersection type well-formed?

JLS 5.1.10, where glb is defined, says:
>  It is a compile-time error if, for any two classes (not interfaces) Vi and Vj, Vi is not a subclass of Vj or vice versa. 

Note that this rule doesn't mention array types. OTOH, there is no class that is a subtype of both `X[]` (for any X) and `Y` (for any class Y other than Object, or interface other than Cloneable, Serializable). So, this might be an omission in 5.1.10?

Hence a change right in `Scope.greaterLowerBound()` might indeed require clarification regarding JLS 5.1.10. 

The next location to check might be well-formedness of the intersection type. JLS 4.9 indeed mentions array types:
>  For each Ti (1 ≤ i ≤ n), let Ci be the most specific class or array type such that Ti <: Ci. Then there must be some Ck such that Ck <: Ci for any i (1 ≤ i ≤ n), or a compile-time error occurs. 

For any pair `Ti` being an interface type or j.l.Object and `Tj` being any array type, `Ck` =  j.l.Object satisfies this rule. This would imply that well-formedness is _not_ violated.

Remains the question of what `CaptureBinding18` should answer in such situations? Perhaps other types below `TypeVariableBinding` are affected, too?

My first guess is: we cannot pretend that the first found array type is indeed the erasure, because it might be incompatible with another intersecting type. So, should we return a `ProblemReferenceBinding` instead? Or should we generalize further and answer j.l.Object as the erasure?

stephan-herrmann:
As a reminder, in case AI was used to generate any of this content, you should be transparent about this, see https://www.eclipse.org/projects/handbook/#genai

martinfrancois:
Thanks. I checked each assignment separately.

In the cases I reproduced, the compiler crash happens while ECJ processes invalid assignments. However, the current patch is not correct: it stops the crash, but it allows two invalid assignments to compile. ECJ also continues to reject two valid array assignments that `javac` accepts.

Therefore, returning the first erasure is not a safe solution.

There is also an error in the PR description. It says that the GLB version broke nested-lambda overload resolution. The same failures still happened after that version was removed. I later found that those failures came from an error in a separate test. I will remove that statement from the description.

The main problem appears to be that `setUpperBounds()` stores the bounds before checking them. It then returns `false`, but `erasure()` can later use those rejected bounds. Returning the first bound makes this invalid state look like a valid type.

I agree that I should keep this state as an error instead. A `ProblemReferenceBinding` looks more suitable than returning the first bound, but I need to check how its callers handle it. I will also test whether returning `Object` is safe and check the related `TypeVariableBinding` implementations.

I will rework the patch and add separate tests for the valid and invalid assignments. The current patch should not be merged as it is.

martinfrancois:
@stephan-herrmann Thanks for the reminder. I did use AI to investigate the issue, review the relevant code, suggest and implement some changes, and help draft the PR description. However, AI assisted only with parts of the work. It did not generate the contribution as a whole. I reviewed the results, changed a lot of them to match what I intended to contribute, and take full responsibility for the contribution.

I had read the main handbook section before submitting the PR. However, this wording:

> For files that are largely or entirely generated by an AI, add a standard copyright header expressing the project licence along with an AI disclosure.

followed by:

> **And** in your commit message:

made me think that both the file-level disclosure **and** the `Assisted-by` footer together were required only when a file was largely or entirely generated by AI.

That did not seem to apply here. Only some parts of the contribution were AI-assisted. Additionally, my code only makes up a fraction of the files, so even if it was fully AI generated, adding a file-level statement saying that such a file was largely AI-generated would therefore have been misleading, so I chose not to add it.

While reading the handbook again for this reply, I searched the page for "AI disclosure" and discovered that there is also a related [FAQ](https://www.eclipse.org/projects/handbook/#genai-faq) entry. I had not seen it before. I would not have expected such an important part of the guidance to appear only in the FAQ without at least being linked from the main GenAI section. Personally, I think it would be clearer to include it directly in the main section, so readers can understand the guidance without having to follow another link.

The FAQ says that not every use of AI needs to be disclosed and gives this rule of thumb:

> If you would give a human credit for the contribution, you should likely provide an AI disclosure.

I still find this unclear. It does not say whether this applies to the `Assisted-by` footer, the file-level disclosure, or both. "Giving a human credit" is also open to interpretation.

For example, I often already know the solution and the code structure I want, and use AI mainly to write the code with less typing, steering it until it matches what I had in mind. Whether a human doing that work would receive credit is unclear. In contrast, if AI independently investigated an issue, designed and implemented the solution, and prepared the PR, a human doing the same work would clearly receive credit.

A clearer rule might be based on whether AI had a meaningful influence on the investigation, design, implementation, tests, documentation, or other important parts of the contribution.

Could you please clarify whether the intended rule is:

- Add an `Assisted-by` footer when AI helped with a contribution in any way.
- Additionally add the file-level disclosure when the file itself was largely or entirely generated by AI.

If so, it may also be useful to state this distinction directly in the main handbook section. I went with this interpretation now and added the `Assisted-by` footer.

martinfrancois:
@stephan-herrmann I also reworked the implementation based on your feedback, can you please have another look and let me know if you agree with the direction I took?

Thanks as well for your detailed questions earlier. To complete the technical points not already covered by my earlier comment or the updated PR description:

> Hence a change right in `Scope.greaterLowerBound()` might indeed require clarification regarding JLS 5.1.10.

I did not change `Scope.greaterLowerBound()`.

The updated code gives the existing GLB calculation the complete set of proper upper bounds. Compatible bounds reduce normally. Bounds that cannot be reduced to one compatible type remain separate and are rejected by the later consistency check.

> For any pair `Ti` being an interface type or `j.l.Object` and `Tj` being any array type, `Ck = j.l.Object` satisfies this rule.

I agree that the existing JLS 4.9 check alone accepts this case. It does not check whether an array can actually satisfy the interface bound.

I added an extra condition for arrays. If the most specific bound is an array, it must be compatible with every other proper bound.

This accepts `Object`, `Cloneable`, `Serializable`, and compatible array types. It rejects an array combined with an unrelated interface such as `List`.

This does not change the meaning of GLB. It rejects an intersection that no array type can satisfy.

> Should we return a `ProblemReferenceBinding` instead? Or should we generalize further and answer `j.l.Object` as the erasure?

Neither option is needed with the revised flow.

Compatible bounds reduce to a valid array type, so returning `Object` would lose that result. Incompatible bounds now make inference fail before the capture state is updated, so creating a `ProblemReferenceBinding` during `erasure()` would happen too late.

I therefore left the existing `erasure()` behavior unchanged and prevented invalid bounds from reaching it.

> Perhaps other types below `TypeVariableBinding` are affected, too?

I checked the relevant subclasses. `CaptureBinding` and `InferenceVariable` do not contain the multi-bound cast used by `CaptureBinding18.erasure()`. `InferenceVariable.erasure()` delegates to the normal `TypeVariableBinding` implementation.

The unsafe multi-bound state is therefore specific to `CaptureBinding18`. The `BoundSet` correction is shared by inference, so inference variables also receive the complete upper-bound set.

I also compared ECJ and `javac` with additional combinations involving arrays, primitive arrays, classes, interfaces, `Object`, `Cloneable`, and `Serializable`. The special array supertypes remain accepted, while unrelated interfaces are rejected.

Please let me know if you agree with this handling of the array intersection case.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericsRegressionTest_1_8.java
```diff
@@ -10964,4 +10964,66 @@ <X, Y extends One<X>> void foo2(Y y1) {
 			"The method bar(One<Inner<?>>) in the type Bug is not applicable for the arguments (One<Inner<X>>)\n" +
 			"----------\n");
 	}
+	public void testIssue5204ValidCaptureArrayErasure() {
```
iloveeclipse: Literal copy from https://github.com/openjdk/jdk/blob/be40b6bcdab37368ea3c769e575b36e290d0c6a1/test/langtools/tools/javac/generics/inference/8062977/T8062977.java

iloveeclipse (dismissed):
See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5204#issuecomment-5020427686.

Please don't literally include GPL'ed code, it is not compatible with our project EPL license.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/GenericsRegressionTest_1_8.java
martinfrancois: Thanks for the note, I'll make that change.

stephan-herrmann:
> [...]
> > **And** in your commit message:
> 
> made me think that both the file-level disclosure **and** the `Assisted-by` footer together were required only when a file was largely or entirely generated by AI.
> 
> That did not seem to apply here. Only some parts of the contribution were AI-assisted.

Given that this is a "rapidly evolving area" and given the goal of transparency, I'd rather add one "Assisted-by" too many rather than keeping silent. If two humans co-operate on a fix, an "Also-by" line also doesn't indicate that that person has contributed an entirely new file.
   
 
> While reading the handbook again for this reply, I searched the page for "AI disclosure" and discovered that there is also a related [FAQ](https://www.eclipse.org/projects/handbook/#genai-faq) entry. I had not seen it before. I would not have expected such an important part of the guidance to appear only in the FAQ without at least being linked from the main GenAI section. Personally, I think it would be clearer to include it directly in the main section, so readers can understand the guidance without having to follow another link.
> ...

The handbook is managed by EMO, so either an email or a ticket at https://gitlab.eclipse.org/eclipsefdn/emo-team/emo might help them to improve.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/BoundSet.java
```diff
@@ -1287,6 +1278,12 @@ protected List<Pair<TypeBinding>> allSuperPairsWithCommonGenericType(TypeBinding
 	private void allSuperPairsWithCommonGenericTypeRecursive(TypeBinding s, TypeBinding t, List<Pair<TypeBinding>> result, HashSet<Integer> visited) {
 		if (s == null || s.id == TypeIds.T_JavaLangObject || t == null || t.id == TypeIds.T_JavaLangObject)
 			return;
+		if (s.isArrayType() && t.isArrayType()) {
+			// Common generic supertypes of array types are found through their component types.
+			allSuperPairsWithCommonGenericTypeRecursive(
+					((ArrayBinding) s).elementsType(), ((ArrayBinding) t).elementsType(), result, visited);
```
stephan-herrmann: I believe we should first check array dimensions. We don't want `Integer[][]` to be seen as compatible with `Number[]` or similar.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
```diff
@@ -2535,6 +2535,16 @@ else if (current.isSubtypeOf(mostSpecific, simulatingBugJDK8026527))
 		else
 			return false;
 	}
+	if (mostSpecific.isArrayType()) {
+		// Apart from array supertypes, an array subtype implements only Cloneable and Serializable.
+		for (TypeBinding intersectingType : intersectingTypes) {
+			if (intersectingType.isTypeVariable() || intersectingType.isWildcard()
+					|| !intersectingType.isProperType(true))
+				return false;
+			if (!mostSpecific.isCompatibleWith(intersectingType))
```
stephan-herrmann: I believe this should use `isSubtypeOf()` like above, rather than `isCompatibleWith()`.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding18.java
```diff
@@ -40,22 +40,25 @@ private CaptureBinding18(CaptureBinding18 prototype) {
...
 					numReferenceInterfaces++;
 			} else if (TypeBinding.equalsEquals(aBound.leafComponentType(), this)) {
 				return false; // cycle detected
 			}
 		}
+		// Publish the new state only after all bounds have passed validation.
+		this.upperBounds = upperBounds;
+		if (upperBounds.length > 0)
+			this.firstBound = upperBounds[0];
+		this.superclass = newSuperclass;
```
stephan-herrmann: I couldn't find a test that needs this group of changes. Can you demonstrate the need with a new test cases, or is this a left-over from previous experiments?

stephan-herrmann (changes requested):
The general strategy looks good to me. Please just attend to the detail comments.

stephan-herrmann:
As for updating the branch: please use rebase only for updating to latest HEAD. I'd like to see any additional changes as a separate commit (no force push here, please), so I can trace how the fix evolves, thanks.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/BoundSet.java
martinfrancois: Fixed in `d76f3270bf93d5012ed3da779460d95387807901`. The code now compares the number of array dimensions first. If the numbers differ, it stops. If they are equal, it continues by comparing the component types.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/ReferenceBinding.java
martinfrancois: Fixed in `aaf7c6e75feccd731374d88a516bf8814d9d75fa`. The array intersection check now uses `isSubtypeOf(...)` instead of `isCompatibleWith(...)`. It also uses the same compatibility flag as the checks above it.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding18.java
martinfrancois: Fixed in `5e0fe8f4166831007373d9582c4470e798474d6b`. I did not find a test that needed the delayed `CaptureBinding18` state update. I therefore removed those changes and restored the original update order.

I tested this version with the JDT regression tests and with Spoon with applied fixes. I also ran strict StoneDetector scans on the Elasticsearch JDK 25 and OpenJDK reference corpora. All 30,894 selected Elasticsearch files and all 52,321 selected OpenJDK files completed successfully. The scans produced no errors and required no retries.

martinfrancois:
@stephan-herrmann Thank you for the review and for clarifying the AI disclosure guidance. I kept the `Assisted-by` footer in each new commit.

I addressed each detail comment in a separate commit and pushed them without rebasing or force-pushing. I also replied to each review thread with more details and the related test results.

Could you please review the changes again?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding18.java
stephan-herrmann: > I also ran strict StoneDetector scans ...

out of curiosity, as I'm not familiar with StoneDetector: what kind of scanning / testing does it provide that is relevant here?

martinfrancois: @stephan-herrmann [StoneDetector](https://github.com/StoneDetector/StoneDetector) is an open source tool for finding duplicated code. It uses [Spoon](https://github.com/INRIA/spoon), which uses Eclipse JDT to read and analyze Java code.

I found this issue while working on [my PR to update StoneDetector for JDK 25](https://github.com/StoneDetector/StoneDetector/pull/1). I ran it on Elasticsearch and OpenJDK's codebases to cover many different real-world Java files and to ensure I cover all new language features. These scans found bugs in both Spoon and JDT, including this one.

To check that all the fixes work together, I temporarily cherry-picked all my JDT fixes into a local branch and published a snapshot to my local Maven repository. I then built Spoon with that JDT snapshot and all my Spoon fixes, and published a local Spoon snapshot. Finally, I used that Spoon snapshot as the version in my StoneDetector PR and ran the scans again. This confirmed that the changes in this PR still fix all the problems I found. I already did this before I submitted all the PRs for the fixes for JDT and Spoon and it worked - so it is kind of a final regression test for confirmation 😉

stephan-herrmann: Cool, thanks for explaining.

stephan-herrmann (approved):
Thanks for addressing my review comments. 

~No we only need one last step before merging: please rebase your changes on HEAD of master - resolving conflicts if any.~

Scratch that. Not sure why github showed me the this branch needs an update. Looks good, actually.

martinfrancois:
No worries @stephan-herrmann thanks for the review and the merge!

PR #5207, opened 2026-07-17
Fix nested lambda inference memory growth #5206

### LINKED ISSUE #5206: Nested lambda overload inference exhausts the heap

## Steps to reproduce

Use these two valid compiler stress tests from `openjdk/jdk@be40b6bcdab37368ea3c769e575b36e290d0c6a1`:

- [NestedLambdaGenerics.java](https://github.com/openjdk/jdk/blob/be40b6bcdab37368ea3c769e575b36e290d0c6a1/test/langtools/tools/javac/lambda/speculative/NestedLambdaGenerics.java)
- [NestedLambdaNoGenerics.java](https://github.com/openjdk/jdk/blob/be40b6bcdab37368ea3c769e575b36e290d0c6a1/test/langtools/tools/javac/lambda/speculative/NestedLambdaNoGenerics.java)

Build the batch compiler from the tested JDT checkout. Compile each source in a separate process:

```text
java -Xms64m -Xmx1g -XX:+ExitOnOutOfMemoryError \
  -jar ecj.jar -25 -d classes NestedLambdaGenerics.java
```

Repeat the command with `NestedLambdaNoGenerics.java`.

I expected: both finite positive tests compile successfully within the configured heap.

But got: the generic source runs out of memory while copying inference bounds. The non-generic source runs out of memory while repeatedly copying and resolving nested lambdas.

Each source contains 31 nested method calls and 30 parameterless lambdas. Each call has six overload candidates. During speculative overload resolution, ECJ tests each candidate without committing to one and reparses the outer lambda for that candidate.

Each source lambda stores inference results for the target functional interface types already tested against it. The original node from which a speculative copy was made is its canonical source lambda. The outer copy keeps that source link, but nested lambdas inside the copy are treated as new originals. They do not reuse their source lambda's existing results, so every candidate repeats inference at every nested level and memory use grows combinatorially.

Nested parameterless copies must keep their canonical source link and reuse only the existing per-target inference cache. Parameterized lambda copies must keep their context-local parameter bindings. The correction must not share mutable parser state or add a global cache.

Generic source:

```text
java.lang.OutOfMemoryError: Java heap space
    at BoundSet$ThreeSets.copy(...)
    at BoundSet.copy(...)
    at InferenceContext18.computeB3(...)
```

Non-generic source:

```text
java.lang.OutOfMemoryError: Java heap space
    at Parser.<init>(...)
    at LambdaExpression.copy(...)
    at LambdaExpression.cachedResolvedCopy(...)
```

- Heap limit: 1 GiB per compiler process
- Existing-issue search: no JDT issue matching either filename or OpenJDK bug 8078093 was found

## Community

martinfrancois:
## What it does

Fixes #5206.

During _speculative overload resolution_, ECJ tries each overload candidate without committing to one. It reparses the outer lambda to create a separate copy for that candidate. Each source lambda stores inference results for the target functional-interface types against which ECJ has already tested it. The lambda from which a copy was made is its _original lambda_.

The outer speculative copy kept that source link, but nested lambdas inside the copy were treated as new originals. They therefore did not reuse the existing per-target results from their original lambdas. Repeating inference for every candidate at every nested level caused combinatorial memory growth and exhausted the heap.

The change links nested speculative copies to their original lambdas. On each branch, it shares the per-target inference cache down to and including the first lambda with parameters. A cache entry is a resolved copy of that lambda and owns its parameter bindings. Nested lambdas may use those bindings, so traversal stops below it and their caches stay local to the enclosing copy. No global cache or shared parser state is introduced.

The source lambda and the copy created by parsing the source again are traversed in the same source order. Each traversal records the current lambda, then stops below it when it has parameters. `shareInferenceCaches()` therefore receives only lambdas allowed by this parameter-binding rule, so no separate eligibility flag is needed.

The independently designed regression suite has three tests. Two generate a generic and a concrete source. Each source contains 24 nested method calls and 23 parameterless lambdas, alternates between `route` and `select`, provides five overload candidates for each selector, and ends at a `Work<String>` field. The two variants exercise the original missing cache reuse with different inference pressure.

The third test compiles and runs code that mixes lambdas with and without parameters. It includes a nested lambda that reads an enclosing parameter and a separate parameterless sibling. There is no public API change.

The linked [NestedLambdaGenerics.java](https://github.com/openjdk/jdk/blob/be40b6bcdab37368ea3c769e575b36e290d0c6a1/test/langtools/tools/javac/lambda/speculative/NestedLambdaGenerics.java) and [NestedLambdaNoGenerics.java](https://github.com/openjdk/jdk/blob/be40b6bcdab37368ea3c769e575b36e290d0c6a1/test/langtools/tools/javac/lambda/speculative/NestedLambdaNoGenerics.java) files provide separate reference validation for the reported behavior; they are not copied into the regression suite.

## How to test

- On unpatched JDT commit `792a9156c119d28063904c85726e486ec2fb2206`, compiling each linked OpenJDK source separately with `-25` and a 1 GiB heap reproduces #5206.
- With the compiler from this PR, the same commands exit 0 and generate six class files each, with no `OutOfMemoryError` or internal compiler error.
- Run `NestedLambdaInferenceTest` across Java 8 through Java 25 compliance levels: 54 tests pass.
- Run `GenericsRegressionTest_1_8` together with `NestedLambdaInferenceTest`: 5,670 tests pass, including the parameterized-lambda regression `testBug496578`.
- Run the full compiler suite (`parser/TestAll`, `regression/TestAll`, and `eval/TestAll`): 263,722 tests pass with no failures, errors, or skipped tests.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
@@ -47,6 +47,7 @@ package org.eclipse.jdt.internal.compiler.ast;
 
 import static org.eclipse.jdt.internal.compiler.ast.ExpressionContext.INVOCATION_CONTEXT;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
@@ -1116,12 +1117,64 @@ public class LambdaExpression extends FunctionalExpression implements IPolyExpre
 		if (copy != null) { // ==> syntax errors == null
 			if (copy.sourceStart != this.sourceStart || copy.sourceEnd != this.sourceEnd)
 				return null; // something wrong
-			copy.original = this;
+			shareInferenceCaches(copy);
 			copy.assistNode = this.assistNode;
 			copy.enclosingScope = this.enclosingScope;
 			copy.text = this.text; // discard redundant textual copy
 		}
 		return copy;
+	}
+
+	private void shareInferenceCaches(LambdaExpression copy) {
+		// A speculative copy can contain nested lambdas. Keep each nested copy linked to its
+		// source lambda when no enclosing lambda declares parameters, so overload checks can
+		// reuse the existing per-target inference cache without reusing parameter bindings.
+		List<CollectedLambda> sourceLambdas = collectLambdas(this);
+		List<CollectedLambda> copiedLambdas = collectLambdas(copy);
+		if (sourceLambdas.size() != copiedLambdas.size())
+			throw new CopyFailureException();
+		for (int i = 0; i < sourceLambdas.size(); i++) {
+			CollectedLambda source = sourceLambdas.get(i);
+			LambdaExpression sourceLambda = source.lambda();
+			LambdaExpression copiedLambda = copiedLambdas.get(i).lambda();
+			if (sourceLambda.sourceStart != copiedLambda.sourceStart || sourceLambda.sourceEnd != copiedLambda.sourceEnd)
+				throw new CopyFailureException();
+			if (!source.cacheShareable()) {
+				if (i == 0)
+					copiedLambda.original = this;
+				continue;
+			}
+			LambdaExpression originalLambda = sourceLambda.original;
+			copiedLambda.original = originalLambda;
+			if (originalLambda.copiesPerTargetType == null)
+				originalLambda.copiesPerTargetType = new HashMap<>();
+			sourceLambda.copiesPerTargetType = originalLambda.copiesPerTargetType;
+			copiedLambda.copiesPerTargetType = originalLambda.copiesPerTargetType;
+		}
+	}
+
+	private record CollectedLambda(LambdaExpression lambda, boolean cacheShareable) {}
+
+	private static List<CollectedLambda> collectLambdas(LambdaExpression root) {
+		List<CollectedLambda> lambdas = new ArrayList<>();
+		root.traverse(new ASTVisitor() {
+			private int parameterizedLambdaDepth;
+
+			@Override
+			public boolean visit(LambdaExpression lambda, BlockScope skope) {
+				if (lambda.arguments.length > 0)
+					this.parameterizedLambdaDepth++;
+				lambdas.add(new CollectedLambda(lambda, this.parameterizedLambdaDepth == 0));
+				return true;
+			}
+
+			@Override
+			public void endVisit(LambdaExpression lambda, BlockScope skope) {
+				if (lambda.arguments.length > 0)
+					this.parameterizedLambdaDepth--;
+			}
+		}, root.enclosingScope);
+		return lambdas;
 	}
 
 	public void returnsExpression(Expression expression, TypeBinding resultType) {
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
@@ -986,9 +986,7 @@ public class LambdaExpression extends FunctionalExpression implements IPolyExpre
 					return null;
 
 				targetType = copy.expectedType; // possibly updated local types
-				if (this.copiesPerTargetType == null)
-					this.copiesPerTargetType = new HashMap<>();
-				this.copiesPerTargetType.put(targetType, copy);
+				this.copiesPerTargetType.put(targetType, copy); // copy() has linked this lambda to the original's cache
 			}
 			if (!requireExceptionAnalysis)
 				return copy;
@@ -1126,28 +1124,22 @@ public class LambdaExpression extends FunctionalExpression implements IPolyExpre
 	}
 
 	private void shareInferenceCaches(LambdaExpression copy) {
-		// A speculative copy can contain nested lambdas. Keep each nested copy linked to its
-		// source lambda when no enclosing lambda declares parameters, so overload checks can
-		// reuse the existing per-target inference cache without reusing parameter bindings.
-		// Both traversals start with their root and then visit nested lambdas in source order.
-		List<CollectedLambda> sourceLambdas = collectLambdas(this);
-		List<CollectedLambda> copiedLambdas = collectLambdas(copy);
+		// A speculative copy can contain nested lambdas. Link each collected lambda to its
+		// original lambda and share the per-target inference cache.
+		// A cache entry is a resolved lambda copy and owns its parameter bindings. Nested
+		// lambdas may use parameters from an enclosing lambda, so collectLambdas() stops
+		// below a lambda with parameters. Caches below that point stay local to the
+		// enclosing copy.
+		// Both traversals start at the root and visit nested lambdas in source order.
+		List<LambdaExpression> sourceLambdas = collectLambdas(this);
+		List<LambdaExpression> copiedLambdas = collectLambdas(copy);
 		if (sourceLambdas.size() != copiedLambdas.size())
 			throw new CopyFailureException();
 		for (int i = 0; i < sourceLambdas.size(); i++) {
-			CollectedLambda source = sourceLambdas.get(i);
-			LambdaExpression sourceLambda = source.lambda();
-			LambdaExpression copiedLambda = copiedLambdas.get(i).lambda();
+			LambdaExpression sourceLambda = sourceLambdas.get(i);
+			LambdaExpression copiedLambda = copiedLambdas.get(i);
 			if (sourceLambda.sourceStart != copiedLambda.sourceStart || sourceLambda.sourceEnd != copiedLambda.sourceEnd)
 				throw new CopyFailureException();
-			if (!source.cacheShareable()) {
-				// The root is the copy requested by the caller. Nested lambdas below a
-				// parameterized lambda are local sources for copies that use the enclosing
-				// copy's parameter bindings, so their inference caches stay in that context.
-				if (sourceLambda == this)
-					copiedLambda.original = this;
-				continue;
-			}
 			LambdaExpression originalLambda = sourceLambda.original;
 			copiedLambda.original = originalLambda;
 			if (originalLambda.copiesPerTargetType == null)
@@ -1157,25 +1149,13 @@ public class LambdaExpression extends FunctionalExpression implements IPolyExpre
 		}
 	}
 
-	private record CollectedLambda(LambdaExpression lambda, boolean cacheShareable) {}
-
-	private static List<CollectedLambda> collectLambdas(LambdaExpression root) {
-		List<CollectedLambda> lambdas = new ArrayList<>();
+	private static List<LambdaExpression> collectLambdas(LambdaExpression root) {
+		List<LambdaExpression> lambdas = new ArrayList<>();
 		root.traverse(new ASTVisitor() {
-			private int parameterizedLambdaDepth;
-
 			@Override
 			public boolean visit(LambdaExpression lambda, BlockScope skope) {
-				if (lambda.arguments.length > 0)
-					this.parameterizedLambdaDepth++;
-				lambdas.add(new CollectedLambda(lambda, this.parameterizedLambdaDepth == 0));
-				return true;
-			}
-
-			@Override
-			public void endVisit(LambdaExpression lambda, BlockScope skope) {
-				if (lambda.arguments.length > 0)
-					this.parameterizedLambdaDepth--;
+				lambdas.add(lambda);
+				return lambda.arguments.length == 0; // nested lambdas may use these parameters
 			}
 		}, root.enclosingScope);
 		return lambdas;
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
@@ -1129,6 +1129,7 @@ public class LambdaExpression extends FunctionalExpression implements IPolyExpre
 		// A speculative copy can contain nested lambdas. Keep each nested copy linked to its
 		// source lambda when no enclosing lambda declares parameters, so overload checks can
 		// reuse the existing per-target inference cache without reusing parameter bindings.
+		// Both traversals start with their root and then visit nested lambdas in source order.
 		List<CollectedLambda> sourceLambdas = collectLambdas(this);
 		List<CollectedLambda> copiedLambdas = collectLambdas(copy);
 		if (sourceLambdas.size() != copiedLambdas.size())
@@ -1140,7 +1141,10 @@ public class LambdaExpression extends FunctionalExpression implements IPolyExpre
 			if (sourceLambda.sourceStart != copiedLambda.sourceStart || sourceLambda.sourceEnd != copiedLambda.sourceEnd)
 				throw new CopyFailureException();
 			if (!source.cacheShareable()) {
-				if (i == 0)
+				// The root is the copy requested by the caller. Nested lambdas below a
+				// parameterized lambda are local sources for copies that use the enclosing
+				// copy's parameter bindings, so their inference caches stay in that context.
+				if (sourceLambda == this)
 					copiedLambda.original = this;
 				continue;
 			}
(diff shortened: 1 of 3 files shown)


srikanth-sankaran:
@stephan-herrmann - given this involves inference, I am assuming you will want to be the first reviewer, I have also signed up to review FYI

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
```diff
@@ -1116,14 +1117,66 @@ LambdaExpression copy() {
...
+		for (int i = 0; i < sourceLambdas.size(); i++) {
+			CollectedLambda source = sourceLambdas.get(i);
+			LambdaExpression sourceLambda = source.lambda();
+			LambdaExpression copiedLambda = copiedLambdas.get(i).lambda();
+			if (sourceLambda.sourceStart != copiedLambda.sourceStart || sourceLambda.sourceEnd != copiedLambda.sourceEnd)
+				throw new CopyFailureException();
+			if (!source.cacheShareable()) {
+				if (i == 0)
+					copiedLambda.original = this;
+				continue;
```
stephan-herrmann: When we reach here with `i>0` then `copiedLambda` will be left with `copiedLambda.original == copiedLambda`, which seems to violate a contract of copies.

If there is deeper reason behind this, I'd appreciate further explanation.

Furthermore, it assumes meaning of how the lists `sourcesLambdas` / `copiedLambdas` are arranged, which, too, deserves an explanation (what makes `i==0` special?)

stephan-herrmann (commented):
Some questions after a first scan of changes:
* could we have some tests that demonstrate the interleaving of lambdas with and without parameters? There is some complexity in this, which I'd like see live in action.
* would it make sense to add a shortcut such that "normal" scenarios could continue to work the "old" way?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
martinfrancois: Thanks @stephan-herrmann. The root lambda is special because it is the copy requested by the caller. Both lists visit that root first and then visit nested lambdas in the same source order.

A nested lambda below a lambda with parameters stays its own `original` on purpose. Its later copies use parameter bindings from the copied outer lambda, so they must use a cache from that local context.

I replaced the `i == 0` check with `sourceLambda == this`, added comments explaining both points, and added a test that mixes lambdas with and without parameters.

Addressed in f1083f93c1.

stephan-herrmann: > The root lambda is special because it is the copy requested by the caller.

Makes sense.
So the contract `copy.original != copy` holds only for a lambda _currently being resolved_ against a specific target type, which is the copy of `this`.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
```diff
@@ -1116,14 +1117,70 @@ LambdaExpression copy() {
...
+		List<CollectedLambda> copiedLambdas = collectLambdas(copy);
+		if (sourceLambdas.size() != copiedLambdas.size())
+			throw new CopyFailureException();
+		for (int i = 0; i < sourceLambdas.size(); i++) {
+			CollectedLambda source = sourceLambdas.get(i);
+			LambdaExpression sourceLambda = source.lambda();
+			LambdaExpression copiedLambda = copiedLambdas.get(i).lambda();
+			if (sourceLambda.sourceStart != copiedLambda.sourceStart || sourceLambda.sourceEnd != copiedLambda.sourceEnd)
+				throw new CopyFailureException();
+			if (!source.cacheShareable()) {
```
stephan-herrmann: At first I read this method name as a command to cache anything that is shareable.

Perhaps `isCacheSharable` is a clearer name for this record component?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
```diff
@@ -1116,14 +1117,70 @@ LambdaExpression copy() {
...
+	private static List<CollectedLambda> collectLambdas(LambdaExpression root) {
+		List<CollectedLambda> lambdas = new ArrayList<>();
+		root.traverse(new ASTVisitor() {
+			private int parameterizedLambdaDepth;
+
+			@Override
+			public boolean visit(LambdaExpression lambda, BlockScope skope) {
+				if (lambda.arguments.length > 0)
+					this.parameterizedLambdaDepth++;
+				lambdas.add(new CollectedLambda(lambda, this.parameterizedLambdaDepth == 0));
```
stephan-herrmann: This makes every lambda with parameters not-sharing. Moving the conditional increment below the addition to `lambdas`  doesn't seem to break any tests.

So, what is the exact condition preventing caching? Having parameters, or: having parameters of an outer lambda in scope?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
```diff
@@ -1116,14 +1117,70 @@ LambdaExpression copy() {
...
+				// parameterized lambda are local sources for copies that use the enclosing
+				// copy's parameter bindings, so their inference caches stay in that context.
+				if (sourceLambda == this)
+					copiedLambda.original = this;
+				continue;
+			}
+			LambdaExpression originalLambda = sourceLambda.original;
+			copiedLambda.original = originalLambda;
+			if (originalLambda.copiesPerTargetType == null)
+				originalLambda.copiesPerTargetType = new HashMap<>();
```
stephan-herrmann, line 1146: Now the responsibility to instantiate that `HashMap` is split between `cachedResolvedCopy()` and this current location. 

Would it make sense to assign the sole responsibility to the current method, so `cachedResolvedCopy()` can safely assume an existing map instance?

Given that linking `copiesPerTargetType` is the main purpose of this method, maybe it should take full responsibility?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
stephan-herrmann: Another way to avoid messing with non-sharable copies might be to just prune the traversal when a lambda has parameters (by returning `false` after handling the current lambda). That way `shareInferenceCaches()` wouldn't even see any non-sharing lambdas. WDYT? Wouldn't this allow you to simplify the code some more?

stephan-herrmann (commented):
I'm starting to understand the strategy 😄 

I suggest to start the next round by looking at my comment in method `collectLambdas()`. If my idea there is sound, then some of the other comments might become obsolete.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
martinfrancois: Agreed. The pruning change removes `CollectedLambda`, so there is no component left to rename. `collectLambdas()` now returns plain `LambdaExpression` values.

Addressed in b25d5f0f94.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
martinfrancois: A lambda can have parameters and still share its own cache. When ECJ creates a cache entry, it resolves a new copy of that lambda. The cache entry owns the bindings for its parameters.

The problem starts below that lambda. A nested lambda may use a parameter from the enclosing lambda. Its resolved references then belong to one specific copy of the enclosing lambda, so its cache must stay with that copy.

For example:

```java
left -> produce(() -> left)
```

`left -> ...` can share its cache. The cache entry owns its binding for `left`. The inner `() -> left` must stay local because it uses the `left` binding from one particular enclosing copy.

I changed the visitor as you suggested:

```java
lambdas.add(lambda);
return lambda.arguments.length == 0;
```

It records the current lambda and then stops before visiting its children when that lambda has parameters. `shareInferenceCaches()` therefore receives the lambda with parameters, but not any lambda inside it. This removes `CollectedLambda`, the depth counter, `endVisit()`, and the skip branch in the linking loop.

This is conservative: it stops below every lambda with parameters, even when its children do not use those parameters. This may share fewer caches than necessary, but it avoids sharing references to bindings from another enclosing copy.

The source lambda and its reparsed copy must contain the same collected lambdas in the same source order. The size check detects a different number of collected lambdas. The source-range check detects a different source position for a pair. After linking, every collected source/copy pair shares the original lambda's cache. Pruned descendants are not linked and keep their local caches.

The interleaving test also contains a parameterless sibling after the parameterized branch, so this structure is covered.

Addressed in b25d5f0f94.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/LambdaExpression.java
martinfrancois: Agreed, done. `shareInferenceCaches()` now creates the map, and I removed the fallback from `cachedResolvedCopy()`.

`put()` is reached only after `copy()` returned a non-null value and the copy resolved successfully. A non-null result from `copy()` means that `shareInferenceCaches()` completed. That method always collects the root and assigns the original lambda's map, so `this.copiesPerTargetType` is non-null before `put()` is reached.

If parsing or cache linking fails, control does not reach `put()`.

Addressed in b25d5f0f94.

martinfrancois:
Thanks, your pruning idea works and made the code much simpler.

`collectLambdas()` now records the current lambda and stops before visiting its children when that lambda has parameters. This removes `CollectedLambda`, the depth counter, `endVisit()`, and the skip branch in `shareInferenceCaches()`. `shareInferenceCaches()` now also creates `copiesPerTargetType` in one place.

I added the details in the inline replies.

After this change, all 54 `NestedLambdaInferenceTest` tests passed. The full compiler suite (`parser/TestAll`, `regression/TestAll`, and `eval/TestAll`) also passed 263,722 tests, with no failures, errors, or skipped tests. The two linked OpenJDK sources also compiled successfully with the same 1 GiB commands used to reproduce #5206. Each command exited 0 and generated six class files.

I also updated the pull request description to match the new approach.

stephan-herrmann:
> Thanks, your pruning idea works and made the code much simpler.

Thanks for making these changes. The PR now looks good to me.

@srikanth-sankaran do you want review this, too?

srikanth-sankaran:
> Thanks for making these changes. The PR now looks good to me.
> 
> @srikanth-sankaran do you want review this, too?

I'll go with your review, but will request @jarthana to stress test these changes - Jay thanks for your help here. This looks good to go for M3.

jarthana:
> > Thanks for making these changes. The PR now looks good to me.
> > @srikanth-sankaran do you want review this, too?
> 
> I'll go with your review, but will request @jarthana to stress test these changes - Jay thanks for your help here. This looks good to go for M3.

Good to go from my side.

srikanth-sankaran:
> No global cache

Curious: Why not ?

srikanth-sankaran:
> A cache entry is a resolved copy of that lambda and owns its parameter bindings. Nested lambdas may use those bindings, so traversal stops below it

[...]

> Each traversal records the current lambda, then stops below it when it has parameters.

Can you give a small test case that shows why this is necessary ? What breaks otherwise ? Thanks

stephan-herrmann:
> > A cache entry is a resolved copy of that lambda and owns its parameter bindings. Nested lambdas may use those bindings, so traversal stops below it
> 
> [...]
> 
> > Each traversal records the current lambda, then stops below it when it has parameters.
> 
> Can you give a small test case that shows why this is necessary ? What breaks otherwise ? Thanks

This is how I understood the strategy:

Given as a minimal example:
```
foo(a -> () -> a);
```
If `foo()` is overloaded, we may do different inference attempts, one per overload of `foo()`. Each of these attempts will create its own copy of the outer lambda, and in that lambda copy the parameter `a` gets a unique binding. If during both attempts the inner lambda is resolved against the same target type, then we might be tempted to share a resolved copy for both attempts. But then the inner reference `a` has already been resolved to the parameter from the outer lambda, i.e., also one of the `a` bindings will be shared during both attempts.

I see two levels to this:

(1) It just feels wrong if we create branching universes of lambda copies, but then share inner copies, such that a name reference is resolved as pointing into the other universe. But perhaps this is harmless, as long as the final resolution (after inference is done) performs its own, consistent resolving.

(2) This may theoretically break when different attempts resolve `a` as having different types, as this type may influence the inference of the inner lambda.  If we take the wrong type for `a` we may get a wrong result for the inner lambda, which may then bubble up again also to the outer lambda etc. Is it possible to create such a test? I could not come up with any.

So for me it is (1) that tells me to better avoid such inter-galactic references :)

martinfrancois:
Thanks, Stephan. Using your example, `foo(a -> () -> a)`, each attempt to resolve the outer lambda creates a separate binding for the parameter `a`. The inner lambda refers to the `a` binding from that specific outer lambda. Sharing the inner lambda between attempts could make it refer to the binding from the wrong attempt.

I could not find a valid example where this causes incorrect output. Stopping cache sharing below a lambda with parameters is still the safer choice.

A global cache would have the same problem. The target type is not enough to identify a safe cached result. The result also depends on the source lambda and its surrounding parameter bindings. Keeping the cache on the original lambda avoids sharing resolved AST state between unrelated attempts.

srikanth-sankaran:
> A global cache would have the same problem. 

I would like to see what breaks, if for nothing else other than to gain a better understanding. Of course, nothing breaking is no proof of correctness - so we may not want to ship such. 

I'll have this in my radar - famous last words 😆

stephan-herrmann:
@srikanth-sankaran from my p.o.v. there is no strong need to understand what would happen if we'd optimize / share more aggressively. For me its more important to be sure that the optimization we apply doesn't break anything. 

Right when I was about to propose to stop here, I came to think of a tricky case we haven't yet considered:
```java
m(() -> {
    var x = n();
    o(() -> p(x));
});
```
The point here is: the outer lambda has no parameters, so we feel safe to share resolved copies of the inner lambda. But again a name, `n`, resolves to a thing from an enclosing scope. Again sharing a resolved inner will cause a reference from one universe into another.

To some degree we can still rely on final resolution to set things straight, provided that inference did "the right thing".

Question: is there any way that different overloads of `m()` cause `x` to have different types? In the given example this shouldn't happen, because the interpretation of `n()` does not depend on the selected signature of `m()`.

From this I'd argue: yes, a lambda parameter is the only way how outer overload resolution could influence inner inference between levels of nesting. If s.o. proves this to be wrong, then we'd need to improve the change. Otherwise I'd still vote for keeping the status reached and be happy.

martinfrancois:
Thanks, @stephan-herrmann. I agree with your reasoning.

In your example:

```java
m(() -> {
    var x = n();
    o(() -> p(x));
});
```

ECJ creates temporary copies of the outer lambda while checking the overloaded `m()` methods. Each temporary copy creates a separate compiler binding for `x`.

If ECJ reuses a cached inner lambda from an earlier overload check, that cached lambda still contains the `x` binding from the earlier temporary copy. This only affects the temporary compiler data used during inference. Final resolution creates the correct bindings used for code generation, as also discussed in [issue #5206](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5206#issuecomment-5272571841).

Checking each overloaded `m()` method cannot change the type of `x`. Java gets the type of a `var` local from its initializer without using the surrounding lambda type. In this example, `n()` determines the type of `x`. This is specified in [JLS 14.4.1](https://docs.oracle.com/javase/specs/jls/se25/html/jls-14.html#jls-14.4.1).

Lambda parameters are different:

```java
m(a -> o(() -> p(a)));
```

Here, each overloaded `m()` method can provide a different type for `a`. For example, one method could provide `String`, while another could provide `Integer`. This is described in [JLS 15.27.3](https://docs.oracle.com/javase/specs/jls/se25/html/jls-15.html#jls-15.27.3).

I also tried to find a counterexample. For this, I built two experimental compiler versions for comparison. One disabled cache sharing for nested lambdas. The other continued sharing inside lambdas with parameters.

I compared whether the test sources were accepted or rejected and checked their errors and warnings. I also checked relevant OpenJDK lambda and inference tests.

For the small test programs written specifically for this investigation, I tested captured `var` locals, local and anonymous classes, pattern variables, generic inference, checked exceptions, and nested method references. I also compared the generated class files and runtime results. I found no differences.

This does not prove that the more aggressive sharing is safe, but I did not find a counterexample. It gives me more confidence that lambda parameters are the important boundary.

I would keep the current implementation. I would also keep the cache on the original lambda instead of making it global, because a cached result belongs to a specific source lambda and its surrounding bindings.

If someone later finds another way for the outer lambda type to change an enclosing binding, we can handle it in a separate follow-up.

PR #5221, opened 2026-07-24
Fix compatible multi-release JAR lookup

### LINKED ISSUE #380: Plexus-compiler-eclipse fails to resolve module dependency
## Description
I ran into an issue where the the plexus-compiler-eclipse 2.15.0 fails to resolve a compile-scoped dependency that is correctly referenced by the module name listed in its module-info.

I have created a minimal Maven project to reproduce the issue that will also show that Javac is able to compile correctly whereas the Plexus-compiler-eclipse is not. Note that Eclipse 2024-03 IDE has no issue resolving the module, so I suspect the issue is with Plexus compiler and not with ECJ itself.

## Reproduction of the issue
Please clone [this repo](https://github.com/xazap/module-path-issue-plexus-compiler) and reproduce the issue with `mvn clean compile -f pom-using-ecj.xml`:
```
[INFO] --- compiler:3.12.1:compile (default-compile) @ test ---
[INFO] Recompiling the module because of changed source code.
[WARNING] File encoding has not been set, using platform encoding UTF-8, i.e. build is platform dependent!
[INFO] Compiling with eclipse [debug release 17 module-path] to target/classes
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /home/xazap/Projects/plexus-bug/src/main/java/module-info.java:[3,22] org.slf4j cannot be resolved to a module
[ERROR] /home/xazap/Projects/plexus-bug/src/main/java/test/Test.java:[3,8] The type org.slf4j.Logger is not accessible
[ERROR] /home/xazap/Projects/plexus-bug/src/main/java/test/Test.java:[4,8] The type org.slf4j.LoggerFactory is not accessible
[ERROR] /home/xazap/Projects/plexus-bug/src/main/java/test/Test.java:[8,23] Logger cannot be resolved to a type
[ERROR] /home/xazap/Projects/plexus-bug/src/main/java/test/Test.java:[8,39] LoggerFactory cannot be resolved
[ERROR] /home/xazap/Projects/plexus-bug/src/main/java/test/Test.java:[11,3] Logger cannot be resolved to a type
[INFO] 6 errors 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
```

The issue does not appear when using Javac as the compiler as proven by  `mvn clean compile -f pom-using-javac.xml`:
```
[INFO] --- compiler:3.12.1:compile (default-compile) @ test ---
[INFO] Recompiling the module because of changed source code.
[WARNING] File encoding has not been set, using platform encoding UTF-8, i.e. build is platform dependent!
[INFO] Compiling 2 source files with javac [debug release 17 module-path] to target/classes
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
```

### LINKED ISSUE #2778: ECJ batch compiler fails to recognize multi-release module on module-path
This is a clone of https://bugs.eclipse.org/bugs/show_bug.cgi?id=577790, as suggested in https://github.com/eclipse-jdt/eclipse.jdt.core/discussions/2773#discussioncomment-10183796

## Version
`ecj-3.38.0.jar` from Maven Central

(and earlier versions, see Bugzilla report)

## Description
Libraries which want to target Java 8 (respectively Android) as lowest version, but also want to provide a module descriptor commonly use multi-release JARs for this, e.g. `META-INF/versions/9/module-info.class`. This is also explicitly mentioned in the [JAR specification](https://docs.oracle.com/en/java/javase/21/docs/specs/jar/jar.html#modular-multi-release-jar-files):
> A module descriptor may be present under a versioned area but not present under the top-level directory.

The problem is that the ECJ batch compiler when used with `--release N` only considers `META-INF/versions/N`, but not any directories < N of libraries in the module path.

This differs from `javac` behavior, from Java runtime behavior (as required by the JAR specification) and also from Eclipse JDT behavior within the IDE (that has a different problem though, see #2495).

This report here is only about `--release`; currently the ECJ batch compiler does not read multi-release classes at all when using `-target`, see #2774.

There was a pull request for this, but it has been abandoned: #61

## Reproduction steps
1. Download the attached [multi-release-test.zip](https://github.com/user-attachments/files/16433306/multi-release-test.zip) and unzip it
2. Download ECJ, for example from https://search.maven.org/artifact/org.eclipse.jdt/ecj
3. Run ECJ with `--release 11`
    ```
    java -jar ./ecj.jar --release 11 --module-path ./multi-release.jar ./module-info.java
    ```
    ❌ Bug: It cannot find the referenced module
4. Run ECJ with `--release 10` (which matches exactly the `META-INF/versions/10` in the JAR)
    ```
    java -jar ./ecj.jar --release 10 --module-path ./multi-release.jar ./module-info.java
    ```
    ℹ️ As expected: Compilation succeeds

wilx:
## What it does

Fixes https://github.com/codehaus-plexus/plexus-compiler/issues/380 and fixes #2778.

ECJ previously looked only in the exact `META-INF/versions/<release>` directory of a multi-release JAR. As a result, compiling for Java 17 did not find a `module-info.class` or class present in `META-INF/versions/9`, even though JEP 238 requires selecting the highest applicable version not greater than the requested release.

For module descriptors, this change probes candidate paths directly from the requested release down to 9, falling back to the descriptor in the JAR root. For ordinary classes, it builds the package cache and descending applicable-version index together during one lazy archive scan.

The regression tests build a multi-release JAR with Java 8 base content, a module descriptor and class in the Java 9 versioned directory, and a competing class in the Java 11 directory. They verify lower-compatible lookup with `--release 17` and that the highest compatible version wins.

AI assistance: OpenAI Codex helped investigate, implement, and test this change; the assistance is also recorded in the commit trailers.

## How to test

- Run `MultiReleaseJarTests`; 27 compliance-expanded tests pass, including `test007_moduleDescriptorFromLowerRelease`, `test008_classFromLowerRelease`, and `test009_highestCompatibleVersionWins`.
- A complete 17-module reactor run of the original fix passed 187,881 tests. After the archive-scan optimizations, the compiler suite passed 58,541 tests and the builder suite passed 386 tests; the model suite had one transient failure in unrelated `JavaProjectTests.testChangeZIPArchive3`, and all 117 `JavaProjectTests` passed on rerun in the Tycho/OSGi harness.
- The original Plexus Compiler issue reproducer was rerun on Java 17 with the newly built `ecj.jar` and compiled successfully.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathMultiReleaseJar.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathMultiReleaseJar.java
@@ -2,14 +2,13 @@ package org.eclipse.jdt.internal.compiler.batch;
 
 import java.io.File;
 import java.io.IOException;
-import java.nio.file.DirectoryStream;
-import java.nio.file.FileVisitResult;
-import java.nio.file.FileVisitor;
-import java.nio.file.Files;
-import java.nio.file.Path;
-import java.nio.file.attribute.BasicFileAttributes;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
 import java.util.Enumeration;
 import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
 import java.util.zip.ZipEntry;
 import org.eclipse.jdt.internal.compiler.classfmt.ClassFileReader;
 import org.eclipse.jdt.internal.compiler.classfmt.ClassFormatException;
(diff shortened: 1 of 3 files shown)


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathMultiReleaseJar.java
```diff
@@ -18,97 +17,85 @@
...
-							}
-
-							@Override
-							public FileVisitResult postVisitDirectory(java.nio.file.Path dir, IOException exc)
-									throws IOException {
-								return FileVisitResult.CONTINUE;
-							}
-						});
+		int prefixLength = Util.METAINF_VERSIONS.length();
+		Set<Integer> versions = new HashSet<>();
```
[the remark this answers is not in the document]
wilx: I noticed this too. I added a test on a separate branch on top of this to test this: https://github.com/wilx/eclipse.jdt.core/commit/2480424379565bb57147b52fbd577a5870dfafed

If it is desired, I can add the fix to this branch.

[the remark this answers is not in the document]
iloveeclipse: Since this is not directly related / needed for this PR, and can / should work also without it, could you please create dedicated PR with your test?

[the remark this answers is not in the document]
wilx: > Since this is not directly related / needed for this PR, and can / should work also without it, could you please create dedicated PR with your test?

Done.

iloveeclipse:
@wilx : can you please squash all 3 commits to one, rebase on head from master & force push? Thanks.

wilx:
> @wilx : can you please squash all 3 commits to one, rebase on head from master & force push? Thanks.

Done.

iloveeclipse:
Also would be nice if your commit message last line would mention that it fixes original issue, like below
`Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2778`

wilx:
> Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2778

Done, too.

iloveeclipse:
[From the spec](https://docs.oracle.com/en/java/javase/17/docs/specs/jar/jar.html#multi-release-jar-files), this is what the PR implements (and which was implemented wrong by ecj before):

> If a multi-release JAR file is deployed on the class path or module path (as an automatic module or an explicit [multi-release module](https://docs.oracle.com/en/java/javase/17/docs/specs/jar/jar.html#modular-multi-release-jar-files)) of major version N of a Java platform release runtime, then a class loader loading classes from that JAR file will first search for class files under the Nth versioned directory, then prior versioned directories in descending order (if present), down to a lower major version bound of 9, and finally under the top-level directory.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathMultiReleaseJar.java
```diff
@@ -18,97 +17,85 @@
...
+			if (fileName.startsWith(Util.METAINF_VERSIONS)) {
+				int separator = fileName.indexOf('/', prefixLength);
+				if (separator == -1) {
+					continue;
+				}
+				String versionSegment = fileName.substring(prefixLength, separator);
+				try {
+					int version = Integer.parseInt(versionSegment);
+					if (version < 9 || version > this.releaseVersion
+							|| !Integer.toString(version).equals(versionSegment)) {
```
iloveeclipse: That `!Integer.toString(version).equals(versionSegment)` surprized me while reading, but the spec expects for `META-INF/versions/N`
```
N must conform to the specification: N:	{1-9} {0-9}*
```
which rules out such things like `009` or `+9`. Can you please add comment above saying something about this?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/batch/ClasspathMultiReleaseJar.java
```diff
@@ -18,97 +17,85 @@
...
 import org.eclipse.jdt.internal.compiler.lookup.BinaryTypeBinding.ExternalAnnotationStatus;
-import org.eclipse.jdt.internal.compiler.util.JRTUtil;
 import org.eclipse.jdt.internal.compiler.util.SuffixConstants;
 import org.eclipse.jdt.internal.compiler.util.Util;
 
 public class ClasspathMultiReleaseJar extends ClasspathJar {
-	private java.nio.file.FileSystem fs = null;
-	Path releasePath = null;
-	String compliance = null;
+	private List<String> releaseVersions = Collections.emptyList();
```
iloveeclipse: 1) I'm a big fan of initializing things in constructor because while debugging constructors we don't jump around, could you please move this to the constructor below?
2) Can you please add comment saying this is versions list for version folders in the jar, sorted in descending order?

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/MultiReleaseJarTests.java
```diff
@@ -31,6 +33,77 @@ public static Test suite() {
 	public static Class<?> testClass() {
 		return MultiReleaseJarTests.class;
 	}
+
+	private String createMultiReleaseJar() throws IOException {
```
iloveeclipse: Please make static and add a comment briefly explaining jar structure, which should help understanding the tests based on that jar

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/MultiReleaseJarTests.java
```diff
@@ -214,4 +287,60 @@ public void test006() {
...
+				"public class X {\n" +
+				"  String value = new Precedence().value();\n" +
+				"}\n"
+			},
+			"\"" + OUTPUT_DIR + File.separator + "src" + File.separator + "X.java\"" +
+			" -classpath \"" + path + "\" --release 17 ",
+			"",
+			"",
+			false);
+	}
```
iloveeclipse, line 372: PR says:

> For module descriptors, this change probes candidate paths directly from the requested release down to 9, falling back to the descriptor in the JAR root.

This is based on [the spect](https://docs.oracle.com/en/java/javase/17/docs/specs/jar/jar.html#multi-release-jar-files))

> A modular multi-release JAR file is a multi-release JAR file that has a module descriptor, module-info.class, in the top-level directory (as for a [modular](https://docs.oracle.com/en/java/javase/17/docs/specs/jar/jar.html#modular-jar-files) JAR file), or directly in a versioned directory.

I miss the test case for that. Is this because it is broken in yet another place?

iloveeclipse (commented):
Looks good, just some nitpics from me and a I have a question regarding test coverage.

wilx:
@iloveeclipse I have pushed update addressing the review comments.

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/MultiReleaseJarTests.java
iloveeclipse: Note: the new module test doesn't fail without the patch. So if you observed some issue with the original code before, the test doesn't reproduce this issue correctly.

iloveeclipse:
> I have pushed update addressing the review comments.

Could you please check my comment regarding the module-info lookup test which doesn't fail **also without** your patch? 
Beside this this change looks good.

wilx:
> > I have pushed update addressing the review comments.
> 
> Could you please check my comment regarding the module-info lookup test which doesn't fail **also without** your patch? Beside this this change looks good.

Without the patch, ECJ still had the fallback, it just looked only for exact version dir. So there is no way to construct a failing test. I just added the test because I don't see any other test testing the fallback.

iloveeclipse:
Thanks. Please rebase other patch on master, and we could merge that too

PR #5237, opened 2026-07-26
[Generics] Switching the order of extended interfaces alters the result of @NonNull annotation analysis

### LINKED ISSUE #4297: [Generics] Switching the order of extended interfaces alters the result of @NonNull annotation analysis

stephan-herrmann:
+ no longer try to find the most specific method based on null annotations
+ but create a view of the method with merged null annotations

The former strategy stems from #2327, were we used null annotations to steer the "arbitrary selection" in case of multiple maximally specific methods. This strategy was incomplete for cases where none of the maximally specific methods was null-compatible to all others. Also modifying the already-complex logic in `mostSpecificMethodBinding()` led to trouble of its own.

The new strategy lets `mostSpecificMethodBinding()` (almost) complete without considering null annotations. Only right when returning the selected method, and if we found several most specific methods, then we create a view of the selected method with null annotations merged from all candidates: pick the weakest parameter types and the strongest return type.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4297

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/NullAnnotationMatching.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/NullAnnotationMatching.java
@@ -851,4 +851,20 @@ public class NullAnnotationMatching {
 		buf.append(" nullStatus="+this.nullStatus); //$NON-NLS-1$
 		return buf.toString();
 	}
+	public static MethodBinding methodWithMergedNullAnnotations(MethodBinding current, MethodBinding[] moreSpecific, int count, LookupEnvironment environment) {
+		if (count < 2)
+			return current;
+		TypeBinding[] parameters = weakerTypes(moreSpecific[0].parameters, moreSpecific[1].parameters, environment);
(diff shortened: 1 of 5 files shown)


stephan-herrmann:
See also https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2325#issuecomment-2050321950 (look for "Re 'contract merge'")

PR #5242, opened 2026-07-27
Honor the multi-release JAR manifest #5241

### LINKED ISSUE #5241: ECJ uses versioned JAR entries without Multi-Release manifest header
## Description

ECJ treats entries under `META-INF/versions/<n>/` as active when compiling with `--release`, even when the JAR's main manifest does not contain `Multi-Release: true`.

This affects both ordinary class lookup and versioned `module-info.class` lookup. For example, given a JAR containing a root class with the constant value `root` and `META-INF/versions/9/` containing the same class with the value `version-9`, but no `Multi-Release` header, javac selects `root` while ECJ selects `version-9`.

The same problem can make ECJ use a versioned module descriptor instead of the root automatic-module metadata.

## Expected behavior

Per [JEP 238](https://openjdk.org/jeps/238) and the [JAR file specification](https://docs.oracle.com/en/java/javase/17/docs/specs/jar/jar.html#multi-release-jar-files), versioned entries should participate in lookup only when the main manifest declares `Multi-Release: true`. Without that header, ECJ should ignore `META-INF/versions/*`, matching javac.

This was identified during review of #5221 and is being tracked separately as requested in [the review discussion](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/5221#discussion_r3656674749).

wilx:
## What it does

Fixes #5241.

ECJ previously used entries under `META-INF/versions/<n>/` with `--release` even when the JAR's main manifest did not declare `Multi-Release: true`. This differed from javac and the JAR specification.

This change gates both versioned class lookup and versioned `module-info.class` lookup on the manifest attribute. It adds a javac-parity regression test for class lookup, adds batch-compiler coverage for versioned module descriptors, and marks the existing MR-JAR fixture as multi-release.

This work was split from #5221 into an independent `master`-based PR at the maintainer's request in [the review discussion](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/5221#discussion_r3656674749).

AI assistance: OpenAI Codex helped investigate, implement, and test this change; the assistance is also recorded in the commit trailers.

## How to test

Before the fix, `MultiReleaseJarCompilerToolTest` fails with:

```text
javac output: root
ECJ output: version-9
```

After the fix:

- `AllCompilerToolTests`: 75 tests passed.
- `MultiReleaseJarTests`: 91 compliance-expanded tests passed, including the missing-header module-descriptor case.

## Author checklist


wilx:
Rebased.

iloveeclipse:
Similar to other PR, please add "Fixes XYZ" line to the last commit message.

wilx:
> Similar to other PR, please add "Fixes XYZ" line to the last commit message.

Done.

iloveeclipse (approved):

iloveeclipse:
Thanks!

PR #5247, opened 2026-07-28
Validate that loaded classes implement javax.annotation.processing.Processor before instantiating

jarthana:
## Author checklist


org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/apt/dispatch/BatchAnnotationProcessorManager.java
```diff
@@ -148,6 +148,10 @@ public ProcessorInfo discoverNextProcessor() {
 				String proc = this._commandLineProcessorIter.next();
 				try {
 					Class<?> clazz = this._procLoader.loadClass(proc);
+					if (!Processor.class.isAssignableFrom(clazz)) {
```
iloveeclipse, line 151: Not sure why do we need this. At the end, the original code should also throw same exception if a class not a `Processor`. And if this is about security, as attacker all what I need is to impement the interface and that will be enough to make this patch useless.

jarthana: Agree. It doesn't take much to overcome this. Will close this.

PR #5256, opened 2026-07-30
Missing unchecked cast warning in polymorphic method call

stephan-herrmann:
Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/3651

Issue #3651, opened 2025-02-01
Missing unchecked cast warning in polymorphic method call


stephan-herrmann:
Ecj exhibits the same bug that javac < 24 did:
```
 import java.lang.invoke.VarHandle;
class VarHandleCast<V> {
     VarHandle vh;
     V method(Object obj) {
         return (V)vh.getAndSet(this, obj);
     }
} 
```
this should trigger an unchecked cast warning, but doesn't.

See https://bugs.openjdk.org/browse/JDK-8343286

PR #5260, opened 2026-07-31
Missing unchecked cast warning in polymorphic method call

stephan-herrmann:
+ Suppress CheckCast after polymorphic method call

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5256

See https://github.com/eclipse-jdt/eclipse.jdt.core/pull/5256#issuecomment-5139794756 f.

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/CastExpression.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/CastExpression.java
@@ -655,10 +655,12 @@ public TypeBinding resolveType(BlockScope scope) {
 	}
 	TypeBinding expressionType = this.expression.resolveType(scope);
 	TypeBinding originalExpressionType = expressionType;
+	boolean isPolymorphicMethod = false;
 	if (this.expression instanceof MessageSend) {
 		MessageSend messageSend = (MessageSend) this.expression;
 		MethodBinding methodBinding = messageSend.binding;
 		if (methodBinding != null && methodBinding.isPolymorphic()) {
+			isPolymorphicMethod = true;
 			messageSend.binding = scope.environment().updatePolymorphicMethodReturnType((PolymorphicMethodBinding) methodBinding, castType);
 			if (TypeBinding.notEquals(expressionType, castType)) {
 				expressionType = castType;
@@ -681,6 +683,8 @@ public TypeBinding resolveType(BlockScope scope) {
 			}
 			boolean isLegal = checkCastTypesCompatibility(scope, castType, originalExpressionType, this.expression, true);
 			if (isLegal) {
+				if (isPolymorphicMethod)
+					this.bits &= ~ASTNode.GenerateCheckcast;
 				this.expression.computeConversion(scope, castType, expressionType);
 				if ((this.bits & ASTNode.UnsafeCast) != 0) { // unsafe cast
 					if (scope.compilerOptions().reportUnavoidableGenericTypeProblems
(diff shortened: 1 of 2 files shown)


iloveeclipse:
Do I see it right, the extra cast was not expected? If so, we should touch the platform bundle again like in https://github.com/eclipse-platform/eclipse.platform/pull/2842 for the next build with this PR merged.

stephan-herrmann:
> Do I see it right, the extra cast was not expected? 

correct

> If so, we should touch the platform bundle again like in [eclipse-platform/eclipse.platform#2842](https://github.com/eclipse-platform/eclipse.platform/pull/2842) for the next build with this PR merged.

Has an I-build already picked up that touching? Otherwise we might be good already?

iloveeclipse:
> Has an I-build already picked up that touching? Otherwise we might be good already?

Yes, I've built one already.
I will tuch that bundle again.

iloveeclipse:
> > Do I see it right, the extra cast was not expected?
> 
> correct

It seem counterintuitive? Could you help to understand this?

stephan-herrmann:
> > > Do I see it right, the extra cast was not expected?
> > 
> > 
> > correct
> 
> It seem counterintuitive? Could you help to understand this?

I'll try, though those polymorphic method calls are not my everyday business either.

The declaration of methods like `VarHandle.getAndSet()` and `MethodHandle.invokeExact()`  has signature `(Object[])Object`, i.e., they accept any arguments in any number, and produce some result.

[JLS 15.2.3](https://docs.oracle.com/javase/specs/jls/se26/html/jls-15.html#jls-15.12.3), however, defines the "compile-time parameter types" and "compile-time result" to be determined from the context, in particular:
>
>...
>* Otherwise, if the method invocation expression is the operand of a cast expression ([§15.16](https://docs.oracle.com/javase/specs/jls/se26/html/jls-15.html#jls-15.16)), the compile-time result is the erasure of the type of the cast expression ([§4.6](https://docs.oracle.com/javase/specs/jls/se26/html/jls-4.html#jls-4.6)).


So, if you write ...
```java
(String) methodHandle.invokeExact(a)
```
... the expression `methodHandle.invokeExact(a)` is considered to have a signature with `String` as its return type. IOW, the enclosing cast is not a conversion but informs the compiler about the assumed shape of the underlying method. Whether or not the actual `MethodHandle` satisfies the assumption is subject to runtime checking inside `invokeExact`.  If you try to fool type checking instead of a ClassCastException you'll be greeted like this:
```
Exception in thread "main" java.lang.invoke.WrongMethodTypeException: handle's method type (CheckCast)CheckCast but found (CheckCast)String
	at java.base/java.lang.invoke.Invokers.newWrongMethodTypeException(Invokers.java:522)
	at java.base/java.lang.invoke.Invokers.checkExactType(Invokers.java:531)
```

Ergo, the invocation already "has" the type requested by the cast, so no additional check via a `checkcast` operation is needed.

Still the compiler should check for unsafe operations due to erasure. In the original example of #3651 a cast to a type variable `V` is something that the runtime cannot check, since `V` has no representation in byte code. Hence the compiler needs to perform that part of analysis that issues "unsafe cast" warnings (#5256), without, however, actually generating a `checkcast` (this PR).

Does this make sense?

I checked with javac, which also does not issue a `checkcast` for the two new tests.

stephan-herrmann:
To further expand on this part:
> IOW, the enclosing cast is not a conversion but informs the compiler about the assumed shape of the underlying method. 

The type information from the cast is then used for the signature in the `invokevirtual` bytecode, like:
```
2: invokevirtual #20  // Method java/lang/invoke/MethodHandle.invokeExact:(Ljava/lang/Object;)Ljava/lang/String;
```

The `String` return type is derived from the cast, and replaces the original return type `Object`.

iloveeclipse:
OK, thanks.

PR #5269, opened 2026-08-06
Bogus error about missing indirectly referenced types during compilation in presence of native method invocation

### LINKED ISSUE #1175: Looks into indirect types during compilation in presence of native method invocation
This is related to issue #543. I am including ([bug.zip](https://github.com/eclipse-jdt/eclipse.jdt.core/files/11819676/bug.zip)) an example Eclipse 2023-06 workspace with a project, a sliced version of https://github.com/pietrobraione/jbse with just two classes. The project is a Java 8 project containing one class called jbse.base.JAVA_CONCURRENTMAP. This class extends java.util.AbstractMap and overrides the hashCode() method. In the code of the overriding implementation of the hashCode() method there is an invocation to a native method refineOnKeyCombinationsAndBranch. If this invocation is removed, the code compiles correctly. If this invocation is kept, compilation fails with the error:

Description	Resource	Path	Location	Type
The type java.lang.invoke.VarHandle cannot be resolved. It is indirectly referenced from required .class files	JAVA_CONCURRENTMAP.java	/jbse/src/main/java/jbse/base	line 1	Java Problem

I am reporting this issue as a related but distinct issue from #543 because I started to experience it only recently, if I do not remember bad it should be since Eclipse 2022-06.

srikanth-sankaran:
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/1175

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/MethodBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/MethodBinding.java
@@ -1456,8 +1456,10 @@ public boolean hasPolymorphicSignature(Scope scope) {
 	}
 	return this.isNative() && this.isVarargs() && this.parameters.length == 1 &&
 			this.parameters[0].leafComponentType().id == TypeIds.T_JavaLangObject &&
-				(TypeBinding.equalsEquals(this.declaringClass, scope.getJavaLangInvokeMethodHandle())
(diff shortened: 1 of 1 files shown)


srikanth-sankaran:
@jarthana - As this does not have a regression test, May I ask for a quick review and verification from you ? 

(requires jdk8 to be the runtime, none of our builds do that I believe)

jarthana (approved):

srikanth-sankaran:
Thanks Jay!

PR #5284, opened 2026-08-13
Disable indexing for restricted binary files

### LINKED ISSUE #4970: Add preference for disabling indexing of specific files
We have some sources, which we want to exclude from the Java indexing. We wish to add a preference for this and mark the files with a session property. If both the preference is set and the session property is present at a file, the file should be excluded from indexing.

See also:
* https://github.com/eclipse-platform/eclipse.platform/issues/2588
* https://github.com/eclipse-platform/eclipse.platform.ui/issues/3815

trancexpress:
Indexing for restricted files is so far disabled only for sources, not for binary files. This change ensures also restricted binary files are not indexed.

Fixes: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4970

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/IndexManagerTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/IndexManagerTests.java
@@ -27,6 +27,7 @@ import junit.framework.Test;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IResource;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.Path;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences;
@@ -186,6 +187,7 @@ public class IndexManagerTests extends ModifyingResourceTests {
 
 	public void testDisableIndexingForRestrictedFile() throws Exception {
 		if (SKIP_TESTS) return;
+		startLogListening();
 		this.indexDisabledForTest = true;
 		IEclipsePreferences node = InstanceScope.INSTANCE.getNode(JavaCore.PLUGIN_ID);
 		String pref = JavaModelManager.DISABLE_RESTRICTED_FILE_INDEXING_PREFERENCE;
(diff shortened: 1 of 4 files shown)


org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/IndexManagerTests.java
```diff
@@ -214,7 +221,14 @@ public void testDisableIndexingForRestrictedFile() throws Exception {
 			assertSearchResults(
 				"src/p/TestClass2.java p.TestClass2 [TestClass2]",
 				collector);
+
+			// ensure we disabled indexing for the binary file as well
+			String binaryIndexerMessage = "Could not index empty /IndexProject/src/p/TestBroken.class";
+			List<IStatus> logs = this.logListener.getLogs();
+			boolean unexpectedLog = logs.stream().map(IStatus::getMessage).anyMatch(m -> binaryIndexerMessage.equals(m));
+			assertFalse("Expected no log from binary indexer", unexpectedLog);
```
trancexpress, line 229: Not the best and definitely not the most robust way to test this... but its simple. I can try to come up with something else, but it won't be simple getting a class file that has the expected matches.

org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/indexing/BinaryIndexer.java
```diff
@@ -973,4 +978,12 @@ private char[] replace(char toBeReplaced, char newChar, char[] array) {
 		}
 		return array;
 	}
+
+	private boolean disabledForFile() {
+		if (JavaModelManager.disableRestrictedFileIndexing()) {
+			IFile file = getDocumentFile(this.document);
```
iloveeclipse, line 984: For binary indexer, in most cases the IFile will not exist, so on every file from a jar or jre it will throw exception, which is not OK.

iloveeclipse:
Ideally we only check at places which indexer has "IFile" in hand, like:
 
* `org.eclipse.jdt.internal.core.search.indexing.IndexManager.addSource(IFile, IPath, SourceElementParser)`
* `org.eclipse.jdt.internal.core.search.indexing.IndexManager.addBinary(IFile, IPath)`

So we don't need to reconstruct the `IFile` from some path which was never `IFile`.

trancexpress:
> Ideally we only check at places which indexer has "IFile" in hand, like:
> 
>     * `org.eclipse.jdt.internal.core.search.indexing.IndexManager.addSource(IFile, IPath, SourceElementParser)`
>     * `org.eclipse.jdt.internal.core.search.indexing.IndexManager.addBinary(IFile, IPath)`
> 
> So we don't need to reconstruct the `IFile` from some path which was never `IFile`.

My main reason for putting code directly in `SourceIndexer` with the first PR was to ensure we always check the session property, regardless of who the caller is.

trancexpress:
So far its seems we won't be needing this, lets close. We can always re-open if we do need the changes.

PR #5288, opened 2026-08-16
Fix to Ensure CaptureBinding.initializeBounds is called only once per

### LINKED ISSUE #4869: Ensure CaptureBinding.initializeBounds is called only once per binding
Fallout from https://github.com/eclipse-jdt/eclipse.jdt.core/pull/4862#issuecomment-3941063688:

> Inside `PTB.capture()` method `initializeBounds()` is invoked on only such elements `capturedArguments[i]` where `freshCaptures[i]` is true, which should "normally" imply that this capture binding has been created just now, i.e., never exposed to any outside code. [...]
>
> Theory thus suggests, that the capture isn't really fresh, i.e., `this.environment.createCapturedWildcard()` probably answered a pre-existing capture binding. This smells like inference computing the same capture more than once, i.e., inference performing several attempts with similar inputs.
>
> That theory also suggest, that it would be prudent to add an early exit to `CaptureBinding.initializeBounds()` to avoid repeated initialized in the first place (that would be an additional safety measure on top of the current fix).

mpalat:
fixes #4869 

## What it does
Ensures CaptureBinding.initializeBounds is called only once per binding

## How to test
a junit test provided

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/lookup/CaptureBinding.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2024 IBM Corporation and others.
+ * Copyright (c) 2000, 2026 IBM Corporation and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -40,6 +40,7 @@ public class CaptureBinding extends TypeVariableBinding {
 	public int start;
 	public int end;
 	public ASTNode cud; // to facilitate recaptures.
+	private boolean boundsInitialized;
 
 	TypeBinding pendingSubstitute; // for substitution of recursive captures, see https://bugs.eclipse.org/456924
 
@@ -92,6 +93,7 @@ public class CaptureBinding extends TypeVariableBinding {
 		this.lowerBound = prototype.lowerBound;
 		this.tagBits |= (prototype.tagBits & TagBits.HasCapturedWildcard);
 		this.cud = prototype.cud;
+		this.boundsInitialized = prototype.boundsInitialized;
 	}
 
 	// Captures may get cloned and annotated during type inference.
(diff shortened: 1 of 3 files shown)


mpalat:
@stephan-herrmann an attempt to address #4869. Once the tests pass, would like to have a quick review. Thanks!

mpalat:
Also, @stephan-herrmann since M3 closed a day before, shall we target this to RC1 (preferred unless you see any possible issues) or 4.42M1?

stephan-herrmann:
> Also, @stephan-herrmann since M3 closed a day before, shall we target this to RC1 (preferred unless you see any possible issues) or 4.42M1?

While the fix is simple, I don't see any urgency to squeeze it in post M3.

I do have two more ideas, though:
* Now that we can detect whether a CaptureBinding has its bounds initialized, we could leverage this already during `ParameterizedTypeBinding.capture(Scope, int, int)` and avoid setting `freshCaptures[i] = true` if the capture isn't really fresh -- this would serve as an optimization to avoid executing the entire try-finally where `initializeBounds()` is invoked.
* If we do this, then we could discuss: do we still expect other call chains the might try double initialization, or could we make the check at the head of `initializeBounds()` an assert rather than an early exit? This would let us detect such unexpected call chains.
   * As a compromise we could do both: assert (only active under `-ea`) + the early exit

WDYT?

mpalat:
@stephan-herrmann  - my comments below:

> * Now that we can detect whether a CaptureBinding has its bounds initialized, we could leverage this already during `ParameterizedTypeBinding.capture(Scope, int, int)` and avoid setting `freshCaptures[i] = true` if the capture isn't really fresh -- this would serve as an optimization to avoid executing the entire try-finally where `initializeBounds()` is invoked.

Agree.  I have added [this in the PR](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/5288/commits/a37e88665dc34d27f6b8e0f08c3399ea8e82d1d4).


> * If we do this, then we could discuss: do we still expect other call chains the might try double initialization, or could we make the check at the head of `initializeBounds()` an assert rather than an early exit? This would let us detect such unexpected call chains.

I see only one (code) reference of `initializeBounds()` implying that I don't see multiple call chains for this. Am I missing something here ? TIA

PR #5292, opened 2026-08-17
Fix StackOverflowError in HierarchyResolver 

MeherSru:
This adds a simple check: track which files are currently being processed, and skip re-processing a file that's already in progress.


iloveeclipse:
Can you please provide steps to reproduce the problem, where the fix would be needed?

MeherSru:
> Can you please provide steps to reproduce the problem, where the fix would be needed?

Sure, here's the exact repro I used, it's the reporter's own script linked in the issue:

- git clone https://github.com/usrlocl/jdtls-stackoverflow-reproduce
- cd jdtls-stackoverflow-reproduce
- JDTLS_HOME=<path-to-a-built-jdtls> node jdtls-calls.mjs fail "fail/src/main/java/foo/Foo.java" 5 8

iloveeclipse:
> it's the reporter's own script linked in the issue

Can you please provide link to the original issue?

MeherSru:
> > it's the reporter's own script linked in the issue
> 
> Can you please provide link to the original issue?

Sorry, missed linking it earlier, here's the issue: https://github.com/eclipse-jdtls/eclipse.jdt.ls/issues/3866

PR #5309, opened 2026-08-22
[performance] compute compiler options lazily in determineIfOnClasspath

### LINKED ISSUE #5308: determineIfOnClasspath: getOption calls cost 78% of the all-projects scan
While testing our Eclipse based MCP server (https://github.com/vogellacompany/eclipse-mcp-server), we found lots of UI freezes in a larger workspace in`JavaModelManager.create:1028`.

`JavaModelManager.create(IFolder, ...)` scans every Java project when a folder is on no classpath (JavaModelManager.java:1026-1035). Per project, `determineIfOnClasspath` calls:

````
String sourceLevel = project.getOption(JavaCore.COMPILER_SOURCE, true);
String complianceLevel = project.getOption(JavaCore.COMPILER_COMPLIANCE, true);
````

Both are computed up front but only read inside the `isPrefixOf` branch. And they are not cheap: `getOption` reaches `hasJavaNature`, which reads the resource tree twice.

If the call is deferred to the needed case, we see a ~75-78% performance improvement off that scan.

Patch is trivial, PR to follow.

vogella:
`determineIfOnClasspath` read the source and compliance level before the classpath entry loop, but both are only used inside the `isPrefixOf` branch, so on a miss they were computed and discarded. They are not free: `getOption` reaches `hasJavaNature`, which reads the resource tree twice.

This sits on the scan that `create(IFolder, IJavaProject)` runs over every Java project when a folder is on no classpath, and unlike a hit that scan caches nothing. Deferring the two calls to the branch that needs them cuts 70 to 78 percent off the scan, measured with an interleaved in-JVM A/B from 50 to 400 projects. The change is behaviour preserving and covered by `JavaModelTests.testCreatePkgHandleInDifferentProject`.

It fixes the constant, not the O(projects) scan itself; caching the negative result the way #3470 caches the positive one would be the follow-up.

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5308

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaModelTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/JavaModelTests.java
@@ -361,6 +361,27 @@ public void testCreatePkgHandleInDifferentProject() throws CoreException {
 }
 
 /*
+ * Ensure that a non-Java-like resource at the root of a classpath entry
+ * defined in a different project resolves to the default package fragment.
+ */
+public void testCreateDefaultPkgHandleInDifferentProject() throws CoreException {
+	try {
+		createJavaProject("P1", new String[] {}, "bin");
+		createFolder("/P1/lib");
+		IFile file = createFile("/P1/lib/readme.txt", "");
+		IJavaProject p2 = createJavaProject("P2", new String[] {}, new String[] {"/P1/lib"}, "");
+		IJavaElement element = JavaModelManager.determineIfOnClasspath(file, p2);
+		assertElementEquals(
+			"Unexpected element",
+			"<default> [in /P1/lib [in P2]]",
+			element
+		);
+	} finally {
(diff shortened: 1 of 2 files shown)


vogella:
FYI: the model suite passed locally (25088 tests, 0 failures) with this change.

skrcode:
I ran JAIPilot Cloud against this exact PR head. It disproved the suspected per-entry repeat because the branch always returns, then found a narrower deterministic follow-up: default-package classpath hits still fetched source and compliance options even though that path never validates a package name.

The change makes that path perform 0 rather than 2 IJavaProject.getOption calls. The same 28 JavaModelTests passed before and after, including a new cross-project default-package case, and the core plus tests.model reactor passed. No wall-clock speed claim is being made.

PR directly onto this source branch: https://github.com/vogella/eclipse.jdt.core/pull/2  
Cloud-generated draft and full evidence: https://github.com/skrcode/eclipse.jdt.core/pull/2

This is optional and deliberately limited to the default-package path.

PR #5310, opened 2026-08-22
external annotations translated to type annotation even if null annotations lack TYPE_USE target

### LINKED ISSUE #2746: external annotations translated to type annotation even if null annotations lack TYPE_USE target
PR #2551 is about to change the expectation of org.eclipse.jdt.core.tests.model.ExternalAnnotations17Test.testBug565246(). The new expectation is wrong, revealing the following bug:

In a project >= 1.8 which has null annotations configured that are declaration annotations, evaluating external annotations will evaluate to annotations on types, rather than declaration annotations.

stephan-herrmann:
+ re-introduce specific eea-walkers for scanField/Method*

Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2746


stephan-herrmann:
Tests raise lots of `Retention cannot be resolved to a type` for which I see no easy remedy.

PR #5317, opened 2026-08-25
Compiler crashes with NullPointerException when enabling null analysis, annotations silently not compiled to bytecode

### LINKED ISSUE #5316: Compiler crashes with NullPointerException when enabling null analysis, annotations silently not compiled to bytecode
This bug concerns Eclipse 2026-06 and	org.eclipse.jdt 3.19.1300.v20260601-0713 with org.eclipse.jdt.core 3.46.0.v20260520-1003. I can't exactly say when this bug was introduced, but it worked with Eclipse 2025-09.

The annotation `@NonNullByDefault` wreaks havoc when combined with `@Nullable` for a field type, and causes other annotations to be ignored. **But only when `@NonNullByDefault` is not the last annotation on the field!**

Take this example class from the reproduction sample project [jdt-bug-project.zip](https://github.com/user-attachments/files/31424374/jdt-bug-project.zip):
```java
package org.eclipse.jdt.bug;

import java.util.Map;

import org.eclipse.jdt.annotation.NonNullByDefault;
import org.eclipse.jdt.annotation.Nullable;

public class MyInnocentClass {
	@NonNullByDefault
	@Nullable
	private Map<String, String> aTotallyNormalMap;
}
```

Observe that it compiles. Now check the "Enable annotation-based null analysis" checkbox in Project Properties > Java Compiler > Errors/Warnings. Nothing else has to be done, the severities of all items can stay the same.

Rebuild the project and observe that the JDT compiler now crashes when compiling the file: Also take note that the second annotation is not syntax-highlighted as an annotation in the editor, but rather as plain text.
<details><summary>JDT stack trace</summary>
<p>

```
java.lang.RuntimeException: Internal Error compiling /bug.mve/src/org/eclipse/jdt/bug/MyInnocentClass.java
	at org.eclipse.jdt.internal.compiler.ProcessTaskManager.processing(ProcessTaskManager.java:139)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: java.lang.NullPointerException: Cannot read field "id" because "annotation.resolvedType" is null
	at org.eclipse.jdt.internal.compiler.ast.ConstructorDeclaration.isValueProvidedUsingAnnotation(ConstructorDeclaration.java:361)
	at org.eclipse.jdt.internal.compiler.ast.ConstructorDeclaration.doFieldReachAnalysis(ConstructorDeclaration.java:345)
	at org.eclipse.jdt.internal.compiler.ast.ConstructorDeclaration.analyseCode(ConstructorDeclaration.java:291)
	at org.eclipse.jdt.internal.compiler.ast.ConstructorDeclaration.analyseCode(ConstructorDeclaration.java:89)
	at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.internalAnalyseCode(TypeDeclaration.java:878)
	at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.analyseCode(TypeDeclaration.java:316)
	at org.eclipse.jdt.internal.compiler.ast.CompilationUnitDeclaration.analyseCode(CompilationUnitDeclaration.java:127)
	at org.eclipse.jdt.internal.compiler.Compiler.process(Compiler.java:817)
	at org.eclipse.jdt.internal.compiler.ProcessTaskManager.processing(ProcessTaskManager.java:135)
	... 5 more
```
</p>
</details> 

Look at the Error Log view and see that there are now multiple errors with problem detection pointing to the same source:
<details><summary>Error Log stack trace</summary>
<p>

```
Java Model Exception: java.lang.NullPointerException: Cannot read field "id" because "annotation.resolvedType" is null
	at org.eclipse.jdt.internal.core.CompilationUnitProblemFinder.process(CompilationUnitProblemFinder.java:353)
	at org.eclipse.jdt.internal.core.CompilationUnitProblemFinder.process(CompilationUnitProblemFinder.java:375)
	at org.eclipse.jdt.internal.core.ReconcileWorkingCopyOperation.makeConsistent(ReconcileWorkingCopyOperation.java:244)
	at org.eclipse.jdt.internal.core.ReconcileWorkingCopyOperation.executeOperation(ReconcileWorkingCopyOperation.java:108)
	at org.eclipse.jdt.internal.core.JavaModelOperation.run(JavaModelOperation.java:751)
	at org.eclipse.jdt.internal.core.JavaModelOperation.runOperation(JavaModelOperation.java:816)
	at org.eclipse.jdt.internal.core.CompilationUnit.reconcile(CompilationUnit.java:1431)
	at org.eclipse.jdt.internal.ui.text.java.JavaReconcilingStrategy.reconcile(JavaReconcilingStrategy.java:132)
	at org.eclipse.jdt.internal.ui.text.java.JavaReconcilingStrategy$1.run(JavaReconcilingStrategy.java:94)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:47)
	at org.eclipse.jdt.internal.ui.text.java.JavaReconcilingStrategy.reconcile(JavaReconcilingStrategy.java:91)
	at org.eclipse.jdt.internal.ui.text.java.JavaReconcilingStrategy.reconcile(JavaReconcilingStrategy.java:158)
	at org.eclipse.jdt.internal.ui.text.CompositeReconcilingStrategy.reconcile(CompositeReconcilingStrategy.java:94)
	at org.eclipse.jdt.internal.ui.text.JavaCompositeReconcilingStrategy.reconcile(JavaCompositeReconcilingStrategy.java:107)
	at org.eclipse.jface.text.reconciler.MonoReconciler.process(MonoReconciler.java:77)
	at org.eclipse.jface.text.reconciler.AbstractReconciler$BackgroundWorker.run(AbstractReconciler.java:203)
	at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: java.lang.NullPointerException: Cannot read field "id" because "annotation.resolvedType" is null
	at org.eclipse.jdt.internal.compiler.ast.ConstructorDeclaration.isValueProvidedUsingAnnotation(ConstructorDeclaration.java:361)
	at org.eclipse.jdt.internal.compiler.ast.ConstructorDeclaration.doFieldReachAnalysis(ConstructorDeclaration.java:345)
	at org.eclipse.jdt.internal.compiler.ast.ConstructorDeclaration.analyseCode(ConstructorDeclaration.java:291)
	at org.eclipse.jdt.internal.compiler.ast.ConstructorDeclaration.analyseCode(ConstructorDeclaration.java:89)
	at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.internalAnalyseCode(TypeDeclaration.java:878)
	at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.analyseCode(TypeDeclaration.java:316)
	at org.eclipse.jdt.internal.compiler.ast.CompilationUnitDeclaration.analyseCode(CompilationUnitDeclaration.java:127)
	at org.eclipse.jdt.internal.compiler.Compiler.resolve(Compiler.java:978)
	at org.eclipse.jdt.internal.compiler.Compiler.resolve(Compiler.java:1017)
	at org.eclipse.jdt.internal.core.CompilationUnitProblemFinder.process(CompilationUnitProblemFinder.java:309)
	... 16 more
Caused by: java.lang.NullPointerException: Cannot read field "id" because "annotation.resolvedType" is null
	at org.eclipse.jdt.internal.compiler.ast.ConstructorDeclaration.isValueProvidedUsingAnnotation(ConstructorDeclaration.java:361)
	at org.eclipse.jdt.internal.compiler.ast.ConstructorDeclaration.doFieldReachAnalysis(ConstructorDeclaration.java:345)
	at org.eclipse.jdt.internal.compiler.ast.ConstructorDeclaration.analyseCode(ConstructorDeclaration.java:291)
	at org.eclipse.jdt.internal.compiler.ast.ConstructorDeclaration.analyseCode(ConstructorDeclaration.java:89)
	at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.internalAnalyseCode(TypeDeclaration.java:878)
	at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.analyseCode(TypeDeclaration.java:316)
	at org.eclipse.jdt.internal.compiler.ast.CompilationUnitDeclaration.analyseCode(CompilationUnitDeclaration.java:127)
	at org.eclipse.jdt.internal.compiler.Compiler.resolve(Compiler.java:978)
	at org.eclipse.jdt.internal.compiler.Compiler.resolve(Compiler.java:1017)
	at org.eclipse.jdt.internal.core.CompilationUnitProblemFinder.process(CompilationUnitProblemFinder.java:309)
	at org.eclipse.jdt.internal.core.CompilationUnitProblemFinder.process(CompilationUnitProblemFinder.java:375)
	at org.eclipse.jdt.internal.core.ReconcileWorkingCopyOperation.makeConsistent(ReconcileWorkingCopyOperation.java:244)
	at org.eclipse.jdt.internal.core.ReconcileWorkingCopyOperation.executeOperation(ReconcileWorkingCopyOperation.java:108)
	at org.eclipse.jdt.internal.core.JavaModelOperation.run(JavaModelOperation.java:751)
	at org.eclipse.jdt.internal.core.JavaModelOperation.runOperation(JavaModelOperation.java:816)
	at org.eclipse.jdt.internal.core.CompilationUnit.reconcile(CompilationUnit.java:1431)
	at org.eclipse.jdt.internal.ui.text.java.JavaReconcilingStrategy.reconcile(JavaReconcilingStrategy.java:132)
	at org.eclipse.jdt.internal.ui.text.java.JavaReconcilingStrategy$1.run(JavaReconcilingStrategy.java:94)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:47)
	at org.eclipse.jdt.internal.ui.text.java.JavaReconcilingStrategy.reconcile(JavaReconcilingStrategy.java:91)
	at org.eclipse.jdt.internal.ui.text.java.JavaReconcilingStrategy.reconcile(JavaReconcilingStrategy.java:158)
	at org.eclipse.jdt.internal.ui.text.CompositeReconcilingStrategy.reconcile(CompositeReconcilingStrategy.java:94)
	at org.eclipse.jdt.internal.ui.text.JavaCompositeReconcilingStrategy.reconcile(JavaCompositeReconcilingStrategy.java:107)
	at org.eclipse.jface.text.reconciler.MonoReconciler.process(MonoReconciler.java:77)
	at org.eclipse.jface.text.reconciler.AbstractReconciler$BackgroundWorker.run(AbstractReconciler.java:203)
	at java.base/java.lang.Thread.run(Thread.java:1583)
```
</p>
</details> 

Now change the order of the two null analysis annotations and observe that this code magically compiles again:
```java
@Nullable
@NonNullByDefault
private Map<String, String> aTotallyNormalMap;
```

This bug is not constrained to a combination of null analysis annotations, but instead it occurs whenever `@NonNullByDefault` is combined with other annotations, and this annotation is the first. However, it gets even weirder: Whenever the annotation is not the **last** annotation on a field, **any later annotation is silently ignored and not compiled into bytecode.**

Take this example:
```java
@Nullable
@NonNullByDefault
@Deprecated
private Map<String, String> aTotallyNormalMap;
```

Note that this compiles perfectly fine, but when you look at the syntax highlighting, you already see something's fishy: the `@Deprecated` annotation is displayed as plain-text, not as an annotation. And when you hover over the field and look at the generated Javadoc, you see that the annotation isn't actually regarded on the field:
Image

Would you compile this class and use reflection to get the annotations on the field, you would also see that the annotation isn't there, even though it should retained in the compiled code. When you re-order the annotations again, so that `@NonNullByDefault` is the last, all annotations work and the bytecode is generated correctly.

Edit: might correlate with #5249, but the error is a different one.

stephan-herrmann:
Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5316

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ASTNode.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ASTNode.java
@@ -855,6 +855,10 @@ public abstract class ASTNode implements Location, TypeConstants, TypeIds {
 							// need to fill the instances array
 							for (int j = 0; j < length; j++) {
 								Annotation annot = sourceAnnotations[j];
+								if (annot.recipient == null || annot.resolvedType == null) {
+									annot.recipient = recipient;
+									annot.resolveType(scope);
+								}
 								annotations[j] = annot.getCompilerAnnotation();
 							}
 						}
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/NullTypeAnnotationTest.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/NullTypeAnnotationTest.java
@@ -20253,4 +20253,24 @@ public void testGH5249() {
 		getCompilerOptions(),
 		"");
 }
+public void testGH5316() throws Exception {
+	runConformTestWithLibs(new String[] {
+			"MyInnocentClass.java",
+			"""
+			import java.util.Map;
+
+			import org.eclipse.jdt.annotation.NonNullByDefault;
+			import org.eclipse.jdt.annotation.Nullable;
+
+			@SuppressWarnings("unused")
+			public class MyInnocentClass {
+				@NonNullByDefault
+				@Nullable
+				private Map<String, String> aTotallyNormalMap;
+			}
+			"""
+		},
+		getCompilerOptions(),
+		"");
+}
 }


stephan-herrmann:
Here's what happened, when null annotations are enabled:
* `SourceTypeBinding.resolveTypeFor(VariableBinding)` wants to know if `@NonNullByDefault` is active for the given field.
  * inside `resolveNullDefaultAnnotations()` that annotation is selectively resolved incl. setting the annotation's recipient. This fact is recorded by setting `NullDefaultAnnotationResolved`
* later during `ASTNode.resolveAnnotations()` for the same field we find the first annotation already having its `recipient` set, and take a shortcut:
  * in this shortcut sibling annotations are treated only by transfering the "compiler annotation" from the recipient to the array of `AnnotationBinding`s
  * this skips two necessary statements that otherwise would happen at the bottom of the enclosing loop
  * in this method `AllAnnotationsResolved` is set so we never come back here

For a moment I was tempted to replace the statement `return annotations;` of this shortcut, but then I found the motivation in https://bugs.eclipse.org/bugs/show_bug.cgi?id=191908: If an annotation has multiple recipients (as in `@Deprecated int x= 5, y= 10;`) then we want to share the one annotation binding into each recipient.

The existing scheme was broken when we started to selectively resolved only `@NonNullByDefault` during early phases (to avoid dangerous order of compilation steps). Previously, the array of annotations was either all resolved or all unresolved (really? didn't we selectively resolved `@Deprecated` for a long time already?). Now this code paragraph has to check for each annotation if it has been resolved already.

stephan-herrmann:
@srikanth-sankaran from my p.o.v. this is a candidate for RC2.

Do you have the time for a quick review? 

It's a real bug with strange effects. I think the fix is quite safe, and explanation is given above.

If you agree, could you please pass this on to a JDT lead for official +1?

If you have doubts I'm also fine with postponing this to the next release.

srikanth-sankaran:
> I think the fix is quite safe

I think the fix is in the right direction - so leaves us better than where we are.

But I am not sure we are doing the right thing by simply returning at the place we do in this block:

```
					case Binding.TYPE_USE: // deliberately not setting tagBits (see above)
						if (annotations != null) {
							// need to fill the instances array
							for (int j = 0; j < length; j++) {
								Annotation annot = sourceAnnotations[j];
								if (annot.recipient == null || annot.resolvedType == null) {
									annot.recipient = recipient;
									annot.resolveType(scope);
								}
								annotations[j] = annot.getCompilerAnnotation();
							}
						}
						return annotations;
```

The effect of that eager return would appear to be that we would skip

- `tagMissingAnalysisAnnotation(scope, recipient, annotation);` call 
- `checkRepeatableMetaAnnotation()` call
- checking for duplicate annotations.

Here is a snippet that shows one of the problems:

Given:

```
			import java.util.Map;

			import org.eclipse.jdt.annotation.NonNullByDefault;
			import org.eclipse.jdt.annotation.Nullable;

			@SuppressWarnings("unused")
			public class MyInnocentClass {
				@NonNullByDefault
				@Nullable @Nullable
				private Map<String, String> aTotallyNormalMap;
			}
```

we don't complain about the repeated `Nullable` annotation but we complain about it in this scenario without the `NonNullByDefault` :

```
			import java.util.Map;

			import org.eclipse.jdt.annotation.NonNullByDefault;
			import org.eclipse.jdt.annotation.Nullable;

			@SuppressWarnings("unused")
			public class MyInnocentClass {
				@Nullable @Nullable
				private Map<String, String> aTotallyNormalMap;
			}
```

There could be other problems. I didn't analyze this case, but it reports a strange error:

```
			import java.util.Map;

			import org.eclipse.jdt.annotation.NonNullByDefault;
			import org.eclipse.jdt.annotation.Nullable;

			@interface I {}

			@SuppressWarnings("unused")
			public class MyInnocentClass {
				@NonNullByDefault
				@Nullable @I @I
				private Map<String, String> aTotallyNormalMap;
			}
```

elicits:

----------\n
1. INFO in MyInnocentClass.java (at line 8)\n
	@SuppressWarnings("unused")\n
	                  ^^^^^^^^\n
At least one of the problems in category 'unused' is not analysed due to a compiler option being ignored\n
----------\n

srikanth-sankaran:
> I think the fix is quite safe, and explanation is given above.
> 
> If you agree, could you please pass this on to a JDT lead for official +1?

The patch is leaving us in a better place than we are. I don't see how it can introduce adverse effects. So, I am ok with proceeding.

Alternately, we can ordinate so that a fuller fix could be a first and sole commit in the first I-build after 4.41 ships.

I'll let you take the call - @mpalat - this is good to go from my pov, needs JDT Lead approval - FYI

mpalat (approved):
+1 from my side as JDT Lead for RC2. Looks good and I did n't see much risk as well.
@stephan-herrmann @srikanth-sankaran

stephan-herrmann:
> > I think the fix is quite safe
> 
> I think the fix is in the right direction - so leaves us better than where we are.
> 
> But I am not sure we are doing the right thing by simply returning at the place we do in this block:

I agree that there is more to be considered, and thanks specifically for new test cases.

Still I'll go with the judgement "leaves us better than where we are", to merge now and handle remaining points as a separete issue.

Issue #4479, opened 2025-10-08
[26] JEP 401: Value Classes and Objects (Preview)


mpalat:
Ref: https://bugs.openjdk.org/browse/JDK-8251554
 https://cr.openjdk.org/~dlsmith/jep401/latest/

mpalat:
Will take it up with the guidance of @srikanth-sankaran. Thanks @srikanth-sankaran for offering to guide

stephan-herrmann:
Caveat, this JEP seems to have recent activity, but **we still do not see it officially proposed for 26**.

So, getting an early start is a great idea, but nothing of this should yet be pushed to BETA_JAVA26.

mpalat:
> Caveat, this JEP seems to have recent activity, but **we still do not see it officially proposed for 26**.
> 
> So, getting an early start is a great idea, but nothing of this should yet be pushed to BETA_JAVA26.


Right. will not be pushed. The PR is flagged as draft - planning to remove that only [if and ] after its targeted

mpalat:
@stephan-herrmann as you hinted, looks like this may not make it to 26 given the (lack) of progress in state. JEP is still in submitted state. Will wait wait for a couple more weeks before I remove the milestone label.

stephan-herrmann:
Not in 26, but I just found [this](https://mail.openjdk.org/pipermail/valhalla-spec-experts/2025-October/002546.html):

> We're happy to report an implementation milestone: an updated EA release of Value Classes and Objects with all core functionality and optimizations implemented. We'll spend the next few months on stabilization and code review as we work towards a preview release in the JDK.

mpalat:
The Java Value types effort will on valtypes branch

mpalat:
Issue created for the corresponding Early Access JVM to be installed - https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/7189

stephan-herrmann:
@mpalat  do you care to explain the branch name `VALCO_PDOT_1`?  When I saw it in today's pull result I was a bit surprised, and didn't quite know what to make of it :)

Should we assume that the previous branch `valtypes-26` is no longer needed and can be removed?

mpalat:
> [@mpalat](https://github.com/mpalat) do you care to explain the branch name `VALCO_PDOT_1`? When I saw it in today's pull result I was a bit surprised, and didn't quite know what to make of it :)

[...]

PR #5350, opened 2026-09-02
Unify the bifurcated constructor body  

### LINKED ISSUE #5349: [Internal] Unify the bifurcated constructor body
Currently, `org.eclipse.jdt.internal.compiler.ast.ConstructorDeclaration` splits the body of the constructor into two holding state variables.

`org.eclipse.jdt.internal.compiler.ast.ConstructorDeclaration.constructorCall` which holds the chaining `ExplicitConstructorCall` (which is a misnomer because it can be an implicit `super()` inserted by the compiler 😆) and `org.eclipse.jdt.internal.compiler.ast.AbstractMethodDeclaration.statements` which holds the rest of the body. If the programmer had supplied an explicit constructor call as the first statement we would move it from the `statements` that constitute the constructor body into the `constructorCall` state variable.

This was all fine once upon a time, but with JEP 513 and JEP 401 things get really complicated.

Here is evolving code from BETA_JAVA28 branch that shows how this bifurcation along with evolving language nuances result in spaghetti code.
See how many invocations of `this.constructorCall.generateCode(...)` and `generateFieldInitializations(...)` need to be sprinkled all over to cover various scenarios.

I believe we could significantly simplify the code and cut back on clutter by unifying constructor body into ole holding state variable `statements`.
The compiler can suitably insert the synthesized `super()` call either at the head or the tail end or not at all as needed.

```
		boolean needFieldInitializations = this.constructorCall == null || this.constructorCall.accessMode != ExplicitConstructorCall.This;

		// Synthetic initializations occur prior to explicit constructor call
		if (needFieldInitializations){
			generateSyntheticFieldInitializationsIfNecessary(this.scope, codeStream, declaringClass);
			codeStream.recordPositionsFrom(0, this.bodyStart > 0 ? this.bodyStart : this.sourceStart);
		}

		if (JavaFeature.FLEXIBLE_CONSTRUCTOR_BODIES.isSupported(this.scope.compilerOptions())) {
			this.scope.enterEarlyConstructionContext();
		}

		// generate constructor call
		if (this.constructorCall != null) {
			if (!declaringClass.isValueClass()) // wait until field initializations for value class
				this.constructorCall.generateCode(this.scope, codeStream);
		}
		final ExplicitConstructorCall lateConstructorCall = getLateConstructorCall();
		// generate field initialization - only if not invoking another constructor call of the same class
		if (needFieldInitializations) {
			if (lateConstructorCall == null) {
				// traditionally field inits are generated before explicit statements
				generateFieldInitializations(declaringType, codeStream, initializerScope, declaringClass.isValueClass() ? FieldsOnly : All);
			}
		}
		if (this.constructorCall != null && !this.constructorCall.isImplicitSuper() && declaringClass.isValueClass()) {
			this.constructorCall.generateCode(this.scope, codeStream);
			generateFieldInitializations(declaringType, codeStream, initializerScope, InitializersOnly);
		}
		// generate statements
		if (this.statements != null) {
			for (Statement statement : this.statements) {
				if (lateConstructorCall == statement && lateConstructorCall.accessMode != ExplicitConstructorCall.This && declaringClass.isValueClass()) {
					// with JEP 401 (value classes) involved field inits are generated *before* chaining to super constructor
					generateFieldInitializations(declaringType, codeStream, initializerScope, FieldsOnly);
				}
				statement.generateCode(this.scope, codeStream);
				if (!this.compilationResult.hasErrors() && (codeStream.stackDepth != 0 || codeStream.operandStack.size() != 0)) {
					this.scope.problemReporter().operandStackSizeInappropriate(this);
				}
				if (lateConstructorCall == statement && lateConstructorCall.accessMode != ExplicitConstructorCall.This) {
					// with JEP 492 (Flexible Constructor Bodies) involved field inits are generated only *after* the explicit constructor for identity classes
					generateFieldInitializations(declaringType, codeStream, initializerScope, declaringClass.isValueClass() ? InitializersOnly : All);
				}
			}
		}
		// if a problem got reported during code gen, then trigger problem method creation
		if (this.ignoreFurtherInvestigation) {
			throw new AbortMethod(this.scope.referenceCompilationUnit().compilationResult, null);
		}
		if ((this.bits & ASTNode.NeedFreeReturn) != 0) {
			if (this.isCompactConstructor()) {
				// Note: the body of a compact constructor may not contain a return statement and so will need an injected return
				for (RecordComponent rc : classScope.referenceContext.recordComponents) {
					LocalVariableBinding parameter = this.scope.findVariable(rc.name);
					FieldBinding field = classScope.referenceContext.binding.getField(rc.name, true).original();
					codeStream.aload_0();
					codeStream.load(parameter);
					codeStream.fieldAccess(Opcodes.OPC_putfield, field, classScope.referenceContext.binding);
				}
			}
			if (this.constructorCall != null && this.constructorCall.isImplicitSuper() && declaringClass.isValueClass()) {
				this.constructorCall.generateCode(this.scope, codeStream);
				generateFieldInitializations(declaringType, codeStream, initializerScope, InitializersOnly);
			}
			codeStream.return_();
		}
```

### LINKED ISSUE #5352: Unused flexible constructor not complained about
Found during code review & white box testing:

The following program generates a `The constructor X(int, int) is never used locally` warning correctly, but no warning is issued if the syso statement is uncommented making the constructor a flexible constructor.

```
public class X {
	
	X(String s) {
		
	}
	
	private X(int x, int y) {
		//System.out.println("Hello"); // uncomment and the warning goes away on this unused constructor
		this("Hello");
	}
	void foo() {
		System.out.println("X::foo");
	}
	
		new X("Hello") {
			@Override
			void foo() {
				System.out.println("Anon::foo");
			}
		}.foo();
	}
}
```

### LINKED ISSUE #5353: Flexible recursive constructors tolerated at compile time lead to StackOverflowError at runtime.
Found by code inspection and white box testing:

The following program compiles with ECJ master/HEAD while is correctly rejected by javac with `X.java:10: error: recursive constructor invocation`

```
public class X {
	
    X(String s) {
    	System.out.println("delegating");
		this(10, 20);
	}
	
	public X(int x, int y) {
		System.out.println("delegating");
		this("xxx");
	}
	
		new X("Hello");
	}
} 
```

ECJ generated program crashes at runtime with a StackOverflowError

### LINKED ISSUE #5358: Explicit constructor call from flexible constructor missing in search for references to the chained constructor
Test case:

```
class A {
    A() {}
    A(int x) {}  
}

class B extends A {
    B() {
    	System.out.println();
        this(10);   // constructor call #1
    }
    B(int n) {
    	System.out.println();
        super(n);   // constructor call #2
    }
}
```

If you search for references to A(int) with `Ctrl+Shift+G` then constructor call #2 does not show up. Without the preceding sysout call it does.

### LINKED ISSUE #5360: ECJ generated code may write to final fields in inner classes with flexible constructors more than once
Given the following program:

```
public class X {
	public class Inner {
		Inner() {
			System.out.println("Chaining to other constructor");
			this("chained");
		}
		
		Inner(String s) {		
		}
	}
	
		new X().new Inner();
	}
} 
```

on master/HEAD we generate the following:

```
  final X this$0;
    descriptor: LX;
    flags: (0x1010) ACC_FINAL, ACC_SYNTHETIC

  X$Inner(X);
    descriptor: (LX;)V
    flags: (0x0000)
    Code:
      stack=3, locals=2, args_size=2
         0: aload_0
         1: aload_1
         **2: putfield      #10                 // Field this$0:LX;**
         5: getstatic     #12                 // Field java/lang/System.out:Ljava/io/PrintStream;
         8: ldc           #18                 // String Chaining to other constructor
        10: invokevirtual #20                 // Method java/io/PrintStream.println:(Ljava/lang/String;)V
        13: aload_0
        14: aload_1
        15: ldc           #26                 // String chained
        17: invokespecial #28                 // Method "<init>":(LX;Ljava/lang/String;)V
        20: return
      LineNumberTable:
        line 3: 0
        line 4: 5
        line 5: 13
        line 6: 20
      LocalVariableTable:
        Start  Length  Slot  Name   Signature
            0      21     0  this   LX$Inner;

  X$Inner(X, java.lang.String);
    descriptor: (LX;Ljava/lang/String;)V
    flags: (0x0000)
    Code:
      stack=2, locals=3, args_size=3
         0: aload_0
         1: aload_1
         2: putfield      #10                 // Field this$0:LX;
         5: aload_0
         6: invokespecial #35                 // Method java/lang/Object."<init>":()V
         9: return
```

See that the first constructor should not emit the `putfield` to `Field this$0` as it chains to the second which will do a `putfield` to this field anyway.

Per my reading of JVMS 6.5 this may not be an error, but is merely wasteful. If you remove the syso preceding the `this(...)` call, we don't emit the putfield in the first inner constructor.

srikanth-sankaran:
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5349
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5353
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5352
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5358
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5360

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
@@ -2827,63 +2827,30 @@ protected void consumeConstructorDeclaration() {
 
 	//statements
 	this.realBlockPtr--;
-	ExplicitConstructorCall constructorCall = null;
-	Statement[] statements = null;
-	if ((length = this.astLengthStack[this.astLengthPtr--]) != 0) {
-		this.astPtr -= length;
-		if (!this.options.ignoreMethodBodies) {
-			if (this.astStack[this.astPtr + 1] instanceof ExplicitConstructorCall explicitCall) {
-				//avoid a isSomeThing that would only be used here BUT what is faster between two alternatives ?
-				System.arraycopy(
-					this.astStack,
-					this.astPtr + 2,
-					statements = new Statement[length - 1],
-					0,
-					length - 1);
-				constructorCall = explicitCall;
-			} else { //need to add explicitly the super();
-				System.arraycopy(
-					this.astStack,
-					this.astPtr + 1,
-					statements = new Statement[length],
-					0,
-					length);
-				constructorCall = SuperReference.implicitSuperConstructorCall();
-			}
-		}
-	} else {
-		boolean insideFieldInitializer = false;
-		if (this.diet) {
-			for (int i = this.nestedType; i > 0; i--){
-				if (this.variablesCounter[i] > 0) {
-					insideFieldInitializer = true;
-					break;
-				}
-			}
-		}
-
-		if (!this.options.ignoreMethodBodies) {
-			if (!this.diet || insideFieldInitializer){
-				// add it only in non-diet mode, if diet_bodies, then constructor call will be added elsewhere.
-				constructorCall = SuperReference.implicitSuperConstructorCall();
-			}
-		}
-	}
+	boolean insideFieldInitializer = false;
+	length = this.astLengthStack[this.astLengthPtr--];
+	this.astPtr -= length;
 
 	// now we know that the top of stack is a constructorDeclaration
 	ConstructorDeclaration cd = (ConstructorDeclaration) this.astStack[this.astPtr];
-	cd.constructorCall = constructorCall;
-	cd.statements = statements;
-
-	//highlight of the implicit call on the method name
-	if (constructorCall != null && cd.constructorCall.sourceEnd == 0) {
-		cd.constructorCall.sourceEnd = cd.sourceEnd;
-		cd.constructorCall.sourceStart = cd.sourceStart;
+	if (!this.options.ignoreMethodBodies) {
+		if (length == 0) {
+			if (this.diet) {
+				for (int i = this.nestedType; i > 0; i--){
+					if (this.variablesCounter[i] > 0) {
+						insideFieldInitializer = true;
+						break;
+					}
+				}
+			}
+		}
+		if (length > 0 || !this.diet || insideFieldInitializer) // if diet mode then constructor body will be added elsewhere.
+			cd.buildBody(this.astStack, this.astPtr + 1, length, this.options);
 	}
 
 	if (!(this.diet && this.dietInt == 0)
-			&& statements == null
-			&& (constructorCall == null || constructorCall.isImplicitSuper())
+			&& cd.statements != null
+			&& (cd.statements.length == 0 || (cd.statements.length == 1 && cd.statements[0].isImplicitSuper()))
 			&& !containsComment(cd.bodyStart, this.endPosition)) {
 		cd.bits |= ASTNode.UndocumentedEmptyBlock;
 	}
@@ -10252,7 +10219,16 @@ public MethodDeclaration convertToMethodDeclaration(ConstructorDeclaration c, Co
 	m.declarationSourceEnd = c.declarationSourceEnd;
 	m.declarationSourceStart = c.declarationSourceStart;
 	m.selector = c.selector;
-	m.statements = c.statements;
+	int length = c.statements != null ? c.statements.length : 0;
+	if (length > 0) {
+		if (c.statements[0] instanceof ExplicitConstructorCall) {
+			System.arraycopy(c.statements, 1, m.statements = new Statement[length - 1], 0, length - 1);
+		} else if (c.statements[length - 1] instanceof ExplicitConstructorCall) {
+			System.arraycopy(c.statements, 0, m.statements = new Statement[length - 1], 0, length - 1);
+		}
+	} else {
+		m.statements = c.statements;
+	}
 	m.modifiers = c.modifiers;
 	m.annotations = c.annotations;
 	m.arguments = c.arguments;
@@ -11827,44 +11803,13 @@ public void parse(ConstructorDeclaration cd, CompilationUnitDeclaration unit, bo
 
 	//statements
 	cd.explicitDeclarations = this.realBlockStack[this.realBlockPtr--];
-	int length;
-	if (this.astLengthPtr > -1 && (length = this.astLengthStack[this.astLengthPtr--]) != 0) {
-		this.astPtr -= length;
-		if (!this.options.ignoreMethodBodies) {
-			if (this.astStack[this.astPtr + 1] instanceof ExplicitConstructorCall explicitCall)
-				//avoid a isSomeThing that would only be used here BUT what is faster between two alternatives ?
-				{
-				System.arraycopy(
-					this.astStack,
-					this.astPtr + 2,
-					cd.statements = new Statement[length - 1],
-					0,
-					length - 1);
-				cd.constructorCall = explicitCall;
-			} else { //need to add explicitly the super();
-				System.arraycopy(
-					this.astStack,
-					this.astPtr + 1,
-					cd.statements = new Statement[length],
-					0,
-					length);
-				cd.constructorCall = SuperReference.implicitSuperConstructorCall();
-			}
-		}
-	} else {
-		if (!this.options.ignoreMethodBodies) {
-			cd.constructorCall = SuperReference.implicitSuperConstructorCall();
-		}
-		if (!containsComment(cd.bodyStart, cd.bodyEnd)) {
-			cd.bits |= ASTNode.UndocumentedEmptyBlock;
-		}
+	int length = this.astLengthPtr > -1 ? this.astLengthStack[this.astLengthPtr--] : 0;
+	this.astPtr -= length;
+	if (!this.options.ignoreMethodBodies) {
+		cd.buildBody(this.astStack, this.astPtr + 1, length, this.options);
 	}
-
-	ExplicitConstructorCall explicitConstructorCall = cd.constructorCall;
-	if (explicitConstructorCall != null && explicitConstructorCall.sourceEnd == 0) {
-		explicitConstructorCall.sourceEnd = cd.sourceEnd;
-		explicitConstructorCall.sourceStart = cd.sourceStart;
-	}
+	if (length == 0 && !containsComment(cd.bodyStart, cd.bodyEnd))
+		cd.bits |= ASTNode.UndocumentedEmptyBlock;
 }
 // A P I
 
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ASTNode.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ASTNode.java
@@ -318,6 +318,7 @@ public abstract class ASTNode implements Location, TypeConstants, TypeIds {
 	// for all reference context entries.
 	public static final int HasFunctionalInterfaceTypes = ASTNode.Bit22;
 
+	public static final Statement [] NO_STATEMENTS = new Statement [0];
 	public static final FieldDeclaration [] NO_FIELD_DECLARATIONS = new FieldDeclaration [0];
 	public static final Argument [] NO_ARGUMENTS = new Argument [0];
 	public static final RecordComponent [] NO_RECORD_COMPONENTS = new RecordComponent [0];
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/AbstractMethodDeclaration.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/AbstractMethodDeclaration.java
@@ -727,7 +727,9 @@ public abstract class AbstractMethodDeclaration
 
 		if (this.statements != null) {
 			resolveStatements(this.statements, this.scope);
-		} else if ((this.bits & UndocumentedEmptyBlock) != 0) {
+		}
+		// An empty constructor gets injected with a super() call, so don't rely on this.statements == null being true for an empty block!
+		if ((this.bits & UndocumentedEmptyBlock) != 0) {
 			if (!this.isConstructor() || this.arguments != null) { // https://bugs.eclipse.org/bugs/show_bug.cgi?id=319626
 				this.scope.problemReporter().undocumentedEmptyBlock(this.bodyStart-1, this.bodyEnd+1);
 			}
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ConstructorDeclaration.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ConstructorDeclaration.java
@@ -60,8 +60,6 @@ public class ConstructorDeclaration extends AbstractMethodDeclaration {
 
 	public static final UnconditionalFlowInfo EMPTY_FLOW_INFO = new UnconditionalFlowInfo();
 
-	public ExplicitConstructorCall constructorCall;
-
 	public TypeParameter[] typeParameters;
 
 	private ExceptionHandlingFlowContext prologueContext;
@@ -101,7 +99,8 @@ public void analyseCode(ClassScope classScope, InitializationFlowContext initial
 
 	try {
 		ExplicitConstructorCall lateConstructorCall = getLateConstructorCall();
-		boolean hasArgumentNeedingAnalysis = this.constructorCall != null && this.constructorCall.hasArgumentNeedingAnalysis();
+		ExplicitConstructorCall earlyConstructorCall = getEarlyConstructorCall();
+		boolean hasArgumentNeedingAnalysis = earlyConstructorCall != null && earlyConstructorCall.hasArgumentNeedingAnalysis();
 		if (mode == AnalysisMode.PROLOGUE
 				&& lateConstructorCall == null
 				&& (!hasArgumentNeedingAnalysis)) {
@@ -131,14 +130,15 @@ public void analyseCode(ClassScope classScope, InitializationFlowContext initial
 		 		}
 				// https://bugs.eclipse.org/bugs/show_bug.cgi?id=270446, When the AST built is an abridged version
 				// we don't have all tree nodes we would otherwise expect. (see ASTParser.setFocalPosition)
-				if (this.constructorCall == null)
+				ExplicitConstructorCall constructorCall = getConstructorCall();
+				if (constructorCall == null)
 					break checkUnused;
 				// https://bugs.eclipse.org/bugs/show_bug.cgi?id=264991, Don't complain about this
 				// constructor being unused if the base class doesn't have a no-arg constructor.
 				// See that a seemingly unused constructor that chains to another constructor with a
 				// this(...) can be flagged as being unused without hesitation.
 				// https://bugs.eclipse.org/bugs/show_bug.cgi?id=265142
-				if (this.constructorCall.accessMode != ExplicitConstructorCall.This) {
+				if (constructorCall.accessMode != ExplicitConstructorCall.This) {
 					ReferenceBinding superClass = constructorBinding.declaringClass.superclass();
 					if (superClass == null)
 						break checkUnused;
@@ -161,7 +161,7 @@ public void analyseCode(ClassScope classScope, InitializationFlowContext initial
 
 			// check constructor recursion, once all constructor got resolved
 			if (isRecursive(null /*lazy initialized visited list*/)) {
-				this.scope.problemReporter().recursiveConstructorInvocation(this.constructorCall);
+				this.scope.problemReporter().recursiveConstructorInvocation(getConstructorCall());
 			}
 			// https://bugs.eclipse.org/bugs/show_bug.cgi?id=385780
 			if (this.typeParameters != null  &&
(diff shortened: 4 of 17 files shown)


srikanth-sankaran:
@stephan-herrmann - you may want to glance through the changes here. This is in preparation for value types work. 

ATM, I have gone to some lengths to keep constructor flow analysis unmodified. I may attempt simplifying it in a future separate PR.

I expect to integrate this today or  tomorrow.

srikanth-sankaran:
> I have gone to some lengths to keep constructor flow analysis unmodified

This is the reason there are now three get methods:

`org.eclipse.jdt.internal.compiler.ast.ConstructorDeclaration.getConstructorCall()`
`org.eclipse.jdt.internal.compiler.ast.ConstructorDeclaration.getEarlyConstructorCall()` and
`org.eclipse.jdt.internal.compiler.ast.ConstructorDeclaration.getLateConstructorCall()`

I expect to prune these in a future PR.

stephan-herrmann:
A very quick response without having looked through all details of this change:

I agree that field `constructorCall` looks from a different era, inlining this into `statements` makes a lot of sense.

Concerning the spaghetti accumulating from different language changes I wonder, if things would be easier if the core part of `internalGenerateCode` would be split into separate methods, like: one method for compact record constructors, one method for constructors of value classes. I'm not sure about the difference between regular constructors with vs without a late constructor call, perhaps those are well served by just one method. In particular the way how constructors in value classes use synthetic locals to defer all field assignments to the end suggests that this is a fundamentally new concept, not just a mild variation of regular constructors.

@srikanth-sankaran WDYT?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
```diff
@@ -2806,6 +2807,29 @@ protected void consumeConstructorBody() {
...
 }
+public static final void attachConstructorBody(ConstructorDeclaration cd, ASTNode [] astStack, int astPtr, int length, /* @Nullable */ProblemReporter problemReporter) {
+	ArrayDeque<Statement> statements = new ArrayDeque<>(length + 1);
+	boolean needImplicitSuperCall = true;
+	boolean superCallPrecedes = true;
+	Statement statement;
+	for (int i = 0; i < length; i++) {
+	    statement = (Statement) astStack[astPtr + i];
+	    if (statement instanceof ExplicitConstructorCall)
+	    	needImplicitSuperCall = i != 0 && problemReporter != null && !JavaFeature.FLEXIBLE_CONSTRUCTOR_BODIES.isSupported(problemReporter.options);
```
stephan-herrmann: dependence on `problemReporter` looks funny. What's wrong with `this.options`?

srikanth-sankaran:
> if things would be easier if the core part of `internalGenerateCode` would be split into separate methods

Thanks, I will give it some thought (although if I decide to do it, I will defer it to a subsequent phase as I am trying to wrap up JEP401 implementation. The first 90% of the work took 90% of the time and the last 10% of the work I would like not to take 90% of the time 😆 )

FWIW, I did something like that with `org.eclipse.jdt.internal.compiler.ast.SwitchStatement` with 

`SwitchTranslator` being the abstract base class and `ClassicSwitchTranslator`, `ClassicEnumSwitchTranslator`, `PatternSwitchTranslator` and `StringSwitchTranslator` specializing behavior. 

Let us see.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
srikanth-sankaran: > dependence on `problemReporter` looks funny. What's wrong with `this.options`?

Yes, that would be better. Planning to move this method to ConstructorDeclaration as an instance method with some simplifications - as suggested by co-pilot, the ArrayDeque route is not really required for master. Even for JEP401 where I need to append the synthesized call at the tail, there could be faster paths that avoid it.

srikanth-sankaran: Done

srikanth-sankaran:
I still have to self-review the following files:


- org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ConstructorDeclaration.java
- org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeDeclaration.java
- org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/parser/Parser.java
- org.eclipse.jdt.core/model/org/eclipse/jdt/internal/compiler/SourceElementNotifier.java
- org.eclipse.jdt.core/search/org/eclipse/jdt/internal/core/search/matching/ConstructorLocator.java


(I always hated this `org.eclipse.jdt.internal.compiler.parser.Parser.convertToMethodDeclaration(ConstructorDeclaration, CompilationResult)` May be time to get rid of that - but first I need to understand what motivated it in the first place)

srikanth-sankaran:
> * org.eclipse.jdt.core/model/org/eclipse/jdt/internal/compiler/SourceElementNotifier.java

Any direct reference to `constructorCall` is guaranteed to be a defect in some scenarios as in a flexible constructor this would be null. Unless we reach into the late constructor, we will miss "processing" the constructor invocation. So in this block of code we have a bug lurking! However, I don't know how to come up with a scenario to reproduce. The proposed PR fixes this and that ought to be enough satisfaction :)

```
         if (this.reportReferenceInfo) {
			ConstructorDeclaration constructorDeclaration = (ConstructorDeclaration) methodDeclaration;
			ExplicitConstructorCall constructorCall = constructorDeclaration.constructorCall;
			if (constructorCall != null) {
				switch(constructorCall.accessMode) {
					case ExplicitConstructorCall.This :
						this.requestor.acceptConstructorReference(
							this.typeNames[this.nestedTypeIndex-1],
							constructorCall.arguments == null ? 0 : constructorCall.arguments.length,
							constructorCall.sourceStart);
						break;
					case ExplicitConstructorCall.Super :
					case ExplicitConstructorCall.ImplicitSuper :
						this.requestor.acceptConstructorReference(
							this.superTypeNames[this.nestedTypeIndex-1],
							constructorCall.arguments == null ? 0 : constructorCall.arguments.length,
							constructorCall.sourceStart);
						break;
				}
			}
		}
```

stephan-herrmann:
> (I always hated this `org.eclipse.jdt.internal.compiler.parser.Parser.convertToMethodDeclaration(ConstructorDeclaration, CompilationResult)` May be time to get rid of that - but first I need to understand what motivated it in the first place)

Does this comment help? https://github.com/eclipse-jdt/eclipse.jdt.core/blob/e81e2252e115745e41997eb3942efdd39d373a68/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeDeclaration.java#L335-L341

srikanth-sankaran:
> > (I always hated this `org.eclipse.jdt.internal.compiler.parser.Parser.convertToMethodDeclaration(ConstructorDeclaration, CompilationResult)` May be time to get rid of that - but first I need to understand what motivated it in the first place)
> 
> Does this comment help?

I am saying one way or other we are going to have reject it, why go through some work before complaining about missing return. Why not complain about malformed constructor right away and not bother with transforming it into a malformed method ?

srikanth-sankaran:
> However, I don't know how to come up with a scenario to reproduce

If `org.eclipse.jdt.core.tests.model.JavaSearchTests.testConstructorReference04()` is modified so it uses flexible constructors then it would trigger a bug on master

srikanth-sankaran:
> > However, I don't know how to come up with a scenario to reproduce
> 
> If `org.eclipse.jdt.core.tests.model.JavaSearchTests.testConstructorReference04()` is modified so it uses flexible constructors then it would trigger a bug on master

Co-pilot to the rescue: https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5358

srikanth-sankaran:
> Any direct reference to `constructorCall` is guaranteed to be a defect in some scenarios

I think I have gone through every such use and we are covered. (Question with the patch is whether this.constructorCall should be replaced with `getEarlyConstructorCall()` or `getConstructorCall()`- I think this has been carefully addressed)

srikanth-sankaran:
@iloveeclipse - this is a non-trivial rewrite of how constructor bodies are encoded and processed. I expect no perturbance, but it may still be a good idea to trigger an early build to ensure any issues are addressed in time for tonight's scheduled I-build

iloveeclipse:
@srikanth-sankaran : I will merge https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/4023 first and trigger SDK build afterwards, otherwise we will not see test results :-)

@akurtakov : I see you have some Java 25 related work in different repos. Should I wait for something or can start SDK right away after https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/4023 merge?

akurtakov:
@iloveeclipse My changes are for various repos verification builds - nothing pending that should affect I-build right now.

iloveeclipse:
@srikanth-sankaran : new SDK build is there, has comparator errors, but they are unrelated to this change. So we are fine here.

See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/issues/4025

srikanth-sankaran:
> So we are fine here.

Thanks, I'll check back the test results later today at https://download.eclipse.org/eclipse/downloads/drops4/I20260904-0846/#tests

iloveeclipse:
Related tests are green.

srikanth-sankaran:
> In particular the way how constructors in value classes use synthetic locals to defer all field assignments to the end suggests that this is a fundamentally new concept,

Thanks for the details in https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4479#issuecomment-5543813688 

Yeah, this is something I hadn't thought about as it falls under the section Safe construction of value objects which I have called out is the one area our implementation has gaps (see https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4479#issuecomment-5452267984)

I have some idea how this can be handled seamlessly: the compiler will issue field write/reads but the field will mutate them to local proxy read/ writes as/when needed. 

Let us see 😊

srikanth-sankaran:
> I have some idea how this can be handled seamlessly

I have a tentative name for the new abstraction BipolarFieldBinding (SchizophrenicFieldBinding ?) 😊

Once outside the early construction context the FieldBinding is "cured" of its split personality disorder! 🤣

To 99.99% of the compiler it is just a FieldBinding

stephan-herrmann:
> > I have some idea how this can be handled seamlessly
> 
> I have a tentative name for the new abstraction BipolarFieldBinding (SchizophrenicFieldBinding ?) 😊
> 
> Once outside the early construction context the FieldBinding is "cured" of its split personality disorder! 🤣
> 
> To 99.99% of the compiler it is just a FieldBinding

:)

Don't SyntheticFieldBinding / SyntheticMethodBinding already go in this direction? E.g., when any access to an outer field may be redirected to a SyntheticMethodBinding which knows what byte codes should represent a source-level field access? In the new case it would be redirecting field access to local var access. These synthetics have the advantage that the pure field binding is not adulterated in any form.

I imagine a tricky part here is: when do we detect the need for "proxying" field access in early construction context? Perhaps it's enough if the analysis phase detects the situation, remembers this in the scope, so that generateCode can introduce such proxy exactly where needed, plus add a "flush" write at the end of early construction. I.e., I'm proposing for this case to let early construction operate on the proxy local var exclusively and indeed defer write access to the point where the object leaves the larval state.

Point to emphasize: I do think that the (method-) **scope** should be involved, because the need for proxying may be specific to a particular constructor, but not to others.

srikanth-sankaran:
> Don't SyntheticFieldBinding / SyntheticMethodBinding already go in this direction? E.g., when any access to an outer field may be redirected to a SyntheticMethodBinding which knows what byte codes should represent a source-level field access? In the new case it would be redirecting field access to local var access. These synthetics have the advantage that the pure field binding is not adulterated in any form.

We won't adulterate the pure field binding. We would subclass it and specialize some behavior. 


> 
> I imagine a tricky part here is: when do we detect the need for "proxying" field access in early construction context? 

I think we can/should do this while resolving constructor body itself. When synthetic local variable requirement is deduced in resolution phase itself things are much simpler. I recall some other situation where we could determine this only much later requiring us to adjust local variable offsets and such. A situation best avoided if possible. 


> Point to emphasize: I do think that the (method-) **scope** should be involved, because the need for proxying may be specific to a particular constructor, but not to others.

Yes. May be specific only to a particular constructor and may be only specific to some fields.

stephan-herrmann:
> > Don't SyntheticFieldBinding / SyntheticMethodBinding already go in this direction? E.g., when any access to an outer field may be redirected to a SyntheticMethodBinding which knows what byte codes should represent a source-level field access? In the new case it would be redirecting field access to local var access. These synthetics have the advantage that the pure field binding is not adulterated in any form.
> 
> We won't adulterate the pure field binding. We would subclass it and specialize some behavior.

**Class** FieldBinding will be unaffected, good, but the same **instance** of FieldBindingInEarlyConstructionContext will need to operate differently depending on context, right? With a synthetic proxy you have both behaviors in separate instances which might reduce the number of spaghetti-ifs. You can also create the proxy after the fact. Otherwise we'd have to decide which kind of binding to create already before the resolve phase, viz. during BUILD_FIELDS_AND_METHODS.

(_transparency notice: with all the work on role objects in ObjectTeams I have a bias towards coding contextual behavior as separate instances_ :) )
 
> > I imagine a tricky part here is: when do we detect the need for "proxying" field access in early construction context?
> 
> I think we can/should do this while resolving constructor body itself.

You're right. I was thinking we need flow to know about early construction context, but `isInsideEarlyConstructionContext()` is already heavily used during resolve phase. 

> > Point to emphasize: I do think that the (method-) **scope** should be involved, because the need for proxying may be specific to a particular constructor, but not to others.
> 
> Yes. May be specific only to a particular constructor and may be only specific to some fields.

exactly

srikanth-sankaran:
> Class FieldBinding will be unaffected, good, but the same instance of FieldBindingInEarlyConstructionContext will need to operate differently depending on context, right?

I am not sure I understand - that is the essence of inheritance and overriding and mutable objects - isn't it ?

> Otherwise we'd have to decide which kind of binding to create already before the resolve phase, viz. during BUILD_FIELDS_AND_METHODS.

Yes, this is a niggle. We may have to opt for the subclass instance defensively even in cases it may not be needed. Not most elegant.

The new abstraction differs from vanilla FieldBinding only in two ways: in name and in tracking a doppelgänger synthetic local.

There is a clean alternative that avoids a new abstraction. A Map in constructor's method scope with FieldBinding as the key and SyntheticLocalVariableBinding as the mapped value. 

I am not hung up on a new abstraction because of having to eagerly build fields ahead of knowing exact needs. I like the cute name possibilities but that can't be driving design choices 😊

What I am keen on is for most of the compiler to think it is working with a FieldBinding and for the switch to happen seamlessly automagically as needed. That would be one 'if' buried deep down. 

At the moment I don't have my laptop. So this feels like being a blind man and having to describe an elephant. 

Let us see.

stephan-herrmann:
> > Class FieldBinding will be unaffected, good, but the same instance of FieldBindingInEarlyConstructionContext will need to operate differently depending on context, right?
> 
> I am not sure I understand - that is the essence of inheritance and overriding and mutable objects - isn't it ?

Behavior depending on an object's own state is normal for mutable objects, right. But behavior depending on context outside the object (e.g., "who is calling me?", "are we in code gen for an early construction context?") is not directly supported in Java et al.

Let's consider `VariableBinding.kind()` as an example and assume the schizo field needs to answer FIELD _sometimes_ and LOCAL _at other times_. To answer "what is my kind right now?"  this class needs to learn how to distinguish between different modes all by itself. If on top of that several code locations need to make the same distinction, then we end up with many locations  saying "if in fieldish-context doThis else doThat". 

If, OTOH, we have separate instances (a) FieldBinding and (b) SyntheticFieldBinding, then it _might_ suffice if generateCode only once says `variableBinding = scope.getEmulationIfNecessary(variableBinding)`, and this statement will answer a binding that unambiguously knows how to answer `kind()` and many other queries. When leaving the early construction context of a given constructor, saying `variableBinding = variableBinding.getOriginal()` would switch back.

Subtyping is good for one-time selection of behavior: at the time of instantiation. For any behavior changes after instantiation you need ifs.

srikanth-sankaran:
> > > Class FieldBinding will be unaffected, good, but the same instance of FieldBindingInEarlyConstructionContext will need to operate differently depending on context, right?
> > 
> > 
> > I am not sure I understand - that is the essence of inheritance and overriding and mutable objects - isn't it ?
> 
> Behavior depending on an object's own state is normal for mutable objects, right. But behavior depending on context outside the object (e.g., "who is calling me?", "are we in code gen for an early construction context?") is not directly supported in Java et al.
> 
> Let's consider `VariableBinding.kind()` as an example and assume the schizo field needs to answer FIELD _sometimes_ and LOCAL _at other times_. To answer "what is my kind right now?" this class needs to learn how to distinguish between different modes all by itself. If on top of that several code locations need to make the same distinction, then we end up with many locations saying "if in fieldish-context doThis else doThat".
> 
> If, OTOH, we have separate instances (a) FieldBinding and (b) SyntheticFieldBinding, then it _might_ suffice if generateCode only once says `variableBinding = scope.getEmulationIfNecessary(variableBinding)`, and this statement will answer a binding that unambiguously knows how to answer `kind()` and many other queries. When leaving the early construction context of a given constructor, saying `variableBinding = variableBinding.getOriginal()` would switch back.
> 
> Subtyping is good for one-time selection of behavior: at the time of instantiation. For any behavior changes after instantiation you need ifs.

Thanks @stephan-herrmann - I'll see what I can do. There is some more work on master before I will attempt it on JAVA28 branch. You will certainly be invited to comment before that PR gets merged;

Issue #2811, opened 2024-08-12
[Bug - Autocompletion] - Does not work with Record implements Interface


mamilic:
When the autocompletion is supposed to kick in within record that implements interface, there is nothing.

##### Steps To Reproduce
1. Open Project 
[test-vs-code.zip](https://github.com/redhat-developer/vscode-java/files/15219960/test-vs-code.zip)
2. Go to Post.java
3. Try to autocomplete List

<details><summary>Server Logs</summary>
<p>

Cannot read the array length because "superPermittedTypes" is null
java.lang.NullPointerException: Cannot read the array length because "superPermittedTypes" is null
	at org.eclipse.jdt.internal.compiler.lookup.SourceTypeBinding.checkPermitsAndAdd(SourceTypeBinding.java:1346)
	at org.eclipse.jdt.internal.compiler.lookup.SourceTypeBinding.checkPermitsInType(SourceTypeBinding.java:1207)
	at org.eclipse.jdt.internal.compiler.lookup.SourceTypeBinding.faultInTypesForFieldsAndMethods(SourceTypeBinding.java:1116)
	at org.eclipse.jdt.internal.compiler.lookup.CompilationUnitScope.faultInTypes(CompilationUnitScope.java:608)
	at org.eclipse.jdt.internal.codeassist.CompletionEngine.complete(CompletionEngine.java:2329)
	at org.eclipse.jdt.internal.core.Openable.codeComplete(Openable.java:135)
	at org.eclipse.jdt.internal.core.CompilationUnit.codeComplete(CompilationUnit.java:360)
	at org.eclipse.jdt.internal.core.CompilationUnit.codeComplete(CompilationUnit.java:350)
	at org.eclipse.jdt.ls.core.internal.handlers.CompletionHandler.computeContentAssist(CompletionHandler.java:287)
	at org.eclipse.jdt.ls.core.internal.handlers.CompletionHandler.completion(CompletionHandler.java:121)
	at org.eclipse.jdt.ls.core.internal.handlers.JDTLanguageServer.completion(JDTLanguageServer.java:633)
	at jdk.internal.reflect.GeneratedMethodAccessor31.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(Unknown Source)
	at java.base/java.lang.reflect.Method.invoke(Unknown Source)
[...]

Issue #1445, opened 2023-09-28
Cleanup with Java 21 support removes wrong code


frankbenoit:
With installed Java 21 in Eclipse, but in a project using Java 17 only
````
public class EBug {
    
    class A{
        
    }
    class A1 extends A {
        
    }
    class A2 extends A {
        
    }
    class A3 extends A {
        
    }
    
    public static void main(String[] args) {
        Object o = "";
        if( o instanceof A1 a1) {
            
        } else if( o instanceof A2 a1) {
            
        } else if( o instanceof A3 a1) {
            
        } else  {
            
        }
    }
}

````
Calling cleanup, afterwards the 2nd case is gone:
````
public class EBug {

    class A {

    }

    class A1 extends A {

    }

    class A2 extends A {

    }

    class A3 extends A {

    }

    public static void main(String[] args) {
        final Object o = "";
        if (o instanceof final A1 a1) {

        } else if (o instanceof final A3 a1) {

        } else {

        }
    }
}


````

PR #5364, opened 2026-09-05
Prevent stale Java options cache publication after concurrent updates

carstenartur:
## What it does

Prevents an in-flight `JavaModelManager.getOptions()` call from publishing an obsolete options map after a concurrent update or cache invalidation has completed.

On a cache miss, the existing implementation reads preferences and then unconditionally stores the resulting map in `optionsCache`. This permits the following ordering:

```text
Reader A starts building a map and reads an old preference value.
Writer B updates the preferences and updates or invalidates the cache.
Writer B returns.
Reader A publishes its older map.
A subsequent reader receives the obsolete value from the cache.
```

The problem is the stale value returned to a subsequent reader after the concurrent operations have finished. An old value observed by the overlapping reader itself is not the failure being addressed. Defensive copies protect the cached map against caller modifications, but do not prevent this publication ordering.

## Background and scope

Related debugging was already documented by @trancexpress in [JDT UI #2811](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2811#issuecomment-3967356378): resuming an indexing thread allowed it to publish an options map without the field-prefix preference previously set by a test. That investigation is relevant prior work for the cache-publication problem addressed here.

The investigation also relates to [JDT UI #1445](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1445), which reports sporadic save-participant formatting failures. This PR does not claim to reproduce or resolve all symptoms of that issue, including the additional space before `=` and the reported `MalformedTreeException`.

[Bug 563771](https://bugs.eclipse.org/bugs/show_bug.cgi?id=563771) is included as related background from the [performance and options-copying discussion in #1445](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1445#issuecomment-2610200744). This PR does not replace the existing defensive copies, introduce a copy-on-write map, or claim to resolve that performance topic.

## Implementation

A reader captures the cache generation before reading the cache or preferences. It may publish a newly computed map only if the generation is still unchanged.

Writer updates and invalidations advance the generation under the same private lock used for the compare-and-publish operation. Invalidating an already-null cache also advances the generation, because an older computation may still be in progress. `setOptions(null)` invalidates again before its final rebuild so that a computation started before the reset cannot subsequently become the cached result.

The lock covers only cache publication and invalidation. Preference access, listener callbacks and map copying remain outside it. Cache hits do not acquire the new lock, and callers continue to receive independent, modifiable maps. This avoids serializing the entire options lookup or holding the new lock across preference callbacks.

The PR also addresses a separate invalidation gap when the instance preference node is removed. The replacement node receives the cache-invalidation listener, the stored instance-listener reference is updated, and the cache is invalidated. Two additional tests cover removal and subsequent changes on replacement nodes.

Production changes are limited to `JavaModelManager`. The PR also adds `OptionCacheTests` and registers it in `AllJavaModelTests`. There are no public API, formatter, editor, or build-configuration changes.

## How to test

Run `org.eclipse.jdt.core.tests.model.OptionCacheTests` as a **JUnit Plug-in Test**, using a disposable runtime workspace. It requires an Eclipse runtime and is also registered in `AllJavaModelTests`.

The current class contains **10 tests**:

| Tests | Coverage |
| ---: | --- |
| 4 | Cache rebuilding versus `setOptions()`, direct preference invalidation, repeated invalidation, and `setOptions(null)` |
| 2 | Instance preference-node removal and continued invalidation after node replacement |
| 4 | Independent returned maps, sequential updates, cache reuse, and a preference listener that reenters `getOptions()` |

The race tests use bounded `CountDownLatch` barriers rather than sleeps. A delegating preference proxy pauses a real preference read after the underlying method has returned. It does not hold the preference implementation's locks or invent preference values.

The writer completes its update before the paused reader resumes. The assertion then checks a fresh `getOptions()` call after both operations have finished. This distinguishes stale cache publication from an old value observed during an overlapping read.

`testCacheIsReused` checks that repeated cache hits do not reread preferences. It is a functional cache-reuse check, not a throughput, allocation, or contention benchmark.

## Before/after comparison and validation status

[Test-only PR #5366](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/5366) contains the same test class and suite registration, but omits all production changes from this PR.

The comparison was prepared from this PR at `f38704d15668ca9bdd9c76a89c279e08aa95b661`, with both variants based on upstream commit `8c40c7d2ae12c0a32ab3cca1ab31b53956c65d51`.

An [earlier SDK-based diagnostic run](https://github.com/carstenartur/eclipse.jdt.core/actions/runs/33955492150) is retained as supplementary evidence. The previously reported eight-cache-test results must not be presented as results for the current ten-test version. That diagnostic setup is also distinct from a normal upstream Tycho regression-suite run.

The required comparison is the same stale-value assertions failing without the fix and passing with it, with all ten tests passing on the fixed version. Individual JUnit results and the tested revisions are needed to establish that comparison.

The [first Jenkins run of #5366](https://ci.eclipse.org/jdt/job/eclipse.jdt.core-Github/job/PR-5366/1/) ended with a pipeline timeout. That is not a JUnit regression result and is not evidence that the proposed fix makes a difference.

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/AllJavaModelTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/AllJavaModelTests.java
@@ -139,6 +139,7 @@ private static Class[] getAllTestClasses() {
 
 		// Options tests
 		OptionTests.class,
+		OptionCacheTests.class,
 
 		// Type hierarchy tests
 		TypeHierarchyTests.class,
--- a/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/OptionCacheTests.java
+++ b/org.eclipse.jdt.core.tests.model/src/org/eclipse/jdt/core/tests/model/OptionCacheTests.java
@@ -1,0 +1,356 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.jdt.core.tests.model;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Proxy;
+import java.util.Hashtable;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
+
+import org.eclipse.core.runtime.Platform;
+import org.eclipse.core.runtime.preferences.IEclipsePreferences;
+import org.eclipse.core.runtime.preferences.InstanceScope;
+import org.eclipse.jdt.core.JavaCore;
+import org.eclipse.jdt.core.formatter.DefaultCodeFormatterConstants;
+import org.eclipse.jdt.internal.core.JavaModelManager;
+
+import junit.framework.Test;
+import junit.framework.TestCase;
+import junit.framework.TestSuite;
+
+/** Tests the real options cache without creating projects, editors or builders. */
+public class OptionCacheTests extends TestCase {
+	private static final long TIMEOUT_SECONDS = 30;
+	private static final String KEY = DefaultCodeFormatterConstants.FORMATTER_INSERT_SPACE_AFTER_CLOSING_PAREN_IN_CAST;
+	private Hashtable<String, String> savedOptions;
+
+	public OptionCacheTests(String name) {
+		super(name);
+	}
+
+	public static Test suite() {
+		return new TestSuite(OptionCacheTests.class);
+	}
+
+	@Override
+	protected void setUp() throws Exception {
+		super.setUp();
+		assertTrue("These tests require an Eclipse runtime", Platform.isRunning());
+		this.savedOptions = JavaCore.getOptions();
+	}
+
+	@Override
+	protected void tearDown() throws Exception {
+		try {
+			if (this.savedOptions != null) {
+				JavaCore.setOptions(this.savedOptions);
+			}
+		} finally {
+			super.tearDown();
+		}
+	}
+
+	public void testCompletedSetOptionsCannotBeOverwritten() throws Exception {
+		assertLaterReadIsCurrent(Update.SET_OPTIONS);
+	}
+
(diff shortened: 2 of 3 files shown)


jukzi:
Hi @carstenartur, the summaries and Test reads like AI slop, so i am not sure if the problem was really understood and fixed. Please submit the Test without the fix in a separate PR to show that it actually fails without the proposed fix.

carstenartur:
> Hi @carstenartur, the summaries and Test reads like AI slop, so i am not sure if the problem was really understood and fixed. Please submit the Test without the fix in a separate PR to show that it actually fails without the proposed fix.

Hi @jukzi,

To be transparent: I used AI to generate the proposed implementation, tests and descriptions. I do not expect the result to be accepted on trust, and using AI does not remove my responsibility as the contributor to provide a sound change and reproducible evidence.

Your request to submit the tests without the fix is reasonable. I have opened [#5366](https://github.com/eclipse-jdt/eclipse.jdt.core/pull/5366) with the same tests and assertions, but without the production changes. I have also revised this PR's description to acknowledge earlier investigations and distinguish the available evidence from the validation that is still missing.

At the same time, I found the “AI slop” wording discouraging. I understand being cautious about generated code, especially in a central synchronization path. Specific concerns about the failure scenario, the test setup, or the locking approach would help me much more than that label.

I have been asking questions about related options and formatting problems in the older Bugzilla reports, and some of those questions still seem unanswered to me. I appreciate the existing investigations, including your input in #1445 and @trancexpress's debugging in [#2811](https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/2811#issuecomment-3967356378). My frustration is not that nobody has worked on this, but that I am still looking for a resolution and would welcome more discussion of the specific mechanism this PR addresses: an older `getOptions()` computation publishing its map after an update or invalidation has completed.

I have also had difficulty obtaining a useful comparison through CI. While #5366 was a draft, the available checks were green, but there was no Jenkins test run. After it was marked ready for review, the first Jenkins run ended with a pipeline timeout in the `sh` step. That is not a JUnit assertion failure, and I am not presenting it as evidence for the fix.

From a contributor's perspective, it is frustrating to have to mark a PR ready for review before obtaining the test results needed to prepare it for review. I am not asking for a lower standard of validation; I am trying to meet that standard.

Could you point me to the preferred way to run just `OptionCacheTests` against both revisions in CI, ideally while the PR is still a draft? I would also appreciate concrete technical feedback on the proposed approach.

jukzi:
i have rerun the test-only build. it showed the test failed i.e works:
image
So this PR seems to have a legit usecase. Please squash to a single commit.

jukzi (approved):

carstenartur:
> i have rerun the test-only build. it showed the test failed i.e works:
> 
> image
> 
> So this PR seems to have a legit usecase. Please squash to a single commit.

Hi @jukzi, thanks for rerunning the test-only build and for your review!

I had assumed the commits would be squashed when merging, since the [JDT contribution guidelines](https://github.com/eclipse-jdt/.github/blob/main/CONTRIBUTING.md#reviewing-a-change) recommend “Squash and Merge” for bug fixes.

Happy to do it beforehand, though—I’ve now squashed the five commits into a single commit and updated the PR branch. The code and tests are unchanged.

jukzi:
@carstenartur thanks for contributing

PR #5371, opened 2026-09-07
Incremental restructuring of constructor flow analysis

### LINKED ISSUE #5369: [Null][Records] Nullness tracking broken inside compact constructors
Found by code inspection:

Given:

```

public record X(@NonNull String s) {
    public X {
        // should be diagnosed as dead code / always false if nullness is properly tracked:
        if (s == null) { // <---- No warning here!!!
            System.out.println("impossible");
        }

        @NonNull String k = "";
        if (k == null) { // correctly triggers Redundant null check: comparing '@NonNull String' & Dead code
            System.out.println("impossible");
        }
        
        // should be diagnosed as potential null dereference if flow thinks s may be null:
        s.toString();
    }
}

record R2(@NonNull String s) {
    public R2(@NonNull String s) {
        if (s == null) { // correctly triggers Redundant null check: comparing '@NonNull String' & Dead code             
            System.out.println("impossible");
        }
        this.s = s;
        s.toString();                  // should be clean
    }
}
```

We don't get redundant null check and dead code warnings in the line tagged with `if (s == null) { // <---- No warning here!!!`

srikanth-sankaran:
* Declutter ConstructorDeclaration.analyzeCode() by moving unused constructor checks and unused type variables checks into separate methods

* Fix broken nullity tracking inside compact constructors

* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5369

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/AbstractMethodDeclaration.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/AbstractMethodDeclaration.java
@@ -252,15 +252,17 @@ public abstract class AbstractMethodDeclaration
 	 * <li>NotOwning - for resource leak analysis
 	 * </ul>
 	 */
-	static void analyseArguments(LookupEnvironment environment, FlowInfo flowInfo, FlowContext flowContext, Argument[] methodArguments, MethodBinding methodBinding) {
+	static void analyseArguments(MethodScope scope, FlowInfo flowInfo, FlowContext flowContext, AbstractVariableDeclaration[] methodArguments, MethodBinding methodBinding) {
 		if (methodArguments != null) {
+			LookupEnvironment environment = scope.environment();
 			boolean usesNullTypeAnnotations = environment.usesNullTypeAnnotations();
 			boolean usesOwningAnnotations = environment.usesOwningAnnotations();
 
 			int length = Math.min(methodBinding.parameters.length, methodArguments.length);
 			for (int i = 0; i < length; i++) {
 				TypeBinding parameterBinding = methodBinding.parameters[i];
-				LocalVariableBinding local = methodArguments[i].binding;
+				AbstractVariableDeclaration methodArgument = methodArguments[i];
+				LocalVariableBinding local = methodArgument instanceof Argument argument ? argument.binding : scope.findVariable(methodArgument.name);
 				if (usesNullTypeAnnotations) {
 					// leverage null type annotations:
 					long tagBits = parameterBinding.tagBits & TagBits.AnnotationNullMASK;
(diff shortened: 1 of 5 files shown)


srikanth-sankaran:
@stephan-herrmann - you may want to glance through these changes. This doesn't touch any of your code for JEP 513. (there is a fix for a records constructor defect I may have caused)

The only exception to that is a couple of places I have dropped the `if` in

```
if (JavaFeature.FLEXIBLE_CONSTRUCTOR_BODIES.isSupported(this.scope.compilerOptions())) {
	this.scope.enterEarlyConstructionContext();
}
```

Let me know if you see a problem with that.

stephan-herrmann:
> The only exception to that is a couple of places I have dropped the `if` in
> 
> ```
> if (JavaFeature.FLEXIBLE_CONSTRUCTOR_BODIES.isSupported(this.scope.compilerOptions())) {
> 	this.scope.enterEarlyConstructionContext();
> }
> ```

Is that "if" provably redundant at those places?

Or are you saying, executing `enterEarlyConstructionContext()` never hurts, even if the feature is not enabled?

srikanth-sankaran:
> > The only exception to that is a couple of places I have dropped the `if` in
> > ```
> > if (JavaFeature.FLEXIBLE_CONSTRUCTOR_BODIES.isSupported(this.scope.compilerOptions())) {
> > 	this.scope.enterEarlyConstructionContext();
> > }
> > ```
> 
> Is that "if" provably redundant at those places?
> 
> Or are you saying, executing `enterEarlyConstructionContext()` never hurts, even if the feature is not enabled?

The latter is what I had in mind. From https://openjdk.org/jeps/513:

>Currently, code that appears in the argument list of an explicit constructor invocation is said to appear in a [static context](https://docs.oracle.com/javase/specs/jls/se23/html/jls-8.html#jls-8.1.3). This means that the arguments to the explicit constructor invocation are treated as if they were code in a static method, in which no instance is available. The technical restrictions of a static context are stronger than necessary, however, and they prevent code that is useful and safe from appearing as constructor arguments.

>Rather than revise the concept of a static context, we introduce the concept of an early construction context that covers both the argument list of an explicit constructor invocation and any statements that appear before it in the constructor body, i.e., in the prologue. Code in an early construction context must not use the instance under construction, except to initialize fields that do not have their own initializers.

From this I inferred it never hurts. The very first statement in JDK24- must be a chaining call. The early construction context will end with it - see that the leave's are unconditional. By turning on `enterEarlyConstructionContext()` unconditionally, the only change we would see is that constructor arguments will appear to be in early construction context - which is fine and correct for Java25+.

For Java24- we are in a more technically restrictive environment anyway - static context imposed by `org.eclipse.jdt.internal.compiler.lookup.MethodScope.isConstructorCall` being true.

As a proof that it doesn't hurt this program continues to be rejected at Java 24- while compiling fine in Java 25+:

```
public class X {
	int abcd;
	
	X(int x) {
		
	}
	
	X() {
		this(abcd = 10);
	}
}
```

I'll add this to the testsuite - although I suspect it is redundant

PR #5372, opened 2026-09-07
Workspace compliance level is not properly "inherited" by projects

### LINKED ISSUE #5281: Workspace compliance level is not properly "inherited" by projects
## Description

In a workspace with a compliance level that is lower than the Java version of the JDK used for compilation and a Java project with no project-specific compliance settings, the compiler will not complain about library contents that are only available in a later Java version. In case the same compliance is set as project-specific setting, the compiler will report an error.

As one example, the `Path.resolve(String, String ...)` method introduced in Java 22 can be used when a JDK 25 is set for the workspace even though workspace compliance level is set to 21. Another example is the usage of Java 25's `ScopedValue`.

### Reproduction

1. Configure your workspace to use compliance 21 and use a Java 25 JDK
   Image
   Image
2. Create a Java project and add code that uses `Path.resolve(String, String ...)`, e.g.:   `Path.of("a").resolve("b", "c");`
3. Ensure that the Java project has no project-specific compliance settings defined
   Image

Here is an according demo project: [JavaComplianceDemo.zip](https://github.com/user-attachments/files/30986672/JavaComplianceDemo.zip)

The code will compile unexpectedly: 
Image

## Expected Behavior

The compiler should report compilation errors for code that is incompatible with the workspace compliance level (in case no project-specific settings are defined), just like it does when setting project-specific compliance:
Image

This conforms to what happens when compiling the class with a Java 25 javac:
`javac.exe -d out --release 21 src\demo\DemoClass.java`

Resulting in:
```
DemoClass.java:7: error: method resolve(String,String) not applicable in Path
      Path.of("a").resolve("b", "c");
```

## Environment
* OS+Version: z.B. Windows 11
* Eclipse IDE/Platform-Version (*Help > About > ...*): Eclipse 2026-09 (I20260803-2300), also reproducible in 2025-12, so probably existed for longer
* Java-Vendor und Version: Temurin 25.0.2

## Workaround
Set the project-specific compliance to the desired level. That one is picked up properly.

## Additional Information
This does only seem to affect library functions and not language constructs. E.g., trying to use flexible constructor bodies in the above example fails with "The Java feature 'Flexible Constructor Bodies' is only available with source level 25 and above".

It might be that I miss something that justifies the current behavior, but having different compilation results depending on whether compliance level is defined by the workspace or the project is something we did not expect.

jarthana:
The --release option should default to workspace setting if not set

## What it does
Fixes #5281

## Author checklist

### DIFF
--- a/org.eclipse.jdt.core.tests.builder/src/org/eclipse/jdt/core/tests/builder/BuilderTests11.java
+++ b/org.eclipse.jdt.core.tests.builder/src/org/eclipse/jdt/core/tests/builder/BuilderTests11.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2022, Andrey Loskutov (loskutov@gmx.de) and others.
+ * Copyright (c) 2022, 2026 Andrey Loskutov (loskutov@gmx.de) and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
@@ -13,6 +13,7 @@
  *******************************************************************************/
 package org.eclipse.jdt.core.tests.builder;
 
+import java.util.Hashtable;
 import junit.framework.Test;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.jdt.core.JavaCore;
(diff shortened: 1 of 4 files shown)


trancexpress:
Looks like this caused a test fail on Java 25 fails: https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/1000

jarthana:
> Looks like this caused a test fail on Java 25 fails: [eclipse-jdt/eclipse.jdt.debug#1000](https://github.com/eclipse-jdt/eclipse.jdt.debug/issues/1000)

Will take a look. Probably can be handled during #995

Issue #5249, opened 2026-07-29
Compilation errors when annotation-based null analysis is enabled


ptziegler:
This is an error that shows up with the I20260728-2300. I didn't see this issue with Mondays build and am fairly certain this is related to 62bd2a36042cc3ed2ba517ece5fedd6b6c29e357.

To reproduce:

- Create a new Java project with annotation-based null analysis enabled
- Add the JDT annotation plugin to the classpath

<img width="1592" height="937" alt="Image" src="https://github.com/user-attachments/assets/8869bb65-50c7-4060-914a-5b36335d7e2e" />

- Add the following class to the project:

```java
package test;

import java.util.Collection;
import java.util.List;

public class Test {
    private ListValuedMap<String, String> multiMap;

    public Test() {
        List<String> values = multiMap.get("");
    }

    public static interface ListValuedMap<K, V> extends MultiValuedMap<K, V> {
        @Override
        List<V> get(K key);
    }

    public static interface MultiValuedMap<K, V> {
        Collection<V> get(K key);
    }
}
```

This should produce the following, invalid, compilation error:

> Type mismatch: cannot convert from Collection<String> to List<String>

If I remove the JDT plugin or disable the analysis, the error disappears.

srikanth-sankaran:
Stephan, thanks for investigating.

stephan-herrmann:
Thanks for the report.

I had naively expected that https://github.com/eclipse-jdt/eclipse.jdt.core/commit/62bd2a36042cc3ed2ba517ece5fedd6b6c29e357 would operate only on methods with the same signature & return type disregarding annotations.

This example proves that we may get methods with covariant return types. It would be tempting to let `NullAnnotationMatching.strongerType()` also pick the most specific type wrt covariance, but in #5250 I decided to restrict the effect of that method to null annotations only. 

This leaves the general question untouched, whether it's kosher that we are seeing methods with different return types in this situation.

ptziegler:
Many thanks! I installed yesterdays build and can confirm that the compilation errors no longer show up in my workspace. 😄

PR #5373, opened 2026-09-08
Preserve covariant return types when merging null annotations

wenytang-ms:
## Summary

Preserve the covariant return type already selected by Java method resolution when merging null annotations from override-equivalent inherited methods.

Addresses redhat-developer/vscode-java#4500. This is a follow-up to eclipse-jdt/eclipse.jdt.core#5250, not a change to the language server's null-analysis settings.

## Root cause

`Scope.mostSpecificMethodBinding()` can select a method whose return type is more specific than the first entry in `moreSpecific`. However, `methodWithMergedNullAnnotations()` initializes the merged return type from `moreSpecific[0]`, rather than from the selected method.

Since `strongerType()` deliberately preserves its first argument's underlying Java type, merging annotations can replace the selected subtype with a supertype. In AWS SDK 2.42.26, this makes `GetParametersByPathRequest.builder().build()` appear to return `SsmRequest` when annotation-based null analysis is enabled.

The fix starts with `current.returnType` and merges annotations from every candidate onto that type. Parameter annotation merging is unchanged.

## Regression history

Historical ECJ builds using the same AWS input and null-analysis settings show:

| Revision | Result |
| --- | --- |
| `27f2d7a`, parent of eclipse-jdt/eclipse.jdt.core#5237 | Compiles |
| `62bd2a3`, eclipse-jdt/eclipse.jdt.core#5237 | Fails with an `Object` return type |
| `618869e`, eclipse-jdt/eclipse.jdt.core#5250 | Fails with an `SsmRequest` return type |

The earlier follow-up fixed the explicit covariant override reported in eclipse-jdt/eclipse.jdt.core#5249, but did not cover the separately inherited generic builder methods used by the AWS SDK.

## Regression coverage

- `testGH5249_inheritedBuilders`: generic and covariant `build()` declarations in both interface orders, resolved from source and then from class files.
- `testGH5249_inheritedNullAnnotations`: preserve the covariant Java type while merging the strongest return nullness from three interfaces, in both orders.

The tests do not depend on the AWS SDK.

## Validation

The Maven/Tycho compiler reactor succeeded, and `NullTypeAnnotationTest` reported **958 tests, 0 failures, 0 errors**. The Maven-built ECJ also compiles both original AWS expressions with null analysis enabled.

```text
mvn -B -ntp -pl org.eclipse.jdt.core.tests.compiler -am verify
  "-Dtest=NullTypeAnnotationTest#testGH5249*+testGH5070*"
  -DfailIfNoTests=false
  "-Dtycho.surefire.argLine=--add-modules ALL-SYSTEM -Dcompliance=1.8,21 -Djdt.performance.asserts=disabled"
  -DcompilerBaselineMode=disable -DcompilerBaselineReplace=none
```

The test provider ran the full `NullTypeAnnotationTest` suite for this invocation.

This draft fixes the compiler. The downstream VS Code issue will additionally require consuming the fixed JDT Core version through JDT LS.

### DIFF
--- a/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/NullAnnotationMatching.java
+++ b/org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/NullAnnotationMatching.java
@@ -856,9 +856,12 @@ public class NullAnnotationMatching {
 		if (count < 2)
 			return current;
 		TypeBinding[] parameters = weakerTypes(moreSpecific[0].parameters, moreSpecific[1].parameters, environment);
-		TypeBinding returnType = strongerType(moreSpecific[0].returnType, moreSpecific[1].returnType, environment);
 		for (int i = 2; i < count; i++) {
 			parameters = weakerTypes(parameters, moreSpecific[i].parameters, environment);
+		}
+		// Preserve the covariant return type selected by method resolution.
+		TypeBinding returnType = current.returnType;
+		for (int i = 0; i < count; i++) {
 			returnType = strongerType(returnType, moreSpecific[i].returnType, environment);
 		}
 		if (parameters != current.parameters || returnType != current.returnType) { //$IDENTITY-COMPARISON$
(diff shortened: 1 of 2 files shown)


wenytang-ms:
## Current

image

image

wenytang-ms:
## Fixed

image

wenytang-ms:
@stephan-herrmann please help review this PR, thanks

wenytang-ms:
@srikanth-sankaran please review this pr.

PR #5382, opened 2026-09-09
Incremental restructuring of flow analysis for flexible constructor bodies 

### LINKED ISSUE #4453: Complete and further test the implementation of UnconditionalDualFlowInfo
Follow-up from #4452:

Newly introduced class `UnconditionalDualFlowInfo` still lacks overrides for a few methods from super.

Most of these relate to null analysis.

To achieve some degree of test coverage we should create some tests where constructor prologues play a relevant role in null analysis.

### LINKED ISSUE #5378: ECJ fails to complain about a flexible constructor failing to initialize a blank final variable.
4.40, 4.41 M2 and master/HEAD all compile the following illegal program:

```
public class X {
    final String s;

    X(String s0) {
        s = s0;
        super();
    }

    X() {
        System.out.println("Blah");
        super();
    }

        System.out.print(new X("OK").s);
    }
}
```

Javac 26 and 28 EA versions correctly reject this with: `X.java:12: error: variable s might not have been initialized`

Note: both constructors are required to reproduce the problem.

### LINKED ISSUE #5396: [flexible-constructors] java.lang.VerifyError: Constructor must call super() or this() before return
From https://github.com/eclipse-jdt/eclipse.jdt.core/pull/5383#issuecomment-5608408274:

We don't handle the program from the comment above (shared below) correctly: 

- On 4.40 GA and 4.41 M2 and on master/HEAD the generated class crashed with:

```
Error: Unable to initialize main class X
Caused by: java.lang.VerifyError: Constructor must call super() or this() before return
Exception Details:
  Location:
    X.<init>()V @14: return
  Reason:
    Error exists in the bytecode
  Bytecode:
    0000000: bb00 0a59 b700 0cbf 2a12 0eb5 0010 b1  
  Stackmap Table:
    same_frame(@8)
```

Test case:

```
public class X {
	String s = "good";
	X() {
		if (true) throw new RuntimeException();
		super(); 
	}
	X (int i) {
		System.out.println(s);
	}
		new X(1);
	}
}
```

### LINKED ISSUE #5400: [Flexible constructors] VerifyError: Bad type on operand stack
On 4.40, 4.41 M2 and master/HEAD the following program compiled by ECJ fails to pass verification:

```
public class X {
    String field = "initialized";
    
    {  // Instance initializer block
        System.out.println(field);
        field = "modified";
    }
    
    // First constructor with unreachable super() call
    X() {
        if (true) throw new RuntimeException();
        super();  // unreachable
    }
    
    // Second constructor that can reach the field initializer
    X(int x) {
        System.out.println(field);  // uses the field
    }
    
        new X(1);
    }
}
```

When run, we get:

```
Error: Unable to initialize main class X
Caused by: java.lang.VerifyError: Bad type on operand stack
Exception Details:
  Location:
    X.<init>()V @18: getfield
  Reason:
    Type uninitializedThis (current frame, stack[1]) is not assignable to 'X'
  Current Frame:
    bci: @18
    flags: { flagThisUninit }
    locals: { uninitializedThis }
    stack: { 'java/io/PrintStream', uninitializedThis }
  Bytecode:
    0000000: bb00 0a59 b700 0cbf 2a12 0eb5 0010 b200
    0000010: 122a b400 10b6 0018 2a12 1eb5 0010 b1  
  Stackmap Table:
    same_frame(@8)


```

### LINKED ISSUE #5404: [Flexible constructors] Incorrect diagnostic: The final field x may already have been assigned
I am not claiming snippets like below are _realistic_, but it does reveal _real_ problem in flow analysis for flexible constructors:

The following program is rejected by 4.40, 4.41 M2 and master/HEAD with 2 errors of the form `The final field x may already have been assigned` while javac compiles them correctly.  See that there are no other assignments to `x` and `z` in the program, so these messages are plain wrong:

```
        public final class X {
          final int x; // 1
          final int y; // 2
          final int z; // 4

          {
              x = 10;
              z = y;
          }

          X() {
              int xx = 123; // 8
              if (true)
                  throw new RuntimeException();
              super();
          }
          X(int a) {
              this.y = 10;
              if (true)
                  throw new RuntimeException();
              super();
          }
        }

```

### LINKED ISSUE #5405: [Flexible constructors] Field initialization analysis should not be influenced by constructor locals initialization status in prologues.
Found by code inspection.

I have been starting at this code on master for some time. I am convinced there is a defect here, but I don't yet have a tangible problem to report because the defect is perhaps masked somehow.

org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.internalAnalyseCode(FlowContext, FlowInfo) reads:

```
...
        // collect field initializations happening in constructor prologues
			FlowInfo prologueInfo = null;
			boolean allConstructorsHavePrologue = true;
			for (AbstractMethodDeclaration method : this.methods) {
				if (method instanceof ConstructorDeclaration constructor) {
					FlowInfo ctorInfo = flowInfo.unconditionalFieldLessCopy();
					constructor.analyseCode(this.scope, initializerContext, ctorInfo, ctorInfo.reachMode(), PROLOGUE_ANALYSIS);
					ctorInfo = constructor.getPrologueInfo();
					if (ctorInfo == null) {
						allConstructorsHavePrologue = false;
					} else if (ctorInfo.hasInits()) {
						if (prologueInfo == null)
							prologueInfo = ctorInfo.copy();
						else
							prologueInfo = prologueInfo.mergeDefiniteInitsWith(ctorInfo.unconditionalInits()); // will only evaluate field inits below
					}
				}
			}
...
```
The prologue gathered from the constructors is still carrying along initialization status for *locals* declared inside the constructor and these are erroneously exposed to influence the field declarations. 

Or in other words, it is missing the equivalent of this block of code from analyzeFieldInitializations() in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/5383.

```
	// field initialization analysis should not see constructor locals.
	LocalVariableBinding[] locals = this.scope.locals;
	if (locals != null) {
		int numLocals = this.scope.localIndex;
		for (int i = 0; i < numLocals; i++)
			flowInfo.resetAssignmentInfo(locals[i]);
	}
```

As a result, initialization status of constructor locals from prologues leaks into the `DualFlowInfo`. I can verify by stepping through the code while compiling:

```
			public final class X {
				final int x; // 1
				final int y; // 2
				final int z; // 4

				{
					final int zzz;
					System.out.println(zzz);
					x = 10;
					z = 20;
				}

				X() {
					int xx = 123; // 8
					super();
					y = 89;
				}

				X(int a) {
					this.y = 10;
					super();
				}
			}
```

that such a leak and overlay does happen between constructor parameter `a` and local `xx` and the initializer local `zzz` but perhaps due to a different bug somewhere else this problem gets masked and no ill effects are observed.

It is certainly an error that constructor locals should carry over to field analysis - this needs investigation.

https://github.com/eclipse-jdt/eclipse.jdt.core/pull/5383 - already addresses this while https://github.com/eclipse-jdt/eclipse.jdt.core/pull/5382 does not.

### LINKED ISSUE #5410: Missing dead code warning
This came up during code review of PR #5383 but is about long standing behavior:

This program generates a `Dead code` warning as it should:

```
public class X {

	{
		if (true)
			throw new RuntimeException();
		System.out.println();
	}
	X(int a) {
		super();
		System.err.println();
		System.out.println();
	}
}
```

while this doesn't : (no justification)

```
public class X {

	{
		if (true)
			throw new RuntimeException();

	}
	X(int a) {
		super();
		System.err.println();
		System.out.println();
	}
}
```

srikanth-sankaran:
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5378
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5396
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5400
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5404
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5405
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4453
* Fixes https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5410


srikanth-sankaran:
This is work in progress - Fails 8 tests in TestAll run at 26:

org.eclipse.jdt.core.tests.compiler.regression.FlowAnalysisTest.testBug338234c()
org.eclipse.jdt.core.tests.compiler.regression.BatchCompilerTest.test032()o
org.eclipse.jdt.core.tests.compiler.regression.SerializableLambdaTest.testbug485333()
org.eclipse.jdt.core.tests.compiler.regression.SerializableLambdaTest.testbug494487()
org.eclipse.jdt.core.tests.compiler.regression.JEP286Test.testBug567183_1()
org.eclipse.jdt.core.tests.compiler.regression.JEP286Test.testBug567183_2()
org.eclipse.jdt.core.tests.compiler.regression.JEP286Test.testBug567183_3()
org.eclipse.jdt.core.tests.compiler.regression.JEP286Test.testBug567183_4()

@stephan-herrmann - you are welcome to take a look - but I expect there will be more changes.

srikanth-sankaran:
@stephan-herrmann - Both the PRs are green. Let us discuss

srikanth-sankaran:
I completed a three-pronged self-review of the changes:

- Studying the implementation on its own (`ConstructorDeclaration.analyseCode` and its caller `TypeDeclaration.internalAnalyseCode` - highly recommended.
- Comparing with master
- Comparing against R4_30 branch

I recommend all three steps for reviewers.

- I still need to fully review and test `DualFlowInfo` and `UnconditionalDualFlowInfo` ( https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4453)

- I need validation from Stephan that this block in `TypeDeclaration.internalAnalyseCode` is doing the right thing:

```
			if (prologueInfo != null) {
				// DAs from EVERY constructor prologue should carry over to main flow info, while potential inits should flow into companion only.
				for (FieldBinding field : this.binding.fields()) {
					if (prologueInfo.isDefinitelyAssigned(field))
						nonStaticFieldInfo.markAsDefinitelyAssigned(field);
				}
				nonStaticFieldInfo = new DualFlowInfo(nonStaticFieldInfo, prologueInfo);
			}
```

Otherwise, this PR looks good to go according to me - if this is the pick over https://github.com/eclipse-jdt/eclipse.jdt.core/pull/5383 that is.

srikanth-sankaran:
Some input for review of #5382 and #5383

#5383 - the new design.  For the meat of the changes, it is perhaps better just junking diffs against master and start directly studying the code in  `ConstructorDeclaration.analyseCode`, its callees and its call site. I suspect contextual diff will be too distracting as it is an entirely different design. For the supporting changes contextual diff will work.

#5382 - new flow analysis implementation for existing design:  It would be a good idea to first glance through the locations called out above and then try diff with master.

One key point to internalize for #5382 is:

In the new implementation every super invoking constructor has a prologue - perhaps an empty one - this maps cleanly with the language. 

Constructors that chain via this (…) are excluded from prologue analysis completely. In the actual control flow, they don’t emit field initializations **at all** by themselves. So, it doesn’t make sense for them to influence field initializations or for field initializations to influence them - any part of them - including epilogue.

As a result, `DualFlowInfo` computes very different things in #5382 than in master.

In #5382 the companion flow info is just the **intersection of all** prologues flowinfo. Conceptually, this is what field analysis will **definitely** see irrespective of which (super chaining) constructor prologue precedes it.

During field analysis, it is to be treated read only - since what happens in field declarations cannot alter the state of the prologue. (although **co-altering** it identically with the main flowinfo will be harmless)

The key output of `DualFlowInfo` then is the flowinfo at the end of the field initialization analysis which when fused with the individual stashed prologue info of ANY constructor would be the accurate flowinfo entering the epilogue in THAT constructor. This fusing happens in EPILOGUE_ANALYSIS at the top in this block:

```
// Retrieve flow info and flow context stashed away during prologue analysis, skip prologue (below) and continue analysis.
			constructorContext = this.prologueContext;
			if ((this.prologueInfo.reachMode() & FlowInfo.UNREACHABLE) != 0)
				epilogReachMode = this.prologueInfo.reachMode(); // field initialization completion status is immaterial if prologue completes abruptly!
			else
				epilogReachMode = flowInfo.reachMode();
			flowInfo = this.prologueInfo.addInitializationsFrom(flowInfo);
```

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/AbstractMethodDeclaration.java
```diff
@@ -171,6 +171,7 @@ public static void computeParamFlowBits(MethodBinding binding, int argLen, long
 	 */
 	public void bindArguments() {
 
+		this.scope.analysisIndex = this.scope.outerMostMethodScope().analysisIndex;
```
srikanth-sankaran, line 174: We use this field in MethodScope which thus far is set _only_ for outerMostMethodScope() to record arg0 position for an inner method scope

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ConstructorDeclaration.java
```diff
@@ -175,19 +173,25 @@ public void analyseCode(ClassScope classScope, InitializationFlowContext initial
 			if (mode == PROLOGUE_ANALYSIS)
 				this.prologueContext = constructorContext; // save for EPILOGUE_ANALYSIS
 
+			epilogReachMode = flowInfo.reachMode();
```
srikanth-sankaran, line 176: May be we should have two variables. `prologueReachMode` (for this branch of `if`) and `epilogueReachMode` for `else`. Reusing the same variable name can be confusing ??

or perhaps just one variable `reachMode` ?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ConstructorDeclaration.java
```diff
@@ -196,16 +200,12 @@ public void analyseCode(ClassScope classScope, InitializationFlowContext initial
 					constructorCall = ecc;
 				if (mode == EPILOGUE_ANALYSIS && inPrologue) {
```
srikanth-sankaran, line 201: May be in prologue analysis we should save the index of start of epilogue and start the analysis there rather than skipping prologue like this. 

Not saying from a performance angle - I doubt that there will be many statements to skip. Prologues will likely be tiny. But may be more readable.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeDeclaration.java
```diff
@@ -751,15 +752,12 @@ private void internalAnalyseCode(FlowContext flowContext, FlowInfo flowInfo) {
 		if (this.methods != null) {
 			// collect field initializations happening in constructor prologues
 			FlowInfo prologueInfo = null;
-			boolean allConstructorsHavePrologue = true;
 			for (AbstractMethodDeclaration method : this.methods) {
-				if (method instanceof ConstructorDeclaration constructor) {
+				if (method instanceof ConstructorDeclaration constructor && constructor.invokesSuper()) {
```
srikanth-sankaran, line 756: In this PR, every super chaining constructor contributes a prologue which may influence field initializations, but no alternate chaining constructor does

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/flow/UnconditionalDualFlowInfo.java
```diff
@@ -96,7 +96,9 @@ public FlowInfo asNegatedCondition() {
 
 	@Override
 	public FlowInfo copy() {
-		return new UnconditionalDualFlowInfo(super.copy(), this.companionInits.copy());
+		FlowInfo copy = new UnconditionalDualFlowInfo(super.copy(), this.companionInits.copy());
```
srikanth-sankaran, line 99: Need to check more cases like this - I deferred it to https://github.com/eclipse-jdt/eclipse.jdt.core/issues/4453

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/flow/UnconditionalFlowInfo.java
```diff
@@ -839,11 +839,15 @@ public UnconditionalFlowInfo discardInitializationInfo() {
...
+ * Remove ALL local variables (including any of enclosing methods) information from this flow info and return this.
+ * @return this, deprived from any and all local variable information
  */
 public UnconditionalFlowInfo discardNonFieldInitializations() {
-	int limit = this.maxFieldCount;
+	return discardNonFieldInitializations(this.maxFieldCount);
+}
+
+
+public UnconditionalFlowInfo discardNonFieldInitializations(int limit) {
```
srikanth-sankaran, line 850: Need documentation here

stephan-herrmann:
I made a first rough scan through the changes in this PR. At this level all changes look _plausible_ to me. Congrats, @srikanth-sankaran  for fixing so many issues with comparably small changes!

I particularly like the removal of field `implicitArguments`, which according to a quick history scan may have survived 25 years without ever having been used at all! :)

I was slightly confused about the connection to all those issues reportedly being fixed. I couldn't find new tests directly referring to #5396 nor #5405.

At this point I'll switch over to the other PR, with a mental note, that in case the other PR should fail review, we have a fallback in this current PR, where I'm confident that I should be able to convince myself of its correctness with reasonable efforts.

srikanth-sankaran:
> which according to a quick history scan may have survived 25 years without ever having been used at all! :)

🤣 

> I was slightly confused about the connection to all those issues reportedly being fixed. I couldn't find new tests directly referring to #5396 nor #5405.

#5405 :

Here is the sequence of events: 

As #5405 "comment#0" reads: `Found by code inspection. I have been starting at this code on master for some time. I am convinced there is a defect here, but I don't yet have a tangible problem to report because the defect is perhaps masked somehow.`

but after investing some amount of time, I didn't quite come up with a test case for #5405 that manifests some ill behavior. But since we can conclude readily it is an error for a prologue to expose its locals to field analysis, I went ahead with the fix in `org.eclipse.jdt.internal.compiler.ast.ConstructorDeclaration.getPrologueInfo()` without a test since it is easy to verify by inspection.

#5396 : 

as "comment#0" calls out the reproducer is actually from your comment in https://github.com/eclipse-jdt/eclipse.jdt.core/pull/5383#issuecomment-5608408274 - the tests capture this URL and don't mention #5396 

(So the verify error was/is happening in master, in #5383 where you tested the snippet, the verify error was already gone, but it was printing `null` instead of `good` - it was fixed there. When the tests were synchronized in both universes, I realized that on master it fails to verify)

org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/SuperAfterStatementsTest.java
```diff
@@ -3817,18 +3813,248 @@ private X(boolean f) {
...
-			"	             ^^\n" +
-			"The blank final field s2 may not have been initialized\n" +
+			"1. ERROR in X.java (at line 5)\n" +
+			"	s = \"updated\";\n" +
+			"	^\n" +
+			"The final field s may already have been assigned\n" +
 			"----------\n");
 	}
+
+	// https://github.com/eclipse-jdt/eclipse.jdt.core/pull/5383#issuecomment-5608408274
```
srikanth-sankaran, line 3887: This test should mention #5396

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ConstructorDeclaration.java
```diff
@@ -73,7 +73,14 @@ public ConstructorDeclaration(CompilationResult compilationResult){
...
 
 FlowInfo getPrologueInfo() {
-	return this.prologueInfo;
+	if (this.prologueInfo == null) // may be null when `this.ignoreFurtherInvestigation` is true; epilogue analysis will be skipped too.
+		return null;
+	/* Field initialization analysis should not see constructor locals! Compute the index of the
+	   first constructor parameter in flow data and discard that and all subsequent locals.
+	*/
+	TypeDeclaration typeDeclaration = this.scope.enclosingClassScope().referenceContext;
+	int limit = typeDeclaration.maxFieldCount + (this.scope != this.scope.outerMostMethodScope() ? this.scope.analysisIndex : 0);
```
stephan-herrmann, line 82: I couldn't immediately see why we can't use `this.scope.analysisIndex` unconditionally, but noticed that this would cause codeGen to throw in TryStatementTest.test025().

Next I tried to literally translate the above comment to code:
```java
	int limit = typeDeclaration.maxFieldCount;
	if (this.arguments != null && this.arguments.length > 0) {
		limit += this.arguments[0].binding.id;
	}
```
But that causes 15 unique regressions in compiler.regression.TestAll.

@srikanth-sankaran can you explain, why your solution beats both my attempts, even though the second one looks very similar to the code comment above?

srikanth-sankaran: Yes. Although every `MethodScope` has this field, all nested `MethodScope`s share the outer most enclosing method scope's `analysisIndex`.  See how `org.eclipse.jdt.internal.compiler.lookup.BlockScope.addLocalVariable(LocalVariableBinding)` and `org.eclipse.jdt.internal.compiler.lookup.BlockScope.registerTrackingVariable(FakedTrackingVariable)` are coded.

So this field is dormant in every *enclosed* method scope - I am using this wasted space to store the index of the first local in that scope. (cf `org.eclipse.jdt.internal.compiler.ast.LambdaExpression.firstLocalLocal` where it is stored in AST)

For the outer most method scope itself, we can't rely on this field to capture the index of the first local in that scope - because it is _volatile_ - right after `org.eclipse.jdt.internal.compiler.ast.AbstractMethodDeclaration.bindArguments()` it is holding the right value, but as nested scopes invoke `org.eclipse.jdt.internal.compiler.lookup.BlockScope.addLocalVariable(LocalVariableBinding)` it no longer will !!! 

But then for the outer most method scope, we know the first local is right past maxFieldCount (so its index is 0) and we can directly use that value.


> I couldn't immediately see why we can't use `this.scope.analysisIndex` unconditionally, but noticed that this would cause codeGen to throw in TryStatementTest.test025().
> 
> Next I tried to literally translate the above comment to code:
> 
> ```java
> 	int limit = typeDeclaration.maxFieldCount;
> 	if (this.arguments != null && this.arguments.length > 0) {
> 		limit += this.arguments[0].binding.id;
> 	}
> ```

Conside this code:

```
class Outer {

    int field;

    void foo(int foo_arg0) {

        class Inner {
            String iField;

             Inner(int inner_arg0) {
             }
        }
}
```

maxFieldCount = 2 for both the classes. It holds culmulative field count of all the nested classes.

foo_arg0 is just after maxFieldCount
inner_arg0 is just after foo_arg0 not after maxFieldCount

stephan-herrmann: OK, this is more complex than I thought.

What do you consider easiest / cleanest?
* update the code comment to reflect what really is happening (constructor arguments don't seem to suffice for the explanation).
* add a field `MethodScope.firstLocalIndex`
* _your resolution here_

srikanth-sankaran: > * add a field `MethodScope.firstLocalIndex`

This would be the cleanest. Trying to reuse a dormant field is cute, but generations of future maintainers (Ahem, just watched an Anderson Cooper interview with an ex-Anthropic employee who says AI will kill all humans by 2030 🤣 - hey look at the bright side, we don't have to deal with defects past that 😆 ) will find it so much easier to understand `firstLocalIndex`

srikanth-sankaran:
Rationale for `org.eclipse.jdt.internal.compiler.flow.InstanceFieldsFlowInfo`:

- `InstanceFieldsFlowInfo` is a new abstraction to model flow analysis of instance field declarations of a class. It is derived from `UnconditionalFlowInfo`.
- The only site that instantiates a `InstanceFieldsFlowInfo` is `org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.internalAnalyseCode(FlowContext, FlowInfo)` in this manner:

```
if (prologueInfo != null)
	nonStaticFieldInfo = new InstanceFieldsFlowInfo(nonStaticFieldInfo.unconditionalInits(), prologueInfo);
```

- The idea behind the abstraction is as follows:
    * In an identity class - which is all we have as of Java 27, field initializations take place **after** the call to `super(...)` (and never in a constructor that chains to another via `this(...)`) and **before** the rest of the body of the constructor past the `super(...)` call that constitutes the epilogue.
    * As a result, instance field declarations are influenced by every prologue of every super-chaining constructor and the end state flow analysis output of instance fields analysis will influence every epilogue of every super-chaining constructor
    *  Instance field declarations are not influenced by any part of a constructor that chains to an alternate constructor via `this(...)` and vice versa. (hereafter all discussion is about super-chaining constructors only and likewise about only instance fields)
    * Fields flow analysis is influenced by every prologue. Conceptually, at the start of fields analysis, instance fields that are initialized by EVERY prologue are definitely assigned for the purposes of (the entire of the) field analysis, while instance fields that are initialized in one or more prologues but not in ALL of them are only potentially assigned.
    * Thus, in order to flow analyze the field declarations, we first analyze all prologues and merge their end state flow data to compute the start state flow data for field analysis.
    * Now as we analyze the fields, we may discover more assignments via initializations and keep updating the flow data to reflect the initialization status of the fields in the class.
    * What is wrong with this picture ? The end state flow data of field analysis - while it contains the accurate picture - is NOT suitable as the start state flow data for the analysis of ANY epilogue - because it is polluted by potential initializations from ALL prologues.
    * What is the correct start state flow data for epilogue analysis in _any_ particular constructor ? It is the prologue flow data of _that_ constructor, fused with fields flow analysis data minus the corrupting influence of every other prologue.
    * As a result, while field analysis is going on, we want the prologue data to be consulted, but right after field analysis is over, we want the end state fields analysis data to reflect the pristine data that would be output IRRESPECTIVE of whichever prologue preceded the field initializations. The field analysis end state data should be valid for EVERY prologue so that it can be used to drive the epilogue analysis of ANY constructor (by using the prologue flow information specific to _that_ constructor)
    * The new abstraction initializes itself so that fields that are initialized in EVERY prologue are DA in its world. It iterates through the fields and augments the flow data with further updates to assignments states as required by the AST. In this phase the prologue flow info is READONLY. What initializations happen in fields block cannot impact prologues as they precede fields initialization. A field is considered DA only if it was DA at entry into fields analysis (i.e was DA in every prologue) or was DA in a field initializer. HOWEVER, for a field to be considered potentially assigned, it may have been assigned in ANY of the prologues or in the field declarations.
    * The new abstraction simply implements the model outline in the previous point.
    * At the end of the field analysis, the abstraction shorn of all prologue data gives the flow info that when fused with the prologue flow info specific to a constructor will result in the right start flow data for epilogue analysis in that constructor

srikanth-sankaran:
Here are some points to reason about the correctness of `org.eclipse.jdt.internal.compiler.flow.InstanceFieldsFlowInfo`:

Now that we have established the raison d'etre for the new abstraction, here are some operational details.

- Prologue info is readonly while field analysis is going on. Also prologue info has only fields data - constructor locals don't make it to there.
- `InstanceFieldsFlowInfo` can simply rely on inherited behavior for most everything. One thing to be careful about is not to allow for object identity change. `copy()` operations must be overridden so that the copy continues to be `InstanceFieldsFlowInfo` retaining the capability to consult prologue info (really only for potential assignments)
- In order to satisfy myself about the correctness, I have made a few changes to the super class `UnconditionalFlowInfo`
- All `UnconditionalFlowInfo` operations that mutate `this` will continue to just work as is because `InstanceFieldsFlowInfo` is a `UnconditionalFlowInfo` - only API calls needing special treatment are
    * org.eclipse.jdt.internal.compiler.flow.InstanceFieldsFlowInfo.isPotentiallyAssigned(FieldBinding) - should also consult prologue info which the super class will be clueless about.
    * org.eclipse.jdt.internal.compiler.flow.InstanceFieldsFlowInfo.hasInits() - ditto.
- All `UnconditionalFlowInfo` operations that return a copy now instantiate a copy not by calling the constructor, but by calling `newInstance()` factory method which is overridden in `InstanceFieldsFlowInfo` - so that a copy of the right class gets instantiated.
- The copy returning APIs are overridden in `InstanceFieldsFlowInfo` so that the copy carries along additional state specific to `InstanceFieldsFlowInfo`

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/flow/UnconditionalFlowInfo.java
```diff
@@ -769,54 +769,52 @@ final public boolean canOnlyBeNull(LocalVariableBinding local) {
...
-	// do not clone the DeadEnd
-	if (this == DEAD_END) {
-		return this;
-	}
-	UnconditionalFlowInfo copy = new UnconditionalFlowInfo();
-	// copy slots
-	copy.definiteInits = this.definiteInits;
-	copy.potentialInits = this.potentialInits;
-	boolean hasNullInfo = (this.tagBits & NULL_FLAG_MASK) != 0;
+public UnconditionalFlowInfo copy(UnconditionalFlowInfo that) { // Tat Tvam Asi!
```
srikanth-sankaran: Note: This is a mirror image of previously existing copy method - which was making a copy of the receiver. It is flipped to make `this` a copy of passed in argument - this is more useful I think as it enables new use cases (fashioning this after a argument `FlowInfo`) while still being able to do previous style of copying by saying `newInstance().copy(this)`

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/flow/UnconditionalFlowInfo.java
```diff
@@ -2135,12 +2141,16 @@ static int numberOfEnclosingFields(ReferenceBinding type){
 	return count;
 }
 
+public UnconditionalFlowInfo newInstance() {
```
srikanth-sankaran, line 2143: Allows injection of a level of indirection at the call site rather than hard coding class id at call site. Subclasses can override to control instantiation

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ConstructorDeclaration.java
```diff
@@ -63,17 +62,35 @@ public class ConstructorDeclaration extends AbstractMethodDeclaration {
 
 	public TypeParameter[] typeParameters;
 
-	private ExceptionHandlingFlowContext prologueContext;
-	private FlowInfo prologueInfo;
+	private ExceptionHandlingFlowContext constructorContext;
```
srikanth-sankaran, line 67: local + field merged to just field

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ConstructorDeclaration.java
```diff
@@ -128,108 +145,104 @@ private void complainOnUnusedTypeVariables() {
...
-			flowInfo = this.prologueInfo.addInitializationsFrom(flowInfo);
+			/* Passed in flowInfo corresponds to instance fields flow analysis output.
+			   Fuse it with prologue flow info to arrive at flow info for epilogue analysis.
+			*/
+			if ((this.prologueInfo.prologueFlowInfo.reachMode() & FlowInfo.UNREACHABLE) != 0)
+				epilogReachMode = this.prologueInfo.prologueFlowInfo.reachMode(); // epilogue is unreachable if prologue completes abruptly!
+			else
+				epilogReachMode = flowInfo.reachMode(); // epilogue is reachable, if last instance field was reachable.
+
+			flowInfo = this.prologueInfo.prologueFlowInfo.addInitializationsFrom(flowInfo); // compose flow info input for epilogue analysis.
```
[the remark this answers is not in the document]
srikanth-sankaran: `org.eclipse.jdt.core.tests.compiler.regression.SuperAfterStatementsTest.testNullWarningsAcrossConstructorCalls()` added

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ConstructorDeclaration.java
```diff
@@ -589,6 +595,11 @@ public boolean isConstructor() {
 	return true;
 }
 
+public boolean invokesSuper() {
```
srikanth-sankaran, line 590: Do we want a bit in AST node ? I don't expect this loop iterate too deep.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/TypeDeclaration.java
```diff
@@ -751,40 +751,21 @@ private void internalAnalyseCode(FlowContext flowContext, FlowInfo flowInfo) {
...
 					FlowInfo ctorInfo = flowInfo.unconditionalFieldLessCopy();
 					constructor.analyseCode(this.scope, initializerContext, ctorInfo, ctorInfo.reachMode(), PROLOGUE_ANALYSIS);
-					ctorInfo = constructor.getPrologueInfo();
-					if (ctorInfo == null) {
-						allConstructorsHavePrologue = false;
-					} else if (ctorInfo.hasInits()) {
+					ctorInfo = constructor.getPrologueFlowInfo();
+					if (ctorInfo != null) {
```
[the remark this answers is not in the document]
srikanth-sankaran: I don't think it matters. It is an error for a prologue to complete with a potentially assign to a final field, but changed anyway.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/flow/InstanceFieldsFlowInfo.java
```diff
@@ -0,0 +1,82 @@
...
+		InstanceFieldsFlowInfo copy = (InstanceFieldsFlowInfo) super.nullInfoLessUnconditionalCopy();
+		copy.prologueInfo = this.prologueInfo;
+		return copy;
+	}
+
+	@Override
+	public UnconditionalFlowInfo unconditionalFieldLessCopy() {
+		InstanceFieldsFlowInfo copy = (InstanceFieldsFlowInfo) super.unconditionalFieldLessCopy();
+		copy.prologueInfo = this.prologueInfo;
+		return copy;
```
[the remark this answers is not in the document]
srikanth-sankaran: Good point

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ConstructorDeclaration.java
srikanth-sankaran: Done

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ConstructorDeclaration.java
srikanth-sankaran: I have restructured it to be a bit more readable

srikanth-sankaran:
More input for reasoning correctness and completeness of the new abstraction `InstanceFieldsFlowInfo`.

What methods of `UnconditionalFlowInfo` MUST it override ? 

- `InstanceFieldsFlowInfo` is a `UnconditionalFlowInfo` for all purposes BUT with an additional piece of state `prologueInfo` that is a merger of all prologues. DAs from `prologueInfo` have been carried over to instance itself see constructor of IFFI.
-  `prologueInfo` is already stripped of all locals information, since constructor locals may not influence fields => No useful nullness information is in prologue too.
- Reachability information from prologues is already "consumed" in the constructor of IFFI to set its reachability.
- Essentially this means ONLY potential initializations in prologues are STILL material. As a result EVERY API that touches potential initializations of fields (not locals) needs override.
- Secondly, while fields analysis is underway the current flow info ALWAYS must be an `InstanceFieldsFlowInfo` - or a `ConditionalFlowInfo` that wraps two `InstanceFieldsFlowInfo` . If this invariant is violated, we may end up with a state where prologue potential initializations are not taken into considerations in some analysis.
- To prevent such from happening, every copy operation must be overridden to ensure the copy handed out is an `InstanceFieldsFlowInfo`
- To encapsulate this better, I have introduced a new instance method `newInstance()` to not have call sites deal with actual object identities which is overridden in IFFI
- All other behavior - which is vast majority of behavior - is suitable as is and can be inherited from `UnconditionalFlowInfo` 
- What methods of `UnconditionalFlowInfo` invoke `newInstance()` ?
    * UnconditionalFlowInfo.copy()
    * UnconditionalFlowInfo.nullInfoLessUnconditionalCopy()
    * unconditionalFieldLessCopy()
These are all overridden in IFFI to hand out the right object instance.

srikanth-sankaran:
I have gone through every method of `UnconditionalFlowInfo` to see which ones need override and added them:

There is some uncertainty about what is the right behavior for these:

org.eclipse.jdt.internal.compiler.flow.InstanceFieldsFlowInfo.mergeDefiniteInitsWith(UnconditionalFlowInfo)
org.eclipse.jdt.internal.compiler.flow.InstanceFieldsFlowInfo.mergedWith(UnconditionalFlowInfo)
org.eclipse.jdt.internal.compiler.flow.InstanceFieldsFlowInfo.discardInitializationInfo()
org.eclipse.jdt.internal.compiler.flow.InstanceFieldsFlowInfo.unconditionalFieldLessCopy()

For the first two, (the scenario in question is when the IFFI is UNREACHABLE_OR_DEAD. See what the super method is doing) it seems intuitive to expect that argument `UnconditionalFlowInfo` will be IFFI already. I feel once we are in field analysis IFFI is the currency and _stays_ the currency due to all copy and new instance operations being intercepted and overridden

For the last two, it feels like the fields analysis should only ever discard own data and never the prologues data.

@stephan-herrmann - let me know if you see any problems here! I suspect these are super corner cases anyways.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/flow/UnconditionalFlowInfo.java
stephan-herrmann: Wouldn't this style require an update also in `ConditionalFlowInfo.copy()`? 

Alternatively, I could think of a set of three methods:
* `newInstance()` - just as seen in this PR
* `copyValuesFrom(UnconditionalFlowInfo other)` - covariant by intention, holding the current body of `copy()` minus the instantiation
* `UnconditionalFlowInfo.copy()` - simply invoking the above two methods, and returning the result. 

In this line of thought `ConditionalFlowInfo` would need no further tweaks, I'd think.

stephan-herrmann:
@srikanth-sankaran as I'll be closely following your movements during review, may I ask you to avoid anything that needs a force push (rebase, amend) for the time being? Thanks.

stephan-herrmann:
> * I need validation from Stephan that this block in `TypeDeclaration.internalAnalyseCode` is doing the right thing:

That block has been removed, so it's not doing anything bad! :sunglasses:

stephan-herrmann:
> Constructors that chain via this (…) are excluded from prologue analysis completely. In the actual control flow, they don’t emit field initializations at all by themselves. So, it doesn’t make sense for them to influence field initializations or for field initializations to influence them - any part of them - including epilogue.

I was wondering about the last part: of course the epilogue of a this(..)-calling constructor will see the initialization state of fields. But in this regard it actually lives against the same contracts that apply also to each regular method, since the instance is fully initialized at that point. Is that the reasoning?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/flow/UnconditionalFlowInfo.java
stephan-herrmann: Meanwhile I found that `UFI.copy()` (no args) still exists, so the only difference between current state in the PR and my proposal is the name of the one-arg copy method :)

srikanth-sankaran:
> @srikanth-sankaran as I'll be closely following your movements during review, may I ask you to avoid anything that needs a force push (rebase, amend) for the time being? Thanks.

Noted, I don't have any changes forthcoming. FTR the force pushes since your last review don't impact the portion already reviewed at first pass. i.e the very first commit is intact

srikanth-sankaran:
> > * I need validation from Stephan that this block in `TypeDeclaration.internalAnalyseCode` is doing the right thing:
> 
> That block has been removed, so it's not doing anything bad! 😎

😆 It is moved in part to inside `org.eclipse.jdt.internal.compiler.flow.InstanceFieldsFlowInfo.InstanceFieldsFlowInfo(UnconditionalFlowInfo, FlowInfo)` 

Essentially I was trying to validate for fields definitely assigned in EVERY prologue we could transfer the DA to the fields analysis flow info ("main" in earlier avatar and `InstanceFieldsFlowInfo` instance in the updated PR) while leaving the potential assignments in the prologue itself. 

I have convinced myself that this is correct.

stephan-herrmann:
> #5405 :
> 
> Here is the sequence of events:
> 
> As #5405 "comment#0" reads: `Found by code inspection. I have been starting at this code on master for some time. I am convinced there is a defect here, but I don't yet have a tangible problem to report because the defect is perhaps masked somehow.`
> 
> but after investing some amount of time, I didn't quite come up with a test case for #5405 that manifests some ill behavior. But since we can conclude readily it is an error for a prologue to expose its locals to field analysis, I went ahead with the fix in `org.eclipse.jdt.internal.compiler.ast.ConstructorDeclaration.getPrologueInfo()` without a test since it is easy to verify by inspection.

In current state in this PR, this is a non-issue in multiple ways:

Even without the fix in `ConstructorDeclaration.getPrologueInfo()` the example from #5405 is correctly handled for these reasons:
* flow info at local position 8 (`1L<<(maxField+0)`) occurs only in `InstanceFieldsFlowInfo.prologueInfo` not in that info's main `*Inits`.
* field `prologueInfo` is accessed only in
   * constructor and copy methods, harmless as we are only propagating this thing
   * `isPotentiallyAssigned(FieldBinding)` -- here locals are clearly irrelevant
   * `hasInits()` -- another piece of DEAD CODE throughout the FlowInfo hierarchy!! @srikanth-sankaran do you want to remove it right now?

srikanth-sankaran:
> > Constructors that chain via this (…) are excluded from prologue analysis completely. In the actual control flow, they don’t emit field initializations at all by themselves. So, it doesn’t make sense for them to influence field initializations or for field initializations to influence them - any part of them - including epilogue.
> 
> I was wondering about the last part: of course the epilogue of a this(..)-calling constructor will see the initialization state of fields. But in this regard it actually lives against the same contracts that apply also to each regular method, since the instance is fully initialized at that point. Is that the reasoning?

I think there are two ways to flow analyze a constructor that chains to an alternate via `this(...) :

One way is to at the `this(...)` call, after arguments are analyzed, _follow the flow of control_, continue the analysis with the prologue of/from the alternate constructor and eventually getting back to the epilogue of the original starting constructor. It is guaranteed that we will get back if the program is well formed - as one or more the constructors chained to via `this(...)` will have to stop the alternate-chaining and opt for super chaining. And an ill formed program with cycles could be detected and recovered.

But such is entirely unnecessary! 

We can assume the chained-to constructor to be well formed (otherwise it would get blamed anyway) and _transfer over flow data that could be assumed to be established at the end of the chained constructor_ - which is what the current implementation is doing in `org.eclipse.jdt.internal.compiler.ast.ConstructorDeclaration.markFieldsAsInitializedAfterThisCall(ExplicitConstructorCall, FlowInfo)` 

If we followed the control flow through various constructors chaining via `this(...)` we would end up with no more, no less, no different flow data than after `org.eclipse.jdt.internal.compiler.ast.ConstructorDeclaration.markFieldsAsInitializedAfterThisCall(ExplicitConstructorCall, FlowInfo)`  _in a well formed program._ So such a rendezvous is silly.

srikanth-sankaran:
> * `hasInits()` -- another piece of DEAD CODE throughout the FlowInfo hierarchy!! @srikanth-sankaran do you want to remove it right now?

Thanks, I hadn't noticed it, I will!

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ConstructorDeclaration.java
stephan-herrmann: ASTNode bits are a scarce resource. I'm fine with the loop, as it is invoked at most twice per constructor.

If you'd like to cache something, perhaps `getConstructorCall()` would be the more worthwhile candidate?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ConstructorDeclaration.java
stephan-herrmann: :+1:

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ConstructorDeclaration.java
srikanth-sankaran: That is an interesting suggestion!

Interesting because ...

That actually in some sense restores the withdrawn field `ConstructorDeclaration.constructorCall` while still retaining ALL the benefits around de-spaghettifying the code in `ConstructorDeclaration` - the benefits that accrued via the commit 0928a9586ed60dfc6d879a94aa034ac0daa1cc3f 

@iloveeclipse ran into some code base that broke build since that commit because it was reaching into internal data of the compiler!

stephan-herrmann:
In this thread we seem to be converging but with quite different wording. Let me see ...

> > > Constructors that chain via this (…) are excluded from prologue analysis completely. In the actual control flow, they don’t emit field initializations at all by themselves. So, it doesn’t make sense for them to influence field initializations or for field initializations to influence them - any part of them - including epilogue.
> > 
> > 
> > I was wondering about the last part: of course the epilogue of a this(..)-calling constructor will see the initialization state of fields. But in this regard it actually lives against the same contracts that apply also to each regular method, since the instance is fully initialized at that point. Is that the reasoning?
> 
> I think there are two ways to flow analyze a constructor that chains to an alternate via `this(...) :
> 
> One way is to at the `this(...)` call, after arguments are analyzed, _follow the flow of control_, continue the analysis with the prologue of/from the alternate constructor and eventually getting back to the epilogue of the original starting constructor.

I was never considering this to be necessary.

> We can assume the chained-to constructor to be well formed (otherwise it would get blamed anyway) and _transfer over flow data that could be assumed to be established at the end of the chained constructor_

I think the statement could be slightly stronger even: A this(..) calling constructor in analysis mode epilogue or full should get the exact same flow info as a regular method declaration. It is not operating on a larval instance (so we could use `flowInfo.copy()` like just below). 

Bonus question: what about field access inside an argument to the chaining constructor? Now? With JEP 401?

srikanth-sankaran:
> > I think there are two ways to flow analyze a constructor that chains to an alternate via `this(...) :
> > One way is to at the `this(...)` call, after arguments are analyzed, _follow the flow of control_, continue the analysis with the prologue of/from the alternate constructor and eventually getting back to the epilogue of the original starting constructor.
> 
> I was never considering this to be necessary.

Of course. 

I was just verbalizing the possibilities.

> 
> Bonus question: what about field access inside an argument to the chaining constructor? Now? With JEP 401?

What about it now? It is illegal as constructor chaining invocation arguments are evaluated in static context upto Java 27 -  right? 

JEP 401, relaxes it for all classes identity and value.

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ConstructorDeclaration.java
stephan-herrmann: > some code base that broke build since that commit because it was reaching into internal data of the compiler!

Let me hazard a guess: lombok?

srikanth-sankaran:
> I think the statement could be slightly stronger even: A this(..) calling constructor in analysis mode epilogue or full should get the exact same flow info as a regular method declaration. It is not operating on a larval instance (so we could use `flowInfo.copy()` like just below).

The part about epilogue of a this(...) chaining constructor not operating on a larval instance - true.

Not clear about the " or full" part though.

Above this(...) it is still in early construction context and so is working with a larval object ?

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ConstructorDeclaration.java
srikanth-sankaran: I think he mentioned lombok IIRC, yes!

stephan-herrmann:
> > Bonus question: what about field access inside an argument to the chaining constructor? Now? With JEP 401?
> 
> What about it now? It is illegal as constructor chaining invocation arguments are evaluated in static context upto Java 27 - right?
> 
> JEP 401, relaxes it for all classes identity and value.

I just wondered if one choice now would make that change for JEP 401 harder than necessary. I'm fine to postpone that discussion, though.

~Do you have a comment regarding `outerInfo.copy()` vs. `flowInfo.copy()`  in this call~ 

scratch that, didn't see your response.

stephan-herrmann:
> > I think the statement could be slightly stronger even: A this(..) calling constructor in analysis mode epilogue or full should get the exact same flow info as a regular method declaration. It is not operating on a larval instance (so we could use `flowInfo.copy()` like just below).
> 
> The part about epilogue of a this(...) chaining constructor not operating on a larval instance - true.
> 
> Not clear about the " or full" part though.
> 
> Above this(...) it is still in early construction context and so is working with a larval object ?

Sorry, I was confused about where such prologue would be analysed. So for ctors with a chaining call, its the full mode that handles the prologue, and that's where the difference to regular method lies.

Should we now be worried, that my proposed change `outerInfo.copy()` -> `flowInfo.copy()` didn't produce any regressions? :)

org.eclipse.jdt.core.compiler.batch/src/org/eclipse/jdt/internal/compiler/ast/ConstructorDeclaration.java
stephan-herrmann: Should we invite the lombok team to participate in ecj maintenance? :)

srikanth-sankaran:
> Should we now be worried, that my proposed change `outerInfo.copy()` -> `flowInfo.copy()` didn't produce any regressions? :)

Good question.

`outerInfo.copy()`  == ` flowInfo.unconditionalFieldLessCopy()` != `flowInfo.copy()`

So, what ill effects could this produce ? or a switch would produce ?

Trying to construct a scenario ...

srikanth-sankaran:
> > Should we now be worried, that my proposed change `outerInfo.copy()` -> `flowInfo.copy()` didn't produce any regressions? :)
> 
> Good question.
> 
> `outerInfo.copy()` == ` flowInfo.unconditionalFieldLessCopy()` != `flowInfo.copy()`
> 
> So, what ill effects could this produce ? or a switch would produce ?
> 
> Trying to construct a scenario ...

Try the following program against the PR v switching `outerInfo.copy()` -> `flowInfo.copy()`:

```
        public class X {
            final int f;

            X() {
                this.f = 1; // establishes an incoming field-init bit in flowInfo

                class L {
                    final int g;

                    L() {
                        this(0); // must NOT see X.f's flow bit as L.g's flow bit
                    }

                    L(int i) {
                        this.g = i;
                    }
                }

                new L();
            }

                new X();
                System.out.println("OK");
            }
        }

```

srikanth-sankaran:
> Try the following program against the PR v switching `outerInfo.copy()` -> `flowInfo.copy()`:

Pushed a regression test - establishing correctness of outerInfo.copy() call being used in invocation of a this-chaining constructor's flow analysis.

Basically the field binding id for a _local_ class ends up overlapping with an enclosing class's field binding id. Each class is responsible for initialization sanity of its fields - so it makes sense to strip it. 

See how nonstaticFieldInfo is initialized - also stripping enclosing fields's flow information in this manner:

`FlowInfo nonStaticFieldInfo = flowInfo.unconditionalFieldLessCopy();`

Ergo we can conclude PR behavior is the intended behavior.

srikanth-sankaran:
> There is some uncertainty about what is the right behavior for these:
...
>org.eclipse.jdt.internal.compiler.flow.InstanceFieldsFlowInfo.unconditionalFieldLessCopy()

The following test case pin points what the correct behavior should be: `unconditionalFieldLessCopy()` should not retain prologue fields info as it would interfere with the flow analysis of a local class from an instance initializer block of a class which has some prologue initializations: The program fails to compile with the PR, but it should and it should print `OK!`

```
			public class X {
				final int f;

				{
					class L {
						final int g;

						L() {
							this(0); // must NOT see X.f's flow bit as L.g's flow bit
						}

						L(int i) {
							this.g = i;
						}
					}

					new L();
				}

				X() {
					this.f = 1; // establishes an incoming field-init bit in flowInfo
					super();
				}

					new X();
					System.out.println("OK!");
				}
			}
```

Fix is easy and will follow shortly

srikanth-sankaran:
> I have gone through every method of `UnconditionalFlowInfo` to see which ones need override and added them:
> 
> There is some uncertainty about what is the right behavior for these:

OK, I have studied the APIs below in detail and updated the implementation.

> org.eclipse.jdt.internal.compiler.flow.InstanceFieldsFlowInfo.mergeDefiniteInitsWith(UnconditionalFlowInfo) 
> org.eclipse.jdt.internal.compiler.flow.InstanceFieldsFlowInfo.mergedWith(UnconditionalFlowInfo) 

No need to override these methods. The question was whether the super implementation is correct when `this` is `UNREACHABLE_OR_DEAD`. Once we enter fields analysis with an InstanceFieldsFlowInfo instance, the "currency" stays InstanceFieldsFlowInfo. What was claimed intuitive in the comment

```
		/* What is the right behavior when this is UNREACHABLE_OR_DEAD ?
		   Are we guaranteed `otherInits` is an IFFI - seems intuitive
	    */
```
can actually be verified by stepping through the upcoming test `SuperAfterStatementsTest.testInstanceFieldsFlowInfoMergedWith()

> org.eclipse.jdt.internal.compiler.flow.InstanceFieldsFlowInfo.discardInitializationInfo() 

Upcoming test `SuperAfterStatementsTest.testDiscardInitializationInfo()` doing the honorable thing per the API contract is harmless.

So at this point, my own questions/doubts about the contract for `InstanceFieldsFlowInfo` stand fully resolved; is only 70 LOC, easily reasonable and looks good to go!

srikanth-sankaran:
@stephan-herrmann, here is a summary of changes since our last discussion:

1. Finalize the contract, cleanup, comment, polish InstanceFieldsFlowInfo
2. More regression tests.
3. Cache the result in restored field and withdraw `ConstructorDeclaration.getConstructorCall()`
4. Include a fix for https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5410 on the basis that we would warn about the first statement of epilogue as being dead code even if we have warned inside an initializer block that preceded it in control flow, since these are lexically disjoint regions. See https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5410#issuecomment-5725906551

stephan-herrmann:
> > > Should we now be worried, that my proposed change `outerInfo.copy()` -> `flowInfo.copy()` didn't produce any regressions? :)
> > 
> > 
> > Good question.
> > `outerInfo.copy()` == ` flowInfo.unconditionalFieldLessCopy()` != `flowInfo.copy()`
> > So, what ill effects could this produce ? or a switch would produce ?
> > Trying to construct a scenario ...
> 
> Try the following program against the PR v switching `outerInfo.copy()` -> `flowInfo.copy()`: [...]

Thanks for the extra scrutiny!

So using outerInfo for constructor analysis is shown to be correct, and flowInfo would be incorrect. I could now turn around and claim that perhaps the same also applies to analysing methods (using `outerInfo.copy()` for methods creates no regressions), but then this difference is probably moot, as field assignedness is not relevant during method analysis.

srikanth-sankaran:
> but then this difference is probably moot, as field assignedness is not relevant during method analysis.

Correct.

stephan-herrmann:
For the records:

I understand that `this(..)` calling constructors are treated differently from ones calling `super(..)` (implicitly or explicitly), and I understand why this is so. No bug here.

Still this introduces what I perceive as a slight inconsistency:

A prologue in a `this(..)` calling constructor is not treated as an explicit prologue. Yes, it respects enter/leaveEarlyConstructionContext, but it will never have a `prologueInfo`. Not a bug, just to keep in mind when working on this code in the future.

There is actually an observable difference resulting from this:
```java
public class X {
	final String s;
	{
		s = ""; // ERROR
	}
	X() {
		s = "";
		super();
	}
	X(int i) {
		s = "";
		this(); // ERROR
	}
}
```

This reports "may already have been assigned" at two locations: inside the instance initializer, and against the `this()` call. If I remove the assignment in the `super()`-calling constructor, the error in the instance initializer disappears, although literally speaking the illegal assignment still happens in that initializer.

Yes, the `this()` call is the location where exactly we know a repeated assignment to happen, and yes, the other constructor doesn't have such a code location to report against (since field initializers are invoked 100% implicitly). Yet, this subtlety will likely not be clear to many users.

I don't have a better solution, and since all mandatory errors are indeed reported, and also javac reports at the same locations, I'm fine with leaving it as it is.

stephan-herrmann (approved):
I've run out of things to complain about 😄 

This is a complex area, both conceptually and thus also in the code. This PR manages to make it (a) more correct and (b) easier understandable. Congratulations. 🎉 🥂

srikanth-sankaran:
> I've run out of things to complain about 😄 
> 
> This is a complex area, both conceptually and thus also in the code. This PR manages to make it (a) more correct and (b) easier understandable. Congratulations. 🎉 🥂 

Thanks Stephan! This brought back fond memories of our collaboration from the past decade. Enjoyed every bit of it. This restructuring should serve us well for JEP 401 - I plan to involve you in full review of it - your schedule permitting. 

I should record: 

Most importantly, thanks for challenging #5383. I now realize I was too close to the problem/solution to be enamored by seductive simplicity of the core of it and not recognize it. 

It is entirely likely that I would have rushed in where angels fear to tread and your cool sane head saved the day by preventing the needless signing up for a certain amount of avoidable pain. 😊🙏

Issue #5389, opened 2026-09-10
ECJ synthesizes bridge method referencing java.util.SequencedMap when compiling -8/-source 8 on a JDK 21+ host, unlike javac


laeubi:
When compiling a class that implements `java.util.SortedMap` (without overriding `reversed()`) at `-source 8 -target 8` (or `-8`) while the compiler itself is hosted on a JDK 21+ runtime, **without** `--release`/an explicit bootclasspath restriction, ECJ 3.46.200 synthesizes an extra bridge method:

```
public java.util.SequencedMap reversed();
    aload_0
    invokeinterface java/util/SortedMap.reversed:()Ljava/util/SortedMap;
    areturn
```

`java.util.SequencedMap` does not exist before JDK 21, so any class file emitted in Java-8 format (major version 52) containing this method fails with `ClassNotFoundException`/`NoClassDefFoundError: java/util/SequencedMap` the moment `reversed()` is invoked (or at verification time, depending on the VM) on any JDK 8-20 runtime.

**Importantly, `javac` given the exact same flags on the exact same JDK 21 host does *not* synthesize this bridge.** It correctly omits it and produces a clean Java-8-compatible class. This shows the behavior is an ECJ-specific inconsistency, not merely an unavoidable consequence of compiling `-source 8` without `--release` on a newer JDK.

### Real-world impact

This was discovered via `org.apache.felix.resolver.util.OpenHashMap` (vendored in Eclipse Equinox's `org.eclipse.osgi` bundle), which implements `SortedMap` without overriding `reversed()`. Built by Tycho's `tycho-compiler-plugin` with `useJDK=BREE` (falls back to running JVM's API when no explicit `--release`/bootclasspath is set) on a JDK 21 build host, the resulting `org.eclipse.osgi` jar throws `ClassNotFoundException: java.util.SequencedMap` at runtime under JDK 17, breaking OSGi TCK test container startup. See https://github.com/eclipse-equinox/equinox/pull/295 for context.

### Standalone reproducer (no Maven/Eclipse needed)

A copy of the source / script is inline below; you can also download `org.eclipse.jdt.core.compiler.batch` from Maven Central (`org.eclipse.jdt:ecj`, matching build version) to compile MyMap.java with ECJ directly.

`MyMap.java`:
```java
import java.util.AbstractMap;
import java.util.Comparator;
import java.util.SortedMap;
import java.util.Set;

[...]

PR #5390, opened 2026-09-10
Add regression test demonstrating SequencedMap bridge leak at -source 8

laeubi:
Adds `SortedMapReversedBridgeTest` to `org.eclipse.jdt.core.tests.compiler`, demonstrating that ECJ synthesizes a bridge method referencing `java.util.SequencedMap` for classes implementing `SortedMap` (without overriding `reversed()`), when compiling at `-source/-target 8` on a JDK 21+ host without an explicit `--release`/bootclasspath restriction.

`javac` given the exact same flags on the same host does **not** synthesize this bridge, showing this is an ECJ-specific inconsistency rather than an unavoidable consequence of missing `--release`.

This test currently **fails**, intentionally demonstrating the bug for review/discussion; it is not accompanied by a compiler fix. See #5389 for a standalone, Maven-free reproducer and full root-cause analysis (discovered via `org.apache.felix.resolver.util.OpenHashMap`, vendored in Eclipse Equinox's `org.eclipse.osgi`, breaking under JDK 17 when built by Tycho on a JDK 21 host).

Related to https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5389

### DIFF
--- a/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/SortedMapReversedBridgeTest.java
+++ b/org.eclipse.jdt.core.tests.compiler/src/org/eclipse/jdt/core/tests/compiler/regression/SortedMapReversedBridgeTest.java
@@ -1,0 +1,95 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Eclipse contributors and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     See git history
+ *******************************************************************************/
+package org.eclipse.jdt.core.tests.compiler.regression;
+
+import java.io.File;
+import java.io.IOException;
+import junit.framework.Test;
+import org.eclipse.jdt.core.ToolFactory;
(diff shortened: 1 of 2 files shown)


akurtakov:
@mpalat Can this one be reviewed with priority by JDT team? Breaking old JVMs compatibility shouldn't happen by running ecj on newer JVM.

laeubi:
I have made some tests with different combinations now see https://github.com/eclipse-jdt/eclipse.jdt.core/issues/5389#issuecomment-5616087538. I'm not sure why `-bootclasspath` preferred over `--release` (it might be because of the combination was somehow illegal) - need to dig more history.

One thing I found, that at least when using java 1.8 in the JDT UI I can't enable `--release` option at all (even though compiler seem to accept that flag).

mpalat:
> @mpalat Can this one be reviewed with priority by JDT team? Breaking old JVMs compatibility shouldn't happen by running ecj on newer JVM.

@akurtakov - we can review this, but this PR has the tests and not the fix as @laeubi mentioned. So we will target this issue for 4.42 timeframe itself on priority.

> One thing I found, that at least when using java 1.8 in the JDT UI I can't enable `--release` option at all (even though compiler seem to accept that flag).

@laeubi - --release option was introduced in Java 9 via [JEP 247.](https://openjdk.org/jeps/247). using it with a 1.8 compiler should have thrown an error. And Thanks @laeubi for extracting a testcase for reproducing this issue.

mpalat:
@jarthana can you please take a look once Java 27 release activities are done?

